PR #197, opened 2014-08-13
`ManuallyDrop` type gives precise control of dtors of inline data.

huonw:
This is a replacement for zeroing, acting like `*_ T` wrt destructors (i.e. they need to be run manually).

PR #1180, opened 2015-06-29
Propose `Interior<T>` data-type, to allow moves out of the dropped value during the drop hook.

aidancully:
[rendered](https://github.com/aidancully/rfcs/blob/interior/text/0000-interior.md)

### DIFF
--- a/text/0000-interior.md
+++ b/text/0000-interior.md
@@ -7,7 +7,7 @@
 Define `Interior<T>` type and associated language item. `Interior<T>`
 is structurally identical to `T` (as in, same memory layout, same
 structure fields, same enum discriminant interpretation), but has no
-traits implemented. Define a new `DropValue` trait that takes an
+traits implemented. Define a new `DropInterior` trait that takes an
 `Interior<T>` argument (instead of `&mut T`). This will allow fields
 to be moved out of a compound structure during the drop glue. The
 drop-glue will change to directly invoke the `DropValue` hook with the
@@ -91,64 +91,156 @@ recursively by default.
 Other solutions to this issue in current-Rust require circumlocutions
 for a `Drop`-implementing compound type to control whether or how its
 embedded fields get dropped. These approaches will be discussed in the
-`Alternatives` section, below.
+**Alternatives** section, below.
 
 # Detailed design
 
 We define an `Interior<T>` type and associated `"interior"` language
-item:
+item. The `T` parameter of `Interior<T>` will be restricted to types
+that implement `Drop`, or other such traits (and this RFC defines one)
+that require the `T` value to be fully populated on clean-up. (For
+purposes of exposition, we will describe all such traits as
+"`Drop`-like traits".) We enforce this restriction by defining a
+`DropGlue` marker-type:
 
 ```rust
-#[lang="interior"]
-struct Interior<T>(T);
+// in libcore::marker:
+
+// marker trait indicates that an instance of this type must be
+// fully-populated on clean-up. For example:
+//     struct MyStruct1;
+//     impl Drop for MyStruct1 { ... }
+//     struct MyStruct2;
+//     fn print_dropglue<T: DropGlue>() {
+//         println!("has drop-glue!");
+//     }
+//     print_dropglue::<MyStruct1>(); // prints "has drop-glue!"
+//     // fails to compile:
+//     // print_dropglue::<MyStruct2>();
+#[lang="drop_glue_marker"]
+pub trait DropGlue;
 ```
+
+With the `DropGlue` marker-type defined we can restrict `Interior<T>`
+such that it is only defined against types that implement `Drop`-like
+traits:
+
+```rust
+// in libcore::interior:
+
+// Presents access to `T` type, as though it did NOT implement any
+// drop-related clean-up functions.
+#[lang="interior"]
+struct Interior<T: DropGlue>(T);
+```
+
+Since `Interior<T>` cannot have drop-glue, this restriction prevents
+construction of `Interior<Interior<T>>` types. Further, we've
+prevented construction of `Interior<T>` types when the underlying `T`
+does not implement a drop-related clean-up function.
 
 An instance of `Interior<T>` is defined to be structurally identical
 to `T`, but is an independent type. Since it is an independent type,
 it can have a unique set of `impl`s defined against it: the `impl`s
-defined against `T` are not inherited by `Interior<T>`. In particular,
-this means that `Interior<T>` does not implement `Drop`, which means
-it is (by default) valid to partially or fully deconstruct an
-`Interior<T>` instance (even where it would not be valid to
-deconstruct an instance of `T`).
+defined against `T` are not inherited by `Interior<T>`. However, it
+will always be valid to partially or fully deconstruct an
+`Interior<T>` instance (though it would not be valid to deconstruct an
+instance of `T`).
 
-We also define a new `DropValue` trait to replace the legacy `Drop`
+We also define a new `DropInterior` trait to replace the legacy `Drop`
 trait, and which takes an instance of `Interior<Self>` by value:
 
 ```rust
-#[lang="dropvalue"]
-trait DropValue {
-    fn drop_value(Interior<Self>);
+#[lang="drop_interior"]
+trait DropInterior {
+    fn drop_interior(Interior<Self>);
 }
 ```
 
-Define the following bare functions to convert between `T` and
+Define the following functions to convert between `T` and
 `Interior<T>`:
 
 ```rust
-fn into_interior<T>(t: T) -> Interior<T>;
-fn from_interior<T>(t: Interior<T>) -> T;
-fn as_interior<T>(t: &T) -> &Interior<T>;
-fn as_interior_mut<T>(t: &mut T) -> &mut Interior<T>;
-fn of_interior<T>(t: &Interior<T>) -> &T;
-fn of_interior_mut<T>(t: &mut Interior<T>) -> &mut T;
-```
-
-And then it is possible to define a compatibility `Drop` trait in
-terms of `DropValue`:
-
-```rust
-impl<T: Drop> DropValue for T {
-  fn drop_value(this: Interior<Self>) {
-    of_interior_mut(this).drop();
-  }
+impl<T> Interior<T> {
+    unsafe fn from_outer_unsafe(outer: T) -> Self;
+    fn into_outer(this: Self) -> T;
+    fn as_outer(this: &Self) -> &T;
+    fn as_outer_mut(this: &mut Self) -> &mut T;
+    fn of_outer(outer: &T) -> &Self;
+    fn of_outer_mut(outer: &mut T) -> &mut Self;
 }
 ```
 
-## Compiler-generated drop-glue.
+(Note that `from_outer_unsafe` is made unsafe to avoid a
+backwards-compatibility risk with current `Drop`. This design point
+will also be discussed under **Alternatives**.)
 
-In this design, the compiler-generated drop-glue for a type gets so
-simplified as to become nearly, or possibly entirely, non-existent.
+These functions will be used to define conversion traits against
+`Interior<T>`:
+
+```rust
+impl<T> AsRef<T> for Interior<T> { ... }
+impl<T> AsMut<T> for Interior<T> { ... }
+```
+
+(With negative bounds, a safe variant of `from_outer` could be defined
+for !Drop types, and `From<T>` could be implemented for `Interior<T:
+!Drop>`. This is discussed under **Alternatives**.)
+
+With these definitions in place, it is possible to define the legacy
+`Drop` trait in terms of `DropInterior`:
+
+```rust
+impl<T: Drop> DropInterior for T {
+    fn drop_interior(mut this: Interior<T>) {
+        Interior::as_outer_mut(&mut this).drop();
+    }
+}
+```
+
+## Structurally identical
+
+`Interior<T>` is defined to be structurally identical to `T`. There
+are several things we mean by that:
+
+* `Interior<T>` will have the same size and alignment as `T`.
+* The fields of `Interior<T>` will have the same names as the fields of
+`T` and will live at the same memory offsets within the structure.
+* If `T` is an enum, then `Interior<T>` will have the same enum
+discriminant and variants.
+* The accessability of the fields of `Interior<T>` will be identical
+to the accessability of the equivalent fields of `T`.
+
+This definition is intended to support having the conversion from `T`
+to `Interior<T>` (and vice-versa) be a no-op at run-time: the
+introduction of the `Interior<T>` type should introduce *no* run-time
+overhead.
+
+## Drop-glue functions
+
+In this design, we have eliminated the need for the drop-glue function
+for a type to be compiler-generated. Rather, we make the drop-glue
+function (called on scope-based clean-up for a `DropGlue`-value)
+explicit, with an associated lang-item:
+
+```rust
+#[lang="drop_glue_fn"]
+fn drop_glue<T: DropInterior>(arg: T) {
+    // using "unsafe" variant to allow working with `Drop` types (as
+    // opposed to `DropInterior` types that do NOT implement `Drop`).
+    DropInterior::drop_interior(unsafe { Interior::from_outer_unsafe(arg) });
+}
+```
+
+(Note that, under this proposal, all drop-hooks are invoked via
+`DropInterior`, so restricting the `T` argument to implementors of
+`DropInterior` will cover all types that require `DropGlue`.) This
+drop-glue function is also used by trait vtables (to support running
+drop-based clean-up when ownership is attached to a trait pointer, as
+in `Box<Trait>`).
+
+## Example clean-up process
+
 Consider the following types:
 
 ```rust
@@ -203,14 +295,15 @@ impl DropValue for Foo {
 }
 impl DropValue for Bar {
   fn drop_value(this: Interior<Bar>) {
+    // we allow partial moves from `this`:
+    mem::drop(this.0);
     println!("drop_bar");
-    // "this" is consumed at function exit, so the compiler
+    // the remainder of "this" is consumed at function exit, so the compiler
     // will generate the following code:
-    // DropValue::<Foo>::drop_value(into_interior(this.0));
-    // DropValue::<Foo>::drop_value(into_interior(this.1));
+    // drop_glue::<Foo>(this.1);
     // and the following will be displayed as this function is called:
-    // drop_bar
     // drop_foo
+    // drop_bar
     // drop_foo
   }
 }
@@ -222,8 +315,8 @@ impl DropValue for Baz {
     // with something like the following code:
     // match this.discriminant {
     //   Empty => (),
-    //   AmFoo(val) => DropValue::<Foo>::drop_value(into_interior(val)),
-    //   AmBar(val) => DropValue::<Bar>::drop_value(into_interior(val)),
+    //   AmFoo(val) => drop_glue::<Foo>(val),
+    //   AmBar(val) => drop_glue::<Bar>(val),
     // }
     println!("drop_baz");
     // note that the println! should be evaluated *after* `this` is
@@ -235,79 +328,138 @@ impl DropValue for Baz {
 }
 ```
 
-The drop-glue function no longer appears to be necessary: all clean-up
-now occurs in the drop-hook itself, in the same way it does for types
-that do not have drop-glue defined.
-
-## `impl DropValue for Interior<T>`
-
-We do not prohibit `impl DropValue for Interior<T>`: it should just
-work in the natural way.
-
-```rust
-struct Foo;
-impl DropValue for Interior<Foo> {
-  fn drop_value(this: Interior<Interior<Foo>>) {
-    println!("drop_interior_foo");
-  }
-}
-impl DropValue for Foo {
-  fn drop_value(this: Interior<Foo>) {
-    println!("drop_foo");
-    // `this` is cleaned up as it falls out of scope. the compiler
-    // will generate something like the following code at this point:
-    // DropValue::<Interior<Foo>>::drop_value(into_interior(this));
-  }
-}
-fn main() {
-  mem::drop(Foo);
-  // the compiler will generate something like the following code
-  // at this point:
-  // DropValue::<Foo>::drop_value(into_interior(Foo));
-}
-// the following text should be displayed on running this program:
-// drop_foo
-// drop_interior_foo
-```
-
-## Interaction with `impl Drop`
-
-The blanket implementation of `DropValue` for `Drop` types means that
-it will be impossible for a user to simultaneously define `DropValue`
-and `Drop` for a given type. On the other hand, it will be possible to
-implement `Drop` for `T` and `DropValue` for `Interior<T>`, or
-vice-versa. The drop-hooks will then be called in the natural way.
-
 ## `Deref` and `DerefMut`
 
 I also suggest implementing `Deref` and `DerefMut` for `Interior<T>`
 (where `Deref::Target` would be `T`). This will allow the drop-hook to
 call functions defined against `T`, without needing to go through
-`of_interior` or `of_interior_mut` (which would be a significant
-ergonomic problem).
+`Interior::as_outer` or `Interior::as_outer_mut` (which would be a
+significant ergonomic problem).
 
 ```rust
 impl<T> Deref for Interior<T> {
   type Target = T;
   fn deref(&self) -> &T {
-    of_interior(self)
+    Interior::as_outer(self)
   }
 }
 impl<T> DerefMut for Interior<T> {
   fn deref_mut(&mut self) -> &mut T {
-    of_interior_mut(self)
+    Interior::as_outer_mut(self)
   }
 }
 ```
 
 # Drawbacks
 
-Additional compiler complexity. Potentially awkward concept to
-document and explain. Major use is for the drop-hook, though such a
-broad-purpose concept seems like it should have other important uses.
-Drop-hook function declaration is unfortunately verbose.
+## Expanding "unsafe"
+
+`Interior::from_outer_unsafe` is not actually "unsafe" (in my
+understanding of Rust's current definition of unsafe), but it does
+pose a backwards-compatibility issue: In current Rust, there is no way
+that the destructor for a field of a container will run *unless* the
+destructor for the container has also run, and it is possible that
+there is Rust code in the wild that relies on this property being
+true. `from_outer_unsafe` changes this: using this function, it's
+possible to cause a field's destructors to be called *without* having
+called the container's destructor. Rust explicitly allows that failing
+to call a destructor is "safe" behavior, so the `from_outer_unsafe`
+behavior should be considered "safe" from a technical perspective.
+But, borrowing an example from @eefriedman:
+
+```rust
+struct MaybeCrashOnDrop { c: bool }
+impl Drop for MaybeCrashOnDrop {
+    fn drop(&mut self) {
+        if self.c {
+            unsafe { *(1 as *mut u8) = 0 }
+        }
+    }
+}
+pub struct InteriorUnsafe { m: MaybeCrashOnDrop }
+impl InteriorUnsafe {
+    pub fn new() -> InteriorUnsafe {
+        InteriorUnsafe { m: MaybeCrashOnDrop{ c: true } }
+    }
+}
+impl Drop for InteriorUnsafe {
+    fn drop(&mut self) {
+        self.m.c = false;
+    }
+}
+```
+
+The above program would *not* crash with safe code in current Rust,
+but could be made to crash, *by safe code*, if `Interior::from_outer`
+were made safe.
+
+An alternative design is discussed below under **Alternatives**, but I
+could not identify an alternative that did all of the following:
+
+* Preserved the current definition of unsafe;
+* Did not pose a backwards-compatibility risk;
+* Integrated the implementations of `Drop` and `DropInterior`.
+
+As such, this proposal requires expanding the definition of "unsafe"
+to also include "backwards compatibility risk". Or, at least, to have
+an exception made that allows the `unsafe` marker on
+`from_outer_unsafe` to mean "backwards compatibility risk", rather
+than "memory unsafe".
+
+## Standard-library complexity
+
+We've expanded the number of standard-library concepts involved in
+`Drop`-based clean-up from one (the `Drop` trait) to four (the
+`DropGlue` marker, `struct Interior<T: DropGlue>`, the `DropInterior`
+trait, and the `drop_glue` function). On the other hand, we've also
+removed some magic (the `drop_glue` functions are no longer
+compiler-generated, and the `DropGlue` marker documents when
+partial-moves are disallowed).
+
+The `Interior<T>` type does not have an obvious parallel in other
+languages, so we will have a higher documentation burden in
+explanation.
+
+## Usage complexity
+
+The `DropInterior` hook function declaration is unfortunately verbose.
 
 # Alternatives
+
+## Variants of this design
+
+### Make `from_outer` safe.
+
+As discussed above (under **Drawbacks**), the `from_outer` behavior is
+not technically unsafe, but it does pose a backwards-compatibility
+risk. We *could* ignore the backwards-compatibility issue, and call
+`from_outer` safe. Or, we could keep the current drop-trait in place,
+unmodified, define two parallel drop mechanisms in the language,
+implement negative bounds, and only define `from_outer` on `T: !Drop`.
+
+### Rely on negative bounds.
+
+There are some aspects of this design that would, in my opinion, be
+better if negative bounds were already supported:
+
+* We could define a safe `from_outer` function (in addition to the
+current `from_outer_unsafe`) that was only implemented for `T: !Drop`
+types.
+* We could use the safe `from_outer` to `impl<T> convert::From<T>
+for Interior<T>`.
+
+I don't think this proposal should block on negative bounds, but I've
+tried to make the design forward-looking, so that when/if negative
+bounds are implemented we can put these improvements in place.
+
+### Drop-glue extensions
+
+This proposal moves the drop-glue function into the language itself,
+which allows considerably more flexibility in how drop-glue can be
+implemented. In the interest of keeping this proposal bounded, I'm not
+pursuing such avenues here.
+
+## Other approaches
 
 * A field of type `U` can be replaced with an `Option<U>`, which can
 be populated with `Some()` for the lifetime of the container, and
@@ -333,9 +485,33 @@ example *should* be implementable with strictly safe code.
 [manuallydrop]: https://github.com/rust-lang/rfcs/pull/197
 [nodrop]: https://crates.io/crates/nodrop
 
-* The `*_interior` functions above could be made `impl` functions for
-`struct Interior<T>`. I chose not to do so in this RFC to avoid any
-potential for conflict with functions defined against `T`.
+* Many have argued for a [consuming destructuring][destructure]
+operation (as in `let MyStruct { field1, field2 } = mystruct`).
+Consuming destructuring, along with a `DropValue` trait which took
+`self` by value, could address many of the same use-cases as the
+mechanism proposed in this RFC. I prefer the solution proposed here
+for a few reasons:
+
+** Consuming destructuring would have the same backwards-compatibility
+concern that the `Interior::from_outer` operation does (current code
+might assume that the containing structure would be dropped before the
+fields are dropped), so consuming-destructuring would probably need to
+be limited to types that don't implement `Drop`. As a result, we would
+probably still require compiler support for two types of `Drop` trait.
+
+** Taking `self` by value during the drop-hook would *require*
+implementors to deconstruct (or otherwise consume) the `self` argument
+in the drop-hook. Failure to do so would result in infinite recursion,
+as drop calls itself when the `self` argument goes out of scope at the
+end of the function. This is fragile, and probably not something we
+want to force on users of the language.
+
+** Some types can implement `Drop`, but are not easily consumed when
+destructured. In particular, enums where no variants have non-copy
+data, or structs in which all fields are `Copy`, cannot be easily
+destructured.
+
+[destructure]: https://internals.rust-lang.org/t/pre-rfc-allow-by-value-drop/1845
 
 # Unresolved questions
 
@@ -345,4 +521,5 @@ The name can be bike-shedded. Alternatives include `Synonym<T>`,
 
 # Acknowledgments
 
-This is based on a design by @eddyb.
+This is based on a design by @eddyb. Thank you to commenters on the
+RFC PR.
--- a/text/0000-interior.md
+++ b/text/0000-interior.md
@@ -1,0 +1,348 @@
+- Start Date: 2015-06-29
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Define `Interior<T>` type and associated language item. `Interior<T>`
+is structurally identical to `T` (as in, same memory layout, same
+structure fields, same enum discriminant interpretation), but has no
+traits implemented. Define a new `DropValue` trait that takes an
+`Interior<T>` argument (instead of `&mut T`). This will allow fields
+to be moved out of a compound structure during the drop glue. The
+drop-glue will change to directly invoke the `DropValue` hook with the
+`Interior<T>` structure of the `T` structure being dropped.
+
+# Motivation
+
+Currently, the `Drop::drop` method does not take its argument by
+value. This is unfortunate, as there exist cases where it may be
+important for a type to control whether and how its embedded fields
+are destructed. Consider the `SmallVec` example from the
+[`ManuallyDrop` RFC][manuallydrop]:
+
+```rust
+struct SmallVec<T> {
+    length: uint,
+    capacity: uint,
+    pointer: *mut T,
+    inline: [T, .. 8]
+}
+```
+
+This type is defined to preferentially use the `inline` field for
+value storage, to avoid unnecessary interactions with the heap when
+only a vew values must be stored. However, whether or not the `inline`
+field is used for value storage, the destructors for all 8 elements of
+the array *will* be called, unless counter-measures are taken.
+Currently, counter-measures are taken: the `SmallVec` implementation
+explicitly sets the drop-flag for each unfilled element of the array,
+such that the drop-glue will not be invoked for that element. But as
+the drop-flag is going away this is not a viable long-term solution.
+
+As another example, consider a `MessageForwarder` type, which promises
+to send an attached message to a recipient when the forwarder is
+dropped:
+
+```rust
+struct Message;
+unsafe impl marker::Send for Message;
+struct MessageForwarder<'a> {
+  c: &'a mpsc::Sender<Message>,
+  m: Message,
+}
+impl<'a> Drop for MessageForwarder<'a> {
+  fn drop(&mut self) {
+    // currently generates compile-error
+    self.c.send(self.m);
+  }
+}
+```
+
+There is nothing memory-unsafe in what `MessageForwarder` is trying to
+do. The language ought to support the desired behavior.
+
+My personal motivation for this RFC is that it can be used by a
+[proposed linear-types facility][lineartypes] to allow wrapping a
+linear-type in an affine type (which can be useful to make linear
+types more ergonomic in some circumstances). In the linear types
+proposal, implementing `Drop` against a compound type containing a
+linear field will allow the compound container type to be treated as
+affine. In that case, the linear field must be moved out during the
+drop-hook to satisfy linearity constraints, but the current drop-hook
+API does not allow partial moves from the dropped value. The
+drop-behavior defined in this RFC *would* be forward-compatible with
+the proposed linear-types mechanism.
+
+[lineartypes]: https://github.com/rust-lang/rfcs/issues/814
+
+This RFC proposes that the drop-hook take a form of `self` by value,
+instead of by reference. I say "a form of", because we want the
+drop-hook to allow moves out of the `self` argument, but partial moves
+are disallowed for types implementing `Drop`. Also, if the argument
+were the direct `self` value, then it would usually imply that new
+drop-glue would be inserted at the end of the drop-hook, resulting in
+an infinite recursive loop at the end of the drop-hook. What this
+approach needs is a different view of the `self` argument to `drop`
+which does *not* implement `Drop`, so that partial moves would be
+allowed, and so that the drop-hook would not re-invoke itself
+recursively by default.
+
+Other solutions to this issue in current-Rust require circumlocutions
+for a `Drop`-implementing compound type to control whether or how its
+embedded fields get dropped. These approaches will be discussed in the
+`Alternatives` section, below.
+
+# Detailed design
+
+We define an `Interior<T>` type and associated `"interior"` language
+item:
+
+```rust
+#[lang="interior"]
+struct Interior<T>(T);
+```
+
+An instance of `Interior<T>` is defined to be structurally identical
+to `T`, but is an independent type. Since it is an independent type,
+it can have a unique set of `impl`s defined against it: the `impl`s
+defined against `T` are not inherited by `Interior<T>`. In particular,
+this means that `Interior<T>` does not implement `Drop`, which means
+it is (by default) valid to partially or fully deconstruct an
+`Interior<T>` instance (even where it would not be valid to
+deconstruct an instance of `T`).
+
+We also define a new `DropValue` trait to replace the legacy `Drop`
+trait, and which takes an instance of `Interior<Self>` by value:
+
+```rust
+#[lang="dropvalue"]
+trait DropValue {
+    fn drop_value(Interior<Self>);
+}
+```
+
+Define the following bare functions to convert between `T` and
+`Interior<T>`:
+
+```rust
+fn into_interior<T>(t: T) -> Interior<T>;
+fn from_interior<T>(t: Interior<T>) -> T;
+fn as_interior<T>(t: &T) -> &Interior<T>;
+fn as_interior_mut<T>(t: &mut T) -> &mut Interior<T>;
+fn of_interior<T>(t: &Interior<T>) -> &T;
+fn of_interior_mut<T>(t: &mut Interior<T>) -> &mut T;
+```
+
+And then it is possible to define a compatibility `Drop` trait in
+terms of `DropValue`:
+
+```rust
+impl<T: Drop> DropValue for T {
+  fn drop_value(this: Interior<Self>) {
+    of_interior_mut(this).drop();
+  }
+}
+```
+
+## Compiler-generated drop-glue.
+
+In this design, the compiler-generated drop-glue for a type gets so
+simplified as to become nearly, or possibly entirely, non-existent.
+Consider the following types:
+
+```rust
+struct Foo;
+impl Drop for Foo { ... }
+
+struct Bar(Foo, Foo);
+impl Drop for Bar { ... }
+
+enum Baz {
+  Empty,
+  AmFoo(Foo),
+  AmBar(Bar),
+}
+impl Drop for Baz { ... }
+```
+
+In current Rust, the drop-glue for these types will look something
+like this:
+
+```rust
+fn drop_glue_Foo(v: Foo) {
+  v.drop();
+  mem::forget(v);
+}
+fn drop_glue_Bar(v: Bar) {
+  v.drop();
+  drop_glue_Foo(v.0);
+  drop_glue_Foo(v.1);
+  mem::forget(v);
+}
+fn drop_glue_Baz(v: Baz) {
+  v.drop();
+  match v {
+    Empty => (),
+    AmFoo(val) => drop_glue_Foo(val),
+    AmBar(val) => drop_glue_Bar(val),
+  }
+  mem::forget(v);
+}
+```
+
+Under this proposal, the drop-glue for member fields is called at
+the point `this` is consumed:
+
+```rust
+impl DropValue for Foo {
+  fn drop_value(this: Interior<Foo>) {
+    println!("drop_foo");
+    // no compiler-inserted clean-up necessary
+  }
+}
+impl DropValue for Bar {
+  fn drop_value(this: Interior<Bar>) {
+    println!("drop_bar");
+    // "this" is consumed at function exit, so the compiler
+    // will generate the following code:
+    // DropValue::<Foo>::drop_value(into_interior(this.0));
+    // DropValue::<Foo>::drop_value(into_interior(this.1));
+    // and the following will be displayed as this function is called:
+    // drop_bar
+    // drop_foo
+    // drop_foo
+  }
+}
+impl DropValue for Baz {
+  fn drop_value(this: Interior<Baz>) {
+    // mem::drop() is #[inline], so the compiler will replace this
+    // call:
+    mem::drop(this);
+    // with something like the following code:
+    // match this.discriminant {
+    //   Empty => (),
+    //   AmFoo(val) => DropValue::<Foo>::drop_value(into_interior(val)),
+    //   AmBar(val) => DropValue::<Bar>::drop_value(into_interior(val)),
+    // }
+    println!("drop_baz");
+    // note that the println! should be evaluated *after* `this` is
+    // dropped. So if `this` were `AmFoo(Foo)`, then the following
+    // would be displayed:
+    // drop_foo
+    // drop_baz
+  }
+}
+```
+
+The drop-glue function no longer appears to be necessary: all clean-up
+now occurs in the drop-hook itself, in the same way it does for types
+that do not have drop-glue defined.
+
+## `impl DropValue for Interior<T>`
+
+We do not prohibit `impl DropValue for Interior<T>`: it should just
+work in the natural way.
+
+```rust
+struct Foo;
+impl DropValue for Interior<Foo> {
+  fn drop_value(this: Interior<Interior<Foo>>) {
+    println!("drop_interior_foo");
+  }
+}
+impl DropValue for Foo {
+  fn drop_value(this: Interior<Foo>) {
+    println!("drop_foo");
+    // `this` is cleaned up as it falls out of scope. the compiler
+    // will generate something like the following code at this point:
+    // DropValue::<Interior<Foo>>::drop_value(into_interior(this));
+  }
+}
+fn main() {
+  mem::drop(Foo);
+  // the compiler will generate something like the following code
+  // at this point:
+  // DropValue::<Foo>::drop_value(into_interior(Foo));
+}
+// the following text should be displayed on running this program:
+// drop_foo
+// drop_interior_foo
+```
+
+## Interaction with `impl Drop`
+
+The blanket implementation of `DropValue` for `Drop` types means that
+it will be impossible for a user to simultaneously define `DropValue`
+and `Drop` for a given type. On the other hand, it will be possible to
+implement `Drop` for `T` and `DropValue` for `Interior<T>`, or
+vice-versa. The drop-hooks will then be called in the natural way.
+
+## `Deref` and `DerefMut`
+
+I also suggest implementing `Deref` and `DerefMut` for `Interior<T>`
+(where `Deref::Target` would be `T`). This will allow the drop-hook to
+call functions defined against `T`, without needing to go through
+`of_interior` or `of_interior_mut` (which would be a significant
+ergonomic problem).
+
+```rust
+impl<T> Deref for Interior<T> {
+  type Target = T;
+  fn deref(&self) -> &T {
+    of_interior(self)
+  }
+}
+impl<T> DerefMut for Interior<T> {
+  fn deref_mut(&mut self) -> &mut T {
+    of_interior_mut(self)
+  }
+}
+```
+
+# Drawbacks
+
+Additional compiler complexity. Potentially awkward concept to
+document and explain. Major use is for the drop-hook, though such a
+broad-purpose concept seems like it should have other important uses.
+Drop-hook function declaration is unfortunately verbose.
+
+# Alternatives
+
+* A field of type `U` can be replaced with an `Option<U>`, which can
+be populated with `Some()` for the lifetime of the container, and
+updated with `None` by the `drop` call. This introduces unnecessary
+overhead, and is not self-documenting. It's not clear in the structure
+definition whether it would be valid for the field to be `None` during
+the active lifetime of the structure without comments, and the
+constraint that the value be populated during the container's active
+lifetime cannot be enforced at the language-level.
+
+* A [`ManuallyDrop`][manuallydrop] or [`NoDrop`][nodrop] wrapper could
+be placed around the field of type `U`. This approach requires unsafe
+code to explicitly enable dropping a member field (using `ptr::read`
+to move droppable data out of a `NoDrop` type). Since it is not
+actually memory unsafe to control whether or not a field's destructor
+will run (at least when the field is considered valid), it seems
+undesirable to force clients to use `unsafe` for this purpose. While
+unsafe code will always be necessary for the `SmallVec` motivating
+example (since `SmallVec` explicitly allows uninitialized fields, and
+these are necessarily unsafe to deal with), the `MessageForwarder`
+example *should* be implementable with strictly safe code.
+
+[manuallydrop]: https://github.com/rust-lang/rfcs/pull/197
+[nodrop]: https://crates.io/crates/nodrop
+
+* The `*_interior` functions above could be made `impl` functions for
+`struct Interior<T>`. I chose not to do so in this RFC to avoid any
+potential for conflict with functions defined against `T`.
+
+# Unresolved questions
+
+The name can be bike-shedded. Alternatives include `Synonym<T>`,
+`Mirror<T>`, `Structure<T>`, `Inner<T>`, `Inside<T>`, `Bare<T>`,
+`Naked<T>`, `Contents<T>`, perhaps there are others?
+
+# Acknowledgments
+
+This is based on a design by @eddyb.


text/0000-interior.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+The drop-glue function no longer appears to be necessary: all clean-up
+now occurs in the drop-hook itself, in the same way it does for types
+that do not have drop-glue defined.
+
+## `impl DropValue for Interior<T>`
+
+We do not prohibit `impl DropValue for Interior<T>`: it should just
+work in the natural way.
```
eefriedman: Is this proposing to add a special case to the orphan rules?

text/0000-interior.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+An instance of `Interior<T>` is defined to be structurally identical
+to `T`, but is an independent type. Since it is an independent type,
+it can have a unique set of `impl`s defined against it: the `impl`s
+defined against `T` are not inherited by `Interior<T>`. In particular,
+this means that `Interior<T>` does not implement `Drop`, which means
+it is (by default) valid to partially or fully deconstruct an
+`Interior<T>` instance (even where it would not be valid to
+deconstruct an instance of `T`).
```
eefriedman: You probably want to explicitly note how private fields of T are handled.

text/0000-interior.md
```diff
@@ -0,0 +1,348 @@
...
+Define the following bare functions to convert between `T` and
+`Interior<T>`:
+
+```rust
+fn into_interior<T>(t: T) -> Interior<T>;
+fn from_interior<T>(t: Interior<T>) -> T;
+fn as_interior<T>(t: &T) -> &Interior<T>;
+fn as_interior_mut<T>(t: &mut T) -> &mut Interior<T>;
+fn of_interior<T>(t: &Interior<T>) -> &T;
+fn of_interior_mut<T>(t: &mut Interior<T>) -> &mut T;
```
eefriedman: Maybe explicitly note that these functions are not `unsafe`?

eefriedman:
Might be worth explicitly describing the shim the compiler has to generate for trait vtables.  (Consider the code necessary to implement `core::intrinsics::drop_in_place::<Any>`.)

Overall, this seems like a solid proposal.

text/0000-interior.md
eefriedman: Actually, I'm not completely sure whether these need to be unsafe.  into_interior and from_interior seem suspicious.

text/0000-interior.md
aidancully: Well, I had thought `DropValue` would be defined in `libcore`, same as `Drop`. So I think this should be valid per orphan rules.

eddyb: `Drop` has special rules so you can't have partial impls for generic types:
```rust
struct Foo<T>(T);
// error: Implementations of Drop cannot be specialized [E0366]
impl Drop for Foo<()> {
    fn drop(&mut self) {}
}
```

pnkfelix:
cc me

text/0000-interior.md
aidancully: The way I see it, the safety of these functions should match the safety of `mem::forget`, which has recently been made safe. But I'll add a note to that effect in the text.

text/0000-interior.md
aidancully: They would continue to be private to the module where `T` was defined, but I can note that explicitly.

text/0000-interior.md
eefriedman: Consider an example like this:

```rust
struct MaybeCrashOnDrop { c: bool }
impl Drop for MaybeCrashOnDrop {
    fn drop(&mut self) {
        if self.c {
            unsafe { *(1 as *mut u8) = 0 }
        }
    }
}
pub struct InteriorUnsafe { m: MaybeCrashOnDrop }
impl InteriorUnsafe {
    pub fn new() -> InteriorUnsafe {
        InteriorUnsafe { m: MaybeCrashOnDrop{ c: true } }
    }
}
impl Drop for InteriorUnsafe {
    fn drop(&mut self) {
        self.m.c = false;
    }
}
```
Currently, safe code can't break safety rules using an InteriorUnsafe, but it could if `into_interior` were considered safe, I think?

aidancully: I'm not the best person to make this argument, but my understanding of Rust's current definition of "unsafe" is that it refers specifically to memory-unsafety - that is, in the ability to access invalid memory. As your example shows, `into_interior` can provide the ability to violate API invariants, in the same way that `mem::forget` can (see the `thread::scoped` discussion, https://github.com/rust-lang/rust/issues/24292, and the mem::forget discussion, https://github.com/rust-lang/rfcs/pull/1066). It is a foot-gun, just like `mem::forget` is. But it does NOT do anything memory unsafe, and it does not cause unsafe behavior (though it can, as you show, cause an invariant to be violated in such a way that it results in unsafe behavior). I'm not persuaded that these functions are unsafe.

Ericson2314:
Thanks for proposing this @aidancully, especially making it backwards compatible!

text/0000-interior.md
```diff
@@ -0,0 +1,348 @@
...
+    println!("drop_foo");
+    // no compiler-inserted clean-up necessary
+  }
+}
+impl DropValue for Bar {
+  fn drop_value(this: Interior<Bar>) {
+    println!("drop_bar");
+    // "this" is consumed at function exit, so the compiler
+    // will generate the following code:
+    // DropValue::<Foo>::drop_value(into_interior(this.0));
```
Ericson2314: Is it possible to do partial moves like this? Otherwise do:

```
let Bar(x, y) = this;
DropValue::<Foo>::drop_value(into_interior(x));
DropValue::<Foo>::drop_value(into_interior(y));
```

Ericson2314:
The point of manual drops it to override the default behavior (duh). So the final drop behavior and the default exist in a sort of self-super relationship.

The point of `Interior<T>` is to expose the default drop implementation. But in that regard, all the freely generated `Interior<...Interior<T>...>` make no sense, as there is no "extra default" implementation.

IMO it is simpler/less magical just to expose a different function which is the default `drop_value` implementation (both would take plan `T`). Ideally this would just be the default implementation of `drop_value`, if it was possible to call the default implementation of a method in the definition of it's override. The point of it will be to avoid infinite recursion with `drop_value`.

The vast majority of implementations will not need to worry about infinite recursion, as all interesting things to do in a `drop_value` implementation involve destructuring this. But it is useful in the definition of the `Drop` => `DropValue` blanket impl: `drop_value` would just call the default implementation after to avoid recursion.

text/0000-interior.md
aidancully: In the proposal, it's possible to do this so long as `DropValue` isn't defined for `Interior<Foo>`: partial moves are allowed when type doesn't implement `DropValue`. In the example, `DropValue` is implemented for `Bar`, but not for `Interior<Bar>`; so a partial move out of `Interior<Bar>` is allowed. That said, the proposal needs better discussion of partial moves.

eddyb: `Interior<T>` cannot implement `DropValue` if the same sanity checks we do for `Drop` are kept.

Ericson2314: As I said like in our discussion long below, I still prefer tying destructuring and functional update to whether the type has private fields, rather than whether any trait is implemented.

Ericson2314: On the other hand, seeing that `mem::forget is safe`, I don't know why we need to care about circumnavigating user-defined destructors at all.

text/0000-interior.md
eefriedman: `mem::forget` is specifically documented as something safe code is allowed to do, `into_interior` is not, and changing that isn't backwards-compatible.

Safe code staying safe depends on unsafe code maintaining its invariants.  One invariant that unsafe code can currently depend on is that the drop implementation for a struct runs before the drop implementations of its members.  Breaking that invariant breaks backward-compatibility: a library whose invariants can't be broken with safe code in 1.0 can have its invariants broken in 1.x.

I think you could get around the safety issue if the members of an `Interior<T>` were also of type `Interior<T>`.  Of course, that leads to ergonomic issues because you can accidentally leak members of a struct.

text/0000-interior.md
aidancully: Well, just because `mem::forget` is safe, doesn't mean we should be encouraging its use...

aidancully:
> IMO it is simpler/less magical just to expose a different function which is the default drop_value implementation (both would take plan T).

I think that would be acceptable if infinite recursion were the only concern... But it doesn't solve the problem of allowing partial moves from the value during the drop-hook: partial moves are still disallowed for types implementing `Drop`, and taking `T` by value in `drop_value` means we're still working with a type that implements `Drop`. Partial moves would still be disallowed...

I know there's been [discussion](https://internals.rust-lang.org/t/pre-rfc-allow-by-value-drop/1845) of using destructuring to allow consumption of the container type without consuming the fields, which would inhibit invocation of the drop-hook while continuing to provide access to fields. The difficulty I've had with this approach is that I haven't seen how it would work with some kinds of enums, which don't have natural destructors:

```rust
enum Foo {
  Bar,
  // the way this enum gets destructured might depend on its discriminant:
  //Baz(Box<u32>),
  // but in this case, doesn't really...
  Qux,
}
impl DropValue for Foo {
  fn drop_value(self) {
    // use `match` to find the discriminant:
    match self {
      when Bar => (),
      when Qux => (),
    }
    // "self" is still accessible? argh, must use `forget`.
    mem::forget(self)
  }
}
```

When this came up before, I [suggested](https://internals.rust-lang.org/t/pre-rfc-allow-by-value-drop/1845/12) a `match move` operation, which would be a consuming destructuring for an enum type. I'd be ok with that, but would argue against any solution that requires `forget`... `forget` can be used to deliberately circumvent an API design, whenever the API has meaning attached to `drop`, its use shouldn't be encouraged, IMO.

Ericson2314:
@aidancully Thanks, I hadn't realized the problems with Copy before. Exposing a "default method" thing would help with the boilerplate, but not with Copy before. [Nor am I willing just say a type can't be DropValue and Copy, linear types that impl Copy might perhaps be useful].

`match move` is a great solution, semantically a lot lighter-weight than `Interior<T>`. I hope we can adopt it.

eddyb:
@Ericson2314 You currently cannot impl `Copy` and `Drop` at the same time.
Abstractions have been built that assume `T: Copy` implies dropping `T` is a noop (`Cell` would be one example).
Allowing destructors on copyable types would be a breaking change.

text/0000-interior.md
```diff
@@ -0,0 +1,348 @@
...
+trait, and which takes an instance of `Interior<Self>` by value:
+
+```rust
+#[lang="dropvalue"]
+trait DropValue {
+    fn drop_value(Interior<Self>);
+}
+```
+
+Define the following bare functions to convert between `T` and
```
eddyb: Why aren't these static methods? `Interior::of(T)`, for example.

text/0000-interior.md
eddyb: I have to agree with @eefriedman here, being able to break the invariants of a type you didn't define, from safe code, is a no-no.
For some reason I thought `T -> Interior<T>` was safe, and it might be in most existing `Drop` implementations, but not all of them: some put their fields in a state that would make the fields' destructors noops, or at least not memory-unsafe.

text/0000-interior.md
aidancully: Because I wanted to avoid possibility of namespace collision. This is nonsensical, but demonstrates the problem I was trying to avoid:
```rust
struct Foo;
impl Foo {
  fn of(&self) -> &Foo {
    self
  }
}
```
If we have `impl Deref for Interior<T>` with `Target == T`, then there's an ambiguity in how `of` would be resolved, right?

eddyb: No, because resolving paths does not touch `Deref` at all.

Ericson2314:
@eddyb OK, but then we don't even need `match move`!

eefriedman:
@Ericson2314 the problem isn't with a type that implements Copy and Drop; the problem is with a type that implements Drop, but all the members are Copy.  There's also the issue that making an unmarked pattern match skip dropping a value might be a bit too easy to screw up.

Ericson2314:
Ah, I see. Well `@-patterns` seem to solve our problem http://is.gd/bbr17H . Granted they are verbose.

text/0000-interior.md
aidancully: I base the argument that `into_interior` is safe around a lot of the discussion I saw in the `mem::forget` discussion. Given that discussion, if, hypothetically, this proposal had been made and accepted before Rust 1.0 shipped, would this function have been called safe or unsafe? Considering comments like [this one](https://github.com/rust-lang/rfcs/pull/1066#issuecomment-95762252), it's hard for me to believe that `into_interior` would have been made unsafe.

I think what's being described is not so much an "unsafety" issue is actually a backwards-compatibility issue... (And, note, this same back-compat issue would also exist under the consuming-destructuring alternative.) The only consistent resolution I see is to disallow `into_interior` for `Drop` types. Which means that both `Drop` and `into_interior` need more special-case logic in the language than I had originally thought...

text/0000-interior.md
aidancully: Right, I've just read the [discussion](https://github.com/rust-lang/rust/issues/8142) around that... You're right, I don't want to change this behavior. I'll amend the RFC.

text/0000-interior.md
aidancully: OK, right - we can make these all static methods on `Interior`. I played it a little too conservative.

aidancully:
One of the reasons I haven't written up `match move` is that it didn't seem to get any interest when I suggested it, but I probably wouldn't approach it in such a narrow way, either... I'd try defining `move` as a low-precedence, right-associative operator, which would locally mask `Copy` behavior for a type, and ensure that the moved value doesn't outlive its expression scope, as in:

```rust
let x = 0u32;
let y = move x; // x is no longer accessible
fn blah(arg: u32) {}
blah(move y);
#[derive(Copy,Clone)]
struct Foo(u32, u32);
let z = Foo(0, 1);
let (a0, a1) = move z;
```

Of course this likely conflicts with other uses of `move`, especially in closure definition, and I still haven't thought about it very seriously.

eddyb:
@aidancully I really don't see the utility of `move expr`, unless only macros or syntax extensions expanded to it (and then it could be prototyped in a compiler-implemented macro).

Ericson2314:
It seems to me that the move should actually be part of the pattern, as we are trying to overrides and "optimization" where patterns that would otherwise consume something don't. Also this allows
`let move _ = something_that_gives_a_result`, which is a nice way to get around must-use.

Diggsey:
Given that `mem::forget()` is safe, is there any reason why destructuring should still be disallowed on drop types, assuming the fields are accessible?

Infinite recursion can be avoided by either:
- Have the compiler add an implicit call to `mem::forget()` at the end of the destructor if `self` is not consumed.
- If that's too magical, a lint to detect the infinite recursion would at least prevent accidental problems, and `mem::forget()` can be called manually. After all, it's a sufficiently rare occurence that I think explicitness wins out over verbosity.

eddyb:
@Diggsey ... "sufficiently rare occurrence"?
We're talking about a replacement for `Drop`, you'd want the default to do *the right thing*, always calling `mem::forget` at the end of destructors to get the correct behaviour seems like an anti-pattern.

Diggsey:
@eddyb It's rare if you continue to use old `Drop` for destructors which only need to borrow self. A `DropValue` destructor *should* consume self - if it doesn't do that via destructuring then it should call `mem::forget()`.

It seems like a fairly straightforward choice - either you special-case destructors in the compiler, or you require the programmer to manually consume `self`. Interior<T> in this proposal is just a very round-about way of doing the former.

eddyb:
@Diggsey I thought the whole point of this proposal was to allow moving out of `self` in the destructor (and also avoid infinite recursion) by having it be `Interior<Self>` and not just `Self`, as the latter would continue to have the restrictions it does today.
What am I missing?

I should also mention that current `Drop` situation is acknowledged as suboptimal.
It's not *supposed* to take `&mut self`, but we didn't have anything better.
`self: Own<Interior<T>>` is the "correct" type, although it could be a while before that can be implemented.

Ericson2314:
What exactly is a partial move? I assume that if you move a non-copy value out of a non-copy value, it consumes the original. I agree with @Diggsey that the only Drop implementation that shouldn't easily consume `self` is the Drop => DropValue adapter.

> as the latter would continue to have the restrictions it does today.

Certainly for backwards compatibility `Drop` would need to prevent all destructuring. Ideally all other destructuring could be privacy based, but that also breaks backward compatibility. It may have to be that the privacy-based destructuring is used only for types that implement DropValue, (and in the future, linear types).

eddyb:
@Ericson2314 if you want to move out of `foo` outside of the destructor, you'll need to first "peel off" the destructor with `Interior::of(foo)`.
That was my intention, at least, you seem to be talking about an alternate proposal that doesn't involve either of `Interior` or `DropValue`.

Ericson2314:
@eddyb Sorry for the confusion, I am talking about the alternative of having `drop_value` take self.

text/0000-interior.md
eefriedman: Yes, you could think of it as a backwards-compatibility issue.

Under the consuming-destructuring proposal, we would probably only allow destructuring Drop types if the type is defined in the same crate as the destructuring use, unless someone has a better idea for a rule.  A bit restrictive, but it dodges the backwards-compatibility concerns.

Diggsey:
While writing down the alternative below, I noticed l'd misunderstood part of the RFC, an important part which is mostly glossed-over:
`Internal<T>` is defined as not implementing `drop`, and yet the RFC says that `mem::drop(this: Internal<T>)` will call the drop glue for those fields of the T which have not been consumed already.

> I thought the whole point of this proposal was to allow moving out of self in the destructor (and also avoid infinite recursion) by having it be Interior<Self> and not just Self, as the latter would continue to have the restrictions it does today.
> What am I missing?

OK, let me present an alternative which has `drop_value` take self:
1) Allow partial moves out of types with destructors. Instead, the restriction is that variables which currently have one or more fields 'moved-out-of' cannot be used or even dropped, and it's a compile error for a variable to go out of scope while in an incomplete state.
2) Add `mem::destructure()`, an intrinsic whose behaviour mimics what happens when an `External<T>` from this RFC is dropped: all fields which have not been 'moved-out-of' are themselves dropped.
3) `drop_value()` has an implict call to `mem::destructure(self)` at the end.
4) (optional) Upgrade `mem::forget()` to also accept incomplete values, as a complement to `mem::destructure()`.

eddyb:
@Diggsey uh oh, all of those are pretty much *hacks*:
2. requires special-casing calls to a specific intrinsic in borrow-checking (and passing information from borrow-checking to codegen).
3. requires special and potentially surprising codegen for an otherwise regular method.
4. more special-casing intrinsics - not to mention that `std::mem` contains wrappers for intrinsics, would they have to be special-cased, too, and how would that be achieved?

I really can't see how that could be accepted without proper refinement types describing the partially-moved-out state, and those are even further out than `Own<T>`.

aidancully:
@eddyb:

> @aidancully I really don't see the utility of move <expr>, unless only macros or syntax extensions expanded to it (and then it could be prototyped in a compiler-implemented macro).

Honestly, I haven't thought deeply about it, maybe it doesn't have much utility, and maybe the way it's presented here simply doesn't work. It came from trying to find a more orthogonal way to capture `match move`'s behavior, so that `match move` wouldn't have to have special-case logic, but was not otherwise motivated.

--

@Diggsey and @Ericson2314:

>  I agree with @Diggsey that the only Drop implementation that shouldn't easily consume self is the Drop => DropValue adapter.

For what it's worth, I am not a fan of a solution that results in *two* blessed `Drop` traits. This is obviously just my opinion, but a `DropValue` that takes `Self` by value makes the language feel worse, since you have to jump through hoops to use it correctly. In the presence of such a `DropValue` trait, `Drop` feels worse, as it starts to feel like a work-around for a badly specified `DropValue` trait. (With the `DropValue` trait in the proposal, `Drop` feels like it's there for backwards-compatibility, while `DropValue` is the "correct" facility for newer code to use.)

I think I'm going to regret throwing out another half-baked idea, but this *might* address the concerns raised so far... Perhaps the idea of the drop-shell is a decent fit for an OIBIT?

```rust
#[derive(DropShell)]
struct MyType {
    // fields omitted
}

#[lang="drop_shell"]
trait DropShell {
    type Inner; // analagous to Interior<T>
    fn into_inner(self) -> Inner;
    fn from_inner(Inner) -> Self;
    fn of_inner(&Inner) -> &Self;
    fn of_inner_mut(&Inner) -> &Self;
    fn as_inner(&self) -> &Inner;
    fn as_inner_mut(&self) -> &Inner;
}

impl<T: DropShell> Deref for T::Inner {
    type Target = T;
    fn deref(&self) -> &T {
        <T as DropShell>::of_inner(self)
    }
}

trait DropValue: DropShell {
    fn drop_value(self::Inner) {}
}
```

And maybe `impl DropValue for T` could implicitly `#[derive(DropShell)]`, so that that first line above would become unnecessary...

Ericson2314:
I talked to @eddyb for a while, on irc. I think we both agree that `drop_value(self)` can be done without hacks if we allow destructuring iff all the fields are visible. There would be no difficulty using it correctly in the `DropValue` impl either as the only interesting thing to do involves moving at least one field out--which is a form of destructuring, and thus prevents infinite recursion.

@eddy points out however that in the same module but not in the `DropValue` impl, one could also destructure to avoid calling the destructure, and it could be done by accident.

eddyb:
@aidancully How is that different from `Interior<T>` other than having to keep it as a projection (`<MyType as DropShell>::Inner`) or using some compiler-internal type?

eddyb:
@Ericson2314 I agree with the point about moving at least one field, but only if you limit the uses of `DropValue` and if you keep `Drop` around, undeprecated (which is in the opposite direction of what @aidancully and I want).

There is also the issue of enums, which do not have privacy but can have destructors right now.
We could allow `Interior<Enum>` but not provide any refutable pattern-matching support (which would simplify the implementation at the cost of usability).
That or `DropValue` could be entirely unimplementable for enums except via `Drop` (which also feels suboptimal).

aidancully:
@eddyb: There are no essential differences in behavior between `Interior<T>` and `<T as DropShell>::Inner` - it's still essentially your design, this is basically a bike-shed. On the other hand, I think `DropShell` may match compiler behavior better, and (as such) be easier to explain than `Interior<T>` is. Of course this is subjective, but IMO:

* When I first saw `Interior<T>`, I immediately wondered what `Interior<Interior<T>>` might mean... If I see `<T as DropShell>::Inner`, it takes a bigger leap for me to ask about `<<T as DropShell>::Inner as DropShell>::Inner`, and it feels more reasonable for the compiler to prevent constructing such a type.

* `DropShell` sounds like something that will prevent partial moves: values are usually encased in a shell, and must be moved as a unit. `DropShell::Inner` sounds like you've unwrapped the contents, so that now you can have partial moves. It's nicely self-documenting.

* This is probably the most subjective point here, but I don't have the same urge to self-bikeshed the names...

but your mileage may vary...

aidancully:
Another iteration of the idea:

```rust
impl<T> From<T> for Interior<T> {
    fn from(this: T) -> Interior<T> { ... }
}

// "blessed" drop-glue supporting trait
trait DropValueHook {
    type Inner: From<Self>;
    fn drop_value_hook(this: Self::Inner);
}
fn drop_value_glue<T: DropValueHook>(t: T) {
    T::drop_value_hook(From::from(t));
}

// if you want to take `self` directly, you can...
trait DropSelf {
    fn drop_self(self);
}
impl<T: DropSelf> DropValueHook for T {
    type Inner = Self;
    fn drop_value_hook(this: Self) {
        this.drop_self();
    }
}

// also compatible with `Interior<T>`.
// `Interior<T>` is still desired, since we still haven't solved the partial-move-during-drop
// problem without it (and I still prefer `Interior` over requiring consuming destructuring,
// for reasons stated here: https://github.com/rust-lang/rfcs/pull/1180/#issuecomment-118587363).
trait DropInterior {
    fn drop_interior(this: Interior<Self>);
}
impl<T: DropInterior> DropValueHook for T {
    type Inner = Interior<Self>;
    fn drop_value_hook(this: Interior<Self>) {
        DropInterior::drop_interior(this);
    }
}
```

This mechanism would also allow users to define their own drop-hook conversions, if they wanted to, which is the possible advantage over the other mechanisms so far discussed.

We still have a backwards-compatibility issue that prevents existing `Drop` from being defined in terms of `DropValueHook` and friends... That could be addressed if we expanded the definition of "unsafe" to also include "backwards-compatibility hazard"... which isn't an argument I wanted to have, here, but I guess it'll have to go into the RFC when I get a chance to update it.

aidancully:
The new version of this RFC makes `Interior::from_outer` unsafe. I think this violates Rust's definition of "unsafe", but I can't find an alternative that I find to be as technically sound that preserves this definition... So I started a thread on [internals](https://internals.rust-lang.org/t/working-around-back-compat-issues-and-safety/2346) to discuss how to handle this abuse of "unsafe"...

Ericson2314:
Honestly, if we go with `Interior<Self>` I'm fine just forcing that and not providing another way to use plain `Self`. It's not *that* big of a deal :)

One thing I would like to think about is how exactly this is compatible with linear types down the road. We have a nice `Copy: Clone` relationship now, what relationships can we enforce beween `Affine`, `DropGlue`, and `DropValue`?

aidancully:
@Ericson2314, I agree we should think about these relationships, but I don't think most of that needs to go into this RFC? At any rate, I think: `DropValue: DropGlue`, `! Affine` implies `! DropGlue` (I think partial moves would always be allowed from linear types, limited by field privacy), but no other relationships need to be defined. Some `Affine` types will be `DropGlue`, others will not. (I should also mention, the current version of the proposal does not strictly require a `DropGlue`: it could use `DropValue` to mean the same thing. I thought `DropGlue` seemed like a better separation of concerns, but I could be convinced to remove the marker trait.)

I think this works out, since linear types never have scope-based clean-up. But I need to update the linear-types proposal... I had a `Finalize` trait in there that I plan to eliminate.

Ericson2314:
I don't think it needs to be a part of this RFC either, just that discussing it in the comments is important to make sure we come up with a good plan.

On thing to consider is that an affine type can have a linear type as a field if the a `drop_value` to clean up that manual field. So there are "two routes" to affinity, manual dropping or automatic dropping. This leads me to think we should have:
 - `AutomaticCleanup` type allows partial moves,and the default cleanup. This is somewhat analogous to `Copy`, and probably an OIBIT.
 - `DropValue: !AutomaticCleanup` (as we get closer to bike shedding time, I'd prefer the name `Destroy`)
 - `impl<T: AutomaticCleanup> Affine for T`
 - `impl<T: DropValue> Affine for T`

This keeps it so all traits add functionality, not remove it.

aidancully:
@Ericson2314 The way I see it, it's the `Drop`-hook that's actually the added behavior, rather than allowing partial moves out: Partial moves are already disallowed before a value is passed to a function, and what `Drop` does is makes sure that the cleaned-up value gets passed to a function when it leaves scope. That the value is being passed to a cleanup function is why partial moves are disallowed on `Drop` types, and asking the compiler to pass the value to a user-defined clean-up function on scope-exit is the new behavior added by `Drop`-like traits. `DropGlue` (in the RFC) is the marker trait for querying whether the compiler will insert such a call on scope-based clean-up.

On the other hand, I think linear types will also have "automatic cleanup" - it's just that automatic cleanup on linear types can't be triggered by scope-destruction, it can only be triggered by unwinding. I *think* that `AutomaticCleanup`, in this reading, is identical to `! Copy`.

As far as how this plays with the `Affine` trait, I think `Affine` should be a built-in trait, fully determined by the compiler (not opt-in), kind of like `Sized`. The compiler will call a type `Affine` if it is not copy, and either none of its fields are linear, *or* the type implements `DropGlue`. If we tie that idea into the design in this RFC, well, `Interior<T>` doesn't implement `DropGlue`, so if `T` has a linear field, and `DropGlue` is implemented for `T`, then `T` will be `Affine`, and `Interior<T>` will be linear.

Ericson2314:
@aidancully So first of all, sorry for calling it `Affine`. This trait is simply supposed to mean the type respects contraction, either "first class" contraction via automatic cleanup or "second class" contraction via the drop hook. In this manner it is the dual of `Clone`, which likewise represents "second class weakening".

`AutomaticCleanup` basically is `!DropGlue`. I think about it in terms of cleanup because the restrictions needed to make sure the drop hook can be used are also the restrictions needed to make the type not respect contraction in the absence of a drop hook. So a type `T: !Copy + !AutomaticCleanup` s is affine, and `T: !Copy + !AutomaticCleanup + DropValue` is linear. It looks like you were on to the same thing when you said "The compiler will call a type Affine if it is not copy, and either none of its fields are linear, or the type implements DropGlue."

> On the other hand, I think linear types will also have "automatic cleanup" - it's just that automatic cleanup on linear types can't be triggered by scope-destruction, it can only be triggered by unwinding.

@eddyb was thinking we can just err if a landing pad is actually generated. This would be useful as in many cases the linear type can be owned by a type with a drop hook. Hopefully that proves adequate.

aidancully:
@Ericson2314

> @eddyb was thinking we can just err if a landing pad is actually generated. This would be useful as in many cases the linear type can be owned by a type with a drop hook. Hopefully that proves adequate.

I'll answer this point first, since I think it means we may be coming at things from different perspectives. I would like to enable patterns like:

```rust
let linearguard = MakeLinear::new(mutex.lock().unwrap());
do_stuff();
linearguard.consume();
```

where, if `do_stuff` generates a panic, the `MutexGuard` embedded in `linearguard` can still have its drop-hook called on the unwinding, and the mutex could still be released. The point of `MakeLinear`, in this example, is that it allows someone to tell the compiler "I don't want this Mutex to be implicitly released", which you might want to do if you've put a lot of thought into the location where it should dropped. But you'd probably still like the mutex released on an unwind...

> So first of all, sorry for calling it Affine. This trait is simply supposed to mean the type respects contraction, either "first class" contraction via automatic cleanup or "second class" contraction via the drop hook. In this manner it is the dual of Clone, which likewise represents "second class weakening".

No worries about calling it `Affine` - I think the last draft I posted of the linear types RFC also called the trait `Affine`, and I think we have the same understanding of its meaning (though I think you have the terms "contraction" and "weakening" swapped? When I quote you below, I'll swap the terms, to try to preserve meaning).

> `AutomaticCleanup` basically is `!DropGlue`.

So, there are cases that weren't obvious to me when I started writing this that have become clearer in the process... In particular, the trait-object vtable contains an entry for the cleanup function for the underlying type - whether or not the type implements `Drop`. And I imagine this will still be the case when stack-based dynamic drop fully lands (since it will still be necessary to support `Box<Trait>`, and I don't see another mechanism for that). In other words, even when `Drop` isn't implemented, there'll still be some need for automatic clean-up functions for many (possibly most) `Affine` types...

> I think about it in terms of cleanup because the restrictions needed to make sure the drop hook can be used are also the restrictions needed to make the type not respect [weakening] in the absence of a drop hook.

Those are also, I think, the same restrictions needed to allow a value to be passed to *any* function, whether that's a clean-up function or not. The restrictions are not about clean-up, per-se, but about the integrity of a value of compound type as we try to use the value...

Ericson2314:
First of all, I realized we don't even need `DropValue` and can put `drop_value` directly in `Affine`. Through `impl<T: AutomaticCleanup> Affine for T` alone, we can ensure that it is not possible to write a custom drop hook if a type unless one opts out of `AutomaticCleanup`. As a backwards-compatible hack, we can just make legacy `Drop` opt one out of `AutomaticCleanup`. I should check if/how this is achieved with `Copy` and `Clone`.

@aidancully 
> ...unwinding...

I just worry that not in all cases will a "backup" drop hook be feasable. For example, one use-case I was telling @eddyb about is that we can phantom-type heap pointers with their allocator, and then reference the allocator just once at the root of the data structure owning the boxes, as opposed to with each box.

If we do want to go that route, in addition to what I said above, we can require all types implement `Affine` and a separate trait can indicate whether it should be implicitly used outside of landing pads, i.e. whether the type is linear, or merely has a drop hook. [With this plan, `Affine` would especially need a new name as truly everything would implement it.]

> I think you have the terms "contraction" and "weakening" swapped?

Oops, I most certainly do. Sorry about that.

> In particular, the trait-object vtable contains an entry for the cleanup function for the underlying type - whether or not the type implements `Drop`.

I think this actually will be less ad-hoc. `Trait` = `Trait + Affine`. With abstract types we don't need to worry about partial moves, and `drop_value` on `Affine` is exactly the additional entry we need in our vtable. `Trait + ?Affine` chucks that entry, and thus we are stuck with an object we must clean up manually. `Trait + AutomaticCleanup + ?Affine` is the awkward case, since `AutomaticCleanup` is shallow and thus we might need to call drop on fields. But if we can arrange it so that `AutomaticCleanup` is both a subtrait of `Affine`, and forces the `Affine` instance to just trigger the automatic cleanup, then we will get the `drop_value` entry in the vtable we need. [The extra `?Affine` will have no effect, and should trigger the warning we already have for superfluous bounds.]

In fact, I think this reformulation of objects alone can be used to justify the omnipresence of `Affine` (equivalently all types implement `DropValue`, since the traits are merged). So if you like that, we can put it in this RFC without entangling it with linear types down the road.

aidancully:
@Ericson2314 

> First of all, I realized we don't even need `DropValue` and can put `drop_value` directly in `Affine`.

I'm trying to understand this, it goes against my intuition about what `DropValue` and `Affine` would mean. Take this example:

```rust
struct Foo(Box<i32>);
```

`Foo` is affine in the type-theoretic sense: you don't need to use the value for the program to type-check. On the other hand, partial moves from `Foo` are allowed, because no user-defined drop-hook is implemented. If a `drop_value` function *were* implemented against `Foo`, then it would presumably rely on `Foo` being fully populated as the function is invoked, right? I'm sorry to get so verbose, but I am confused, and I find being explicit helpful... Say we had the following code:

```rust
struct Foo(Box<i32>);
fn f1() {
  let x = Foo(Box::new(0));
  mem::drop(x.0);
}
fn f2() {
  let _ = Foo(Box::new(0));
}
```

In your view, would the compiler insert a call to `Foo::drop_value` in either `f1` or `f2`?

> I just worry that not in all cases will a "backup" drop hook be feasable. For example, one use-case I was telling @eddyb about is that we can phantom-type heap pointers with their allocator, and then reference the allocator just once at the root of the data structure owning the boxes, as opposed to with each box.

It sounds like you might be talking about intrusive structures? For what it's worth, I agree that there will be scenarios when we can't define a "correct" backup drop-hook (and intrusive structures will cause a lot of those scenarios). On the other hand, if the alternative proposal is to generate an error when we hit a landing pad that tries to clean up after a linear type, well, it should be easy for a programmer to opt-in to the "backup drop-hook" for that linear type to generate that same error. A backup drop-hook seems strictly more powerful, to me, than a language-level decision to error when we are asked to unwind past a value of linear type...

Ericson2314:
I'm going to call `Affine` `Destroy` now, because in neither of these proposals is it always a drop hook, and in the second it doesn't even correspond to "second class weakening".

> `struct Foo(Box<i32>);`

The key is that when `AutomatedCleanup` is implemented, `destroy` is defined to do nothing more than the normal automatic cleanup. That means the compiler is free to *not* add it as a hook for automatically cleaned up types without changing program behavior. And in the case of partial moves, it cannot call it. That is very different way of thinking that the current one, but it isn't more complex. The new algorithm for the compiler is:

1. If the type implements `AutomaticCleanup`, automatically clean it up (partial moves allowed, etc).
2. If the type doesn't implement `AutomaticCleanup` but does implement `Destroy`, it must be cleaned up with `destroy`.
3. If the type doesn't implement `Destroy` it must be destroyed manually; no more weakening.

So `Foo`'s `drop_value` need never be called in that example.

To accommodate backup dropping for linear types, we can add a new trait `DropHook: !AutomaticCleanup + Destroy` and change the rules to this:

1. If the type implements `AutomaticCleanup`, automatically clean it up (partial moves allowed, etc).
2. If the type implements `DropHook`, it must be cleaned up with `destroy`.
3. If the type implements `Destroy` alone, it must be cleaned up manually except for unwinding.
4. If the type doesn't implement `Destroy` it must be destroyed manually, and cannot be cleaned up during unwinding.

These two options become pretty indistinguishable absent of linear types, as I suppose we can always retroactively add some sort of opt-out `DropHook`.

aidancully:
(Sorry for the delayed response... In general, I don't have much time to work on this, so progress is going to be slow even in the best case.)

OK, so I think I understand your counter-proposal for the most part, now. But I still think that emphasizing `DropGlue` as the positive trait (rather than `AutomaticCleanup`) is the way to go (assuming I understand you correctly that my idea of `DropGlue` is the same as your idea of `!AutomaticCleanup`. There are two arguments I'll make: one practical, and one philosophical.

The practical argument: `impl Drop for T` implies that partial moves are disallowed for `T`. `DropGlue` indicates that partial moves are disallowed, so this relationship is easily represented in the trait system as `trait Drop: DropGlue`. (Note that this same argument applies any time there's a clean-up hook for an object. If/when Rust gets GC, and we need something similar to Java's `finalize` method, a `Finalize` trait would also require that the value be fully populated when `Finalize::finalize` is invoked, which would be representable as `trait Finalize: DropGlue`.) If we used `AutomaticCleanup` as the marker trait, then implementing `Drop` would *take away* the implemention of `AutomaticCleanup`... This is, I think, more confusing than making `DropGlue` be the positive trait.

The philosophical argument: If I understand you correctly, your preference for `AutomaticCleanup` is that you believe it is a positive trait: when `AutomaticCleanup` is implemented, it's possible to do more things with a value than when it isn't. That's true. On the other hand, `DropGlue` can also be considered a positive trait: when `DropGlue` is implemented, the compiler will insert a call to a drop-hook when resources needs to be reclaimed. That is, we're asking the compiler to do something new with the value, that it wouldn't have done in the absence of `DropGlue` being implemented. This new thing is causing an old constraint (values must be fully populated before move) to be applied when it otherwise wouldn't need to be: When `DropGlue` is implemented, the compiler moves the value to the drop-glue function as part of clean-up. The value must be fully populated to support this.

So, I'm not arguing that `AutomaticCleanup` can't be considered a positive trait... But I do think that `DropGlue` is just as arguably a positive trait, and that using `DropGlue` as the positive trait will fit more naturally in the language.

glaebhoerl:
I find I'm having a hard time following along, which bothers me a little bit as linear types are a thing I'm also interested in. (In particular I want `&out`/`&uninit`, perhaps the prototypical example of a linear type in Rust.) My best guess is that this is because I'm trying to track you all through the underbrush and the weeds, while not having a good high-level overview of the forest or even the trees. Or in other words, this RFC and discussion feels somewhat *in medias res* to me, like I'm missing a whole bunch of context which everyone else seems to already have.

So, only if it's not too much work, could someone please briefly describe the big picture view of what capabilities we (eventually) want, what constraints we are faced with, from among these which are due to the nature of the problem itself, and which are due to the requirement to preserve backwards compatibility with the language as it is, and finally, where this cornucopia of different types and traits (all apparently intended to address the same issues) which have grown up around the `Drop` trait fits in with that? (Perhaps it might also be useful to outline what an ideal design might look like if we still had the liberty of starting from a blank slate?)

aidancully:
@glaebhoerl - I know the linear types stuff is distracting, but do you find the RFC itself unclear?

Anyways, I don't have more time, but I started looking at this facility because I wanted [this part][https://github.com/aidancully/rfcs/blob/linear-trait/text/0000-linear-type.md#making-a-linear-type-droppable] of the linear types RFC (among other reasons). But I wanted this to stand on its own, too, so I'd be curious to know why or if it doesn't...

Ericson2314:
@aidancully And now it is my turn to apologize for a late response. I'll try to respond to your practical and theoretical concerns, while also giving @glaebhoerl and whoever else is reading this a better background.

First, where I think @aidancully and I agree. The RFC gives good examples of where the current drop trait causes problems, but without those, there is something disconcerting about drop taking and `&mut self`. Programming practice says destruction requires ownership, even outside Rust where "destruction" and "ownership" have only vague meaning. But `&mut` is about borrowing and always leaves the object intact for it's original owner. The `Drop` trait really about rattling off a few side effects before a value is destroyed in a predetermined fashion, not controlling the process of destruction itself.

My first plan, which I put forth on I think @aidancully's pre-RFC long ago, was very simple and elegant. `Copy` Indicates that values of the type can be copied by just copying the underlying bits. `Drop` indicates that value of the type can be destroyed by just forgetting the underline bits. `Clone` indicates the type can be copied with some assistance / effects, and `Destroy` indicates the type can be destroyed with some assistance / effects. Inline with my sentiments expressed in the previous paragraph, `Destroy` has one method `fn destroy(self)`. Naturally, `Copy: Clone`, and `Drop: Destroy`.

The biggest changed that I noticed at the time is that whereas today types containing types implementing a drop hook need have it's only drop hook, a type must only contain (the new) `Drop` fields if itself is going to implement the `Drop` trait. This would be of little practical concern however as one could derive `Destroy` just as they do `Clone` today.

Cloning and destruction would thus handled orthogonally and symmetrically. The only asymmetry would be that in the interest of preserving the current ergonomics, calls to `destroy` (the one method on `Destroy, which consumes `self` by value) were automatically inserted. The call could be injected in all cases, relying on inlining to remove them when the type as implemented `Drop`, or just in the case that the type implemented `Destroy` alone.

Also, linear types just fell naturally out of this plan: a type that doesn't implement `Destroy` (or `Drop`, since it is a sub trait), cannot be destroyed at all unless the implementer make another consuming function. It even makes sense as to think of any type that doesn't implement `Drop` as a linear type, with the special case of those that implement `Destroy` alone achieving a poor-man's affinity with the auto generated `destroy` calls.

The glaring problem with this plan is that I completely overlooked partial moves. Certainly that would rule out inserting `destroy` calls for `Drop` types, but that wasn't a requirement. But my plan would also prohibit partial moves on types that didn't have a drop hook but with having fields that did, since that type would need to derive a drop hook (`Destroy`) to call it's fields' drop hooks. Even without the burden of backward compatibility, that would be an unacceptable regression.

So, after much discussion, I arrived at the plan I presented in the last few comments. First of all, in the interest of backwards compatibility, I could not steal the name `Drop`, so, in the absence of a non-shitty name, I anointed its successor `AutomaticCleanup`. Secondly, I had to make the trait "shallow" so that instead of requiring that implementers' fields also implement `AutomaticCleanup`, require that they merely implement `Destroy`. Whereas my `Drop` represented a nice, clear concept: one can destroy simply forgetting about underlying bytes, the meaning of `AutomaticCleanup` would have to be a bit muddled: the type has no special dropping requirements of its own, just that of its fields, and thus it permits permits partial moves.

On to @aidancully's philosophical argument, the point having "positive traits" is that they demonstrate some capability that is of use of the consumer. For example, I consider bounded quantification like
```rust
for<A> fn(&A) -> Foo where A: Bar
```
 a stand-in for (in Agda-ish notation):
```agda
(A : Type) -> (proof : Bar A) -> (_ : & A) -> Foo
```

Now the point of having a drop hook is to preserve some invariant. We can imagine that the implementer of the type would like a proof saying that the destructor is always run, and use that proof to make another proving that the invariant is always maintained. Of course, the type system is nowhere near ready to handle that: but more importantly, the invariant only matters to the definer of the type. To the user of the type, the purpose of the droop hook is unknown, and it's presence is not an extra ability, but a burden.

I concede that it's totally true that the point of prohibiting partial moves is so we can ensure that the destructor is always run, not prohibiting partial moves for it's own sake. We can reconcile this with what I wrote above however. It's probably easiest to illustrate the reconciliation by supposing our anthropomorphized compiler give an error message like "The only way to forget about IDENTIFIER is to call it's `destroy` method, but I cannot insert the call to that method for you while IDENTIFIER is in an inconsistent state because [FIELD is moved out at LOCATION,]*." It's simplest to say `AutomaticCleanup` allows partial moves, but the fully-correct statement is that offers an additional way to destroy a value, and one that happens to impose no prohibition of partial moves.

A summary of that last bit:
```rust
// You are cursed with an `A` you can't get rid of.
for<A> fn(A)
```
```rust
// You have an `A` you can only get rid of with that `destroy` method
for<A: Destroy> fn(A)
```
```rust
// You have an `A` you can get rid either with `destroy` or the "normal way"
for<A: AutomaticCleanup> fn(A)
```
 
On to the practical concerns, there is a whole lot of complexity on what is automatically implemented, what is automatically implemented but with an opt-in, and what is bounded by default. Adding these traits and making them work with legacy `Drop`, all backwards compatibly, will make things even more complex, but I beleive it is better to have good traits and this awkwardness, than confuse our traits in the interest of less confusing boilerplate / confusing lack of boilerplate. Misguided traits hurts our semantics on a much deeper level, and is harder to clean up down the road with a Rust 2.0.

Ericson2314:
So `self` vs `Interior<Self>` as actually a largely orthogonal concern. When I was ignorant of partial moves, I assumed that moving out a field would consume the others right then and there. Instead of having `Drop` (or whatever the drop hook trait is) just prohibit such things, I advocated a privacy based approach where one could only move a field out (or do a functional update) if all other fields were public. In the implementation of the destructor, the very least that could be done was to move all fields out so that their destructors would be run, and through simple structural recursion everything would eventually be consumed with finite steps. Without destructuring, the destructor would diverge, but at least we could catch that with a lint (one might exist already).

The intuition was that to call the destructors of the other fields, one has to know what those fields are. The drop hook trait would offer the weaker guarantee (to the implementer) that the type must be destroyed by the drop hook, or through destructuring. It would then be up to the the implementer to "seal the leak" by making sure that at least one of the fields was private. 

Acknowledging partial moves at the very least make things weird. For example, if one moves a field out, but then moves a new value back in, should the destructor be run, or is the value now "tainted" and only it's fields' destructors be run? The role of `AutomaticCleanup` now gets even more confusing, it now means something like "by cheating and getting around the destructor requirement, nothing actually different happens, so haha you are foiled!". To use my three examples from above:

```rust
// You are cursed with an `A` you can't get rid of unless you can destructure it
for<A> fn(A)
```
```rust
// You have an `A` you can only get rid of with that `destroy` method, or destructuring.
for<A: Destroy> fn(A)
```
```rust
// You have an `A` you can get rid either with `destroy` or through destructuring, but they do the exact same thing.
for<A: AutomaticCleanup> fn(A)
```

Needless to say, there are a number of weaknesses with this approach. The simplest is that it would be hard/impossible to do backwards-compatibly. Another that @eddyb pointed out is that it is easy to shoot yourself in the foot in a scope where the private fields are visible (i.e. the module where the type is defined). I grudgingly admit that while it would be nice to avoid a magical `Interior<T>`, doing so makes so many other things weird or intractable that avoiding it is not worth it.

Ericson2314:
I feel pretty sure about my proposals so far. The one area of uncertainty I still have is whether partial moves or complete destructuring should ever be permitted on linear types. The justification that they are prohibited so that the destructor can be called falls flat on its face when there is no destructor to call.

I think that everyone agrees that absent all trait bounds, an abstract type is linear (if we have linear types at all). With my proposal, such an abstract type wouldn't so that all `Destroy` does is introduce a way to destroy things. I suppose we could have the `Destroy` trait remove the ability, and then `AutomaticCleanup` give it back, but that is rather complex. I'm not sure how @aidancully's proposal addresses as one one hand `DropHook` indicates the obligation to use a drop hook, but  on the other linear types were supposed to have a fallback drop hook for unwinding.

I think always allowing it would make for a rather "toothless" linearity as it is easy to circumvent, but forcing the using of `Interior<T>` all over the place seems annoying too. One of the few advantages of my privacy based destructuring alternative is that it does decouple those abilities from the presence of a drop hook in a (somewhat) principled manner.

text/0000-interior.md
```diff
@@ -0,0 +1,525 @@
...
+fn drop_glue<T: DropInterior>(arg: T) {
+    // using "unsafe" variant to allow working with `Drop` types (as
+    // opposed to `DropInterior` types that do NOT implement `Drop`).
+    DropInterior::drop_interior(unsafe { Interior::from_outer_unsafe(arg) });
+}
+```
+
+(Note that, under this proposal, all drop-hooks are invoked via
+`DropInterior`, so restricting the `T` argument to implementors of
+`DropInterior` will cover all types that require `DropGlue`.) This
```
rkjnsn: How would this interact with hypothetical immovable types? There are a couple of situations where having the ability to specify that a type can not be moved would be beneficial (e.g., a mutex that doesn't require an internal heap allocation, at the cost of not being able to move it around, or an offset-pointer type that can be serialized or stored in shared memory, but couldn't be moved without updating the offset). I wouldn't want to preclude immovable types in the quest for linear ones by requiring that drop glue always operated by value.

Ericson2314: Once we have linear types we can make `&in` and `&out` and then we'd want to have ` drop_pointer(&out Interior<Self>)`.

I don't think we can convince everybody we need linear types, non-movable types, changed drop, etc in one swoop, however. Baring that the best we can do is add it all incrementally and don't stabilized until they are all there.

eddyb: Don't you need the opposite, an "in"-pointer? Isn't that what used to be referred to as `&'a move T`? The latest is `Own<'a, T>`, actually, AFAIK.
There's certain design problems with an `Own<'a, T>` pointer that have resulted in it not being proposed, e.g. how do you create such a pointer to a local variable?

glaebhoerl:
Yeah, the names have gotten completely confused, I think it was in one of mahkoh's RFCs where `&out` was used to mean the opposite of what it had theretofore been. Anyways, I personally use `&out` to mean a reference to uninitialized memory which must be initialized (an "out-parameter").

text/0000-interior.md
glaebhoerl: Yeah, the names have gotten completely confused, I think it was in one of mahkoh's RFCs where `&out` was used to mean the opposite of what it had theretofore been. Anyways, I personally use `&out` to mean a reference to uninitialized memory which must be initialized (an "out-parameter").

Ericson2314: Yeah the naming is a  horrible mess. I use `&out` to mean you can move the data out and `&in` to mean you can move the data in, but I guess I am the only one.

I do think whatever the names they should be `&<something>`, rather than a lang-itemed struct, as they are quite fundamental, but we can worry about that later.

aidancully: To answer the original question, I saw this being forward-compatible with immovable types by something like how this RFC addressed backwards-compatibility with `Drop`:

```rust
trait DropPointer {
  fn drop_pointer(&in Interior<Self>);
}
impl<T: DropValue> DropPointer for T {
  fn drop_pointer(this: &in Interior<Self>) {
    T::drop_value(*this);
  }
}
```

That is, we would at some point stop having `DropInterior` be the lang-item drop-hook trait, in favor of `DropPointer`, and define a blanket implementation of `DropPointer` for `DropInterior` types. This way, the compiler will still (in principle) only need to worry about a single drop-hook lang-item.

eddyb: We can't exactly add new keywords, and for owning pointers, references are nothing more than a special case - try writing the implementation of `DerefOwn` for `Box<T>` - @glaebhoerl and I hit this last year.
I guess I should not speak only in hypotheticals, I can instead resurrect an older draft to show you what I mean.

Here it is: [`OwnIn<T, P>`, `Own<'a, T>` and `DerefOwn<P>`](https://gist.github.com/eddyb/cce96357c8083d88819a).

I chose to use "place" this time because it matches the terminology @pnkfelix used for in-place construction (and the same abstractions could be reused - shallow allocation cleanup has to be provided in @pnkfelix's design already).

rkjnsn: @aidancully 
> That is, we would at some point stop having DropInterior be the lang-item drop-hook trait, in favor of DropPointer, and define a blanket implementation of DropPointer for DropInterior types. This way, the compiler will still (in principle) only need to worry about a single drop-hook lang-item.

That makes sense, thanks.

@eddyb 

> We can't exactly add new keywords

`in` is already reserved, and the argument I heard for not reserving more keywords for the 1.0 release was that there would be a mechanism to add keywords in a backward-compatible way. I agree with @Ericson2314 that it should be `&<something>`, since they would be fundamental language constructs, which distinct compiler handling from other references. Since they would be mainly useful for function parameters, I personally prefer `&in` and `&out` for in parameters and out parameters, respectively. Even without a general mechanism for adding keywords, I believe `&out` could be introduced as a backward-compatible contextual keyword with only a single token of look-ahead.

aidancully: @eddyb, Yeah, without a better specified `Own` design, anything I write here is going to be inexact. The forward compatibility story here would rely on being able to construct `Own` pointers for any movable type, and on a `DerefMove`-type operation being defined against `Own` pointers.

eddyb: @rkjnsn You're missing the part where `Own<'a, T>` is not something by itself, but rather an alias for `OwnIn<T, Stack<'a>>`.
Without being generic over the "place" of the value, you cannot really write abstractions like `DerefOwn`.
I *guess* you could have `in(P) T` or some other construct, but it can get pretty messy trying to shove all that into custom syntax instead of having clean library types with special semantics.

Anyway, this is a pointless discussion, because the type syntax can be made to work.
The real obstacle is expression syntax for grabbing an owning pointer to a stack local.

Something coercion based would be too magical, e.g. `(*((x: T): Own<T>)): T` - if you remove the intermediary type annotation, you're left with the nonsensical `(*(x: T)): T`.

It could be done similarly to auto-ref, only on method calls, so you would need some trait with a method taking and returning `Own<Self>`, that you would call like `x.as_owning_ref()`, to get access to an `Own<T>`.
That would allow `DerefOwn` to function, and the `into_iter` method from my sketch.

However, it would be painful to pass arguments, if the callee wants them wrapped in `Own` (arguably, you would mainly want this for DSTs, where we could have sugar for passing unsized lvalues by-"value").

If you want to use `in` in the expression syntax for building an `Own<T>`, you have to somehow avoid ambiguities with the use of `in` for in-place initialization (which corresponds to out-pointers, as if this wasn't confusing enough).

rkjnsn: @eddyb What is place? Why is it needed? Here's how I envision `&in` working:

A function could have an in reference parameter as follows:

    fn consume(val: &in SomeType) { ... }

It would be called analogously to other functions taking references:

    consume(&in foo);

In the caller, the borrow checker treats this the same as moving out of `foo`. Inside `consume` the compiler would implicitly drop `val` at the end of `consume`'s scope or in the event of a panic (unless it is moved or dropped explicitly before then), the same as with an argument passed by value.

For completeness, here is how I envision `&out` operating:

Functions are defined and called analogously to `&in`:

    fn produce(val: &out SomeType) { ... }
    produce(&out foo);

Like with assignment, `foo` must either have no value yet, or be mutable. In the case of a mutable variable with an existing value, the existing value would be dropped before the call. The compiler would ensure that all code paths in `produce` assign a value to `val` (similarly to how in ensures all code paths in a function return a value, today). In the event that a panic occurs during the execution of `produce`, `produce` is responsible for dropping the contents of `val`, if necessary, and the caller assumes `foo` is in an uninitialized state. This is because `produce` knows if `val` has been initialized or not at the point the panic occurred, but the caller does not. Because this means a panic can unwind out of the caller with `foo` uninitialized, `foo` must be legal to leave uninitialized in the event of a panic. In practice, I believe this means it can be a local variable or accessed via an `&in` or `&out` reference, but not an `&mut` reference.

aidancully:
@glaebhoerl I've got a bit more time, now, and will try to address your questions and, hopefully, @Ericson2314's responses...

> So, only if it's not too much work, could someone please briefly describe the big picture view of what capabilities we (eventually) want, what constraints we are faced with, from among these which are due to the nature of the problem itself, and which are due to the requirement to preserve backwards compatibility with the language as it is, and finally, where this cornucopia of different types and traits (all apparently intended to address the same issues) which have grown up around the Drop trait fits in with that? (Perhaps it might also be useful to outline what an ideal design might look like if we still had the liberty of starting from a blank slate?)

What do we want? Well, I can't speak for a "we"... What *I* want is to be able to use Rust in a way that does not require run-time use of a general-purpose dynamic allocator (such as `malloc`/`free`). (Note I'm referring specifically to *general-purpose* dynamic allocators. I'm more comfortable with allowing special-purpose dynamic allocators: I can design them to prevent heap fragmentation, and they allow me to tightly restrict the places where they're used, so I need to worry much less about designing for allocation failure.) Linear types will make avoiding a dynamic allocator significantly easier: they'll allow me to enforce that only appropriate resource reclamation mechanisms will be used. (I've also got a strong interest in intrusive structures, which I believe improve locality of reference, improve cache utilization, and reduce dynamic allocations, and which I also believe will be safer when using linear types, see https://github.com/aidancully/containerof/blob/master/src/lib.rs#L174 .)

So, I'm approaching this from an end goal: I want linear types in the language, because I believe they are extremely helpful in addressing the domains that interest me. I'm hopeful that this effort will be successful, because I believe that the domains that interest me (high-reliability low-resource embedded systems) are ones that, I think, the language tries to address.

Because I'm approaching this from a more-or-less practical standpoint (I mean, how practical is it, after all, for me to try to change a programming language to do what I want?), I want any design I work on to have the best possible chance of incorporation. So I don't want to break backwards compatibility, and I want the design to fit as well as possible within the language, as the language is viewed by the core team (and others with influence in the project). Both of the linear types RFC I filed, as well as this RFC, are not really my design: In both cases, I tried to write up designs originally by @eddyb, because I think his designs fit the language *better* than the ones I tried to come up with on my own (though I did try to fill out his designs as best I could).

So, what capabilities do I want?

* I want compilation to fail if a value of linear type falls out of scope in an uncontrolled way.
* I want linear types to cooperate with exception-type unwinding. Rust already has a strong commitment to unwinding, and it appeared to me (from reading previous discussion on the subject) that one of the reasons Rust doesn't have linear types already is that there wasn't a good story for how linear types should interact with unwinding. Note that supporting unwindingg will mean that the linear types I propose are not "true" linear types: unwinding allows clean-up of linear types without explicit use.
* (I also want to be able to disable exception-type unwinding, at some point in the future. I don't have a good recovery path for exceptional conditions in my domain, I'd rather just stop the whole system from running on exception, then figure out why I hit that exception, then rewrite the code to not hit that exception again. In this scenario, we *would* have "true" linear types.)
* I want users to be able to put an Affine wrapper around a linear type. This would resolve a potential ergonomic problem with linear types: If a user really prefers to use affine types, they can just make the drop-hook for the affine type clean up its linear component on `Drop`. (I don't personally consider this critical, I don't use *that* much externally-written code in the sorts of projects I'm interested in. But I strongly suspect that this is important to address if linear types are going to be accepted into the language.)

So, where does this proposal fit in? Well, this RFC is a pre-condition to putting an affine wrapper around a linear type - the last point in the list above. For details, I'll refer you to the linear types RFC I linked above. As far as I'm concerned, that is the only relationship of this RFC to the linear types RFC. This RFC is supposed to be more general than just linear types... For example, I think this also addresses the same issues addressed in the [ManuallyDrop](https://github.com/rust-lang/rfcs/pull/197) RFC, and (I think, and @eddyb [seems to agree](https://github.com/rust-lang/rfcs/pull/197#issuecomment-110850383)) does so in a way that feels more in keeping with the language (this RFC allows you to use the existing `mem::forget` primitive to prevent a destructor from being run - which is the canonical way of preventing a destructor from running. `ManuallyDrop` allows you to use `ptr::read` to allow a destructor to be run, which is, AFAICT, decidedly uncommon), though at the expense  of adding a new concept (`Interior<T>`).

I've written a lot at this point, and need to call it a night, but I'll try to talk about the associated `DropGlue` trait as soon as I get a chance.

text/0000-interior.md
eddyb: @rkjnsn That is good and all, but it's incomplete.
Look at the definition of `DerefOwn` and the implementation of `DerefOwn` for `Box` and try to write it without `OwnIn` being generic over the "place" that manages the allocation of the value.
Remember, such a pointer always owns the *value*, but ownership and cleanup of the underlying memory may vary from case to case.

Your use of `in` is also ambiguous in parsing with the planned in-place construction syntax whether it's `in(place) expr` or `in place {expr}`.

At least with `in` you could try to accommodate the two together, but `out` cannot be used like that because it's not already a keyword, not even reserved.
We *do not* have a magical way to deal with unreserved keywords, except in contextual-keyword-friendly contexts (such as those where an item always starts with a keyword).
Expressions *are not* friendly to anything, really, they're quite nasty.
Simple counter-example: `let out = x; if out == &out {y} else {z}`.

aidancully:
So, to try to answer the rest of the design questions. Implementing `Drop` for a type has some strange effects on the type:

1. (In current Rust): it adds a drop-flag field to instances of the type. (In future Rust, it may result in creation of drop-flags on the stack.)
2. It prevents partial moves from values of that type.
3. It cannot be implemented on `Copy` types.

In trying to conceptualize this, it occurred to me that implementing `Drop` can be seen as creating a "shell" around the underlying type. Basically, implementing `Drop` behaves like the following actions take place (in imperative style, that I know is not valid syntax, hopefully the idea comes across):

```rust
impl Drop for MyType {
}
```

means something like:

```rust
// remember old definition of `MyType`
let InnerType = MyType;
// and decorate old definition of `MyType` in a drop-shell wrapper.
let MyType = struct {
  inner: InnerType,
  drop_flag: bool,
};
// for most intents and purposes, MyType can be treated like InnerType,
// but moves out of MyType instances are disallowed:
impl Deref for MyType {
  type Target = InnerType;
  fn deref(&self) -> &InnerType { &self.inner }
}
impl DerefMut for MyType {...}
```

That is, `MyType` behaves like an aliasing decorator for an inner type, against which `Drop` is not implemented. (The shell mechanism prevents partial moves from instances, and makes drop-glue invocation idempotent.) I sketched some of this out in a gist, [here](https://gist.github.com/aidancully/4d9decbcee5d5c0ca62a).

Now, for backwards-compatibility, for ergonomic reasons, and to try to follow the principle of parsimony and not introduce too much change to the language at once, it'd be awkward to make users define a drop-shell *and* the inner type as two distinct types. So I think of `Interior<T>` as a means of making this inner type (which does not implement `Drop`) accessible, again, at the language level. The `DropGlue` marker trait in the RFC represents whether a drop-shell is present for a type. And `Interior<T>` is only defined when `T: DropGlue` (since, otherwise, there is no "hidden" inner type.) (In a future draft of the RFC, I intend to rename the `DropGlue` marker trait to `DropShell`.)

For what it's worth, I do not think of this design as having much to do with linear types, besides that it allows my preferred mechanism of wrapping a linear type in an affine type. For linear types, my old proposal suggested that we would define an `Affine` OIBIT that would apply to all non-linear, non-Copy types. I'll refer, again, to the old proposal to see what I had in mind, most of that proposal won't have changed (though I'm [still thinking about](https://internals.rust-lang.org/t/pre-rfc-excdrop-trait-for-different-drop-glue-during-unwinding/2242) how to make an exception-path-only drop-hook work)... I still feel the discussion has been distracting for *this* RFC.

glaebhoerl:
@Ericson2314 @aidancully Thank you both, that really helps a lot. Now to re-read the discussion with these things in mind... (in the meantime, have you seen #713?)

aidancully:
@glaebhoerl Yes, I saw that recently... I found that proposal interesting and insightful, and perhaps more internally consistent than the "destructuring" alternative discussed [above](#issuecomment-118402123). But I have to agree where you say you can't think of a use-case... And I also think it goes against the spirit of what I'm trying to do in this proposal. In #713, you're integrating the current, mutating `Drop` (that disallows moves during the drop hook) further into the language, by creating another circumstance in which it would be called. In this proposal, I'm trying to allow current `Drop` to be deprecated in favor of a more powerful facility, and which addresses a real pain-point in the language today.

Ericson2314:
Yeah I'd rather have real consuming destructors than the #713 features. I do know what I'd use the former for, but not sure what I'd use the latter for.

@aidancully Ah I hadn't realized the meaning behind the name "Interior" before--seeing that definitely warms me to concept further.

> For linear types, my old proposal suggested that we would define an Affine OIBIT that would apply to all non-linear, non-Copy types.

What is the point of excluding Copy types? I'd think the thing we really care about is whether the type can be dropped, with or without a drop hook., not affinity per-se.

aidancully:
@Ericson2314 
> What is the point of excluding Copy types? I'd think the thing we really care about is whether the type can be dropped, with or without a drop hook., not affinity per-se.

You're absolutely right. When I update that RFC, I'll rename the trait `CanDrop`, and it will apply to all non-linear types.

Ericson2314:
Sweet! In that case, should it contain a method which corresponds to the drop hook included with all objects today? Granted right now OIBITs must have no items, but since it doesn't make sense to `unsafe impl` `CanDrop`, just opt out, I don't a fundamental problem.

aidancully:
@Ericson2314 Maybe? I promise I'll come back to your question when I start looking at linear types again... I don't think it affects *this* RFC, I don't intend to put a `CanDrop` in, here.

Ericson2314:
I ask because if so, we maybe just want to mark that method in the `CanDrop` trait as `#[lang="drop_glue_fn"]`, rather than introduce the top-level `drop_glue` function as proposed in this RFC.

aidancully:
I think that doesn't work, since `CanDrop` would also be defined on `Copy` types, and `Copy` types should never have drop-glue... I should update the RFC to remove the `drop_glue` routine... I think something like it is a good idea, but I'm not sure it's the right time to define it (would the `drop_glue` signature change when / if we get immovable types? how will it interact with affine and linear types?), and it isn't particularly relevant to the main thrust of `Interior<T>`. `drop_glue` can be moved into its own RFC when the time comes...

nrc:
@pnkfelix this RFC hasn't had any comments in a year, should we FCP? Could you summarise the status please?

Ericson2314:
@nrc if we got `&move` this would have no life breathed into it.

eefriedman:
Given that #1444 was accepted, this isn't very important. SmallVec can be implemented in a straightforward manner on top of a union.  And it's possible to write a wrapper type which allows zero-overhead safe consume-on-drop:

```
// This API is terrible and incomplete, but it's enough to make my point
union NoDrop<T> {
    t: T
}
pub struct ConsumeOnDrop<T, F: FnMut(T)> {
    t: NoDrop<T>
    f: F
}
impl<T, F: FnMut(T)> ConsumeOnDrop<T, F> {
    pub fn new(t: T, f: F) -> ConsumeOnDrop<T, F> {
        ConsumeOnDrop { t: NoDrop { t: t }, f: f }
    }
}
impl<T, F: FnMut(T)> Drop for ConsumeOnDrop<T, F> {
    fn drop(&mut self) {
        unsafe {
            (self.f)(ptr::read(&mut t.t));
        }
    }
}
```

Ericson2314:
@eefriedman I think moving-drop is still value because a) it accurately reflects the semantics of destruction and b) it confines all mandatory unsafety to `drop_glue`.

nikomatsakis:
**Hear ye, hear ye!** This RFC is now entering **final comment period**. The @rust-lang/lang team is inclined to **close**, since the addition of unsafe unions allows for things like the small vector use case to be handled that way, and in general there doesn't seem to be a great amount of urgency for this change.

Ericson2314:
Could we close with the understanding that it could be revisited if we get `&move` / used as justification for `&move`?

jethrogb:
> it's possible to write a wrapper type which allows zero-overhead safe consume-on-drop

For some definitions of "safe". In terms of current alternatives: I find neither the Option version nor the union version terribly ergonomic, in particular because they require `unwrap` or `unsafe` at every point of use.

Ericson2314:
My main motivation is, well, a belief that `&mut` for drop just is disgustingly incorrect---to the point where it probably makes the language harder to learn by creating a special case where the usual rules for what makes a `&mut` borrow safe don't apply.

ubsan:
@Ericson2314 implementing drop glue by yourself sucks :(

Ericson2314:
@ubsan hmm? This RFC doesn't get rid of automatic drop glue, especially that in the presence of partial moves.

jethrogb:
Is this a safe public interface?

```
pub union NoDrop<T> {
    inner: T,
}

impl<T> NoDrop<T> {
    fn into_inner(self) -> T { /*...*/ }
}

impl<T> From<T> for NoDrop<T> { /*...*/ }

impl<T> Deref for NoDrop<T> {
    type Target = T;
    //...
}

impl<T> DerefMut for NoDrop<T> { /*...*/ }
```

I think so, and I think it would fix the ergonomics issues I was talking about before.

ubsan:
@Ericson2314 Afaict, any person that implements `DropInterior` must deal with drop glue themselves. A struct that does not implement a `Drop` function may not have to, but, then, that's the point :)

Ericson2314:
@ubsan hmm `Interior<T>`'s fields aren't themselves "`Interior`'d", so I don't think that's the case. `Interior<T>` has no drop instance, but not no drop glue.

ubsan:
@Ericson2314 I guess I see it. Still seems... overcomplicated.

aidancully:
I'm out of practice with Rust, but I'm not sure why @eefriedman's [union][union] is used? Is it just to prevent the interior type from implementing `Drop`? Because I don't see that the `union` RFC actually does prevent `Drop` on fields: it generates a lint, but does not actually prevent `Drop` fields. Which is an improvement, but if the embedded type _does_ define `Drop`, then I think the `union` approach ends up calling `Drop` twice on the contained value. (Once in the consuming function, and once in the drop glue for the ConsumeOnDrop type.) Which would be bad.

[union]: https://github.com/rust-lang/rfcs/pull/1180#issuecomment-231269532

I think the `DropGlue` marker-trait from this RFC, with negative trait bounds, would make this more robust, but my test code showed this basic approach to be pretty awkward to use. (Which could well be my fault, I'm out of practice.) I made another variant that was a little nicer:

```rust
trait DropConsumer: !DropGlue {
  fn drop_consume(self);
}
struct ConsumeOnDrop<T: DropConsumer> {
  t: T,
}
impl<T: DropConsumer> ConsumeOnDrop<T> {
  fn new(t: T) -> Self {
    ConsumeOnDrop { t: t }
  }
}
impl<T: DropConsumer> Drop for ConsumeOnDrop<T> {
  fn drop(&mut self) {
    unsafe {
      ptr::read(&self.t).drop_consume()
    }
  }
}
```

I don't see this as _that_ different than the RFC... There are still two types, but instead of being `T` and `Interior<T>`, it's `ConsumeOnDrop<T>` and `T`; and it's up to the user to explicitly opt into using consuming drop, which makes it a certainty that consuming drop will never be the default drop mechanism... And I think that's a shame, because consuming drop is (IMO) more "correct". On the other hand, it's certainly true that the explicit drop-consumer is easier to understand than the weird implicitly-generated `Interior<T>` type.

Either way, I won't have time in the foreseeable future to put serious effort into this or other RFCs, so I can't object to closing it.

ubsan:
@aidancully union types don't have any drop glue.

```rust
union Forget<T> { t: T }

Forget { t: ... };
```

is equivalent to

```rust
std::mem::forget(...);
```

aidancully:
@ubsan, Thanks for pointing that out, I didn't read the union RFC carefully enough and missed that. Using `union` would allow the consume-on-drop implementation described above to work. And I think would work with the scheme I described, too, just replacing `struct ConsumeOnDrop<T: DropConsumer>` with `union ConsumeOnDrop<T: DropConsumer>`.

I think my point that consuming drop, while possible, would never become as ergonomic as `&mut` drop remains...

nikomatsakis:
> My main motivation is, well, a belief that &mut for drop just is disgustingly incorrect

I'm aware you feel this way, but I continue to think this is an exaggeration. From the point of view of the *Rust type system guarantees alone*, `&mut T` is a perfectly valid type. That is, the data is *not* owned by the drop code -- the overarching drop glue will go ahead and expect those fields to have valid types, ready to be dropped in turn -- so they are borrowed.

Certainly it's true that Drop marks a state transition, and if you consider an extended type, the extended permissions on drop are thus different from other methods that use `&mut` (the type goes from "normal" to "post-drop", basically, which means some invariants might not hold). But then again, if you have an `unsafe` fn, the same could be true. So in general I think the "flow" of these logical invariants requires richer type signatures that what the Rust itself provides.

None of this is to say that I think `&mut` is a perfect fit -- but it's also not "disgusting".

Diggsey:
In that case I think that `Drop` is not really equivalent to destructors in other languages, but is very similar to eg. `finalize` in .net. Now that drop-flags are gone, it means that to actually implement a destructor you have to extend the set of possible states for your type to be in to include "post-drop", which may come at a memory/performance/maintenance cost.

nrc:
Agree we should close - we might want to consider something like this, but I think we should re-evaluate with unions, and after considering `&move` in its own right, etc. Doesn't seem much gain in leaving this RFC open

ubsan:
@Diggsey I think it's exactly equivalent. `~T` in C++ is:

```C++
class This {
    ...
public:
    ...
    ~This() {
        ...
    }
    // not
    ~This() && {
    }
};

nikomatsakis:
Thanks everyone for the discussion. The @rust-lang/lang team has decided to **close** this RFC, as [previously proposed](https://github.com/rust-lang/rfcs/pull/1180#issuecomment-239560263). As @Ericson2314 [noted](https://github.com/rust-lang/rfcs/pull/1180#issuecomment-239561652), this proposal could be revisited in the future when we have more experience with `union` and/or other language features.

PR #1190, opened 2015-07-05
Style: function declarations

nrc:
How should we format function declarations?

### DIFF
--- a/text/0000-style-fn-decl.md
+++ b/text/0000-style-fn-decl.md
@@ -1,0 +1,274 @@
+- Feature Name: N/A
+- Start Date: 2015-07-06
+- RFC PR: 
+- Rust Issue: 
+
+# Summary
+
+How should we format function declarations?
+
+# Motivation
+
+Since the last time this was discussed, function declarations have got busier -
+we've added `where` clauses and function types have got larger and are more
+often used in generics. The style I see in the compiler is also different to
+that used in some of the libraries and has moved away from the style guide
+(which is underspecified in any case).
+
+For reference the current text in the style guide states:
+
+> For multiline function signatures, each new line should align with the
+> first parameter. Multiple parameters per line are permitted:
+> 
+> ``` rust
+> fn frobnicate(a: Bar, b: Bar,
+>               c: Bar, d: Bar)
+>               -> Bar {
+>     ...
+> }
+> 
+> fn foo<T: This,
+>        U: That>(
+>        a: Bar,
+>        b: Bar)
+>        -> Baz {
+>     ...
+> }
+> ```
+
+I believe that this does not work very well, since when scanning it is hard to
+distinguish arguments from type parameters and the return type. It also does not
+specify anything for where clauses.
+
+# Detailed design
+
+## Examples
+
+```
+// a comment
+#[an_attribute]
+#[another_attribute]
+pub unsafe extern "C" fn foo<A, B, C>(x: SomeType,
+                                      y: Vec<B>,
+                                      z: SomeOtherType<B, C>)
+    -> ReturnType<A>
+    where A: Foo + Bar,
+          B: Baz<B, C>,
+{
+```
+
+```
+// a comment
+fn foo<A, B, C>(x: SomeType,
+                y: Vec<B>,
+                z: SomeOtherType<B, C>)
+    -> ReturnType<A>
+{
+```
+
+```
+// a comment
+fn foo(x: SomeType,
+       y: Vec<B>,
+       z: SomeOtherType<B, C>)
+{
+```
+
+```
+// a comment
+pub unsafe extern "C" fn foo<A, B, C>(x: SomeType,
+                                      y: Vec<B>,
+                                      z: SomeOtherType<B, C>)
+    where A: Foo + Bar,
+          B: Baz<B, C>,
+{
+```
+
+```
+fn foo<T>(x: T, y: A) -> T where T: Bar {
+
+// alternative
+fn foo<T>(x: T, y: A) -> T
+    where T: Bar
+{
+```
+
+```
+fn foo<T>(x: T, y: A) -> T {
+```
+
+## Rules
+
+* Comments first, then attributes, one line each, then the rest of the signature on a new line;
+* if the whole signature fits on one line (excluding comments and attributes), do it;
+* if the signature requires more than one line,
+  - each argument gets its own line, argument names are aligned;
+  - the return type gets its own line, with block indent + 4;
+  - the where clause gets its own line, aligned with the return type (or block indent + 4, if no return type);
+  - each predicate in a where clause gets its own line, with the start of the predicates aligned;
+  - the opening brace gets a new line at block indent.
+
+Prefer not to split type parameters. If it is necessary, then the opening
+parenthesis and opening angle bracket should be aligned. Type parameter names
+should be aligned (but prefer one line if possible):
+
+```
+// Prefer this
+fn foo<A, B, C>
+      (x: SomeType,
+       y: Vec<B>,
+       z: SomeOtherType<B, C>)
+{
+
+// If we need more room (but prefer a where clause, if it is bounds causing the problem):
+fn foo<A,
+       B,
+       C>
+      (x: SomeType,
+       y: Vec<B>,
+       z: SomeOtherType<B, C>)
+{
+```
+
+If more space is necessary for arguments, then the opening parenthesis may be on
+a new line with block indent. Consider this only in an emergency. E.g.,
+
+```
+pub unsafe extern "C" fn foo<A, B, C>
+    (x: SomeType,
+     y: Vec<B>,
+     z: SomeOtherType<B, C>)
+    -> ReturnType<A>
+    where A: Foo + Bar,
+          B: Baz<B, C>,
+{
+```
+
+When to use inline bounds vs where clauses:
+
+* If there is a where clause, use it for all bounds (i.e., never have both inline bounds and a where clause);
+* if there is not enough space for bounds without splitting lines between type parameters or splitting arguments, use a where clause;
+* if a bound is a function type, use a where clause;
+* if all bounds are 'small', prefer inline bounds;
+* if in doubt, use a where clause.
+
+
+## Justification
+
+I hope most of these guidelines are common sense. I'll justify some of the more
+controversial aspects.
+
+My motivations (in order):
+
+* readability, especially when scanning code;
+* keep small functions small;
+* consistency;
+* avoiding change for most Rust code today.
+
+
+### Indenting return type
+
+Aligning the return type with the arguments makes it hard to scan for (one could
+argue that the argument type is just another component of the function type, but
+in practice I find myself often wanting to scan functions for the return type,
+or to quickly notice if a function returns something or not). Indenting the
+return type at block indent is hard for scanning if there is only one argument:
+
+```
+fn foo<A, B, C>(x: SomeType)
+-> ReturnType<A>
+{
+```
+
+Indenting at block indent + 4 works in all circumstances and is easy to scan for.
+
+
+### Where clauses
+
+The same arguments about indenting return types apply here. I believe
+distinguishing `->` from `where` is easy enough that having where clauses and
+return types at the same level of indent is OK.
+
+Where clause predicates can be long; aligning to block indent + 4 rather than to
+the arguments gives more space for long predicates.
+
+Starting the line with `where` makes the clause easy to scan for (as opposed by
+having it on the same line as the last argument).
+
+Aligning the type variables is tidy and consistent with arguments. Having one
+predicate per line, rather than sometimes putting the whole clause on one line
+is consistent and does not affect the principle of keeping small functions small,
+since a function with a multiple-predicate where clause is not a small function.
+Furthermore, predicates can be syntatically complex and separating them by line
+is useful for readability.
+
+
+### The opening brace
+
+Keeping the opening brace on the last line of the signature where there is a
+return type or where clause means that there is no whitespace between the
+function signature and body. E.g.,
+
+```
+fn foo<A, B, C>(x: SomeType)
+    -> ReturnType<A> {
+    let a = bar(x);
+    let b = baz(a, x);
+    b
+}
+```
+
+That is horrible for scanning (and just really ugly).
+
+For the sake of consistency, we should use the newline brace style for any
+multi-line function, even if there is no where clause or return type. I think it
+is OK to use same line brace for one line function signatures for the sake of
+keeping small functions small, but it seems that arguing for consistency here is
+also reasonable. E.g.,
+
+```
+fn foo<T>(x: T, y: A) -> T
+{
+```
+
+Note that both same-line and newline brace styles are common in most programming
+languages, although usually they are not mixed. Also note that the style we use
+for functions does not necessarily have to affect the style used for block
+expressions (e.g., the Mozilla C++ style guide requires newline braces for all
+functions and same-line braces for all block statements).
+
+
+## Questions
+
+* Should there be a terminating comma for where clauses?
+* Should the opening brace always be on a new line? What about one line signatures (the shortest examples)?
+* Should where clauses always get their own line, even for short functions (see alternatives in examples)?
+
+
+# Drawbacks
+
+These guidelines could be quite disruptive, especially the brace on a new line.
+This is alleviated somewhat by [Rustfmt](https://github.com/nrc/rustfmt). I
+propose that if this RFC is accepted, a version of Rustfmt is made available
+which only changes function signatures to the style described here. All code can
+then be formatted automatically (modulo some exceptions for macros).
+
+
+# Alternatives
+
+So many. Some discussed inline.
+
+
+# Unresolved questions
+
+Function types - I want to leave this for later, they complicate things because
+they look like function signatures themselves and tend to be very long, often
+needing multiple lines.
+
+What if we need to break the line inside a type or pattern?
+
+Specific line length. This is somewhat orthogonal and dealt with elsewhere in
+the style guide.
+
+Spaces per tab. I've assumed four in this RFC, but it is really orthogonal, and
+is dealt with elsewhere in the style guide.


text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+                                      z: SomeOtherType<B, C>)
+    where A: Foo + Bar,
+          B: Baz<B, C>,
+{
+```
+
+```
+fn foo<T>(x: T, y: A) -> T where T: Bar {
+
+// alternative
```
nagisa: This alternative is what I personally prefer whenever `fn foo<T>(x: T, y: A) -> T` fits into a single line and `where` part does not.

gsingh93:
I'm really glad we're finally talking about this, I've been wondering about this for a while.

- I don't think it should be required for where clause predicates to all be placed on their own lines. This should be preferred if the predicates are fairly complicated, but there are cases where I wouldn't want this to be a strict requirement.
- I like trailing commas for struct declarations, so I like them for where clauses as well in principle, although I don't like it as much visually. (I don't know why the trailing commas in struct declarations don't bother me as much).
- In the "emergency" case where the first argument must begin on a new line, I usually leave the `(` on the first line and begin the first argument at previous indent + 4 on the next line. This is the precedent in other languages I believe.
- I really, really, don't like having the curly brace (`{`) on a newline. I've always hated that. Can we just make the initial indent two tabs instead of one so the alignment works out? This convention would fix the issue for other types of blocks as well:

```
if foo
    && bar {
    // This wouldn't be a problem if `&& bar` was indented once more
}
```
The alternative would be to have a newline brace for both functions *and* other types of blocks, which I think more people are against.

gsingh93:
One more point. The RFC proposes that if we need more than one line for function arguments, then each argument must get it's own line. I think this rule is a bit extreme. I personally use this rule on a case by case basis. Sometimes having multiple arguments per line is fine, and avoids very long function declarations. Not to mention that as soon as my function declaration gets too long I have to do (slightly) more work to put each argument on a new line.

Tools like `rustfmt` can have a (possibly configurable?) choice on this matter, but I think in general the developer should be allowed to choose.

nrc:
@gsingh93 could you give an example of the where clause all on one line please? My feeling is that if the predicates are so simple that one line is OK, then using inline bounds is OK.

A hanging `(` would also work, however it does mean the arguments are inline with the return type and where clause, which is sub-optimal.

I think one arg per line makes function signatures much easier for the reader to parse, especially when scanning for argument. Multiple args per line (except for all on one line) is rare in Rust code today (and in other statically typed languages). I guess I don't really value compactness that highly vs readability (likewise, `{` on its own line doesn't bother me - note that blocks do not have to follow functions here [e.g.](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Control_Structures)).

eddyb:
My main issue with `where` clause alignment is that without a `{` between the head and the body, I would end up with `where` at the same indentation level as the first line inside the body.

And then I've seen @DanielKeep use the style where the `where` clauses start at the same level as the declaration, in the [test case code he sent me](https://github.com/eddyb/rust/blob/calendar-driven-development/src/test/run-pass/anon-types-example-calendar.rs#L554-L556), and it looks really good IMHO:
```rust
impl<'a, It: 'a> FormatWeeks<'a> for It
where It: Iterator, It::Item: DateIterator {
    type WeekDays = impl Iterator<Item=String> + 'a;

    fn format_weeks(self) -> Self::WeekDays {...}
}
```

There's no scanning issues, as both `impl` (in that case) and `where` are highlighted, and the `{` is still on the previous row, allowing an immediate down-right cascade into the block (`impl` in this case).

Another example (cleaned up from that same file, I guess I used mixed styles when I modified it):
```rust
    fn group_by<G, F>(self, f: F) -> GroupBy<Self, Fn0<F>>
    where F: FnMut(&Self::Item) -> G,
          G: Eq {
        GroupBy {
            it: self.peekable(),
            f: f.copyable(),
        }
    }
```

tshepang:
I like the idea of opening braces being alone on their own line:

- This would be consistent with closing braces, which are also alone in their line
- This would help with cleaner diffs, say if you change argument names

The exception is when the functions are super short, e.g. `fn add_one(x: i32) -> i32 { x + 1 }`

text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+Note that both same-line and newline brace styles are common in most programming
+languages, although usually they are not mixed. Also note that the style we use
+for functions does not necessarily have to affect the style used for block
+expressions (e.g., the Mozilla C++ style guide requires newline braces for all
+functions and same-line braces for all block statements).
+
+
+## Questions
+
+* Should there be a terminating comma for where clauses?
```
Gankro: Yes. Any multiline comma-delimited-list should place a trailing comma. However function arguments make this nasty:

```
fn foo(a: A,
       b: B,
       c: C)
{
```

vs

```
fn foo(a: A,
       b: B,
       c: C,
){
```

vs 

```
fn foo(
    a: A,
    b: B,
    c: C,
){
```

text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+* Should the opening brace always be on a new line? What about one line signatures (the shortest examples)?
```
Gankro: Plz no. I like simple functions to be compact. Although I probably take this to a perverse limit in my personal code:

```rust
#[inline] fn simple_op(input: u32) -> { input * 2 + 1 }
```

text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+* Should where clauses always get their own line, even for short functions (see alternatives in examples)?
```
Gankro: Yes, they blur into the return type too easily otherwise.

```rust
fn foo<T>(&mut self, input_a: Foo, deltarizer: Bar) -> Frob<T> where T: Clone + Ord {
```

Fundamentally any `where` that is simple enough to merit "on the same line" seems like it should fall under "shouldn't have been a where"

text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+# Alternatives
+
+So many. Some discussed inline.
+
+
+# Unresolved questions
+
+Function types - I want to leave this for later, they complicate things because
+they look like function signatures themselves and tend to be very long, often
+needing multiple lines.
```
Gankro: Also 

```rust
where for<'a> F: Fn(&'a u8, &'a u8) -> &'a u8
```

vs

```rust
where F: for<'a> Fn(&'a u8, &'a u8) -> &'a u8
```

:P

text/0000-style-fn-decl.md
cmr: vs
```
fn foo(
    a: A
    , b: B
    , c: C
){
```

cmr:
@eddyb :+1: on that style, it's the most palatable to me.

text/0000-style-fn-decl.md
nrc: My rule of thumb is that lists where the enclosing symbol follows immediately don't get trailing commas, so no in arg lists or type param lists, but yes for where clauses, fields in structs, etc.

retep998:
I favor the style being proposed by @eddyb.

BurntSushi:
I am quite partial to this style (which I don't see in the RFC---but perhaps I missed it---although I think I saw @Gankro post it):

```rust
fn foo<A, B, C>(
    a: A,
    b: B,
    c: C,
) -> Foo {
    // ...
}
```

How `where` fits into this isn't always clear, but just fishing through my code, I see this:

```rust
    pub fn send<'c: 'a, 'b: 'a, C, T, F>(
        mut self,
        chan: C,
        val: T,
        mut run: F,
    ) -> Select<'a> where C: Sender<Item=T> + Clone + 'c,
                          T: 'b,
                          F: FnMut() + 'a {
    // ...
}
```

I like my first example. I am less happy with the second.

adudney:
In my personal preference, I would like if simple function declarations (Ones that easily fit on a single line, not including body) were allowed to keep the brace at the end of the line, rather than always requiring a new line. Other than that this looks good.

This line shows the style I prefer for simple functions: https://github.com/rust-lang/rfcs/pull/1190/files#diff-b9d68e978b763d44ed0f1b4fffd666e0R97

Icabob4:
I agree with @eddyb

chris-morgan:
@eddyb’s style of `where` not indented and the curly brace not on a new line is my favoured style.

ssokolow:
I have to agree with @gsingh93 about the curly brace on a new line. It wastes precious vertical space on monitors that have only gone more widescreen historically.

(It also makes people rely more on smart IDEs to mitigate flaws in the language and I firmly believe that to be one of the contributing factors to the culture of "build and trust leaky abstractions" that I've noticed in the Java world.)

I'm obsessive enough with it that I'll fight an IDE's auto-indent to enforce it. (I think I might have even wrapped a style checker in grep-based filter script once when it refused to provide an option to shut up about that.)

The suggested double indent for block-starting continuations is also in line with what pleases `pep8.py` in Python, which would join the many names and syntactic decisions inspired by that language:

```python
if (long_thing and
        other_long_thing):
   stuff()

**Comment by rpjohnst** [comment]
I prefer continued lines' indentation not be at the mercy of the previous line. For example, instead of this:

```
fn foo(a: A,
       b: B,
       c: C) -> R where A: T,
                         B: T {
}
```

This would be nice (or use double indent like @ssokolow suggests):

```
fn foo(
    a: A,
    b: B,
    c: C
) -> R where
    A: T,
    B: T {
}
```

**Comment by reem** [comment]
I also prefer @eddyb's noted style, with the where clause on the next line and unindented.

I do not think it's necessary for each argument to get its own line, this just makes for unnecessarily tall argument lists when we have lots of horizontal space to use. In my code, I usually continue to list arguments until I run out of horizontal space, then go to the next line and continue (with the start of each line lining up with the first argument). For example:

```rust
pub fn listen_with<A>(mut self, addr: A, threads: usize,
                      protocol: Protocol) -> HttpResult<Listening>
where A: ToSocketAddrs {
```

As you can see in the example, I also do not think the return type needs its own line should it fit on the end of the last line of arguments. 

I am also in the "I hate hate hate { on a new line" club, and most rust code I've seen in the wild leaves it on the same line.

reem:
I would also just like to quietly note that all of the examples combining `pub unsafe extern "C" fn` with type parameters don't really make any sense, since extern functions almost never have type parameters, so we should avoid optimizing their readability at the cost of readability of other kinds of declarations.

sinistersnare:
All I want to say is that I think it is a bad idea to use any subjective wording in here, you claim that certain styles are 'just ugly' and 'harder to read', although that is highly dependant on the person.

nrc:
More motivation for the proposed design c.f., @eddyb's counter-proposal: I believe having only the start of functions against the left margin makes it easyier to scan for functions:

```
fn foo1<T>(argument1: Type1,
           argument1: Type1,
           argument1: Type1,
           argument1: Type1)
    where T: Foo,
          T: Bar,
{
    let fsfas;
    asfdasf
    asdfsd
    asdfsad
}

fn foo2<T>(argument1: Type1,
           argument1: Type1,
           argument1: Type1,
           argument1: Type1)
    where T: Foo,
          T: Bar,
{
    let fsfas;
    asfdasf
    asdfsd
    asdfsad
}



// vs
fn foo1<T>(argument1: Type1,
           argument1: Type1,
           argument1: Type1,
           argument1: Type1)
where T: Foo,
      T: Bar {
    let fsfas;
    asfdasf
    asdfsd
    asdfsad
}

fn foo2<T>(argument1: Type1,
           argument1: Type1,
           argument1: Type1,
           argument1: Type1)
where T: Foo,
      T: Bar {
    let fsfas;
    asfdasf
    asdfsd
    asdfsad
}

```

When I'm scanning a large file, I find the former, more spaced out layout much easier to pick functions out. The latter is harder for me to scan because the `where`s interrupt my scanning.

In general, I much prefer 'spacier' layouts to more compact ones - I'd rather scan more lines quickly, than fewer lines slowly.

reem:
I personally prefer shorter source files with shorter declarations, so it's easier to keep more things in view at once. This makes it much easier to see the structure of the code then when everything is spaced out so it can only be viewed one piece at a time.

gsingh93:
@nrc: @eddyb gave a good example of where `where` clause predicates are simple enough to be on one line:
```
impl<'a, It: 'a> FormatWeeks<'a> for It
where It: Iterator, It::Item: DateIterator {
    type WeekDays = impl Iterator<Item=String> + 'a;

    fn format_weeks(self) -> Self::WeekDays {...}
}
```
I don't think you can reference associated types in type parameters, correct?

I also like @eddyb's where clause style and have seen multiple community members already using it. Your counter proposal doesn't have me convinced, I still find it fairly easy to scan (but I think I'm biased because I *really* don't want the brace on a newline as the alternative).

And I'd like more opinions on the point @reem and I brought up. I don't think it should be necessary in all cases to place all the arguments and the return type on new lines. I also often add arguments until I run out of space and then wrap the line. In certain cases I find it cleaner to put each argument on a newline, but I do that on a function by function basis.

retep998:
I definitely prefer the style for function definitions that I use in `winapi`. For example:
```Rust
    pub fn GetVolumeInformationByHandleW(
        hFile: HANDLE, lpVolumeNameBuffer: LPWSTR, nVolumeNameSize: DWORD,
        lpVolumeSerialNumber: LPDWORD, lpMaximumComponentLength: LPDWORD,
        lpFileSystemFlags: LPDWORD, lpFileSystemNameBuffer: LPWSTR, nFileSystemNameSize: DWORD,
    ) -> BOOL;
```
It saves a ton of space thus making it quicker to scroll through and do things.
As for how it would look combined with @eddyb 's where clause style:
```Rust
fn some_function<A, B, C>(
    reallylongarg1: ReallyLongType1, reallylongarg2: ReallyLongType2,
    reallylongarg3: ReallyLongType3, reallylongarg4: ReallyLongType4,
) -> ReallyReallyReallyLongReturnType
where A: IsSomethingReallyLong, B: IsAlsoSomethingReallyLong,
      C: OkayAreYouActuallyReadingThis {
    unimplemented!()
}

**Comment by DanielKeep** [comment]
Since I've been summoned...

I don't agree with aligning arguments with the inside of the opening paren.  I fell in love with this style when I first used IDLE (the editor that ships with Python), but fell *out* of love when I couldn't find a single other editor that would do it.  It's a massive pain having to manually align all those arguments manually.  Worse, it interacts badly with changes to the rest of the signature: if you change the name, or add a generic parameter, or really do just about *anything*, suddenly you have to re-align everything.  By hand.  In practice, I'd just ignore this part of the style on the basis of being too damn fiddly.

I don't agree that the style @eddyb showed interferes too significantly with scanning; it just makes a given function "anchor" somewhat fatter.  I imagine it'd only be a problem if you were looking at a list of functions, sans bodies, all packed together with no whitespace.

Also, to expand on what I do when the `where` clause gets more complicated:

```rust
fn many_args_and_much_where<R, T, U>(
    first_arg: T,
    // ..
    last_arg: U,
) -> R
where
    T: Clone + Debug + Ord + Sized + DivisibleBySix,
    U: Sugar + Spice + Nice + Vegemite,
{
    // ...
}
```

With this, changes to a single argument or clause affects only that line, and there is *never* a line that looks like it might be part of the function body (by virtue of having an unbroken left margin in common).

I mean, yes, it squats on the page like a giant toad that's been flattened by a road train, but it really only becomes necessary on complicated function signature where readability trumps conciseness.

chris-morgan:
@DanielKeep: Vim does visual indent, incidentally.

mdinger:
@GBGamer You may have an opinion on this since trying to [discuss this](https://internals.rust-lang.org/t/multiple-line-fn-and-impl-declarations/1106) in the past.

liigo:
another +1 to @eddyb 's code style

ssokolow:
I have no problem with the style @eddyb borrowed from @DanielKeep and reccommended.

With or without double-indenting the `where` line, it's easy to scan, easy to parse, vertically compact, and keeps the `{` from wasting a line.

Ericson2314:
I have long basically done what this RFC advocated. I think with fat signatures, a line with just `{` is nice to separate the signature from the body. A long first line and a long where clause can run together really nastily. [Hopefully that is different than the "scanning" points people have brought up ready.]

ruud-v-a:
I am in favour of `{` on its own line. I find big blocks of characters hard to read. With the current style of `{` after the declaration, I often resort to adding a blank line after it for readability. I prefer readability over cramming as much as possible into one screen.

Anyhow, this is one of those things where although there are a few objective arguments for every style, personal preference dominates. (My main objection to `{` after the declaration, apart from readability, is still “it’s ugly.”) I don’t think there will ever be consensus on this, so just pick a style and be consistent. From a consistency point of view, we might as well keep the `{` after the declaration.

oli-obk:
I personally use the style shown by @BurntSushi and @rpjohnst 

```rust
fn lorem<
    T,
    U: Cake,
> (
    ipsum: T,
    aquarium: U,
) -> Laudanium where
    T: Babaorum,
{
    let x = 1;
    code.goes(here);
}
```

Of course, everything may get compressed if it's short:


```rust
fn lorem<T, U: Cake> (
    ipsum: T,
    aquarium: U,
) -> Laudanium where T: Babaorum {
    let x = 1;
    code.goes(here);
}
```

There's partial precedent (generics fit, args kindof fit) in [Ada's style guide](https://en.wikibooks.org/wiki/Ada_Style_Guide/Source_Code_Presentation) (it even has rationales to explain them)

```ada
generic
    type T is new Babaorum;
    type U is new Cake;
function lorem
    (ipsum:    T;
     aquarium: U)
    return Laudanium
is -- only in its own line if there's anything before the `begin`
    x: Integer := 1;
begin
    code.goes(here);
end lorem;
```

Note that the Ada style-guide is a little ambiguous, so the following is legal, too


```ada
generic
    type T is new Babaorum;
    type U is new Cake;
function lorem(
    ipsum:    T;
    aquarium: U
) return Laudanium
is -- only in its own line if there's anything before the `begin`
    x: Integer := 1;
begin
    code.goes(here);
end lorem;
```

and

```ada
generic
    type T is new Babaorum;
    type U is new Cake;
function lorem(ipsum:    T;
               aquarium: U
) return Laudanium
is -- only in its own line if there's anything before the `begin`
    x: Integer := 1;
begin
    code.goes(here);
end lorem;
```

daboross:
Adding another option here, as I'm not sure if it's been posed in this thread:

Function signatures with +4 more indentation for continuing function declaration lines, and no extra line before the opening `{`.

I find the following just as readable as the first example in the rfc, if not more:

```
// a comment
#[an_attribute]
#[another_attribute]
pub unsafe extern "C" fn foo<A, B, C>(x: SomeType,
                                      y: Vec<B>,
                                      z: SomeOtherType<B, C>)
        -> ReturnType<A>
        where A: Foo + Bar,
              B: Baz<B, C> {
    let x = func();

    x
}


fn foo<A, B, C>(x: SomeType,
                y: Vec<B>,
                z: SomeOtherType<B, C>)
        -> ReturnType<A>
        where A: Foo + Bar,
              B: Baz<B, C> {
    let x = func();

    x
}
```

The extra indentation makes it very clear which lines are in the declaration, and which are in the function body. The version proposed by the rfc, with `{` on a separate line, and @eddyb's version with `where` indented the same as the function also do this - what I've proposed is just another option for what to choose.

mitchmindtree:
Re having the `{` on a separate line, my thought process of late has been:

If the function signature would only require a single line, place the `{` on the same line.

Otherwise, if the function signature contains more than one line, place the `{` on its own line to more clearly distinguish between where the multi-line signature ends and the body starts.

bachm:
I don't believe that there is one perfect style suitable for both complex and simple signatures. We should either select on general style and create a low complexity and high complexity variant of it, or select one style optimized to handle the average joe function signature.

rkruppe:
Like everyone else, I have many opinions on style, but in the interest of quick(er) convergence, let me just say I am fully OK with the style @DanielKeep presents and also fully okay with most variants in the RFC text, with the exception of the position of the opening brace. I definitely want it to be on the same line for one-line signatures. I also see the argument for (and find it more pleasing, visually, than the alternative) to have it on a line of its own for multi-line signatures. This is inconsistent, but it's better than either consistent option IMHO. The [Kernel style guide](https://www.kernel.org/doc/Documentation/CodingStyle) justifies K&R braces like this:

> However, there is one special case, namely functions: they have the
opening brace at the beginning of the next line, thus:
>
> [example code]
>
>Heretic people all over the world have claimed that this inconsistency
is ...  well ...  inconsistent, but all right-thinking people know that
(a) K&R are _right_ and (b) K&R are right.  Besides, functions are
special anyway (you can't nest them in C).

Besides the tounge-in-cheek appeal to authority, this does make a good point... *for C*. In Rust it does not apply, functions are ordinary blocks, so if loops and conditionals and `match` put the opening brace on the same line, so should functions &mdash; *by default*. Opening brace on the same line works better if there is a *single* line that indicates the start of the block. Inside function bodies, the loop header/conditional/`match` can always be refactored to be a single line (helper function for the condition, temporary variable for the thing being matched). Only function signatures will unavoidably be so long that the opening brace would be on a different line than the start of the signature anyway. That, and the fact that multi-line signatures *look* very different, makes me okay with this inconsistency.

bluss:
@mitchmindtee: :+1:  this looks neat and is what I try to do. Incidentally, the same rule works well with `for` and `if` too.

I favour the RFC's current proposal, the where clauses aligned and with trailing comma, and the `{` on a new line if there are where clauses.

nielsle:
I like @BurntSushi's code example where aned enclosing ')' is never separated from the '-> returnvalue' 

/u/protestor said something similar on reddit (linking for

Anachron:
I'm just a beginner but this is what I have in mind: 
```rust
fn foo( a: TypeA, b: TypeB, c: TypeC )
    where {
        a: Foo+Bar, b: Foo(), c: Baz(42)
    } 
    -> Bar {
    return a   
}

fn dat(
        a: VeryLongComplicatedAndNestedType,
        b: GodKnowsWhatThisIs,
        c: HeyLookItsMe<Vec>
    ) where {
        a: Shorty
        b: JustFifty
        c: NotAgain
    }
    -> int {
    return 42
}
```

Maybe it will cause some new ideas?

sgkim126:
I love @oli-obk's one, because it keeps the git history even the first or last argument is changed.

dotdash:
So far, I prefer the style shown by @DanielKeep, especially regarding the alignment of the function arguments. Having them shifted to the right by varying amounts (depending on the function name) makes it incredibly hard for me to quickly scan for them when I'm looking at the function body and want to look up the type of an argument. And then there's the diff-noise caused by re-aligning them, which I absolutely hate.

text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+       B,
+       C>
+      (x: SomeType,
+       y: Vec<B>,
+       z: SomeOtherType<B, C>)
+{
+```
+
+If more space is necessary for arguments, then the opening parenthesis may be on
+a new line with block indent. Consider this only in an emergency. E.g.,
```
Florob: It says "block indent" here, but the example shows "block indent + 4".

text/0000-style-fn-decl.md
```diff
@@ -0,0 +1,274 @@
...
+fn foo<A,
+       B,
+       C>
+      (x: SomeType,
+       y: Vec<B>,
+       z: SomeOtherType<B, C>)
+{
+```
+
+If more space is necessary for arguments, then the opening parenthesis may be on
```
oli-obk: I like how in struct constructors the arguments all look the same (`name: Type,`), no matter if they are the first or the last. I suggest doing the same for function generics and arguments. Imo this increases readability significantly and allows easy adding/removing of function generics and arguments (add/remove a line). Therefor I suggest that the opening parenthesis is always on the same line, and if the previous styles take up more than ~40 chars of width, use the style below

tafia:
:+1: for @DanielKeep style. In particular this part:
> I don't agree with aligning arguments with the inside of the opening paren

... and not only for `fn` declaration.
In addition to what he already said, I think it just looses too much space. Once you start having nested functions (with iterators, it's easy) you *quickly* end up having lot of thing on the very right side.

BurntSushi:
:+1: for @DanielKeep too. I think I like how you handle `where` better than my example. I also very strongly agree with this opinion:

> I don't agree with aligning arguments with the inside of the opening paren

I've tried this in code because I also write a lot of Python where this style is popular, but I've never been happy with it in Rust.

For the folks saying that @DaneilKeep's style uses too much vertical space: what if we didn't require each argument be on its own line? I happen to like it because it's diff friendly and makes it easy to add/remove arguments if a function has a lot of them. But I can see how it might not scale well to functions with lots of parameters.

DanielKeep:
@BurntSushi If you have a function with so many arguments that it no longer fits on the screen, I'd argue you have more fundamental problems than the formatting style. :)  If anything, I'd hope it would provide impetus to refactor the function's interface.

And anyway, style guidelines almost always fall apart at the extremes.  It's how it works on the common 90% of code that really matters.

Diggsey:
I'm not too fussed on the rest of the style, but I strongly agree with @DanielKeep about never aligning things with the inside of an opening brace for all of the reasons given.

ben0x539:
I don't think having inconsistent rules for whether `{`s go on their own line is the way forward, even if some other programming languages go with that. If we decide opening braces of functions need to be on their own line, we should adopt the same rule for block expressions, braces in control flow constructs, macros, various type definitions, etc.

I also think enshrining vertical alignment of parameters or bounds after a `where` in the official style guide is not a good match for Rust. It'd be a neat idea for a language like Haskell that has syntax based on vertical alignment to begin with, but I think we should just stick to defining leading whitespace in terms of indentation levels. That has the added benefit of supporting non-monospace depictions of code, which is apparently a thing some people prefer, and a thing other people can't seem to avoid in publications. Maybe more significantly, it also avoids code drifting far rightwards with long function names which looks kind of silly.

`->` should probably be at the beginning of its line for easier scanning, so no cute `) -> T` lines imo. Give the `)` another line of its own if you have to, or just leave it with the last parameter in place of the `,`.

Kimundi:
I'm also for @DanielKeep's proposal, mainly because I just want to align with a single tab keystroke rather than with manual spaces :)

yazaddaruvala:
> I don't agree with aligning arguments with the inside of the opening paren

:+1:

tshepang:
I don't agree with aligning arguments with the inside of the opening paren

:+1:

iliekturtles:
> I don't agree with aligning arguments with the inside of the opening paren

:+1:

That a change to a function or variable name can have formatting consequences elsewhere in the code is frustrating.

tbelaire:
(+1) for @DanielKeep but please keep the `{` on the same line for normal functions that don't require breaking into so many lines or don't have `where` clauses.

gsingh93:
I just want to mention again that the "body alignment problem" with using only a single indent for a newline applies to things other than functions (i.e. if statements). I'm only bringing this up because it would be nice if the fix for all cases was the same. I do like the no-indent `where` clause style, but that can't be applied to `if` statements. On the other hand, using two indents is a fix that can be applied to both.

In any case, I'm fine with either of these fixes if that means the opening brace stays on the same line.

I'm also fine with both styles of argument alignment. I'm still more worried about the requirement that each argument gets it's own line.

nikomatsakis:
So, to start, I hate all styles that are not my own with the passion of a thousand suns. And I change my own style regularly because I hate it too. That said, I guess we have to settle on *something*. It seems useful to come up with criteria to me, and see if that leads us places.

To that end, I have to give a :+1: to the idea that you should be able to rename the fn without destroying indentation. So much as I love indenting with the inside of an open `(` (those of you who don't like it just aren't using emacs, vi, or some other "modern" editor, clearly), that suggests that if the whole fn can't fit on one line, you should indent the arguments onto the next line.

I personally prefer the `->` to be aligned with the arguments (contra both this RFC and @DanielKeep), but I can perhaps get behind the diff-friendly style @DanielKeep advocates, where each argument has a trailing comma, and the `) -> ` is on the following line.

nikomatsakis:
Hmm. I decided I was hasty to comment last night. I'm going to do some experimentation locally as I write code this week and see what I think about these various styles...in particular, I just now noticed a function as I was editing:

```rust
    fn relate_free_regions(&mut self,
                           fn_sig_tys: &[Ty<'tcx>],
                           body_id: ast::NodeId,
                           span: Span) {
        debug!("relate_free_regions >>");
```

and I thought to myself how much I appreciate that, in our current style, we can use this "lightweight form" when applicable, only falling back to "heavier" forms when needed (i.e., longer function names, longer parameter names, where clauses, etc).

Now, I DO still get annoyed when changing a fn name means I have to re-indent all the parameters, but then part of the point of this discussion is that @nrc is working on this nifty rustfmt tool that would do that for me, so perhaps this consideration is moot.

aidancully:
If I can, I'd like to suggest that the style should be constrained by some maintainability principles... I'd like to suggest a few:

* The style should allow the code structure (branch / merge points, scope definitions, function definitions / declarations, statements) to be understood, without requiring reading the code text.
* The style should require as little as possible from its tools, for either editing or display.
* The style should impose minimal burden on maintainers.

If we can agree on principles like these, it gives a basis for accepting / rejecting specific styles. In particular, aligning function arguments with the open-paren goes against, *ahem*, all of these principles, I think. The way this violates the second and third principles are discussed by other commenters, so I'll talk about the first principle, instead...

McConnell had a good technique in Code Complete for checking whether the style represented code-structure information: If you were to black out all the characters in a block of code, how much information could you still get from reading it? Take an example:

```
// function name is three characters - one less than default tab-stop of 4, so
// that the open-paren aligns to the tab-stop.
fn foo(...);
fn bar() {
    foo(arg1, arg2, arg3,
        arg4, arg5, arg6,
        arg7);
    something_else(arg1, arg2, arg3,
                   arg4, arg5, arg6);
    if blah {
        lorem ipsum dolor sit amet;
    } else {
        consectetur adipiscing elit;
    }
}
```

After replacing all characters (save leading whitespace) of `bar` with `.`, we get:

```
..........
    .....................
        .................
        ......
    ................................
                   ..................
    .........
        ...........................
    ........
        ............................
    .
.
```

So, two things to point out here... 1, can you really tell the difference between the `foo()` invocation and the `if` block?, and 2, the `foo()` invocation and the `something_else()` invocation have the same structure, but they look completely different when you just look at the indentation. The style is obscuring the structure.

I personally prefer a style where arguments are half-indented compared to control-flow (*e.g.* 2 spaces, when the control-flow nesting indentation is 4), and if I need to split a function's arguments across multiple lines, then I don't put any arguments on the same line as the function:

```
fn bar() {
    foo(
      arg1, arg2, arg3,
      arg4, arg5, arg6,
      arg7);
    something_else(
      arg1, arg2, arg3,
      arg4, arg5, arg6);
    if blah {
        lorem ipsum dolor sit amet;
    } else {
        consectetur adipiscing elit;
    }
}
```

liigo:
please place `{` at the end of a line.

blaenk:
I also follow the style that has been attributed to @DanielKeep with the `where` at the beginning of its own line. When there are a lot of parameters, I don't like to align them with the opening parenthesis, but instead do what @BurntSushi showed in [his comment](https://github.com/rust-lang/rfcs/pull/1190#issuecomment-118697011).

So to change his second code example in that comment to what I do, it's:

``` rust
pub fn send<'c: 'a, 'b: 'a, C, T, F>(
    mut self,
    chan: C,
    val: T,
    mut run: F,
) -> Select<'a>
where C: Sender<Item=T> + Clone + 'c,
      T: 'b,
      F: FnMut() + 'a {
    // code
}
```

I view that method of writing many parameters as the "lesser of evils." It's not very common in my code, but when it is it looks much better IMO than overly-indented parameters. It is much quicker to scan and see each individual parameter, and to make modifications to the parameters.

The type bounds can even be placed on their own line like the parameters are to make it even quicker to scan, and less indented:

``` rust
pub fn send<'c: 'a, 'b: 'a, C, T, F>(
    mut self,
    chan: C,
    val: T,
    mut run: F,
) -> Select<'a>
where
    C: Sender<Item=T> + Clone + 'c,
    T: 'b,
    F: FnMut() + 'a {
    // code
}
```

gsingh93:
I still can't get over the fact that with this proposed style *every* one of my two line function declarations will now be 10+ lines.

blaenk:
@gsingh93 can you post an example? I see none that you've posted in this issue. I thought it was obvious that these styles are for when there are just too many function or type parameters, i.e. the extreme and general case. If there are a reasonable amount then you can fit everything in two lines:

``` rust
fn foo<T, U>(one: T, two: U) -> i32
where T: Shape, U: Color {
    // code
}
```

gsingh93:
@blaenk I'll quote @reem on this one:
> I do not think it's necessary for each argument to get its own line, this just makes for unnecessarily tall argument lists when we have lots of horizontal space to use. In my code, I usually continue to list arguments until I run out of horizontal space, then go to the next line and continue (with the start of each line lining up with the first argument). For example:

> ```
> pub fn listen_with<A>(mut self, addr: A, threads: usize,
                      protocol: Protocol) -> HttpResult<Listening>
> where A: ToSocketAddrs {
> ```
> As you can see in the example, I also do not think the return type needs its own line should it fit on the end of the last line of arguments.

Obviously that's not the worst case example, as 3 lines are turned into 7 or 8 lines (depending on if the predicate gets it's own line), but it would be easy to throw some more variable names in there and make it fit.

I completely understand the argument for putting arguments and predicates on their own lines for readability, but I don't think this is necessary in *every* case for readability. I use both the newline for each argument style and the fit as many arguments per line as possible style in my code, depending on the function.

Other than the compactness issue, I'm worried about the amount of churn necessary to convert a one line function signature to a multiline one as soon as I add an argument that pushes it over the limit. It's going to get annoying very quickly.

rpjohnst:
I do not believe anyone has actually proposed *always* putting everything on its own line. The purpose here is to specify where to put linebreaks and indentation when they actually are necessary. In the common case the entire function signature should go on one line, and in the middling case multiple arguments should share lines.

gsingh93:
@rpjohnst The RFC states that arguments should not share lines:

> if the signature requires more than one line, each argument gets its own line

blaenk:
@gsingh93 

> Other than the compactness issue, I'm worried about the amount of churn necessary to convert a one line function signature to a multiline one as soon as I add an argument that pushes it over the limit. It's going to get annoying very quickly.

I think this is a very good point. Personally I think weighed against the other advantages I believe this style has, I still prefer this style, and given that the RFC is in part created to know what style rustfmt should enforce, I think it'd be a non-issue. It'll be like gofmt where you write your code however and then just run rustfmt on it (usually bound to some key binding in your editor) to automatically conform to style. So in this scenario, you would write it out in-line or however you want, then trigger some binding e.g. `<leader>-f` in vim to automatically re-format everything to the rustfmt style.

ssokolow:
@blaenk *nod* That's also why I've tried so hard to find objective arguments for not putting the curly brace on its own line.

I'm stubborn enough about the styling of my own code that, if such a style made it into a rustfmt implementation, I'd be torn between not using rustfmt and maintaining a patched fork... neither of which I want to do.

gsingh93:
I guess I'm a little more convinced with the `rustfmt` argument. The thing is I've never really used a code formatter, so I didn't really see myself using `rustfmt` in the future. But if `rustfmt` is pushed as something everyone should be using (and maybe with possible Cargo integration?), I'd be OK with the multiline function signatures.

@ssokolow The curly brace on it's own line is something I'm still against, even with `rustfmt`. It's pointless, doesn't help readability, and wastes space. I believe it was only suggested to solve the body alignment problem, and the no-indent where clause style solves that.

blaenk:
Yeah having some formatting program like that is inherently conflicting I think, because everyone has their own style that they have refined over the years of experience they've had, that they believe to be the one true style and everyone else is crazy to do their own thing.

I think when we have a formatting program it becomes a matter of deciding if personal style is secondary to the benefits of there being a single or at least majority style that everyone is familiar with and can get to working with. It actually removes the whole time-consuming question of style; you don't have to worry about it or what style project A uses and how it differs from the style project B uses, you just throw rustfmt at it and have it over with. If we later determine that some other style is superior, just update rustfmt and re-run the code through it.

There definitely could be some cargo integration for it, or hooks such as git's could be leveraged otherwise. There's bound to be some tool to automatically run it whenever you change a file, which is something easily done in an editor as well.

Personally I really dislike for example the fact that gofmt uses tabs and no spaces, or how it removes spacing around bitwise operators e.g. `color&mask`, which I don't see the reasoning for. But since gofmt exists and everyone uses it, I don't bother spending too much time caring about it, as long as I can consume (read/write) the produced code just as well.

retep998:
@gsingh93 The way I'd format that example would be

```Rust
pub fn listen_with<A>(
    mut self, addr: A, threads: usize, protocol: Protocol,
) -> HttpResult<Listening>
where A: ToSocketAddrs {
```

Only 4 lines. Hardly a disaster compared to the 3 lines of the other version.

gsingh93:
@retep998 Sure, that's a reasonable alternative, but it's not in the RFC if that's what you're saying.

retep998:
@gsingh93 So let's use my style instead of what the RFC proposes!

blaenk:
I think I would be ok with bunny's proposal.

benaryorg:
If there would be an automatic formatting tool in e.g. cargo, which was highly configurable, everyone could set up their projects how they wanted (five minutes of configuring would be okay for a large project I think) and use that formatting forever.

PRs would be well formatted as cargo does that for them (and if I want to help a project I will honour their formatting guidelines).

Everybody could choose for their projects what they want.

Teams had no conflicts as they would agree on a formatting from the beginning.

People like me, who have a very unusual (and most of the time unwanted) formatting style would (hopefully) agree to the guidelines given and use their own for their projects.

The project wide configuration file could be enforced by both, cargo and the editors.

It would definitively ensure a consistent format and would completely erase the problems arising from standardised language-wide formatting guidelines.

ssokolow:
I think we should still endeavour to come up with some kind of PEP8-like "good enough for most people" formatting standard. Despite being stubborn, I found myself drifting to using underscore_function_names in Python simply because Prospector demands pep8-naming and I was too busy to fork it to produce a camelCaseFunctionNaming version.

(Though it did make my code slightly less readable in a quantifiable way given the interaction between the extra line length from the underscore requirement and the word-wrapping requirement. I wound up either using stranger word-wrapping patterns or making function and variable names more cryptic to compensate.)

BurntSushi:
@ssokolow Agreed. There should at least be a canonical format that is officially recommended. The argument of how configurable `rustfmt` should be is orthogonal.

tshepang:
@benaryorg nice points there

benaryorg:
@tshepang If you have a formatting that is accepted by nobody, you learn to defend your formatting with good arguments ;-)

liigo:
+1

rkjnsn:
Big :-1: to aligning arguments on following lines with the open parens, or anything else that results in not using a consistent multiple of the indent width.

benaryorg:
Firt of all, please use tabs for intending.
If you use tabs, you can write your wonderful 2-space code without having to worry that I am gonna get mad because I don't like that.
Tabs make the code more universal.

The rest could look like this:

```rust
pub fn send<'c: 'a, 'b: 'a, C, T, F>
(
    mut self,
    chan: C,
    val: T,
    mut run: F
) -> Select<'a> where
    C: Sender<Item=T> + Clone + 'c,
    T: 'b,
    F: FnMut() + 'a
{
    //code
}
```

nrc:
Thanks for all the feedback. My plan is to make sure that rustfmt can support as options all of the popular suggestions here, that will make it easy to see how much of a change that makes and what a style looks like over large, real files. I'll then post some examples and hopefully we can decide on something for the style guide and rustfmt defaults (whether rustfmt should allow non-default formats is a discussion for another day...).

nrc:
I believe that with [this PR](https://github.com/nrc/rustfmt/pull/238) rustfmt can now handle pretty much every variation mentioned in this thread, although not all combinations have tests.

blaenk:
awesome work @nrc

killercup:
Wow, that looks amazing, @nrc!

nrc:
We briefly discussed this at the lang team meeting today and we decided to close it for now, with no decision made.

Rustfmt can now support all (I think) of the options discussed in this thread and I believe the defaults are conservatively close to the current situation in the Rust repo. As discussed [here](http://ncameron.org/blog/rustfmt-ing-rust/), I am trying to rustfmt the Rust repo, bit by bit. Once the whole repo can be rustfmt'ed then style becomes much more mutable than it is currently (without tool support a reformatting to some minor style change is an Herculean task). At that point I plan to open a bunch of style RFCs to establish formal guidelines for Rust programming style and thus for the rustfmt default options (which one day, may or may not be the only options available).

Issue #947, opened 2015-03-06
Open ranges in pattern matching


nrc:
Allow `...x` and `y....` in matches. Proposed in #880.

PR #1192, opened 2015-07-07
RFC for inclusive ranges with ...

huonw:
[Rendered](https://github.com/huonw/rfcs/blob/inclusive/text/0000-inclusive-ranges.md)

### DIFF
--- a/text/0000-inclusive-ranges.md
+++ b/text/0000-inclusive-ranges.md
@@ -1,0 +1,79 @@
+- Feature Name: inclusive_range_syntax
+- Start Date: 2015-07-07
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Allow a `x...y` expression to create an inclusive range.
+
+# Motivation
+
+There are several use-cases for inclusive ranges, that semantically
+include both end-points. For example, iterating from `0_u8` up to and
+including some number `n` can be done via `for _ in 0..n + 1` at the
+moment, but this will fail if `n` is `255`. Furthermore, some iterable
+things only have a successor operation that is sometimes sensible,
+e.g., `'a'..'{'` is equivalent to the inclusive range `'a'...'z'`:
+there's absolutely no reason that `{` is after `z` other than a quirk
+of the representation.
+
+The `...` syntax mirrors the current `..` used for exclusive ranges:
+more dots means more elements.
+
+# Detailed design
+
+`std::ops` defines
+
+```rust
+pub struct RangeInclusive<T> {
+    pub start: T,
+    pub end: T,
+}
+```
+
+Writing `a...b` in an expression desugars to `std::ops::RangeInclusive
+{ start: a, end: b }`.
+
+This struct implements the standard traits (`Clone`, `Debug` etc.),
+but, unlike the other `Range*` types, does not implement `Iterator`
+directly, since it cannot do so correctly without more internal
+state. It can implement `IntoIterator` that converts it into an
+iterator type that contains the necessary state.
+
+The use of `...` in a pattern remains as testing for inclusion
+within that range, *not* a struct match.
+
+The author cannot forsee problems with breaking backward
+compatibility. In particular, one tokenisation of syntax like `1...`
+now would be `1. ..` i.e. a floating point number on the left, however, fortunately,
+it is actually tokenised like `1 ...`, and is hence an error.
+
+# Drawbacks
+
+There's a mismatch between pattern-`...` and expression-`...`, in that
+the former doesn't undergo the same desugaring as the
+latter. (Although they represent essentially the same thing
+semantically.)
+
+The `...` vs. `..` distinction is the exact inversion of Ruby's syntax.
+
+Only implementing `IntoIterator` means uses of it in iterator chains
+look like `(a...b).into_iter().collect()` instead of
+`(a..b).collect()` as with exclusive ones (although this doesn't
+affect `for` loops: `for _ in a...b` works fine).
+
+# Alternatives
+
+An alternate syntax could be used, like
+`..=`. [There has been discussion][discuss], but there wasn't a clear
+winner.
+
+[discuss]: https://internals.rust-lang.org/t/vs-for-inclusive-ranges/1539
+
+This RFC doesn't propose non-double-ended syntax, like `a...`, `...b`
+or `...` since it isn't clear that this is so useful. Maybe it is.
+
+# Unresolved questions
+
+None so far.
--- a/text/0000-inclusive-ranges.md
+++ b/text/0000-inclusive-ranges.md
@@ -29,25 +29,25 @@ more dots means more elements.
 pub struct RangeInclusive<T> {
     pub start: T,
     pub end: T,
+    pub finished: bool,
 }
 ```
 
 Writing `a...b` in an expression desugars to `std::ops::RangeInclusive
-{ start: a, end: b }`.
+{ start: a, end: b, finished: false }`.
 
 This struct implements the standard traits (`Clone`, `Debug` etc.),
-but, unlike the other `Range*` types, does not implement `Iterator`
-directly, since it cannot do so correctly without more internal
-state. It can implement `IntoIterator` that converts it into an
-iterator type that contains the necessary state.
+and implements `Iterator`. The `finished` field is to allow the
+`Iterator` implementation to work without hacks (see Alternatives).
 
 The use of `...` in a pattern remains as testing for inclusion
 within that range, *not* a struct match.
 
 The author cannot forsee problems with breaking backward
 compatibility. In particular, one tokenisation of syntax like `1...`
-now would be `1. ..` i.e. a floating point number on the left, however, fortunately,
-it is actually tokenised like `1 ...`, and is hence an error.
+now would be `1. ..` i.e. a floating point number on the left,
+however, fortunately, it is actually tokenised like `1 ...`, and is
+hence an error with the current compiler.
 
 # Drawbacks
 
@@ -58,10 +58,8 @@ semantically.)
 
 The `...` vs. `..` distinction is the exact inversion of Ruby's syntax.
 
-Only implementing `IntoIterator` means uses of it in iterator chains
-look like `(a...b).into_iter().collect()` instead of
-`(a..b).collect()` as with exclusive ones (although this doesn't
-affect `for` loops: `for _ in a...b` works fine).
+Having an extra field in a language-level desugaring, catering to one
+library use-case is a little non-"hygienic".
 
 # Alternatives
 
@@ -73,6 +71,28 @@ winner.
 
 This RFC doesn't propose non-double-ended syntax, like `a...`, `...b`
 or `...` since it isn't clear that this is so useful. Maybe it is.
+
+The `finished` field could be omitted, leaving two options:
+
+- `a...b` only implements `IntoIterator`, not `Iterator`, by
+  converting to a different type that does have the field. However,
+  this means that `a...b` behaves differently to `a..b`, so
+  `(a...b).map(|x| ...)` doesn't work (the `..` version of that is
+  used reasonably often, in the author's experience)
+- `a...b` can implement `Iterator` for types that can be stepped
+  backwards: the only case that is problematic things cases like
+  `x...255u8` where the endpoint is the last value in the type's
+  range. A naive implementation that just steps `x` and compares
+  against the second value will never terminate: it will yield 254
+  (final state: `255...255`), 255 (final state: `0...255`), 0 (final
+  state: `1...255`). I.e. it will wrap around because it has no way to
+  detect whether 255 has been yielded or not. However, implementations
+  of `Iterator` can detect cases like that, and, after yielding `255`,
+  backwards-step the second piece of state to `255...254`.
+
+  This means that `a...b` can only implement `Iterator` for types that
+  can be stepped backwards, which isn't always guaranteed, e.g. types
+  might not have a unique predecessor (walking along a DAG).
 
 # Unresolved questions
 
--- a/text/0000-inclusive-ranges.md
+++ b/text/0000-inclusive-ranges.md
@@ -31,13 +31,18 @@ pub struct RangeInclusive<T> {
     pub end: T,
     pub finished: bool,
 }
+
+pub struct RangeToInclusive<T> {
+    pub end: T,
+}
 ```
 
 Writing `a...b` in an expression desugars to `std::ops::RangeInclusive
-{ start: a, end: b, finished: false }`.
+{ start: a, end: b, finished: false }`. Writing `...b` in an
+expression desugars to `std::ops::RangeToInclusive { end: b }`.
 
-This struct implements the standard traits (`Clone`, `Debug` etc.),
-and implements `Iterator`. The `finished` field is to allow the
+`RangeInclusive` implements the standard traits (`Clone`, `Debug`
+etc.), and implements `Iterator`. The `finished` field is to allow the
 `Iterator` implementation to work without hacks (see Alternatives).
 
 The use of `...` in a pattern remains as testing for inclusion
@@ -59,7 +64,9 @@ semantically.)
 The `...` vs. `..` distinction is the exact inversion of Ruby's syntax.
 
 Having an extra field in a language-level desugaring, catering to one
-library use-case is a little non-"hygienic".
+library use-case is a little non-"hygienic". It is especially strange
+that the field isn't consistent across the different `...`
+desugarings.
 
 # Alternatives
 


Kimundi:
Imo we should try providing a `(x..y).inclusive()` method first. I'm also worried about potentially stealing the syntactic space for variadic generics

text/0000-inclusive-ranges.md
```diff
@@ -0,0 +1,79 @@
...
+    pub end: T,
+}
+```
+
+Writing `a...b` in an expression desugars to `std::ops::RangeInclusive
+{ start: a, end: b }`.
+
+This struct implements the standard traits (`Clone`, `Debug` etc.),
+but, unlike the other `Range*` types, does not implement `Iterator`
+directly, since it cannot do so correctly without more internal
```
gsingh93: Can you clarify why more state is needed? Why not just increment start while start is less than or equal to end?

Gankro: @gsingh93 The problem is that `0...u8::MAX` has 257 states: Some(i) for all i; and None. 
So you can't modify only one bound and get all the states. You have two options:

* extra flag for "inclusive case yielded"
* do a checked_add on each step, and if it fails then decrement the maximum instead of increment the minimum (symmetric logic applies for `next_back`)

eddyb:
@Kimundi VG can still use `<Args...>` and `(args...)` because `start...` is not exactly meaningful (it goes up to infinity... inclusively?), though I expect `...end` to work.

text/0000-inclusive-ranges.md
oli-obk: so why can't RangeInclusive simply get another field that is set during desugaring?

nikomatsakis:
I want inclusive ranges, and I want the syntax to be `x...y`, because it's analogous with patterns. I also want exclusive patterns, but I'm content to wait. Similarly, I think we should support `...y`. It seems clear that it is useful in the same way that `..y` is useful. If I want the (bigint) numbers from 0 to n inclusive, I should be able to write `...BigInt(N)`, and not `BigInt(0)...BigInt(N)`.

As for variadics, if the syntaxes turn out to be in conflict in some way, I'd rather have `...` for inclusive ranges, given the precedent from patterns. I think it'd be more confusing otherwise. (My impression when we settled on changing patterns to `...` was that we aimed to eventually support both inclusive/exclusive patterns/ranges).

nikomatsakis:
Oh, also: I think that `x...y` should be an iterator, even if it requires an extra field. I think we should make it work as analogously to `x..y` as possible. Any discrepancies will be annoying warts. I don't think people store "inclusive ranges" as a datastructure very often (if ever), so the fact that the struct has an extra field shouldn't be a big deal. The usual way to use this is either as an iterator (in which case you want the field), or `&vec[x...y]`, in which case hopefully everything is inlined and you don't ever create the struct in the first place.

Gankro:
Agreed with niko, though I'd be interested in investigating the API and performance impact of doing the hack I suggest to avoid a second field. You need to do a check in *every* step anyway. Might as well fold it into the add, right?

Things that come to mind: 

* You can't have inclusive range iterators of unit types (what a tragedy!)
* I find the step API to be frustratingly opaque, but it *seems* like it has all the parts for doing checked adds. For types like BigInts these can return unconditional Some and therefore the checks should be const-folded fairly easy.

text/0000-inclusive-ranges.md
huonw: I discussed it with a few people while writing the API, but got such negative feedback that I didn't even think to put it as an alternative. However, several people seem to be in support of it.

petrochenkov:
How common are inclusive ranges? I'm wary of putting a language sugar into places where a programmer benefits from it like once a year. (Library sugar like `(x..y).inclusive()` is ok.)

On the other side, inclusive ranges do improve consistency and completeness of the language, that already has inclusive range patterns and exclusive range expressions. On this ground, I'd suggest to extend the RFC with half-open inclusive range `...a` and exclusive range patterns and position it as a complete "language beauty package".

nikomatsakis:
@petrochenkov I agree that this works best if we have half-open inclusive ranges and exclusive range patterns as well. If we didn't already have inclusive patterns, I would prefer a method too.

quantheory:
CC #947

blaenk:
Yeah LGTM.

huonw:
I've updated the RFC to include `...b`, and to give a design that allows `a...b` to implement `Iterator` directly, to match `a..b`.

nikomatsakis:
Hear ye, hear ye. This RFC is entering **final comment period**.

nikomatsakis:
@huonw looks good to me.

rkruppe:
I realize this problem already exists with patterns, but I don't like that `...` is so close to `..`, visually. Mixing them up can result in off-by-one errors, one of two hard problems in computer science, so debugging them should not be made even harder by by making the two kinds of ranges identical from afar or in a hurry. To quote `python-ideas` (referring to Tim Peters):

> Syntax must not look like grit on Tim's screen!

Therefore I favor `(a..b).inclusive()`, not just as conservative first step, but as permanent solution.

tbu-:
I think it's too similar to `...`, I wouldn't add it just for the sake of consistency with patterns. Python, for example, only has exclusive ranges as well, and is one of the more convenient programming languages – if we introduce this I'd be in favor of a method like `(a..b).inclusive()`, because that makes it loud and obvious that this is not a normal range.

alexcrichton:
The libs team discussed this at triage today, and our conclusion was that we're ok with the structures defined here. The best route forward is probably to leave the `finished` field as `#[unstable]` for the time being to leave the door open to possible future improvements, but otherwise we're ok with it being public.

bluss:
Will slices and vectors be indexed with inclusive ranges?

huonw:
@bluss not sure; definitely not initially, and seems like that would be best be considered/proposed as a separate RFC.

bluss:
Slice use of inclusive ranges seems to be an aspect that needs to be covered as well. Without it, this RFC adds an inconsistency. If we don't want to allow `&foo[a...b]`, then maybe we should not have this syntax? Or will users be ok with the logic that half open ranges are index ranges, and closed ranges are not?

blaenk:
Heh, it didn't occur to me that this RFC wouldn't include slicing of slices/vectors with inclusive ranges. I do think we should support it, I think it'll be weird to have inclusive ranges but not support them in the area that is probably most commonly used, even if it eventually comes later. It feels to me like it should be a complete package.

bombless:
>I realize this problem already exists with patterns, but I don't like that ... is so close to ..

agree with that, how about `..~`

SimonSapin:
> Will slices and vectors be indexed with inclusive ranges?

I think they should not. I prefer the mental model where indices in a slice refer not to items but to positions between items. This makes the meaning of `Index<Range>` obvious. (`Index<usize>` gives the item *after* a position.)

https://blog.nelhage.com/2015/08/indices-point-between-elements/

text/0000-inclusive-ranges.md
```diff
@@ -0,0 +1,106 @@
...
+# Alternatives
+
+An alternate syntax could be used, like
+`..=`. [There has been discussion][discuss], but there wasn't a clear
+winner.
+
+[discuss]: https://internals.rust-lang.org/t/vs-for-inclusive-ranges/1539
+
+This RFC doesn't propose non-double-ended syntax, like `a...`, `...b`
+or `...` since it isn't clear that this is so useful. Maybe it is.
```
RalfJung: Actually, it now does.

blaenk:
I didn't think we meant to replace the current indexing with inclusive ranges, I thought maybe we could support both, is that not possible?

SimonSapin:
> I didn't think we were saying we should replace the current indexing with inclusive ranges

Of course, removing existing `Index<Range>` impls would be backward-incompatible.

> I thought maybe we could support both, is that not possible?

It is possible, multi-dispatch allows adding more impls. But that doesn’t necessarily make it desirable. Inclusive ranges have their use for iteration (e.g. `for i in 0..255_u8`) but how do they help for slicing, compared to "half-open" ranges? Again, think of indices as positions between items.

It’s also a matter of design principles, I come from Python with its [There should be one-- and preferably only one --obvious way to do it](https://www.python.org/dev/peps/pep-0020/) zen.

blaenk:
Yeah I generally agree with the single way principle, I just personally don't really see it as much of a problem. Exclusive ranges would continue to be the norm the majority of the time, it just seems inconsistent that if you have `b` and you want to slice from `a` up to and including `b`, you have to do `&vec[a .. (b + 1)]` when you otherwise have this inclusive range syntax in the language that would allow you to do: `&vec[a ... b]`. I don't really see how it affects anything else, and would be nice and useful when needed, while omitting it would feel like an inconsistency/half-baked feature.

Veedrac:
I really dislike the idea that `...` exists but can only be used in a small and seemingly arbitrary subset of circumstances where `..` can be used, despite both being so syntactically and semantically similar. It just seems super confusing.

If we use `...`, I want it to do *everything* that `..` can do. If we don't want to do that, let's please use a method. `(a..b).inclusive()` seems fine to me.

bluss:
The motivation mentions loops from `a` to `b` inclusive, this can be covered with a simple adaptor method on the `a..b` range.
 
We already have `[a, b)` ranges, this rfc proposes to add `[a, b]` ranges (using `[` for an inclusive end and `)` for an exclusive/open end).  We already have use cases for `(a, b)` ranges too (BTreeMap's range and drain, and accessors), but that case is not resolved in this RFC. So the added `[a, b]` is neat, but does not give us the full power we actually need.

Veedrac:
FWIW, if your only argument right now is `for x in a...b`, just do `for x in (a..b).chain(Some(b))`. Unless this comes up everywhere, I'm not really convinced we even need a method, never mind syntax.

bluss:
@Veedrac I'd suspect that the chain solution wouldn't actually compile well enough. The inclusive range iteration needs its own iterator implementation to be efficient.

Veedrac:
I looked for the obvious cases in rustc where this would be used.

No character ranges were found.

For a search of `..` followed by `+ 1` (regex: `[^.]\.\.[^.].*\+\s*1`), and excluding all tests except for the shootout, we have three groups of usages. One is looping with `for`:

Veedrac:
@bluss Suboptimal code is unfortunate, but this shouldn't always be the `std`'s responsibility. It depends how common and how important to optimize the code is.

---

I looked for the obvious cases in rustc where this would be used.

No character ranges were found.

For a search of `..` followed by `+ 1` (regex: `[^.]\.\.[^.].*\+\s*1`), and excluding all tests except for the shootout, we have three groups of usages. One is looping with `for`:

    for deref in 1..autoderefs + 1 {
    for i in 1..lines + 1 {
    for _ in 0..(backslashes+1) {
    for i in 1 .. iterations + 1 {
    for i in 1..n as usize + 1 {
    for j in 0..i + 1 {
    for n in 1..max + 1 {
    for i in 2 .. n_tasks + 1 {

The second is indexing:

    for (place, val) in pp.iter_mut().zip(&self.perm.p[..i+1]) {
    self.edges.$index(.. (pos + 1))
    &ss[last_start..i + 1]),
    let n = try!(self.inner.write(&buf[..i + 1]));
    buf.push_all(&available[..i + 1]);
    rotate(&mut perm[..i + 1]);
    try!(wr.write(&line[..nb+1]));

And the third is iteration outside of `for`.

    let messages = (min_depth..max_depth + 1).step_by(2).map(|depth| {
    let mut dcol: Vec<_> = (0..t.len() + 1).collect();

This means 9/17 instances would soley need iteration, 7/17 would need indexing and 1 would need `step_by`. Not all of these would actually want to switch, though.

---

The tests gave these extra results:

    for at in 0..len + 1 {
    for i in 1..DATASZ + 1 {
    for j in i+1..s.len() + 1 {
    for len in 0..cap + 1 {
    for tail_pos in 0..max_cap + 1 {
    for to_remove in 0..len + 1 {
    i += s[i + 10..].find("stack backtrace").unwrap() + 10;
    let data = (0..4*100+1).map(|i| {
    let data = (0..4*100+1).map(|i| {
    let data = (1..DATASZ + 1).collect::<Vec<_>>();
    let expected: Vec<_> = (0..i+1).collect();
    let expected: Vec<_> = (0..i+1).collect();
    let expected_left: Vec<_> = (-last..j+1).rev().collect();
    let expected_left: Vec<_> = (-last..j+1).rev().collect();

for which only one required indexing.

---

Given this, I still think this is not common enough to have dedicated syntax.

If we do wish to have this in some form, it would seem an oversight to not include indexing, as about 40% of the use-cases are indexing. This means half-open ranges as well.

That said, the indexing case is already effectively bounds checked by construction so a `+1` isn't too worrisome. That would leave no more than 10 discovered use-cases in the standard library, and some would be better off not switching.

I guess the point of what I'm saying is that I don't see the justification here. Having a subtle difference between `...` and `..` seems like an order of magnitude larger a problem than having slightly suboptimal inclusive ranges.

Blackhole16:
I do like the idea of having a lot of syntactic sugar, therefore I rate for `...`. I see the point, that inclusive ranges are not that common, but as rust depends a lot on ranges, still wanting to have maximum speed, I'd like to see `...` becoming part of the language.

Featurewise `...` should be equivalent to `..` as otherwise there would be confusing discrepances.

My only problem is, that `...` is not enough distinguishable from `..`. Therefore I'd also go with bombless's idea of `..~`.

In conclusion I like the idea of inclusive range syntactic sugar, but the visual operator should be discussed more when the idea is through.

erikjohnston:
I'm also concerned that the syntax `...` will be confusing, especially for either a) people new to the language or b) someone not familiar with rust doing maintenance on an existing code base. Off by one errors are hard to track down to begin with, let alone when you don't even know about the distinction between `..` and `...`.  It's little things like this that will put off people using rust recreationally.

FWIW, i've been vaguely dabbling with rust for a few months now and its only after I saw this RFC that I realised why I could never remember if it was `..` or `...`.

(I'm *almost* half tempted by the idea of `[a..b)`, `[a..b]`, etc.)

rkjnsn:
:+1: Since we already have `...` for matches, I think it makes sense to have it for ranges, too.

liigo:
> My only problem is, that `...` is not enough distinguishable from `..`

`...` has one more dot than `..`, that is a significant difference.
Though `:` and `;` do have insignificant difference, we still use them at the same time.

One more dot, one more element in range, which i like. +1 to `...`

rkruppe:
@liigo One dot is not a significant difference. Although one dot versus none may be, a sequence of identical symbols is hard on the eyes (as an extreme example, consider 99999999 vs 999999999). The comparison with `:` and `;` does not hold water because the two are never interchangeable: First, context always tells me which one I should expect (`foo(bar):`, `let x; i32;`, `let x: <statement>`, `field; Type`, `<T; Debug>` and so on are all obviously wrong). Second, when I typo `:` for `;` or vice versa and not noticed it (because, as you point out, they look quite alike), then the parser points it out for me. On the other hand, `a..b` versus `a...b` are very similar in all aspects: In many contexts not even the type checker can help me when I wrote one and meant the other, the mistake will only manifest itself as an off-by-one error.

nagisa:
If we’re going with anything else than ellipsis, I hope its `.inclusive()` method on the half-open range.

withoutboats:
I agree with @rkruppe. This seems like a change that's guaranteed to cause a disproportionate number of off-by-one errors that will be difficult to identify and debug. If one were guaranteed never to accidentally type the wrong number of periods, this addition would be quite elegant and I definitely see the appeal of it. But as we are imperfect humans, `..`/`...` is a significance of syntax that I think Rust should avoid.

I also agree that an `.incluse()` method seems like the best alternative. Unusual syntaxes like `..=` and `..~` are not aesthetically pleasing to me.

glaebhoerl:
As a compromise option likely to please no one, we could just add even more periods to make them harder to confuse, like `....` ;)

nagisa:
How will ambiguous syntax be parsed in case we pick `...`?

Currently something like `2.....` fails with error about finding a `...`, however if `...` for inclusive ranges is agreed upon, you could parse the previous expression as either `(2..)...` or `(2...)..`. Will it make `2....` work (parsed as `(2..)..`)?

huonw:
@nagisa This won't change the tokeniser (which is essentially what controls all those cases), so `2 ... ..` for the first case, and error for the second (its token stream is `2 ... .`)

pix64:
What about [0..n]?

rkruppe:
`[0..n]` is currently an array of type `[Range<_>; 1]`. While that's probably not very useful (so you could argue for breaking backwards compatibility here), it seems very strange to me to make a particular instance of array literals do something completely differently. And `(0..n)` would become misleading, it could then be interpreted as an open range. I also have doubts whether such syntax could be reconciled with the desire to need very little and constant lookahead in the parser.

qres:
I also think that `..` and `...` look to similar and would prefer the `(a..b).inclusive()` suggestion. So there is only _one_ simple way to do these kind of ranges and it is clearer which range is actually used.
Additionnaly I think, that it also makes it easier and less confusing for new people to understand the two versions.

rhoark:
Typo waiting to happen. I suggest `a..+b`

Blackhole16:
@rhoark The problem with `a..+b` is, that it could be read by developers as `a..abs(b)`.
Even though rust does not support this syntax, it could lead to misunderstandings for developers not familiar with rust.

Otherwise I would +1 `a..+b` :)

olivren:
When I was writing Ruby code (a long time ago), I clearly remember being annoyed by the existence of two ellipsis versions (.. and ...) in the language. My problem was not to recognize which one was used (it's pretty clear when you use a monospace font), but to remember which one was the exclusive variant. Each time I read a line of code using it, I had to stop for 2 seconds to apply a mnemonic I made up, to answer the question "is this range inclusive or exclusive?".

The fact is, the meaning given to each variant is arbitrary. You could switch the two symbols and it would still be reasonable (like in Ruby). In current Rust, when I see an ellipsis I instantly know this is  an exclusive range, because the language has only exclusive range literals. To me, this is an easy rule to remember, because I know this that exclusive ranges are the most useful range type, and thus the logical choice. I admit this logic may not be as straightforward to a less experienced developer.

I think I like the proposed `.inclusive()` method. I don't have an opinion about the use of the ellipsis in patterns (never used them there).

pzol:
+1 for .inclusive()

Emerentius:
`(a..b).inclusive()` is like a caricature of Rust verbosity.
I imagine that almost everyone would simply use `a..b+1` for integers and `(a..b).inclusive()` for the rare occasion where that doesn't work, if they find the method in the docs before they hack something together.
For that reason I believe that an `.inclusive()` method would violate the single obvious way rule.

Ticki:
@Emerentius Wouldn't that still be a problem when having `...`?

Emerentius:
@Ticki Partially, yes. `a..b+1` will always be the most obvious way, but the verbosity of `.inclusive()` will hinder adoption. You use it less, other use it less, you don't see it as often.

Virtlink:
I do see use cases for inclusive ranges, but I don't see a clear winner in the proposed syntaxes. I especially dislike the `..` vs `...` syntax, for the following reasons:

1. As [rkruppe](https://github.com/rkruppe) points out, they are too similar. I'm afraid we're introducing off-by-one errors that are hard to spot.
2. The only other language I know that uses the ellipses is Ruby, and there `..` is inclusive and `...` exclusive, so the other way around. This gets confusing for some new users. I see at least one Rubyist that [doesn't like their syntax](https://www.reddit.com/r/ruby/comments/15omrb/why_is_ruby_range_syntax_this_way_15_exclusive/c7ohhfo) for being too similar either.

---

I think `.inclusive()` is the best proposal here currently.

And perhaps this inspires some syntax ideas:

	[a..b)    // from a (inclusive) to b (exclusive)
	[a..b]    // from a (inclusive) to b (inclusive)
	
I don't think the above syntax clashes with existing bracket syntax (such as tuples or array indices) due to the presence of `..` in there, but I'm probably wrong.

Blackhole16:
Actually I like `[a..b]` for exclusive and `[a..b)` for inclusive more, as `[a..b]` just looks better and is the 'normal' case, and `)` just looks rounder, has a bigger stomach and therefore contains b ;)

One problem could be, that it breaks current programs, as `a..b` would not work anymore.

But in general +1 for this other syntax.

tbu-:
@Virtlink That clashes with how we parse the source code. The first thing that is done is parsing the brackets, and for this step they must match.

erikjohnston:
@Blackhole16 `[a..b]` and `[a..b)` are standard maths notation for inclusive and exclusive. See: https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

Blackhole16:
@erikjohnston oh, damn. I didn't say anything ;)

Then it's my most preferred syntax atm. Will `[a..b)` just result in an array then? As `a..b` does atm? Because in that case, that syntax makes even more sense then the current one, indicating that it is an array!

nagisa:
* `[a..b)` is meaningless, because that’s exactly what `a..b` is and we ain’t getting rid of it;
* `[a..b]` is a value of type `[Range<T>; 1]` and cannot be used for closed intervals;
* `(a..b)` is a value of type `Range<T>` and cannot be used for open intervals;
* `(a..b]` is the only thing that you cannot exactly write using any of current syntaxes without doing the arithmetic yourself.

`[a..b]` and `(a..b)` and `[a..b)` therefore are not an option.

rkjnsn:
We already have `...` meaning inclusive for match patterns, so I think reusing that for inclusive ranges makes the most sense. I'd also like to see `..` usable in match expressions.

For those who find `..` vs. `...` confusing in Ruby, I think this is at least partially because their meaning in Ruby is backward from what is (or at least what I find) to be intuitive.

Virtlink:
@nagisa `a..b` could be syntactic sugar for `[a..b)`, to be consistent. And in the other cases there is alternative syntax that's semantically the same as the existing syntax that would get replaced (e.g. `[a..b; 1]`, when `[a..b]` denotes a closed interval).

However, this was just an idea. It may not be feasible from a backward-compatibility or syntactic standpoint. Perhaps there are other brackets that are similar enough and available, or there's a prefix, infix or suffix operator that can be used to distinguish between interval brackets and normal parentheses/brackets.

placrosse:
@rkjnsn +1 for a...b

Consistency with match patterns for both .. and ... will result in the least surprising behavior

liigo:
@rkruppe 

> as an extreme example, consider 99999999 vs 999999999

You are just quibbling. Yes 99999999 vs 999999999 is not a significant difference, so does `........` vs `.........`. But we are comparing `..` and `...`, *two* dots and *three* dots, not many dots.

gilescope:
+1 for ... as inclusive.

main--:
While I agree that the current proposal may lead to confusion because of the small visual difference, I really don't think there's any way around this unless patterns are changed as well.

Otherwise, the resulting syntax asymmetry is simply much worse.

glaebhoerl:
I think the fact that we already have `...` patterns is what sways me towards that syntax as well.

nikomatsakis:
My feeling is that we already had this debate back when we adopted `..` (for ranges) and `...` for patterns. I think the consistency there is important and worthwhile.

nikomatsakis:
**Huzzah!** The language design subteam [has decided][1] to accept this RFC. There were two major points of contention:

- whether or not `...` ought to be an iterator, or merely support into iterator. The decision was to make `...` an interator, which implies an extra field in the implementation to indicate termination.
- whether to accept the `...` syntax at all, or some other alternative such as an `inclusive` method. The conclusion was that consistency with `...` patterns was desirable, as well as concise notation for use cases like `vec.remove(1...5)`. 

[1]: https://docs.google.com/spreadsheets/d/18FtPnEI8RTPdFhGDtj21ITndakEtxVrs8pVzRNx-jL8/edit#gid=0

blaenk:
Awesome.

Stebalien:
So, after the fact (sorry if this is way to late but this hasn't been stabalized yet)... How about using an enum to represent `RangeInclusive`? For example:

```rust
enum RangeInclusive<T> {
    Empty,
    NonEmpty {
        start: T,
        end: T,
    }
}
```

Rational:

1. `finished` is very iterator specific. Regardless of what happens, I think this field should be called `empty`.
2. `start`/`end` don't make sense if the range is empty. Using an enum prevents coders from using the `start`/`end` of spent ranges. Basically, this makes it impossible for the coder to do something like `foo(my_range.take(10)); bar(my_range)` and forget to check `finished` in `bar`.
3. If we ever get more space optimizations (specifically, utf8 code point ones) `'a'...'z'` should be the same size as `'a'..'z'`.

purpleposeidon:
I rather missed the boat on this alternative syntax suggestion: `low ==> high`.

PR #1194, opened 2015-07-08
RFC: Add item recovery collection APIs

apasel422:
[Rendered](https://github.com/apasel422/rfcs/blob/collection-recovery/text/0000-set-recovery.md)

### DIFF
--- a/text/0000-collection-recovery.md
+++ b/text/0000-collection-recovery.md
@@ -11,7 +11,7 @@ in `std` in order to facilitate this.
 # Motivation
 
 Sets are sometimes used as a cache keyed on a certain property of a type, but programs may need to
-access the type's other properties for efficiency or functionailty. The sets in `std` do not expose
+access the type's other properties for efficiency or functionality. The sets in `std` do not expose
 their elements (by reference or by value), making this use-case impossible.
 
 Consider the following example:
--- a/text/0000-collection-recovery.md
+++ b/text/0000-collection-recovery.md
@@ -5,14 +5,14 @@
 
 # Summary
 
-Add item-recovery methods to the set types in `std`. Add key-recovery methods to the map types in
-`std` in order to facilitate this.
+Add element-recovery methods to the set types in `std`. Add key-recovery methods to the map types
+in `std` in order to facilitate this.
 
 # Motivation
 
 Sets are sometimes used as a cache keyed on a certain property of a type, but programs may need to
 access the type's other properties for efficiency or functionailty. The sets in `std` do not expose
-their items (by reference or by value), making this use-case impossible.
+their elements (by reference or by value), making this use-case impossible.
 
 Consider the following example:
 
@@ -74,27 +74,27 @@ fn main() {
     // replaced copper with zinc
     // ```
     //
-    // However, `HashSet` does not expose its items via its `{contains, insert, remove}` methods,
-    // instead providing only a boolean indicator of the item's presence in the set, preventing us
-    // from implementing the desired functionality.
+    // However, `HashSet` does not expose its elements via its `{contains, insert, remove}`
+    // methods,  instead providing only a boolean indicator of the elements's presence in the set,
+    // preventing us from implementing the desired functionality.
 }
 ```
 
 # Detailed design
 
-Add the following item-recovery methods to `std::collections::{BTreeSet, HashSet}`:
+Add the following element-recovery methods to `std::collections::{BTreeSet, HashSet}`:
 
 ```rust
 impl<T> Set<T> {
-    // Like `contains`, but returns a reference to the item if the set contains it.
-    fn item<Q: ?Sized>(&self, item: &Q) -> Option<&T>;
+    // Like `contains`, but returns a reference to the element if the set contains it.
+    fn element<Q: ?Sized>(&self, element: &Q) -> Option<&T>;
 
-    // Like `remove`, but returns the item if the set contained it.
-    fn remove_item<Q: ?Sized>(&mut self, item: &Q) -> Option<T>;
+    // Like `remove`, but returns the element if the set contained it.
+    fn remove_element<Q: ?Sized>(&mut self, element: &Q) -> Option<T>;
 
-    // Like `insert`, but replaces the item with the given one and returns the previous item if the
-    // set contained it.
-    fn replace(&mut self, item: T) -> Option<T>;
+    // Like `insert`, but replaces the element with the given one and returns the previous element
+    // if the set contained it.
+    fn replace(&mut self, element: T) -> Option<T>;
 }
 ```
 
@@ -118,8 +118,7 @@ impl<K, V> Map<K, V> {
 }
 ```
 
-For completion, add the following key-recovery methods to
-`std::collections::{btree_map, hash_map}::OccupiedEntry`:
+Add the following key-recovery methods to `std::collections::{btree_map, hash_map}::OccupiedEntry`:
 
 ```rust
 impl<'a, K, V> OccupiedEntry<'a, K, V> {
@@ -137,24 +136,7 @@ impl<'a, K, V> OccupiedEntry<'a, K, V> {
 }
 ```
 
-# Drawbacks
-
-This complicates the collection APIs.
-
-The distinction between `insert` and `replace` may be confusing. It would be more consistent to
-call `Set::replace` `Set::insert_item` and `Map::replace` `Map::insert_key_value`, but `BTreeMap`
-and `HashMap` do not replace equivalent keys in their `insert` methods, so rather than have
-`insert` and `insert_key_value` behave differently in that respect, `replace` is used instead.
-
-# Alternatives
-
-Do nothing.
-
-# Unresolved questions
-
-Are these the best method names?
-
-Should `std::collections::{btree_map, hash_map}::VacantEntry` provide methods like
+Add the following key-recovery methods to `std::collections::{btree_map, hash_map}::VacantEntry`:
 
 ```rust
 impl<'a, K, V> VacantEntry<'a, K, V> {
@@ -168,6 +150,23 @@ impl<'a, K, V> VacantEntry<'a, K, V> {
     fn into_key(self) -> K;
 }
 ```
+
+# Drawbacks
+
+This complicates the collection APIs.
+
+The distinction between `insert` and `replace` may be confusing. It would be more consistent to
+call `Set::replace` `Set::insert_element` and `Map::replace` `Map::insert_key_value`, but
+`BTreeMap` and `HashMap` do not replace equivalent keys in their `insert` methods, so rather than
+have `insert` and `insert_key_value` behave differently in that respect, `replace` is used instead.
+
+# Alternatives
+
+Do nothing.
+
+# Unresolved questions
+
+Are these the best method names?
 
 Should `{BTreeMap, HashMap}::insert` be changed to replace equivalent keys? This could break code
 relying on the old behavior, and would add an additional inconsistency to `OccupiedEntry::insert`.
--- a/text/0000-collection-recovery.md
+++ b/text/0000-collection-recovery.md
@@ -167,6 +167,3 @@ Do nothing.
 # Unresolved questions
 
 Are these the best method names?
-
-Should `{BTreeMap, HashMap}::insert` be changed to replace equivalent keys? This could break code
-relying on the old behavior, and would add an additional inconsistency to `OccupiedEntry::insert`.
--- a/text/0000-collection-recovery.md
+++ b/text/0000-collection-recovery.md
@@ -1,0 +1,173 @@
+- Feature Name: collection_recovery
+- Start Date: 2015-07-08
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Add item-recovery methods to the set types in `std`. Add key-recovery methods to the map types in
+`std` in order to facilitate this.
+
+# Motivation
+
+Sets are sometimes used as a cache keyed on a certain property of a type, but programs may need to
+access the type's other properties for efficiency or functionailty. The sets in `std` do not expose
+their items (by reference or by value), making this use-case impossible.
+
+Consider the following example:
+
+```rust
+use std::collections::HashSet;
+use std::hash::{Hash, Hasher};
+
+// The `Widget` type has two fields that are inseparable.
+#[derive(PartialEq, Eq, Hash)]
+struct Widget {
+    foo: Foo,
+    bar: Bar,
+}
+
+#[derive(PartialEq, Eq, Hash)]
+struct Foo(&'static str);
+
+#[derive(PartialEq, Eq, Hash)]
+struct Bar(u32);
+
+// Widgets are normally considered equal if all their corresponding fields are equal, but we would
+// also like to maintain a set of widgets keyed only on their `bar` field. To this end, we create a
+// new type with custom `{PartialEq, Hash}` impls.
+struct MyWidget(Widget);
+
+impl PartialEq for MyWidget {
+    fn eq(&self, other: &Self) -> bool { self.0.bar == other.0.bar }
+}
+
+impl Eq for MyWidget {}
+
+impl Hash for MyWidget {
+    fn hash<H: Hasher>(&self, h: &mut H) { self.0.bar.hash(h); }
+}
+
+fn main() {
+    // In our program, users are allowed to interactively query the set of widgets according to
+    // their `bar` field, as well as insert, replace, and remove widgets.
+
+    let mut widgets = HashSet::new();
+
+    // Add some default widgets.
+    widgets.insert(MyWidget(Widget { foo: Foo("iron"), bar: Bar(1) }));
+    widgets.insert(MyWidget(Widget { foo: Foo("nickel"), bar: Bar(2) }));
+    widgets.insert(MyWidget(Widget { foo: Foo("copper"), bar: Bar(3) }));
+
+    // At this point, the user enters commands and receives output like:
+    //
+    // ```
+    // > get 1
+    // Some(iron)
+    // > get 4
+    // None
+    // > remove 2
+    // removed nickel
+    // > add 2 cobalt
+    // added cobalt
+    // > add 3 zinc
+    // replaced copper with zinc
+    // ```
+    //
+    // However, `HashSet` does not expose its items via its `{contains, insert, remove}` methods,
+    // instead providing only a boolean indicator of the item's presence in the set, preventing us
+    // from implementing the desired functionality.
+}
+```
+
+# Detailed design
+
+Add the following item-recovery methods to `std::collections::{BTreeSet, HashSet}`:
+
+```rust
+impl<T> Set<T> {
+    // Like `contains`, but returns a reference to the item if the set contains it.
+    fn item<Q: ?Sized>(&self, item: &Q) -> Option<&T>;
+
+    // Like `remove`, but returns the item if the set contained it.
+    fn remove_item<Q: ?Sized>(&mut self, item: &Q) -> Option<T>;
+
+    // Like `insert`, but replaces the item with the given one and returns the previous item if the
+    // set contained it.
+    fn replace(&mut self, item: T) -> Option<T>;
+}
+```
+
+In order to implement the above methods, add the following key-recovery methods to
+`std::collections::{BTreeMap, HashMap}`:
+
+```rust
+impl<K, V> Map<K, V> {
+    // Like `get`, but additionally returns a reference to the entry's key.
+    fn key_value<Q: ?Sized>(&self, key: &Q) -> Option<(&K, &V)>;
+
+    // Like `get_mut`, but additionally returns a reference to the entry's key.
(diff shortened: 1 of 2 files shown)


Gankro:
CC @cmr @eddyb @bluss @seppo0010 (can't remember all the people interested...)

text/0000-collection-recovery.md
```diff
@@ -0,0 +1,173 @@
...
+This complicates the collection APIs.
+
+The distinction between `insert` and `replace` may be confusing. It would be more consistent to
+call `Set::replace` `Set::insert_item` and `Map::replace` `Map::insert_key_value`, but `BTreeMap`
+and `HashMap` do not replace equivalent keys in their `insert` methods, so rather than have
+`insert` and `insert_key_value` behave differently in that respect, `replace` is used instead.
+
+# Alternatives
+
+Do nothing.
```
pnkfelix: hold on, to be clear: "Do nothing" here means "Do nothing and let users write such caches via, e.g., `HashMap<T, ()>` " ... right?

I don't particular *mind* adding the functionality described here to `HashSet`, but I'm also not sure its strictly necessary, unless I have missed something with how `HashMap<T, ()>` would work.

apasel422: It's not possible to use `HashMap` that way, because it doesn't provide any methods that return `&K` (or `K`) other than via its iterators.

blaenk:
It'd be cool if you could add another copy of the code that demonstrates the problem but showing how to use your proposed APIs to resolve it.

apasel422:
@blaenk I'd actually like to use a more concrete motivating example, but I'll add the revised code as well.

apasel422:
@Gankro Do you have any ideas for a better motivating example, like an algorithm that uses a set as a cache?

eddyb:
@apasel422 that's the usecase in the compiler: sets of hundreds of thousands of elements, used for interning/caching, that have to be identity maps right now, wasting some memory space.

benaryorg:
@eddyb Does that mean, if this RFC is implemented, the compiler can be tweaked to use less memory?

apasel422:
I've added a WIP implementation of this RFC for `BTreeMap` at https://github.com/apasel422/rust/tree/rfc-1194.

Gankro:
@apasel422 It could be argued by metaphor to the current naming in https://github.com/rust-lang/rfcs/pull/1195 that these methods could just called `get_eq`, `remove_eq`, etc...

Gankro:
This is a bit more dubious for HashMap; but not crazy.

apasel422:
@Gankro I actually had that same thought a little while ago, but I don't think it fully translates, and would be even weirder for entries:

```rust
impl<'a, K, V> OccupiedEntry<'a, K, V> {
    fn get(&self) -> &V;
    fn get_eq(&self) -> (&K, &V); // what does eq have to do with this?
    ...
}
```

`{key_value, key_value_mut, remove_key_value}` are quite verbose, but have parallel set names `{element, remove_element}`. `{kv, kv_mut, remove_kv}` are more succinct, but don't have an obvious set version.

Gankro:
Hmm... Entry does seem to mess things up. That said; is it a tragedy if it's a bit misaligned from everything else?

apasel422:
I think they should be consistent.

Here are some options that work for both maps and occupied entries (in addition to `Map::replace`, which I think is non-controversial):

1. `{kv, kv_mut, remove_kv}` with `VacantEntry::insert_kv`
2. `{get_kv, get_kv_mut, remove_kv}` with `VacantEntry::insert_kv`
3. `{key_val, key_val_mut, remove_key_val}` with `VacantEntry::insert_key_val`
4. `{key_value, key_value_mut, remove_key_value}` with `VacantEntry::insert_key_value`

And here are some options for sets (assuming that the changes in rust-lang/rust#27135 canonicalize "element" over "value" when referring to sets):

1. `{elem, remove_elem}`
2. `{element, remove_element}`

i30817:
One common optimization that can't be done in java because of set item recovery is to just store hashes instead of elements, for the case where identity hash-mapping is not desirable. Are you going to give up this special case?

apasel422:
@i30817 Rust's `HashMap` and `HashSet` already prevent that optimization: The keys (elements) must implement `Eq` in addition to `Hash`, and the types provide iterators over their contents. A hypothetical map or set providing that optimization could omit these operations, but would also have to omit iteration. In any case, just storing hashes seems more like a bloom filter than a hash table, due to the absence of a check on `Eq`.

i30817:
Mmm makes sense. Still, it's a somewhat common optimization, maybe a bloom filter type could be added to the language.

apasel422:
@i30817 Off-topic, but [https://crates.io/search?q=bloom filter](https://crates.io/search?q=bloom filter)

shepmaster:
I'm definitely in favor of the ideas laid out here. I have the same motivating problem - a cache of strings. I've used some unsafe code to avoid double-allocating the strings, but I still have essentially a `HashMap<&str, &str>`.

bkoropoff:
How would you feel about modifying `OccupiedEntry` to hold on to the key passed to `entry` so you can recover it?

apasel422:
@bkoropoff That could be done, but it has the problem that a new key is only present for `OccupiedEntry`s that are obtained through `Map::entry`. The theoretical `{max_entry, min_entry, ...}` provided by RFC #1195 will not have a new key to return.

I'm therefore more inclined to add that kind of key-recovery functionality as

```rust
pub enum Entry<'a, K: 'a, V: 'a> {
    Occupied(OccupiedEntry<'a, K, V>, K),
    Vacant(VacantEntry<'a, K, V>),
}
```

instead of

```rust
pub struct OccupiedEntry<'a, K: 'a, V: 'a> {
    new_key: K,
    // ...
}

impl<'a, K, V> OccupiedEntry<'a, K, V> {
    /// Returns the key that was used to acquire this entry.
    // This could return `Option<K>` in order to better model the `max_entry` situation
    pub fn into_new_key(self) -> K { self.key }
}

```
but that would not be a backwards-compatible change. Additionally, storing the new key in the struct itself has the benefit of allowing us to provide an additional `OccupiedEntry::replace_key` method that updates the new key and returns the old one, much as `Map::replace` would:

```rust
impl<'a, K, V> OccupiedEntry<'a, K, V> {
    /// Replaces the entry's key with the one that was used to acquire this entry, if any, and
    /// returns the old key.
    ///
    /// This method always return `None` after the first call to it and for all entries
    /// acquired through `max_entry` etc.
    pub fn replace_key(&mut self) -> Option<K>;
}
```

This adds some complexity to the API surface and makes it harder to reason about what the behavior is. It's possible that we could add what you're proposing in a subsequent RFC instead.

Gankro:
:bell: HERE YE HERE YE THIS RFC IS ENTERING ITS FINAL COMMENT PERIOD :bell:

alexcrichton:
I agree with the motivation of this RFC, and I would indeed like to see it implemented. I'm not 100% comfortable with the names right now, but this will initially land as unstable anyway so we'll have time to change them.

Some thoughts about naming:

* Adding `replace` to maps/sets seems a bit innocuous, it's pretty similar to the existing `insert` methods. The consistency between collections is nice, however.
* Getting "the thing" back differs between sets/maps where sets are `element` and maps are `key_value` (as a base name for methods). We've mostly avoided these kinds of base names for various methods and it might be nice to avoid them if possible. For example `HashSet::element` could perhaps be `HashSet::get` and `remove_element` could be something like `delete` (or just a synonym for remove). Maps could then perhaps follow suit somewhat.
* The `key_value` method on maps should probably have "get" somewhere in the name as it's basically the same operation as `get`.

Unfortunately I don't have any concrete suggestions right now as to what I might prefer the names to be, but I'm ok landing this as `#[unstable]` to get some usage on nightly.

apasel422:
@alexcrichton Thanks for the feedback. I agree that it would be nice to avoid the base name paradigm, but I was also trying to go for consistency between the `Map` and `Entry` APIs. How would you feel about the following? (And as you said, we can mess around with the names while these are unstable.)

```rust
impl<K, V> Map<K, V> {
    fn find<Q: ?Sized>(&self, key: &Q) -> Option<(&K, &V)>;
    fn find_mut<Q: ?Sized>(&mut self, key: &Q) -> Option<(&K, &mut V)>;
    fn delete<Q: ?Sized>(&mut self, key: &Q) -> Option<(K, V)>;
    fn replace(&mut self, key: K, val: V) -> Option<(K, V)>;
}

impl<T> Set<T> {
    fn find<Q: ?Sized>(&self, key: &Q) -> Option<&T>;
    fn delete<Q: ?Sized>(&mut self, key: &Q) -> Option<T>;
    fn replace(&mut self, elem: T) -> Option<T>;
}

impl<'a, K, V> OccupiedEntry<'a, K, V> {
    fn key_val(&self) -> (&K, &V); // or `key_value`
    fn key_val_mut(&mut self) -> (&K, & mut V); // ditto
    fn into_key_val_mut(self) -> (&'a K, &'a mut V); // ditto
    fn delete(self) -> (K, V);
}

impl<'a, K, V> VacantEntry<'a, K, V> {
    fn key(&self) -> &K;
    fn into_key(self) -> K;
    fn insert_key_val(self, val: V) -> (&'a K, &'a mut V); // `insert_key_val_mut` would be more consistent
}
```

alexcrichton:
Hm interesting! I know we axed `find` before in favor of `get` to be consistent with vectors, but this seems like a relatively appropriate use of it... It does have a bit of a danger of being easily confused with `get` (I still write `find` before the compiler tells me to use `get` today), but perhaps it's not the end of the world.

I do kinda like `delete` as the "this is remove but give me back as much as possible" method.

I think that `Set::find` should perhaps be `get`, however, as `get` is the universal "get an optional thing from this collection" accessor, and the name isn't already taken for another method on sets.

I'm still a little on the fence about the entry methods, I feel like `key_val` kinda sticks out. Perhaps `pair` could be used there, but that also may still feel a bit odd to me...

(sorry if this isn't the most helpful comment!)

apasel422:
An alternative to `find` is `retrieve`. I agree that `key_val` is a bit strange, so `pair` is an option.

apasel422:
This seems relatively consistent across the board:

```rust
impl<K, V> Map<K, V> {
    fn get_pair<Q: ?Sized>(&self, key: &Q) -> Option<(&K, &V)>;
    fn get_pair_mut<Q: ?Sized>(&mut self, key: &Q) -> Option<(&K, &mut V)>;
    fn delete<Q: ?Sized>(&mut self, key: &Q) -> Option<(K, V)>;
    fn replace(&mut self, key: K, val: V) -> Option<(K, V)>;
}

impl<T> Set<T> {
    fn get<Q: ?Sized>(&self, key: &Q) -> Option<&T>;
    fn delete<Q: ?Sized>(&mut self, key: &Q) -> Option<T>;
    fn replace(&mut self, elem: T) -> Option<T>;
}

impl<'a, K, V> OccupiedEntry<'a, K, V> {
    fn pair(&self) -> (&K, &V);
    fn pair_mut(&mut self) -> (&K, &mut V);
    fn into_pair_mut(self) -> (&'a K, &'a mut V);
    fn delete(self) -> (K, V);
}

impl<'a, K, V> VacantEntry<'a, K, V> {
    fn key(&self) -> &K;
    fn into_key(self) -> K;
    fn insert_pair(self, val: V) -> (&'a K, &'a mut V);
}
```

sfackler:
rust-phf has had some of this functionality for a while, primarily for use in interning schemes: http://sfackler.github.io/rust-phf/doc/phf/struct.Map.html#method.get_key http://sfackler.github.io/rust-phf/doc/phf/struct.Set.html#method.get_key

glaebhoerl:
What about `take` instead of `delete`?

frankmcsherry:
Generally, ++ on the change, but with some comments about methodolgy:

I don't know if this is on the table but: it would also be great if `Entry`'s `or_insert_with` method got extended to take a reference to the key. This is roughly what you want/need for any memoization, and is the source of several key clones in my code.

Random thought: is there a need to literally add these to `HashMap` et al vs provide a more feature complete hash map (and friends) implementation, not as part of the stdlib, and make the existing `HashMap` and its methods wrap the new one with their stability-guaranteed-but-feature-limited methods? Just toss it up on crates.io as "LinearHashMap" or something, and make the RFCs be for when it is critical to add methods to `HashMap` backed by this implementation. For example, maybe it is too hard to work with `HashMap`s others provide, so having these methods on all `HashMap`s would be best, but if one just wraps the other it would be trivial to convert between them.

0. Although I like all these methods more than those of the current `HashMap`, people who want these methods are probably more ownership saavy than normal users, and know how crates.io works. I didn't get from the RFC why it was important to add these methods to the stdlib. 
1. It is a pity having the funny names just because the previous `HashMap` squatted the good ones. 
2. I would be more likely to test this out if I didn't have to wager on whether / how it would eventually stabilize, and move off of stable for a few months (or maintain several forks). It takes about a day for me to re-thread all of the ownership changes through my code, audit for extra uses of `clone`, etc.
3. There should just be a good linear-type friendly version of each of these datastructures somewhere without the constraints of backwards compatibility, name bikeshedding, and forward stability guarantees. I'd really like that to exist, almost more than I'd like the stdlib to get one with 90% of the methods.

jmesmon:
I'd also like this. Having to clone keys out of my structs for HashMap has never made me happy.

arthurprs:
I'd appreciate this inclusion.

muja:
I have to agree with @alexcrichton on the names, especially the `remove_element` doesn't sound any different than `remove` in my eyes so I would confuse them every single time and have to look it up. Naming it `remove_get` would make it clearer for example.  

On different note: Was there a reason why `remove` wasn't implemented to return `Option<T>`? It has all information a bool has and more.

Gankro:
@muja A set is a trivial wrapper around a map, and maps don't give that information.

arthurprs:
I'm not really into bike-shedding today, but the most readable/consistent option to me is the _pair suffixes as @apasel422 suggested with the exception of Map<>.delete() which should probably be remove_pair(). The only inconsistency left then is Set<>.remove/delete() which is kinda unfortunate.

Gankro:
The libs team has decided to merge this with precise naming to be re-evaluated when the feature is brought up for stabilization. (scared to push merge -- leaving that to Alex because The Process)

cristicbz:
I like the latest names (`pair`-s) with exception of `delete`. How about merging with @glaebhoerl 's suggestion of `take` and call it `take_pair`?

aturon:
Sorry to be late to this party (I also had to miss the libs meeting this week). I'm on board with the basic motivation here, and regret the stabilization of the `bool`-centric methods on sets. As @Gankro said, the team decided to go forward with this, modulo some bikeshedding.

That said, I feel like the RFC is proposing significantly more API expansion than is actually needed to solve the original problem -- in particular, I don't see why any changes to the entry API are needed. Could we instead take the following as a starting point (bikesheds painted in my favorite colors):

```rust
impl<T> Set<T> {
    // Like `contains`, but returns a reference to the element if the set contains it.
    fn get<Q: ?Sized>(&self, element: &Q) -> Option<&T>;

    // Like `remove`, but returns the element if the set contained it.
    fn take<Q: ?Sized>(&mut self, element: &Q) -> Option<T>;

    // Like `insert`, but replaces the element with the given one and returns the previous element
    // if the set contained it.
    fn replace(&mut self, element: T) -> Option<T>;
}

impl<K, V> Map<K, V> {
    // Like `get`, but additionally returns a reference to the entry's key.
    fn key_value<Q: ?Sized>(&self, key: &Q) -> Option<(&K, &V)>;

    // Like `get_mut`, but additionally returns a reference to the entry's key.
    fn key_value_mut<Q: ?Sized>(&mut self, key: &Q) -> Option<(&K, &mut V)>;

    // Like `remove`, but additionally returns the entry's key.
    fn remove_key_value<Q: ?Sized>(&mut self, key: &Q) -> Option<(K, V)>;

    // Like `insert`, but additionally replaces the key with the given one and returns the previous
    // key and value if the map contained it.
    fn replace(&mut self, key: K, value: V) -> Option<(K, V)>;
}
```

In particular, the fact that the entry APIs need an *owned* key to use (today, at least) seems to make the key-accessing functionality questionable. But maybe I'm missing something?

apasel422:
@aturon We will presumably want the entry methods once #1195 is accepted, but they could be omitted for now. I think that both RFCs need to be considered together though, and it probably makes sense to avoid a proliferation of `{get, get_mut, remove, and entry}` methods when just `{get, entry}` suffice. We don't need to add `take` or `key_value_mut` on `Map` directly if we just add them to `OccupiedEntry` now, and this will benefit the queries added in #1195 as well. For example:

```rust
impl<K, V> Map<K, V> {
    fn get_pair(&self, key: &Q) -> Option<(&K, &V)>;
    fn get_entry(&mut self, key: &Q) -> Option<OccupiedEntry<K, V>>;
    fn replace(&mut self, key: K, val: V) -> Option<(K, V)>;

    fn get_max(&self) -> Option<(&K, &V)>;
    fn max_entry(&mut self) -> Option<OccupiedEntry<K, V>>>;

    fn get_lt(&self, key: &Q) -> Option<(&K, &V)>;
    fn lt_entry(&mut self, key: &Q) -> Option<OccupiedEntry<K, V>>>;

    // get_* and *_entry for le, ge, gt, min
}

impl<'a, K, V> OccupiedEntry<'a, K, V> {
    fn pair(&self) -> (&K, &V);
    fn pair_mut(&mut self) -> (&K, &mut V);
    fn into_pair_mut(self) -> (&'a K, &'a mut V);
    fn take(self) -> (K, V);
}
```

alexcrichton:
The libs team discussed this RFC today, and our conclusion was that it may be best to hone this down to what's precisely necessary to satisfy the motivation in the outset. To that end would it be possible to *only* include the set methods? Specifically:

```rust
impl<T> Set<T> {
    fn get<Q: ?Sized>(&self, element: &Q) -> Option<&T>;
    fn take<Q: ?Sized>(&mut self, element: &Q) -> Option<T>;
    fn replace(&mut self, element: T) -> Option<T>;
}
```

Gankro:
Specifically, I believe the supporting map methods were also decided to just be `doc(hidden)` and unstable.

alexcrichton:
I'd personally prefer the methods to be freestanding in the module so they're private to the outside world but public to the crate rather than having them in the inherent API at all.

Gankro:
@alexcrichton How does that work? Privacy can only reach up, and not down or sideways. Maps and Sets are defined in sibling modules.

sfackler:
They could be defined in a crate private trait. That's how I've gotten
around visibility issues before.

Gankro:
@apasel422 Can you amend the RFC to be minimal per aturon's request? I think we're good to go when that's done.

seppo0010:
I don't understand the motivation to allow item recover from a `Set` but not key recover from a `Map`.

I was actually expecting that feature to move items from one `Map` to another without cloning its keys.

SimonSapin:
Same here, the use case that got me here was with a `HashMap`, not a set.

eddyb:
@seppo0010 The usecase I hit that needed the feature for `Set` was sort of a cache.
One quick example would be `HashSet<Rc<str>>`, you want to be able to find an existing key with `&str` and clone the `Rc` handle.
To get closer to the compiler, `HashSet<&'arena str>` could be used as a cache, queried with a temporary `&str`, that would get copied on the arena if not found in the set.

alexcrichton:
@apasel422 ping about the RFC update, would love to merge!

apasel422:
@alexcrichton I haven't updated yet because it seems like there's still some dissent, based on the last few comments.

shepmaster:
As another voice, only having it on sets would be acceptable for me. I am in the same boat as @eddyb — a cache.

apasel422:
@alexcrichton Updated.

alexcrichton:
Ok, thanks @apasel422! The consensus of the libs team is that this is a great step forward for sets and we can continue to explore the problem space for maps as the needs arise, but it seems like the most pressing parts to work with are sets today.

And of course, thanks again for the RFC @apasel422!

rpjohnst:
I find myself needing this for maps.

In my case, I am building a string interner using a `HashMap<Box<str>, u32>` where the `u32`s track the order of insertion. I use this to determine whether a string is in one of a few fixed sets of symbols by doing range checks on its corresponding `u32`.

The [`OrderMap` crate](https://crates.io/crates/ordermap) provides this part of the API under the name `get_pair`, so that name now has some precedent.

Alternatively, it would be useful for the `Entry` API to provide a reference to the key as well as the value after an `or_insert`/`or_insert_with`-like operation. This might be the better option since it allows the lookup to be reused when inserting new entries.

What is the best route forward here? Should I write up a new RFC?

fschutt:
Well, I needed this for a function where I insert into a set, but then I immediately want an iterator to that last, inserted element in the set. Since a `BTreeSet` is just a BTreeMap internally, this should be possible. 

The application is a scanline algorithm, where the set consists out of ordered points. I need to insert a point into a scanline and then know where it has been inserted (the position), so that I can construct an iterator to the next and previous point in the (ordered) scanline.

So for now I've forked the `std::collections::BTreeSet`. Watching this for when it eventually gets merged.

PR #1199, opened 2015-07-10
SIMD groundwork

huonw:
[Rendered](https://github.com/huonw/rfcs/blob/simd/text/0000-simd-infrastructure.md).

### DIFF
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -196,16 +196,10 @@ platform specific intrinsic for shuffling.
 
 ```rust
 extern "platform-intrinsic" {
-    fn simd_shuffle2<T, Elem>(v: T, w: T, i0: u32, i1: u32) -> Simd2<Elem>;
-    fn simd_shuffle4<T, Elem>(v: T, w: T, i0: u32, i1: u32, i2: u32, i3: u32) -> Sidm4<Elem>;
-    fn simd_shuffle8<T, Elem>(v: T, w: T,
-                              i0: u32, i1: u32, i2: u32, i3: u32,
-                              i4: u32, i5: u32, i6: u32, i7: u32) -> Simd8<Elem>;
-    fn simd_shuffle16<T, Elem>(v: T, w: T,
-                               i0: u32, i1: u32, i2: u32, i3: u32,
-                               i4: u32, i5: u32, i6: u32, i7: u32
-                               i8: u32, i9: u32, i10: u32, i11: u32,
-                               i12: u32, i13: u32, i14: u32, i15: u32) -> Simd16<Elem>;
+    fn simd_shuffle2<T, Elem>(v: T, w: T, idx: [i32; 2]) -> Simd2<Elem>;
+    fn simd_shuffle4<T, Elem>(v: T, w: T, idx: [i32; 4]) -> Sidm4<Elem>;
+    fn simd_shuffle8<T, Elem>(v: T, w: T, idx: [i32; 8]) -> Simd8<Elem>;
+    fn simd_shuffle16<T, Elem>(v: T, w: T, idx: [i32; 16]) -> Simd16<Elem>;
 }
 ```
 
@@ -214,10 +208,8 @@ time, ensure that `T` is a SIMD vector, `Elem` is the element type of
 `T` etc. Libraries can use traits to ensure that these will be
 enforced by the type checker too.
 
-This approach has some downsides: `simd_shuffle32` (e.g. `Simd32<u8>`
-on AVX, and `Simd32<u16>` on AVX-512) and especially `simd_shuffle64`
-(e.g. `Simd64<u8>` on AVX-512) are unwieldy. These have similar type
-"safety"/code-generation errors to the vectors themselves.
+This approach has similar type "safety"/code-generation errors to the
+vectors themselves.
 
 These operations are semantically:
 
@@ -225,10 +217,10 @@ These operations are semantically:
 // vector of double length
 let z = concat(v, w);
 
-return [z[i0], z[i1], z[i2], ...]
+return [z[idx[0]], z[idx[1]], z[idx[2]], ...]
 ```
 
-The indices `iN` have to be compile time constants. Out of bounds
+The index array `idx` has to be compile time constants. Out of bounds
 indices yield unspecified results.
 
 Similarly, intrinsics for inserting/extracting elements into/out of
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -276,6 +276,33 @@ of the arithmetic overflow checks of `-C debug-assertions`): explicit
 SIMD is essentially only required for speed, and checking inflates one
 instruction to 5 or more.
 
+### Why not inline asm?
+
+One alternative to providing intrinsics is to instead just use
+inline-asm to expose each CPU instruction. However, this approach has
+essentially only one benefit (avoiding defining the intrinsics), but
+several downsides, e.g.
+
+- assembly is generally a black-box to optimisers, inhibiting
+  optimisations, like algebraic simplification/transformation,
+- programmers would have to manually synthesise the right sequence of
+  operations to achieve a given shuffle, while having a generic
+  shuffle intrinsic lets the compiler do it (NB. the intention is that
+  the programmer will still have access to the platform specific
+  operations for when the compiler synthesis isn't quite right),
+- inline assembly is not currently stable in
+  Rust and there's not a strong push for it to be so in the immediate
+  future (although this could change).
+
+Benefits of manual assembly writing, like instruction scheduling and
+register allocation don't apply to the (generally) one-instruction
+`asm!` blocks that replace the intrinsics (they need to be designed so
+that the compiler has full control over register allocation, or else
+the result will be strictly worse). Those possible advantages of hand
+written assembly over intrinsics only come in to play when writing
+longer blocks of raw assembly, i.e. some inner loop might be faster
+when written as a single chunk of asm rather than as intrinsics.
+
 ## Platform Detection
 
 The availability of efficient SIMD functionality is very fine-grained,
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -348,6 +348,11 @@ cfg_if_else! {
   to be worth it. (Each "feature" would essentially be a platform
   specific `cfg` anyway.)
 - Check vector operators in debug mode just like the scalar versions.
+- Make fixed length arrays `repr(simd)`-able (via just flattening), so
+  that, say, `#[repr(simd)] struct u32x4([u32; 4]);` and
+  `#[repr(simd)] struct f64x8([f64; 4], [f64; 4]);` etc works. This
+  will be most useful if/when we allow generic-lengths, `#[repr(simd)]
+  struct Simd<T, n>([T; n]);`
 
 # Unresolved questions
 
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -19,6 +19,24 @@ This RFC lays the ground-work for building nice SIMD functionality,
 but doesn't fill everything out. The goal here is to provide the raw
 types and access to the raw instructions on each platform.
 
+## Where does this code go? Aka. why not in `std`?
+
+This RFC is focused on building stable, powerful SIMD functionality in
+external crates, not `std`.
+
+This makes it much easier to support functionality only "occasionally"
+available with Rust's preexisting `cfg` system. If it were in `std`,
+there would need to be some highly delayed `cfg` system so that
+functions that only work with (say) AVX-2 support:
+
+- don't break compilation on systems that don't support it, but
+- are still usable on systems that do support it.
+
+With an external crate, we can leverage `cargo`'s existing build
+infrastructure: compiling with some target features will rebuild with
+those features enabled.
+
+
 # Detailed design
 
 The design comes in three parts:
@@ -39,113 +57,42 @@ There is definitely a common core of SIMD functionality shared across
 many platforms, but this RFC doesn't try to extract that, it is just
 building tools that can be wrapped into a more uniform API later.
 
-## Background: Where does this code go?
-
-This RFC is focused on building stable, powerful SIMD functionality in
-external crates, not `std`. This makes it much easier to support
-functionality only "occasionally" available with Rust's preexisting
-`cfg` system. If it were in `std`, there would need to be some highly
-delayed `cfg` system so that functions that only work with AVX-2
-support:
-
-- don't break compilation on systems that don't support it, but
-- are still usable on systems that do support it.
 
 ## Types & traits
 
-A type designed to be used as a SIMD vector is indicated by the
-`repr(simd)` attribute. A type marked as such will be compiled to
-behave like a SIMD register (as well as the target platform can
-support it).
-
-The types/traits will be defined as follows:
+There are two new attributes: `repr(simd)` and `simd_primitive_trait`
 
 ```rust
 #[repr(simd)]
-struct Simd2<T: SimdPrim>(T, T);
+struct f32x4(f32, f32, f23, f23);
+
 #[repr(simd)]
-struct Simd4<T: SimdPrim>(T, T, T, T);
-#[repr(simd)]
-struct Simd8<T: SimdPrim>(T, T, T, T, T, T, T, T);
-#[repr(simd)]
-struct Simd16<T: SimdPrim>(T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T);
-#[repr(simd)]
-struct Simd32<T: SimdPrim>(T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T,
-                           T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T);
-#[repr(simd)]
-struct Simd64<T: SimdPrim>(T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T,
-                           T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T,
-                           T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T,
-                           T, T, T, T, T, T, T, T, T, T, T, T, T, T, T, T);
+struct Simd2<T>(T, T);
 
-trait SimdVector {
-    type Elem: SimdPrim;
-    type Bool: SimdVector<Elem = <Self::Elem as SimdPrim>::Bool>;
-}
-
-impl<T: SimdPrim> for Simd2<T> {
-    type Elem = T;
-    type Bool = Simd2<T::Bool>;
-}
-impl<T: SimdPrim> for Simd4<T> {
-    type Elem = T;
-    type Bool = Simd4<T::Bool>;
-}
-// ...
-impl<T: SimdPrim> for Simd64<T> {
-    type Elem = T;
-    type Bool = Simd64<T::Bool>;
-}
-
-#[simd_prim_trait]
-trait SimdPrim {
-    type Bool: SimdPrim;
-}
-
-// boolean types, see below
-struct bool8i(...);
-struct bool16i(...);
-struct bool32i(...);
-struct bool64i(...);
-struct bool32f(...);
-struct bool64f(...);
-
-// specifying what types are SIMD-able.
-impl SimdPrim for u8 { type Bool = bool8i; }
-impl SimdPrim for i8 { type Bool = bool8i; }
-impl SimdPrim for u16 { type Bool = bool16i; }
-// ...
-impl SimdPrim for i64 { type Bool = bool64i; }
-
-impl SimdPrim for f32 { type Bool = bool32f; }
-impl SimdPrim for f64 { type Bool = bool64f; }
-
-impl SimdPrim for bool8i { type Bool = bool8i; }
-// ...
-impl SimdPrim for bool64i { type Bool = bool64i; }
-
-impl SimdPrim for bool32f { type Bool = bool32f; }
-impl SimdPrim for bool64f { type Bool = bool64f; }
+#[simd_primitive_trait]
+trait SimdPrim {}
 ```
-
-It is illegal to take an internal reference to the fields of a
-`repr(simd)` type.
 
 ### `repr(simd)`
 
 The `simd` `repr` can be attached to a struct and will cause such a
-struct to be compiled to a SIMD vector. It is required that the
-monomorphised vector consist of only a single "primitive" type,
-repeated some number of times. The restrictions on the element type
-are exactly the same restrictions as `#[simd_primitive_trait]` traits
-impose on their implementing types.
+struct to be compiled to a SIMD vector. It can be generic, but it is
+required that any fully monomorphised instance of the type consist of
+only a single "primitive" type, repeated some number of times. The
+restrictions on the element type are exactly the same restrictions as
+`#[simd_primitive_trait]` traits impose on their implementing types.
 
 The `repr(simd)` may not enforce that the trait bound exists/does the
 right thing at the type checking level for generic `repr(simd)`
 types. As such, it will be possible to get the code-generator to error
 out (ala the old `transmute` size errosr), however, this shouldn't
 cause problems in practice: libraries wrapping this functionality
-would layer type-safety on top (i.e. the `SimdPrim` trait).
+would layer type-safety on top (i.e. generic `repr(simd)` types would
+use the `SimdPrim` trait as a bound).
+
+It is illegal to take an internal reference to the fields of a
+`repr(simd)` type, because the representation of booleans may require
+to change, so that booleans are bit-packed.
 
 ### `simd_primitive_trait`
 
@@ -159,35 +106,6 @@ distinct traits in a compilation. The attribute just adds the
 restriction and possibly tweaks type's internal representation (as
 such, it's legal for a single type to implement multiple traits with
 the attribute, if a bit pointless).
-
-### Booleans
-
-SIMD booleans are non-trivial. Many conventional APIs e.g. SSE, and
-NEON, use "wide booleans": a large number of bits set to all-zeros
-(false) or all-ones, e.g. equality between `Simd4(0_u32, 1, 2, 3)` and
-`Simd4(0_u32, 0, 2, 3)` gives (on the CPU) `Simd4(!0_u32, 0, !0,
-!0)`. Hence, the boolean types need to have width. It's tempting to
-just use the integer types of the appropriate width, but this falls
-down for two reasons:
-
-1. booleans aren't always this format
-2. the source of the boolean matters
-
-The second is easiest: CPUs are complicated beasts, and the hardware
-that handles floating point vector operations may be very different to
-the hardware that handles integer ones: instructions use different
-execution units. It can take several cycles to transfer data between
-them. Encoding the provenance/execution unit of the value in the type
-makes costs explicit.
-
-The first is much harder to solve. Some architectures/instruction sets
-model booleans as single bits. For example, equality between
-`Simd4(0_u32, 1, 2, 3)` and `Simd(0_u32, 0, 2, 3)` gives `1 + 4 + 8 ==
-0b1101`. One example is AVX-512 which essentially replaces all of the
-older SSE through AVX2 boolean-returning instructions with versions
-that return those. Using separate types for booleans (and restricting
-their API) allows for some serious magic: `Simd4<bool32>` becomes
-`u4`. (This is where the reference-restriction above comes in.)
 
 ## Operations
 
@@ -295,19 +213,23 @@ Comparisons are implemented via intrinsics, because the current
 comparison operator infrastructure doesn't easily lend itself to
 return vectors, as required.
 
-A library could give signatures like:
+The raw signatures would look like:
 
 ```rust
 extern "rust-intrinsic" {
-    fn simd_eq<T: SimdVector>(v: T, w: T) -> T::Bool;
-    fn simd_ne<T: SimdVector>(v: T, w: T) -> T::Bool;
-    fn simd_lt<T: SimdVector>(v: T, w: T) -> T::Bool;
-    fn simd_le<T: SimdVector>(v: T, w: T) -> T::Bool;
-    fn simd_gt<T: SimdVector>(v: T, w: T) -> T::Bool;
-    fn simd_ge<T: SimdVector>(v: T, w: T) -> T::Bool;
+    fn simd_eq<T, U>(v: T, w: T) -> U;
+    fn simd_ne<T, U>(v: T, w: T) -> U;
+    fn simd_lt<T, U>(v: T, w: T) -> U;
+    fn simd_le<T, U>(v: T, w: T) -> U;
+    fn simd_gt<T, U>(v: T, w: T) -> U;
+    fn simd_ge<T, U>(v: T, w: T) -> U;
 }
 ```
 
+However, these will be type checked, to ensure that `T` and `U` are
+the same length, and that `U` is appropriately shaped for a boolean. A
+library actually importing them might use some trait bounds to get
+actual type-safety.
 
 ### Built-in functionality
 
@@ -340,8 +262,10 @@ exact target e.g.
 - compiling with `-C target-cpu=native` on a modern CPU might set
   `target_feature = "avx2"`, `target_feature = "avx"`, ...
 
-(There are other non-SIMD features that might have `target_feature`s
-set too, such as `popcnt` and `rdrnd` on x86/x86-64.)
+The possible values of `target_feature` will be a selected whitelist,
+not necessarily just everything LLVM understands. There are other
+non-SIMD features that might have `target_feature`s set too, such as
+`popcnt` and `rdrnd` on x86/x86-64.)
 
 With a `cfg_if_else!` macro that expands to the first `cfg` that is
 satisfied (ala [@alexcrichton's cascade][cascade]), code might look
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -87,12 +87,6 @@ would layer type-safety on top (i.e. generic `repr(simd)` types would
 use some `unsafe` trait as a bound that is designed to only be
 implemented by types that will work).
 
-It is illegal to take an internal reference to the fields of a
-`repr(simd)` type, because the representation of booleans may require
-modification, so that booleans are bit-packed. The official external
-library providing SIMD support will have private fields so this will
-not be generally observable.
-
 Adding `repr(simd)` to a type may increase its minimum/preferred
 alignment, based on platform behaviour. (E.g. x86 wants its 128-bit
 SSE vectors to be 128-bit aligned.)
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -324,11 +324,11 @@ not necessarily just everything LLVM understands. There are other
 non-SIMD features that might have `target_feature`s set too, such as
 `popcnt` and `rdrnd` on x86/x86-64.)
 
-With a `cfg_if_else!` macro that expands to the first `cfg` that is
-satisfied (ala [@alexcrichton's cascade][cascade]), code might look
+With a `cfg_if!` macro that expands to the first `cfg` that is
+satisfied (ala [@alexcrichton's `cfg-if`][cfg-if]), code might look
 like:
 
-[cascade]: https://github.com/alexcrichton/backtrace-rs/blob/03703031babfa87cbe2c723ad6752131819dc554/src/macros.rs
+[cfg-if]: https://crates.io/crates/cfg-if
 
 ```rust
 cfg_if_else! {
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -64,7 +64,7 @@ There are two new attributes: `repr(simd)` and `simd_primitive_trait`
 
 ```rust
 #[repr(simd)]
-struct f32x4(f32, f32, f23, f23);
+struct f32x4(f32, f32, f32, f32);
 
 #[repr(simd)]
 struct Simd2<T>(T, T);
@@ -261,9 +261,10 @@ SIMD vectors can be converted with `as`. As with intrinsics, this is
 their lengths match and their elements are castable (i.e. are
 primitives), there's no enforcement of nominal types.
 
-All of these are never checked: explicit SIMD is essentially only
-required for speed, and checking inflates one instruction to 5 or
-more.
+All of these operators and conversions are never checked (in the sense
+of the arithmetic overflow checks of `-C debug-assertions`): explicit
+SIMD is essentially only required for speed, and checking inflates one
+instruction to 5 or more.
 
 ## Platform Detection
 
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -132,6 +132,33 @@ in a "duck-typed" manner: it will just ensure that the types are SIMD
 vectors with the appropriate length and element type, it will not
 enforce a specific nominal type.
 
+NB. The structural typing is just for the declaration: if a SIMD intrinsic
+is declared to take a type `X`, it must always be called with `X`,
+even if other types are structurally equal to `X`. Also, within a
+signature, SIMD types that must be structurally equal must be nominal
+equal. I.e. if the `add_...` all refer to the same intrinsic to add a
+SIMD vector of bytes,
+
+```rust
+// (same length)
+struct A(u8, u8, ..., u8);
+struct B(u8, u8, ..., u8);
+
+extern "rust-intrinsic" {
+    fn add_aaa(x: A, y: A) -> A; // ok
+    fn add_bbb(x: B, y: B) -> B; // ok
+    fn add_aab(x: A, y: A) -> B; // error, expected B, found A
+    fn add_bab(x: B, y: A) -> B; // error, expected A, found B
+}
+
+fn double_a(x: A) -> A {
+    add_aaa(x, x)
+}
+fn double_b(x: B) -> B {
+    add_aaa(x, x) // error, expected A, found B
+}
+```
+
 There would additionally be a small set of cross-platform operations
 that are either generally efficiently supported everywhere or are
 extremely useful. These won't necessarily map to a single instruction,
@@ -173,9 +200,9 @@ extern "rust-intrinsic" {
 ```
 
 The raw definitions are only checked for validity at monomorphisation
-time, ensure that `T` is a SIMD vector, `U` is the element type of `T`
-etc. Libraries can use traits to ensure that these will be enforced by
-the type checker too.
+time, ensure that `T` is a SIMD vector, `Elem` is the element type of
+`T` etc. Libraries can use traits to ensure that these will be
+enforced by the type checker too.
 
 This approach has some downsides: `simd_shuffle32` (e.g. `Simd32<u8>`
 on AVX, and `Simd32<u16>` on AVX-512) and especially `simd_shuffle64`
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -1,4 +1,4 @@
-- Feature Name: simd_basics
+- Feature Name: simd_basics, cfg_target_feature
 - Start Date: 2015-06-02
 - RFC PR: (leave this empty)
 - Rust Issue: (leave this empty)
@@ -25,12 +25,12 @@ This RFC is focused on building stable, powerful SIMD functionality in
 external crates, not `std`.
 
 This makes it much easier to support functionality only "occasionally"
-available with Rust's preexisting `cfg` system. If it were in `std`,
-there would need to be some highly delayed `cfg` system so that
-functions that only work with (say) AVX-2 support:
-
-- don't break compilation on systems that don't support it, but
-- are still usable on systems that do support it.
+available with Rust's preexisting `cfg` system. There's no way for
+`std` to conditionally provide an API based on the target features
+used for the final artifact. Building `std` in every configuration is
+certainly untenable. Hence, if it were to be in `std`, there would
+need to be some highly delayed `cfg` system to support that sort of
+conditional API exposure.
 
 With an external crate, we can leverage `cargo`'s existing build
 infrastructure: compiling with some target features will rebuild with
@@ -39,11 +39,11 @@ those features enabled.
 
 # Detailed design
 
-The design comes in three parts:
+The design comes in three parts, all on the path to stabilisation:
 
-- types
-- operations
-- platform detection
+- types (`feature(simd_basics)`)
+- operations (`feature(simd_basics)`)
+- platform detection (`feature(cfg_target_feature)`)
 
 The general idea is to avoid bad performance cliffs, so that an
 intrinsic call in Rust maps to preferably one CPU instruction, or, if
@@ -92,7 +92,9 @@ use the `SimdPrim` trait as a bound).
 
 It is illegal to take an internal reference to the fields of a
 `repr(simd)` type, because the representation of booleans may require
-to change, so that booleans are bit-packed.
+to change, so that booleans are bit-packed. The official external
+library providing SIMD support will have private fields so this will
+not be generally observable.
 
 ### `simd_primitive_trait`
 
@@ -107,6 +109,13 @@ restriction and possibly tweaks type's internal representation (as
 such, it's legal for a single type to implement multiple traits with
 the attribute, if a bit pointless).
 
+This trait exists to allow new-type wrappers around primitives to also
+be usable in a SIMD context. However, this only works in limited
+scenarios (i.e. when the type wraps a single primitive) and so needs
+to be an explicit part of every type's API: type authors opt-in to
+being designed-for-SIMD. If it was implicit, changes to private fields
+may break downstream code.
+
 ## Operations
 
 CPU vendors usually offer "standard" C headers for their CPU specific
@@ -116,10 +125,13 @@ x86(-64)][x86].
 [armneon]: http://infocenter.arm.com/help/topic/com.arm.doc.ihi0073a/IHI0073A_arm_neon_intrinsics_ref.pdf
 [x86]: https://software.intel.com/sites/landingpage/IntrinsicsGuide
 
-All of these would be exposed as (eventually) stable intrinsics with
-names very similar to those that the vendor suggests (only difference
-would be some form of manual namespacing, e.g. prefixing with the CPU
-target), loadable via an `extern` block with an appropriate ABI.
+All of these would be exposed as compiler intrinsics with names very
+similar to those that the vendor suggests (only difference would be
+some form of manual namespacing, e.g. prefixing with the CPU target),
+loadable via an `extern` block with an appropriate ABI. This subset of
+intrinsics would be on the path to stabilisation (that is, one can
+"import" them with `extern` in stable code), and would not be exported
+by `std`.
 
 ```rust
 extern "rust-intrinsic" {
@@ -164,18 +176,23 @@ platform specific intrinsic for shuffling.
 
 ```rust
 extern "rust-intrinsic" {
-    fn simd_shuffle2<T: SimdVector>(v: T, w: T, i0: u32, i1: u32) -> Simd2<T::Elem>;
-    fn simd_shuffle4<T: SimdVector>(v: T, w: T, i0: u32, i1: u32, i2: u32, i3: u32) -> Simd4<T::Elem>;
-    fn simd_shuffle8<T: SimdVector>(v: T, w: T,
-                                    i0: u32, i1: u32, i2: u32, i3: u32,
-                                    i4: u32, i5: u32, i6: u32, i7: u32) -> Simd8<T::Elem>;
-    fn simd_shuffle16<T: SimdVector>(v: T, w: T,
-                                     i0: u32, i1: u32, i2: u32, i3: u32,
-                                     i4: u32, i5: u32, i6: u32, i7: u32
-                                     i8: u32, i9: u32, i10: u32, i11: u32,
-                                     i12: u32, i13: u32, i14: u32, i15: u32) -> Simd16<T::Elem>;
+    fn simd_shuffle2<T, Elem>(v: T, w: T, i0: u32, i1: u32) -> Simd2<Elem>;
+    fn simd_shuffle4<T, Elem>(v: T, w: T, i0: u32, i1: u32, i2: u32, i3: u32) -> Sidm4<Elem>;
+    fn simd_shuffle8<T, Elem>(v: T, w: T,
+                              i0: u32, i1: u32, i2: u32, i3: u32,
+                              i4: u32, i5: u32, i6: u32, i7: u32) -> Simd8<Elem>;
+    fn simd_shuffle16<T, Elem>(v: T, w: T,
+                               i0: u32, i1: u32, i2: u32, i3: u32,
+                               i4: u32, i5: u32, i6: u32, i7: u32
+                               i8: u32, i9: u32, i10: u32, i11: u32,
+                               i12: u32, i13: u32, i14: u32, i15: u32) -> Simd16<Elem>;
 }
 ```
+
+The raw definitions are only checked for validity at monomorphisation
+time, ensure that `T` is a SIMD vector, `U` is the element type of `T`
+etc. Libraries can use traits to ensure that these will be enforced by
+the type checker too.
 
 This approach has some downsides: `simd_shuffle32` (e.g. `Simd32<u8>`
 on AVX, and `Simd32<u16>` on AVX-512) and especially `simd_shuffle64`
@@ -191,7 +208,8 @@ let z = concat(v, w);
 return [z[i0], z[i1], z[i2], ...]
 ```
 
-The indices `iN` have to be compile time constants.
+The indices `iN` have to be compile time constants. Out of bounds
+indices yield unspecified results.
 
 Similarly, intrinsics for inserting/extracting elements into/out of
 vectors are provided, to allow modelling the SIMD vectors as actual
@@ -199,13 +217,14 @@ CPU registers as much as possible:
 
 ```rust
 extern "rust-intrinsic" {
-    fn simd_insert<T: SimdVector>(v: T, i0: u32, elem: T::Elem) -> T;
-    fn simd_extract<T: SimdVector>(v: T, i0: u32) -> T::Elem;
+    fn simd_insert<T, Elem>(v: T, i0: u32, elem: Elem) -> T;
+    fn simd_extract<T, Elem>(v: T, i0: u32) -> Elem;
 }
 ```
 
 The `i0` indices do not have to be constant. These are equivalent to
-`v[i0] = elem` and `v[i0]` respectively.
+`v[i0] = elem` and `v[i0]` respectively. They are type checked
+similarly to the shuffles.
 
 ### Comparisons
 
@@ -226,10 +245,10 @@ extern "rust-intrinsic" {
 }
 ```
 
-However, these will be type checked, to ensure that `T` and `U` are
-the same length, and that `U` is appropriately shaped for a boolean. A
-library actually importing them might use some trait bounds to get
-actual type-safety.
+These are type checked during code-generation similarly to the
+shuffles. Ensuring that `T` and `U` has the same length, and that `U`
+is appropriately "boolean"-y. Libraries can use traits to ensure that
+these will be enforced by the type checker too.
 
 ### Built-in functionality
 
@@ -333,3 +352,4 @@ cfg_if_else! {
 
 - Should integer vectors get `/` and `%` automatically? Most CPUs
   don't support them for vectors.
+- How should out-of-bounds shuffle and insert/extract indices be handled?
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -196,17 +196,17 @@ platform specific intrinsic for shuffling.
 
 ```rust
 extern "platform-intrinsic" {
-    fn simd_shuffle2<T, Elem>(v: T, w: T, idx: [i32; 2]) -> Simd2<Elem>;
-    fn simd_shuffle4<T, Elem>(v: T, w: T, idx: [i32; 4]) -> Sidm4<Elem>;
-    fn simd_shuffle8<T, Elem>(v: T, w: T, idx: [i32; 8]) -> Simd8<Elem>;
-    fn simd_shuffle16<T, Elem>(v: T, w: T, idx: [i32; 16]) -> Simd16<Elem>;
+    fn simd_shuffle2<T, U>(v: T, w: T, idx: [i32; 2]) -> U;
+    fn simd_shuffle4<T, U>(v: T, w: T, idx: [i32; 4]) -> U;
+    fn simd_shuffle8<T, U>(v: T, w: T, idx: [i32; 8]) -> U;
+    fn simd_shuffle16<T, U>(v: T, w: T, idx: [i32; 16]) -> U;
 }
 ```
 
 The raw definitions are only checked for validity at monomorphisation
-time, ensure that `T` is a SIMD vector, `Elem` is the element type of
-`T` etc. Libraries can use traits to ensure that these will be
-enforced by the type checker too.
+time, ensure that `T` and `U` are SIMD vector with the same element
+type, `U` has the appropriate length etc. Libraries can use traits to
+ensure that these will be enforced by the type checker too.
 
 This approach has similar type "safety"/code-generation errors to the
 vectors themselves.
@@ -362,6 +362,17 @@ cfg_if_else! {
   pointers(/references?) in `repr(simd)` types.
 - allow (and ignore for everything but type checking) zero-sized types
   in `repr(simd)` structs, to allow tagging them with markers
+- the shuffle intrinsics could be made more relaxed in their type
+  checking (i.e. not require that they return their second type
+  parameter), to allow more type safety when combined with generic
+  simd types:
+
+      #[repr(simd)] struct Simd2<T>(T, T);
+      extern "platform-intrinsic" {
+          fn simd_shuffle2<T, U>(x: T, y: T, idx: [u32; 2]) -> Simd2<U>;
+      }
+
+  This should be a backwards-compatible generalisation.
 
 # Alternatives
 
@@ -403,7 +414,6 @@ cfg_if_else! {
   complicated set of traits (with compiler integration).
 - use generic intrinsics like shuffles for the arithmetic operations,
   instead of providing the operations implicitly.
-
 
 # Unresolved questions
 
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -126,11 +126,11 @@ extern "platform-intrinsic" {
 These all use entirely concrete types, and this is the core interface
 to these intrinsics: essentially it is just allowing code to exactly
 specify a CPU instruction to use. These intrinsics only actually work
-on a subset of the CPUs that Rust targets, and are only be available
-for `extern`ing on those targets. The signatures are typechecked, but
-in a "duck-typed" manner: it will just ensure that the types are SIMD
-vectors with the appropriate length and element type, it will not
-enforce a specific nominal type.
+on a subset of the CPUs that Rust targets, and will result in compile
+time errors if they are called on platforms that do not support
+them. The signatures are typechecked, but in a "duck-typed" manner: it
+will just ensure that the types are SIMD vectors with the appropriate
+length and element type, it will not enforce a specific nominal type.
 
 NB. The structural typing is just for the declaration: if a SIMD intrinsic
 is declared to take a type `X`, it must always be called with `X`,
--- a/text/0000-simd-infrastructure.md
+++ b/text/0000-simd-infrastructure.md
@@ -390,6 +390,17 @@ cfg_if_else! {
   compiler can know this. The `repr(simd)` approach means there may be
   more than one SIMD-vector type with the `Simd8<u32>` shape (or, in
   fact, there may be zero).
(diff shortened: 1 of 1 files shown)


eternaleye:
X-posting my late comment from the pre-RFC:

What I was saying was a bit more than that.

1. From my earlier post, these may not need to be intrinsics _at all_ - one-instruction-of-`asm!()` functions  with `#[inline(always)]` and proper register specifiers on the `asm!()` can do the job _except_ for the quirky magic structural typing.

2. The ergonomics of strict types, whether `[u32; 4]` or `Simd4<u32>`, really aren't that bad for low-level building blocks that will mostly live behind prettier interfaces.

3. What problems there are with the ergonomics can be largely resolved with `T: Structural<Layout=[u32; 4]>> + SimdSafe`, where SimdSafe is a marker trait denoting the same things as `#[repr(simd)]`, and possibly added by it.

And yeah, I edited "alignment voodoo" into my post before you mentioned that :stuck_out_tongue: 

Anyway, the result of the above is that one only really _needs_ two changes to the compiler:

1. `#[repr(simd)]`
2. `#[lang_item="simd_repr_marker"]` (added by `#[repr(simd)]`)

`Structural` _can_ be done without any help from the compiler, but would benefit a lot from a `#[derive]`

But as far as _benefits_, this avoids a large mass of worryingly magical (regarding parameter types) intrinsics being added to the compiler.

alexcrichton:
> these may not need to be intrinsics at all

Unfortunately if we want to get SIMD on stable Rust this would necessitate a stable `asm!` macro, and we've got a much stronger story for stabilizing these intrinsics than we do that macro

> What problems there are with the ergonomics can be largely resolved with T: Structural<Layout=[u32; 4]>> + SimdSafe, where SimdSafe is a marker trait denoting the same things as #[repr(simd)], and possibly added by it.

I believe a core aspect of this RFC is that it's stabilizing the absolute bare minimum of what the compiler needs to support SIMD. If we add in a few language traits and other various types here and there it's more surface area that will have to be stabilized. If this stuff can be built externally in a library, that'd be great! This RFC, however, is just focused on the compiler support.

eternaleye:
> This RFC, however, is just focused on the compiler support.

Sure! I just think that the way it makes the intrinsics unify types with sufficiently similar layouts under `#[repr(simd)]` is neither necessary nor advisable, and wanted to lay out a library-based alternative.

huonw:
> some subset of intrinsics magically ignoring that feels very questionable to me.

Hm, maybe I haven't been clear. They're not really ignoring it.

Suppose we have:

```rust
#[repr(simd)]
struct A(f64, f64);
#[repr(simd)]
struct B(f64, f64);

extern {
      fn some_simd_intrinsic(x: A);
}
```

It's not legal to call `some_simd_intrinsic(B(0.0, 1.0))` (type error). The "structural typing" of intrinsics just means it's valid to also write the above and (if desired) `extern { fn some_simd_intrinsic(x: B); }` elsewhere. I see it some what similar to how importing C functions work: if some C function takes a struct `Foo`, we don't require that there's some single canonical type (or structural generic) always passed to every import of that C function, just that the function passed has the right layout.

(That said, it's slightly different, since the compiler doesn't enforce that arguments have the right layout for C functions, whereas the SIMD intrinsics do have layout-enforcement.)

eternaleye:
Ah, that does make rather more sense, and is not the impression I had gotten.

main--:
>There would additionally be a small set of cross-platform operations that are either generally efficiently supported everywhere or are extremely useful. These won't necessarily map to a single instruction, but will be shimmed as efficiently as possible.
>* shuffles and extracting/inserting elements
>* comparisons
>
>Lastly, arithmetic and conversions are supported via built-in operators.

The Motivation section mentions how this RFC aims to provide just some ground-work on top of which nice SIMD functionality could be built. While builtin arithmetic, shuffles etc for `repr(simd)` types is nice and convenient, providing it at *this* level seems questionable. I think something like this could be accomplished inside the to-be-written SIMD library as well with some operator overloading and the intrinsic functions for basic arithmetic.

>The indices iN have to be compile time constants.

I have a bad feeling about this. A regular method call shouldn't require parameters to be compile time constants. Using generics to express this requirement as shown [here](https://github.com/rust-lang/rfcs/pull/1062#issuecomment-119752339) depends on #1062, but it would be a much cleaner solution.

>Out of bounds indices yield unspecified results.

Wouldn't a compile error be nicer here?

---

That aside, when reading this RFC I had the same thought as @eternaleye: Why implement compiler magic when these functions could be implemented in plain rust with `asm!()`? The absence of stable inline asm in a systems programming language is annoying and while this RFC attempts to make up for that by simply importing **all** C/C++ intrinsics into Rust, this set grows regularly, requiring a compiler update every time when an update of the [system LLVM](https://internals.rust-lang.org/t/building-with-system-llvm/1647) would be sufficient. (Well, it's mostly the maintenance overhead for rust. Why maintain a set of intrinsics when someone else is already doing it?)

Additionally, inline asm allows the programmer to influence things like instruction scheduling and register allocation (within the asm section), in case the compiler is doing a bad job in that regard.

So I'd suggest solving the "Operations" section of the RFC in a way that doesn't require any compiler changes (at least not specifically for SIMD).

---

I'm unsure about what the `repr(simd)` introduced here really *does*. I guess its primary purpose is signaling to the compiler that this struct can live in the SIMD registers and be subject to SIMD operations (like this builtin arithmetic).

People often argue that vectorization is best left to the compiler and as Rust uses LLVM, many simple cases can benefit from great optimizations. But unfortunately optimizers aren't perfect and simply unable to handle sufficiently complex code, so obviously an explicit way for programmers to express how they want code to be vectorized is necessary. Somewhere above, I suggested removing even those basic operations from `repr(simd)` types and doing them in intrinsics (or preferably inline asm) instead. (Danger ahead: I fear these ideas might require significant LLVM changes and therefore be infeasible) Then we could think further and consider removing `repr(simd)`: The inline asm constraints (or the intrinsics) should ensure that these values stay in the SIMD registers while we're working with them. While we aren't, it's up to the compiler. As a result, the compiler could decide to hold even non-SIMD values inside the SIMD registers which is probably not a big issue as I remember the Intel optimization manual mentioning how spilling to XMM registers can be faster than spilling to memory.

---

So to summarize, I'm in favor of simplifying/breaking up the "types" and the "operations" sections of the RFC by building on top of much less specific compiler features and pushing as much work as possible into the to-be-written SIMD crate (and moving the "platform detection" section into its own RFC).

This feels appropriate as the RFC's intention was just basic groundwork. Or in other words: I'm arguing that "the absolute bare minimum of what the compiler needs to support SIMD" is zero. The missing parts aren't necessarily SIMD-specific.

eternaleye:
@main--
> I'm unsure about what the `repr(simd)` introduced here really _does_. I guess its primary purpose is signaling to the compiler that this struct can live in the SIMD registers and be subject to SIMD operations (like this builtin arithmetic).

1. Can live in SIMD registers (though, not sure how it'll handle someone trying to apply it to `[u64; 37]` or other silliness)
2. Interior references are forbidden
3. Tweaks the in-memory layout to match SIMD for platform
4. Subject to SIMD operations
5. **subject to SIMD alignment constraints**

@huonw, did I miss anything? ISTR you mentioning it didn't validate things like that they need to be homogenous, and that such things would be left up to `impl`'ing some `unsafe trait`

pcwalton:
Arithmetic identity-based optimizations apply just as well to SIMD operations as they do to non-SIMD operations. By using the intrinsics as opposed to inline asm we give LLVM the ability to do those optimizations.

main--:
@pcwalton Oh, I didn't know that! Yes, that's a big advantage of the intrinsics then.

RalfJung:
Concerning the structural typing when importing the intrinsics: Please be careful that this does not end up allowing people to "peek through" private abstractions of data-types. That would be a horrible mess of a safety issue.

Essentially, such merely structural typing should only be allowed if the module had access to all the fields of the type anyway: Either because they were all public (all the way down), or because the type was defined in the same module.
What's the reason for choosing this unconventional approach to typing, rather than using tuples, or arrays, or lang-items for the Simd* types?

Aatch:
Just a note to all the people that keep mentioning inline assembly: `asm!` is pretty much a black box for LLVM. You can tell it a decent amount about the contents, but at the end of the day, it's going to have to take a conservative approach to it. LLVM can reason a lot more about regular operations on vectors (arithmetic, equality, shuffling) and intrinsic functions than it can single-instruction inline asm segments.

huonw:
> The Motivation section mentions how this RFC aims to provide just some ground-work on top of which nice SIMD functionality could be built. While builtin arithmetic, shuffles etc for repr(simd) types is nice and convenient, providing it at this level seems questionable. I think something like this could be accomplished inside the to-be-written SIMD library as well with some operator overloading and the intrinsic functions for basic arithmetic.

I agree that it isn't totally necessary to actually use the arithmetic operators: we could instead use a generic intrinsic similar to the comparison operators. However, I think it is important we do more than the platform intrinsics: LLVM (and compilers in general) knows more about its internal `add` instruction than arbitrary platform specific intrinsics, and so may be able to optimise it more aggressively.

For shuffles, the optimisation applies: the compiler can e.g. simplify a sequence of shuffles into a single one. Also. the RFC discusses this. One point in it is the compiler synthesizing an optimal (/close to optimal) sequence of instructions for an arbitrary shuffle, instead of forcing the programmer to think about doing that themselves.

> I have a bad feeling about this. A regular method call shouldn't require parameters to be compile time constants. Using generics to express this requirement as shown here depends on #1062, but it would be a much cleaner solution.

This isn't a regular method call: intrinsics are special in many ways. Note that my solution on #1062 that you link to just calls the intrinsic. This is the low-level API, people generally won't be calling the intrinsics directly.

> Wouldn't a compile error be nicer here?

Yes, sort of. However, using the trick mentioned in #1062 would result in very poor error messages, since the shuffle order may be passed through multiple layers of generic function calls possibly in external crates, meaning the out-of-bounds error is generated deep inside code that the programmer didn't write.

> Why implement compiler magic when these functions could be implemented in plain rust with asm!()?

As other have said, `asm!` is a black-box, and seriously inhibits optimisations.

> Additionally, inline asm allows the programmer to influence things like instruction scheduling and register allocation (within the asm section), in case the compiler is doing a bad job in that regard.

Neither of these apply to this: the API is essentially exposing individual CPU instructions, i.e. each `asm!` block is a single instruction. Hence, there's no scheduling benefit, and none of the `asm!` blocks would use concrete registers: they'd all be "generic", to let the compiler allocate registers as it sees fit.

These reasons apply if one was, say, writing an entire inner loop as a single `asm!` block, but it doesn't apply here.

> I'm unsure about what the repr(simd) introduced here really does. I guess its primary purpose is signaling to the compiler that this struct can live in the SIMD registers and be subject to SIMD operations (like this builtin arithmetic).

Yes. `repr(simd)` changes how a type is represented. E.g. it changes the alignment, imposes element constraints, and even changes its ABI (for function/FFI calls).

> Concerning the structural typing when importing the intrinsics: Please be careful that this does not end up allowing people to "peek through" private abstractions of data-types. That would be a horrible mess of a safety issue.

It sort-of does, but in a very very restricted way, that's already possible with `transmute`.

>  What's the reason for choosing this unconventional approach to typing, rather than using tuples, or arrays, or lang-items for the Simd* types?

Tuples and arrays don't have the right low-level details.

`repr(simd)` is essentially acting as a lang-item that can be defined multiple times. All of the actual lang items (i.e. `#[lang = "..."]`) in the compiler can only be defined once in the entire hierarchy of dependencies of a compilation target, which means we'd either have to allow multiple versions of these lang items, or just disallow linking multiple SIMD crates into a project (e.g. two different crates that define low-level SIMD interfaces, or even just versions `0.1` & `0.3` or `1.0` & `2.3` or ... of a single SIMD crate).

RalfJung:
> It sort-of does, but in a very very restricted way, that's already possible with transmute.

Transmute requires unsafe. It shouldn't be possible for safe code to violate abstraction boundaries. What you are proposing (if I follow your RFC correctly) is essentially that `#[repr(simd)]` implies that all fields are public, but the programmer doesn't have to write `pub`.

The restriction that the type used for the intrinsic has to be defined in the same module, shouldn't be a problem for the implementations you envision (with some crate(s) taking care of providing a descent abstraction) , right?

huonw:
Hm, I misunderstood what you were talking about. I'm unsure what the problematic situation you're envisioning could be. Is it something like:

Crate `foo` defines a simd `Simd`, crate `bar` depends on `foo` and loads the `Simd` type. `bar` `extern`'s in an intrinsic like `extern { fn some_simd_intrinsic(x: Simd); }`... and then something bad happens? NB. the only way to call intrinsics is with `unsafe`.

(I'm not against the privacy restriction, I'm just trying to understand the motivation more concretely.)

glaebhoerl:
> `repr(simd)` is essentially acting as a lang-item that can be defined multiple times.

Is a lang item the best intuition here? Wouldn't a closer analogy be `repr(C)`? In both cases the semantic content of the type is (mostly) unaffected and you're just specifying its underlying representation, which is mostly relevant at the ABI, rather than the API level.

> The `repr(simd)` may not enforce that any trait bounds exists/does the right thing at the type checking level for generic `repr(simd)` types. As such, it will be possible to get the code-generator to error out (ala the old `transmute` size errors), 

Would it be possible to just make this best-effort, and fall back to laying the type out normally if it doesn't meet the SIMD requirements (perhaps with a warning)? That seems cleaner than implicit requirements at the code generator level, which feels like a contract violation (if it passes the typechecker, it should compile), and perhaps more in the spirit of `#repr`.

> It is illegal to take an internal reference to the fields of a `repr(simd)` type, because the representation of booleans may require modification, so that booleans are bit-packed.

I wonder if the fact that the borrow checker enforces limits on observability wouldn't actually let us support the interior reference-taking, just in a less efficient way, by first copying the field out onto the stack when taking a shared `&` reference, and in the case of `&mut`, also copying it *back* when it goes out of scope. (Off the top of my head, the potential complication that comes to mind here is generic code - copying the value back is effectively a `Drop` impl for the `&mut`. But it does seem like it could actually be implemented in precisely that way... at least, I can't immediately think of why not.)

In this way it would be even more truly the case that `#repr` has no effect on the semantics, only on the representation and performance characteristics. (I had a similar idea, earlier, [here](https://github.com/rust-lang/rfcs/issues/311).)

> Any type marked `repr(simd)` automatically has the `+`, `-` and `*` operators work. The `/` operator works for floating point, and the `<<` and `>>` ones work for integers.

I might be more comfortable with this if you had to explicitly write `derive(Add)`, and so on, to get the desired operations, even if that in turn just ended up calling out to appropriate compiler magic. On the one hand, it does seem logical that "why would you `repr(simd)` if not to get the SIMD operations", but on the more important-seeming hand, I think there should be a separation of concerns, and `#repr` should really *only* affect the representation (as far as possible).

RalfJung:
> NB. the only way to call intrinsics is with unsafe.

This essentially means its much less of an issue than I thought.

> Hm, I misunderstood what you were talking about. I'm unsure what the problematic situation you're envisioning could be.

Which part is unclear - whether these rules allow code to get around the restriction that usually apply to private fields, or whether getting access to private fields is an issue? The answer to the latter is that it violates parametricity - for now I'll just assume that as accepted. Please tell me if I should elaborate on that. Unsafe code (through transmute) can violate parametricity anyways, but I would still prefer if no additional violations would be introduced.

Regarding the first part, let me try to come up with some examples. I assume "A::Simd" is some other crate's `Simd` type, with all fields private, and `B::Simd` is our own `Simd`.
If we have `fn x86_mm_add_epi16(a: Simd8<i16>, b: Simd8<i16>) -> Simd8<i16>` and the types all just match structurally, I could convert any `B::Simd` to an `A::Simd` by adding a 0, and choosing the argument types to be `B::Simd` and the return type to be `A::Simd`. Similarly, I can convert `A::Simd` to `B::Simd`. This gives me full access to all the private fields.
If we have `fn simd_shuffle2<T, Elem>(v: T, w: T, i0: u32, i1: u32) -> Simd2<Elem>`, I can choose `T` to be `A::Simd` and the return type to be `B::Simd` and convert from `A::Simd` to `B::Simd` with the appropriate options for the shuffles - and back, with a similar trick.

huonw:
Ah, I think you might be working with the same misunderstanding as @eternaleye, that the intrinsics can be *called* with any type that matches structurally, it's just that they can be *declared* with any type that matches structurally. See https://github.com/rust-lang/rfcs/pull/1199#issuecomment-120537682 .

(I 100% agree that being able to write a transmute via SIMD intrinsics would be unfortunate.)

RalfJung:
I had that misunderstanding at first, but the post above was written without that assumption.

The first one, regarding add, can't I declare this with `fn x86_mm_add_epi16(a: A::Simd, b: A::Simd) -> B::Simd`?
And the second one, with shuffle, is actually explicitly *declared* generically, so it can be used with any `T`, right? Otherwise, why the distinction between explicit generics for shuffle, and implicit for `abs`?

huonw:
> Is a lang item the best intuition here? Wouldn't a closer analogy be repr(C)? In both cases the semantic content of the type is (mostly) unaffected and you're just specifying its underlying representation, which is mostly relevant at the ABI, rather than the API level.

I agree that `repr(C)` is probably closer, however I was responding to a comment talking about lang-items. :)

> Would it be possible to just make this best-effort, and fall back to laying the type out normally if it doesn't meet the SIMD requirements (perhaps with a warning)? That seems cleaner than implicit requirements at the code generator level, which feels like a contract violation (if it passes the typechecker, it should compile), and perhaps more in the spirit of #repr.

I think we can relax this in future if we find the RFC doesn't work well in practice. (It's part of the reason I proposed a hard error.)

> I wonder if the fact that the borrow checker enforces limits on observability wouldn't actually let us support the interior reference-taking, just in a less efficient way, by first copying the field out onto the stack when taking a shared & reference, and in the case of &mut, also copying it back when it goes out of scope. (Off the top of my head, the potential complication that comes to mind here is generic code - copying the value back is effectively a Drop impl for the &mut. But it does seem like it could actually be implemented in precisely that way... at least, I can't immediately think of why not.)

Interesting idea, however it seems relatively complicated, and not worth it for SIMD: efficient SIMD code won't be handling/mutating individual elements like this very much.

> I might be more comfortable with this if you had to explicitly write derive(Add), and so on, to get the desired operations, even if that in turn just ended up calling out to appropriate compiler magic. On the one hand, it does seem logical that "why would you repr(simd) if not to get the SIMD operations", but on the more important-seeming hand, I think there should be a separation of concerns, and #repr should really only affect the representation (as far as possible).

Another alternative is just providing arithmetic intrinsics.

> The first one, regarding add, can't I declare this with fn x86_mm_add_epi16(a: A::Simd, b: A::Simd) -> B::Simd?

Oh, I see. It seems sensible to disallow it. I.e. have nominal equality constraints *within* a definition.

And for the `shuffle`, there's an implicit checked-at-code-gen link between `T` and `Elem` (i.e. `Elem` is the actual element type of `T`).

Shuffle is generic because it can be used with literally any SIMD type, i.e. it's not restricted to some subset of types with the same structure.

huonw:
(I just pushed a few more commits to (hopefully!) improve the RFC based on the discussion so far.)

RalfJung:
> Oh, I see. It seems sensible to disallow it. I.e. have nominal equality constraints within a definition.

That would fix the concrete issue I mentioned, but it would still be used to violate the invariants of other types. Like (warning: silly example) some `Simd8<u16>` that ensures that all components are either 0 or 1. Imagine that's somehow crucial for safety. If now a 3rd-party module defines the `add` function that that type, suddenly the invariant can be violated and things can crash.

> And for the shuffle, there's an implicit checked-at-code-gen link between T and Elem (i.e. Elem is the actual element type of T).

Oh, wow. That's very surprising. And I agree with the comment made above that well-typed code should always compile. What's the reason this one is a generic function with code-gen check, while the others can only be imported as non-generic functions? Also, even with the check, how does that help to prevent converting between different nominal types from different modules, that happen to share their element type?

huonw:
> That would fix the concrete issue I mentioned, but it would still be used to violate the invariants of other types. Like (warning: silly example) some Simd8<u16> that ensures that all components are either 0 or 1. Imagine that's somehow crucial for safety. If now a 3rd-party module defines the add function that that type, suddenly the invariant can be violated and things can crash.

I think this is partly a problem with the choice of types. It is solved by storing a wrapper around `u16` rather than a `u16` itself (this is explicitly supported: it is the approach most appropriate for booleans when they are defined as either all 0s or all 1s (bitwise)). However I suppose we could use a privacy-based rule as you suggested earlier.

> Oh, wow. That's very surprising. And I agree with the comment made above that well-typed code should always compile. What's the reason this one is a generic function with code-gen check, while the others can only be imported as non-generic functions?

The shuffles/inserts/etc. don't care about either the vector type or the element type at all; it just cares that the input is a vector; having to import a version for every single type that could be shuffled would get... tiresome. As stated in the RFC the well-typing is trivial to achieve with a trait bound on the intrinsic import, e.g. `T: Simd<Elem = Elem>`, but I do not think we should start hard-coding relatively complicated traits like that into the compiler. At least, not right now.

> Also, even with the check, how does that help to prevent converting between different nominal types from different modules, that happen to share their element type?

You can't at the moment, it's part of what `repr(simd)` means and why it's strictly opt-in. Specialised semantics shouldn't be added by creating new vector types, they should only be added via special element types or wrappers around vector types. Also handled via your privacy rule.

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+These are type checked during code-generation similarly to the
+shuffles. Ensuring that `T` and `U` has the same length, and that `U`
+is appropriately "boolean"-y. Libraries can use traits to ensure that
+these will be enforced by the type checker too.
+
+### Built-in functionality
+
+Any type marked `repr(simd)` automatically has the `+`, `-` and `*`
+operators work. The `/` operator works for floating point, and the
+`<<` and `>>` ones work for integers.
```
eefriedman: Does this mean Simd4<u8> implements unary negation?  If so, is that intentional?

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+
+SIMD vectors can be converted with `as`. As with intrinsics, this is
+"duck-typed" it is possible to cast a vector type `V` to a type `W` if
+their lengths match and their elements are castable (i.e. are
+primitives), there's no enforcement of nominal types.
```
eefriedman: What happens if an f32->u32 conversion is out of range?  (This hasn't been resolved for scalars either, but we might pick a different solution for SIMD vectors?)

text/0000-simd-infrastructure.md
huonw: I'm not clear on which part of `Simd4<u8>` you're thinking is special? The fact it's `u8`, or the fact it's only 32-bits long?

Either way, no, I only had binary operators in mind when I wrote this. In any case, I'm changing the whole section to just use intrinsics and let the `std::ops` traits handle operator overloads.

eefriedman: The fact that it's u8, and unary negation on unsigned integers will be illegal.

If you're changing this to use intrinsics, we can probably just name the intrinsics so it's clear they're wrapping.

text/0000-simd-infrastructure.md
huonw: I'm basically punting that to depend on how we solve it for scalars. Also, as an (unsafe) intrinsic, we can punt it entirely to a library concern: the intrinsic just has UB when floats are too large.

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+representation as `Simd2<u64>`.
+
+The `repr(simd)` may not enforce that any trait bounds exists/does the
+right thing at the type checking level for generic `repr(simd)`
+types. As such, it will be possible to get the code-generator to error
+out (ala the old `transmute` size errors), however, this shouldn't
+cause problems in practice: libraries wrapping this functionality
+would layer type-safety on top (i.e. generic `repr(simd)` types would
+use some `unsafe` trait as a bound that is designed to only be
+implemented by types that will work).
```
eefriedman: This isn't great... ideally, we want as few errors as possible coming out of monomorphization.  (Currently, essentially the only errors that are fundamentally tied to monomorphization are recursion depth limit errors, and it should stay that way if possible.)  If `Simd4<T>` isn't exposed outside of the library, it's just an implementation convenience; if it is exposed, it's encouraging people to write code that gives weird errors.  Using `unsafe` doesn't really improve anything (that's memory safety, not crash-the-compiler safety), and Rust doesn't have sealed traits at the moment.

I think Rust's trait system is powerful enough that you could come up with a reasonable SIMD library without going down that path.  We already have traits for most of the common operations (arithmetic, conversions, etc.), and you can write additional traits for SIMD-specific operations like shuffles and splitting. This allows the library to completely hide the fact that `repr(simd)` exists; actually, a user could even implement SIMD traits for types that aren't `repr(simd)`.

huonw: > If Simd4<T> isn't exposed outside of the library, it's just an implementation convenience; if it is exposed, it's encouraging people to write code that gives weird errors

I also really really dislike monomorphisation errors, however, I don't think this is case too problematic. In practice, the generics will be bounded by traits, and users opt-in to getting a weird error by implementing the trait. I imagine that almost all instances of monomorphisation errors will be caught immediately: an implementation of the trait will presumably accompany uses of the type in SIMD vectors, resulting in the errors.

>  Using unsafe doesn't really improve anything (that's memory safety, not crash-the-compiler safety), and Rust doesn't have sealed traits at the moment.

One can seal traits via privacy. It's not sealing in a way that the compiler will reason about for type-system purposes, but it is enough for this. However, I'm not sure it makes sense.

> I think Rust's trait system is powerful enough that you could come up with a reasonable SIMD library without going down that path

This is very unclear to me. I suspect there will be a large amount of duplicated code, exponentially much, in fact: every operation that works on, say, 4-element vectors needs to be duplicated across every concrete type with that shape, e.g. `Simd4<i32>`, `Simd4<u32>`, `Simd4<f32>`, `Simd4<bool32>` definitely exist (and possibly more).

I'm very happy to be proved wrong, and this RFC is expressly designed to allow fairly free experimentation for higher level APIs. Also, it will remain unstable for a while, while we tweak (or, massively-change, if necessary) it.

One approach that I've discussed a bit is to "eventually" require some sort of trait bound on any generic `repr(simd)`'s, however, we don't know what this trait/traits should look like yet, and getting freeform experience will help guide that decision.

> actually, a user could even implement SIMD traits for types that aren't repr(simd).

This seems like something that's perfectly possible even with generic `repr(simd)`. (I'm missing the point.)

eefriedman: > > actually, a user could even implement SIMD traits for types that aren't repr(simd).
> 
> This seems like something that's perfectly possible even with generic repr(simd). (I'm missing the point.)

Err, I meant that if Simd4<T> is a struct, and you encourage people to write code generic over Simd4<T>, that limits the usage of the code, as opposed to a Simd4<T> trait which could be implemented by anyone.  I suppose the same argument could be made for other generic types, but it seems more important here because of the limitations on T in the struct Simd4<T>.

> This is very unclear to me. I suspect there will be a large amount of duplicated code, exponentially much, in fact: every operation that works on, say, 4-element vectors needs to be duplicated across every concrete type with that shape, e.g. Simd4<i32>, Simd4<u32>, Simd4<f32>, Simd4<bool32> definitely exist (and possibly more).

Any SIMD library is going to have macros all over the place in the implementation anyway, so I'm not sure this actually matters.  The set of allowed types is "small" (around 10 or so).  We can always experiment, I guess.

RalfJung:
>I think this is partly a problem with the choice of types. It is easily solved by storing a wrapper around u16 rather than a u16 itself (this is explicitly supported: it is the approach most appropriate for booleans when they are defined as either all 0s or all 1s (bitwise)).

I don't understand. How would a wrapper type help, since intrinsics can be imported based on structural matching, ignoring any wrappers?

arielb1:
I don't think the "structual typing" effect is too abstraction-breaking - you can only reach the abstraction-breakage if you declare an intrinsic with a "specially-crafted" signature, and declaring extern functions with "specially-crafted" signatures is already equivalent to a transmute (you can e.g. declare a pointer-sized identity function as `for<'a> foo(&'a Foo) -> &'a Bar`).

The "monomorphization-error" part is a bit more annoying. I prefer to think of it as undefined behaviour detected as compile-time (monomorphized unsafe code can of-course exhibit guaranteed UB even without intrinsics). Theoretically, we could just emit an `intrinsics::unreachable()` - this can even be part of a useful program, if the UB isn't actually reachable - but we prefer to abort compilation.

cesarb:
If I read this RFC correctly, it doesn't allow compiling a single executable or library which can use the best SIMD functionality available at runtime. For instance, an executable compiled for AVX2 wouldn't run on an older CPU, and an executable compiled for SSE2 wouldn't be able to make use of the AVX2 extensions.

For C, there's a [GCC function attribute](https://gcc.gnu.org/onlinedocs/gcc-5.2.0/gcc/Function-Attributes.html) (`__attribute__((target("...")))`) which temporarily changes the SIMD instructions allowed within a single function. That allows you to have for instance a function compiled for AVX2 while everything else is compiled for SSE2 only.

And GCC has another cool extension ([Function Multiversioning](https://gcc.gnu.org/onlinedocs/gcc-5.2.0/gcc/Function-Multiversioning.html)) which builds on the `target` attribute to automatically select at runtime which version of the function to use.

For Rust, it would be like:

```rust
#[target_feature = "avx2"]
fn foo(...) -> ... {
    // Implementation used when AVX2 is available.
    // AVX2 intrinsics can be used here, the compiler can generate AVX2 instructions here.
}

#[target_feature = "sse2"]
fn foo(...) -> ... {
    // Implementation used when AVX2 is not available but SSE2 is available.
    // AVX2 intrinsics cannot be used here, the compiler cannot generate AVX2 instructions here.
}

#[target_feature = "default"]
fn foo(...) -> ... {
    // Implementation used when neither is available.
    // What's available here depends on the compiler command line (same as for functions without #[target_feature]).
}
```

The same function is defined three times, but only one will be called at runtime, depending on the available CPU features. The order doesn't matter, it chooses the best one available.

eternaleye:
@cesarb
Function multiversioning is VERY platform-dependent - it requires support from the dynamic linker, because it's built on top of ifuncs. I don't think LLVM can even generate those at present, and they only work on Linux+glibc last I checked.

cesarb:
@eternaleye 

> Function multiversioning is VERY platform-dependent - it requires support from the dynamic linker, because it's built on top of ifuncs.

It is that way on C. It doesn't have to be that way on Rust. For instance, it could use ifuncs where available, and do something else where not available.

eternaleye:
@cesarb
Perhaps, but that "something else" would also need specified. Something that works along the lines of `lazy_static!` under the hood could probably function almost identically to ifuncs for example. However, that could be a whole RFC in its own right, and then function multiversioning a _second_ RFC over the top of that.

Aatch:
@cesarb to clarify this RFC does not *specify* such a dispatch mechanism, but to my understanding nothing in it makes that mechanism not possible. If you have specific concerns about the current design that you think might prevent an auto-dispatch mechanism from being implemented, feel free to voice that concern. Otherwise, as @eternaleye says, this RFC is for getting base-line support for SIMD into the language, additional functionality can be worked out with further RFCs.

cesarb:
@eternaleye , @Aatch Perhaps I shouldn't have mentioned auto-dispatch, it ended up obfuscating what I wanted to mention, which is the *other* effect of GCC's `__attribute__((target("...")))`: it changes the target CPU and/or features while compiling the function.

The lack of something like that makes creating your own auto-dispatch mechanism much harder, since you have to compile each variant as a separate library (with different compiler options) and link them together.

But I agree that it's a separate feature which can be discussed later in its own RFC.

Now, as for the rest of this RFC. I played a bit with SIMD in my [blake2-rfc](https://github.com/cesarb/blake2-rfc) crate, to get the feel of it before giving my opinion. Sorry for the very long comment, but here it goes.

First of all, this RFC already does more than one thing: it has *four* features which can evolve separately. I believe that each one should have its own separate feature flag, so they can be stabilized separately if necessary.

# Types

The `repr(simd)` attribute is mostly a rename of the current `simd` attribute. It does *two* things: declare a type to have a "SIMD vector" representation, and implicitly add a few primitive operations to it.

Since the point of this RFC is to allow the creation of SIMD helper crates, and not to be used directly by the end developer, wouldn't it be more flexible to not implement the primitive operations, and instead let the crate developer implement the `Add`, etc. traits by calling the corresponding intrinsic?

That would allow for instance a SIMD crate which implemented a different `Mul` operation, or which made `Add` be saturating. Wrapping the SIMD type in a single-element struct just to be able to implement traits on it doesn't work as expected; I tried, and it confused the optimizer so much that the speed halved (it was moving all the time between SIMD and normal registers, even though everything was force-inlined).

I don't see the point of flattening the types, instead of allowing only structs with the same element type repeated a few times. That part needs additional justification for why it won't be just adding unnecessary complexity to the implementation.

The SIMD types by themselves are already very useful; they have the correct alignment, so they can be passed to SIMD code written in C even if the rest of this RFC isn't stable yet.

# CPU instruction intrinsics

These are intrinsics which map directly to a CPU instruction. We already have a bit of it, via the `link_llvm_intrinsics` feature, but the proposal in this RFC is cleaner and more complete (as far as I could find, some intrinsics seem to not be available through the `link_llvm_intrinsics` feature; for instance, I couldn't find the u64x2 shuffle intrinsics).

I'd like to see even basic things like adding two vectors exposed as intrinsics (`x86_mm_add_epi32` for instance). There are many types of addition, and just `a + b` doesn't make it obvious which one it is (wrapping? saturating?).

And a real-life example of "why not inline asm": LLVM was able to convert a left-shift of `1` (using what you'd call the `x86_mm_slli_epi64` intrinsic) into an add of the vector with itself. That would not happen with inline asm.

Just leaving this link I found here, as it can be useful: https://software.intel.com/sites/landingpage/IntrinsicsGuide/

# Generic intrinsics

These are the intrinsics which are the same everywhere, and might map to more than one CPU instruction. It would be better for them to be a separate feature from the CPU instruction intrinsics, since while the design of the CPU instruction intrinsics is uncontroversial (just copy the C intrinsics), this part can lead to more discussion, so they might need more time to become stable.

The generic shuffles are useful, but I see only two-vector shuffles. A single-vector shuffle is also useful, and `simd_shuffle_single_4(v, 1, 2, 3, 0)` is easier to understand than `x86_mm_shuffle_epi32(v, 0b00_11_10_01)`. And sometimes a single intrinsic isn't available; I couldn't find so far a single intrinsic to swap the two halves of an u64x2. (Of course, another option would be to do a `simd_shuffle4(v, v, 1, 2, 3, 0)`, and hope that the compiler can notice that only the first vector is being used.)

An alternative design for the generic shuffles would be: `simd_shuffle(v, w, s)` where `s` is some kind of `Shuffle` object, similar to `Range` (but restricted to be a compile-time constant). That might allow separating the shuffle specification from the call; for instance, one could have a (compile-type constant) `fn swap_consecutive64() -> Shuffle` which returned a `Shuffle` which swaps consecutive objects (`1, 0, 3, 2, ...`).

If the vectors are restricted to just "structs with a repeated element type", insert and extract of a constant index don't need an intrinsic; they're just `v.0 = ...` and `... = v.0` (or `v.x = ...` and `... = v.x` if they're named). Of course, that doesn't work if the index is not a constant, unless the vector was declared as an array (`#[repr(simd)] struct u32x4([u32; 4])` as mentioned in the alternatives).

The result of the comparison intrinsics needs to be better explained; do they return vectors of `bool`, or vectors of arbitrary-sized elements where all-ones is true and all-zeros is false (like with `x86_mm_cmpeq_epi32` or similar)?

# cfg(target_feature)

The `cfg(target_feature = "...")` is one of these "how come this doesn't already exist?" things. It's unnecessarily hard to use anything above SSE2 without it.

# Real-world use

Are there real-world examples of how a SIMD crate would look like?

While playing with SIMD for my [blake2-rfc](https://github.com/cesarb/blake2-rfc) crate, I wrote a miniature SIMD module (https://github.com/cesarb/blake2-rfc/tree/master/src/simd) with both a "fallback" non-SIMD variant and a SSE2 variant. The same core BLAKE2 code is used for both non-SIMD and SIMD; the difference is completely contained within that SIMD module.

While I implemented only what little functionality BLAKE2 needs (basically xor, add, rotate, and a few simple shuffles), it is a working example of a SIMD "crate", and already gives an idea of a few things that would be useful to have from the compiler: a two-vector shuffle intrinsic would be useful, having `cfg(target_feature=...)` could allow the use of rotate intrinsics and 256-bit vectors, and I'd like a guarantee that the addition wraps.

(Unfortunately, my attempt at SIMD did not work as I had expected. When I enable the SIMD code with `cargo bench --features="bench simd"`, it is ~7% slower (blake2b) or only 1% faster (blake2s) than the fallback code. Not only that, but for blake2s the fallback code is somehow faster than the code before it was converted into vectors, even though it's not using any SIMD types or intrinsics!)

Does anyone know of other real-world examples of a SIMD crate being used by real code (or even Rust code using SIMD directly without giving up and either using inline asm or calling into C code)?

cesarb:
I played a bit more with SIMD in my [blake2-rfc](https://github.com/cesarb/blake2-rfc) crate, and here are my new observations:

* Not having a way to tell cargo to pass `-C target-feature=+neon` to the compiler is a pain. Without it, LLVM explodes with a `LLVM ERROR: Do not know how to split the result of this operator!` message. By the way, the `+` is required, and it won't warn you if it's missing.
* Now I understand better the proposed simd_shuffleN intrinsic: it's LLVM's [__builtin_shufflevector](http://clang.llvm.org/docs/LanguageExtensions.html#builtin-shufflevector) (aka `shufflevector` in the LLVM IR). Not having something like it means that (if there's no intrinsic available, which is often the case since LLVM expects you to use `shufflevector`) you have to unpack and repack by hand, and hope that the compiler guesses correctly that you meant to do a shuffle.
* Having a way to convert directly between vectors with different element sizes could be useful. For instance, BLAKE2 has many rotates by "easy" amounts (16, 12, and 8 for the 32-bit BLAKE2s; 32, 24, and 16 for the 64-bit BLAKE2b). If I could get the u32x4 or u64x2 and cast it directly to a u8x16 (without changing its bit representation), and I had a decent `shufflevector`, I could replace these rotates by shuffles. (Of course, for the particular case of BLAKE2, it would be better if I had a "rotate" intrinsic and the compiler knew how to optimize it into a shuffle.)
* SIMD really makes a difference for 32-bit processors.

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+(x86-64's Skylake, Haswell, Nehalem, ...).
+
+There is definitely a common core of SIMD functionality shared across
+many platforms, but this RFC doesn't try to extract that, it is just
+building tools that can be wrapped into a more uniform API later.
+
+
+## Types
+
+There is a new attributes: `repr(simd)`.
```
shepmaster: "a new attribute"

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+on a subset of the CPUs that Rust targets, and are only be available
+for `extern`ing on those targets. The signatures are typechecked, but
+in a "duck-typed" manner: it will just ensure that the types are SIMD
+vectors with the appropriate length and element type, it will not
+enforce a specific nominal type.
+
+NB. The structural typing is just for the declaration: if a SIMD intrinsic
+is declared to take a type `X`, it must always be called with `X`,
+even if other types are structurally equal to `X`. Also, within a
+signature, SIMD types that must be structurally equal must be nominal
```
shepmaster: "nominally" ?

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+    fn simd_ne<T, U>(v: T, w: T) -> U;
+    fn simd_lt<T, U>(v: T, w: T) -> U;
+    fn simd_le<T, U>(v: T, w: T) -> U;
+    fn simd_gt<T, U>(v: T, w: T) -> U;
+    fn simd_ge<T, U>(v: T, w: T) -> U;
+}
+```
+
+These are type checked during code-generation similarly to the
+shuffles. Ensuring that `T` and `U` has the same length, and that `U`
```
shepmaster: "have the"

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,404 @@
...
+  and the generic intrinsics. This would probably require a relatively
+  complicated set of traits (with compiler integration).
+- use generic intrinsics like shuffles for the arithmetic operations,
+  instead of providing the operations implicitly.
+
+
+# Unresolved questions
+
+- Should integer vectors get `/` and `%` automatically? Most CPUs
+  don't support them for vectors. However
```
shepmaster: Rogue "however". Maybe the rest of the sentence got lost?

cesarb:
> Having a way to convert directly between vectors with different element sizes could be useful.

Actually, we already have it, and it works as expected: `mem::transmute()`. The only thing missing is a decent shuffle intrinsic (or a "rotate" intrinsic which the compiler knew how to optimize into a shuffle for constant amounts).

It would be good to add to the RFC a guarantee that `mem::transmute()` works and does what is expected (reinterpret the vector register as another type, so for instance a `u64x2` can interpreted as a `u32x4` without actually changing its contents).

(For NEON, I gave up and used inline assembly, since unlike with SSE2 there are *no* exposed shuffle intrinsics I could find.)

huonw:
@RalfJung 

>> I think this is partly a problem with the choice of types. It is easily solved by storing a wrapper around u16 rather than a u16 itself (this is explicitly supported: it is the approach most appropriate for booleans when they are defined as either all 0s or all 1s (bitwise)).

> I don't understand. How would a wrapper type help, since intrinsics can be imported based on structural matching, ignoring any wrappers?

We get to control the exact definition of "structural equality" for these purposes, and I think ensuring that this does (well, doesn't) work is a good idea.

---

@arielb1 

> The "monomorphization-error" part is a bit more annoying. I prefer to think of it as undefined behaviour detected as compile-time (monomorphized unsafe code can of-course exhibit guaranteed UB even without intrinsics). Theoretically, we could just emit an intrinsics::unreachable() - this can even be part of a useful program, if the UB isn't actually reachable - but we prefer to abort compilation.

That's a good way too look at it, although I wonder if it's too much of a post-hoc rationalisation.

---

@cesarb 

Thanks for the thoughts

> If I read this RFC correctly, it doesn't allow compiling a single executable or library which can use the best SIMD functionality available at runtime. For instance, an executable compiled for AVX2 wouldn't run on an older CPU, and an executable compiled for SSE2 wouldn't be able to make use of the AVX2 extensions.

As @Aatch and @eternaleye said, this RFC is designed for the bare minimum. Fancier dispatch based rules can and should come later. It's definitely something I've been thinking about since starting this work, and was even discussed on the pre-RFC. There's several problems to solve in that space, and it'd be good to get the basic ground-work: there's non-trivial work one can do with SIMD even in a broadly-supported cross-platform manner.

(For one, my understanding is that LLVM is still just migrating to an architecture to support per-function target specialisations, so any design proposed for allowing `__attribute__((target))` wouldn't be able to implemented in any existing Rust compiler (i.e. `rustc`) without waiting for/doing a pile of deeper LLVM work.)

> Since the point of this RFC is to allow the creation of SIMD helper crates, and not to be used directly by the end developer, wouldn't it be more flexible to not implement the primitive operations, and instead let the crate developer implement the Add, etc. traits by calling the corresponding intrinsic?

Yes this was discussed a little above, and the RFC has now been changed to use that.

> I don't see the point of flattening the types, instead of allowing only structs with the same element type repeated a few times. That part needs additional justification for why it won't be just adding unnecessary complexity to the implementation.

This is designed to allow

```rust
struct bool32(u32);
#[repr(simd)]
struct bool32x4(bool32, bool32, bool32, bool32);
```

However, this is becoming less important to me at the moment.

> some intrinsics seem to not be available through the link_llvm_intrinsics feature; for instance, I couldn't find the u64x2 shuffle intrinsics

Yes, LLVM doesn't offer intrinsics as functions for things that exist in its language.

> A single-vector shuffle is also useful, and simd_shuffle_single_4(v, 1, 2, 3, 0) is easier to understand than x86_mm_shuffle_epi32(v, 0b00_11_10_01). [...] Of course, another option would be to do a simd_shuffle4(v, v, 1, 2, 3, 0), and hope that the compiler can notice that only the first vector is being used

That is the approach that LLVM explicitly uses. I don't know about other Rust back-ends (since they don't exist yet), but it seems trivial to handle lowering `simd_shuffle4(v, v, ...)` to whatever single-argument shuffle one likes.

> And sometimes a single intrinsic isn't available; I couldn't find so far a single intrinsic to swap the two halves of an u64x2

I don't understand what you mean. The `simd_shuffle2` is explicitly designed for this.

> An alternative design for the generic shuffles would be: simd_shuffle(v, w, s) where s is some kind of Shuffle object, similar to Range (but restricted to be a compile-time constant). That might allow separating the shuffle specification from the call; for instance, one could have a (compile-type constant) fn swap_consecutive64() -> Shuffle which returned a Shuffle which swaps consecutive objects (1, 0, 3, 2, ...).

Yeah, I've considered this. However, I think this is better handled by the higher level APIs, not the raw intrinsics, e.g.

```rust
impl u32x4 {
     fn swizzle<T: Shuffle>(self) -> u32x4 {
         unsafe { simd_shuffle4(self, self, T::N0, T::N1, T::N2, T::N3) }
     }
}
trait Shuffle4 {
     const N0: u32;
     const N1: u32;
     const N2: u32;
     const N3: u32;
}
struct SwapConsec;
impl Shuffle4 {
     const N0: u32 = 1;
     const N1: u32 = 0;
     const N2: u32 = 3;
     const N3: u32 = 2;
}

// used like

let w = v.swizzle::<SwapConsec>();
```

> If the vectors are restricted to just "structs with a repeated element type", insert and extract of a constant index don't need an intrinsic; they're just v.0 = ... and ... = v.0 (or v.x = ... and ... = v.x if they're named). Of course, that doesn't work if the index is not a constant, unless the vector was declared as an array (#[repr(simd)] struct u32x4([u32; 4]) as mentioned in the alternatives).

This definitely has upsides, but there's a few downsides, like the indexing one, and it also sort-of starts to think of/use SIMD vectors as memory objects rather than registers which is their usual.

However, the major one is API uniformity, with boolean vectors. Some platforms/instruction sets represent vector boolean results as "wide" booleans, e.g. a comparison of two `f32x4`'s results in (basically) a `u32x4` where each element is either all 0s (false) or all 1s (true). But, some platforms represent them as single bits, e.g. the result of that comparison would be a `u8` (or `u32` or something) where the bottom four bits are either 0 or 1 as appropriate. The field approach doesn't work for such booleans.

... Then again, these concerns are at a higher level than this RFC: I could imagine the `extract`/`replace` API being implemented internally by directly accessing private fields, and `extract`/`replace` probably don't make much sense with the compact version anyway.



> Not having a way to tell cargo to pass -C target-feature=+neon to the compiler is a pain. Without it, LLVM explodes with a LLVM ERROR: Do not know how to split the result of this operator! message. By the way, the + is required, and it won't warn you if it's missing.

This is https://github.com/rust-lang/cargo/issues/1137. NB. `cargo rustc` allows adding features in some situations.

> Having a way to convert directly between vectors with different element sizes could be useful

As you say, this is `transmute`. Doing it safely isn't a concern of this RFC.

cesarb:
> (For one, my understanding is that LLVM is still just migrating to an architecture to support per-function target specialisations, so any design proposed for allowing `__attribute__((target))` wouldn't be able to implemented in any existing Rust compiler (i.e. `rustc`) without waiting for/doing a pile of deeper LLVM work.)

If LLVM can't do that even for C yet, I agree that it's better to wait. And now that I got a bit more of experience playing with SIMD, I can see that it's not the whole answer. It's useful if you write a fully specialized core of your algorithm for each variant, so you have for instance a SSE2 core, a SSSE3 core, and an AVX2 core; but if you write it in a generic way, like in this [BLAKE2 core](https://github.com/cesarb/blake2-rfc/blob/80b1ac828067e7d2ef0021d85cf7c00e9cf855d2/src/blake2.rs#L224-L259), it's hard to have a specialized variant for each SIMD implementation, since the specialization for each SIMD variant is in a separate module. And that's probably how a generic SIMD crate would be used; that SIMD module I wrote could be thought of as a "mini-SIMD crate with software fallback". 

> Yes, LLVM doesn't offer intrinsics as functions for things that exist in its language.

And unfortunately, Rust doesn't expose `shufflevector` yet. This RFC would fix that.

> That is the approach that LLVM explicitly uses.

From what I've seen, it does something slightly different: the equivalent of `simd_shuffle4(v, uninitialized(), ...)`. I think `simd_shuffle4(v, v, ...)` is cleaner, and I hope it works the same.

> I don't understand what you mean. The simd_shuffle2 is explicitly designed for this.

And it doesn't exist yet :cry: 

Fortunately, I've found that extracting and reconstructing works quite well; the compiler was able to determine that I meant to do a shuffle every time I tried (of course, not all attempts at a shuffle are successful; it depends on the instructions the compiler knows how to use). So something like `u64x2(foo.1, foo.0)` turns into a single shuffle instruction (like in [these shuffles](https://github.com/cesarb/blake2-rfc/blob/80b1ac828067e7d2ef0021d85cf7c00e9cf855d2/src/simd.rs#L152-L165), or even in the ones where I [abuse `mem::transmute`](https://github.com/cesarb/blake2-rfc/blob/80b1ac828067e7d2ef0021d85cf7c00e9cf855d2/src/simd.rs#L200-L227) - they do work, and generate faster code than the [generic version](https://github.com/cesarb/blake2-rfc/blob/80b1ac828067e7d2ef0021d85cf7c00e9cf855d2/src/simd.rs#L134-L141)).

> Yeah, I've considered this. However, I think this is better handled by the higher level APIs, not the raw intrinsics, e.g.

Great example!

With something like that, it's not that much of a problem to have a 66-argument shuffle intrinsic: it can be hidden behind a saner API. And since most of the point of this RFC, as far as I can now see, is to directly expose more of LLVM's internal SIMD API, if said API can have a 66-argument `shufflevector`, why not expose it?

> As you say, this is `transmute`. Doing it safely isn't a concern of this RFC.

The safety needed is just a guarantee that `mem::transmute` on a SIMD value does what one would expect: reinterpret the contents of the register (or memory) as having a different number of lanes and/or different types, without actually changing its contents. Yes, the result will depend on the lane and type layout (is lane 0 first or last? are the values two's-complement little-endian or big-endian?), but if you are using `transmute`, you are already deep in the bit-level representation of your SIMD types.

huonw:
> And unfortunately, Rust doesn't expose shufflevector yet. This RFC would fix that.

> And it doesn't exist yet :cry: 

Ah, so I was a little confused by your commentary, some of it is general comments about the current state of the world (which is quickly disappearing) and some of it is feedback on this RFC in particular.

> From what I've seen, it does something slightly different: the equivalent of simd_shuffle4(v, uninitialized(), ...). I think simd_shuffle4(v, v, ...) is cleaner, and I hope it works the same.

Yes, LLVM will presumably normalise to uninitialised internally. I was just talking about shuffles taking two arguments.

> The safety needed is just a guarantee that mem::transmute on a SIMD value does what one would expect: reinterpret the contents of the register (or memory) as having a different number of lanes and/or different types, without actually changing its contents

I was referring to safety in the Rust sense. I.e. exposing a function that doesn't require `unsafe` to call, since the fundamental operation isn't `unsafe`.

RalfJung:
> We get to control the exact definition of "structural equality" for these purposes, and I think ensuring that this does (well, doesn't) work is a good idea.

We are on the same side then: There has to be a way to make sure that my type is not considered structurally equal to any outside type. Restricting the check to visible (or just public) fields should provide that, and I think it'd be the cleanest approach. (You could then remove again the restriction that, e.g., `add` cannot mix different types.) But in the end, I don't care too much how this is enforced.

cesarb:
I ported my blake2-rfc crate to the current version of the code available at https://github.com/rust-lang/rust/pull/27169, and here are my comments (you can see what I changed at https://github.com/cesarb/blake2-rfc/commit/9e5a416d79a9e75eb474acb7f6d313c490a23036):

* The basic intrinsics like `simd_add` should be gated by the `simd_basics` feature, instead of the `platform_intrinsics` feature. The later should be reserved for platform-specific intrinsics.
* The basic intrinsics like `simd_add` should be safe to use, why do I need an `unsafe` block to call them?
* (The above also applies to shuffles, insert/extract, and compares)
* When you use `repr(C)`, it implies `allow(non_camel_case_types)`. Perhaps the same could be done for `repr(simd)`?
* Why do the shuffle intrinsics have two type parameters?

```
src/simd.rs:220:5: 222:66 error: intrinsic has wrong number of type parameters: found 1, expected 2 [E0094]
src/simd.rs:220     fn simd_shuffle8<T>(v: T, w: T,
src/simd.rs:221                         i0: u32, i1: u32, i2: u32, i3: u32,
src/simd.rs:222                         i4: u32, i5: u32, i6: u32, i7: u32) -> T;
```

I have to use `fn simd_shuffle8<T, Elem>(v: T, w: T, ...) -> T`, which means the compiler can't infer all the type parameters, so the call is an ugly `simd_shuffle8::<u16x8, u16>(tmp, tmp, ...)`, instead of the `simd_shuffle8(tmp, tmp, ...)` I could have if it had only one type parameter.

* If I add `-C target-feature=+ssse3` to the compiler command line, the SIMD code becomes slower. That is, allowing the compiler to use more instructions makes it generate slower code. Why???
* I haven't played with `cfg_target_feature` yet. I wanted to use it to know when I can use SSSE3's `pshufb`, but since merely enabling SSSE3 made the code slower, I don't know if it'll be worthwhile.

Finally, an unrelated thought about the RFC:

> scatter/gather operations allow (partially) operating on a SIMD vector of pointers. This would require allowing pointers(/references?) in repr(simd) types.

Not necessarily. At least the Intel gather instructions (`vpgather*`) operate on a scaled vector of offsets. For instance, the variant to gather a `u32x4` using 32-bit offsets from a base address is represented in C as `__m128i _mm_i32gather_epi32 (int const* base_addr, __m128i vindex, const int scale)`. In Rust, I'd expect something like `fn gather32_u32(base: &[u32], index: u32x4) -> u32x4 { ... }`

(I did an experiment calling `llvm.x86.avx2.gather.d.d` and `llvm.x86.avx2.gather.d.q.256`, but the result was much slower. I don't know if it's LLVM's fault, or if that instruction isn't as fast as I'd hoped.)

cmr:
> If I add  -C target-feature=+ssse3  to the compiler command line, the SIMD code becomes slower. That is, allowing the compiler to use more instructions makes it generate slower code. Why???

Without looking at the generated code, I have two suspicions:

1. Depending on your processor, those instructions could be microcoded and using the direct SSE2 equivalent avoids going through microcode
2. You aren't allowing *enough* extra instructions so it needs to generate stupid code to emulate something that supports using some SSSE3 feature. Try enabling all the features.

Aatch:
> * Why do the shuffle intrinsics have two type parameters?

Because the inputs might not match the output type, if you use `simd_shuffle8` you can give it two `Simd4`s, and get an `Simd8` out.

huonw:
@cesarb Thanks for taking the time to experiment with it, I'll be sure to take a look at your code tomorrow. I'll just reemphasise that this API is designed to be the absolute lowest level, for people to build better functionality APIs above, not to be used directly regularly.

> The basic intrinsics like simd_add should be gated by the simd_basics feature, instead of the platform_intrinsics feature. The later should be reserved for platform-specific intrinsics.

The intention is for the feature gate to disappear, so the details don't seem to matter that much to me. Maybe you see a deeper reason they should be separate?

> Why do the shuffle intrinsics have two type parameters?

The current design with generic SIMD types in mind, to allow the signature given in the RFC, so that one can shuffle more than just within a single type, e.g. on an x86 CPU with AVX, one could join two xmm registers (say f32x4) into a single ymm one (f32x8), or, visa versa: truncating a ymm register to an xmm. Shuffles are the general purpose way to handle this sort of munging.

> The basic intrinsics like simd_add should be safe to use, why do I need an unsafe block to call them?
(The above also applies to shuffles, insert/extract, and compares)

All `extern` functions are unsafe to call. Maybe it makes sense to make them sometimes safe to call if they happen to be so, but it's a more general "problem", and, these aren't meant to be called directly.

> When you use repr(C), it implies allow(non_camel_case_types). Perhaps the same could be done for repr(simd)?

That's sounds reasonable.

> If I add -C target-feature=+ssse3 to the compiler command line, the SIMD code becomes slower. That is, allowing the compiler to use more instructions makes it generate slower code. Why???

This is very likely to be caused by LLVM, not rustc or the functionality I added. It may be worth investigating the IR/asm and filing an LLVM bug if you can narrow it down.  `-C target-cpu=native` might be a useful tool in comparing the native code LLVM wants to generate for your CPU with what it is assuming runs fast with just `+ssse3`.

> (I did an experiment calling llvm.x86.avx2.gather.d.d and llvm.x86.avx2.gather.d.q.256, but the result was much slower. I don't know if it's LLVM's fault, or if that instruction isn't as fast as I'd hoped.)

Combination of both I imagine.

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,422 @@
...
+#[repr(simd)]
+struct Simd2<T>(T, T);
+```
+
+The `simd` `repr` can be attached to a struct and will cause such a
+struct to be compiled to a SIMD vector. It can be generic, but it is
+required that any fully monomorphised instance of the type consist of
+only a single "primitive" type, repeated some number of times. Types
+are flattened, so, for `struct Bar(u64);`, `Simd2<Bar>` has the same
+representation as `Simd2<u64>`.
```
pnkfelix: i'm wondering if this "flatten-able" repr, combined with the structural-typing (aka psuedo-duck-typing) discipline for intrinsics, may have utility for other non-simd intrinsics.

If so, then perhaps we could separate the two concerns here, by having `#[repr(simd)]` (which on its own could only be used for structs composed from primitive numeric types) and `#[repr(flatten)]`, which would be used for structs that opt-in to the duck-typed intrinsic discipline.

----

(also, is it worth considering actually specifying the subset of `struct` definitions that can be used with this flattening system?  Perhaps  just specified via a grammar?)

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+These all use entirely concrete types, and this is the core interface
+to these intrinsics: essentially it is just allowing code to exactly
+specify a CPU instruction to use. These intrinsics only actually work
+on a subset of the CPUs that Rust targets, and are only be available
+for `extern`ing on those targets. The signatures are typechecked, but
+in a "duck-typed" manner: it will just ensure that the types are SIMD
+vectors with the appropriate length and element type, it will not
+enforce a specific nominal type.
```
pnkfelix: as I noted above, this duck-typing may be useful for other intrinsics.  But maybe I am just being overly optimistic and/or aggressive. 

(I basically am just idly thinking about a [recent PR](https://github.com/rust-lang/rust/pull/27215) I wrote with a [commit](https://github.com/pnkfelix/rust/commit/75da81e78231a6fb5d90b3918bebe867249eb14c) that changed the signature of `move_val_init` from taking `&mut T` to taking `*mut T`, and wondering if some pain would have been avoided if the intrinsic system were more flexible in this regard. Its probably a bad idea.)

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,422 @@
...
+platform specific intrinsic for shuffling.
+
+```rust
+extern "platform-intrinsic" {
+    fn simd_shuffle2<T, Elem>(v: T, w: T, i0: u32, i1: u32) -> Simd2<Elem>;
+    fn simd_shuffle4<T, Elem>(v: T, w: T, i0: u32, i1: u32, i2: u32, i3: u32) -> Sidm4<Elem>;
+    fn simd_shuffle8<T, Elem>(v: T, w: T,
+                              i0: u32, i1: u32, i2: u32, i3: u32,
+                              i4: u32, i5: u32, i6: u32, i7: u32) -> Simd8<Elem>;
+    fn simd_shuffle16<T, Elem>(v: T, w: T,
```
pnkfelix: I assume that there's some fundamental reason (e.g. intrinsic codegen issues and/or const-eval issues) that these are taking all of these arguments explicitly rather than `[u32; 2]`, `[u32; 4]`, `[u32; 8]` and `[u32; 16]`?

cesarb:
After posting my previous comment, I received an email with a [failure](https://travis-ci.org/cesarb/blake2-rfc/jobs/74510567) from travis-ci which points to another problem:

```
src/simd.rs:33:8: 33:28 error: illegal ABI: expected one of [cdecl, stdcall, fastcall, aapcs, win64, Rust, C, system, rust-intrinsic, rust-call], found `platform-intrinsic`

src/simd.rs:33 extern "platform-intrinsic" {

                      ^~~~~~~~~~~~~~~~~~~~

src/simd.rs:41:8: 41:28 error: illegal ABI: expected one of [cdecl, stdcall, fastcall, aapcs, win64, Rust, C, system, rust-intrinsic, rust-call], found `platform-intrinsic`

src/simd.rs:41 extern "platform-intrinsic" {

                      ^~~~~~~~~~~~~~~~~~~~
```

It seems that `#[cfg(feature = "...")]` was not enough to hide the `extern` block from the compiler, it insists in validating the ABI name even though it's supposed to be ignoring the whole block. And that's the "stable" compiler, it was supposed to work because I define no cargo features (the "nightly" features are all gated by cargo features in that crate), so the `extern` blocks were supposed to be ignored.

cesarb:
@cmr 

> You aren't allowing enough extra instructions so it needs to generate stupid code to emulate something that supports using some SSSE3 feature. Try enabling all the features.

@huonw 

>  `-C target-cpu=native` might be a useful tool in comparing the native code LLVM wants to generate for your CPU with what it is assuming runs fast with just `+ssse3`

The problem is, `target-cpu=native` enables AVX2, which somehow makes it go even slower. A few benchmarks with all the cargo features enabled follow. Enabling just the "simd" cargo feature is slower than the non-SIMD code, probably because this CPU can run 4 integer instructions in parallel, and SIMD rotates need 3 instructions, while each non-SIMD rotate needs 1 instruction (and the 4 rotates can run in parallel). The gains when I enable the "simd_opt" cargo feature all come from replacing the SIMD rotates by single-instruction (or two-instruction for u64x4) shuffles.

* default (the fastest I've got so far):

```
test blake2b::bench::bench_16  ... bench:         198 ns/iter (+/- 2) = 80 MB/s
test blake2b::bench::bench_4k  ... bench:       5,766 ns/iter (+/- 44) = 710 MB/s
test blake2b::bench::bench_64k ... bench:      92,115 ns/iter (+/- 1,288) = 711 MB/s
test blake2s::bench::bench_16  ... bench:         141 ns/iter (+/- 1) = 113 MB/s
test blake2s::bench::bench_4k  ... bench:       7,997 ns/iter (+/- 86) = 512 MB/s
test blake2s::bench::bench_64k ... bench:     127,775 ns/iter (+/- 675) = 512 MB/s
```

* with `-C target-features=+ssse3`:

```
test blake2b::bench::bench_16  ... bench:         208 ns/iter (+/- 37) = 76 MB/s
test blake2b::bench::bench_4k  ... bench:       5,939 ns/iter (+/- 39) = 689 MB/s
test blake2b::bench::bench_64k ... bench:      95,356 ns/iter (+/- 833) = 687 MB/s
test blake2s::bench::bench_16  ... bench:         141 ns/iter (+/- 1) = 113 MB/s
test blake2s::bench::bench_4k  ... bench:       7,995 ns/iter (+/- 1,659) = 512 MB/s
test blake2s::bench::bench_64k ... bench:     127,766 ns/iter (+/- 758) = 512 MB/s
```

* with `-C target-cpu=native` (which uses AVX2 instructions everywhere):

```
test blake2b::bench::bench_16  ... bench:         230 ns/iter (+/- 7) = 69 MB/s
test blake2b::bench::bench_4k  ... bench:       6,592 ns/iter (+/- 70) = 621 MB/s
test blake2b::bench::bench_64k ... bench:     105,376 ns/iter (+/- 345) = 621 MB/s
test blake2s::bench::bench_16  ... bench:         143 ns/iter (+/- 1) = 111 MB/s
test blake2s::bench::bench_4k  ... bench:       8,181 ns/iter (+/- 57) = 500 MB/s
test blake2s::bench::bench_64k ... bench:     130,863 ns/iter (+/- 4,278) = 500 MB/s
```

> This is very likely to be caused by LLVM, not rustc or the functionality I added.

I agree.

> I'll just reemphasise that this API is designed to be the absolute lowest level, for people to build better functionality APIs above, not to be used directly regularly.

That's what  I'm experimenting with: the `lib/simd.rs` file builds a better functionality API above it, while the `lib/blake2.rs` file uses it (and required no change when porting to this proposed RFC).

cesarb:
> The current design with generic SIMD types in mind, to allow the signature given in the RFC, so that one can shuffle more than just within a single type, e.g. on an x86 CPU with AVX, one could join two xmm registers (say f32x4) into a single ymm one (f32x8), or, visa versa: truncating a ymm register to an xmm. Shuffles are the general purpose way to handle this sort of munging.

I see. I have one place which might be able to use it:

```
fn u64x4_rotate_right_u8(vec: u64x4, n: u8) -> u64x4 {
    let tmp0 = vext_u64_u8(u64x2(vec.0, vec.1), n);
    let tmp1 = vext_u64_u8(u64x2(vec.2, vec.3), n);
    u64x4(tmp0.0, tmp0.1, tmp1.0, tmp1.1)
}
```

Could be replaced by something like:

```
fn u64x4_rotate_right_u8(vec: u64x4, n: u8) -> u64x4 {
    let tmp0 = vext_u64_u8(simd_shuffle_2(vec, vec, 0, 1), n);
    let tmp1 = vext_u64_u8(simd_shuffle_2(vec, vec, 2, 3), n);
    simd_shuffle4(tmp0, tmp1, 0, 1, 2, 3)
}
```

But first I'd have to change all the vector types to be generic in the element type.

cesarb:
> But first I'd have to change all the vector types to be generic in the element type.

Which is harder than it looks. This doesn't work:

```
#[derive(Clone, Copy, Debug)]
#[repr(simd)]
pub struct Simd4<T>(pub T, pub T, pub T, pub T);

pub type u32x4 = Simd4<u32>;
```

Since it doesn't find the constructor e.g. `u32x4(0, 0, 0, 0)`. Either I add a `fn new(...) -> Self` to the type's `impl` (touching every place that constructs a SIMD value), or I leak the fact that it's a generic type outside the SIMD code (touching every place that uses a SIMD value).

I don't know which would be the better design. The traditional Rust SIMD design of `u32x4`, or the generic `Simd4<u32>` variant. This [SIMD crate prototype](https://github.com/huonw/simd) uses the former, while this RFC tends towards the later.

Aatch:
@cesarb 

> But first I'd have to change all the vector types to be generic in the element type.

What was the exact issue you had? I haven't looked at the prototype, but it's possible that it doesn't yet have the relevant functionality to support that case. Please remember that this comment thread is for the RFC, not the prototype implementation, so pointing out issues with the prototype is only relevant if the behaviour is consistent with the RFC.

text/0000-simd-infrastructure.md
arielb1: `transmute` is already duck-typed. There is nothing new here.

text/0000-simd-infrastructure.md
huonw: Hm, are you visualising something like the following being valid?

```rust
struct Foo(f32);
#[repr(simd, flatten)]
struct Foox4(Foo, Foo, Foo, Foo);
```

(And if there was no `flatten`, it would be invalid?)

---

In any case, I'm actually going to remove this aspect of the RFC: I'm not sure it's pulling its weight: it is moderately non-trivial for me to implement, and my experimentations have ended up working well without it.

text/0000-simd-infrastructure.md
huonw: There's no fundamental reason for it that I know of (although I'm not 100% sure if the compiler currently supports the necessary const-eval stuff, but that should just be an implementation detail). In any case, your approach does seem like it may be better.

huonw:
Updated. Highlights:

- removal of struct flattening and the internal-reference restriction
- shuffles use @pnkfelix's suggestion for an array of indices (so much nicer!)

text/0000-simd-infrastructure.md
```diff
@@ -0,0 +1,423 @@
...
+to the `ADC` instruction).
+
+### Shuffles & element operations
+
+One of the most powerful features of SIMD is the ability to rearrange
+data within vectors, giving super-linear speed-ups sometimes. As such,
+shuffles are exposed generally: intrinsics that represent arbitrary
+shuffles.
+
+This may violate the "one instruction per instrinsic" principal
```
ranma42: "principal" seems to be a typo (for "principle")

nikomatsakis:
**Hear ye, hear ye.** This RFC is entering **final comment period.**

glaebhoerl:
What's the basic plan with respect to stabilization of these features? As long as the idea is just to get these into the unstable compiler, so that the implementation PR can land and we can start experimenting more seriously with higher-level APIs, further evolving the lower-level ones, and whatever else, that's *completely* cool. But the number of places (there's a few) where we're currently forced to say "enforcing this in the type system is difficult-to-impossible right now, so let's punt it to the backend" still bothers me if it's something we'd be committing to support as a stable feature forever (or until 2.0, anyways).

nikomatsakis:
@glaebhoerl I think there is no plan to stabilize until experience has been gained, but of course the further we go down this road, the less likely we'll back up and start from another.

glaebhoerl:
Basically, I suspect the approach I'd prefer is that once we have this infrastructure, and we've used it to gain some experience and to figure out the best way to formulate higher-level "type-safe" (i.e. without checks deferred to codegen) abstractions for SIMD (various `trait`s, etc.), we should go ahead and stabilize *those*.

nikomatsakis:
To be clear, I agree with you Gabor -- I'd like to gain experience in what
we want to do, first, and then come back and see if we can find the most
elegant way to do it. (Not that this approach is unacceptable, it's quite
elegant in its own way.) I'm just acknowledging the power of incumbency. :)

nikomatsakis:
**It's official.** The language subteam [has decided](https://docs.google.com/spreadsheets/d/18FtPnEI8RTPdFhGDtj21ITndakEtxVrs8pVzRNx-jL8/edit#gid=0) to accept this RFC.

nikomatsakis:
Tracking issue is https://github.com/rust-lang/rust/issues/27731

PR #1201, opened 2015-07-11
Add support for naked functions.

tari:
[Rendered](https://github.com/tari/rfcs/blob/naked-fns/text/0000-naked-fns.md).

### DIFF
--- a/text/0000-naked-fns.md
+++ b/text/0000-naked-fns.md
@@ -87,17 +87,14 @@ any calling convention the compiler is compatible with, calls to naked functions
 from within Rust code are forbidden unless the function is also declared with
 a well-defined ABI.
 
-The function `call_foo` in the following code block is an error because the
-default (Rust) ABI is unspecified and as such a programmer can never write code
-in `foo` which is compatible:
+Defining a naked function with the default (Rust) ABI is an error, because the
+Rust ABI is unspecified and the programmer can never write a function which is
+guaranteed to be compatible. For example, The function declaration of `foo` in
+the following code block is an error.
 
 ```rust
 #[naked]
-fn foo() { }
-
-fn call_foo() {
-    foo();
-}
+unsafe fn foo() { }
 ```
 
 The following variant is not an error because the C calling convention is
@@ -107,28 +104,36 @@ function:
 ```rust
 #[naked]
 extern "C" fn foo() { }
-
-fn call_foo() {
-    foo();
-}
 ```
 
 ---
 
-The current support for `extern` functions in `rustc` generates a minimum of two
-basic blocks for any function declared in Rust code with a non-default calling
-convention: a trampoline which translates the declared calling convention to the
-Rust convention, and a Rust ABI version of the function containing the actual
-implementation. Calls to the function from Rust code call the Rust ABI version
-directly.
+Because the compiler cannot verify the correctness of code written in a naked
+function (since it may have an unknown calling convention), naked functions must
+be declared `unsafe` or contain no non-`unsafe` statements in the body. The
+function `error` in the following code block is a compile-time error, whereas
+the functions `correct1` and `correct2` are permitted.
 
-For naked functions, it is impossible for the compiler to generate a Rust ABI
-version of the function because the implementation may depend on the calling
-convention. In cases where calling a naked function from Rust is permitted, the
-compiler must be able to use the target calling convention directly rather than
-call the same function with the Rust convention.
+```
+#[naked]
+extern "C" fn error(x: &mut u8) {
+    *x += 1;
+}
 
----
+#[naked]
+unsafe extern "C" fn correct1(x: &mut u8) {
+    *x += 1;
+}
+
+#[naked]
+extern "C" fn correct2() {
+    unsafe {
+        *x += 1;
+    }
+}
+```
+
+## Example
 
 The following example illustrates the possible use of a naked function for
 implementation of an interrupt service routine on 32-bit x86.
@@ -162,6 +167,21 @@ fn main() {
 }
 ```
 
+## Implementation Considerations
+
+The current support for `extern` functions in `rustc` generates a minimum of two
+basic blocks for any function declared in Rust code with a non-default calling
+convention: a trampoline which translates the declared calling convention to the
+Rust convention, and a Rust ABI version of the function containing the actual
+implementation. Calls to the function from Rust code call the Rust ABI version
+directly.
+
+For naked functions, it is impossible for the compiler to generate a Rust ABI
+version of the function because the implementation may depend on the calling
+convention. In cases where calling a naked function from Rust is permitted, the
+compiler must be able to use the target calling convention directly rather than
+call the same function with the Rust convention.
+
 # Drawbacks
 
 The utility of this feature is extremely limited to most users, and it might be
@@ -179,8 +199,20 @@ external libraries such as `libffi`.
 
 It is easy to quietly generate wrong code in naked functions, such as by causing
 the compiler to allocate stack space for temporaries where none were
-anticipated. It may be desirable to require that all statements inside naked
-functions be inside `unsafe` blocks (either by declaring the function `unsafe`
-or including `unsafe { }` in the function body) to reinforce the need for
-extreme care in the use of this feature. Requiring that the function always be
-marked `unsafe` is not desirable because its external API may be safe.
+anticipated. There is currently no restriction on writing Rust statements inside
+a naked function, while most compilers supporting similar features either
+require or strongly recommend that authors write only inline assembly inside
+naked functions to ensure no code is generated that assumes a particular stack
+layout. It may be desirable to place further restrictions on what statements are
+permitted in the body of a naked function, such as permitting only `asm!`
+statements.
+
+The `unsafe` requirement on naked functions may not be desirable in all cases.
+However, relaxing that requirement in the future would not be a breaking change.
+
+Because a naked function may use a calling convention unknown to the compiler,
+it may be useful to add a "unknown" calling convention to the compiler which is
+illegal to call directly. Absent this feature, functions implementing an unknown
+ABI would need to be declared with a calling convention which is known to be
+incorrect and depend on the programmer to avoid calling such a function
+incorrectly since it cannot be prevented statically.
--- a/text/0000-naked-fns.md
+++ b/text/0000-naked-fns.md
@@ -10,12 +10,19 @@ function attribute.
 
 # Motivation
 
-Some systems programming tasks require that machine state not be modified at all
-on function entry so it can be preserved- particularly in interrupt handlers.
-For example, x86\_64 preserves only the stack pointer, flags register, and
-instruction pointer on interrupt entry. To avoid corrupting program state, the
-interrupt handler must save the registers which might be modified before handing
-control to compiler-generated code. Consider a contrived interrupt handler:
+Some systems programming tasks require that the programmer have complete control
+over function stack layout and interpretation, generally in cases where the
+compiler lacks support for a specific use case. While these cases can be
+addressed by building the requisite code with external tools and linking with
+Rust, it is advantageous to allow the Rust compiler to drive the entire process,
+particularly in that code may be generated via monomorphization or macro
+expansion.
+
+When writing interrupt handlers for example, most systems require additional
+state be saved beyond the usual ABI requirements.  To avoid corrupting program
+state, the interrupt handler must save the registers which might be modified
+before handing control to compiler-generated code. Consider a contrived
+interrupt handler for x86\_64:
 
 ```rust
 unsafe fn isr_nop() {
@@ -47,14 +54,25 @@ stack layout for any given function are not predictable (and may change with
 compiler version or optimization settings), attempting to predict the stack
 layout to sidestep this issue is infeasible.
 
-In other languages (particularly C), "naked" functions omit the prologue and
-epilogue (represented by the modifications to `rsp` in the above example) to
-allow the programmer complete control over stack layout. This makes the
-availability of stack space for compiler use unpredictable, usually implying
-that the body of such a function must consist entirely of inline assembly
-statements (such as a jump or call to another function).
+When interacting with FFIs that are not natively supported by the compiler,
+a similar situation arises where the programmer knows the expected calling
+convention and can implement a translation between the foreign ABI and one
+supported by the compiler.
 
-The [LLVM language
+Support for naked functions also allows programmers to write functions that
+would otherwise be unsafe, such as the following snippet which returns the
+address of its caller when called with the C ABI on x86.
+
+```
+    mov 4(%ebp), %eax
+    ret
+```
+
+---
+
+Because the compiler depends on a function prologue and epilogue to maintain
+storage for local variable bindings, it is generally unsafe to write anything
+but inline assembly inside a naked function.  The [LLVM language
 reference](http://llvm.org/docs/LangRef.html#function-attributes) describes this
 feature as having "very system-specific consequences", which the programmer must
 be aware of.
@@ -64,23 +82,83 @@ be aware of.
 Add a new function attribute to the language, `#[naked]`, indicating the
 function should have prologue/epilogue emission disabled.
 
-For example, the following construct could be assumed not to generate extra code
-on entry to `isr_caller` which might violate the programmer's assumptions, while
-allowing the compiler to generate the function definition as usual:
+Because the calling convention of a naked function is not guaranteed to match
+any calling convention the compiler is compatible with, calls to naked functions
+from within Rust code are forbidden unless the function is also declared with
+a well-defined ABI.
+
+The function `call_foo` in the following code block is an error because the
+default (Rust) ABI is unspecified and as such a programmer can never write code
+in `foo` which is compatible:
 
 ```rust
 #[naked]
-unsafe fn isr_caller() {
-    asm!("push %rax
-          call other_function
-          pop %rax
-          iretq" :::: "volatile");
-    core::intrinsics::unreachable();
+fn foo() { }
+
+fn call_foo() {
+    foo();
+}
+```
+
+The following variant is not an error because the C calling convention is
+well-defined and it is thus possible for the programmer to write a conforming
+function:
+
+```rust
+#[naked]
+extern "C" fn foo() { }
+
+fn call_foo() {
+    foo();
+}
+```
+
+---
+
+The current support for `extern` functions in `rustc` generates a minimum of two
+basic blocks for any function declared in Rust code with a non-default calling
+convention: a trampoline which translates the declared calling convention to the
+Rust convention, and a Rust ABI version of the function containing the actual
+implementation. Calls to the function from Rust code call the Rust ABI version
+directly.
+
+For naked functions, it is impossible for the compiler to generate a Rust ABI
+version of the function because the implementation may depend on the calling
+convention. In cases where calling a naked function from Rust is permitted, the
+compiler must be able to use the target calling convention directly rather than
+call the same function with the Rust convention.
+
+---
+
+The following example illustrates the possible use of a naked function for
+implementation of an interrupt service routine on 32-bit x86.
+
+```rust
+use std::intrinsics;
+use std::sync::atomic::{self, AtomicUsize, Ordering};
+
+#[naked]
+#[cfg(target_arch="x86")]
+unsafe fn isr_3() {
+    asm!("pushad
+          call increment_breakpoint_count
+          popad
+          iretd" :::: "volatile");
+    intrinsics::unreachable();
 }
 
-#[no_mangle]
-pub fn other_function() {
+static bp_count: AtomicUsize = ATOMIC_USIZE_INIT;
 
+#[no_mangle]
+pub fn increment_breakpoint_count() {
+    bp_count.fetch_add(1, Ordering::Relaxed);
+}
+
+fn register_isr(vector: u8, handler: fn() -> ()) { /* ... */ }
+
+fn main() {
+    register_isr(3, isr_3);
+    // ...
 }
 ```
 
@@ -93,22 +171,16 @@ considered.
 # Alternatives
 
 Do nothing. The required functionality for the use case outlined can be
-implemented outside Rust code (such as with a small amount of externally-built
-assembly) and merely linked in as needed.
-
-Add a new calling convention (`extern "interrupt" fn ...`) which is defined to
-do any necessary state saving for interrupt service routines. This permits more
-efficient code to be generated for the motivating example (omitting a 'call'
-instruction which is necessary for any non-trivial ISR), but may not be
-appropriate for other situations that might call for a naked function.
-Implementation of additional calling conventions like this in the current
-`rustc` would involve significant modification to LLVM to support it (whereas
-the proof-of-concept patch for `#[naked]` is less than 10 lines of code).
+implemented outside Rust code and linked in as needed. Support for additional
+calling conventions could be added to the compiler as needed, or emulated with
+external libraries such as `libffi`.
 
 # Unresolved questions
 
 It is easy to quietly generate wrong code in naked functions, such as by causing
 the compiler to allocate stack space for temporaries where none were
-anticipated. It may be desirable to allow the `#[naked]` attribute on `unsafe`
-functions only, reinforcing the need for extreme care in the use of this
-feature.
+anticipated. It may be desirable to require that all statements inside naked
+functions be inside `unsafe` blocks (either by declaring the function `unsafe`
+or including `unsafe { }` in the function body) to reinforce the need for
+extreme care in the use of this feature. Requiring that the function always be
+marked `unsafe` is not desirable because its external API may be safe.
--- a/text/0000-naked-fns.md
+++ b/text/0000-naked-fns.md
@@ -1,0 +1,114 @@
+- Feature Name: `naked_fns`
+- Start Date: 2015-07-10
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Add support for generating naked (prologue/epilogue-free) functions via a new
+function attribute.
+
+# Motivation
+
+Some systems programming tasks require that machine state not be modified at all
+on function entry so it can be preserved- particularly in interrupt handlers.
+For example, x86\_64 preserves only the stack pointer, flags register, and
+instruction pointer on interrupt entry. To avoid corrupting program state, the
+interrupt handler must save the registers which might be modified before handing
+control to compiler-generated code. Consider a contrived interrupt handler:
+
+```rust
+unsafe fn isr_nop() {
+    asm!("push %rax"
+         /* Additional pushes elided */ :::: "volatile");
+    let n = 0u64;
+    asm!("pop %rax"
+         /* Additional pops elided */ :::: "volatile");
+}
+```
+
+The generated assembly for this function might resemble the following
+(simplified for readability):
+
+```x86
+isr_nop:
+    sub $8, %rsp
+    push %rax
+    movq $0, 0(%rsp)
+    pop %rax
+    add $8, %rsp
+    retq
+```
+
+Here the programmer's need to save machine state conflicts with the compiler's
+assumption that it has complete control over stack layout, with the result that
+the saved value of `rax` is clobbered by the compiler. Given that details of
+stack layout for any given function are not predictable (and may change with
+compiler version or optimization settings), attempting to predict the stack
+layout to sidestep this issue is infeasible.
+
+In other languages (particularly C), "naked" functions omit the prologue and
+epilogue (represented by the modifications to `rsp` in the above example) to
+allow the programmer complete control over stack layout. This makes the
+availability of stack space for compiler use unpredictable, usually implying
+that the body of such a function must consist entirely of inline assembly
+statements (such as a jump or call to another function).
+
+The [LLVM language
+reference](http://llvm.org/docs/LangRef.html#function-attributes) describes this
+feature as having "very system-specific consequences", which the programmer must
+be aware of.
+
+# Detailed design
+
+Add a new function attribute to the language, `#[naked]`, indicating the
+function should have prologue/epilogue emission disabled.
+
+For example, the following construct could be assumed not to generate extra code
+on entry to `isr_caller` which might violate the programmer's assumptions, while
+allowing the compiler to generate the function definition as usual:
+
+```rust
+#[naked]
+unsafe fn isr_caller() {
+    asm!("push %rax
+          call other_function
+          pop %rax
+          iretq" :::: "volatile");
+    core::intrinsics::unreachable();
+}
+
+#[no_mangle]
+pub fn other_function() {
+
+}
+```
+
+# Drawbacks
+
+The utility of this feature is extremely limited to most users, and it might be
+misused if the implications of writing a naked function are not carefully
+considered.
+
+# Alternatives
+
+Do nothing. The required functionality for the use case outlined can be
+implemented outside Rust code (such as with a small amount of externally-built
+assembly) and merely linked in as needed.
+
+Add a new calling convention (`extern "interrupt" fn ...`) which is defined to
+do any necessary state saving for interrupt service routines. This permits more
+efficient code to be generated for the motivating example (omitting a 'call'
+instruction which is necessary for any non-trivial ISR), but may not be
+appropriate for other situations that might call for a naked function.
+Implementation of additional calling conventions like this in the current
+`rustc` would involve significant modification to LLVM to support it (whereas
+the proof-of-concept patch for `#[naked]` is less than 10 lines of code).
+
+# Unresolved questions
+
+It is easy to quietly generate wrong code in naked functions, such as by causing
+the compiler to allocate stack space for temporaries where none were
+anticipated. It may be desirable to allow the `#[naked]` attribute on `unsafe`
+functions only, reinforcing the need for extreme care in the use of this
+feature.


eddyb:
:+1: This would allow using only Rust + inline assembly in a kernel.
Definitely need to restrict this to either `unsafe` functions, or have `unsafe` in the name of the attribute (I prefer the former).

DanielKeep:
How does this interact with the calling convention?

I recently looked into writing some inline assembly, using `extern "C"` so that I knew the layout of the stack, but it looks like in Rust this just creates an `extern "C"` landing pad that calls an `extern "rust"` function which is where the code *actually* lives.

If all this does is let us replace the contents of an `extern "rust"` function, and the Rust ABI is deliberately undefined, how is one supposed to write code that accesses arguments or returns anything?

I'm inclined to suggest that `#[naked]` be restricted not only to `unsafe` functions, but functions that use a "well-defined" ABI such as `C` or `stdcall`.  Unless the Rust ABI is defined, there's no way `#[naked] unsafe extern "rust" fn() { ... }` can have any well-defined meaning.

Tobba:
This doesn't seem like a terribly good idea, LLVM lacked this for a very long time due to the undefined and extremely iffy semantics involved, essentially the *only* thing that can safely be done in a naked function is using a single asm block. LLVM already has had support for module-level assembly blocks as long as I can remember.

Although I wouldn't let any remotely complicated hand-written code get within 50 feet of LLVMs built-in assembler (it tends to be somewhat broken)

tari:
> If all this does is let us replace the contents of an `extern "rust"` function, and the Rust ABI is deliberately undefined, how is one supposed to write code that accesses arguments or returns anything?

I think you're right that anything but parameterless functions returning `()` are difficult or impossible to write as naked. Interrupt handlers always match those criteria, but that seems like a good case in support of the `extern "interrupt"` alternative since doing anything else with a naked function is so sketchy.

> LLVM already has had support for module-level assembly blocks as long as I can remember.

Without Rust support though, that doesn't get us very far. The advantage of putting this directly on functions is that we allow the compiler to generate a function definition, including name mangling.

---

Despite the much greater implementation complexity, I'm coming around to the `extern "interrupt"` option as more robust and safer than this one, especially if nobody can come up with a non-interrupt use case for naked functions.

DanielKeep:
@tari: My point wasn't "you can't write naked functions that take arguments", it was "you can't write naked functions that take arguments using the undefined `rust` ABI".  I've only written a few naked functions over the years, but every time I did, they *had* to be naked, and they *had* to take arguments.  I think they were *all* trampolines of one kind or another.

I think the *good* solution is to make sure that naked functions aren't "desugared" into `extern "rust"` functions wrapped in an `extern "$OTHER"` trampoline.  After all, I would only expect them to contain a single `asm!` directive, anyway, so the compiler shouldn't *need* to translate the ABI.

`extern "interrupt"`, on the other hand, feels more like a "this is the only use case *I* care about" solution. :P

Tobba:
`extern "interrupt"` gives the impression that the ABI is only for interrupt handlers, while that is not the case. How  about `extern "naked"`?

Either way, the compiler could provide name mangling for module-level assembly blocks, which would be much saner.

Diggsey:
Isn't name mangling seperate from the ABI already in rust, ie. `extern "C"` doesn't disable name mangling?

One problem with `extern "interrupt"`/`extern "naked"` is that it doesn't specify how the function should be called. For interrupts it's not a problem because you never call them directly, but there may be cases where you need to call a naked function directly from other rust code - in that case the compiler needs to know which ABI to use to call the function.

For this reason, I think it would be better to leave `#[naked]` as an attribute, but disallow its use in conjunction with `extern "rust"` functions.

Finally, there may be cases where the naked function doesn't follow any standard ABI, in which case it shouldn't be callable at all from rust code. This "unspecified" ABI could be selected via `extern "unspecified"`, in which case the compiler would prevent it being called except via assembly.

tari:
> I think the good solution is to make sure that naked functions aren't "desugared" into `extern "rust"` functions wrapped in an `extern "$OTHER"` trampoline. After all, I would only expect them to contain a single `asm!` directive, anyway, so the compiler shouldn't need to translate the ABI.

The trouble with that approach I see is that it becomes impossible to safely call such a naked function from Rust. We could choose to forbid native rust calls to such a function, but that seems to me like an unusual exception to the rule that all declared functions are callable.

> `extern "interrupt"` gives the impression that the ABI is only for interrupt handlers, while that is not the case. How about `extern "naked"`?

Right; I started conflating the goal and means a bit here. Clarifying: my ultimate goal is to support writing interrupt handlers in Rust-only. The proposed means to that is introducing naked functions which allow the programmer to use an arbitrary calling convention.

As `extern` functions are implemented today, we cannot implement arbitrary ABI trampolines because the Rust ABI is unspecified and `extern` functions are implemented by the compiler as a trampoline to translate the calling convention to Rust and call the Rust-native version of a function. This is not a concern for interrupt handlers because they don't take any parameters, nor do they return any values.

Changing this approach would have the implication that some calls to functions defined in Rust become wrong because they assume a different calling convention- a hypothetical `extern "naked"` function implemented as a module-level assembly block would be unsafe to call from Rust because the ABIs will conflict and it is not possible for the compiler to generate a Rust-ABI version of the function.

---

Given the previous points, the simplest approach to supporting arbitrary ABIs appears to be implementing them in the compiler. This is much less flexible than supporting naked functions, but permits more efficient code generation since the compiler has detailed knowledge of what the system is doing. For the motivating use case this could be `extern "interrupt"` to use a target-specific interrupt ABI, or something like `extern "fortran"` for Fortran FFI.

Addition of calling conventions doesn't preclude a generic mechanism for naked functions, but it does sidestep the current issues involved in making naked functions useful.

DanielKeep:
> The trouble with that approach I see is that it becomes impossible to safely call such a naked function from Rust. We could choose to forbid native rust calls to such a function, but that seems to me like an unusual exception to the rule that all declared functions are callable.

A naked function is no less safe to call than any other non-local `extern "C"` function.  Heck, that C ABI function you're linking to might very well be implemented in pure assembly for all you know!

I get the feeling that we're starting to talk across purposes, here, so let's clarify.  I'm talking about something like:

```rust
#[cfg(target_arch="x86")]
extern "C" fn add_two(a: i32, b: i32) -> i32 {
    #![naked]
    asm!(r#"
        mov eax, [esp+4]
        mov ecx, [esp+8]
        add eax, ecx
    "#)
}

fn main() {
    assert_eq!(add_two(4, 6), 10);
}
```

Unless I've got the assembler wrong (it's been a while), the above should be perfectly safe to call.  It's a function that uses the C ABI... that it's written in assembler and doesn't have the standard function prelude is just an implementation detail the user doesn't need to worry about.

I mean, `rustc` **has** to be able to call `extern "C"` functions; it's just that the code it generates when it's defining them is a bit unexpected.

tari:
> A naked function is no less safe to call than any other non-local extern "C" function. Heck, that C ABI function you're linking to might very well be implemented in pure assembly for all you know!

The difference is that the only reason to use a naked function is when the compiler does not understand the ABI you need to use and is thus virtually guaranteed to generate code that will do something unexpected (and wrong) if you try to call it. You *can* implement a naked function that uses the C ABI, but there's no reason to do so because the compiler already understands how to deal with that for you.

Diggsey:
@tari That's simply not true - there are many reasons to require a naked function even when using an ABI that the compiler understands.

tari:
Can you provide some examples of situations you'd want to do so? The way I see it, naked functions are useful to fill holes in the compiler's ABI support and I don't see the point of using one when the compiler is capable of doing what you need.

Diggsey:
Sometimes you can implement a method more efficiently than the compiler's default prolog/epilog would allow. For example, I needed to write a function which returned the address of the caller. With a naked function, this boils down to two instructions:
```asm
mov eax, [ebp+4]
ret
```
There's just no way to generate code like that without getting rid of the default prolog and epilog.

main--:
@Diggsey While specific stack layout (e.g. caller address being at `[ebp+4]`) is undefined in the rust ABI, prolog/epilog optimization is generally the compiler's job. So if we just ignore the fact that the result of this function is essentially undefined, you can do this:

```rust
#![feature(asm)]

pub fn myfunc() -> u32 {
    let ret: u32;
    unsafe { asm!("mov $0, [ebp+4]" : "=r"(ret) ::: "intel"); }
    return ret;
}
// yes, I'm compiling this on x86_64, so it won't ever work like this anyways but it really doesn't matter for this example
```

```
$ rustc -O --crate-type rlib test.rs
$ objdump -CdMintel libtest.rlib 
Disassembly of section .text._ZN6myfunc20h9064a5bb1ed9f519eaaE:

0000000000000000 <myfunc::h9064a5bb1ed9f519eaa>:
   0:	67 8b 45 04          	mov    eax,DWORD PTR [ebp+0x4]
   4:	c3                   	ret    
```

And there you go. Exactly what you wanted, without a naked function.

tari:
It seems we have two somewhat distinct use cases for naked functions, where I only had the first in mind while writing this RFC:
 * Papering over holes in the compiler's ABI support (allowing a user to handle the ABI themselves).
 * Implementing optimized or machine-specific functions.

Neither of these works well with the current proposal, because the ABI of function calls generated by the compiler is unspecified. For a compiler-generated call to such a function, we should require that the ABI in use be well-defined (specifically it must not use the Rust ABI, and we may want to restrict it to `extern "C"` only). This matches the semantics of naked functions in C, where the ABI of a given function is always known from source code and platform alone, and is usually (but not always) the default calling convention.

If the calling convention of a naked function which is called from Rust code is constrained to be well-defined, the problems are purely in implementation- `rustc` must learn how to use non-Rust calling conventions for functions it defines (rather than ones it merely links against).

---

@main-- that may be technically possible, but it's too fragile for me to be comfortable encouraging such a thing. We can make the static guarantees of such a function stronger by allowing it to be declared as naked without affecting the rest of the language, so I believe we should.

main--:
I agree that especially for things like interrupt handlers, naked functions are simply a necessity. Of course one could try and hack around that by calling into an embedded inline asm block, but that's far from perfect.

I do have doubts regarding ABI emulation (providing functions in a specific ABI without compiler support): Every function needs their own trampoline, right? So one would probably write a macro to create all those trampolines which ... surprisingly isn't even that messy, but still less pretty than it could be (`extern "fancyabi"`).

When calling a method, we need to know how. In case of a "standard" naked function, this gets us into a funny situation: Just like any normal function, we're supposed to use the Rust ABI which is fine for us - but at the same time we're **sure** that they (callee) can *never* handle this correctly due to that ABI being undefined. I'd suggest making those functions uncallable and forcing them to be parameterless.

On the other hand, when a naked function's ABI has a well-known definition (like `extern "C"`), all is well; no need to impose any further restrictions.

The other difficult part are local variables. Should you write "heavy" rust code within naked functions? Probably not. But it should definitely be possible within certain limits. Why else would you use Rust if you could just write your function in plain asm (much better tooling than inline asm)? Straight up banning locals is not a great solution, but since the compiler can't really reserve stack space either (naked --> no prolog) something like MSVC's [__LOCAL_SIZE](https://msdn.microsoft.com/en-us/library/6xy06s51.aspx) is probably inevitable. The *really ugly* part about that particular approach is that the compiler is forced to emit even a useless prologs and epilogs (when no stack spilling is required). So one can either sacrifice performance by having them *or* sacrifice non-optimized builds altogether since rustc (like many compilers) likes to just hold everything on the stack and then rely heavily on the optimizer to eliminate these useless stack operations, simply resulting in broken code without that.

aidancully:
Sorry for this, but... is this a problem that needs solving? I do a lot of systems programming, and do my best to avoid inline assembly where I can. Usually, what I'll do instead (especially for something like an interrupt handler), is write an assembly file and have it call out to a C-ABI routine to get a trampoline to more maintainable code. I don't yet see how `#[naked]` is technically superior to just writing assembly when you want to use assembly, and then interacting through an FFI?

eddyb:
@aidancully arbitrary codegen driven by Rust code, basically.
I don't know why this is not mentioned, but being able to output naked functions from macros and even monomorphization is quite useful:
```rust
static HANDLERS: [AtomicPtr<()>; 256] = [AtomicPtr::new(); 256];
#[naked]
unsafe fn handle_interrupt<F: Fn()>(number: u8) {
    (*(HANDLERS[number as usize].load(Relaxed) as *const F))();
    asm!("iret");
}
```

nikomatsakis:
Are there links to LLVM discussion on this topic?

tari:
I haven't found any meaningful LLVM discussion on it. Support was added in 2009, [r76198](http://llvm.org/viewvc/llvm-project?view=revision&revision=76198), which may be a good starting point for searching.

Tobba:
@eddyb That code is most certainly broken. If the compiler ever spills a register, you're fucked. Since it's also an interrupt handler, it should avoid clobbering registers, which is impossible to do efficiently in any way that will please LLVM.

In general the only way you're ever going to get the correct output is by having a single asm block within the function, which is just a hacky version of module-level assembly.

main--:
>If the compiler ever spills a register, you're fucked.

While the RFC currently doesn't address this issue it's not a problem with naked functions in general as solving it is possible (and probably necessary; see my previous comment).

aidancully:
@eddyb For what it's worth, I find that type of use plausible as a motivator, but I also worry that it might be *very* challenging to make `#[naked]` functions work in a generic fashion. And in that specific case, I don't think I'd write that function in a generic way: why should I need to monomorphize a function that just dereferences a function pointer and calls it? The code that will be generated during monomorphization will likely be identical, I don't *want* multiple copies! And, I want to make sure that the way I call the function will be valid, which can be hard to do without knowing the ABI of the called routine... But if I need to know the ABI, do I possibly *require* a prologue or epilogue for some types of function call? Maybe there's a solution that allows us to use the ABI as a parameter? Maybe something like this:

```rust
static HANDLERS: [AtomicPtr<()>; 256] = [AtomicPtr::new(); 256];
#[naked]
unsafe fn handle_interrupt<F: Fn()>(number: u8)
{
    // F::ABI's static functions are #[naked] #[inline].
    F::ABI::create_calling_env();
    (*(HANDLERS[number as usize].load(Relaxed) as *const F))();
    F::ABI::destroy_calling_env();
    asm!("iret");
}
```

But if this sort of thing is necessary to safely support generic naked functions, it'd probably need to be reflected in the RFC. (Note that I don't think the whole change would need to *land* at once, the initial patch sounds pretty small and may be acceptable in the short term. But the design space in which the feature would be used probably needs to be explored better... I think there's a significant risk that, even though `#[naked]` functions themselves are pretty simple, making them useful in a Rust context might get pretty complicated.)

eddyb:
@aidancully if you don't generate multiple monomorphizations, you need to use indirect calls, which might be a measurable cost during interrupt handling - but it is a micro-optimization and I could very well be wrong about it.

tari:
If I understand the current tack of the discussion now, the main concern is that the effect of writing code that is not inline assembly in a naked function is unpredictable. The correctness of the `handle_interrupt` example depends entirely on what codegen yields as currently proposed. Some users might be willing to accept the risk of doing so in the name of small gains, but doing so is fundamentally unsafe. This is the logic behind the current requirement that the body of a naked function contain only `unsafe` statements- inline assembly already requires `unsafe`, and the result of anything else cannot be guaranteed by the compiler.

A way to ask the compiler how much space it needs for locals would permit a solution, but there's no mechanism to enable that in LLVM. I see a way forward in having an LLVM intrinsic function that is guaranteed to expand to a constant load representing the necessary size of locals, but I think such a significant change is beyond the scope of this proposal.

With the `__LOCAL_SIZE`-like solution off the table, the only way to enforce (relative) safety is if only assembly is permitted in naked functions. This approach gives the compiler freedom to implement such a thing as either a module-level assembly block (with name mangling) or a true naked function, while permitting future extensions to relax the requirements (with a `__LOCAL_SIZE` workalike for example). On the downside, users willing to accept a certain amount of undefined behavior (trusting the compiler's behavior will be stable enough to make use of) can't choose to make that tradeoff anymore.

tari:
> On the downside, users willing to accept a certain amount of undefined behavior (trusting the compiler's behavior will be stable enough to make use of) can't choose to make that tradeoff anymore.

Both cases could be accommodated by putting support for Rust statements in naked functions behind a feature gate; something like `error: Rust statements in naked functions are extremely unsafe and may require additional guarantees in the future`. This permits the statically-safe version (assuming the programmer doesn't write bugs) with inline assembly only, while allowing adventurous users to be extremely unsafe until a better solution can be chosen.

nikomatsakis:
In a recent lang subteam meeting, we decided to close this PR. This RFC seems to have a lot in common with #55, which was closed by @brson with the following rationale:

> So far we've resisted exposing platform-specific linkage options, besides the limited ability to declare weak extern statics. If we were to pursue that I think I would want a more comprehensive proposal.
>
> Naked functions clearly have some use but there is a workaround of just writing it in assembly.

The situation seems not to have changed very much -- basically it seems a bit premature to be introducing these sorts of constructs. The concerns about running Rust code in naked functions seem particularly salient.

(Taking off my lang subteam hat for a moment and speaking personally, I think I would prefer to address this by writing naked functions in pure assembly and linking them in with LTO -- or having a variant on inline assembly that generates an entire function.)

tari:
Per discussion on #1275, reopening this one (doesn't look like I can do it since @nikomatsakis closed it?). I haven't had time to do any re-drafting, but anybody interested in doing so is welcome to have a go.

nikomatsakis:
Reopened.

jackpot51:
For reference: https://github.com/rust-lang/rust/pull/29189 is an implementation of naked functions

rasendubi:
I'd like to add that it's possible to use naked functions without assembly.

On ARM's Cortex-M architectures, the processor automatically stacks its state on exception entry and unstacks this state on exception exit, with no instruction overhead. It's ok for the interrupt handler to use any register, so it's possible to write interrupt handlers without assembly.

rasendubi:
The other use case for naked functions without assembly is a task function. In FreeRTOS, a task function is a function that runs in a separate thread with the new stack. It is not allowed to return, so it's essentially a diverging function.

I guess any diverging function may be safely marked as naked. I've checked, the compiler generates prologue for diverging functions now.

text/0000-naked-fns.md
```diff
@@ -0,0 +1,186 @@
...
+function should have prologue/epilogue emission disabled.
+
+Because the calling convention of a naked function is not guaranteed to match
+any calling convention the compiler is compatible with, calls to naked functions
+from within Rust code are forbidden unless the function is also declared with
+a well-defined ABI.
+
+The function `call_foo` in the following code block is an error because the
+default (Rust) ABI is unspecified and as such a programmer can never write code
+in `foo` which is compatible:
```
nikomatsakis: This will be hard to enforce given how the type system works. In particular, the type of `foo` is `extern "Rust" fn()`, and so writing `let x = foo;` would be legal, and then one could later write `x()`. So I think what would be better is to have a kind of "unknown" ABI that one can use for such cases:

```
extern "unknown" fn foo()
```

The type of `foo` is then `extern "unknown" fn()`, and we would just prohibit calling functions with that ABI.

nikomatsakis:
@tari so reviewing the discussion from the interrupt handler RFC, I think that to go forward with this RFC, we would want to edit it to include:

1. the constraint that `#[naked]` fns cannot have Rust ABI, since that is unspecified;
2. the constraint that the fn be either declared as unsafe or have an unsafe body;
3. we *could* add an "unknown" ABI, but we might be able to get by with just declaring functions with no known ABI as `unsafe`. This could also be added later.

I would probably add an "UNRESOLVED QUESTION" about firmly settling the `unsafe` rule.

arielb1:
Don't we need some way to prevent the optimizer from randomly spilling registers just because it feels like it? I mean, if a naked function consists only of inline assembly then there is no reason for it to do so, but optimizers tend to have their own reasons.

In other words: what is the exact semantics we want for naked functions? Having them be just a way to create a Rust static in `.text` that references assembly code would be one idea, but apparently some of the people here want to put Rust code within. Should it just be an error if LLVM can't compile the code without spilling registers?

main--:
@arielb1 According to @tari's comments on that topic above, the RFC as written would mean that users simply pray and hope that the compiler won't use any (more than they allocated) stack space - I believe.

It was [suggested](https://github.com/rust-lang/rfcs/pull/1201#issuecomment-122580133) that feature gating Rust code inside naked functions might improve the situation.

After re-reading [my old comment on that](https://github.com/rust-lang/rfcs/pull/1201#issuecomment-121413631) and considering the fact that there appears to be no solution (no suggestions except __LOCAL_SIZE which was determined infeasible), #1275 does look like the better solution after all, as it *can* deal with these problems much better.

arielb1:
@main-- 

I feel like just adding a "use rustc as an assembler" mode - forcing naked functions to consist of only an asm block - is the best option here: having rustc support the ABI for each interrupt calling convention is kind of nice, but people will always come up with new ABIs with their own needs. I feel that if you are afraid of the overhead of a single normal ABI call you should just write the relevant code in assembly anyway.

If we go with that option, an `"unknown"` ABI that renders functions inert would be fine, but allowing declaring "assembly" functions with an ordinary ABI would be nice too (of course, as the Rust ABI is unstable, what happens when a Rust ABI assembly function is called is unspecified, but I see no reason to prohibit trying).

main--:
>I feel that if you are afraid of the overhead of a single normal ABI call you should just write the relevant code in assembly anyway.

I agree. If one critically needs the compiler to output some instructions, one can just write that manually in the first place.

But if a naked function can't contain Rust code, then what's its purpose? Why would one prefer it over an actual assembler?

>what happens when a Rust ABI assembly function is called is unspecified, but I see no reason to prohibit trying

Thought: This can trigger UB. Therefore, it must be unsafe. But I'd say unsafe is not entirely

eddyb:
@main-- Generating many such functions using Rust macro could be an usecase.

main--:
>I feel that if you are afraid of the overhead of a single normal ABI call you should just write the relevant code in assembly anyway.

I agree. If one critically needs the compiler to output some instructions, one can just write that manually in the first place.

But if a naked function can't contain Rust code, then what's its purpose? Why would one prefer it over an actual fully-featured assembler?

>what happens when a Rust ABI assembly function is called is unspecified, but I see no reason to prohibit trying

Thoughts: This can trigger UB. Therefore, it must be unsafe. But I'd say that it's even more unsafe than Rust's standard notion of unsafety: While there are valid usecases for things like transmute, calling a naked Rust-ABI function is (according to my understanding) *always UB*. No matter what you do, it's wrong and may break whenever the compiler wants it to. From that perspective, I don't see why it should be allowed as it's unusable anyways. So are there any arguments in favor of banning it? Yes: It's a pitfall users can and should be protected from. While checking for the numerous conditions that may make a transmute valid is nontrivial, I'd expect forbidding a certain ABI to be much easier.

I see that a major drawback of architecture-specific interrupt ABIs is the sheer number (infinity) of variations people would want. I believe this would have to be configurable: Users could tell the compiler how to generate code for whatever ABI they need, perhaps through a compiler plugin.

Then again, due to LLVM integration problems I'm unaware of this approach may be infeasible as well.

arielb1:
@main-- 

> But if a naked function can't contain Rust code, then what's its purpose? Why would one prefer it over an actual fully-featured assembler?

If you use an assembler, you have to manage an additional build tool with its name-mangling scheme and build-system. That is quite annoying. We ideally want to build the full executable image with a single call to `rustc` with a linker script (+ maybe another pass to add the bootloader + image compression + maybe initrd + whatever).

> But I'd say that it's even more unsafe than Rust's standard notion of unsafety.

Not really. The Rust ABI is unspecified. This means that programs that use Rust ABI functions not generated via rustc are not strictly conforming.

However, Rust *does* have an ABI: it is currently allowed (and I think is contractually guaranteed to be allowed in the future) for (userspace) Rust programs to `dlopen` a crate compiled with the same version of rustc and call exported Rust ABI functions from it (as long as the types match). This implies that copying assembly code generated by an instantiation of rustc to a function will result in a perfectly-legal Rust program (whether that copying is done to a `#[naked]` function or by manually `mmap`-ing some rwx memory and loading the code). This at least contradicts such calls *always* being UB.

The Rust Team is not really willing to support interrupt handlers on odd architectures, and there are other uses for functions with odd ABI (e.g. context-switching trampolines, crazy JIT stuff). Trying to add that kind of support will only lead to endless rounds of complaints between the users and the Rust Team.

Ticki:
@arielb1 Also, you potentially wouldn't get all the optimizations by using external assembly. This is critical when doing osdev, since performance is very important.

arielb1:
@ticki

What kinds of optimizations are you talking about? LLVM is not supposed to look into `asm!` statements.

Ticki:
@arielb1 LLVM _is_ doing optimizations on assembly, it's just much more limited. But the main performance disadvantage of external assembly is the very limited inlining.

main--:
@arielb1 I now understand your reasoning. Thank you for the clarification.

lhecker:
Another use-case for naked functions: Coroutines!
[context-rs](https://github.com/zonyitoo/context-rs) is currently using `gcc` to compile the assembly seperately. If Rust had naked functions the code could be moved into a `asm!` block, e.g. without getting the frame pointer `%rbp` pushed onto the stack first. Thus users of `context-rs` wouldn't be required to have a seperate compiler installed anymore.

Ticki:
Yeah. @zonyitoo had a lot of problems due to the lack of naked functions.

mahkoh:
```rust
pub unsafe fn isr_3_wrapper() {
    asm!(
        "
        .globl isr_3
        isr_3:
            pusha
            call increment_breakpoint_count
            popa
            iret
        " :::: "volatile");
}
```

mahkoh:
>If you use an assembler, you have to manage an additional build tool with its name-mangling scheme and build-system. That is quite annoying. We ideally want to build the full executable image with a single call to rustc with a linker script (+ maybe another pass to add the bootloader + image compression + maybe initrd + whatever).

Who's going to write thousands of lines of assembly in inline assembly? Nobody. But if you really want to do it then you can already do so without a `#[naked]` attribute as the code above shows.

rasendubi:
The feature is pretty useful without inline assembly. I've described a couple of use cases for `#[naked]` that don't involve assembly at all, but I'm afraid they're lost in the disscussion: https://github.com/rust-lang/rfcs/pull/1201#issuecomment-161519313

mahkoh:
@rasendubi Why can't you use a normal function for this? Maybe `extern`.

rasendubi:
That's what I'm doing now. I'm not saying I can't live without the feature but rather it would be really nice to have it (save little stack and CPU cycles). I feel pity ARM designed the architecture that automatically stores the context for interrupts and it's all ruined by function prologues.

Ticki:
@mahkoh I don't think you get what's the issue here: First of all, when writing 20 lines of assembly in rust code, inline assembly is much easier and more clean to work with. Naked functions is a feature missing, you could as well say: "X is not necessary to have in Rust, since you can just use C to do it." Furthermore, you do not get all the opltimizations possible by using external assembly. This is important, because when doing OS development (like Redox), performance matters.

jackpot51:
@mahkoh We need naked functions so we can write interrupt handlers in Rust, not in assembler.

mahkoh:
>First of all, when writing 20 lines of assembly in rust code, inline assembly is much easier and more clean to work with.

I disagree. There are good reasons to use inline assembly over freestanding assembly files if you need inlining or similar. E.g. if you want to use atomic operations, disable or enable interrupts, then those are good cases where one can use inline assembly. On the other hand, longer functions are much better written in freestanding assembly files which have proper syntax highlighting. E.g., [this](https://github.com/lrs-lang/lib/blob/e7292da1b7b964e62036b7e8977940c6c655ad97/asm/aarch64-unknown-linux-gnu/src/stop_thread.s) is code to terminate a thread on aarch64. This could easily be written in inline assembly but since this function does not profit from inlining, it is much better to write it in a separate file.

>Furthermore, you do not get all the opltimizations possible by using external assembly. 

Which optimizations are you referring to besides inlining? Are there any other operating systems that use inline assembly over freestanding assembly files to get significant performance improvements. I think linux doesn't do this and since linux is already using lots of compiler extensions, using naked functions seems like something they would do if it were to give them a significant performance boost.

>We need naked functions so we can write interrupt handlers in Rust, not in assembler.

Doesn't the code I posted above already solve this problem without the need for naked functions? Or do you want to use rust code in naked functions? It seems that at this moment nobody has explained how this would even work.

It seems that @nikomatsakis has already suggested a much better solution to the performance problems: LTO with assembly files. Though I don't know if this is even possible with llvm right now and inlining of rust code into assembly functions seems to have the same problems as using rust code in naked functions.

What seems to be much more interesting is allowing assembly files to be compiled with rustc which is another justification for this RFC. Since @arielb1 already suggesting allowing rustc to be used with a linker file, having multiple arguments seems to be inevitable. E.g.,

```
rustc .... main.rs start.s entry.s --linker-file kernel.ld
```

It's not super clear how this would really work since assembly files often use assembler-specific features. If the rustc assembler isn't GAS compatible, then it's not clear how useful it really is. But the same applies to inline assembly. The LLVM inline assembler is supposedly trying to be GAS compatible, but I'm not sure I'd trust it with the following code:

```asm
	.macro vector dst
	.balign 0x80
	b	\dst
	.endm

	.section ".text"
	.balign	0x1000
	.globl vectors
vectors:
	vector e1
	vector e2
	vector e3
	vector e4
	vector e5
	vector e6
	vector e7
	vector e8
	vector e9
	vector e10
	vector e11
	vector e12
	vector e13
	vector e14
	vector e15
	vector e16
```

(The aarch64 exception vector.)

To wrap this up:

1. Naked functions alone do not make it a reasonable possibility to only use rustc when building a kernel.
2. The one feature that everyone agrees would be possible with naked functions (a single asm block) is already possible today without naked functions.
3. It hasn't been shown how using anything but a single asm block would work.

mahkoh:
Also note that sufficiently complicated assembly functions often make use of CPP when compiled with GAS to avoid the use of magic constants. How would this work with inline assembly?

mahkoh:
```rust
rustc .... main.rs start.s entry.s --linker-file kernel.ld
```

Well, this is not really necessary. Compiling/assembling one file at a time and then linking them all in the end would be better if rustc exposed a way to invoke the linker directly.

farcaller:
@mahkoh yes, I plan to write interrupt handlers in pure rust and I am one of those guys who have their MCUs to manage the stack for them.

mahkoh:
>What seems to be much more interesting is allowing assembly files to be compiled with rustc which is another justification for this RFC. 

To add to this. @arielb1 and @eddyb have said that this would allow kernels to be written without the need for another toolchain (another assembler, linker.) This is not quite true because at the very least you probably want a linker file. But even if rustc accepts a linker file, you still have to write lots and lots of assembly and I don't think anybody wants to write 1000s of lines of assembly in inline assembly. It's already possible to write lots of this in inline assembly using the technique I showed above. If nobody is doing this now, nobody will do it if naked functions are added.

mahkoh:
@farcaller Maybe I missed the post, but I think the latest was that it wasn't clear if a naked function could contain anything but a single asm block. Some have said that other compilers support this, but from my 1 minute research it appears that GCC clearly says that naked function can only contain inline assembly. GCC has other attributes that allow the saving/restoring of registers to be omitted which appears to be what you want. But those attributes are platform specific and different from naked. See e.g. [here](https://gcc.gnu.org/onlinedocs/gcc/AVR-Function-Attributes.html).

farcaller:
@mahkoh well, what I'm looking for is literally a function with no prologue/epilogue generated by compiler.

arcnmx:
As a random data point, this is a macro from one of my projects:

```rust
asm_fn! {
	pendsv_imp: unsafe fn pendsv("i"(sched::sched)) {
		"push {r4 - r11, lr}
		mov r0, sp
		b $0"
	}
}
```

I've encountered two main drawbacks to it...

1. The need to specify an ident for the dummy function because we don't have either `concat_idents!()` or module-level asm blocks. The dummy wrapper function is required, and ugly.
2. No ability to use `include()`, which would allow for external files and syntax highlighting.

> Naked functions alone do not make it a reasonable possibility to only use rustc when building a kernel.

I've found it to be easy enough to avoid an external assembler and only use inline asm with the method mentioned above. Naked functions make it nicer and possible to define inline asm functions in a "rusty" way that feels like it fits into the language. Avoiding a linker is more tricky, but rustc already has the ability to specify linker args. My own arm thing just uses `cargo rustc --bin/--example` to build the entire kernel.

mahkoh:
Can we see some of that code? I've written a little aarch64 kernel to test lrs's new freestanding mode and the minimal work I have to do to call into rust code (set up various system registers, create page tables, etc.) is well over 200 lines of assembly (including comments.) Would you actually suggest converting this into inline assembly for the sole purpose of avoiding an assembler? And why do you want to avoid an assembler? I'm also using various GAS features to make the code clearer. Is this possible in inline assembly?

```asm
// register aliases
	kernel_map		.req	x24

// ldr, =
	ldr	kernel_map, =_kernel_map
```

mahkoh:
To get some perspective: The arch/arm64 in the linux tree contains over 9000 lines of assembly code in assembly files. Most of them use CPP to include header files which contain constants and macros used in the assembly. Converting those 9000 lines into rust inline assembly would force one to manually expand all macros and to replace constants by magic numbers.

arcnmx:
@mahkoh My goal was to use as little asm as possible, so everything that can be is written in Rust.

For example, my interrupt table looks like so:

```rust
#[no_mangle]
#[linkage = "external"]
#[link_section = ".vector_table"]
pub static VECTOR_TABLE: VectorTable<()> = VectorTable {
	exception_handlers: ExceptionVectors {
		svcall: IrqHandler::new(svcall as *mut _),
		systick: IrqHandler::new(systick as *mut _),
		pendsv: IrqHandler::new(pendsv as *mut _),
		.. ConstDefault::DEFAULT
	},
	interrupt_handlers: (),
};

```

The entry point is pure rust, most interrupt handlers are tiny asm stubs that call into rust functions (as in my last comment), cpu registers are all set via structs and volatile types, etc.

mahkoh:
@arcnmx By entry point do you mean the code followed by the exception vector (I'm not sure if pre-v8 arm uses this terminology) or do you mean the code called after a reset? I don't think the code called after a reset can be written in rust directly since you shouldn't even have a stack. How does the reset part look in your code?

In your example you put `VECTOR_TABLE` into its own section. I assume this is so that you can properly align it. At least on armv8 it has to be aligned to a page table. I've also considered this but I'm unsure about the guarantees made by the linker. If a section contains only symbol, is this symbol guaranteed to be located at the start of the section? In the assembly version this is no problem because I can align the vectors directly with various directives so that I don't have to rely on the linker doing what I want.

arcnmx:
@mahkoh 

Yup, you can make all the assertions you want about how it places data in a linker script. It'll align everything as you need as long as you specify it all properly (and make sure your inner member alignments are all correct too, of course)! A single symbol in a section will indeed start at its beginning and only fill as much as it needs.

CPU registers can be modelled as `&*(0x12345678 as *const SomeRegisterStruct)` as long as `SomeRegisterStruct` is `repr(C)` and contains members like `VolatileCell<usize>`. Make some more abstractions for bit banding... Most constants and setup like that can just be moved over to rust, really.

As for entry point, most exception vectors work with pure rust in armv7m. Exceptions set up the stack before executing the handler, and the vector table has a separate entry for the stack value when the cpu is reset. So that all just works! And in the cases (or architectures) where it doesn't, well, that's what tiny stubs and macros are for!

mahkoh:
@arcnmx I'll try to port my startup code to mostly rust and report back.

rasendubi:
@mahkoh My kernel starts without a single line of assembly: https://github.com/rasendubi/bkernel.
Look at https://github.com/rasendubi/bkernel/blob/master/stm32f4/isr_vector.rs and .isr_vector section (https://github.com/rasendubi/bkernel/blob/master/stm32_flash.ld). That's all code needed for startup.

nikomatsakis:
Per our recent comments, I'd like to see some motion on this RFC. I think we're going to need do more experimentation to reach firm conclusions on the best design and overall utility of naked fns.

What I think should happen is that we land the RFC and then @jackpot51's implementation (once it is brought in sync with the RFC), all under a feature gate and with the general understanding that this is all provisional.

When we reopened the RFC, we had in mind two edits:

- [ ] the constraint that #[naked] fns cannot have Rust ABI, since that is unspecified;
- [ ] the constraint that the fn be either declared as unsafe or have an unsafe body;

and also some unresolved questions:

- what restrictions should be placed on the body (if any)
- should we adjust the rule requiring the use of `unsafe` to be more flexible?
- do we want an "unknown" ABI, or can we just specify an incorrect ABI (like "C") and mark the function as `unsafe` (I'm inclined for that, personally)

@tari if you don't have time to edit the RFC, perhaps @jackpot51 or someone else would like to adopt it?

arielb1:
@nikomatsakis 

The view I got from this is that the only restriction on naked fns is that they are supposed to consist of just an `asm` block.

nikomatsakis:
On Fri, Jan 08, 2016 at 01:10:19PM -0800, arielb1 wrote:
> @nikomatsakis 
> 
> The view I got from this is that the only restriction on naked fns
> is that they are supposed to consist of just an `asm` block.

I would be ok with such a restriction as well. I can certainly see the
justification (it's hard to tell if the code that the rust compiler
would generate makes any sense absent the usual prologue).

rasendubi:
I want:
1) To write naked functions without assembly. I plan my naked functions to be pure Rust.
2) Safe bodies for naked functions. I want compiler check I'm not doing unsafe things. I fully support calling a naked function in unsafe, though.

rasendubi:
Ok, I've read more docs on both llvm and clang. Now I see what I really want is clang's `interrupt` attribute. However, llvm itself doesn't have it, but it has `naked` which is doing almost the same thing on ARM-M architecture.

Now I prefer to have `interrupt` attribute (or calling convention) because it adds more safety and allows everything I need. But I see interrupt calling convention proposal (https://github.com/rust-lang/rfcs/pull/1275) was closed in favor of this one. I'm kind of ok with that. I'll be satisfied with either naked functions with non-asm (possibly safe) bodies or proper interrupt calling convention. The later much better meets my needs, though.

eddyb:
> @tari commented 23 minutes ago (in #1275)
> Datum: LLVM support for x86 interrupts landed in [r256155](http://reviews.llvm.org/rL256155), and should be present in releases beginning with 3.8.

AFAICT we could support that calling convention by restricting the signature to `fn(*mut T)` and `fn(*mut T, usize)`.

Ericson2314:
Yeah, if upstream went for it, its not much a burden on us to support it. I'd also like to see unstable compilers allow for any llvm-supported calling convention, to allow messing around with things as they appear upstream

@eddyb I rather just let LLVM do the error-handling for that than duplicate it. Firstly, one may want to use there own types of the same size---e.g. bitflags and nicer reference types. Secondly, as far as safety is concerned, installing an interrupt handler is still unsafe, so were LLVM to erroneously allow interrupt calling conventions with absurd signatures, we wouldn't break any of our invariants.

Ticki:
@mahkoh 

ITT "we don't need this, because you can do what you want in assembly." That's right, but you can do everything, you want to, in brainfuck as well. Rust is a systems programming language. It should be able to handle these kind of things without hassle. Using "use [insert language here] instead" as an argument is not valid.

mahkoh:
>ITT "we don't need this, because you can do what you want in assembly." 

How fitting since a naked function only allows a single block of assembly as its body.

eddyb:
@Ericson2314 Code that passes Rust checks should not end up in LLVM asserts or other kinds of *post-monomorphization* error - although I'm not sure if we wrote that down somewhere official.

jackpot51:
@nikomatsakis What is the status of this?

There is still a missing piece of Redox, the interrupt handler, that must be done in assembly.
https://www.reddit.com/r/rust/comments/4656db/this_week_in_redox_11_x86_64_support_and_much_more/d046at4

nikomatsakis:
@jackpot51 I've been waiting on the edits described here:

https://github.com/rust-lang/rfcs/pull/1201#issuecomment-170111445

jackpot51:
@nikomatsakis Unfortunately, I do not know how to get rustc to require an unsafe signature. I believe the requirement for an extern ABI is not required, as naked modifies how a function is compiled internally, not how it is called.

eddyb:
In the current implementation, non-Rust ABIs wouldn't even work, as they *still* introduce a shim. Hopefully that won't be true for long.

tari:
> I've been waiting on the edits described here:
>
> #1201 (comment)

I thought I addressed all of those in https://github.com/tari/rfcs/commit/1cc857fb4239990e0620a11ae504ece7f410fcaa. Is anything missing?

jackpot51:
@tari I don't think anything is missing

nikomatsakis:
On Tue, Feb 23, 2016 at 12:54:07PM -0800, Peter Marheine wrote:
> > I've been waiting on the edits described here:
> >
> > #1201 (comment)
> 
> I thought I addressed all of those in https://github.com/tari/rfcs/commit/1cc857fb4239990e0620a11ae504ece7f410fcaa. Is anything missing?

Well I'll be. I didn't notice that commit I guess. My apologies!
(Also for tardiness of this reply; I'm working off an e-mail backlog
of notifications.)

nikomatsakis:
Nominating for discussion at lang meeting.

eddyb:
After rust-lang/lang#32080 is merged, adding custom ABIs that interact well with Rust code will be relatively easy, just need to have `abi::FnType` configured correctly.

We can also do interesting things like have simple MIR passes for denying *anything* that would create variables on the stack, and then make `AllocaFcx` a hard error.
That will let you have `asm!` statements without fearing the codegen's whims.

nagisa:
I had the opportunity to use the naked functions in clang to great extent lately. In my eyes the restriction to only have a single `asm!` block was not restricting at all, but there were a few other things that were nagging me. 

Namely, I disliked a lot the fact that you can specify arguments, but not use them as inputs to inline assembly block (with some restrictions). I would have liked for code like this to work:

```c
__attribute__((naked, ms_abi)) uint32_t
nakedfn(uint32_t c /* uses ecx */)
{
    __asm__("..." : /* a way to get return value register (rax for ms_abi)? :( */ 
                  : "r"(c) /* (ab)uses the same register */);
}
```

All these could be handled implicitly by some sort of restricted `asm!` macro which doesn’t take any sections, other than the assembly section, explicitly and generates inputs+outputs from the function signature instead. E.g.

```rust
#[naked]
extern "win64" fn nakedfn(a: u32) -> u32 {
     asm!("mov %[return], %[a]"); // cannot explicitly provide any outputs or inputs
     // turns into mov %ecx, %eax; 
     // for ABIs where argument cannot be passed in without generating extra code, this would fail.
}
```

eddyb:
@nagisa We can allow arguments and enforce that they are immediates.
Also maybe we should model LLVM's `inlineasm` better?
It actually *returns* the results instead of pointers to lvalues, so you could imagine a form of `asm!` that produces a value.

nagisa:
> We can allow arguments and enforce that they are immediates.

There’s no necessity to do that. All what *really* matters that caller will put argument somewhere, and rust would automatically generate the constraint and input depending on the ABI. It should work for both arguments passed through registers and arguments passed on stack (memory constraint, resolving to `[rsp - whatever]`).

eddyb:
@nagisa But I'd rather have a principled solution that a hack...

nikomatsakis:
**Hear ye, hear ye.** This RFC is hereby promoted to **final comment period**. 

This issue has been around a long time and many of the arguments for and against are not fresh on everyone's mind, I suspect. So if you harbor strong feelings -- or prefer an alternate design -- please do try and make your case, or at least leave a link to an earlier comment that summarizes your position!

jackpot51:
This is the only change that I know of that would allow critical functions to be written in pure Rust, without requiring (but still allowing) external assembly *or* inline assembly.

It is a small change (20 lines of code) that will make Rust competitive with C for low-level development.

Many OS projects written in Rust have struggled with this aspect of the language, and merging the PR mentioned above would alleviate problems that they have in context switching and interrupt handling.

The other proposals (custom ABI, module level assembler) would not be as flexible, and would not give us what we really want, which is the ability to write functions that need to be trusted in a language that can be trusted.

rasendubi:
I would like to have non-asm naked functions. They are useful for writing interrupt handlers in embedded/kernel-level programming (at least on ARM architecture). See comment https://github.com/rust-lang/rfcs/pull/1201#issuecomment-161519313 for use case.

I also found that my need is better covered with interrupt calling convention (https://github.com/rust-lang/rfcs/pull/1201#issuecomment-170169360), but the appropriate RFC (https://github.com/rust-lang/rfcs/pull/1275) was closed in favor of this one.

tari:
> I also found that my need is better covered with interrupt calling convention (#1201 (comment)), but the appropriate RFC (#1275) was closed in favor of this one.

Given the concerns presented on #1275, I think support for arbitrary calling conventions as provided by LLVM is a more generally useful feature. The suggested `"unknown"` calling convention for naked functions could be supplemented by arbitrary LLVM specifications (`"llvm_cc(n)"` perhaps). Where the backend supports a calling convention it could be used, otherwise a naked function can provide the necessary functionality.

---

I am in favor of permitting `asm!`-only naked functions now, with the possibility of future extension to permit non-asm functions with compiler checking or arbitrary backend calling conventions. Non-asm naked functions could be allowed in an unstable fashion with per-function opt-in in the meantime.

eddyb:
@rasendubi I really think we should have both - `naked` only for situations where a dedicated calling convention (i.e. interrupts) doesn't exist.

Ericson2314:
https://github.com/rust-lang/rfcs/pull/1201#issuecomment-194918762 makes me happy that this can evolve into something more interesting. And I whole-hardheartedly agree with everything that @tari said about allowing arbitrary LLVM calling conventions---probably best to do that as a permanently-unstable escape hatch.

text/0000-naked-fns.md
```diff
@@ -0,0 +1,218 @@
...
+```rust
+#[naked]
+extern "C" fn foo() { }
+```
+
+---
+
+Because the compiler cannot verify the correctness of code written in a naked
+function (since it may have an unknown calling convention), naked functions must
+be declared `unsafe` or contain no non-`unsafe` statements in the body. The
```
nrc: I feel the second clause here is unnecessarily complex, just require that the function must be unsafe.

text/0000-naked-fns.md
```diff
@@ -0,0 +1,218 @@
...
+The following example illustrates the possible use of a naked function for
+implementation of an interrupt service routine on 32-bit x86.
+
+```rust
+use std::intrinsics;
+use std::sync::atomic::{self, AtomicUsize, Ordering};
+
+#[naked]
+#[cfg(target_arch="x86")]
+unsafe fn isr_3() {
```
nrc: you state above that this declaration would be an error since it uses the Rust ABI. Should there be an `extern ...` here, or did you mean (above) that calling is illegal, ubt declaration is OK?

text/0000-naked-fns.md
nikomatsakis: On Sun, Mar 13, 2016 at 04:42:35PM -0700, Nick Cameron wrote:
> I feel the second clause here is unnecessarily complex, just require that the function must be unsafe.

While I initially thought as you do, I found [this argument] compelling:

> The main issue I have with this is that doesn't allow you to assert
> that you're maintaining the ABI/invariants/etc. and provide a safe
> implementation.

[this argument]: https://github.com/rust-lang/rfcs/pull/1275#issuecomment-150670118

nikomatsakis:
@jackpot51 

> This is the only change that I know of that would allow critical functions like interrupt handlers to be written in pure Rust, without requiring (but still allowing) external assembly and inline assembly.

So I have seen several assertions that you can use Rust code in naked functions and I admit that this part still confuses me. It's by far the part of this RFC that bothers me the most. I'd feel a lot more comfortable if someone could point me at existing LLVM or clang documentation about how to use "naked" with anything other than inline assembly (afaik, no such documentation exists, because LLVM doesn't really expect you to do anything other than inline assembly).

Since the Redox project is apparently making use of actual Rust code (and not just inline assembly), can you maybe give some examples of things that work? (I think you have done so in the past). I'm wondering if there are general patterns that one could extract. For example, maybe it is reasonable to access statics, but not local variables or parameters (though that is far from clear to me; I imagine some architectures may require some kind of stack frame or temporary to access a static).

All in all, it seems to me that permitting Rust code inside the body of a "naked function" is still a rather dubious concept. After all, compiling general Rust code requires a general "stack frame" to be established -- we may need to e.g. spill a temporary into a stack slot, reference a local variable, reference a parameter (if there are parameters), and so forth. I understand that your patch permits Rust code -- and this RFC leaves it as an unresolved question whether to forbid it -- but it seems to be assuming that the compiler will do **something** in those cases, and I feel like, whatever that **something** is, it seems to be a bit... undefined by its very nature.

For example, if you use Rust code inside the function, whose job is it to allocate the stack space for the current function (that is, adjust the stack pointer appropriately)? Normally, it is the compiler's job, since only the compiler knows how many stack slots it requires (that will, after all, depend on how effectively the compiler optimizes, and so forth).

In practice, the "semantics" here are just plain inherited from LLVM, and maybe that's good enough. I imagine those semantics, in turn, are kind of "inherited" from gcc -- that is, this works "well enough" in practice without anyone having really tried to define very precisely what will happen (at least, the last time I searched, I couldn't find anything).

text/0000-naked-fns.md
tari: Yes, that should be `extern`; this example missed being updated. Ideally it would be `extern "unknown"` but if that were to not exist it would need to be declared with a known-wrong calling convention.

text/0000-naked-fns.md
Ericson2314: An idea I've been mulling for this and other related situation is `safe { }` blocks.

Ericson2314:
I do agree we should always be wary of blindly inheriting semantics from LLVM/GCC/C/Wherever. But I think we can have are cake and eat it too by adding such easy-to-implement-but-hard-to-define features like this as not up for stabilization.

Ericson2314:
And as an aside, I worry that since 1.0 the RFC process has been too geared towards features we basically envision stabilizing at some point. I'd hope it could still be used for both temporary fixes (as this can be construed), and more experimental features that's hard to pass judgement on without trying out for a while.

Obviously the core team doesn't want to be stuck maintaining others' hacks and experiments, but perhaps some means to delegate maintainer responsibility for these things, with the penalty of having them chucked from the tree if the delegates don't deliver, would take care of that.

vitiral:
I've only worked with interrupts in microcontrollers, but why can't you have the interrupt just use the existing stack? It is the nature of interrupts that they have to return before  anything else can run. (this is how it works on a microcontroller)

Amanieu:
I've added an RFC (#1548) for module-level inline assembly. I think this feature can replace the need for naked functions in many cases.

thepowersgang:
@Amanieu - That sounds like a much better idea.

I've been pretty hesitant on naked functions, because they're either unknown magic, or exceptionally limited.

All you get over module-level inline assembly is free naming and symbol size (assuming the pure-bare version of naked functions).

arcnmx:
> All you get over module-level inline assembly is free naming and symbol size (assuming the pure-bare version of naked functions).

As I touched on in https://github.com/rust-lang/rfcs/pull/1201#issuecomment-168777234, I personally feel like this is actually the main advantage of the RFC. You should be able to simulate both module-level asm and asm-only naked functions with a macro, so mostly I see this as an ergonomics discussion. Using rust code in naked functions is something I'm not really missing and don't have as much of an opinion on, but maybe following the llvm/clang/gcc semantics is good enough...

I'd also support module-level inline assembly, as it's a rather useful feature, though if you can't use symbols/constants/etc. that's a pretty unfortunate limitation... Both module-level asm and naked functions are just plain things that should exist in Rust IMO. All too often the lower level aspects get neglected despite being branded as a systems language...

Ketsuban:
I have to admit the comments on this RFC have kind of lost me. Perhaps this just needs the ELI5 treatment.

I started off following the [interrupt calling convention RFC](https://github.com/rust-lang/rfcs/pull/1275) and followed the pointer here; I got the impression there were other unrelated problems and naked functions are a more general solution to both domains.

Now I'm here, this RFC is in FCP and I still don't really understand why naked functions which can only contain a single block of inline assembly are useful. That's enough to call a proper function, sure, but at that point I might as well just make a small assembly file—I already have a larger assembly file which does the necessary prep work to start running Rust code in the first place, so it's not like I can avoid invoking an assembler at all.

By comparison with the Visual C++ example on the OSdev wiki (the only listed compiler which uses naked functions rather than a special function annotation to indicate an interrupt handler) this is what I see a Rust interrupt handler built out of a naked function looking like.

```rust
naked fn interrupt_handler() {
    asm!("push rax");
    asm!("push rcx");
    // push everything ...
    // Rust code to service the interrupt goes here
    // pop everything ...
    asm!("pop rcx");
    asm!("pop rax");
    asm!("iret");
}
```

Am I mistaken?

arcnmx:
> I already have a larger assembly file which does the necessary prep work to start running Rust code in the first place, so it's not like I can avoid invoking an assembler at all.

Note that this isn't really true in all cases, so there's value in being able to avoid bringing in an external assembler entirely! I have an armv7m kernel that only uses asm for interrupt handler trampolines, all inline, and the entry point is pure rust, for example.

Ticki:
> Perhaps this just needs the ELI5 treatment.

The proposed *#[naked]* attribute indicates that the specified function does not need prologue/epilogue sequences generated in the compilation process. It is up to the programmer to provide these sequences.

In basic terms, this makes rustc capable of doing all the low-level stuff which C is able to. Such as:

- Possibility of implementations of green threading.
- Interrupt handling.
- Context switching.

*All in Rust, without needing external files.*

---

This is especially important when dealing with low-level stuff, such as OS development.

We for example need this in Redox (an OS written in pure Rust) to get x86 support working. So this is really an important feature to us.

------

An example for the curious would be a custom interrupt handler function:


    #[cfg(target_arch = "x86")]
    fn interrupt_handler() {
        asm!("iretd");
    }



This function will currently generate code as:


    interrupt_handler:
      push ebp
      mov ebp, esp
      iretd
      ret


(truncated for simplicity)

With the #[naked] attribute (which is implemented with PR #29189), it will bypass the prologue and epilogue sequences. And the output will be like:

    interrupt_handler:
      iretd

Namely, it is a way of bypassing the automatically generated sequences, which pop/push to the stack.

naicode:
Given that rust is meant to be a system level programming language, I think it is needed to have this feature (or a similar on) in rust. 

Through it is needed for low level tasks for many people given that this feature is not  needed or even bad style in less low level code I would propose that (if available in stable) there is per default lint like `#[deny(naket_fn)]` or similar. 

Also even through module level asm seems to* be able to emulate naked fn's I think naked fn's are much clearer for use cases like interrupts.

(* I might have missed something, but naked fn's can contains (some) "normal" rust code, but module level asm can't, or?)

eddyb:
@naicode Naked functions have to be assembly-only.
However, platform-specific interrupt ABIs would be much more useful for interrupt handlers, as you could have Rust code inside them and not care about, e.g. needing `iret` to return.

naicode:
Ok, so module level asm and naked functions most likely can emulate each other.

Hm, I't might be able to implement platform-specific interrupt handling by combining macros with naked functions, leaving naked functions still there for the other use cases.
(through not implementing them as simple rust macros might enable a higher degree of validity checking, my interrupt related knowledge is a bit limited, so I can't judge if it is possible)

eddyb:
@naicode You would want to implement the whole Rust language in macros?

naicode:
@eddyb  No! that would be horribly ;=)

What I meant, is that, if the naked fn's can only contain asm, you would either write the complete ISR in assembly or delegate to a rust-fn at some point. This boiler part could be be written as a macro...

e.g. the example from the rfc:
```rust
#[naked]
#[cfg(target_arch="x86")]
unsafe fn isr_3() {
    asm!("pushad
          call increment_breakpoint_count
          popad
          iretd" :::: "volatile");
    intrinsics::unreachable();
}
```

could be written like:
```rust
delegate_isr_3!(increment_breakpoint_count)
```

Through in the end having a rust native ISR ABI might be more performant/useful

Also you could pass a codeblock to a macro, but it probably wouldn't be a good idea to use it without wrapping it in a function because it might contain a return statement (and possible other "bad" thinks), but you could also have a way to forbid return statements in code block passed to macro, well that is a different topic.

nikomatsakis:
**Huzzah!** The @rust-lang/lang team [has decided][] to **accept** this RFC. As previously stated, this is being accepted on a particularly experimental basis, especially with respect to the semantics (or lack thereof) of Rust code embedded in a naked function. One plausible route for stability would be to stabilize the `#[naked]` attribute when the function contains only assembler, or meets other similar criteria -- but without having some experience with the attribute, we'll never know what those criteria ought to be. 

[has decided]: https://docs.google.com/spreadsheets/d/18FtPnEI8RTPdFhGDtj21ITndakEtxVrs8pVzRNx-jL8/edit#gid=0

nikomatsakis:
Tacking issue rust-lang/rust#32408.

Ticki:
Thanks, thanks, thanks, @rust-lang/lang.

Ticki:
PR submitted: https://github.com/rust-lang/rust/pull/32410

PR #1209, opened 2015-07-13
Update FOLLOW set for `ty` tokens

alexcrichton:
Accounts for rust-lang/rust#27000

cc @cmr

### DIFF
--- a/text/0550-macro-future-proofing.md
+++ b/text/0550-macro-future-proofing.md
@@ -106,7 +106,7 @@ The current legal fragment specifiers are: `item`, `block`, `stmt`, `pat`,
 
 - `FOLLOW(pat)` = `{FatArrow, Comma, Eq}`
 - `FOLLOW(expr)` = `{FatArrow, Comma, Semicolon}`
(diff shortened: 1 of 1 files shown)


cmr:
r+

nikomatsakis:
cc @pnkfelix

nikomatsakis:
Hard to argue against this!

nikomatsakis:
To be clear, did code for this already land?

alexcrichton:
@nikomatsakis yeah it landed in https://github.com/rust-lang/rust/pull/27000 (although we can revert if we don't want to accept this RFC)

dgrunwald:
Any reason we don't allow `{` after `ty`?
In particular, I'd like to write a macro that mimics function definition syntax:
`($paramname:id) -> $rettype:ty $body:block`

nikomatsakis:
@dgrunwald yes, seems like there isn't a good one. also `where` keyword.

huonw:
cc #733

nikomatsakis:
Hear ye, hear ye. This RFC is entering **final comment period**.

/me thought it already was

PR #1235, opened 2015-08-03
More control over `BufReader`

cybergeek94:
[Rendered.](https://github.com/cybergeek94/rfcs/blob/master/text/0000-bufreader-methods.md)


text/0000-fill-buf-min.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Detailed design
+
+The solution the author proposes is to add a new method to the `BufRead` trait, `fill_buf_min()`. The proposed signature is as follows:
+
+```rust
+pub trait BufRead: Read {
+    // Existing methods omitted for brevity.
+    
+    fn fill_buf_min(min: usize) -> io::Result<&[u8]>;
```
seanmonstar: Does this return everything in the buffer (so `&self.buf[..self.len]`) or just the amount that was read in to fill up to the minimum?

cybergeek94: Yeah, it returns the whole buffer, just like `fill_buf()`. The only difference is that if the buffer is less than `min` bytes long, it should perform another read before returning.

text/0000-fill-buf-min.md
```diff
@@ -0,0 +1,81 @@
...
+}
+```
+
+The semantics are similar to the existing `fill_buf()` method, with one exception: if the length of the buffer is less than `min`,
+the implementation should perform another bulk read from the source. 
+
+However, the implementation is not *required* to meet this minimum,
+because it may not be able to for some reason, e.g. because the stream is at EOF or because it would take several expensive reads to do so. Thus,
```
oli-obk: Doesn't the statement about expensive reads make this function totally useless? I see the point about EOF being signalled by not filling the buffer at least to min, but that should be the only reason filling to min should fail.

text/0000-fill-buf-min.md
```diff
@@ -0,0 +1,81 @@
...
+* Implement `fill_buf_min()` as an inherent method on `BufReader`. This avoids breaking downstream implementors altogether, and also doesn't
+require addressing the buffering semantics of other `BufRead` implementations. If it turns out this behavior is desired across all `BufRead` implementations,
+the method can be moved to the trait relatively seamlessly, since almost all uses of `BufReader` are for its `BufRead` impl<sup>[citation needed]</sup>. The transition can be eased
+by employing the previous alternative as well.
+
+    * A preliminary implementation of this approach is available [here](https://github.com/cybergeek94/multipart/blob/boundary_fix/src/server/buf_read.rs#L34).
+
+# Unresolved questions
+
+* Is `fill_buf_min()` a descriptive enough method name?
```
oli-obk: not for the way you are suggesting it should work. The currently described behavior is "do a read and append to the buffer". It doesn't actually guarantee that the buffer is filled to the desired amount.

oli-obk:
I think you need to decide on one behavior.

1. do a single read that appends to the buffer
2. or read until EOF or buffer.len() == n

Right now the function suggests 2. but the guaranteed behavior is actually 1.

So either drop the `min` argument and rename to `read_to_buf`or sth, or change the function to guarantee EOF or filled to the minimum.

cybergeek94:
@oli-obk Ideally, `fill_buf()` would read into the buffer every time it's called. I can *kind of* see why that would be undesirable, but the method name sort of implies that's what it does.

Looping until EOF or the minimum is met would be the best bet, but it would break back-compat if it was added to the `BufRead` trait because it couldn't be implemented in terms of `fill_buf()`, at least with its current semantics. I do ask that question in the last section; I don't know if you read that far. But even then, the buffer length can't be guaranteed because the source could be at EOF, so the user would still have to check it.

oli-obk:
You are adding a new function. That function could by default simply return an `Err(Error::new(ErrorKind::Other, FillBufMinNotImplemented))`. The standard library can then actually implement that function for all trait implementors. This would cause some repetition, but it would not be a breaking change.

alternatively as @seanmonstar suggested, the body could simply be `unimplemented!()`.

>  Ideally, fill_buf() would read into the buffer every time it's called. I can kind of see why that would be undesirable, but the method name sort of implies that's what it does.

yea, this was discussed here: https://github.com/rust-lang/rfcs/pull/1015

I wonder if an RFC changing `fill_buf()` to read on every call (until the buffer is full?) would be accepted.

cybergeek94:
Returning an error or panicking by default is undesirable because it would violate users' expecations. I don't think any other part of the stdlib stubs APIs out like that. However, it might be good to discuss this so we can have an established strategy for adding methods to traits.

It doesn't look like the semantics change to `fill_buf()` that @seanmonstar proposed as an alternative was actually discussed at all. That could definitely be a viable approach, and I don't think it would actually break any existing usage of the method.

Another alternative would be to add just `read_into_buf()` and provide a default impl that returns `try!(self.fill_buf()).len()`. Or implement `read_into_buf()` and `get_buf()` as inherent methods on `BufReader`, since that's the struct that's primarily giving us trouble here.

oli-obk:
> Or implement read_into_buf() and get_buf() as inherent methods on BufReader, since that's the struct that's primarily giving us trouble here.

Why not combine both:

`get_buf()` simply forwards to `fill_buf()` but is overridden in all std-impls to not cause a read.

`fill_buf()` always reads if the buffer isn't full yet. (in case the backend is a fixed sized array e.g.)

cybergeek94:
> `get_buf()` simply forwards to `fill_buf()` but is overridden in all std-impls to not cause a read.

How would that work?

oli-obk:
http://is.gd/mqRi6W

alexcrichton:
@cybergeek94 I'm concerned about this RFC for a few reasons, and I'll try to explain them here:

* Primarily, this RFC as written is a breaking change, which unfortunately is no longer acceptable. This must be written assuming that the change is backwards compatible, so one of the alternatives you've listed may want to be taken.
* I'm a little uneasy about the motivation here. You don't have any guarantee about the internal size of the buffer for each buffered reader, so it's somewhat unclear to me in what case you can rely on it being able to give a larger buffer after this method is called.
* If you want previous data to be preserved, it may force all existing data in the buffer to be shuffled down to the beginning of the vector its residing in.
* The documentation on this function says that implementations are not required to meet this minimum, but this seems to me like it's saying you basically can't rely on this function acting its purpose? That seems somewhat odd to me.

cybergeek94:
@alexcrichton

[I did submit a pre-RFC on Internals](https://internals.rust-lang.org/t/pre-rfc-add-bufread-fill-buf-min/2418), yet after almost a week and over a hundred views and one "like", no one has responded to it. I am aware Rustcamp happened last weekend, but other similar discussions that were submitted later have loads of comments. I assumed that meant no one had a strong enough opinion on it to speak their piece.

The primary motivation is made very clear, in that without an API change like this, there's no way to explicitly tell the `BufRead` impl that you want more data in the buffer without emptying it. This is unergonomic for the reasons stated in the RFC. There's no promise about the minimum length of the buffer because it can't always be guaranteed, such as when the source is at EOF. 

The API contract only guarantees a "best effort" to meet that minimum, which could be one bulk read, or as many as needed until the minimum is met or the source returns EOF.

> Primarily, this RFC as written is a breaking change, which unfortunately is no longer acceptable. This must be written assuming that the change is backwards compatible, so one of the alternatives you've listed may want to be taken.

I'm very much aware it is unacceptable because it is breaking. That's precisely the reason why I provided the two alternatives, because I didn't expect the base proposal to be accepted at face-value. I structured it that way to give the base design first, and then I expand on the more practical approaches. If this isn't exactly in the spirit of the RFC process, I apologize. It's only my first go at it.

My preferred alternative is the latter of the two (implement as an inherent method on `BufReader`), because that is what I ended up implementing in my own custom solution. The semantics of `BufReader` is really the primary concern here, because most of the other concrete `BufRead` impls (e.g. `&[u8]`, `Cursor<Vec<u8>>`) return all the data that will ever be available. The only real exception is `StdinLock`, which may benefit from a similar API. 

> If you want previous data to be preserved, it may force all existing data in the buffer to be shuffled down to the beginning of the vector its residing in.

This is something I hadn't considered until recently, but I wasn't sure what should be done about it (my implementation actually does nothing). If cache locality isn't a huge concern, then the data only needs to be shifted down once the end of the buffer allocation, or some threshold, has been reached. If all the data in the buffer ends up being consumed, the impl can just start back at the beginning (which is what mine does).

cybergeek94:
@oli-obk I don't think anyone would go for that because it would depend too much on method selection semantics, which could change later on and silently break it.

oli-obk:
@cybergeek94 they can't change, that would break ridiculous amounts of code. And it would not ever break it truly, just be less efficient due to reading on every `get_buf` call (which can actually be unit tested against, so shouldn't be a problem).

alexcrichton:
> I assumed that meant no one had a strong enough opinion on it to speak their piece.

Unfortunately we're all quite busy so we don't always have time to vet all RFCs, Pre-RFCs, *and* internals posts. Silence doesn't necessarily mean consensus!

> There's no promise about the minimum length of the buffer because it can't always be guaranteed, such as when the source is at EOF, or when the requested size is larger than the internal buffer can hold. The API contract only guarantees a "best effort" to meet that minimum, which could be one bulk read, or as many as needed until the minimum is met or the source returns EOF.

The weird part about this API to me is that it basically provides no guarantees. It says that in some situations you may be able to read some more data, but you can never concretely say in *which* situations without otherwise knowing everything else in play. 

As a result this sounds like a much more appropriate method for concrete instances rather than the trait. For example it doesn't necessarily make sense for anyone operating over a generic `BufRead` to make use of this method because there are so few guarantees about what happens after it's called.

> If this isn't exactly in the spirit of the RFC process, I apologize. It's only my first go at it.

No worries! It's much easier to focus discussion on an RFC if the text is basically perpetually in a state of "ready to be merged".

cybergeek94:
@alexcrichton What might be the best idea here is to go back to the previous RFC and take the `read_into_buf()` and `get_buf()` methods it proposed for the `BufRead` trait, and implement those as inherent methods on `BufReader`. Then its `BufRead` impl can be implemented in terms of those, without changing the semantics. That would suit my use case without breaking anything.

Should I update the RFC in-place or withdraw, revise, and resubmit?

alexcrichton:
> Should I update the RFC in-place or withdraw, revise, and resubmit?

Feel free to update in-place!

> What should people do, then, if they feel like their post isn't getting enough attention?

I think you basically did the right thing opening an RFC afterwards, just don't be super surprised if you hear some new opinons :)

cybergeek94:
@alexcrichton Beginning my revised draft now. I do have one concern we haven't talked about: should we provide a method to allow the user to grow the internal buffer if they need more than 64k bytes? It seems like a desirable feature since we're already giving them control over when the buffer is read into.

cybergeek94:
@alexcrichton I have these in my "Unresolved Questions" section. Perhaps we should address them before I commit the revisions so I can leave that section empty:

* Should `read_into_buf()` return the new length of the buffer instead of the number of bytes read?
    * This would be more ergonomic for the author's specific use-case but this might not always be the case.
    * It might be useful to do `while try!(buf.read_into_buf()) > 0 {}` to make sure the buffer is filled.

* Should another method be added that allows the user to grow `BufReader`'s internal buffer, in case they need more bytes in-memory than `BufReader` allocates by default? (64KiB (64 * 1024) at the time of this writing)

* Should `read_into_buf()` shuffle existing bytes down if there is room at the beginning of the buffer?
    * Perhaps if there is more room at the beginning than there is at the end?

retep998:
I think there should definitely be a method to grow the internal buffer.

cybergeek94:
> <acrichto> DroidLogician: perhaps, yeah, although there may be some form of composition where you get the buffer/cursor out and then put it back in later
> <acrichto> e.g. so you can modify the buffer or w/e
> <acrichto> DroidLogician: for now though I'd keep the RFC as slim as possible

cybergeek94:
So another RFC later on, perhaps.

cybergeek94:
I have pushed the new revisions.

alexcrichton:
@cybergeek94 do you have some example code which makes use of these methods? I'm still a little uneasy about the guarantees that functions like `read_info_buf()` will provide. For example if you're working with a generic `BufReader` then there's no guarantee that any data was actually read, so you can't rely on the method. Additionally the `get_buf` method seems so close to `fill_buf` that I'm not totally convinced its worth it.

That being said, I'm curious to see how this would be leveraged, as I may just be missing something?

seanmonstar:
@alexcrichton an implementation of this can be found in hyper: https://github.com/hyperium/hyper/blob/master/src/buffer.rs

And usage of those methods: https://github.com/hyperium/hyper/blob/master/src/http/h1.rs#L698-L718

Key difference between `get_buf` and `fill_buf`: `get_buf` will **NOT** perform a `read` internally. This is important if you want to get the buffer, and not have the chance of blocking.

cybergeek94:
@alexcrichton While this uses my implementation of `fill_buf_min()`, the principle is the same: https://github.com/cybergeek94/multipart/blob/master/src/server/boundary.rs#L36

Basically, if the buffer isn't long enough, more data needs to be read into it. The guarantee? That `BufReader` will actually *try* to read into the buffer if there's room. `fill_buf()` will only read if the buffer is completely empty, which is highly problematic if you're trying to work with data in the buffer and you realize you need more.

While this would ideally be fixed by `fill_buf()` performing a read every time it's called, I can understand why that would be undesirable. That was the problem that `fill_buf_min()` was supposed to solve.

seanmonstar:
I may publish the buffer used in hyper, if this isn't accepted. Or perhaps
that's the best course?

cybergeek94:
My own custom impl works for me. I just figured it'd be nice to have it back in the stdlib type I started from so I didn't have to maintain my own copy.

alexcrichton:
@seanmonstar 

Doesn't that example function return spurious errors sometimes though? If the buffered reader had a only a few bytes left in it but located at the very end, `read_into_buf` would always return `Ok(0)`, and wouldn't that function return an error? E.g. it'd return `Error::TooLarge` for having only read 2 bytes so far but there may still be bytes on the stream which can be read.

--

@cybergeek94 

I think what you've got makes sense, but what's the benefit over just calling `fill_buf`? It looks like it's basically the same logic as it has to handle the case that the buffer returned is quite small and doesn't have much data inside it. In that sense this may be a performance optimization, but I don't think it helps structure code any differently?

cybergeek94:
@alexcrichton 

I dunno if there's some part of this that I'm not getting across very well. The problem: I have some data in the buffer. I realize that I need more, because there's not enough data in the buffer to make a decision on. If that data isn't what I'm looking for, I need to pass it on to the user of my API and not just discard it, because it's probably meaningful to them in some way.

With the current `BufReader` implementation of `fill_buf()`, I can't always work directly with the buffer because it might give me an incomplete picture. Sometimes, the one read it performs on the source doesn't give me enough data in the buffer to work with. So I need more, but not without discarding the existing data from memory.

I have two possible solutions: clear the buffer and push it to a temporary allocation, or copy `BufReader` and make it work for my purposes. Neither is satisfactory. 

If I work with a temporary allocation, that's more code to shuffle data in and out of it, and to decide when I need to use it and when not. I need to store that buffer in my struct because I might need it later, that's another 12 or 24 bytes, and another field to initialize and drop later. It's also more strain on the CPU cache hierarchy. Some might call this premature optimization, but I would call it "not doing more work than necessary".

Copying the struct to my crate increases my crate's footprint, it's more to maintain, download and build. And technically, it violates the DRY principle because it's an almost-exact copy of an in-distribution struct that doesn't quite work for me as-is. 

I still personally believe in `fill_buf_min()`, I just think I failed to clearly establish its motivations and utility. It still bugs me that "guarantees" are an issue, because in comparison, `fill_buf()` guarantees *nothing*. It's unintuitive that a method named `fill_buf()` would return a nearly empty buffer while more remains to be read just because it's not *actually* empty yet. I understand that performing a read every time it's called defeats the purpose of buffering, but there could at least be a saner threshold than *empty*.

`read_into_buf()` lets the user tell `BufReader` *directly*, "I need more data in your buffer, so you should perform a read *now*, not when it's empty." `get_buf()` is kind-of a gimme from my perspective; you shouldn't have to call `read_into_buf()` and *then* `fill_buf()` to get the buffer, because then you have to deal with two `io::Result` values and the second wouldn't really *mean* anything if you handled the return value of the first.

For `read_into_buf()` to work sanely, it should copy bytes down to the beginning of the buffer after some threshold is reached, like there's more room at the beginning of the buffer than at the end.

alexcrichton:
@cybergeek94 I think I understand your use case here, and I definitely agree that this sort of abstraction should be consolidated and it shouldn't need to be maintained separately. That being said, I'm still hung up on this point:

> Sometimes, the one read it performs on the source doesn't give me enough data in the buffer to work with. So I need more, but without discarding the existing data from memory.

In your specific use case this may work out, but in general this isn't a problem the current `BufReader` type can solve. A `BufReader` does not dynamically grow its internal buffer, so it's possible that a request to read more data simply can't succeed. This means that the API proposed here, `read_info_buf`, doesn't actually provide the guarantee that it reads any more data, so you *already have to be ready for this situation*. As a result, you'll already need to have your own side buffer and handle this logic, so this extra method would purely be a performance optimization rather than a fundamental addition I think.

retep998:
@alexcrichton But if I can guarantee that I'll only need a block up to a certain size, and I can create my `BufReader` with at least that capacity, then I know that I can keep calling `read_into_buf` until I either get enough or the socket errors, no need for an extra side buffer.

For example, if I want to wrap around the schannel API for SSL, I have to provide entire encrypted blocks to schannel, I can't give them piecemeal otherwise it'll error and tell me 
tell me to read more first. However I do at least have the guarantee of SSL record size limits, so as long as my buffer capacity is greater than the limit, using `read_into_buf` will work perfectly.

alexcrichton:
@retep998 you aren't always guaranteed though that the data in your buffer is always at the beginning? The implementations of `read_into_buf` so far all don't shuffle data to the beginning of the buffer, so even if your buffer is large enough for the largest message you could be looking at the buffered data at the very end, unable to move it down and read in some more.

seanmonstar:
True, I should add a 'reset_buf' or something to reset after a request is
completed, so it's always starting at the beginning.

cybergeek94:
@alexcrichton I've made it pretty clear that at some point, `read_into_buf` should move the existing data back to the beginning before reading in more. It's mostly about deciding when is the optimal time to do so. This calls for the kind of benchmarking and microoptimizing that I don't really specialize in. I'm not averse to adding a method that lets the user do this themselves, but I'm sure you can find some problem with this too.

I think there's a critical difference you're trivializing between the case when there's no more room in the buffer to read into and when the buffer is almost empty and won't be read into until it's completely empty. The user has control over the length of the buffer, at least when creating the `BufReader`. If they know ahead of time what they're looking for, they can make sure it's long enough. A method to grow the buffer would help, but you've already said it's out of scope for this discussion.

alexcrichton:
Yeah certainly this case can be handled, but what I've been pointing out is that your existing implementation of `read_into_buf` doesn't handle this case, nor does the RFC in its current form mention this. 

To me this seems like a pretty serious drawback of the functionality as you're not always in a concrete context with complete control over a `BufReader` and in a generic context this function unfortunately doesn't have much meaning (b/c you don't know the size of the internal buffer)

cybergeek94:
@alexcrichton Actually it's in the "Unresolved Questions" section. That's why I wanted to discuss it.

PR #1240, opened 2015-08-07
References into repr(packed) structs should be `unsafe`.

huonw:
[Rendered](https://github.com/huonw/rfcs/blob/repr-packed-unsafe-ref/text/0000-repr-packed-unsafe-ref.md).

cc rust-lang/rust#27060

### DIFF
--- a/text/0000-repr-packed-unsafe-ref.md
+++ b/text/0000-repr-packed-unsafe-ref.md
@@ -1,0 +1,438 @@
+- Feature Name: NA
+- Start Date: 2015-08-06
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Taking a reference into a struct marked `repr(packed)` should become
+`unsafe`, because it can lead to undefined behaviour. `repr(packed)`
+structs need to be banned from storing `Drop` types for this reason.
+
+# Motivation
+
+Issue [#27060](https://github.com/rust-lang/rust/issues/27060) noticed
+that it was possible to trigger undefined behaviour in safe code via
+`repr(packed)`, by creating references `&T` which don't satisfy the
+expected alignment requirements for `T`.
+
+Concretely, the compiler assumes that any reference (or raw pointer,
+in fact) will be aligned to at least `align_of::<T>()`, i.e. the
+following snippet should run successfully:
+
+```rust
+let some_reference: &T = /* arbitrary code */;
+
+let actual_address = some_reference as *const _ as usize;
+let align = std::mem::align_of::<T>();
+
+assert_eq!(actual_address % align, 0);
+```
+
+However, `repr(packed)` allows on to violate this, by creating values
+of arbitrary types that are stored at "random" byte addresses, by
+removing the padding normally inserted to maintain alignment in
+`struct`s. E.g. suppose there's a struct `Foo` defined like
+`#[repr(packed, C)] struct Foo { x: u8, y: u32 }`, and there's an
+instance of `Foo` allocated at a 0x1000, the `u32` will be placed at
+`0x1001`, which isn't 4-byte aligned (the alignment of `u32`).
+
+Issue #27060 has a snippet which crashes at runtime on at least two
+x86-64 CPUs (the author's and the one playpen runs on) and almost
+certainly most other platforms.
+
+```rust
+#![feature(simd, test)]
+
+extern crate test;
+
+// simd types require high alignment or the CPU faults
+#[simd]
+#[derive(Debug, Copy, Clone)]
+struct f32x4(f32, f32, f32, f32);
+
+#[repr(packed)]
+#[derive(Copy, Clone)]
+struct Unalign<T>(T);
+
+struct Breakit {
+    x: u8,
+    y: Unalign<f32x4>
+}
+
+fn main() {
+    let val = Breakit { x: 0, y: Unalign(f32x4(0.0, 0.0, 0.0, 0.0)) };
+
+    test::black_box(&val);
+
+    println!("before");
+
+    let ok = val.y;
+    test::black_box(ok.0);
+
+    println!("middle");
+
+    let bad = val.y.0;
+    test::black_box(bad);
+
+    println!("after");
+}
+```
+
+On playpen, it prints:
+
+```
+before
+middle
+playpen: application terminated abnormally with signal 4 (Illegal instruction)
+```
+
+That is, the `bad` variable is causing the CPU to fault. The `let`
+statement is (in pseudo-Rust) behaving like `let bad =
+load_with_alignment(&val.y.0, align_of::<f32x4>());`, but the
+alignment isn't satisfied. (The `ok` line is compiled to a `movupd`
+instruction, while the `bad` is compiled to a `movapd`: `u` ==
+unaligned, `a` == aligned.)
+
+(NB. The use of SIMD types in the example is just to be able to
+demonstrate the problem on x86. That platform is generally fairly
+relaxed about pointer alignments and so SIMD & its specialised `mov`
+instructions are the easiest way to demonstrate the violated
+assumptions at runtime. Other platforms may fault on other types.)
+
+Being able to assume that accesses are aligned is useful, for
+performance, and almost all references will be correctly aligned
+anyway (`repr(packed)` types and internal references into them are
+quite rare).
+
+The problems with unaligned accesses can be avoided by ensuring that
+the accesses are actually aligned (e.g. via runtime checks, or other
+external constraints the compiler cannot understand directly). For
+example, consider the following
+
+```rust
+#[repr(packed, C)]
+struct Bar {
+    x: u8,
+    y: u16,
+    z: u8,
+    w: u32,
+}
+```
+
+Taking a reference to some of those fields may cause undefined
+behaviour, but not always. It is always correct to take
+a reference to `x` or `z` since `u8` has alignment 1. If the struct
+value itself is 4-byte aligned (which is not guaranteed), `w` will
+also be 4-byte aligned since the `u8, u16, u8` take up 4 bytes, hence
+it is correct to take a reference to `w` in this case (and only that
+case). Similarly, it is only correct to take a reference to `y` if the
+struct is at an odd address, so that the `u16` starts at an even one
+(i.e. is 2-byte aligned).
+
+# Detailed design
+
+It is `unsafe` to take a reference to the field of a `repr(packed)`
+struct. It is still possible, but it is up to the programmer to ensure
+that the alignment requirements are satisfied. Referencing
+(by-reference, or by-value) a subfield of a struct (including indexing
+elements of a fixed-length array) stored inside a `repr(packed)`
+struct counts as taking a reference to the `packed` field and hence is
+unsafe.
+
+It is still legal to manipulate the fields of a `packed` struct by
+value, e.g. the following is correct (and not `unsafe`), no matter the
+alignment of `bar`:
+
+```rust
+let bar: Bar = ...;
+
+let x = bar.y;
+bar.w = 10;
+```
+
+It is illegal to store a type `T` implementing `Drop` (including a
+generic type) in a `repr(packed)` type, since the destructor of `T` is
+passed a reference to that `T`. The crater run (see appendix) found no
+crate that needs to use `repr(packed)` to store a `Drop` type (or a
+generic type). The generic type rule is conservatively approximated by
+disallowing generic `repr(packed)` structs altogether, but this can be
+relaxed (see Alternatives).
+
+Concretely, this RFC is proposing the introduction of the `// error`s
+in the following code.
+
+```rust
+struct Baz {
+    x: u8,
+}
+
+#[repr(packed)]
+struct Qux<T> { // error: generic repr(packed) struct
+    y: Baz,
+    z: u8,
+    w: String, // error: storing a Drop type in a repr(packed) struct
+    t: [u8; 4],
+}
+
+let mut qux = Qux { ... };
+
+// all ok:
+let y_val = qux.y;
+let z_val = qux.z;
+let t_val = qux.t;
+qux.y = Baz { ... };
+qux.z = 10;
+qux.t = [0, 1, 2, 3];
+
+// new errors:
+
+let y_ref = &qux.y; // error: taking a reference to a field of a repr(packed) struct is unsafe
+let z_ref = &mut qux.z; // ditto
+let y_ptr: *const _ = &qux.y; // ditto
+let z_ptr: *mut _ = &mut qux.z; // ditto
+
+let x_val = qux.y.x; // error: directly using a subfield of a field of a repr(packed) struct is unsafe
+let x_ref = &qux.y.x; // ditto
+qux.y.x = 10; // ditto
+
+let t_val = qux.t[0]; // error: directly indexing an array in a field of a repr(packed) struct is unsafe
+let t_ref = &qux.t[0]; // ditto
+qux.t[0] = 10; // ditto
+```
+
+(NB. the subfield and indexing cases can be resolved by first copying
+the packed field's value onto the stack, and then accessing the
+desired value.)
+
+## Staging
+
+This change will first land as warnings indicating that code will be
+broken, with the warnings switched to the intended errors after one
+release cycle.
+
+# Drawbacks
+
+This will cause some functionality to stop working in
+possibly-surprising ways (NB. the drawback here is mainly the
+"possibly-surprising", since the functionality is broken with general
+`packed` types.). For example, `#[derive]` usually takes references to
+the fields of structs, and so `#[derive(Clone)]` will generate
+errors. However, this use of derive is incorrect in general (no
+guarantee that the fields are aligned), and, one can easily replace it
+by:
+
+```rust
+#[derive(Copy)]
+#[repr(packed)]
+struct Foo { ... }
+
+impl Clone for Foo { fn clone(&self) -> Foo { *self } }
+```
+
+Similarly, `println!("{}", foo.bar)` will be an error despite there
+not being a visible reference (`println!` takes one internally),
+however, this can be resolved by, for instance, assigning to a
+temporary.
+
+# Alternatives
+
+- A short-term solution would be to feature gate `repr(packed)` while
+  the kinks are worked out of it
+- Taking an internal reference could be made flat-out illegal, and the
+  times when it is correct simulated by manual raw-pointer
+  manipulation.
+- The rules could be made less conservative in several cases, however
+  the crater run didn't indicate any need for this:
+  - a generic `repr(packed)` struct can use the generic in ways that
+    avoids problems with `Drop`, e.g. if the generic is bounded by
+    `Copy`, or if the type is only used in ways that are `Copy` such
+    as behind a `*const T`.
+  - using a subfield of a field of a `repr(packed)` struct by-value
+    could be OK.
+
+# Unresolved questions
+
+None.
+
+# Appendix
+
+## Crater analysis
+
+Crater was run on 2015/07/23 with a patch that feature gated `repr(packed)`.
+
+High-level summary:
+
+- several unnecessary uses of `repr(packed)` (patches have been
+  submitted and merged to remove all of these)
+- most necessary ones are to match the declaration of a struct in C
+- many "necessary" uses can be replaced by byte arrays/arrays of smaller types
+- 8 crates are currently on stable themselves (unsure about deps), 4 are already on nightly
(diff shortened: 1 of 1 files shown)


arcnmx:
Taking a reference to a member of a packed struct is a very wrong thing to do, unless...

1. You first verify that the alignment is correct.
    - Subpoint: Your type has an alignment of `1`, and it is therefore always correct.
2. You actually just want a pointer to it, to transmute, or use in asm, etc.
3. You want to call a method by `&self`/`&mut self` that doesn't actually dereference `self`..?
4. We modify the language to have typed references with tagged alignments, so the compiler can emit a proper load/store for everything and suddenly it's all safe and good.

Now going to step back and consider what the actual uses of `packed` are...

1. Mapping packed data structures in memory from external sources, such as C libraries or I/O.
    - This can be done safely with unaligned types and `#[repr(C)]` (I have implementations of this in [packed](http://crates.io/crates/packed) and [pod](http://crates.io/crates/pod)). Slight syntax overhead required.
2. You want to save space in your type and have very peculiar requirements. This also can be done safely as above.

Given the above, this sounds like the correct approach to take. I just have a few points to bring up:

1. Can we define 1-aligned types as a safe exception for referencing within packed types (and therefore other packed types would be as well)?
2. I don't really feel like the generic restriction is necessary. As long as all fields can be asserted to maintain the no-Drop requirement, they should be harmless.
3. Same for indexing and referencing fields as long as you're not taking references. Depending on how the compiler is structured this may be additional work to support, though?
4. The inability to use `struct<T>([u8; ::std::mem::size_of::<T>()])` hurts a lot for exposing safe abstractions around packed types :/

huonw:
@arcnmx

> this sounds like the correct approach to take

I'm confused, I don't really understand your comment in the light of https://github.com/rust-lang/rust/issues/27060#issuecomment-128552030 . Is "this" the design the RFC has? (i.e. are you broadly in favour?)

> I don't really feel like the generic restriction is necessary. As long as all fields can be asserted to maintain the no-Drop requirement, they should be harmless.

NB. I do cover this in the Alternatives section. However, I think it's relatively low-benefit and can be introduced later if necessary: I don't recall crater bringing up a single instance of `repr(packed)` on a generic struct that was useful.

> Same for indexing and referencing fields as long as you're not taking references. Depending on how the compiler is structured this may be additional work to support, though?

Yes, also covered in Alternatives. Like with generics, it should be backwards compatible to allow, and the main reason it is explicitly not supported is to make the implementation easier.

> The inability to use struct<T>([u8; ::std::mem::size_of::<T>()]) hurts a lot for exposing safe abstractions around packed types :/

This seems unrelated to this RFC?

> We modify the language to have typed references with tagged alignments, so the compiler can emit a proper load/store for everything and suddenly it's all safe and good.

(Incidentally we can probably do this in libraries if we had some way to specify the desired alignment of a struct, via something similar to ["`Unaligned`"](https://www.reddit.com/r/rust/comments/3dfqn2/efficient_optional_values/ct4xk87).)

arcnmx:
> I do cover this in the Alternatives section

Right, just wanted to cast my vote in.

> I'm confused, I don't really understand your comment in the light of rust-lang/rust#27060 (comment) . Is "this" the design the RFC has? (i.e. are you broadly in favour?)

I changed my mind! As I went through the use cases, my thought process drifted from "packed is nice for mapping data" to "packed is almost completely useless what we actually need is ergonomic safe abstractions instead". So yes, in favour of this RFC - though I worry people will ignore the warning and take unsafe references without understanding the concequences. It's rather subtle and dangerous, though I'm not a fan of denying the references outright either.

> This seems unrelated to the changes this RFC is proposing?

Just another pain point of dealing with packed types. Tangential, yes, but I'm very concerned about the ergonomics surrounding packed data!

arcnmx:
> (Incidentally we can probably do this in libraries if we had some way to specify the desired alignment of a struct, via something similar to "`Unaligned`".)

Also to expand slightly on the constexpr point, if that were working, it would be possible to write the generic `Unalign` type you linked without `#[repr(packed)]` at all, and expose safe unaligned accessors. `packed` would be unnecessary, modulo a bit uglier syntax compared to built-in copy/move/assign. Currently, [you have to manually implement it for specific types](http://arcnmx.github.io/nue/packed/trait.Aligned.html).

bluss:
Really great RFC work! Preparation, motivation, all of it!

It's a good plan, but I'd like to do it without allowing references at all — it's an additional rule for unsafe blocks, and do we really need that?

nikomatsakis:
@bluss 

> It's a good plan, but I'd like to do it without allowing references at all — it's an additional rule for unsafe blocks, and do we really need that?

We debated that, but there are legitimate use-cases for creating a reference, and if we don't permit `&` notation, then one is reduced to doing manual memory arithmetic. As a simple example of when it could be useful, you may know that some fields *are* aligned (even when packed) even if all fields are not. For example, you may wish to obtain a pointer to a `[u8;512]` field embedded in your struct (such a field is always aligned).

arcnmx:
@nikomatsakis 

> For example, you may wish to obtain a pointer to a [u8;512] field embedded in your struct (such a field is always aligned).

Would it then make sense to only disallow references to improperly aligned fields (field_alignment > containing_struct_alignment)? In the context of packed, anything with an alignment of 1, such as u8 arrays, would still be safe to reference.

I dislike removing functionality completely, but an unaligned reference is pretty much always the wrong thing to do. If anything, you want a pointer, not a reference. Which could still be done:

```
#[repr(C, packed)]
struct Something {
    data: [u8; 0x1f],
    unaligned_ptr: (),
    unaligned: u16,
}

let something = Something { .. };
let x = &something.data; // safe and valid
let x = &something.unaligned; // error: cannot take references to improperly aligned fields
let x = &something.unaligned_ptr as *const () as *const u16; // if you *really* want to access the field somehow... Go for it but don't dereference this!
```

bluss:
I think creating an unaligned reference breaks against [Behavior considered undefined](http://doc.rust-lang.org/reference.html#behavior-considered-undefined) which I think unsafe code blocks shouldn't violate either.

nikomatsakis:
On Fri, Aug 07, 2015 at 06:54:24AM -0700, arcnmx wrote:
> Would it then make sense to only disallow references to improperly aligned fields (field_alignment > containing_struct_alignment)? In the context of packed, anything with an alignment of 1, such as u8 arrays, would still be safe to reference.

This would be a more specific rule, yes. It is harder to enforce
because (currently) the type-checking code is platform independent.

> I dislike removing functionality completely, but an unaligned
> reference is pretty much always the wrong thing to do. If anything,
> you want a pointer, not a reference. Which could still be done:

This is an interesting pattern, but it seems unfortunate to be forced
to insert a dummy field of `()` type. Also, the value of a reference
of type `&()` is undefined (well, it can't be NULL, but other than
that...), so I don't know that we would guarantee any particular value
for the pointer you are creating there. I guess we could say that we
guarantee it, but it may require extra logic that reasons about the
origin of a reference, which in general is not knowable, rather than
simply its type.

arcnmx:
@nikomatsakis 

> This would be a more specific rule, yes. It is harder to enforce because (currently) the type-checking code is platform independent.

A limited implementation would be to whitelist the types that are guaranteed to be 1-aligned on all architectures, which I believe is anything of size 1 or 0? Which again, may not be guaranteed for many types, but at least u8, (), [_; 0]... bool? Not sure. Could probably be done as an OIBIT marker trait. We can also use OIBIT markers with `#[config(target = "etc")]` to make it arch-dependent without introducing target-specific code into the lint / type check itself?

> This is an interesting pattern, but it seems unfortunate to be forced to insert a dummy field of `()` type

Unfortunate, sure, but there are *very* few legitimate reasons for doing this if 1-aligned type references are safe and allowed. It was more an example of how to do it if you *really* needed to, without manual pointer offset counting.

> Also, the value of a reference of type `&()` is undefined (well, it can't be NULL, but other than that...), so I don't know that we would guarantee any particular value for the pointer you are creating there.

Hm, even with `repr(C)`? How about other zero-sized types such as [_; 0], struct X, enum Void { }..? Unfortunate, but makes sense :/

nikomatsakis:
@arcnmx 

> Hm, even with repr(C)? How about other zero-sized types such as [_; 0], struct X, enum Void { }..? Unfortunate, but makes sense :/

Well, I don't know that we've ever thought about it that hard before. But I wouldn't be surprised if there were paths where the compiler would say "ah, that's a `&()` pointer, which could never be dereferenced, I'll just replace it with a constant 0x1" and strip out some address arithmetic. Or if no such paths exist today, I wouldn't be surprised if we tried to add some in the future. :)

arielb1:
@nikomatsakis 

`&()` is pretty much equivalent to `NonZero<*const ()>`. trans randomly replacing it with `&*1` would be a bug.

nikomatsakis:
@arielb1 

> &() is pretty much equivalent to NonZero<*const ()>. trans randomly replacing it with &*1 would be a bug.

Well, that is the question, isn't it? I don't think this is entirely clear. That said, it's not clear how we would be able to really optimize by doing such changes, and it's certainly counterintuitive.

Aatch:
To address some comments so far:

Being able to determine which fields are definitely aligned is difficult to do currently. However, even if it was possible, I don't think it should be done. A simple rule like "Cannot take references to fields of packed structures in safe code" will be easier to explain and maintain compared to "Cannot take references to fields of packed structures, unless we're sure they're aligned properly". The idea that taking a reference to one field is safe, but another is not seems strange. i'd much rather that the behaviour was consistent across all the fields of the struct.

Aatch:
I think that an intrinsic for doing unaligned loads would be useful given the issue this is solving. The linked issue is related to SIMD where (on x86) there are separate instructions for aligned and unaligned loads and the alignment information we pass to LLVM causes it use the former based on the assumption that the location is aligned.

Some platforms don't allow unaligned loads at all, so being able to still load from an unaligned (for example) `*const i64` would still be useful.

bluss:
The usual way I know to write an unaligned load is to use memcpy / `ptr::copy_nonoverlapping`

arcnmx:
Hm, I suppose it could be seen as a bit of an inconsistent rule. Is it that difficult to implement using OIBIT? My view as someone trying to use it would basically be "this is perfectly safe why is the compiler getting in my way here?!"

In any case I'm more and more convinced that using `packed` is simply the wrong thing to use in today's Rust. `repr(C)` works alright as long as you can avoid pitfalls, and type-safe abstractions for unaligned accesses are just fine.

retep998:
I'm okay with making references into packed structs be unsafe to create. Just please do __not__ temporarily feature gate `#[repr(packed)]` since the Windows SDK uses packed structs quite a bit so it is essential for `winapi`.

nikomatsakis:
**Hear ye, hear ye.** This RFC is entering **final comment period**.

glaebhoerl:
Would this be forwards compatible with later making it safe again by means of first copying the value out onto the stack (and then back afterwards for `&mut`), as discussed in https://github.com/rust-lang/rust/issues/27060?

arcnmx:
I feel like this is the easiest approach to take for now, but it certainly merits some further discussion in a followup RFC on how to handle it properly. A `packed(alignment)` syntax as well as safe ways to reference packed data would be nice at some point.

> Would this be forwards compatible with later making it safe again by means of first copying the value out onto the stack (and then back afterwards for &mut), as discussed in rust-lang/rust#27060?
This seems strange, especially if the lifetime of the object outlives the current scope.

huonw:
@glaebhoerl I'm not sure. It's certainly unlikely to be backwards incompatible due to causing some `unsafe` blocks to become unneeded (the warning about that is just a lint), but I suspect the lifetime shortening discussed in https://github.com/rust-lang/rust/issues/27060#issuecomment-124863033 may cause issues: given `x: &'a PackedFoo`, `&x.y` wouldn't be able to have lifetime `'a` and changing semantics/lifetimes due to having an `unsafe` block vs. not seems sad. That said, I suspect that the lifetime problems mean we probably don't ever want `&` to do that magically anyway.

huonw:
(The `repr(packed(n))` extension should be just that, a back-compat extension.)

text/0000-repr-packed-unsafe-ref.md
```diff
@@ -0,0 +1,438 @@
...
+
+let x = bar.y;
+bar.w = 10;
+```
+
+It is illegal to store a type `T` implementing `Drop` (including a
+generic type) in a `repr(packed)` type, since the destructor of `T` is
+passed a reference to that `T`. The crater run (see appendix) found no
+crate that needs to use `repr(packed)` to store a `Drop` type (or a
+generic type). The generic type rule is conservatively approximated by
```
vadimcn: Just want to point out that we already have generic `repr(packed)` structs [in rustc](https://github.com/rust-lang/rust/blob/master/src/libstd/rt/dwarf/mod.rs#L29).

retep998: Oh the irony.

huonw: Haha, rustc isn't covered by crater.

It looks like that case could be handled by a trait that uses a monomorphic `#[repr(packed)]` internally to do an unaligned load, since it is just used to load `{u,i}{8,16,32,64,usize}` data?

glaebhoerl:
> I suspect the lifetime shortening discussed in rust-lang/rust#27060 (comment) may cause issues: given `x: &'a PackedFoo`, `&x.y` wouldn't be able to have lifetime `'a` and changing semantics/lifetimes due to having an `unsafe` block vs. not seems sad.

Yeah, that's the kind of thing I was concerned about as well.

> That said, I suspect that the lifetime problems mean we probably don't ever want `&` to do that magically anyway.

I'm not certain. I think we'd pretty likely want it for [`#[repr(compact)]`](https://github.com/rust-lang/rfcs/issues/311) (because there's no _other_ way to get references into those, aligned or not), and given that `repr`s `packed`, `simd`, and `compact` (and potentially other nonstandard `repr`s, say bitfields) all have in common a restriction on interior references, making this kind of thing potentially useful for them, we may at some point want them to work consistently in this respect.

text/0000-repr-packed-unsafe-ref.md
vadimcn: In this case - yes (though it'd be still annoying with 10 permutations).  
But imagine you wanted to write a generic (de)serializer function...   If we go with "no generics" rule, IMHO we should consider adding `unaligned_load`/`store` intrinsics.

rkjnsn:
Is the idea that the syntax for taking a reference into a packed struct would be the same, it would just be required to be within an unsafe block or function?

I'm concerned that if the syntax is the same as the common, safe operation of taking a reference, it would be easy to take a reference to a member of a packed struct without realizing it, because one is already writing an unsafe function, or inside an unsafe block for other reasons. It would also be harder to catch during code reviews, as it would be easy to look over. Knowing it was unsafe would require realizing the struct definition (which may be in a different part of the code) specified `repr(packed)`.

I agree that obtaining a reference to a member of a packed struct should be unsafe, but I also think doing so should have a different syntax from the normal, safe operation to make it more apparent that special care is required. Additionally, I would expect such an operation to return a raw pointer.

sourcejedi:
Has anyone looked at function arguments which _aren't_ references, but might be optimized to a pointer?  Maybe `f(a.b)` where `a.b : [u32: 512]`.

retep998:
In those cases the correct thing to do would be to copy it out onto the stack to get an aligned value and then pass a pointer. Whether we actually do that, I don't know.

sourcejedi:
@retep998 Agreed, that sounds reasonable given how _un_reasonable my example code is.  If there's ever a real (performance) problem from that, you'd just have to find it by profiling.

I guess the answer for correctness is to add a second test, for that case.  (Just checking the alignment, no need to try forcing a crash with simd).

nikomatsakis:
@rkjnsn 

> I agree that obtaining a reference to a member of a packed struct should be unsafe, but I also think doing so should have a different syntax from the normal, safe operation to make it more apparent that special care is required. Additionally, I would expect such an operation to return a raw pointer.

I find this appealing. I wonder what the syntax would be, though?

eddyb:
@nikomatsakis Another usecase for getting a raw pointer could be for uninitialized memory (potentially going through some sort of safe uninitialized reference?).

retep998:
I like the idea of having alternate syntax which gets a raw pointer directly and can be used when getting a safe pointer wouldn't be safe to do.

arcnmx:
Also in favour of an alternative that retrieves a pointer (and doesn't require `unsafe`), as I mentioned previously. Then references can simply be completely disallowed for now. The problem is that we suddenly need to introduce some new syntax for it...

We could introduce an `offset_of!(type, field_ident)` macro maybe?

retep998:
And if we do introduce new syntax we'd need to make sure that while it is unstable, the old method of just taking a reference would still work in stable Rust, albeit with a warning.

arcnmx:
Wasn't the point of this RFC to be an intentional breaking change and remove packed references in stable though, because they're completely wrong and unsafe?

retep998:
Oh yeah, and on second thought, I don't need to be able to get pointers to those fields regardless, I just need to be able to have those packed structs, and be able to access their fields at all. So as long as that stays in, I don't mind.

arcnmx:
Mm, the uses for `packed` are small enough as they are, the uses for **taking references into packed values** are even smaller. At this point I wouldn't mind simply disallowing them entirely until we figure out a better alternative. I do also agree that the "no generics" restriction seems a bit unnecessary.

nikomatsakis:
@huonw 

> Crater was run on 2015/07/23 with a patch that feature gated repr(packed).

So, I'm slow I guess, but for some reason I thought that this patch actually detected cases that took the address of a packed structure. But re-reading this, I realize that this was just a way to detect where `repr(packed)` was being used period. Do you have any idea how many people actually try to take addr of packed fields?

arcnmx:
@nikomatsakis I was, but I've moved on to using `repr(C)` instead.

I can see it being somewhat useful (mostly for referencing other inner packed members, or calling methods by `&self` for wrapper types), but at this point I don't think it would hurt anyone too much if we just made it flat-out invalid until a proper solution comes up.

bill-myers:
This seems too limiting since for instance on x86 unaligned accesses work fine on scalars and there are instructions for unaligned SIMD access.

A more general solution would seem to be adding an "unaligned" modifier to pointers, so that taking such an unaligned reference would produce an "&unaligned T".

Obviously this would not be compatible with normal references, so for instance methods that take "&self" and "&mut self" would not be callable; on the other hand, it should be possible to implicitly convert "&T" to "&unaligned T" and explicitly converting &unaligned T to something like Either<&T, &unaligned T> with a runtime check.

nikomatsakis:
@bill-myers introducing `&unaligned T` does not seem justified vs unsafe
pointers. Unaligned pointers just aren't that important.

nikomatsakis:
**Huzzah!** The language design subteam [has decided](https://docs.google.com/spreadsheets/d/18FtPnEI8RTPdFhGDtj21ITndakEtxVrs8pVzRNx-jL8/edit#gid=0) to accept this RFC.

Regarding the proposal to remove such references altogether, or introduce a new operator, the general feeling was that (as discussed earlier on the thread) references have value, but an entirely new operator did not feel warranted.

Aatch:
I just noticed a hole in this RFC, it doesn't cover by-ref bindings in patterns. So *technically* `let Foo(_, ref x) = foo;` isn't considered unsafe by this RFC while `let x = &foo.1;` is. The RFC probably should be amended to include by-ref bindings, as it seems like a simple oversight.

nikomatsakis:
Indeed. I consider that to be implied. It's worth noting that in the MIR,
by-ref bindings and explicit `&` expressions are the same thing, so if we
moved effect checking to the MIR (as I've been considering) this would be
another example where things get simpler.

Aatch:
@nikomatsakis still, it seems like it should probably be mentioned in the RFC text. At least for the sake of being thorough.

nikomatsakis:
@aatch

> still, it seems like it should probably be mentioned in the RFC text. At
least for the sake of being thorough.

agreed

gnzlbg:
> Huzzah! The language design subteam has decided to accept this RFC.

This is a message from the future: does anyone know where I can find logs of the rationale for this decision ? Meeting minutes, IRC logs, etc ? I'm writing a post about what we can learn about what happened with `repr(packed)`, and what could we do to perform better in the future.

Also if anyone that was around here back then wants to proof-read it, I would appreciate it.

huonw:
@gnzlbg You may already know of these, but there's https://github.com/rust-lang/meeting-minutes and https://github.com/rust-lang/subteams with historical info... however neither of them have much info about `repr(packed)`.

(I'd be happy to proof-read and/or answer questions about the history. Feel free email me at the email in my profile.)

PR #1241, opened 2015-08-07
Forbid wildcard dependencies on crates.io

sfackler:
[Rendered](https://github.com/sfackler/rfcs/blob/no-wildcard-deps/text/0000-no-wildcard-deps.md)

### DIFF
--- a/text/0000-no-wildcard-deps.md
+++ b/text/0000-no-wildcard-deps.md
@@ -1,0 +1,113 @@
+- Feature Name: N/A
+- Start Date: 2015-07-23
+- RFC PR:
+- Rust Issue:
+
+# Summary
+
+A Cargo crate's dependencies are associated with constraints that specify the
+set of versions of the dependency with which the crate is compatible. These
+constraints range from accepting exactly one version (`=1.2.3`), to
+accepting a range of versions (`^1.2.3`, `~1.2.3`, `>= 1.2.3, < 3.0.0`), to
+accepting any version at all (`*`). This RFC proposes to update crates.io to
+reject publishes of crates that have compile or build dependencies with
+version constraints that have no upper bound.
+
+# Motivation
+
+Version constraints are a delicate balancing act between stability and
+flexibility. On one extreme, one can lock dependencies to an exact version.
+From one perspective, this is great, since the dependencies a user will consume
+will be the same that the developers tested against. However, on any nontrival
+project, one will inevitably run into conflicts where library A depends on
+version `1.2.3` of library B, but library C depends on version `1.2.4`, at
+which point, the only option is to force the version of library B to one of
+them and hope everything works.
+
+On the other hand, a wildcard (`*`) constraint will never conflict with
+anything! There are other things to worry about here, though. A version
+constraint is fundamentally an assertion from a library's author to its users
+that the library will work with any version of a dependency that matches its
+constraint. A wildcard constraint is claiming that the library will work with
+any version of the dependency that has ever been released *or will ever be
+released, forever*. This is a somewhat absurd guarantee to make - forever is a
+long time!
+
+Absurd guarantees on their own are not necessarily sufficient motivation to
+make a change like this. The real motivation is the effect that these
+guarantees have on consumers of libraries.
+
+As an example, consider the [openssl](https://crates.io/crates/openssl) crate.
+It is one of the most popular libraries on crates.io, with several hundred
+downloads every day. 50% of the [libraries that depend on it](https://crates.io/crates/openssl/reverse_dependencies)
+have a wildcard constraint on the version. Almost all of them them will fail
+to compile against version 0.7 of openssl when it is released. When that
+happens, users of those libraries will be forced to manually override Cargo's
+version selection every time it is recalculated. This is not a fun time.
+
+Bad version restrictions are also "viral". Even if a developer is careful to
+pick dependencies that have reasonable version restrictions, there could be a
+wildcard constraint hiding five transitive levels down.  Manually searching the
+entire dependency graph is an exercise in frustration that shouldn't be
+necessary.
+
+On the other hand, consider a library that has a version constraint of `^0.6`.
+When openssl 0.7 releases, the library will either continue to work against
+version 0.7, or it won't. In the first case, the author can simply extend the
+constraint to `>= 0.6, < 0.8` and consumers can use it with version 0.6 or 0.7
+without any trouble. If it does not work against version 0.7, consumers of the
+library are fine! Their code will continue to work without any manual
+intervention. The author can update the library to work with version 0.7 and
+release a new version with a constraint of `^0.7` to support consumers that
+want to use that newer release.
+
+Making crates.io more picky than Cargo itself is not a new concept; it
+currently [requires several items](https://github.com/rust-lang/crates.io/blob/8c85874b6b967e1f46ae2113719708dce0c16d32/src/krate.rs#L746-L759) in published crates that Cargo will not:
+
+ * A valid license
+ * A description
+ * A list of authors
+
+All of these requirements are in place to make it easier for developers to use
+the libraries uploaded to crates.io - that's why crates are published, after
+all! A restriction on wildcards is another step down that path.
+
+Note that this restriction would only apply to normal compile dependencies and
+build dependencies, but not to dev dependencies. Dev dependencies are only used
+when testing a crate, so it doesn't matter to downstream consumers if they
+break.
+
+# Detailed design
+
+Alter crates.io's pre-publish behavior to check the version constraints of all
+compile and build dependencies, and reject those that have no upper bound. For
+example, these would be rejected:
+
+ * `*`
+ * `> 0.3`
+ * `>= 0.3`
+
+While these would not:
+
+ * `>= 0.3, < 0.5`
+ * `^0.3`
+ * `~0.3`
+ * `=0.3.1`
+
+# Drawbacks
+
+The barrier to entry when publishing a crate will be mildly higher.
+
+In theory, there could be contexts where an unbounded version constraint is
+actually appropriate?
+
+# Alternatives
+
+We could continue allowing these kinds of constraints, but complain in a
+"sufficiently annoying" manner during publishes to discourage their use.
+
+# Unresolved questions
+
+Should crates.io also forbid constraints that reference versions of
+dependencies that don't yet exist? For example, a constraint of `>= 0.3, < 0.5`
+where the dependency has no published versions in the `0.4` range.
--- a/text/0000-no-wildcard-deps.md
+++ b/text/0000-no-wildcard-deps.md
@@ -11,7 +11,7 @@ constraints range from accepting exactly one version (`=1.2.3`), to
 accepting a range of versions (`^1.2.3`, `~1.2.3`, `>= 1.2.3, < 3.0.0`), to
 accepting any version at all (`*`). This RFC proposes to update crates.io to
 reject publishes of crates that have compile or build dependencies with
-version constraints that have no upper bound.
+a wildcard version constraint.
 
 # Motivation
 
@@ -40,10 +40,13 @@ guarantees have on consumers of libraries.
 As an example, consider the [openssl](https://crates.io/crates/openssl) crate.
 It is one of the most popular libraries on crates.io, with several hundred
 downloads every day. 50% of the [libraries that depend on it](https://crates.io/crates/openssl/reverse_dependencies)
-have a wildcard constraint on the version. Almost all of them them will fail
-to compile against version 0.7 of openssl when it is released. When that
-happens, users of those libraries will be forced to manually override Cargo's
-version selection every time it is recalculated. This is not a fun time.
+have a wildcard constraint on the version. None of them can build against every
+version that has ever been released. Indeed, no libraries can since many of
+those releases can before Rust 1.0 released. In addition, almost all of them
+them will fail to compile against version 0.7 of openssl when it is released.
+When that happens, users of those libraries will be forced to manually override
+Cargo's version selection every time it is recalculated. This is not a fun
+time.
 
 Bad version restrictions are also "viral". Even if a developer is careful to
 pick dependencies that have reasonable version restrictions, there could be a
@@ -77,37 +80,52 @@ build dependencies, but not to dev dependencies. Dev dependencies are only used
 when testing a crate, so it doesn't matter to downstream consumers if they
 break.
 
+This RFC is not trying to prohibit *all* constraints that would run into the
+issues described above. For example, the constraint `>= 0.0.0` is exactly
+equivalent to `*`. This is for a couple of reasons:
+
+* It's not totally clear how to precisely define "reasonable" constraints. For
+example, one might want to forbid constraints that allow unreleased major
+versions. However, some crates provide strong guarantees that any breaks will
+be followed by one full major version of deprecation. If a library author is
+sure that their crate doesn't use any deprecated functionality of that kind of
+dependency, it's completely safe and reasonable to explicitly extend the
+version constraint to include the next unreleased version.
+* Cargo and crates.io are missing tools to deal with overly-restrictive
+constraints. For example, it's not currently possible to force Cargo to allow
+dependency resolution that violates version constraints. Without this kind of
+support, it is somewhat risky to push too hard towards tight version
+constraints.
+* Wildcard constraints are popular, at least in part, because they are the
+path of least resistance when writing a crate. Without wildcard constraints,
+crate authors will be forced to figure out what kind of constraints make the
+most sense in their use cases, which may very well be good enough.
+
 # Detailed design
 
-Alter crates.io's pre-publish behavior to check the version constraints of all
-compile and build dependencies, and reject those that have no upper bound. For
-example, these would be rejected:
+The prohibition on wildcard constraints will be rolled out in stages to make
+sure that crate authors have lead time to figure out their versioning stories.
 
- * `*`
- * `> 0.3`
- * `>= 0.3`
-
-While these would not:
-
- * `>= 0.3, < 0.5`
- * `^0.3`
- * `~0.3`
- * `=0.3.1`
+In the next stable Rust release (1.4), Cargo will issue warnings for all
+wildcard constraints on build and compile dependencies when publishing, but
+publishes those constraints will still succeed. Along side the next stable
+release after that (1.5 on December 11th, 2015), crates.io be updated to reject
+publishes of crates with those kinds of dependency constraints. Note that the
+check will happen on the crates.io side rather than on the Cargo side since
+Cargo can publish to locations other than crates.io which may not worry about
+these restrictions.
 
 # Drawbacks
 
 The barrier to entry when publishing a crate will be mildly higher.
 
-In theory, there could be contexts where an unbounded version constraint is
-actually appropriate?
+Tightening constraints has the potential to cause resolution breakage when no
+breakage would occur otherwise.
 
 # Alternatives
 
 We could continue allowing these kinds of constraints, but complain in a
 "sufficiently annoying" manner during publishes to discourage their use.
 
-# Unresolved questions
-
-Should crates.io also forbid constraints that reference versions of
-dependencies that don't yet exist? For example, a constraint of `>= 0.3, < 0.5`
-where the dependency has no published versions in the `0.4` range.
+This RFC originally proposed forbidding all constraints that had no upper
+version bound but has since been pulled back to just `*` constraints.


seanmonstar:
:+1:

I warn people not to do so with hyper, but it still happens, and I break
lots of people when releasing new breaking versions.

lfairy:
An alternative is to continue allowing wildcard bounds, but recruit a team of trusted users to tighten the bounds if any breakage comes up.

Another thing to consider is an `--allow-newer` option, which tells Cargo to *ignore* upper bounds during dependency resolution. This is useful when you know a crate will build against a newer dependency, but the maintainer hasn't updated it yet.

Both solutions are used by the Haskell community, which has grappled with dependency issues for quite some time.

Ericson2314:
Definitely +10000 for post-hoc changing of the constraints. I'd also love it for users to suggest changes to the constraints, and a flag for cargo to go by the user consensus instead of the official constraints (consider abandoned libraries)

This is still useful if we decide its better to start conservative and then relax constraints.

sfackler:
Changing dependency constraints for published crates seems like it would wreak havoc on build reproducibility. We could perhaps add a fourth "level" that could only be incremented on the crates.io side, so version 3.2.4's dependencies could be corrected in version 3.2.4.1.

Ericson2314:
Well definitely log changes.

I suppose one depend on a specific subsets of that "fourth level" but I'd wait to see if reproducibility is a problem in practice. I guess my initial impression is that the change is *supposed* to be unobservable, but I haven't thought that through.

sfackler:
Hmm, I think I might have been thinking about that wrong and reproducibility wouldn't be an issue.

Diggsey:
Crazy idea: disallow wildcard dependencies, but allow lower bounds. In that case, when a dependency is updated to a new major version, automatically build and test with the updated dependency. If the build, or any of the tests fail, automatically insert the correct upper bound.

Realistically, that's all the crate maintainer is going to do anyway, so may as well automate it!

alexcrichton:
I'm in favor of this RFC, the motivation is quite right in that not having an upper bound on a version constraint is basically guaranteed to break at some point in the future.

---

@lfairy 

> An alternative is to continue allowing wildcard bounds, but let maintainers tighten the bounds post-hoc if any breakage comes up.

I agree with @sfackler that this is not a great idea for reproducibility. A lock file indicates a solution to the dependency graph provided by the set of `Cargo.toml` files in play, but this only works because the constraints in place by the dependency graph are constant. If they were allowed to change over time (e.g. the bounds are changed) then previously generated lockfiles which represented a solution will no longer be a valid solution.

It could in theory be possible to track revisions and then you lock to a particular index file at a particular point in time, but that seems much more complicated than just rejecting `*` dependencies which, as @sfackler pointed out, are basically guaranteed to break in the future.

---

@Diggsey 

> Crazy idea: disallow wildcard dependencies, but allow lower bounds. In that case, when a dependency is updated to a new major version, automatically build and test with the updated dependency. If the build, or any of the tests fail, automatically insert the correct upper bound.

Unfortunately I don't think this can be done in a bulletproof fashion. Your dependencies can become part of your own public API if you just export or return one of the types, so an upgrade across the major version of a dependency may not break you but could break your own downstream clients. It generally seems like the best practice is no matter what still "have an upper bound".

If you really want to keep the current behavior then dependencies can look like:

```toml
foo = ">= 0.0, <= 10000"
```

(or something like that)

bluss:
I wish to solve this too, the way `"*"` is used as default now is worrying.

I imagine a few crates actually want the convenience of `"*"`, if they are constantly changing. For example crates tracking nightly, syntax extensions could be in this position.

Diggsey:
> If you really want to keep the current behavior then dependencies can look like:

I was specifically interested in avoiding the time pressure: when an upstream crate upgrades to a new major version, it basically has a domino effect, where all downstream crates have to update as quickly as possible, as soon as all their dependencies are updated to use the new version. Invariably, there's one crate whose maintainer is on holiday or busy or whatever, which results in the entire ecosystem being held back.

On the other hand, allowing upper bounds to get ahead of the actual versions of crates seems to defeat the point of removing wildcard support: if you can continue to use "<= 1000" then for all intents and purposes, it's a wildcard bound.

While automated upgrades as I described would only be as good as the tests, there's nothing to say that their maintainers do more than just rerun their test suite with the new version anyway.

Another option would be to allow upper bounds to only go say one or two major versions ahead of the actual published crate. With a little help from cargo, that could be quite powerful (eg. an easy way for the maintainer to switch from using the latest published version of a crate, to one tagged 'beta' in the associated git repo, would allow them to also test one version ahead, and allow everyone to update their crates in parallel ready for when one of their upstream crates has a new major version)

gsingh93:
While I'm mostly for this RFC, I'm concerned about the point @bluss brings up that there may be some legitimate uses for wildcard dependencies. Maybe there could be some opt-in way to allow wildcard dependencies, which would discourage the majority of people from enabling it unless it was really needed? Or maybe if this is something only unstable projects need, we could allow this for projects with a major revision number of zero (I don't know if this actually means unstable in semver, but it's how I use it)?

sfackler:
Super unstable projects are not really served well by crates.io's current setup - you end up with tons of versions that live in S3 forever wasting space.

Maven has `SNAPSHOT` versions in addition to normal release versions, which differ from normal release versions in that they're mutable. The dependency is simply overwritten every time it's updated.

seanmonstar:
Could "super unstable projects" just be depended on via a git dependency?

```toml
[dependencies.super_unstable]
git = "https://githuib.com/super/unstable"
```

sfackler:
Yeah, that is definitely the other alternative, and now that I think about it, seems like a better method than SNAPSHOT publishes given that Cargo has built in git support.

jmesmon:
I'm not really a fan of this. Most of the time, my code is fine with newer versions of other crates. Forbidding wildcards just means that I'll try to work around it (by trying to set very high max versions) or just have to spend more time needlessly bumping versions in my Cargo.toml files, pushing new versions to git, and publishing a new version on crates.io. We'll be trading an occasional need to fix things with an ongoing desire to bump version numbers of dependencies.

Without some type of automation to handle updating Cargo.toml files like we have for Cargo.lock files, I think this is definitely a bad idea. Even with that automation, I'm concerned this will result in more things than necessary being linked into projects (more than 1 version of a crate) even in cases where 1 version of a crate would work.

sfackler:
"Most of the time" isn't really good enough if your constraint choices end up breaking random downstream users.

seanmonstar:
@jmesmon very likely scenario: your crate `foo` depends on crate `zed` with a `*` version. zed releases version `2.0` with breaking changes the same time you're on vacation. All users of your crate suddenly find that they are downloading `zed@2.0`, and their project is no longer compiling.

Worse still, what if you no longer care about crate `foo`. You've moved on to better, green crates. So you don't pay attention to when it suddenly cannot build anymore. Users scramble trying to figure out how to do so.

----

Compare to you depending on `zed = "1.0"`, which will accept **ALL** versions of `1.x`. Users will automatically get minor and patch releases, such as `1.7.89`. Assuming `zed` follows semver, everyone is happy.

gsingh93:
I just realized one reason I use wildcards in a lot of my projects is I don't want to bother looking up the version number of the project. I often know the name of the crate, but don't know the version off the top of my head. I'd be much more willing to ditch wildcard dependencies if I could ask Cargo to add a dependency, and Cargo would add a reasonable default version (maybe only the major and minor version numbers, i.e. 1.0) or maybe the default could be configured. There are some package managers that have this functionality (i.e. `bower install jquery --save`), having this in Cargo might help here.

Ericson2314:
As convenient as it would be, I think it is hypocritical to allow "git dependencies" but not wildcards. Both offer the same impossible promise that some library will never make a breaking change to it's interface. If we need one, I suggest we allow both, but banish package (releases) that use them to some "unsafe sublisting" of packages.

@alexcrichton Revising the bounds is done because it is impossible to get the bounds both sound and complete---because upstream revisions that haven't been made yet. I am OK with shrinking the bound to make it sound being a patch version bump---since one can rely on bugs in the code, one can rely on bugs produced by an unsound bound that nevertheless results in a successful build. But extending the bound (in attempt to make it complete) is *always* a harmless change---even when the bound is extended unsoundly---since lock files ensure reproducible builds.

This reminds me, does cargo attempt to avoid earlier patch versions? Since the only reason to publish a new patch version is a performance increase or bug fix, it seems between foolish and dangerous to accept earlier patch versions---don't want to get downstream "accustomed" to an upstream bug, especially one that has been fixed. By contrast, a bump in the minor version may just indicate an extended interface, in which case there is there no harm whatsoever in using the older version.

bluss:
@gsingh93 Cargo will still support "*", so you can use that during initial development. Then just write in the versions you used before publishing?

sfackler:
@Ericson2314 git dependencies are already forbidden on crates.io. Nothing in here proposes to change that.

I tentatively think that loosening constraints after publish would be a reasonable thing to do, but that seems out of scope for this RFC.

text/0000-no-wildcard-deps.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Detailed design
+
+Alter crates.io's pre-publish behavior to check the version constraints of all
+compile and build dependencies, and reject those that have no upper bound. For
+example, these would be rejected:
+
+ * `*`
+ * `> 0.3`
+ * `>= 0.3`
```
bluss: I assume `0.3.*` will still be ok (it is in common use). What about `0.*`?

sfackler: Yeah, both of those would be fine (I didn't even realize they existed!)

sfackler: `0.*` is a poor choice, but I'm not sure if it's worth explicitly prohibiting. I think we may want to wait and see.

text/0000-no-wildcard-deps.md
```diff
@@ -0,0 +1,113 @@
...
+
+While these would not:
+
+ * `>= 0.3, < 0.5`
+ * `^0.3`
+ * `~0.3`
+ * `=0.3.1`
```
bluss: In that case, I'll ask, will `^0 := >=0.0.0 <1.0.0` be ok?  See http://doc.crates.io/crates-io.html

sfackler: `^0`, `0.*`, `>= 0.0.0, < 1.0.0` etc are probably bad choices but my gut feeling is that we shouldn't try to get too smart about forbidding these kinds of things for now. We can always add more restrictions later if we feel it's necessary. I'll add a thing in the unresolved questions about those kinds of ranges.

jmesmon:
@seanmonstar @sfackler Isn't using/shipping a Cargo.lock file supposed to prevent some of the issues you're concerned about?

Right now downstream crates can avoid bumping the versions by controlling their Cargo.lock file a bit more directly.

Perhaps an alternate that doesn't increase the burden of maintaining crates would be to provide tools that allow more fine grained control of Cargo.lock?

sfackler:
A crate is published once but depended on literally thousands of times. I am more than happy to impose a small cost on the publish side to minimize costs on the consuming side.

withoutboats:
Huge :+1:, this will help a lot to prevent crates.io from falling into the kind of transitive dependency hell situation other package repositories have been plagued with.

As a general principle, I think crates.io should strive provide these "invariants":
* The dep and transitive deps will always exist (achieved by disallowing non-crates.io deps)
* The dep and transitive deps will be unambiguously licensed (achieved by requiring a valid license field)
* IFF the dep and transitive deps follow the semver standard, stable deps will always compile on the stable branch version `^[the most recent stable branch]`.

This goes partway toward achieving the last one, though it would further require a few "additional standards" to be applied to crates marked 1.0.0 or greater, specifically that:
* They compile on the most recent stable Rust.
* All of their dependencies and transitive dependencies are 1.0.0 or greater.

alexcrichton:
This RFC is now entering its week-long **final comment period**

retep998:
I'd just like to mention that if this lands I will have to modify the dependencies of 420 crates.

huonw:
@retep998 this should be trivial to script a replacement, e.g. replace `"*"` with `">= 0.0, <= 10000"` as @alexcrichton suggests above if you wish to continue using the suboptimal version spec.

sfackler:
This RFC also doesn't propose to do anything to existing versions that have already been updated. You'll need to adjust the constraints the next time you release one of those 420 (!) crates, but presumably be touching the source anyway.

jimmycuadra:
I'm very against this proposal.

My experience with this issue that lead to me being in favor of upper-boundless dependencies was in maintaining [Lita](https://github.com/jimmycuadra/lita), a chat bot for Ruby similar to GitHub's Hubot, where users write their own plugins which are published to the central RubyGems server (rubygems.org) as gems. rubygems.org and "gems" are equivalent to crates.io and crates in the Rust world, respectively.

One of Lita's features is the ability to generate a skeleton gem with all of the necessary files to quickly publish it as a plugin to rubygems.org. This is roughly equivalent to the `cargo new` command. Initially, the generated plugins were locked to the major version of Lita that generated them. For example, if I used Lita 2.1 to generate a plugin, the gemspec file (equivalent to Cargo.toml) would depend on Lita ~> 2.1. That is, 2.1.x as the minimum, and anything in the 2.x.x series, disallowing 3.0.0 and higher.

When I released Lita 3.0, I took great care to make sure that the new version did not break existing plugins published to rubygems.org, because I wanted end users to be able to upgrade Lita itself without waiting for all the maintainers of the plugins they used to upgrade their plugins to be compatible. Unfortunately, all the plugins that had used the generated gemspec were locked to the old version of Lita, so even though they would have worked, every single published plugin had to be updated by its author, which was not well received, and deservedly so.

Recall that in Semantic Versioning, a major version change is required for breaking changes, but it doesn't mean that breaking changes are required. It's possible if not probable that specific breaking changes don't actually affect many programs that depend on the library because they don't happen to use the code paths that changed.

In Lita, I follow an approach where major version changes involve two things:

1. Deprecations of APIs that will be removed in the *next* major version.
2. Removal of APIs that were deprecated in the *last* major version.

This has the following effects for plugin authors, assuming their plugin has no upper bound on its Lita dependency:

* If their plugin does not use any of the APIs newly deprecated, they don't have to do anything at all. Their plugin continues to work as it always did.
* If their plugin uses deprecated APIs, it still continues to work for now, and they have an entire major release cycle in which to migrate their plugin to use new, future-compatible APIs.

If the dependency on Lita were locked to a major version instead, they would *always* have to update their plugin just to change the version dependency, even if no code changes were required. This in turn would result in every user of Lita needing to update the version of each plugin they use. Some of these plugins may not even be maintained anymore, and end users are then forced to fork the original code to make the trivial version dependency change just to keep running it otherwise unaltered. Very frustrating churn for everyone involved.

Yes, it's possible that a future update to a dependent library will break your project. But it is not a guarantee, and the nature of software is to move forward. If a library intends to stay active and engaged with its community, I'd argue it is the responsibility of its maintainers to update it when breaking changes in its dependencies actually happen in a way that affects it. Disallowing upper bounds doesn't prevent that update from being necessary when the dependencies do break. That will happen either way. But it does force both users and library authors to do essentially pointless work in the most common case.

BurntSushi:
@jimmycuadra One way to solve your specific problem is to use a version bound of `>= 0.0, <= 10000`.

> If a library intends to stay active and engaged with its community, I'd argue it is the responsibility of its maintainers to update it when breaking changes in its dependencies actually happen in a way that affects it.

Agreed. But in the mean time, let's not break the builds for everyone else needlessly.

jmesmon:
@BurntSushi if everyone can just switch to using `>= 0.0, <= 10000`, I don't see a point in banning `*`.

Is our goal just to make people type more? Will we forbid `>= 0.0, <= 10000` in the future and force a second migration to some other alternative?

bluss:
The goal is to have users pin versions like  "x.*" or "0.y.*" , in a way that avoids breaking upgrades, as far as possible.

daboross:
@jmesmon I believe the idea is that the majority of libraries on crates.io will, in fact, have breaking changes on new major versions.

If someone is determined to allow their library to break on new versions of other libraries, they are still free to do that, with extra work. However, this RFC will require everyone which is using `*` to re-evaluate their decision at the very least, and if they are completely sure that they want their library to break they are still free to use `>= 0.0 <= 10000`. However, I believe the majority of people using `*` are pretty much just being lazy about dependency management, and don't actually want to have their library fail to build in the future.

In your specific case, `*` doesn't seem like the best idea either. `*` is the easiest solution, but I think `>=2.0 < 4.0` would probably be better, since two major versions = breaking changes.

jmesmon:
@daboross my point is that we're just making `>= 0.0 <= 10000` into the new "easiest solution"

daboross:
@jmesmon `>= 0.0 <= 10000` is definitely explicitly a "hack" construct though. I mean `*` could definitely look idiomatic, while `>= 0.0 <= 10000` makes it very explicit that this is what you are intending to do, and that it wasn't intended for you to do it.

jimmycuadra:
You shouldn't have to use a hack for what is a completely valid use case. FWIW, in the Ruby ecosystem, if you have open-ended dependecies, you get a warning to stderr when building the package, but it is not rejected outright by rubygems.org. I still think it's not the package manager's business to have an opinion about this, but a warning is a lot better than what is being proposed in this RFC.

retep998:
Perhaps you should need to pass a force flag of some sort in order to `cargo publish` when you have a `"*"` dependency?

BurntSushi:
> I still think it's not the package manager's business to have an opinion about this

It won't. This is a restriction put on crates.io.

Ericson2314:
If we do care about soundness, make it so constraints can only refer to versions that are already published.

seanmonstar:
> It won't. This is a restriction put on crates.io.

Exactly. It's useful to remember this only affects publishing to crates.io. Any other registry can have it's own opinion, and you can also happily use `*` locally.

jimmycuadra:
Sorry, I used the wrong term there. I was referring to crates.io. I don't think it is Cargo *or* crates.io's business to have an opinion about this.

sfackler:
@jimmycuadra 

> In Lita, I follow an approach where major version changes involve two things:
>
> * Deprecations of APIs that will be removed in the next major version.
> * Removal of APIs that were deprecated in the last major version.

That is certainly a excellent way to manage a library! Cargo's version constraint support allows developers to take care of that sort of strategy without needing to resort to wildcard constraints. If I'm working on a library that depends on Lita 2, the most recent version, then the conservative approach is to add a constraint of `^2.0.0`, as you mentioned. However, if I'm sure that I'm not using any deprecated functionality (and I trust that only deprecated functionality will break in the next release), I can then adjust my constraints out to `>= 2.0.0, < 4.0.0`, allowing the library to be used with both version 2 *and* version 3. When version 3 comes out, I have an entire release cycle to figure out if I need to remove deprecated functionality and adjust the version constraints as appropriate.

> But it is not a guarantee, and the nature of software is to move forward. If a library intends to stay active and engaged with its community, I'd argue it is the responsibility of its maintainers to update it when breaking changes in its dependencies actually happen in a way that affects it.

It would be fantastic for authors to always update their libraries immediately when something happens to break them, but things don't always work out that way. Maybe the author is on vacation for a month and isn't checking his or her email. Maybe the library's simply been abandoned (this is not at all uncommon in ecosystems with a long history like Java's. For example, Trove is a fairly popular library that hasn't seen an update since 2012 and is very much not feature complete). In both cases, users are stuck doing painful hackery to "recover" from a `cargo update`.

Things can [go *really* wrong](https://www.kickstarter.com/projects/2066438441/haunts-the-manse-macabre/posts/373466) if you don't have any constraints on what versions of dependencies you pull in, and your users will *not* be happy to have to go spelunking through transitive dependencies to figure out which library they have to force back to what version.

Even if you keep your library up to date such that future releases of your dependencies don't break you, a wildcard version means your library supports *all past versions* as well, which is almost *certainly* not the case. To me, a wildcard constraint just means "versioning is hard, I give up".

It's also not a fundamentally terrible thing for libraries to not use the most recent bleeding edge version of a library (speaking as someone who has burnt days to searching through the source versions of Jackson and Dropwizard to track down bugs introduced in a new release, staying off the bleeding edge is sometimes *preferable*). If a library depends on an old version of another and nobody cares, does it matter? I maintain a decent number of crates.io packages, all of which have bounded version constraints. Over the last year or so of crates.io being a thing, I think I've been asked twice or maybe three times to expand a constraint. With those requested bumps and bumps that I've done manually to pick up new features, I'd say that the new version did break things for my library maybe 50% of the time.

> Sorry, I used the wrong term there. I was referring to crates.io. I don't think it is Cargo or crates.io's business to have an opinion about this.

crates.io does have a responsibility to make sure the ecosystem of crates uploaded to it has some sense of reasonableness in my opinion. It doesn't allow crates to be hard deleted (except in extreme circumstances) so that users can be relatively confident that they'll be able to build their code indefinitely. It requires a license because using an unlicensed library is often totally impossible. It requires a description so users have some idea of what the thing is and if it might fit their use case. We would add this new restriction because it would give users the confidence that the libraries they depend on will compile, and continue to compile over time.

@BurntSushi 

I wouldn't encourage the use of things like `> 1, < 100000`. If those kinds of constraints start popping up, we're really no better off than before and there will probably be a follow up RFC to deal with that.

@Ericson2314 

I'm a little hesitant to be that strict up front - there are use cases like @jimmycuadra's where you'd want to be able to look a little bit into the future. It might make sense to allow constraints to refer to the next unreleased version, but no further though.

@retep998 

That's certainly an alternative, though I would personally push for something that's a bit more "in your face" than a simple "--force" flag or something like that.

jimmycuadra:
Good points, @sfackler. I definitely wasn't advocating for the "*" wildcard – I'm only concerned with forcing an upper bound. I think there are some trade offs here, both positive and negative effects for the main library author and the plugin authors (continuing my Lita example.) Let me see if I can enumerate them, partially for my own clarity of thought:

The columns here represent two options for crates.io's behavior as we are discussing. The rows represent the role of the crate (or application in the case of the end user), and the time, effort, attention, and responsiveness of their respective authors.

 | No upper bounds (current) | Upper bounds (proposed)
--- | --- | ---
Library author | If any plugins or end users have direct dependencies on some of the same crates the main library also depends on, they are free to use the newest version without having to wait for an update for the library. The newest version of the shared dependency may break the main library in some way, but will be apparent immediately if it causes the plugin to stop working. The library author does not need to continuously release new versions of the library every time a new version of one of its dependencies is released. | Plugin authors are bound to the versions of any shared dependencies required by the main library. If they want to use a newer one, they must either wait for an update to the main library that relaxes the version constraint on the shared dependency in question, or contact the author directly to nudge them to do it. If the shared dependency's new version includes fixes for any security issues, the plugin is blocked on addressing these issues until the main library allows it to by relaxing the version constraint.
Plugin author | Updates to the library may cause the currently published version of the plugin to break. This is the situation this RFC is trying to prevent. However, if the library never makes changes that would break the plugin, the plugin will continue to work for end users even if the author stops maintaining it. This is a very valuable property with a large ecosystem of plugins. | Updates to the library *always* require an update to the plugin in order for end users to use the latest version of the library. If there were breaking changes, code must be changed, otherwise only the version requirements must change. There is no danger of the current published crate breaking at any time, which is a valuable guarantee for crates hosted on crates.io.
End user | When a new version of the main library is released, it's a drop-in upgrade as long as the plugins they use continue to work. However, some plugins might break. If this happens, the user must either wait for the plugin author to issue a new version, or contact the author directly about the problem. If the author no longer maintains the plugin or is otherwise unresponsive, the user must either stop using the plugin or maintain their own fork and fix the issue themselves. The end user's program also exhibits the same relationship to its upstream dependencies as the plugin does to the main library, in that it can use the latest version of any shared dependencies without requiring a change to any of the upstream libraries. | When the main library is updated, all plugins will now be unusable until they are each updated as well. If the maintainer of a plugin is unresponsive, the user will either have to stop using the plugin (which may have continued to work just fine without the upper version constraint) or fork, fix, and maintain a version of the plugin themselves.

Overall, I think the trade off boils down to which is more important at the expense of the other: a crate that will always compile, or increased churn for all programmers all the way down through the ecosystem of a project, whether or not real changes are needed.

I do see the value of both, but again, I think it's a subtle enough issue that as a library maintainer I should be able to make the choice for myself and still enjoy the benefits of having my crate hosted on crates.io.

jimmycuadra:
I'd also like to quote @Diggsey from much earlier in this discussion, as he articulated some of my concerns with this proposal quite well:

> I was specifically interested in avoiding the time pressure: when an upstream crate upgrades to a new major version, it basically has a domino effect, where all downstream crates have to update as quickly as possible, as soon as all their dependencies are updated to use the new version. Invariably, there's one crate whose maintainer is on holiday or busy or whatever, which results in the entire ecosystem being held back.

> On the other hand, allowing upper bounds to get ahead of the actual versions of crates seems to defeat the point of removing wildcard support: if you can continue to use "<= 1000" then for all intents and purposes, it's a wildcard bound.

Ericson2314:
> It might make sense to allow constraints to refer to the next unreleased version, but no further though.

These properties preserve soundness iff the bound is sound at publication time, provided upstream does not break semver:
 - The starting version in a range should always be published
 - The end version in a range must either
    - be published
    - be an exclusive bound on the next unpublished major.

These properties disallow constraints assuming upstream will violate Semver in the future (but allow for violations that already occurred)
 - The end version may be a major release
 - The end version may be a minor or patch release bound is exclusive and it is published

These two lists combine.

----------------

@jimmycuadra

This is related to my soundness completeness thoughts earlier. Soundness is no bad builds, completeness is maximally (i.e. without breaking soundness) relaxed version constraints. With most phenomena baring these names, soundness is more important, but I agree it's less clear here since failing builds are easier to diagnosis than over-tight constraints.

But consider these two things:
 - The churn would a big problem but for allowing package maintainers to retroactively fix mistakes. Then only immediate consumer need to be edited, rather than also transitive consumers.
 - We've talked about automatically comparing interfaces to guard against obvious breakage. That same infrastructure could propose constraint relaxations.

These lead me to believe we might as well go for soundness cause we can, and the churn fallout can be ameliorated now, and moreso in the future when the growth of crates.io would make it worse.

Also, regardless of the severity of the churn issues, this is an issue that affects library consumers more than producers, so IMO library maintainers morally don't "deserve" more discretion on the matter.

jimmycuadra:
> Also, regardless of the severity of the churn issues, this is an issue that affects library consumers more than producers, so IMO library maintainers morally don't "deserve" more discretion on the matter.

Perhaps what you're saying here still applies to my case, but my primary motivation is not to make my job easier, it's to keep the plugin authors (the middle of a three-tier hierarchy) from having to make updates to otherwise working software in order for the end users (the bottom tier) to continue using the plugins as the main library evolves.

withoutboats:
I don't think allowing `0 < 12456789.0.0` is the same as allowing `*` because `*` is the easiest semantic versioning requirement to use; it seems to me like the obvious default for as long as you are the only one building your crate (i.e. when your crate is not on crates.io). Typing out a ridiculous version bound is much more explicitly disregarding versioning reliability, and is much more of an act that the library author has to decide to do. I don't have a strong opinion about whether or not to allow fictionally large upper bounds, but I don't think allowing that makes disallowing `*` pointless.

@jimmycuadra there is a big difference between not being able to use the newest version of a library and not being able to use it at all. It seems much less onorous to me to have to wait on your dependencies to be able to upgrade to a breaking new version of a library than to have a situation where your build is non-functional because one of your dependencies is making false statements about its transitive dependency versioning requirements.

jimmycuadra:
All of this discussion also assumes that SemVer is actually something you can rely on, which you most certainly cannot. People make breaking changes in minor and patch releases all the time, sometimes intentionally because they don't follow SemVer, or unintentionally because of a bug or because it's just really difficult to tell if you're breaking backwards compatibility in some way. Even with upper bounds enforced, there's no guarantee that a crate that built at one point will build again unless you lock every dependency down through the chain to an exact version number.

BurntSushi:
@jimmycuadra I don't see where that assumption is made. I don't think anyone is under the illusion that there is some iron clad guarantee of non-breakage in the absence of a tool that can automatically check semantic versioning for you. This is about doing what we can to mitigate breaking changes. A worthy cause, IMO!

text/0000-no-wildcard-deps.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Drawbacks
+
+The barrier to entry when publishing a crate will be mildly higher.
+
+In theory, there could be contexts where an unbounded version constraint is
+actually appropriate?
```
jmesmon: From the RFC above, `Bad version restrictions are also "viral".`, meaning it's possible for crates very far down the chain to have overly restrictive versions, leading to bloat in final programs due to unnecessarily including multiple versions.

huonw: It's totally possible for a crate to say `foo = ">= 1.0.0, < 1.0.9"`, when it should just be using `foo = "1.0.0"`, since it is really actually compatible with the whole `1.*.*` series (e,g, the author of `foo` followed semver), but it's definitely not clear when the correct version restriction is completely unrestricted: if you're using a dependency you presumably want to have some guarantees about what API it provides i.e. exactly what the semver approach to major versions is. Maybe there is such a case when having absolutely no guarantees about the API/behaviour of a dependency is useful (what this point is asking), but seems kinda unlikely.

jmesmon: My comment is merely stating that we probably want to add a note that this could cause size bloat in the drawbacks section.

erikjohnston:
I would be much happier with this if I could override the version constraints of transitive dependencies in my projects. Relying on upstream to bump their dependency versions sounds very annoying, and will lead to perfectly usable crates being forked repeatedly just so that people can bump the dependencies.

---

> All of this discussion also assumes that SemVer is actually something you can rely on, which you most certainly cannot. People make breaking changes in minor and patch releases all the time, sometimes intentionally because they don't follow SemVer, or unintentionally because of a bug or because it's just really difficult to tell if you're breaking backwards compatibility in some way. Even with upper bounds enforced, there's no guarantee that a crate that built at one point will build again unless you lock every dependency down through the chain to an exact version number.

I totally agree here. Unless a library explicitly specifies a specific version of a dependency there is no way a library author can make any guarantee that their code will work with all later minor/patch versions of the dependency.

It seems odd to me to suggest that library authors not pin versions of dependencies - presumably so that multiple crates depending on the same crate are more likely to be compatible - and yet suggest we force specifying an upper bound. Specifying an upper bound does *not* guarantee that the library is going to work with all versions anyway, so this seems like a) an arbitrary line in the sand and b) that this is a more generic problem.

---

FWIW, I (alas) spend most of my days working in python. The vast majority of my dependencies (and in other projects) don't specify an upper bound unless its known to break at a specific version. This has yet to cause much of an issue.

My understanding is that not specifying an upper bound is the recommended practice in most package systems, so I would be very interested to hear peoples experiences in these.

---

- This seems like a more generic problem: how do you deal with dependencies that break with newer/particular versions of specific libraries? (These could be API changes in major version bumps, accidental API changes in minor versions, bugs, etc.)
- Forcing an upper bound is going to be a pain in the long run if you can't override the version constraints of dependencies. Especially in the presence of unmaintained, but still usable, libraries.
- These problems are going to be especially noticeable now since we have a higher ratio of new unstable libraries vs old stable libraries.

sfackler:
@erikjohnston 

> It seems odd to me to suggest that library authors not pin versions of dependencies - presumably so that multiple crates depending on the same crate are more likely to be compatible - and yet suggest we force specifying an upper bound. Specifying an upper bound does not guarantee that the library is going to work with all versions anyway, so this seems like a) an arbitrary line in the sand and b) that this is a more generic problem.

It's not about guarantees - it's about finding a middle ground between being completely inflexible to the point of too-frequent breakage in dependency resolution and being completely flexible to the point of too-frequent breakage in compilation. Semver is the norm for crates.io packages - if an author breaks back-compat within the same major version, you can expect that people will be filing bugs on the project to get the release yanked, but there's no such norm for major version bumps.

The appropriate constraint depends on the dependency. For a library whose author provides strong guarantees about breaking change patterns, you can extend your version constraints much further than you'd be able to for a library with no such guarantees.

------

Thanks for the comments, everyone! I still feel that unbounded-from-above dependencies are a bit sketchy, but there are clearly compelling arguments that they work out in practice. How about we adjust the RFC to prohibit only true wildcard (`*`) dependencies. This will make library authors think to at least some degree about what versions of things they need and see how that goes. If we see the need to go further in the future, we'll probably want to pair the restriction with the addition of some tools to help with it, such as the ability to loosen constraints of published crates and the ability to force versions out of their constraint bounds on the consumer side.

Ericson2314:
@sfackler Limiting the proposal like that might build support now, but also just begs for a followup RFC, by which point more crates in violation could be published. What's wrong with adding all the restrictions we want (we can relax them later), along with something to ameliorate the concern about about churn?

tbu-:
In the current version the build can break if someone uses `"*"` you update your dependencies. This is clearly a bad thing. You can work around this by specifying a working version in your `Cargo.lock`.

In the proposed version, the upgrade of a library can be blocked by any crate in your dependency tree. This is also bad and **cannot** be worked around currently, so I don't think this RFC should be accepted until it addresses this, currently it just trades one issue for another, where the former has a work-around and the latter does not.

text/0000-no-wildcard-deps.md
```diff
@@ -0,0 +1,113 @@
...
+have a wildcard constraint on the version. Almost all of them them will fail
+to compile against version 0.7 of openssl when it is released. When that
+happens, users of those libraries will be forced to manually override Cargo's
+version selection every time it is recalculated. This is not a fun time.
+
+Bad version restrictions are also "viral". Even if a developer is careful to
+pick dependencies that have reasonable version restrictions, there could be a
+wildcard constraint hiding five transitive levels down.  Manually searching the
+entire dependency graph is an exercise in frustration that shouldn't be
+necessary.
```
tbu-: Bad version restrictions are also viral the other way around: There can always be a unneccessarily strict version constraint hiding "five transitive levels down". Manually searching the entire dependency graph is *not* a solution to this, as you'd have to publish your own versions of all the crates that recursively depend on this overly strict crate if the author does not update the version requirements by themselves – which means that the alternative this RFC is suggesting also carries this risk.

BurntSushi:
@tbu- I'm not sure how this RFC *trades* one issue for another. Even today, if a crate has an overly strict version bound, then you'll have the problem you described. This RFC does nothing to change that.

tbu-:
@BurntSushi Well, it pushes people away from the `"*"` problem which *has* a work-around to stricter version bounds which *don't have* a work-around currently. Once there is a way to get around a crate's overly restrictive bounds, this RFC will pose no problem.

daboross:
@tbu- I don't really think `Cargo.lock` is really a workaround, due to the fact that this RFC would *only* affect libraries on crates.io, and it's currently idiomatic to *not* commit Cargo.lock to a version repository at all for library crates.

I mean, `*` will still work in binary crates and local crates regardless of whether this RFC happens or not, and - idiomatically - `Cargo.lock` is only used in those two scenarios (binary crates and local crates).

text/0000-no-wildcard-deps.md
huonw: NB. this exact point is made above https://github.com/rust-lang/rfcs/pull/1241#discussion_r38805850 .

tbu-:
@daboross It's a work around in the way that that root node of the dependency tree can decide about all library's version, including library's dependencies. `Cargo.lock` used by binary crates also works for the library crates that the binary crate transitively depends on.

But as you say, all libraries would have to rely on the binary crate actually using them maintains a proper `Cargo.lock`, which isn't a nice thing either... I mean I really like this RFC, but we don't have override mechanisms for cases in which an author is not available. (As another example of this: Firefox regulary updates. Firefox addons usually include the next version as compatible as well, just because the author can't immediately release the next version when the next Firefox comes out. Firefox does come with an override tool (Compatiblity Reporter), so it's not that bad if you know about that tool.)

Ericson2314:
The current situation is not great, but I see @tbu-'s point in that sloppy versions are a way to work around the upgrade churn issue.

aturon:
cc @wycats

sfackler:
Updated to only block wildcard deps, and roll out with a warning only for one release cycle.

alexcrichton:
This libs team discussed this RFC today, and decided that we were in favor but in light of the recent change by @sfackler we're going to re-up this for another week-long FCP.

tbu-:
I appreciate the change, thank you!

jimmycuadra:
I have no problems with this proposal now that it has been changed to target only wildcards.

text/0000-no-wildcard-deps.md
```diff
@@ -0,0 +1,131 @@
...
+Absurd guarantees on their own are not necessarily sufficient motivation to
+make a change like this. The real motivation is the effect that these
+guarantees have on consumers of libraries.
+
+As an example, consider the [openssl](https://crates.io/crates/openssl) crate.
+It is one of the most popular libraries on crates.io, with several hundred
+downloads every day. 50% of the [libraries that depend on it](https://crates.io/crates/openssl/reverse_dependencies)
+have a wildcard constraint on the version. None of them can build against every
+version that has ever been released. Indeed, no libraries can since many of
+those releases can before Rust 1.0 released. In addition, almost all of them
```
daboross: I think `came` is meant here instead of `can` twice

wycats:
I've been thinking about this a lot, and I have a few thoughts.

I'm in favor of forbidding wildcard dependencies, as the current draft of the RFC proposes.

Version ranges make me uneasy, because they make it much more likely that a resolved dependency graph will still fail to compile, which is something I would like to eventually rule out altogether. I am not suggesting any changes to the current draft of this RFC related to version ranges, but I'll explore this topic further.

#### Shared Types

The core tension here is about shared types. To give a concrete example, let's say that `format-time` and `enriched-time` both depend on a `time` crate.

```rust
// format-time
extern crate time;
pub fn format_time<T: time::Time>(time: T) -> String {
  // format the time
}
```

```rust
// enriched-time
extern crate time;

pub struct EnrichedTime { /* ... */ }
impl time::Time for EnrichedTime { /* ... */ }
```

If a parent crate depends on both `format-time` and `enriched-time`, and tries to pass an instance of `EnrichedTime` to `format_time`, they better both be talking about the same `time` crate.

#### Cargo's Coarse-Grained Solution

Today's Cargo attempts to address this issue coarsely:

First, Cargo will automatically coalesce dependencies within the same major version into a single shared dependency across the entire graph. This means that if `format-time` depends on `time 1.6` and `enriched-time` depends on `time 1.7`, they will still see the identical `time::Time` trait.

Second, Cargo assumes compliance with semantic versioning to make this work. This assumption is fundamental: if Cargo was not allowed to assume that `time 1.7` was compatible with `time 1.6`, `format-time` and `enriched-time` (and any other crate with a shared dependency on `time`) would be forced to coordinate version bumping.

Third, depending on a range of versions (`time 1.x to time 3.x`) means that you are depending on a subset of the types in that range that are compatible across those versions, but without any enforcement.

If you are sharing those types in your public interface, or passing instances of those types to other dependencies, this creates a "matrix" problem: the only way to be confident that your crate will compile for all downstream crates is to build your crate will all combinations of upstream versions.

#### Shared vs. Internal Dependencies

One of the weaknesses of the current model is that it doesn't distinguish between dependencies used purely internally, as part of your implementation and dependencies exposed as part of your public interface.

Internal-only dependencies have a much higher-tolerance for using version ranges to select a subset of a crate's types that are compatible across major versions, because the code with the internal-only dependency knows exactly how the type is being used.

For example, even if a particular method on `time::Time` has changed between `1.x and 2.x`, if your code doesn't use it, it is still safe to depend on the `1.x to 2.x` range.

In contrast, dependencies exposed as part of your public API by definition do not know all of the ways those types are being used by downstream crates. As a result, it can only use version ranges if all of the types exposed in their public interface are compatible across those versions.

#### The Role of Small Crates and Semantic Versioning

Instead of trying to teach all crate authors the above rules, which are devilishly hard to get right, Cargo has so far relied on relatively small crates and semantic versioning to allow crate authors to describe compatibility for a small cluster of types.

Instead of asking `format-time` and `enriched-time` to coordinate a version bump of the `time` dependency across every "tiny" version bump, they are required to coordinate version bumps across major versions only.

The fact that people want sometimes to express their dependencies as ranges implies two things to me:

1. There are times where a subset of all of the types in a crate are compatible across major versions, and a crate wants to depend on that subset.
2. People are often using a crate as an internal-only dependency, where they only have to worry about finding a compatible subset of functions, not entire types.

#### A Proposal

I think we can address this problem with a small improvement:

Add a new `matrix-build` command to Cargo, which will build your crate against all combinations of (major versions of) upstream dependencies, so that you can figure out before publishing whether your version ranges are as compatible as you think they are.

#### Version Ranges Are Problematic

That said, I think that depending on version ranges for crates that you use as part of your public interface is a smell.

It assumes that there is a subset of a package's types and functions that are stable across major versions.

For example, if the `time` crate exposes both a relatively stable `time::Time` trait and more unstable utility methods, people may find themselves depending on the range of `time` versions that describe an implicit version of the `time::Time` trait.

If the `time` crates bumps its major version, all crates that depend on a range of `time` will need to check whether `time::Time` is still compatible, and bump the range. Of course, many crates in practice will forget to do this, meaning that perfectly compatible combinations of crates that share `time::Time` as a public dependency will no longer compile.

In this case, it behooves the author of the `time` crate to break out the more stable `time::Time` trait (`time-traits` and `time-utils`) and version it explicitly. This allows downstream dependencies that rely on `time-traits` to maintain compatibility with each other even when `time-utils` changes.

**I see dependency ranges as an interim workaround, rather than a long-term solution.**

#### Matrix Builds

We can reduce the risk of dependency hell caused by version ranges through `cargo matrix-build`. I'd like to make `matrix-build` a sanity check done by `cargo publish` (as `cargo build` is today), to ensure that combinations of upstream dependencies will compile downstream.

However, that can only work if dependency ranges are used as a last resort, and crate authors are encouraged to break out more stable interfaces from less stable implementation.

The simple advice amounts to:

> If you're a crate that exports widely shared types or interfaces, you want to create a stable crate (in the same repository) that exposes just a set of stable traits that downstream dependencies can rely on.

**This doesn't mean that everyone needs to preemptively create interface crates just in case.** It just means that once a crate discovers that downstream crates are often depending on version ranges, the "right thing to do" is to break out the more stable interfaces from the less stable implementation. (if you're building a crate you intend to be widely shared, maybe you should do it preemptively.)

alexcrichton:
The library subteam discussed this RFC yesterday and the conclusion was to accept it, thanks again @sfackler!

PR #1270, opened 2015-09-05
RFC: #[deprecated] for Everyone

llogiq:
Library authors need a way of evolving their API painlessly. This is it.

[Rendered](https://github.com/llogiq/rfcs/blob/deprecate/text/0000-deprecation.md)

### DIFF
--- a/text/0000-deprecation.md
+++ b/text/0000-deprecation.md
@@ -40,13 +40,14 @@ strongly advised to make use of it to convey the reason for the deprecation to
 users of their library. The string is interpreted as plain unformatted text 
 (for now) so that rustdoc can include it in the item's documentation without 
 messing up the formatting.
-* `use`, if included, must be the import path (or a comma-separated list of 
+* `use`, if included, must be the import path (or a semicolon-delimited list of 
 paths) to a set of API items that will replace the functionality of the 
 deprecated item. All crates in scope can be reached by this path. E.g. let's 
 say my `foo()` item was superceded by either the `bar()` or `baz()` functions
-in the `bar` crate, I can `#[deprecate(use="bar::{bar,baz}")] foo()`, as long 
-as I have the `bar` crate in the library path. Rustc checks if the item is 
-actually available, otherwise returning an error.
+in the `bar` crate, in conjunction with the `bruzz(_)` function in the `baz`
+crate, I can `#[deprecate(use="bar::{bar,baz};baz::bruzz")] foo()`, as long 
+as I have the `bar` and `baz` crates in the library path. Rustc checks if the 
+item is actually available, otherwise returning an error.
 
 On use of a *deprecated* item, `rustc` will `warn` of the deprecation. Note 
 that during Cargo builds, warnings on dependencies get silenced. Note that 
--- a/text/0000-deprecation.md
+++ b/text/0000-deprecation.md
@@ -9,7 +9,7 @@ This RFC proposes to allow library authors to use a `#[deprecated]` attribute,
 with optional `since = "`*version*`"`, `reason = "`*free text*`"` and 
 `use = "`*substitute declaration*`"` fields. The compiler can then
 warn on deprecated items, while `rustdoc` can document their deprecation
-accordingly. 
+accordingly.
 
 # Motivation
 
@@ -29,29 +29,26 @@ Public API items (both plain `fn`s, methods, trait- and inherent
 fields and enum variants) can be given a `#[deprecated]` attribute. All
 possible fields are optional:
 
-* `since` is defined to contain the exact version of the crate that
-deprecated the item, as defined by Cargo.toml (thus following the semver
-scheme). It makes no sense to put a version number higher than the current
-newest version here, and this is not checked (but could be by external
-lints, e.g. [rust-clippy](https://github.com/Manishearth/rust-clippy).
-To maximize usefulness, the version should be fully specified (e.g. no
-wildcards or ranges).
+* `since` is defined to contain the version of the crate at the time of
+deprecating the item, following the semver scheme. It makes no sense to put a 
+version number higher than the current newest version here, and this is not 
+checked (but could be by external lints, e.g. 
+[rust-clippy](https://github.com/Manishearth/rust-clippy).
 * `reason` should contain a human-readable string outlining the reason for
 deprecating the item. While this field is not required, library authors are
-strongly advised to make use of it to convey the reason to users of their
-library. The string is required to be plain unformatted text (for now) so that
-rustdoc can include it in the item's documentation without messing up the 
-formatting.
-* `use` should be the full path to an API item that will replace the 
-functionality of the deprecated item, optionally (if the replacement is in a 
-different crate) followed by `@` and either a crate name (so that 
-`https://crates.io/crates/` followed by the name is a live link) or the URL to 
-a repository or other location where a surrogate can be obtained. Links must be 
-plain FTP, FTPS, HTTP or HTTPS links. The intention is to allow rustdoc (and
-possibly other tools in the future, e.g. IDEs) to act on the included 
-information. The `use` field can have multiple values.
+strongly advised to make use of it to convey the reason for the deprecation to 
+users of their library. The string is interpreted as plain unformatted text 
+(for now) so that rustdoc can include it in the item's documentation without 
+messing up the formatting.
+* `use`, if included, must be the import path (or a comma-separated list of 
+paths) to a set of API items that will replace the functionality of the 
+deprecated item. All crates in scope can be reached by this path. E.g. let's 
+say my `foo()` item was superceded by either the `bar()` or `baz()` functions
+in the `bar` crate, I can `#[deprecate(use="bar::{bar,baz}")] foo()`, as long 
+as I have the `bar` crate in the library path. Rustc checks if the item is 
+actually available, otherwise returning an error.
 
-On use of a *deprecated* item, `rustc` should `warn` of the deprecation. Note 
+On use of a *deprecated* item, `rustc` will `warn` of the deprecation. Note 
 that during Cargo builds, warnings on dependencies get silenced. Note that 
 while this has the upside of keeping things tidy, it has a downside when it 
 comes to deprecation:
@@ -62,11 +59,11 @@ try to build `foobar` directly. We may want to create a service like `crater`
 to warn on use of deprecated items in library crates, however this is outside
 the scope of this RFC.
 
-`rustdoc` should show deprecation on items, with a `[deprecated since x.y.z]`
-box that may optionally show the reason and/or link to the replacement if
-available.
+`rustdoc` will show deprecation on items, with a `[deprecated]`
+box that may optionally show the version, reason and/or link to the replacement 
+if available.
 
-The language reference should be extended to describe this feature as outlined
+The language reference will be extended to describe this feature as outlined
 in this RFC. Authors shall be advised to leave their users enough time to react
 before *removing* a deprecated item.
 
@@ -83,7 +80,7 @@ quite complex.
 
 * Do nothing
 * make the `since` field required and check that it's a single version
-* Optionally the deprecation lint chould check the current version as set by
+* Optionally the deprecation lint could check the current version as set by
 cargo in the CARGO_CRATE_VERSION environment variable (the rust build process 
 should set this environment variable, too). This would allow future 
 deprecations to be shown in the docs early, but not warned against by the
@@ -93,13 +90,12 @@ be `Allow` by default)
 * `reason` could include markdown formatting
 * The `use` could simply be plain text, which would remove much of the
 complexity here
-* The `use` field contents could make use of the context in finding
-replacements, e.g. extern crates, so that `time::precise_time_ns` would resolve
-to the `time::precise_time_ns` API in the `time` crate, provided an
-`extern crate time;` declaration is present
 * The `use` field could be left out and added later. However, this would
 lead people to describe a replacement in the `reason` field, as is already
 happening in the case of rustc-private deprecation
+* Optionally, `cargo` could offer a new dependency category: "doc-dependencies"
+which are used to pull in other crates' documentations to link them (this is
+obviously not only relevant to deprecation).
 
 # Unresolved questions
 
--- a/text/0000-deprecation.md
+++ b/text/0000-deprecation.md
@@ -6,10 +6,9 @@
 # Summary
 
 This RFC proposes to allow library authors to use a `#[deprecated]` attribute,
-with optional `since = "`*version*`"`, `reason = "`*free text*`"` and 
-`use = "`*substitute declaration*`"` fields. The compiler can then
-warn on deprecated items, while `rustdoc` can document their deprecation
-accordingly.
+with optional `since = "`*version*`"` and `reason = "`*free text*`"`fields. The 
+compiler can then warn on deprecated items, while `rustdoc` can document their 
+deprecation accordingly.
 
 # Motivation
 
@@ -30,29 +29,20 @@ fields and enum variants) can be given a `#[deprecated]` attribute. All
 possible fields are optional:
 
 * `since` is defined to contain the version of the crate at the time of
-deprecating the item, following the semver scheme. It makes no sense to put a 
-version number higher than the current newest version here, and this is not 
-checked (but could be by external lints, e.g. 
-[rust-clippy](https://github.com/Manishearth/rust-clippy).
+deprecating the item, following the semver scheme. Rustc does not know about
+versions, thus the content of this field is not checked (but will be by external
+lints, e.g. [rust-clippy](https://github.com/Manishearth/rust-clippy).
 * `reason` should contain a human-readable string outlining the reason for
 deprecating the item. While this field is not required, library authors are
 strongly advised to make use of it to convey the reason for the deprecation to 
 users of their library. The string is interpreted as plain unformatted text 
 (for now) so that rustdoc can include it in the item's documentation without 
 messing up the formatting.
-* `use`, if included, must be the import path (or a semicolon-delimited list of 
-paths) to a set of API items that will replace the functionality of the 
-deprecated item. All crates in scope can be reached by this path. E.g. let's 
-say my `foo()` item was superceded by either the `bar()` or `baz()` functions
-in the `bar` crate, in conjunction with the `bruzz(_)` function in the `baz`
-crate, I can `#[deprecate(use="bar::{bar,baz};baz::bruzz")] foo()`, as long 
-as I have the `bar` and `baz` crates in the library path. Rustc checks if the 
-item is actually available, otherwise returning an error.
 
 On use of a *deprecated* item, `rustc` will `warn` of the deprecation. Note 
-that during Cargo builds, warnings on dependencies get silenced. Note that 
-while this has the upside of keeping things tidy, it has a downside when it 
-comes to deprecation:
+that during Cargo builds, warnings on dependencies get silenced. While this has 
+the upside of keeping things tidy, it has a downside when it comes to 
+deprecation:
 
 Let's say I have my `llogiq` crate that depends on `foobar` which uses a
 deprecated item of `serde`. I will never get the warning about this unless I
@@ -60,9 +50,8 @@ try to build `foobar` directly. We may want to create a service like `crater`
 to warn on use of deprecated items in library crates, however this is outside
 the scope of this RFC.
 
-`rustdoc` will show deprecation on items, with a `[deprecated]`
-box that may optionally show the version, reason and/or link to the replacement 
-if available.
+`rustdoc` will show deprecation on items, with a `[deprecated]` box that may 
+optionally show the version and reason where available.
 
 The language reference will be extended to describe this feature as outlined
 in this RFC. Authors shall be advised to leave their users enough time to react
@@ -71,38 +60,69 @@ before *removing* a deprecated item.
 The internally used feature can either be subsumed by this or possibly renamed
 to avoid a name clash.
 
+# Intended Use
+
+Crate author Anna wants to evolve her crate's API. She has found that one
+type, `Foo`, has a better implementation in the `rust-foo` crate. Also she has
+written a `frob(Foo)` function to replace the earlier `Foo::frobnicate(self)`
+method. 
+
+So Anna first bumps the version of her crate (because deprecation is always
+done on a version change) from `0.1.1` to `0.2.1`. She also adds the following 
+prefix to the `Foo` type:
+
+```
+extern crate rust_foo;
+
+#[deprecated(since = "0.2.1", use="rust_foo::Foo", 
+    reason="The rust_foo version is more advanced, and this crates' will likely be discontinued")]
+struct Foo { .. }
+```
+
+Users of her crate will see the following once they `cargo update` and `build`:
+
+```
+src/foo_use.rs:27:5: 27:8 warning: Foo is marked deprecated as of version 0.2.1
+src/foo_use.rs:27:5: 27:8 note: The rust_foo version is more advanced, and this crates' will likely be discontinued
+```
+
+Rust-clippy will likely gain more sophisticated checks for deprecation:
+
+* `future_deprecation` will warn on items marked as deprecated, but with a
+version lower than their crates', while `current_deprecation` will warn only on
+those items marked as deprecated where the version is equal or lower to the
+crates' one.
+* `deprecation_syntax` will check that the `since` field really contains a
+semver number and not some random string.
+
+Clippy users can then activate the clippy checks and deactivate the standard
+deprecation checks.
+
 # Drawbacks
 
-* The required checks for the `since` and `use` fields are potentially
-quite complex.
 * Once the feature is public, we can no longer change its design
 
 # Alternatives
 
 * Do nothing
 * make the `since` field required and check that it's a single version
-* Optionally the deprecation lint could check the current version as set by
-cargo in the CARGO_CRATE_VERSION environment variable (the rust build process 
-should set this environment variable, too). This would allow future 
-deprecations to be shown in the docs early, but not warned against by the
-stability lint (there could however be a `future-deprecation` lint that should
-be `Allow` by default)
 * require either `reason` or `use` be present
 * `reason` could include markdown formatting
-* The `use` could simply be plain text, which would remove much of the
-complexity here
-* The `use` field could be left out and added later. However, this would
-lead people to describe a replacement in the `reason` field, as is already
-happening in the case of rustc-private deprecation
+* rename the `reason` field to `note` to clarify it's broader usage.
+* add a `note` field and make `reason` a field with specific meaning, perhaps
+even predefine a number of valid reason strings, as JEP277 currently does
+* Add a `use` field containing a plain text of what to use instead
+* Add a `use` field containing a path to some function, type, etc. to replace
+the current feature. Currently with the rustc-private feature, people are 
+describing a replacement in the `reason` field, which is clearly not the 
+original intention of the field
 * Optionally, `cargo` could offer a new dependency category: "doc-dependencies"
 which are used to pull in other crates' documentations to link them (this is
-obviously not only relevant to deprecation).
+obviously not only relevant to deprecation)
 
 # Unresolved questions
 
 * What other restrictions should we introduce now to avoid being bound to a 
 possibly flawed design?
-* How should the multiple values in the `use` field work? Just split by
-comma or some other delimiter?
 * Can / Should the `std` library make use of the `#[deprecated]` extensions?
 * Bikeshedding: Are the names good enough?
--- a/text/0000-deprecation.md
+++ b/text/0000-deprecation.md
@@ -17,17 +17,13 @@ Namely the following items:
    code paths in the compiler.
 2. Add an optional `rust = "..."` package attribute to Cargo.toml, 
    which `cargo new` pre-fills with the current rust version.
-3. Allow `std` APIs to declare an 
-   `#[insecure(level="Warn", reason="...")]`  attribute that will 
-   produce a warning or error, depending on level, that cannot be 
-   switched off (even with `-Awarning`)
-4. Add a `removed_at="..."` item to `#[deprecated]` attributes that 
+3. Add a `removed_at="..."` item to `#[deprecated]` attributes that 
    allows making API items unavailable starting from certain target 
    versions.
-5. Add a number of warnings to steer users in the direction of using 
+4. Add a number of warnings to steer users in the direction of using 
    the most recent Rust version that makes sense to them, while making 
    it easy for library writers to support a wide range of Rust versions
-6. (optional) add a `legacy="..."` item to `#[deprecated]` attributes
+5. (optional) add a `legacy="..."` item to `#[deprecated]` attributes
    that allows grouping API items under a legacy flag that is already 
    defined in RFC #1122 (see below)
 
@@ -45,7 +41,7 @@ As a background, in no particular order:
 In addition, there has been an ongoing 
 [discussion on internals](https://internals.rust-lang.org/t/thoughts-on-aggressive-deprecation-in-libstd/2176/55) 
 about how we are going to evolve the standard library, which this
-proposal is mostly based on.
+proposal is somewhat based on.
 
 Finally, the recent discussion on the first breaking change 
 ([RFC PR #1156 Adjust default object bounds](https://github.com/rust-lang/rfcs/pull/1156))
@@ -63,7 +59,7 @@ the `-Awarnings` argument.
 
 ## 1. Language / `std` Evolution
 
-The following motivates items 1 and 2 (and to a lesser extent 5)
+The following motivates items 1 and 2 (and to a lesser extent 4)
 
 With the current setup, we can already evolve the language and APIs,
 albeit in a very limited way. For example, it is virtually impossible
@@ -89,7 +85,7 @@ libraries, in which deprecation doesn't exist!
 
 ## 2. User Experience
 
-The following motivates items 2, 4 and 5.
+The following motivates items 2, 3 and 4.
 
 Currently, there is no way to make an API item unavailable via 
 deprecation. This means the API will only ever expand, with a lot of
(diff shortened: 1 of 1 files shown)


llogiq:
[Rendered](https://github.com/llogiq/rfcs/blob/deprecate/text/0000-deprecation.md)

seanmonstar:
I have never found value in a 'since' label for deprecations. *How long* an
item has been deprecated isn't too useful to know. It *would* be useful to
know when a deprecated item will be completely removed. Though, I wouldn't
want that to be a requirement, either.

Also, I would find it useful to be able to mark structs, enums, modules,
and even public fields as deprecated as well.

sfackler:
The `since` label would be useful for tooling, though. If you have a crate that has a wide version constraint on a dependency, it may be using a deprecated method to ensure that it works with older versions of the crate so it might not make sense for rustc to warn about deprecated items that were deprecated in the middle of the supported version range.

llogiq:
@seanmonstar Good point about `since`; I mainly added it as a proxy for when the item will be removed. Since the latter is in the future, it would probably be wrong anyway, and I would mostly expect items that have been deprecated in earlier versions to be removed earlier, too. Perhaps we should make `since` optional and require either `reason` or `surrogate` be present.

And yes, I see types and public fields as API items, too. I forgot that `item` is a reserved word in the Rust compiler. How do I call it to avoid this misunderstanding?

olivren:
A `deprecate` annotation is very desirable, yes.

It seems odd to me that the `since` field can accept a version range. I would expect it to be a precise version. I don't even understand what it would mean for a function to be deprecated since version `3.*`.

I think it is fine to make the `since` field mandatory, because even if it is not a super useful information, its content is trivial and unambiguous to write. A library author who cares about annotating deprecated items in the first place will not be annoyed by typing the next version number of his library. Also, this information can give a sense of when the item will be removed, in particular when the project has an official deprecation policy.

I would not introduce an hypothetical `until` field (the version the item will be removed). If the project has a deprecation policy, then this information can be derived from the `since` field itself. If it doesn't have one, I don't see how the library author would be able to give a value to this field. In any case the content of this field would be unreliable.

I like the idea of the `surrogate` field, it makes it possible to generate a clickable link in the rustdoc. Sometimes, a deprecated item can be replaced by more than one item, either because there are multiple valid alternatives, or because the deprecated item should be replaced by a combination of other items. I propose that surrogate can contain multiple values.

llogiq:
@olivren Thank you; those are good suggestions. I figure I was a bit inconsistent in my thinking; It *is* a proxy for the probability of the item being removed, with the benefit that the removal is in the future, and thus hard to predict, whereas the deprecation is in the present/past, and thus doesn't require a chrystal ball.

bluss:
The staging_api attribute is called `deprecated` (with a d). Your attribute is that one, right? Or are you proposing a new and parallel attribute?

llogiq:
Well, I initially skipped the final `d` to allow for parallel implementations (because the semantics are a tiny bit different), but that probably doesn't make sense. I'm changing it to `#[deprecated]` everywhere to reduce confusion.

bluss:
Best to reduce confusion is just to clearly say which relation it has to the existing `deprecated`. It makes sense to use the same attribute as staging api does (to me at least), and it means that `deprecated` will move from being a staging-api feature to being a separate one.

steveklabnik:
But they have different needs, which is why they're different. Hence the different attribute.

llogiq:
@bluss Ok, I initially wrote the RFC in relation to the existing feature. huonw and brson convinced me to rewrite it as if the feature didn't exist.

@steveklabnik we can bikeshed the name later; for now I want to get the restrictions right. Also I don't think that they have different needs – in fact the initial draft of this included `#[unstable]`.

bluss:
@steveklabnik  The RFC needs to be explicit that this is or is not related to staging-api. Pretending `deprecated` does not already exist is just confusing.

steveklabnik:
It should be explicit, absolutely, but I don't think that it should necessarily just be a port of `deprecated` or remove it.

text/0000-deprecation.md
```diff
@@ -0,0 +1,105 @@
+- Feature Name: Public Stability
+- Start Date: 2015-09-03
+- RFC PR: 
+- Rust Issue: 
+
+# Summary
+
+This RFC proposes to allow library authors to use a `#[deprecated]` attribute,
+with `since="`(version)`"`, `reason="`(free text)`"` and 
+`surrogate="`(text or surrogate declaration)`"` fields. The compiler can then
```
chris-morgan: <code>&lt;code&gt;since = "\*version\*"&lt;/code&gt;</code> produces <code>since = "*version*"</code> which typically looks better. (Also note the whitespace around the equals sign, which is conventional.)

text/0000-deprecation.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Detailed design
+
+Public API items (both plain `fn`s, methods, trait- and inherent 
+`impl`ementations as well as `const` definitions, type definitions, struct
+fields and enum variants) can be given a `#[deprecated]` attribute.
+
+This attribute *must* have the `since` field, which contains the exact
+version of the crate that deprecated the item, as defined by Cargo.toml 
+(thus following the semver scheme). It makes no sense to put a version number
```
huonw: I think this field should just be opaque text. What benefit do we get from validating semver correctness (or whatever) at the language level? It would force rustc to understand cargo's behaviour and read Cargo.toml's or whatever.

Ericson2314: Good point. But I'd like to ensure from the get-go that text does correspond to an already-published version when pushing to crates.io.

huonw:
The question of being a different attribute or not is a little misguided, the internal one isn't exposed and hence it is fluid (and mostly irrelevant, it sets the scene/providing context/suggesting ideas that work/don't work, but "just do what the compiler does" doesn't work well as a longer-term reference document: this is what I meant when I encouraged the RFC to be more independent from the internal system). The external one should chose the best name/functionality, and we can tweak the internal one as necessary to ensure the external one can do everything it wants.

---

I think one of the most important features for an external `#[deprecated]` is being able to be precise about ignoring deprecation warnings. I don't see any mention of silencing the warning in the current RFC (i.e. like the current `allow(deprecated)`) but I think it is required, and I think it needs to not be global like `allow(deprecated)`. For example, a developer could control a group of several crates, which use deprecated functionality from each other (perfectly OK: the developer has full control) and so will want to silence the warnings of deprecated items in those crates, but presumably will want to know about deprecated things in external dependencies. I.e. there should be some attributes or something that mean that this works (once those attributes have been inserted):

```rust
extern internal_crate;
extern external_crate;

fn main() {
    internal_crate::deprecated(); // no warning
    external_crate::deprecated(); // warn: deprecated item
}
```

I think `std` also could benefit from this, but it isn't quite as important because it is much more restricted. External crates are unrestricted by design.

Havvy:
Alternative: Leave out `surrogate` until it is better defined. The reason field can accept a surrogate in plain text in the mean time.

text/0000-deprecation.md
llogiq: @huonw I see where you're coming from, but if we don't validate the contents, we don't need to require the field at all. Unless we have a valid version, there's nothing we can do with the information. If we have one, however, we can act on it. Perhaps an IDE could propose pinning the version (or one before it) in `Cargo.toml` or something like that.

huonw: An external tool (e.g. cargo) can still work with them, handling mistakes/emitting errors appropriately. I really don't think it is the language's place to be opinionated about what a valid version is.

llogiq:
@huonw That's an interesting idea, and certainly one we should explore here. My preliminary idea:

* deprecation is a lint (and thus can be set to allow/warn/deny
* we want more fine-grained control however; e.g. warn on deprecation from crate `foo`, but not `bar` (I think the crate level should be sufficient here)
* since the lint settings don't allow for additional information, we need a side channel to declare that we want to ignore deprecation in certain crates, e.g. a `#![deprecation(on="bar", on="baz", level="allow")]` attribute. The lint could look for this attribute via parenting and omit ignored warnings.

bluss:
@huonw I just asked for clarification, guess it's just my confusion.

llogiq:
@bluss I don't want people to be confused. :smile: I'll try and see if I can make the relation even more clear.

eddyb:
@llogiq Parametric lints (e.g. `#[allow(deprecation(bar))]`) seem like a long-term win IMHO, I'd rather have that than yet another attribute..

llogiq:
@eddyb Full ack. I've wanted to be able to configure lints for some time, resorting to splitting lints into different parts so that each part can be activated independently. It works, but it's more complex than needed.

Do we have parametric lints on the roadmap or is someone going to write the RFC for it?

gkoz:
I have a somewhat exotic use case, that might clash with the `since` field requirement.

Imagine a crate that implements bindings to a third party library (e.g. GTK+ 3). There are various versions of that library in the wild the crate is expected to work with. The library versions might range from 3.4 to 3.16 while the crate is in the 0.1.x territory.

The library API is evolving of course. Suppose a user targets the library version 3.12 and calls a function deprecated in 3.8. The crate needs to communicate this somehow, but its own version number is immaterial in this case.

On the question of warnings control the user would ideally want module-, possibly even item-level granularity, because not all deprecations are created equal.

seanmonstar:
If I didn't explicitly state it in my first comment, I'd rather if `since` was not required.

llogiq:
Ok, folks, this is all great feedback. Perhaps plain `#[deprecated]` isn't so bad after all – it clearly states intent (though it leaves the "when" and "why" completely open).

@gkoz: I'm not sure I want to cater to that use case for now. A possible solution would be to deprecate the item for *all* versions that target gtk versions with a deprecation, on the grounds that a user could update the gtk version behind our backs. As long as we have control over what gets warned about, I don't see a big problem with this.

In the meantime, I'd like to better define `surrogate`s. I'd like it to be easy, so we should perhaps be able to name `use`d APIs directly – basically everything in the current namespace. However, I don't want to require `extern crate` definitions for inter-crate surrogates. What do you folks think?

pornel:
I like this feature.

I would prefer `since` to be optional. It's nice when it's there, but I don't see a reason to make it required (even without it it's a useful information). In some situations functions are not deprecated because they're going to be removed, but because new, better, faster, higher fidelity alternatives are available, but the old one will never be removed (e.g. Web's DOM APIs).

Having `surrogate`s is very useful, so it's good to have a field to encourage it. However, the name is strange to me (I'm not a native English speaker though). In UIs it's usually displayed as "See also" or "Use X instead", so maybe `see_also`, `use_instead`, or `replaced_with`, etc.

llogiq:
@pornel Thanks for the feedback. I just want to note that the `since` field does not state when or why it's going to be removed, just since which version the API was actually deprecated. My idea to make it a required field was informed by current usage of a similar feature in rustc as well as in the Java standard library as well as some other popular Java libraries (which however use `@deprecated` items in javadoc in combination with `@Deprecated` annotations). Also if we had a semver number, we could instruct cargo to pin dependency versions to before the deprecation, or report how many warnings we'd get with which versions, etc. The possibilities are limitless :smile:

Thus, seeing that the cost of writing out the current crate version is quite low (and once IDEs auto-generate annotations, becomes even lower), I erred on the side of conservativeness: If we require the field now, we can make it optional later. The reverse direction is not possible. However, given the current feedback, I'm considering making the field optional and unchecked.

As for the *why*, that's what the `reason` is for. I noticed that its current use is actually overloaded; many `reason`s state which API to use instead. Regarding the name of `surrogate`, I'm open to bikeshedding discussions. I wanted to have a proper noun instead of `instead`, and "surrogate" is two characters shorter than "replacement". I'd like to add that this is part of an annotation, not a docstring. `see also` is something I'd expect in the latter.

glaebhoerl:
I'm essentially a native speaker, and "surrogate" is too jargony for my tastes as well. If I didn't have context, I probably couldn't have guessed what it was supposed to refer to. "Replacement" is much clearer, at least. (I don't think character count is the most important thing to optimize for here, to put it mildly.)

llogiq:
@glaebhoerl: Ok, let's put our bikeshedding-goggles on. How about:

* replacement (:+1: clear and to the point :-1: long)
* ersatz (a word from German I'm reliably informed has been assimilated into English :+1: short :-1: perhaps not universally known)
* alternative (could be a noun)
* alternate
* alt (short for alternative? Its use on usenet has given it some subtext... :+1: very short, :-1: needs explanation)
* instead (:-1: not a noun)
* replace_by (:+1: a direct call to action, :-1: can we do without underscore?)
* use_instead (same as replace_by, only one more character)
* by (:+1: short as hell, :-1: unclear; one could think it referred to the person)

gkoz:
"surrogate", "ersatz" imply that the replacement is inferior to the deprecated function. "alternative" feels confusing too because they're probably not equal alternatives (one is deprecated for a reason).

pornel:
@llogiq crates.io has history of all releases (and won't let you delete anything), so it can automatically find accurate 'since' information without any annotations from developers.

llogiq:
True that. The only downside is reduced locality (e.g. some people don't have fast internet connection), and that rustdoc cannot show the version at which it was deprecated without a value (unless we want rustdoc to hammer crates.io with requests).

You folks convinced me that the `since` field is not actually required, and should not be. I'll rewrite that part of the spec to make the field optional and unchecked (but if present, it should represent a crate version, as some tools may or may not use it in the future). I still think it should be available as an optional field, though.

briansmith:
> `use` should be the full path to an API item that will replace the functionality of the deprecated item.

Please specify this syntax or link to the specification of the syntax. Also, I suggest that it should be allowed to link to an API item in the same module using just its unqualified name. It will often be the case that the replacement(s) is/are in the same module.

Also, sometimes it is useful to specify the replacement as a code fragment.

Finally, sometimes there is no clear replacement. For example, if an implementation of a crypto algorithm is being removed from a library because the algorithm itself is deprecated, then there will be no replacement function that makes sense, necessarily. So, I think `use` should definitely be optional.

> (so that https://crates.io/crates/ followed by the name is a live link)

NIT: Not necessarily, if the crate isn't on crates.io.

> Links must be plain FTP, FTPS,

Please no FTP. Also, very few clients can even do FTPS. For example, browsers can't.

> The use field can have multiple values.

Please clarify the syntax of the use field with multiple values and provide an example of such a multi-valued `use` in the documentation.

Otherwise, LGTM.

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+# Detailed design
+
+Public API items (both plain `fn`s, methods, trait- and inherent 
+`impl`ementations as well as `const` definitions, type definitions, struct
+fields and enum variants) can be given a `#[deprecated]` attribute. All
+possible fields are optional:
+
+* `since` is defined to contain the exact version of the crate that
+deprecated the item, as defined by Cargo.toml (thus following the semver
+scheme). It makes no sense to put a version number higher than the current
```
brson: This sentence isn't clear to me: 'exact version of the crate that deprecated the item' is a value that changes over time, but 'since' isn't going to change every time the crate version bumps.

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+newest version here, and this is not checked (but could be by external
+lints, e.g. [rust-clippy](https://github.com/Manishearth/rust-clippy).
+To maximize usefulness, the version should be fully specified (e.g. no
+wildcards or ranges).
```
brson: 'should'? This sentence doesn't add anything if it's not enforced. Is it required or not that this is a full version?

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+* `reason` should contain a human-readable string outlining the reason for
+deprecating the item. While this field is not required, library authors are
+strongly advised to make use of it to convey the reason to users of their
+library. The string is required to be plain unformatted text (for now) so that
+rustdoc can include it in the item's documentation without messing up the 
+formatting.
```
brson: What does 'plain unformatted text' mean? What's the alternative?

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+* `use` should be the full path to an API item that will replace the 
+functionality of the deprecated item, optionally (if the replacement is in a 
+different crate) followed by `@` and either a crate name (so that 
+`https://crates.io/crates/` followed by the name is a live link) or the URL to 
+a repository or other location where a surrogate can be obtained. Links must be 
+plain FTP, FTPS, HTTP or HTTPS links. The intention is to allow rustdoc (and
+possibly other tools in the future, e.g. IDEs) to act on the included 
+information. The `use` field can have multiple values.
```
brson: Another 'should' statement, with very precise rules about network protocols, without any specced validation. I'm skeptical.

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+while this has the upside of keeping things tidy, it has a downside when it 
+comes to deprecation:
+
+Let's say I have my `llogiq` crate that depends on `foobar` which uses a
+deprecated item of `serde`. I will never get the warning about this unless I
+try to build `foobar` directly. We may want to create a service like `crater`
+to warn on use of deprecated items in library crates, however this is outside
+the scope of this RFC.
+
+`rustdoc` should show deprecation on items, with a `[deprecated since x.y.z]`
```
brson: 'should' again. This sounds like a requirement. If this is going to be implemented by the RFC, then please change it to 'will' or 'must'.

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+box that may optionally show the reason and/or link to the replacement if
+available.
+
+The language reference should be extended to describe this feature as outlined
+in this RFC. Authors shall be advised to leave their users enough time to react
+before *removing* a deprecated item.
```
brson: 'should' again. This is not optional.

text/0000-deprecation.md
```diff
@@ -0,0 +1,111 @@
...
+* `use` should be the full path to an API item that will replace the 
+functionality of the deprecated item, optionally (if the replacement is in a 
+different crate) followed by `@` and either a crate name (so that 
+`https://crates.io/crates/` followed by the name is a live link) or the URL to 
+a repository or other location where a surrogate can be obtained. Links must be 
+plain FTP, FTPS, HTTP or HTTPS links. The intention is to allow rustdoc (and
+possibly other tools in the future, e.g. IDEs) to act on the included 
+information. The `use` field can have multiple values.
+
+On use of a *deprecated* item, `rustc` should `warn` of the deprecation. Note
```
brson: 'should' -> 'will'

brson:
Basically looks good to me, though I had a lot of nits about the language and the lack of required validation.

llogiq:
Thanks @brson – I'll change the wording. And I agree with you on the validation, though I'm not sure if invalid definitions should warn or err – I fear people may omit the field altogether after their first error.

llogiq:
@briansmith: The crates.io prefix is a default so one can link crates that are on crates.io directly by their names.

I'll add an example of multi-use.

briansmith:
> @briansmith: The crates.io prefix is a default so one can link crates that are on crates.io directly by their names.

IMO, the name resolution should be done the same way it is done by the compiler, and shouldn't have special logic for crates.io. In particular, this is a language feature, and the meaning of the Rust language shouldn't depend on crates.io (a layering violation).

llogiq:
I don't think it is a layering violation. Note that this default only applies to crates that are not already `use`d – those can be anywhere. Also crates.io *is* the default repository for crates. But I'm fine with making this default configurable via rustc cmd args / Cargo.toml respectively.

withoutboats:
`rustc` doesn't ehave a concept of a crate registry, much less an awareness that crates.io is the default (currently only) crate registry. Only `cargo` has that concept. `rustc` has a concept of extern crate identifiers which are associated with .rlib files. I don't think that `cargo` infrastructure should become language-level features.

I also think the field is underspecified in general. I don't think requiring that these be URLs in one of four protocols (aside: why not e.g. `git://`?) is enough to guarantee this field will be useful, and if its not necessarily useful then why make any requirements about what it contains, and then how is it different from `reason`? This problem seems better solved by simply allowing markdown in the `reason` field so as to include links.

llogiq:
Thanks, but the difference between `reason` and `use` is that the latter is specific enough to use in other programs besides rustdoc – I'm thinking about a cargo subcommand that pulls in all replacements to deprecated APIs, or an IDE feature.

llogiq:
Thinking some more about it, perhaps we should let cargo deal with it. `[doc-dependencies]` anyone?

alexcrichton:

llogiq:
It's too late for today to push an update, but here's a heads-up: The [JEP 277](http://openjdk.java.net/jeps/277), which extends Java's `@Deprecated` annotation goes a pretty similar route.

text/0000-deprecation.md
llogiq: Is "the version of the crate at the time of deprecating the item" clearer?

text/0000-deprecation.md
llogiq: I think we don't gain enough value from this field to enforce its content, so I'll remove the sentence.

text/0000-deprecation.md
llogiq: markdown.

text/0000-deprecation.md
llogiq: You're right about that. I'm changing this to just use paths (like import paths) – let's just restrict the search to crates on the path. This means we can easily check if the item exists and is public.

withoutboats:
> the difference between `reason` and `use` is that the latter is specific enough to use in other programs besides rustdoc

I understand that that is the intention of the `use` field (and would like to see that feature someday!), but you haven't described any implementation of the mechanism to demonstrate that the proposed restrictions are actually sufficient make it work. The current syntax and restrictions feel a bit like a shot in the dark hoping that it will be useful when tooling is written.

To me, the incompleteness of the design of the `use` field is the only thing holding back this obvious and straightforward language improvement. But the field is optional and can always be added in a separate RFC with a detailed design of the proposed mechanism for locating a the source code of a lang item from the content it provides.

llogiq:
> The current syntax and restrictions feel a bit like a shot in the dark hoping that it will be useful when tooling is written.

The idea is very similar to what the designers of [JEP 277](http://openjdk.java.net/jeps/277) came up with, so it's not completely baseless. The Java guys may have different restrictions when evolving their language, but they sure know their tooling.

Toby-S:
I'm not sure `use` is going to work as well in practice as expected. Although verifying the path is a great way to ensure the documentation is never stale, what about if the alternative is in another crate (like when num et al were removed from std) or the alternatives require elaboration (e.g. if behaviour is split into two different functions)? It seems like that would lead to the info being shoehorned into `reason` instead, which is what `use` is trying to avoid.

A possible solution would be to only verify the path if the `use` value is a valid rust path (if not, treat as plain text), but this does make `use` even more complicated than it already is...

llogiq:
Are you saying that writing valid rust paths (with direct access to every `extern crate` and `use`d items in the namespace, no less!) is so hard that people will opt to continue the current bad practice?

Well, if that is the case then so be it. I think there will be enough people capable and willing to use `use` as defined here to still make it useful.

Toby-S:
No, I am talking about situations where it is not possible to describe the new function as a rust path, such as when code is moved to a completely separate crate which is not referenced by the original crate (a la num). I like `use` and think the path being verified is great (as I mentioned in my previous comment), but I am questioning whether only allowing a rust path is being too restrictive.

llogiq:
I see three options:

* use `extern crate` anyway (and specify that the warnings should be omitted if it's in a `use` field
* allow for `extern xy::foo::bar` to reference `foo::bar` from the extern crate `xy`
* automatically search the library path for extern crates if a definition isn't in scope.

Personally I'm leaning towards the first option, because it's the most consistent one. However the second option may be a good shortcut.

Toby-S:
I think all three of those are quite cumbersome: the first two are of no use unless the crate is added as a dependency (which is fine if it is already a dependency, but a nuisance if it isn't necessary for anything other than `use`), and the third seems quite overboard.

Alternative options:
* Allow `use` to be plain text (but if the value matches a rust path, verify that it exists).
* Allow `use` to be a URL as well as a rust path (for linking to crates that aren't `extern`).
* Do nothing. `use` is probably fine for the vast majority of uses anyway.

:+1: from me regardless

llogiq:
Actually, plain text *is* allowed, but will elicit a warning (which can be switched off with `#[allow(...)]`). However I still think that checking and warning carries its weight, because otherwise we cannot easily distinguish between `use`s that are intentionally not paths and incorrectly written paths. Users would have to look in their docs to see if the paths are correctly linked, which should be the compiler's task.

Toby-S:
@llogiq I don't think the RFC says this, unless I'm missing something. Perhaps it should be updated to reflect this?

llogiq:
Will do.

glaebhoerl:
Perhaps "plain text `use`" and "actually-existing path `use`" should not both be called `use`?

llogiq:
@glaebhoerl I think the "plain text `use`" case is actually not very "use"ful. While it costs us little to allow `allow`ing it, I would like to discourage this use of `use`. If people want to misuse a field they should either misuse `reason` or `#[allow(..)]` the misuse of `use` explicitly.

caipre:
> `since` is defined to contain the version of the crate at the time of deprecating the item, following the semver scheme. It makes no sense to put a version number higher than the current newest version here

I may be misreading this, but it's actually desirable to use the to-be-released version here, which would have a version number one higher than the current newest release. If an upcoming release `B` deprecates a feature of `A`, `since` should read `B`, not `A`.

llogiq:
@caipre Rustc shouldn't need to know about the crate version; therefore it makes little sense to declare a future deprecation (i.e. `since="<current version + 1>"`). Usually the course of events is: 

* bump version
* mark items as deprecated since the recently bumped version
* publish the new version

withoutboats:
I still don't see how one could write a program that could reliably make use of the `use` field as specified in this proposal, even now that its been restricted to not include URLs. The only example given as to its semantics is in the proposal is this:

>E.g. let's say my foo() item was superceded by either the bar() or baz() functions in the bar crate, in conjunction with the bruzz(_) function in the baz crate, I can #[deprecate(use="bar::{bar,baz};baz::bruzz")] foo()

So semicolon seperators denote different items that are supposed to be "used in conjunction"? This notation doesn't even specify a method of composition: do you apply `bruzz()` to the output of `bar()` or `baz()`, or vice versa, or is it a method chain, and if so in what order? or perhaps `foo()` returned a tuple and each half of this returns one item of that tuple? The possibilities are myriad. What if I have multiple alternatives that are not in the same module (e.g. `bar::foo::bar` and `bar::quux::baz`)? There is no way to express that using the proposed syntax.

Again, I'm not saying that this feature shouldn't be added to the language, just that it shouldn't be half-baked. That Java uses something similar is not an argument that this will work with Rust, which has a very different module system and language semantics.

I think a nub of a `use` field (that could always be expanded later!) which has unambiguous semantics would be to allow it to contian a single path to an existent item (note that this is more restrictive than what `use` takes, no `{}`, `*`, `self` or `super`). The use case of this field would be when the new item is a drop in replacement for this one, most commonly because of renames. This would allow for expressing less cases than the version in the RFC, but it would not be ambiguous and as tooling is written using it, the field could be modified to accept more inputs that make sense for that tooling, based on actual experience working with the feature.,

mrhota:
Bikeshed/nitpick alert :)

> `since` is defined to contain the version of the crate at the time of deprecating the item, following the semver scheme. It makes no sense to put a version number higher than the current newest version here, and this is not checked

Why "require" or mention semver? As mentioned in the final sentence the field won't be checked by `rustc`. Version schemes should be left to the developer's tastes and possibly external tools.

Gankro:
@mrhota because cargo uses semver to decide what version of a crate to use -- in particular when resolving multiple dependencies with different version requirements. It's an important aspect of reproducible/reliable builds, which deprecation is part of.

mrhota:
@Gankro You're right. I only mention it because I'm not sure why semver is mentioned if the field isn't going to be checked. Moreover, the field can be checked by cargo, which is probably the proper place for the check anyway, since it's part of the manifest spec.

Cargo, and cargo alone, should enforce the usage of semver. If I'm not using cargo, then it shouldn't matter, right?

mrhota:
@llogiq 
> On use of a _deprecated_ item, `rustc` will `warn` of the deprecation. Note that during Cargo builds, warnings on dependencies get silenced. Note that while this has the upside of keeping things tidy, it has a downside when it comes to deprecation:

(Is that how Cargo works now?)

What if Cargo didn't bother silencing deprecation warnings on dependencies? I'm not sure that would be so bad. It could be potentially useful information (you can file reports with the relevant package's maintainers). Plus you'd get to KISS.

llogiq:
Since the semver check would be done as a lint and cargo silences lints on dependencies, yes, this is how it works.

And no, this is done in rustc, not cargo.

aturon:
@llogiq Thanks for this RFC! Definitely a worthwhile feature for the Rust ecosystem.

I have a couple concerns:

1. The `use` field. In my experience with the standard library, it's pretty unusual for a replacement to be as simple as a new path -- that basically amounts to moving/renaming an API. For more complicated replacements or generalizations, we've often written short snippets of code or prose to explain how to do the change. In some cases, we've pointed at different crates (when things are getting split out). I wonder: can we just have `since` and `message`, where the latter works exactly like `reason` does today? That tilts things toward flexibility for library authors.

2. Just a note, it can be useful for the `since` field to be a future version in some cases. We've talked about doing this for the standard library in cases where we know an API is going to land/be stabilized in some particular version, and we don't want the deprecation warning to fire until that version is reached (this doesn't work today, but it should!)

3. I don't see why this feature requires instant stabilization, as the RFC seems to imply. What's the thinking there?

4. I agree with @huonw's concerns about fine-grained message control; before this feature is stabilized, it'd be good to have some basic vision for how to turn off deprecation warnings in a controlled way, to make sure we've accounted for that. However, I suspect that the attribute won't need to change much, if at all -- and that any needed changes could be done through additional, optional fields. So I don't see this concern as a blocker.

alexcrichton:
@llogiq any thoughts on @aturon's comment? The libs team discussed this during triage today and we were ok merging modulo removing the `use` parts for now.

llogiq:
OK, I'll move it into alternatives.

aturon:
@llogiq 

For (2), I don't see why `since` would have to be required -- more that if it exists and is for a future version, the deprecation doesn't yet apply. FWIW, we could probably make this rustc-specific, unless people thing it'd see use in the wider ecosystem; the use case is tied to our release channels and train model.

For (3), yeah, I was saying that we could introduce it as unstable first. Though, if we end up going with the simplified design that's basically how the feature has been working forever, I don't know that we need a long stabilization process.

For (4), I'm not overly worried; we should be able to layer something on later on.

llogiq:
@aturon: Fine with me. Unfortunately, I'm a bit short on time, so I plan to have the update some time next week.

To sum it up: since will stay optional, but if there is checked by the `deprecation_syntax` lint to contain a version. The `deprecation` lint will not warn if the contained version isn't reached yet. We may want to add a `future_deprecation` lint that warns for future deprecations, if not in rustc, then in clippy. The whole deprecation feature will start out as unstable (which means nightly only), but is expected to be stabilized rather soon. We *do* want finer grained control on deprecation warnings (e.g. by crate of the deprecated API), but that will be done in a followup RFC, once I manage to find the time to write up that lint argument RFC...

alexcrichton:
The compiler currently has no knowledge of crate versions, actually, it's all opaque information coming from Cargo which passes through the compiler. Along those lines I don't think that the compiler itself should have any validation of the `since` field, but an external lint seems fine.

llogiq:
I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

llogiq:
@aturon I finally came around to update the RFC. Before we get to implement it, I have another question to you folks: Should we rename the `reason` field to `note`? Because as it is currently used, most people don't actually write the reason for deprecating their APIs there, and renaming it would leave us the option to declare a more specific `reason` field later on.

Toby-S:
`note

aturon:
@llogiq I agree that `reason` is often a misnomer, and `note` sounds fine to me.

alexcrichton:
Thanks for the update @llogiq! I'm going to go ahead and merge this for now and we can always tweak minor details like naming here and there after this is merged. Note that this will continue to be unstable to start out with and we'll stabilize it at a future date at which point the naming can be clarified.

alexcrichton:
[Tracking issue](https://github.com/rust-lang/rust/issues/29935)

quodlibetor:
The rendered link in the original post here points to a file which no longer exists, and all the commits in this thread refer to an RFC that still specifies `reason` instead of `note`.

Since this is still a popular target for search engines maybe someone could update the original "rendered" link to point to https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md ?

sfackler:
@quodlibetor Fixed!

PR #1284, opened 2015-09-17
Machine-readable output of tests

hauleth:
[Rendered](https://github.com/hauleth/rfcs/blob/machine-readable-tests-output/text/0000-machine-readable-tests-output.md)

### DIFF
--- a/text/0000-machine-readable-tests-output.md
+++ b/text/0000-machine-readable-tests-output.md
@@ -1,0 +1,146 @@
+- Feature Name: machine\_readable\_tests\_output
+  - Start Date: 2015-09-17
+  - RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Replace current test output with machine-readable one and add thin compatibility
+layer on top of that.
+
+# Motivation
+
+Currently when testing Rust provide only human-readable output which is nice
+in most cases as we are humans or other humanoids. But from time to time we need
+to feed machine with results of our tests (i.e. some kind of CI that will report
+which tests failed or something like that) and there is no way to do that in
+"civilised" way. We need to parse non-machine-readable output which isn't nice.
+
+# Detailed design
+
+Replace current, human-readable, output with JSON-based output based on
+[TAP-J][tap-j]. For compatibility with existing work flow there should be added
+built-in parser for that protocol which will output in current format.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][rfc2119].
+
+## Protocol description
+
+Output of test MUST be stream of JSON objects that can be parsed by external
+tools.
+
+### Structure
+
+Output is stream of lines containing JSON objects separated by new line. Each
+object MUST contain `type` field which designates `suite`, `test`, `bench` or
+`final`. Any document MAY have `extra` field which contains an open mapping
+for additional, unstandardized fields.
+
+### Suite
+
+  ```json
+{
+  "type": "suite",
+  "name": "Doc-Test",
+  "build": "2015-08-21T10:03:20+0200",
+  "count": 13,
+  "rustc": "2a89bb6ba033b236c79a90486e2e3ee04d0e66f9"
+}
+```
+
+Describes tests suite. It MUST appear only once at the beginning of each stream.
+
+Fields:
+
+| Name    | Description                                                    |
+| ----    | -------------------------------------------------------------- |
+| `type`  | MUST be a `suite`                                              |
+| `build` | MUST be ISO8601 timestamp of build.                            |
+| `name`  | OPTIONAL suite name i.e. "Tests", "Benchmarks" or "Doc-Tests". |
+| `count` | MUST be count of all tests (including ignored).                |
+| `rustc` | MUST be version of Rust compiler used to build test suite.     |
+
+### Test
+
+```json
+{
+  "type": "test",
+  "subtype": "should_panic",
+  "status": "ok",
+  "label": "octavo::digest::md5::tests::test_md5",
+  "file": "src/digest/md5.rs",
+  "line": 684,
+  "stdout": "",
+  "stderr": "",
+  "duration": 100
+}
+```
+
+Describes test. Each test MUST produce only one test object.
+
+Fields:
+
+| Name         | Description                                                                    |
+| ----         | -----------                                                                    |
+| `type`       | MUST be `test`.                                                                |
+| `subtype`    | SHOULD be one of `test`, `bench` or `should_panic`.                            |
+| `status`     | MUST be one of `ok`, `fail` or `ignore`.                                       |
+| `reason`     | MUST describe of failure if `status` is `fail`. Otherwise MUST NOT be present. |
+| `label`      | MUST be unique identifier of test.                                             |
+| `file`       | OPTIONAL file name containing test.                                            |
+| `line`       | OPTIONAL line in `file` that contains test.                                    |
+| `stdout`     | OPTIONAL output of standard output for given test.                             |
+| `stderr`     | OPTIONAL output of standard error for given test.                              |
+| `duration`   | OPTIONAL test execution duration in nanoseconds. MUST be present in benchmark. |
+| `throughput` | OPTIONAL test throughput in case of benchmark test.                            |
+
+### Final
+
+```json
+{
+  "type": "final",
+  "results": {
+    "ok": 10,
+    "fail": 0,
+    "ignore": 2
+  }
+}
+```
+
+Finish test suite and closes stream. Parser MUST ignore all other input unless
+it is new suite.
+
+Fields:
+
+| Name      | Description                                                                                                                                 |
+| ----      | -----------                                                                                                                                 |
+| `type`    | MUST be `final`.                                                                                                                            |
+| `results` | MUST be object containing only 3 fields: `ok`, `fail` and `ignore` which describe how many test passed, failed or was ignored respectfully. |
+
+
+# Drawbacks
+
+This is breaking change in tooling and will require new tool that will provide
+current functionality for compatibility reasons, but IMHO small pain for big gain.
+
+# Alternatives
+
+Do nothing.
+
+# Unresolved questions
+
+Test object fields:
+
+- Should there be any additional fields? How should benchmark test description
+  look like?
+- Maybe there should be additional, optional field `nspi` containing times
+  for each iteration that would help with statistical analysis of benchmarks.
+
+Object separator: is newline enough? Maybe something that will be simpler to parse.
+
+Format: should additional formats be available? Is support only for JSON enough?
+
+[tap-j]: https://github.com/rubyworks/tapout/wiki/TAP-Y-J-Specification
+[rfc2119]: https://www.ietf.org/rfc/rfc2119.txt
--- a/text/0000-machine-readable-tests-output.md
+++ b/text/0000-machine-readable-tests-output.md
@@ -5,8 +5,7 @@
 
 # Summary
 
-Replace current test output with machine-readable one and add thin compatibility
-layer on top of that.
+Replace current test output with machine-readable one.
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


text/0000-machine-readable-tests-output.md
```diff
@@ -0,0 +1,146 @@
...
+interpreted as described in [RFC 2119][rfc2119].
+
+## Protocol description
+
+Output of test MUST be stream of JSON objects that can be parsed by external
+tools.
+
+### Structure
+
+Output is stream of lines containing JSON objects separated by new line. Each
```
skade: It would make sense to agree on something like http://dataprotocols.org/ndjson/ or http://jsonlines.org/ for that or is that up to TAP-J to define?

text/0000-machine-readable-tests-output.md
```diff
@@ -0,0 +1,146 @@
...
+| Name      | Description                                                                                                                                 |
+| ----      | -----------                                                                                                                                 |
+| `type`    | MUST be `final`.                                                                                                                            |
+| `results` | MUST be object containing only 3 fields: `ok`, `fail` and `ignore` which describe how many test passed, failed or was ignored respectfully. |
+
+
+# Drawbacks
+
+This is breaking change in tooling and will require new tool that will provide
+current functionality for compatibility reasons, but IMHO small pain for big gain.
```
skade: Is the test output a committed API? Using the test API directly requires #[feature(test)], so it is not breaking a committed API.

text/0000-machine-readable-tests-output.md
hauleth: It could also make usage of `\1E` ASCII code which refers to "Record Separator". This is still under consideration, but 1 record per line is simple enough.

skade: I'd take '\n' or '\r\n', it's a pretty common format in the NoSQL-world (e.g. Elasticsearch bulk uses it) and JSON can be written newline-free very well.

text/0000-machine-readable-tests-output.md
hauleth: No, test isn't commited. But testing is available in stable Rust and change in that matter will be visible to non-compiler-devs also so I marked it as a breaking change.

text/0000-machine-readable-tests-output.md
hauleth: But is inconsistent between platform as Windows still uses `\r\n` as a line separator and I believe that `println!` macro respect that. Although this needs reconsideration as I chosen new line as it was simple enough to implement in first draft of specs.

skade: That inconsistency doesn't matter in that case as '{"test": "foo"}\r' is the same JSON value as '{"test": "foo"}'.

text/0000-machine-readable-tests-output.md
```diff
@@ -0,0 +1,146 @@
+- Feature Name: machine\_readable\_tests\_output
+  - Start Date: 2015-09-17
+  - RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+Replace current test output with machine-readable one and add thin compatibility
+layer on top of that.
```
killercup: > and add thin compatibility layer on top of that.

What do you mean by "compatibility layer"? The JSON-based output (instead of binary Rust structs)? In _Drawbacks_ you write that new tooling will be required for compatibility but this sentence sounds to me like that tooling is actually part of the RFC.

killercup:
I really like this proposal. Thanks for writing this!

text/0000-machine-readable-tests-output.md
hauleth: I mean display. I don't know how tooling should be done. There should be external program that will parse output and display it in the same way it is. Maybe it should be included into `rustc` but maybe into `cargo`. I have no idea and this need to be resolved.

alexcrichton:
cc @rust-lang/tools, definitely our territory!

Thanks for the RFC @hauleth! Some thoughts of mine:

* Perhaps there could be a record for when a test starts, as well as when it finishes? For interactive displays it could perhaps be nice to see what tests are in progress.
* We probably want to start off pretty conservative and not have many fields required by default unless necessary. For example could we drop `build` and `rustc` from the `suite` record?
* For a `test` record, the `duration` field could be dropped if we had start/stop records.
* Could this spec how the libtest library will be changed? E.g. it's good that this is preserving backwards compatibility, but we'll probably want a new flag to test binaries which generates this form of output.
* Does this output also affect `rustdoc --test`? I think it probably should and we'd just want to forward flags to the "test binary" basically.

I personally always find it a good exercise to implement features like this ahead of time to get a good feeling about what's needed to implement the current functionality we have today. Along those lines it'd certainly help weed out what's needed for maintaining the same output in terms of benchmarks and tests, but it may also be a pretty significant chunk of work!

text/0000-machine-readable-tests-output.md
masklinn: Why not make the machine-readable output an option and keep the human-readable output the default? Machine-readable output by default makes for a dreadful first-time experience.

FWIW pytest outputs a human-readable format to stdout and optionally a machine-readable one to a specified file.

The ability to output both human-readable and machine-readable is actually convenient in CI, you can send the human-readable output to a human-readable log and the machine-readable output to the CI's parser.

hauleth:
About first point. I assumed that `suite` object will be the beginning of tests so additional object would be unneeded.

About second. It is completely possible to do so. I've just added that to be sure that we are running valid binary.

About third. Of course it could be solution, but I don't see reason behind providing timestamp of beginning and ending of test, duration will do as well (maybe it's only me, but I doesn't care when my test started, I want to know how long it has been running).

About last two it would be recommended to do so. It could simplify calls.

About implementing that: I had that in mind, but currently I hadn't enough time to dig in `rustc` to find where and what so I what I loved in Ruby and power it up to work nice with Rust and provide functionality I need in my work. So it is very opinionated.

text/0000-machine-readable-tests-output.md
hauleth: That's the point of drawbacks. That this should be written, but IMHO isn't point of this RFC. I can rewrite it to provide flag that will print out machine-readable style.

About writing to file machine-readable and on stdout human-readable then I think that one should use Unix tools:

    cargo test | tee output.log | formatter

Where `formatter` is tool that change machine-readable input into desired human-readable output.

jgraham:
For comparison we designed a [similar format](http://mozbase.readthedocs.org/en/latest/mozlog.html#data-format) for logging browser test results at Mozilla. You probably don't need everything in that format, and may want some different things, but it's another point in the possible design space to consider. One of the driving considerations there was when you have tests from a third party source (e.g. because you are implementing some specification) there can be tests that you expect to fail but which you don't wish to edit. If that isn't a case you care about here you are likely to make differnt design choices.

cmr:
In general I'm a big fan of anything TAP based, and this JSON encoding removes most of problems with TAP. Human-readable output should definitely be the default, though. Perhaps an environment variable such as RUST_MACHINE_OUTPUT or a universally understood command-line to enable machine output.

These tools are primarily for people, not machines. Changing Cargo to  consume the machine-readable format is not an issue.

Looking forward to seeing something like this develop!

diwic:
`cargo test` should continue to have human readable format, and machine output could be available with `cargo test --json` or so.

erickt:
This is a great start! I think it could also be useful to capture the rustc command line arguments so that we can observe optimization levels, enabled feature flags, and etc.

For reference, a bunch of other test libraries produce the [xUnit XML File Format](https://xunit.codeplex.com/wikipage?title=XmlFormat).

tomjakubowski:
> In general I'm a big fan of anything TAP based, and this JSON encoding removes most of problems with TAP. Human-readable output should definitely be the default, though. Perhaps an environment variable such as RUST_MACHINE_OUTPUT or a universally understood command-line to enable machine output.

I don't see why it should be an environment variable; environment variables are much harder to reason about than command line flags (it's a bit like dynamic scope vs. lexical scope), and it's not like we need to adjust some behavior deep within a stack of executing programs.

A flag on the test runner (`--format=tap`, bikeshedding welcome) combined with a flag on `cargo test` itself seems like the right fit to me.

alexcrichton:
@hauleth 

> About first point. I assumed that suite object will be the beginning of tests so additional object would be unneeded.

Ah yeah I meant in addition to the suite object there'd also be an object for "this test has started to run". That gives consumers a notion of what tests are currently running, e.g. those you've seen start records for and haven't seen end records for. Additionally if a consumer of the output wants to provide a progress bar this would be useful information perhaps.

> About third. Of course it could be solution, but I don't see reason behind providing timestamp of beginning and ending of test, duration will do as well (maybe it's only me, but I doesn't care when my test started, I want to know how long it has been running).

Ah just in the sense that we don't have a lot of timestamp support in the standard library just yet, so it'd be difficult to implement this in-tree (e.g. whereas it'd be pretty easy to do it out-of-tree), so for ease of implementation we may want to avoid this for now.

> About implementing that: I had that in mind, but currently I hadn't enough time to dig in rustc to find where and what so I what I loved in Ruby and power it up to work nice with Rust and provide functionality I need in my work. So it is very opinionated.

Ah yeah no worries! You're not on the hook to implement this or anything like that, just some musings from me!

yoshuawuyts:
What are the limitations on TAP that warrant the invention of another format? To my understanding this  format would be `TAP-J` *based*, meaning it's something new.

Though being a bit inconvenient to parse, I think that using [TAP](https://testanything.org/tap-specification.html) yields more benefits in terms of tooling, adoption, familiarity and interoperability than creating something new.

It's quite annoying if every language comes with their own custom way of formatting test output. E.g. I think Rust shouldn't fall into the trap Golang fell into with their test output:
```txt
$ go test -v
=== RUN TestPrintSomething
Say hi
--- PASS: TestPrintSomething (0.00 seconds)
    v_test.go:10: Say bye
PASS
ok      so/v    0.002s
```

hauleth:
@yoshuawuyts the only problem with TAP is that this is primitive format, that doesn't provide way to message a lot of things (like type of test, performance, etc.) in uniform way. RusTAP is created to fit into Rust testing framework as it has some quirks that original TAP-J doesn't cover (i.e. benches).

IanConnolly:
I'd be happy to help with moving this along, as I've been wanting this myself recently.

brson:
My quick comments:

* Never heard of TAP-J. Need to consider it.
* Doesn't consider cargo integration
  - people do testing through cargo
  - cargo has other types of tests and I want to be able to run them
* Printing to stdout has problems
  - cargo also prints to stdout
  - test cases can print to stdout when the test runner is not capturing

hauleth:
Closing as I want to rewrite it for [TAP13](https://testanything.org/tap-version-13-specification.html) protocol (wider usage, and already existing tools).

tj:
IMO the problem with Go's is that you can't replace the output generation. It would be so much nicer if you could just:

```go
import (
  _ "my/fancy/test/output"
)
```

Then you don't have to fight about JSON, TAP, etc, just use whatever you like. Maybe Rust could go that route instead of a base format? Otherwise JSON makes more sense to me, anything else can be added on top, dissecting info from TAP is a pain.

andrew-d:
:+1: for that - being able to write a "test output plugin" (or whatever it ends up being called) seems like the best way to handle this.

jonastepe:
Then there still has to be a sensible default or fallback for those that do not provide such a plugin.

jonastepe:
However, I am also in favor of this plugin approach.

hauleth:
I am in the progress of rewriting `libtest` to allow writing reporters in sensible way. About TAP - TAP version 13 allow YAML test description after each test which seems IMHO good approach. As YAML is superset of JSON it solves both problems. 

--
Łukasz Jan Niemier

Dnia 9 sty 2016 o godz. 09:55 Jonas Tepe <notifications@github.com> napisał(a):

kamalmarhubi:
@hauleth
> Closing as I want to rewrite it for TAP13 protocol (wider usage, and already existing tools).

Do you need any help with this? I'd very much like testing to get better in Rust!

milgner:
Is this still being worked on? I think Rust would greatly benefit from this as it makes integration of tests into CI systems much more elegant. TAP 13 seems like a good format, too, even if it doesn't explicitly contain a differentiation between regular tests and benchmarks. But I guess a `# benchmark` directive could be added to address this?

denniscollective:
Can you reopen this?

PR #1288, opened 2015-09-21
Improvements to the Time APIs

wycats:
[Rendered](https://github.com/wycats/rfcs/blob/time-improvements/text/0000-time-improvements.md)

### DIFF
--- a/text/0000-time-improvements.md
+++ b/text/0000-time-improvements.md
@@ -145,7 +145,7 @@ represents an opaque (non-serializable!) timestamp that is guaranteed to
 be monotonic when compared to another `Instant`.
 
 > In this context, monotonic means that a timestamp created later in real-world
-> time will always be larger than a timestamp created earlier in real-world
+> time will always be not less than a timestamp created earlier in real-world
 > time.
 
 The `Duration` type can be used in conjunction with `Instant`, and these
@@ -162,7 +162,7 @@ difference between an earlier and a later `Instant` also produces a positive
 `Duration` when used correctly.
 
 This design does not assume that negative `Duration`s are never useful, but
-rather than the most common uses of `Duration` do not have a meaningful
+rather that the most common uses of `Duration` do not have a meaningful
 use for negative values. Rather than require each API that takes a `Duration`
 to produce an `Err` (or `panic!`) when receiving a negative value, this design
 optimizes for the broadly useful positive `Duration`.
@@ -227,7 +227,8 @@ impl Duration {
 benchmarks.**
 
 A `SystemTime` represents a time stored on the local machine derived from the
-system clock. For example, it is used to represent `mtime` on the file system.
+system clock (in UTC). For example, it is used to represent `mtime` on the file
+system.
 
 The most important caveat of `SystemTime` is that it is **not monotonic**. This
 means that you can save a file to the file system, then save another file to


text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+coding mistakes, the problem of system clock time travel means that programmers
+often produce negative durations that they did not expect, and APIs that
+liberally accept negative durations only propagate the error further.
+
+As a result, this RFC makes a number of simplifying assumptions that can be
+relaxed over time with additional types or through further RFCs:
+
+It provides convenience methods for constructing Durations from larger units
+of time (minutes, hours, days), but gives them names like
+`Duration.from_standard_hour`. A standard hour is 3600 seconds, regardless of
```
nagisa: regardles of…?

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+
+* Add a `Duration` to a `ProcessTime`, producing a new `ProcessTime`
+* compare two `ProcessTime`s to each other
+* subtract a `ProcessTime` from a later `ProcessTime`, producing a `Duration`
+* ask for an amount of time elapsed since a `ProcessTime`, producing a `Duration`
+
+Asking for an amount of time elapsed from a given `SystemTime` is a very common
+operation that is guaranteed to produce a positive `Duration`. Asking for the
+difference between an earlier and a later `SystemTime` also produces a positive
+`Duration` when used correctly.
```
nagisa: s/SystemTime/ProcessTime/ in this paragraph?

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+
+This design does not assume that negative `Duration`s are never useful, but
+rather than the most common uses of `Duration` do not have a meaningful
+use for negative values. Rather than require each API that takes a `Duration`
+to produce an `Err` (or `panic!`) when receiving a negative value, this design
+optimizes for the broadly useful positive `Duration`.
+
+```rs
+impl SystemTime {
```
nagisa: ProcessTime?

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+  /// Panics if `earlier` is later than &self.
+  /// Because SystemTime is monotonic, the only ime that `earlier` should be
+  /// a later time is a bug in your code.
+  pub fn duration_from_earlier(&self, earlier: SystemTime) -> SystemTime;
```
nagisa: ProcessTime

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+
+  /// Panics if self is later than the current time (can happen if a SystemTime
+  /// is produced synthetically)
+  pub fn elapsed(&self) -> Duration;
+}
+
+impl Add<Duration> for SystemTime {
```
nagisa: ProcessTime

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+  type Output = SystemTime;
+}
+
+impl Sub<Duration> for SystemTime {
```
nagisa: ProcessTime

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+kinds of mistakes (unexpected travel **back in time**) before the mistake
+propagates.
+
+```rs
+impl SystemTime {
+  /// Returns an `Err` if `earlier` is later
+  pub fn duration_from_earlier(&self, earlier: SystemTime) -> Result<Duration, ()>;
+
+  /// Returns an `Err` if &self is later than the current system time.
+  pub fn elapsed(&self) -> Result<Duration, ()>;
```
nagisa: `Err = ()` is not sufficient, because system time APIs might error-out for their own reasons.

wycats:
@nagisa thanks for the early feedback. I'll correct these typos ASAP!

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+```
+
+### ProcessTime
+
+`ProcessTime` is the simplest of the instant types. It represents an opaque
+(non-serializable!) timestamp that is guaranteed to be monotonic throughout
+the timeframe of the process it was created in.
+
+> In this context, monotonic means that a timestamp created later in real-world
+> time will always be larger than a timestamp created earlier in real-world
```
sfackler: Pedantically speaking, "no smaller than", since you might pull the time twice within the same tick.

text/0000-time-improvements.md
```diff
@@ -0,0 +1,327 @@
...
+rather than the most common uses of `Duration` do not have a meaningful
+use for negative values. Rather than require each API that takes a `Duration`
+to produce an `Err` (or `panic!`) when receiving a negative value, this design
+optimizes for the broadly useful positive `Duration`.
+
+```rs
+impl SystemTime {
+  /// Panics if `earlier` is later than &self.
+  /// Because SystemTime is monotonic, the only ime that `earlier` should be
+  /// a later time is a bug in your code.
```
sfackler: Are we going to guarantee that the time will "never" overflow? Windows guarantees that overflow won't happen for at least ~200 years of uptime, but I'm not sure about OSX and Linux.

text/0000-time-improvements.md
wycats: Yep!

text/0000-time-improvements.md
wycats: That is referring to `duration_from_earlier`, not the `Sub` operator :smile:

gsingh93: But `duration_from_earlier` doesn't seem to produce a `Duration` either?

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+human terms.
+
+As a result, this RFC chose to make these differences explicit, allowing
+ergonomic, uncaveated use of monotonic time, and a small speedbump when
+working with times that can move both forward and backward.
+
+# Alternatives
+
+One alternative design would be to attempt to have a single unified time
+type. The rationale for now doing so is explained under Drawbacks.
```
leodasvacas: Don't you mean "rationale for *not* doing so"?

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+* Compare two instants
+* Add a time period to an instant
+* Subtract a time period from an instant
+* Compare an instant to "now" to discover time elapsed
+
+However, there are a number of problems that arise when trying to define these
+types and operations.
+
+First of all, with the exception of instants created in the rutime of a single
+process, instants are not monotonic. A simple example of this is that if a
```
jmesmon: Ok, so we're talking about the proposed API here? Because both Posix and windows provide functions to get a monotonic system time (`clock_gettime` + `CLOCK_MONOTONIC` or `GetTickCount64()`/`GetTickCount()`).

Might want to clearly delineate where we're talking about actual system restrictions vs restrictions that just occur in the Rust api.

The simple example happens to be true because (generally) one cannot specify the clock used to generate a file timestamp.

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+program creates two files sequentially, it cannot assume that the creation time
+of the second file is later than the creation time of the first file.
+
+This is because NTP (the network time protocol) can arbitrarily change the
```
jmesmon: Might be a good idea to clarify to:
> NTP (the network time protocol) and any process/user with sufficient privileges

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+to reliably add a number of seconds to a particular instant represented as a
+human date and time ("1 million seconds from 2015-09-20 at midnight").
+
+They also mean that seemingly simple concepts, like "1 minute", have caveats
+depending on exactly how they are used. Caveats related to leap seconds
+create real-world bugs, because of how unusual leap seconds are, and how
+unlikely programmers are to consider "12:00:60" as a valid time.
+
+Certain kinds of seemingly simple operations may not make sense in
+all cases. For example, adding "1 year" to February 29, 2012 would produce
```
jmesmon: I'd replace `would` with `could` or `might`, unless we intend to have the Rust api generate invalid dates. (The subtext here is that there are other ways to resolve these than generating invalid dates)

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+The `Duration` type can be used in conjunction with `ProcessTime`, and these
+operations have none of the usual time-related caveats.
+
+* Add a `Duration` to a `ProcessTime`, producing a new `ProcessTime`
+* compare two `ProcessTime`s to each other
+* subtract a `ProcessTime` from a later `ProcessTime`, producing a `Duration`
+* ask for an amount of time elapsed since a `ProcessTime`, producing a `Duration`
+
+Asking for an amount of time elapsed from a given `ProcessTime` is a very common
+operation that is guaranteed to produce a positive `Duration`. Asking for the
```
jmesmon: Should this really be `positive`? Or just `non-negative`? (ie: might help to clarify if a elapsed time can be 0)

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+
+**This type should not be used for in-process timestamps, like those used in
+benchmarks.**
+
+A `SystemTime` represents a time stored on the local machine derived from the
+system clock. For example, it is used to represent `mtime` on the file system.
+
+The most important caveat of `SystemTime` is that it is **not monotonic**. This
+means that you can save a file to the file system, then save another file to
+the file system, **and the second file has an `mtime` earlier than the second**.
```
jmesmon: `s/has an/could have an`

SimonSapin:
Where should `Duration::span` go? An associated function of `ProcessTime`? A free function? (Having it in a single expression with a closure is sometimes nicer than `let start = ProcessTime::now(); /* do something */; start.elapsed()`

By the way, should `ProcessTime` and `SystemTime` have `now()` constructors?

retep998:
I like having the distinction between monotonic and system time.

aturon:
cc @lifthrasiir

asb:
Might it be worth contrasting this API with that offered by e.g. Jodatime,
Haskell's time library (
https://www.reddit.com/r/haskell/comments/3j2o7a/finally_a_great_tutorial_for_the_time_library/),
Go's time library, ...?

yazaddaruvala:
Hopefully not just `JodaTime` but also `java.time`
Which is a redesign of `JodaTime`, by its author, for Java 8

The article is clearly really old, but might help this design:
http://blog.joda.org/2009/11/why-jsr-310-isn-joda-time_4941.html

yazaddaruvala:
@wycats from the wording in the current draft it seems like the biggest problem isn't actually monotonicity of time but the creation of negative durations. If there are no other issues with time lacking monotonicity, one alternative I haven't seen discussed but would like to mention is to instead split up `Duration`.

`SystemTime`, `ProcessTime` -> `Instant`
`Duration` -> `uDuration`, `iDuration` (aka `UnsignedDuration`, `SignedDuration`)

Apart from naming conventions, I kinda like the consistency it has with numbers.
APIs that don't deal with negative 64bit numbers only accept `u64`.
APIs that don't deal with negative durations only accept `UnsignedDuration`.

I don't think it even changes all that much:
```
impl SignedDuration {
  // Maybe use `as` for conversions?
  pub fn to_unsigned(&self) -> Result<UnsignedDuration, Error>;

  ...
}

impl Instant {
  pub fn duration_from(&self, other: Instant) -> SignedDuration;

  pub fn duration_from_now(&self) -> SignedDuration;

  /// Panics if `earlier` is later than &self.
  /// Because ProcessTime is monotonic, the only time that `earlier` should be
  /// a later time is a bug in your code.
  pub fn duration_from_earlier(&self, earlier: Instant) -> UnsignedDuration {
    self.duration_from(earlier).to_unsigned().expect('Sorry, that wasn't earlier..')
  }

  /// Panics if self is later than the current time
  pub fn elapsed(&self) -> Duration;
}
```

Pros:
`Duration` could be a trait, allowing generic extensions like impl formatting, and user defined `Duration`s (i.e. `BigIntDuration`)
Allows for use of `SignedDuration` when appropriate
Given an understanding of Rust's numbers, `SignedDuration` vs `UnsignedDuration` might be more intuitive, while `ProcessTime` vs `SystemTime` would require a tiny explanation

Cons:
Leaving methods like `duration_from_earlier` and `elapsed` on `Instant` may cause unexpected prod bugs which we are trying to avoid. With some ergonomic sacrifice, only expose `duration_from/_now` and let users handle conversion.
How should a `SignedDuration` convert to an `UnsignedDuration`?


Let me know what downsides I'm not aware of or am currently missing. Thanks.

P.S. After writing this up, if it helps, I did find: https://internals.rust-lang.org/t/unsigned-version-of-duration/893

Veedrac:
I have nothing against a positive-first API, but one should not ignore negative durations either. That said, I can't really think of a case where you'd want signed durations and not want to branch off their direction.

So my suggestion is that the `Result<Duration, SystemTimeError>` should be changed to `Result<Duration, Duration>`, where the later `Duration` is the negative of the difference. This makes it clear that the only case `SystemTimeError` was accounting for is time travel, and makes it really obvious how to deal with negative cases. When you do want a `SignedDuration`, `Result<Duration, Duration>` is only a little bigger (and could reasonably be optimized to the same size in the future by re-using padding).

text/0000-time-improvements.md
eternaleye: There's also that there are multiple _kinds_ of monotonic time on some systems.

In particular, while CLOCK_MONOTONIC generates comparable and orderable _instants_, it does not generate comparable _durations_ because it is still affected by NTP's ability to alter the length of a second. Linux at least has CLOCK_MONOTONIC_RAW (whose ticks are of constant length), but _it_ makes no guarantee of the relationship between its units and human-meaningful ones.

In addition, suspend/hibernate is not counted for CLOCK_MONOTONIC, so durations from it are even more suspect - but Linux has the (underdocumented) CLOCK_BOOTTIME, which does count that... but isn't clear what its guarantees on the length of a tick are (consistent vs. meaningful).

arielb1:
I don't think libstd should contain a calendar - these contain lots of subjective edge cases. However, we probably want some form of absolute time for cross-system timestamps (we would need to choose between classic UNIX, UTS, or TAI). Leap-second-aware format conversions would be nice.

eternaleye:
Personally, I strongly favor TAI, for the following reasons:

- POSIX time is incoherent-as-specified, as it mandates both an 86,400-second day _and_ that it is UTC, which due to leap seconds does not have an 86,400-second day
- UTC, strictly speaking, only applies to the string-based format in which "XX:YY:60" can be represented
- The duration between two UTC times can change at any date before the latter of them has occurred, due to leap seconds being awful

cristicbz:
Two minor suggestions:
  * `ProcessTime` could be `MonotonicTime` instead, to emphasise the difference to `SystemTime` or (`SteadyTime` as in http://en.cppreference.com/w/cpp/chrono/steady_clock).
  * Instead of `from_standard_hour` I would strongly prefer newtypes and `std::convert` conversions

``` rust
    pub struct StdHours(pub u64);
    pub struct StdMinutes(pub u64);
    // ...
    impl From<StdHours> for Duration {
        // ...
    }
    // ...
```
Then conversions become as simple as: `Duration::from(StdHours(3))` and it ensures that I can store strongly typed numbers of hours/minutes etc. without getting them mixed up.

What do you think?

wycats:
Replying to @cristicbz (more later on other comments):

I don't love MonotonicTime since I expect this to be the primary API developers use for things like benchmarks, and an obscure name might put them off. 

The point about being allowed to use BOOT_TIME is good, though. Can you think of a friendly name that communicates monotonicity? TimeThatObeysTheLawsOfPhysics :p

I don't object to conversions at first glance and agree that they read nicely, provide nice ergonomics for mixing, and make it easy to avoid mistakes.

I'll let others weigh in on the panicking issue before I form a strong opinion though :) @aturon?

eternaleye:
I think that C++'s `steady_time` would be effectively perfect (as `SteadyTime`) for `CLOCK_MONOTONIC_RAW`, but for `CLOCK_MONOTONIC` there's some difficulty. I think `SequentialTime` is about as good as it can get:

- To a first approximation, monotonic : sequential :: ordinal : cardinal
- Unlike `SteadyTime` it does not imply the intervals are identical
- Sequential is a much more common word than monotonic

For `CLOCK_BOOTTIME`, there's the complication that it _behaves_ like `CLOCK_MONOTONIC` in every way, except that it also covers time suspended. Thus, both have equal claim to `SequentialTime`, but constructions like `SequentialTimeWithoutSuspended` are, to put it frankly, ugly as sin.

As far as use cases go, what will each be used for?

- `CLOCK_MONOTONIC_RAW` - anything that needs to (computationally) compare intervals/durations. Benchmarks are a big one.
- `CLOCK_BOOTTIME` - Anything that wants to give relative times in human units that correspond to external events ("Your last sync was six hours ago")
- `CLOCK_MONOTONIC` - Anything that wants to give relative times in human units that correspond to internal events ("This process has been hung for five minutes. Kill it?")

arielb1:
@eternaleye 

TAI has the problem that you can't convert ISO8601 timestamps a few years into the future into stable instants (e.g. if you compute `2020-01-01T00:00:00Z` today, and a new leap second is added in the next 4 years, then when you convert your timestamp back to GMT it will end up as `2019-12-31T23:59:59Z`). UTS is a decent enough and stable enough approximation of UT1 for ordinary computers.

jmesmon:
@wycats @cristicbz panicing on those conversions would prevent direct use of them with user-provided values.

Programmers would need to separately validate arguments to them, silly because they could do the validation themselves, and dangerous because programmers could get the verification wrong and inadvertently introduce denial of service bugs.

That said, I do like the type structuring. Perhaps a `From`-workalike that gives a result, with the panicking `From` implemented using that & `unwrap()`?

arielb1:
@jmesmon 

You probably want a tighter bound than `u64::MAX` to prevent durations from exploding within your system, so I think panicking on integer overflow is OK. 

@eternaleye 

If you want to measure events on the timescale of "hours", you probably care about reboots and therefore need a real-time clock.

Steady, monotonic clocks are for control systems (benchmarks, timeouts, that kind of thing). In that case, you need to think about suspends, and pick either monotonic or boot clocks as appropriate (raw clocks have better short-term but worse long-term stability than non-raw clocks, if you care about that). If you want to run at the same rate as some other system, you need to do that yourself.

jmesmon:
@arielb1 
I don't think that's probably the case. There are plenty of systems that won't really care how long a duration is as long as they can store it.

I think the crux of this is that it's easy to go from Result to panic (using `unwrap()`) but not very nice to go from panic to result (using `catch_panic()`).

arielb1:
@jmemson

These applications may want to add some small constant to the duration, or convert it to some other format, and then crash badly.

jmesmon:
@arielb1 
not if all the coversions provide a result and the app is using them :)

I think my concern here is that we've got 2 potential sources for these durations, and as a result we probably may want 2 different ways to construct them (with 2 different failure indications)
 - something the programmer inputs (in which case we'd prefer panicking
 - something the program gets from elsewhere (in which case a Result is preferable).

I'd prefer the second because I can adapt it easily, but I'm not opposed to having both.

eternaleye:
@arielb1
> TAI has the problem that you can't convert ISO8601 timestamps a few years into the future into stable instants.

You can't convert them into stable instants in UTC, either. If a leap second is added into the middle, the textual form would have a ":60" if you iterated it up where it didn't before the declaration, while the integer nanoseconds you already converted doesn't change.

If you convert UTC -> Instant, a leap second is declared, and you convert it back to UTC, you get a timestamp that says it's one second earlier.

Also, you're assuming the ISO8601 timestamp is not for the TAI zone, which is included in modern timezone databases. If it is, then the conversion _is_ stable.

> If you want to measure events on the timescale of "hours", you probably care about reboots and therefore need a real-time clock.

On a client machine. Besides, there's also the IRC standby of "Server has not responded for 1m40s".

> level-triggered CLOCK_BOOTTIME is the right choice for timeouts

Disagree - some benefit from MONOTONIC, as they don't care about time suspended, and rather about time running. The WM deciding whether to show the dialog offering to kill a window as hung, for example. (KWin, my WM, does exactly this).

arielb1:
@eternaleye 

> You can't convert them into stable instants in UTC, either. If a leap second is added into the middle, the textual form would have a ":60" if you iterated it up (where it didn't before the declaration occurred), while the integer nanoseconds you already converted doesn't change.
> If you convert UTC -> Instant, a leap second is declared, and you convert it back to UTC, you get a timestamp that says it's one second earlier.

I guess UTS is also unstable during the last 1000 seconds before a leap second, but that seems less scary - you won't get 00:00:00 rolling to 23:59:59 the previous day (you can still have 23:30:00 turn into 23:29:59, but that's less bad), and its limited to a 1 second delta during the 1000 critical seconds (rather than every second after the leap).

> Also, you're assuming the ISO8601 timestamp is not for the TAI zone, which is included in modern timezone databases. If it is, then the conversion is stable.

ISO8601 timestamps "in the wild" are supposed to be in solar time, whatever that is (not in TAI anyway - that would cause a half-minute unsynchronization)

eternaleye:
> its limited to a 1 second delta during the 1000 critical seconds (rather than every second after the leap).

Er, what?

Let D be the time at which a leap second is declared.
Let T be when that leap second occurs.

If at any time prior to D I convert an ISO 8601 UTC timestamp that is later than T to _any_ integral form (such as nanoseconds since the epoch, in an `Instant` - TAI, UTC, it doesn't matter how the integral form is _interpreted_), and then at any time later than D convert it _back_ to the string form, **it will be a different string**.

arielb1:
@eternaleye 

UNIX time and UTS don't have the discontinuity when mapping from UTC timestamps to counts-of-second instants (UTS has the 1000 second glitch, but that's not too bad).

eternaleye:
And as I said above in my line comment, POSIX time is incoherent as specified due to specifying both 86,400-second days and UTC, which does not have 86,400-second days in Annex B. It's a non-normative annex, but it means that no matter what behavior it has, it's contrary to the docs.

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+use for negative values. Rather than require each API that takes a `Duration`
+to produce an `Err` (or `panic!`) when receiving a negative value, this design
+optimizes for the broadly useful positive `Duration`.
+
+```rs
+impl ProcessTime {
+  /// Panics if `earlier` is later than &self.
+  /// Because ProcessTime is monotonic, the only time that `earlier` should be
+  /// a later time is a bug in your code.
+  pub fn duration_from_earlier(&self, earlier: ProcessTime) -> ProcessTime;
```
Ryman: Should return `Duration`?

text/0000-time-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+
+  /// Panics if self is later than the current time (can happen if a ProcessTime
+  /// is produced synthetically)
+  pub fn elapsed(&self) -> Duration;
+}
+
+impl Add<Duration> for ProcessTime {
+  type Output = SystemTime;
```
Ryman: `ProcessTime`?

alexcrichton:
As an update, @wycats, has pushed an update which primarily renames `ProcessTime` to `Instant` and I believe addresses a few other minor concerns. The libs team will likely look to moving this into FCP soon as well!

text/0000-time-improvements.md
```diff
@@ -0,0 +1,332 @@
+- Feature Name: time_improvements
+- Start Date: 2015-09-20
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes several new types and associated APIs for working with times in Rust.
+The primary new types are `Instance`, for working with time that is guaranteed to be
```
jmesmon: `s/Instance/Instant`

text/0000-time-improvements.md
```diff
@@ -0,0 +1,332 @@
...
+monotonic, and `SystemTime`, for working with times across processes on a single system
+(usually internally represented as a number of seconds since an epoch).
```
jmesmon: Why is it important to mention "across processes on a single system"? How does that tell me "this is the one that might move around and tries to be as close to absolute time as possible"?

I thought the RFC used `Instant` as a moment in time, in effect, `SystemTime::now()` would be an `Instant`, and so would the previous `ProcessTime::now()`. With `ProcessTime` (aka `MonotonicTime`) now renamed to `Instant`, what happens to instants (ie: moments in time, as defined below)?

seanmonstar: Indeed, in jodatime/threeten, both would be an 'Instant'.

yazaddaruvala:
I'm kinda curious, it's not very clear which of these types should be used to represent a time parsed from a file, database or request header? `SystemTime`?

Or given that these are all `u64` based, how would one represent Julius Caesar's birthday?

It's totally fine if these use cases are left to other libraries, but I think it would be really helpful to make this more clear

text/0000-time-improvements.md
```diff
@@ -0,0 +1,332 @@
...
+  /// a later time is a bug in your code.
+  pub fn duration_from_earlier(&self, earlier: Instant) -> Duration;
+
+  /// Panics if self is later than the current time (can happen if a Instant
+  /// is produced synthetically)
+  pub fn elapsed(&self) -> Duration;
+}
+
+impl Add<Duration> for Instant {
+  type Output = SystemTime;
```
SimonSapin: s/SystemTime/Instant/ ?

SimonSapin:
Overall +1, just some nitpicks:

> It [`Instant`] represents an opaque (non-serializable!) timestamp

This suggests that an `Instant` is only meaningful within the process that created it. (Perhaps the “zero” instant is when the process was created.) This should be documented at length in the eventual library.

Is `SystemTime` opaque? If not, is the “zero” instant defined? (For example 1970-01-01 00:00:00 UTC.) Presumably, future extensions of `std::time` or other libraries will want to convert `SystemTime` to a calendar date that can be shown meaningfully to users.

wycats:
@SimonSapin the original RFC had explicit language about the expected 0 time, but the existence of BOOT_TIME means that we're likely to use Instant both for process_start=0 and boot=0, depending on whether BOOT_TIME was available. I don't think we'd want a separate type just for this purpose.

wycats:
@SimonSapin I intend for `SystemTime` to be opaque for now, but this could be loosened along with a future DateTime API.

jmesmon:
I had forgotten about the non-pub nature of all the types.

If we can't actually look at the time they represent, how are we supposed to use these at all?

ie:
 - can't set a posix timer because we can't get enough info to convert to a timespec
 - can't convert a time to a string

Essentially: we can put data into a Duration, but are very limited in what we can get out (ordering only), and `moments`/`Instants`/`SystemTime` can doesn't appear to provide us a way to construct it except as an offset from the current time. Similar to `Duration`, we can only extract limited info from it after construction.

I suppose part of the problem is that I might not be aware what the intention for using these types is? At the moment, it's hard to imagine how libraries without private access will be able to utilize them.

jmesmon:
@wycats when would a `boottime` not be avaliable?

In any case, `boottime`'s avaliability shouldn't affect `SystemTime`, `boottime` would be used for `Instant`/`ProcessTime`/`MonotonicTime`.

sfackler:
`CLOCK_BOOTTIME` wasn't added to Linux until 2.6.39 and there may or may not be equivalent APIs on other platforms.

SimonSapin:
@wycats As far as I understand, the role of `Instant` is to only be meaningful within one process. So it’s fine (and probably less error-prone) for it to be opaque and an arbitrary zero, as long as that’s all documented.

`SystemTime` however is how a Rust program interacts with the outside world. I really think it should not be completely opaque, as that severely limits how it’s useful. It may be a long time before `std::time` gains calendar APIs. In the meantime, there should be a way to extract from `SystemTime` a pair of integers that can be interpreted by another library. One such library might make it into `std::time` eventually, but we’ll want to experiment in crates.io and rust-lang-nursery before committing to a stable API.

If we’re not ready to commit to a “zero” instant for `SystemTime`, we could instead keep it opaque for now but have at least one “well known” instant in `std::time` that can be used with `duration_from_earlier`:

```rust
// std::time

/// 1970-01-01 00:00:00 UTC
const UNIX_EPOCH: SystemTime = /* … */;
```

```rust
// Example user code

// Some instant…
let mtime: SystemTime = std::fs::metadata("foo/bar.txt").unwrap().modification_time();
// Drop sub-second precision and get an integer to send to the client in a JSON response:
// FIXME: remove .unwrap() and deal with negative timestamps?
let timestamap: u64 = mtime.duration_from_earlier(std::time::UNIX_EPOCH).unwrap().as_secs();
```

This is not a very convenient API, but at least the meaning of that integer is obvious when reading the code.

jmesmon:
@sfackler 

It looks like support for the last pre-2.6.39 kernel is drying up sometime this year (https://www.kernel.org/category/releases.html).

Do we have data on how much use EOL'd linux versions get?

Are there any specific platforms that lack a `boottime`-like clock that you have in mind?

wycats:
@jmesmon Skylight, the production app I work on, supports 2.6.18 and likely will for some time longer. RHEL (and CentOS) still haven't EOLed that kernel, and we still have a surprising amount of customers running versions of Linux at 2.6.18.

SimonSapin:
One of the Unresolved Questions in the current RFC is “What should `SystemTimeError` look like?”. I think the FIXME comment in my previous message leads to one possible answer: it should be renamed and defined as:

```rust
#[derive(all the things)]
struct NegativeDuration(pub Duration);
```

The return value of `SystemTime::duration_from_earlier` becomes `Result<Duration, NegativeDuration>`, which is conceptually equivalent to `(bool, Duration)` where the boolean is a sign bit. One way to deal with it might be:

```rust
match /* … */ {
    Ok(duration) => duration.as_secs() as i64,
    Err(negative_duration) => -(negative_duration.0.as_secs() as i64),
}
```

In the future `std::time` could have a `SignedDuration` type to make this more convenient.

jmesmon:
@wycats ah, even the slightly less old RHEL 6 has a kernel without `CLOCK_BOOTTIME`.

Note however that `CLOCK_MONOTONIC` is a system-wide clock. The sole difference between it and `CLOCK_BOOTTIME` is that `CLOCK_BOOTTIME` includes time where the system was suspended.

I'm pretty sure even RHEL 5 has `CLOCK_MONOTONIC`.

That said, I still don't think it was resolved whether which of `CLOCK_BOOTTIME` and `CLOCK_MONOTONIC` was correct to use. Last I heard from @eternaleye it sounded like we want both.

alexcrichton:
@yazaddaruvala, @SimonSapin 

I thought y'all brought up some good points about introspecting a `SystemTime`! I liked @SimonSapin's suggestion and the RFC is now updated with a `UNIX_EPOCH` constant which represents a fixed point in time. The `duration_from_earlier` method provides the ability to know concretely where a `SystemTime` lies in time by learning about the duration from the epoch (and also going backwards with the `Err` case).

I also fleshed out the design of `SystemTimeError` a bit more to just have a field accessor called `duration` for now.

alexcrichton:
Ah also this was updated to have `::now` constructors for both `Instant` and `SystemTime`

SimonSapin:
Should `SystemTimeError::duration` be renamed `opposite_duration` or something to that effect? (“Opposite” as in mathematical additive inverse.)

alexcrichton:
Yeah I wasn't so sure what to call it, I felt like `duration` was pretty unoffensive and I couldn't think of an adjective to put in front that didn't seem odd. Other thoughts I had were `negative_duration` (but it's not actually negative, that's the `Err` part), and `opposite_duration` I feel is kinda the same, the `Err` part already indicates the "opposite-ness" so I'm not sure it needs to be redundantly stated. Open to suggestions though!

nagisa:
`negate_duration`?

nagisa:
`absolute_duration`? `negate_duration`? `inverse_duration`?

retep998:
I'm a fan of `absolute_duration`

alexcrichton:

jmesmon:
I'd like to re-iterate the following:

 - `Instant` is not a good name for `MonotonicTime`/`ProcessTime` (time that only goes forward)
 - It isn't defined if `SystemTime` is UTC or TAI. TAI would be very nice to have.
 - It isn't defined if `Instant` counts time over system suspend or not. Providing both options would be preferred. On some systems both cannot be provided, so we should provide a mechanism to have a default on all systems with a way to query, and the greater number of options on systems that do support it.
 - Doesn't cover `SteadyTime` (ie: monotonic + seconds don't change in length)

alexcrichton:
@jmesmon one core part of this RFC is to provide as few guarantees as possible, so I think some of the questions you have are intentionally left unanswered and left as "open extensions" for future revisions of the API. 

As an example, do we actually need to define whether `SystemTime` is UTC or TAI? The only way to actually inspect it is to extract a `Duration` relative to `UNIX_EPOCH` (a fixed point in time), so I'm at least under the impression that it's not necessary to provide additional details like UTC or not. Is that incorrect, though? I may be missing something!

Next, for precisely what `Instant` is, I believe the intention of this RFC is to currently do the intersection of what platforms support, documenting only what is absolutely necessary in terms of guarantees. This should hopefully give us maximal flexibility to define multiple types in the future (such as the various clocks you mention) while still maintaining backwards compatibility.

So along those lines, how do you feel about the extra-conservative stance of this RFC? The points you raise are definitely very good questions, and I think the standard library will have support at some point, but if we can land something sooner rather than later that's a bit more conservative and leaves the door open, that'd definitely be nice!

Finally, for the naming of `Instant`, I think there's certainly a few minor naming open questions (such as `duration` on `SystemTimeError`) which can continue to be hammered out during FCP or while the API is unstable.

retep998:
@alexcrichton UTC or TAI determines whether there were leap seconds in the time between that time and `UNIX_EPOCH` which would affect the calculated `Duration`.

SimonSapin:
Is TAI typically available in OS-provided APIs?

sfackler:
It looks like Linux didn't add `CLOCK_TAI` until kernel 3.10.

retep998:
Windows only supports UTC. Conversions to and from TAI have to be done manually.

jmesmon:
@alexcrichton 

- as @retep998 noted, we probably need to define it as one or the other, otherwise leap seconds will (or won't) be included adding a few seconds of uncertainty to every `SystemTime`. That is, unless we're willing to allow a time api that specifies "SystemTime might be a few seconds off".

- Yep, looking back it appears the only promise about `Instant` is that it is "monotonic". To avoid confusion, it would be helpful to enumerate some of the common things that it doesn't promise (steady seconds, etc). At the end of the day, the clocks can be (mostly) viewed as being formed from a limited set of properties (`no_jumps`, `includes_sleep`, `includes_leap_seconds`), to clarify what the intention of the clocks of this RFC provide, it would be useful to enumerate the parts that are specified and those that are unspecified for each type. I've included some of this explicit info extracted from the prose of the RFC below.

- Being conservative is fine (great even!) as long as we're very clear about what is promised so others don't start relying on implementation details that are later refined. My concern is that this RFC doesn't quite nail down what is promised & what isn't.

- On hammering out naming: well, that is why I brought up all these issues (so they could be resolved). I want a time API too :) . 

From my reading of the RFC, here's a preliminary chart of things that are & aren't promised.
`Y` are things we promise to have (or allow to happen), `N` are things we promise not to have, `?` are things that are just never mentioned in the RFC, and `M` are things the RFC says may happen or be included (there are no `M`s in the chart).

clock | can_jump_forwards | can_jump_backwards | constant_second_length | includes_sleep | includes_leap_seconds | comparable_on_host
----- | ------- | -------- | --------- | --------- | ------ | -----
Instant | ? | N | ? | ? | ? | ? | ?
SystemTime | Y | Y |  ? | ? | ? | ?
SteadyTime * | N | N | Y | ? | ? | ?

`*` : from the time crate's docs


It is understood that it would be impossible to enumerate all clock properties, but most of the ones people care about could be noted (and provide direction when we're looking at what things to add). 

Again, having something that communicates the above info in the RFC explicitly would be a really good idea.

alexcrichton:
Ah yes, thanks for the clarification about TAI everyone! In the interest of platform compatibility it looks like we may want to go with UTC? That's well supported across Windows/OSX/Linux and is supported on Linux back to 2.6.18 (our current minimum required version).

@jmesmon also thanks for making such a handy table! I think in general if something is omitted from the RFC it can be considered a "maybe", in which case the behavior all seems pretty reasonable to me. My thoughts on that would basically look like:

* can_jump_forwards - can you clarify what this means? How is this different from a constant second length?
* constant_second_length - this seems fine to leave as just a "maybe" for now and perhaps refine it over time.
* includes_sleep - I think this only applies to an instant? In that sense it seems fine to leave as a maybe in that case, and for system time it seems like it should definitely include sleep (as it's a wall-clock-thing).
* leap seconds - it sounds like everything should be defined to be UTC
* comparable_on_host - I'm not sure what this means, could you clarify?

I'm curious if any other APIs provide these sorts of guarantees, and I'm a little worried from what I'm seeing now as well. When thinking about the implementation of these APIs, we'll probably do something along the lines of:

* Linux - using `CLOCK_MONOTONIC` for Instant and `CLOCK_REALTIME` for SystemTime 
* OSX - using `mach_absolute_time` for Instant and `gettimeofday` for SystemTime. 
* Windows - using `QueryPerformanceCounter` for Instant and `GetSystemTimeAsFileTime` for SystemTime.

This basically mirrors what libtime does today, and seems to at least somewhat correspond to what others do. There are... a [number of downsides though](http://stackoverflow.com/questions/13478093/boost-chrono-endtime-starttime-returns-negative-in-boost-1-51):

* On OSX, `gettimeofday` only has microseconds, not nanoseconds. Not a dealbreaker but may be the best we can do. 
* Looks like `mach_absolute_time` isn't actually monotonic. It uses [`rdtsc` as an implementation](http://www.opensource.apple.com/source/Libc/Libc-320.1.3/i386/mach/mach_absolute_time.c)  which is [known to not be steady](http://stackoverflow.com/a/19719739).
* On Windows `QueryPerformanceCounter` is [claimed to not be monotonic](http://stackoverflow.com/a/13478977) but I can't seem to find much more information on this, so I'm not sure whether it's actually true or not.

I'm not sure what our alternatives really are, though. The LLVM libcxx implementation [uses `mach_absolute_time` on OSX](https://github.com/llvm-mirror/libcxx/blob/25bcb0a5a950256af201d36b427f570410e70093/src/chrono.cpp#L85-L91) [as does boost](https://github.com/boostorg/chrono/blob/49d9ee8b0f5f2a155b4e638b64ab1a26c779bd5a/include/boost/chrono/detail/inlined/mac/chrono.hpp#L80-L84), and boost also uses [`QueryPerformanceCounter` on Windows](https://github.com/boostorg/chrono/blob/49d9ee8b0f5f2a155b4e638b64ab1a26c779bd5a/include/boost/chrono/detail/inlined/win/chrono.hpp#L47). So I guess if it's good enough for them it's good enough for us as well.

I generally feel that if something's good enough for C++/boost and there are no known major issues that it's fine for us to do as well, and otherwise this basically closely resembles the APIs found there. We may possibly consider renaming `Instant` to `SteadyTime`, but other than that it seems like we just need to be careful what we document.

wycats:
> In the interest of platform compatibility it looks like we may want to go with UTC? That's well supported across Windows/OSX/Linux and is supported on Linux back to 2.6.18

Is there a reason we have to make guarantees about TAI vs. UTC?

wycats:
> I'm curious if any other APIs provide these sorts of guarantees, and I'm a little worried from what I'm seeing now as well.

We will probably, eventually, have a bunch of more niche time types that handle some of these cases on platforms that support them.

@alexcrichton is right that for maximum portability, we're only guaranteeing "at least monotonic" and "at least UTC", and making no assurances about things like steadiness or leap seconds in this API.

eternaleye:
@alexcrichton 

> can_jump_forwards - can you clarify what this means? How is this different from a constant second length?

It's important to differentiate between changing the _length_ of a second and the _current second number_.

Jumping is setting the current second to some value - forward when it's greater than the old value; backwards when it's smaller. This is often used by NTP daemons when the correct time is very different from the system clock.

Changing the length of the second (also known as "slewing" the clock) really _should_ only happen if the length of a tick is not the correct fraction of the length of a second. However, many NTP daemons also use it if the delta between current time and correct time is small, so as to avoid discontinuities.

> constant_second_length - this seems fine to leave as just a "maybe" for now and perhaps refine it over time.

Sadly, it's really not - it's exactly the difference between "monotonic" and "steady", which you later conflate. If a tick is not steady (a consistent length), this makes benchmarks on the same machine incomparable. Monotonic merely means "never jumps".

> includes_sleep - I think this only applies to an instant? In that sense it seems fine to leave as a maybe in that case, and for system time it seems like it should definitely include sleep (as it's a wall-clock-thing).

Correct on what it applies to - this is the distinction between CLOCK_MONOTONIC (does not include sleep) and CLOCK_BOOTTIME (does include sleep). This cannot, however, be a maybe - if one sets (say) a "process hung" timer expecting MONOTONIC and gets BOOTTIME instead, it'll spuriously think a process is dead. This would have really nasty consequences for stuff like writing a window manager in Rust (which has, in fact, been done).

Also, CLOCK_BOOTTIME is relatively recent - it's not hard to find systems that lack it.

> leap seconds - it sounds like everything should be defined to be UTC

As much as I wish it was not the case, it's what needs done for portability, despite bringing its own hazards (see also, the bugs that happen every time leap seconds occur, like the same Linux kernel bug recurring two leap seconds in a row despite being fixed, and the Java hang.)

> comparable_on_host - I'm not sure what this means, could you clarify?

I'm pretty sure it means that _different_ processes on the _same_ machine can compare times safely - for example, both MONOTONIC and BOOTTIME have this property, REALTIME and TAI do not (because it may have jumped between measurements), and the per-process clocks cannot for obvious reasons.

> * Looks like mach_absolute_time isn't actually monotonic. It uses rdtsc as an implementation which is known to not be steady.

Monotonic != steady. Monotonic merely means that the clock can never go backwards.

nagisa:
> can_jump_forwards - can you clarify what this means? How is this different from a constant second length?

NTP may adjust time by either immediately setting time to the target time or by slowing down or speeding up ticking until the desired time is achieved.

It sounds like constant_second_length implies lack of those speed ups or downs, while that still does not mean clock cannot just jump 10 years into the future.

> On Windows QueryPerformanceCounter is claimed to not be monotonic but I can't seem to find much more information on this, so I'm not sure whether it's actually true or not.

It [is monotonic](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx#Is_the_performance_counter_monotonic__non-decreasing__), but might rollover.

AFAIR the issue it suffers from is the fact that it may be influenced by the CPU clock speed, which is pretty commonly used on processors as a energy saving method.

I found it interesting that this timer keeps ticking during sleep/hibernation etc.

> Looks like mach_absolute_time isn't actually monotonic. It uses rdtsc as an implementation which is known to not be steady.

Monotonic ≠ steady…

seanmonstar:
> Is there a reason we have to make guarantees about TAI vs. UTC?

At this moment, time in TAI is 36 seconds ahead of UTC.

eternaleye:
@wycats 

> Is there a reason we have to make guarantees about TAI vs. UTC?

Yes - otherwise, if you calculate the difference between two times (for a timeout, or any other purpose) you may be off by up to a second in either direction. This exact issue (though with Unix time rather than TAI, the same property holds) was the cause of a JVM hard-hang at the leap second, as well as a Linux kernel bug where timers would fire as they were issued in an infinite loop after crossing the leap second - the latter was fixed after the second most recent leap second, and then _recurred_ by the time of the most recent.

sfackler:
> It is monotonic, but might rollover.

It's guaranteed to last at least 100 years of uptime before rolling over so I don't think we need to worry about that though.

wycats:
> Yes - otherwise, if you calculate the difference between two times (for a timeout, or any other purpose) you may be off by up to a second in either direction.

In Rust, you should never use SystemTime for calculating timeouts, because the problem isn't limited to leap seconds; NTP can produce the same problem.

How is the leap second problem any worse than the usual NTP problem?

eternaleye:
> I'm curious if any other APIs provide these sorts of guarantees, and I'm a little worried from what I'm seeing now as well.

The RISC-V CPU architecture provides RDCYCLE (per-processor, monotonic, not necessarily steady), RDTIME (global sync <=1 tick, tick <= 100ns, steady, monotonic), and RDINSTRET (instruction counter, per-process, non-steady, monotonic)

eternaleye:
> In Rust, you should never use SystemTime for calculating timeouts, because the problem isn't limited to leap seconds; NTP can produce the same problem.

Timeouts were the first example to come to mind, and system time isn't always avoidable with them - see also, at(1) and cron.

However, the issue comes up in other contexts as well regarding leap seconds even when doing pure calculations of time - If you want to ask some server for all events older than $timestamp, and you get the difference between it and the epoch as seconds, and then use that, you may be asking the server the wrong question by half a minute if you are using TAI unawares (while almost every such API is based on UTC).

Which the API provides is only important if either:

* The user does not know which they got
* The user has an outdated database of leap seconds, and thus cannot convert

We can't do much about the latter, but we can't really afford to mess up the former.

wycats:
> If you want to ask some server for all events older than $timestamp, and you get the difference between it and the epoch as seconds, and then use that, you may be asking the server the wrong question by half a minute if you are using TAI unawares

But can't you be wrong by even more time than that if you're using the **same** system but have a bunch of clock drift and are relying on NTP?

eternaleye:
> But can't you be wrong by even more time than that if you're using the same system but have a bunch of clock drift and are relying on NTP?

No - when I say "pure computations of time", I mean "the user has asked for the server's intrusion logs between midnight on the 10th of September, 2015, and the 11th of September, 2015".

You get the seconds value between each stamp and UNIX_EPOCH, and send that pair to the server - but, whoops, that was TAI, and now you are offset by 36s. You miss the log entry of the compromise.

text/0000-time-improvements.md
```diff
@@ -0,0 +1,350 @@
...
+
+### Instant
+
+`Instant` is the simplest of the types representing moments in time. It
+represents an opaque (non-serializable!) timestamp that is guaranteed to
+be monotonic when compared to another `Instant`.
+
+> In this context, monotonic means that a timestamp created later in real-world
+> time will always be larger than a timestamp created earlier in real-world
+> time.
```
kryptan: Shouldn't it be "not less" instead of "larger"?

alexcrichton:
@eternaleye 

Thanks for the clarifications! I'm curious, however, as to your thoughts are on this API striving to be conservative. You mention a few times that it's important that we define various aspects of Instant/SystemTime, but this goes against what @jmesmon and I were [talking about earlier](https://github.com/rust-lang/rfcs/pull/1288#issuecomment-152288079). The idea is that we want this to be as conservative as possible to start out while allowing us room to expand. This should give us time to play around with the APIs on stable Rust while still reserving the right to make changes.

For example in the use cases that you mentioned, it would perhaps simply be a case where the APIs aren't used at this time. If you want a hard guaranteed about various aspects you'll have to use an external binding to something on crates.io (which may eventually move into the standard library itself). As a concrete example, I don't think that we *must* define Instant as steady to be useful for benchmarking. I don't think our current benchmarking framework is cast aside as having "incomparable results" because I think there's an important difference between "may be unsteady" vs "is unsteady" (e.g. the unsteadi-ness is likely quite rare in practice).

Sticking with the conservative choices (to allow expansion in the future) seems to comfortable allow us to say "maybe" to many of the various aspects of the timers in play. This means we can change implementation details over time or perhaps expand the API to take advantage of these looser guarantees.

---

@nagisa 

Thanks for the clarification on `QueryPerformanceCounter`. @wycats investigated a bit more and found [this documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408%28v=vs.85%29.aspx#guidance) as well which indicates that some exceptions to the monotinicity are:

* Older systems running XP (this is fine for us)
* A problem with +- 1 accuracy.

I believe this basically means that a timestamp of 1 is not equal to a timestamp of 2, but 1-2=0. In other words, we will not panic if an instant looks like it goes back in time by one unit as this seems like it may be expected behavior.

Otherwise it looks like they've definitely still got a lot going on in terms of ensuring it's monotonic, so I think we're covered there!

Also yes, sorry I mistook steady/monotonic as synonyms. Regardless, though, it's still the case that `mach_absolute_time` is not guarantee to be monotonic in the strictest sense, but seeing how it's in a few C++ implementations I'm not worried about relying on it as well.

---

As to the TAI/UTC question, I'd also be totally fine if we could not guarantee either but still provide enough to be usable, but in any case it sounds like UTC-as-an-implementation sounds like the right way forward due to compatibility?

ruud-v-a:
This RFC is thought through very well. One minor bikeshed:

* I think `from_standard_minutes`, `from_standard_hours`, `from_standard_days` can be renamed to just `from_minutes`, `from_hours`, `from_days`. What would a non-standard minute be? Would a hypothetical `from_nonstandard_minutes` make one of the minutes 61 seconds if it is called during a minute that has a leap second? Such a thing would be evil, I would expect the functions to be pure. Would `from_nonstandard_minutes(x)` return a duration of `x * 60 * 1436.0683 / 1440` seconds? (To ensure that one revolution of the earth takes 1440 non-standard minutes?). That would be crazy! I don’t think there can be any confusion about what these functions do.

  When local time is involved, it might not be obvious that a hypothetical `some_time.time_of_day()` and `(some_time + Duration::from_days(1)).time_of_day()` are not necessarily equal, but this is more of a quirk of local time. Or when you set a timer of `Duration::from_hours(1)` from `begin = SystemTime::now()`, and when the timer fires you find that `SystemTime::now() - begin` does not equal `Duration::from_hours(1)`. But again this is a quirk of the system time, not of `Duration`.

> Is there a reason we have to make guarantees about TAI vs. UTC?

No, there is no reason at all. **One must not confuse the notion of time with a system for labeling events.** Local time with daylight saving time has the unfortunate effect that distinct labels can refer to the same event, that a single label can refer to distinct events, and that certain labels that appear valid at first sight do not label an event at all. UTC has the issue that some labels that do not appear valid at first sight actually label an event. (This is true for “February 29” too, though it is more well-known that this is “sometimes a valid date”.) But that the wrapping rules for numbers in these labels can be context-sensitive or event depend on the positions of the stars, are idiosyncrasies of the labelling system.

Sampling the system clock is a measurement of time, and measurements can have errors. If the clock detects it was wrong and corrects itself after you sample it, that will not fix earlier measurements, and measurements done before and after the correction are incompatible to some degree. The reason for the correction is irrelevant.

How does this relate to the system clock? It boils down to one question: **does `SystemTime::now() - UNIX_EPOCH` really tell you the time elapsed since epoch?** If `SystemTime::now()` returns “local time now – 1970-01-01 00:00:00 as if there were 3600 * 24 seconds in a day”, then it definitely does not. If `SystemTime::now()` returns “UTC now – 1970-01-01 00:00:00 as if there were 3600 * 24 seconds in a day” it does not. If `SystemTime::now()` returns “TAI now - 1970-01-01 00:00:00 as if there were 3600 * 24 seconds in a day” it returns the right value plus a fixed offset. Most likely it is not the intention of the system clock to get the value of `SystemTime::now() - UNIX_EPOCH` right. But the way it gets it wrong is up to the system clock.

eternaleye:
@alexcrichton 
> You mention a few times that it's important that we define various aspects of Instant/SystemTime, but this goes against what @jmesmon and I were talking about earlier.

The question here is on what exactly "conservative" means - in particular, are we:
- Not providing anything we can't make guarantees about
- Not making guarantees about the things we provide

I personally am strongly in favor of the former, as our ability to make guarantees should only improve with time. However, if we _ever_ stabilize an API with fuzzy guarantees, it's unreliable forever. For instance, if we create `Instant` but make no guarantees of steadiness, then even if we tighten it up later _nobody can rely on that improvement_. It's not reflected in the type system. As a result of that, I'd strongly favor being conservative in what _rows_ of @jmesmon's matrix we expose via the API, but any we _do_ expose are fully pinned-down with no "maybe" entries.

For example, that'd mean that we would not currently expose _any_ steady clock, and on systems that expose a clock that provides stronger guarantees, we (for now) use it as a source for the merely monotonic time _type_, and either under `std::os` (or more publicly later) expose a steady type.

There's a side of me that's tempted to expose the matrix itself in the type system under `#[unstable]`, and then just use type aliases or newtypes for time types that are candidates for `#[stable]`.

arielb1:
@eternaleye 

UTC seconds = TAI seconds. The question is whether `SystemTime::now()` counts in TAI seconds or in some kind of "unspecified-UT1" seconds (which is what a computer clock is typically made to provide).

eternaleye:
@arielb1 

Not quite - my objection is to _not specifying_ which of those is the case.

jmesmon:
I agree with @eternaleye 's evaluation that later tightening the bounds of exposed clocks isn't very helpful. Realistically, if we want tighter bounds we'll want to expose new clocks.

That said, if these clocks can be usable with most things left as "maybe", it still might be useful to expose them. I imagine we'll need to have some clocks that are only available depending on OS/hardware/environment support, and as a result we'll probably want to have some way of saying "I need X & Y, would prefer Z, and don't mind W either way". Clocks with "maybes" for some attributes are one way of fulfilling that.

Note: I don't know if the clocks in this RFC are specified enough to be usable.

vadimcn:
Regarding monotonicity of [`QueryPerformanceCounter()`](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408.aspx): Although the main section on MSDN clearly states that it **is** monotonic, in the comments section, somebody reports having seen non-monotonicity in the wild.
This is further corroborated by the fact that apparently the .NET `Stopwatch` class [specifically accounts](http://stackoverflow.com/questions/1008345/system-diagnostics-stopwatch-returns-negative-numbers-in-elapsed-properties) for this possibility. 
I am afraid that panicking on a negative `Instant`-`Instant` result is going to be a source of major headache for developers, when their program would *sometimes* crash on *one of* customer's machines.

eternaleye:
@vadimcn 

Is there any indication of a fixed upper-bound on the non-monotonicity, such that it could be avoided by masking low bits?

nagisa:
> Is there any indication of a fixed upper-bound on the non-monotonicity, such that it could be avoided by masking low bits?

That’s not a valid solution, is it?

Regarding @vadimcn’s link, virtualised environments are a pretty interesting use case/test which might uncover a whole can of worms with every platform we support (and possibly various different processors that might or might not have “interesting” clocks, if any), since they might not allow virtualised kernel to access hardware directly and other small details.

eternaleye:
@nagisa 

I was hoping it was like the architectural restrictions on `RDTIME` on the RISC-V architecture, which permits 1 tick of inconsistency across cores.

vadimcn:
@eternaleye: I don't recall seeing the 2-3s number anywhere.  I would guess that non-monotonicity is on the order of as few ticks (as returned by `QueryPerformanceFrequency()`, but who knows...  Obviously, Microsoft did not intend there to be any non-monotonicity, so it's a hardware/HAL bug. 
I don't think we should try to paper it over, though I'd be in favor of providing APIs that help in dealing with such errors (for example deriving a non-negative timeout from two `Instant`s).

eternaleye:
@vadimcn 

> I don't recall seeing the 2-3s number anywhere.

It's a comment by the original reporter in response to a question about the magnitude.

>  What is the value of ElapsedMilliseconds when negative? – Michael Haren Jun 17 '09 at 17:04

> On my VM it around -3000 to -2000 – Vadym Stetsiak Jun 17 '09 at 17:11

vadimcn:
@eternaleye: he probably meant 'ticks', not 'milliseconds'.

eternaleye:
@vadimcn 

ElapsedMilliseconds is consistently used - both in the original post's code example _and_ the comments.

arielb1:
@eternaleye 

Yeah - having a `UNIX_EPOCH` without specifying whether we're counting SI or UT1 is stupid.

kryptan:
Seems that no one mentioned but [according to wikipedia](https://en.wikipedia.org/wiki/Leap_second#Proposal_to_abolish_leap_seconds) the desicion whether to keep or abolish leap seconds will be made at World Radiocommunication Conference which is scheduled for November 2–27 (i.e. it is happening now). Indeed, [their page](http://www.itu.int/net/pressoffice/press_releases/2015/Advisory-08.aspx) states:

> WRC-15 will address a number of key issues, in particular:
> * ...
> * ... the feasibility of achieving a continuous reference time-scale (UTC) with the possible suppression of the “leap second”

Lets hope that they will do the right thing.

text/0000-time-improvements.md
```diff
@@ -0,0 +1,350 @@
...
+* subtract a `Instant` from a later `Instant`, producing a `Duration`
+* ask for an amount of time elapsed since a `Instant`, producing a `Duration`
+
+Asking for an amount of time elapsed from a given `Instant` is a very common
+operation that is guaranteed to produce a positive `Duration`. Asking for the
+difference between an earlier and a later `Instant` also produces a positive
+`Duration` when used correctly.
+
+This design does not assume that negative `Duration`s are never useful, but
+rather than the most common uses of `Duration` do not have a meaningful
```
boxofrox: Typo: s/rather than/rather that/

text/0000-time-improvements.md
boxofrox: "not less" implies a timestamp **may be equal** or greater than.  This paragraph describes a timestamp **created later** which, ideally, may never be equal to the first, unless we consider the fact that measurements of time in a computer are discrete and may cause two sequential timestamps to match when the duration between them fails to exceed the time resolution.  May be worth considering if your datatype holds seconds, much less so if it holds nanosecs.  In the context of time corrections skewing timestamps (e.g. with NTP), the original text is accurate.

eternaleye: Consider a 4ghz processor. It completes four cycles per nanosecond; if getting the "now" Instant is single-cycle (as rdtsc can be, for example), then it most certainly can happen.

alexcrichton:
The libs team discussed this RFC during triage last week and after some final clarifications we've decided to merge. The last pieces were indicating that `Instant` is not always increasing, but rather never decreasing, and guaranteeing that `SystemTime` is UTC.

For now we've decided to take the conservative route of not trying to exhaustively specify all aspects of the two types here, but as usual there will be a period of `#[unstable]` for these implementations in the standard library where final tweaks and changes can be made (especially around documentation and such). The implementations are likely to follow what already exists in the `time` crate as well as similar C++ implementations.

Thanks again for all the discussion everyone!

alexcrichton:
[Tracking issue](https://github.com/rust-lang/rust/issues/29866)

tioover:
Why not implement `Hash` to `Duration` and `Instant`?

For example

```rust
#[derive(Hash)]
struct ID {
    random: u32,
    instant: Instant,
}
```

sfackler:
@tioover Duration now implements Hash: https://github.com/rust-lang/rust/pull/30818. The lack of an implementation for Instant is just an oversight - feel free to make a PR adding the impl.

Since this RFC has been merged, https://github.com/rust-lang/rust/issues/29866 is the new place for discussion, by the way.

PR #1328, opened 2015-10-19
Allow a custom panic handler

sfackler:
[Rendered](https://github.com/sfackler/rfcs/blob/panic-handlers/text/0000-global-panic-handler.md)

### DIFF
--- a/text/0000-global-panic-handler.md
+++ b/text/0000-global-panic-handler.md
@@ -1,0 +1,183 @@
+- Feature Name: panic_handler
+- Start Date: 2015-10-08
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+
+When a thread panics in Rust, the unwinding runtime currently prints a message
+to standard error containing the panic argument as well as the filename and
+line number corresponding to the location from which the panic originated.
+This RFC proposes a mechanism to allow user code to replace this logic with
+custom handlers that will run before unwinding begins.
+
+# Motivation
+
+The default behavior is not always ideal for all programs:
+
+* Programs with command line interfaces do not want their output polluted by
+  random panic messages.
+* Programs using a logging framework may want panic messages to be routed into
+  that system so that they can be processed like other events.
+* Programs with graphical user interfaces may not have standard error attached
+  at all and want to be notified of thread panics to potentially display an
+  internal error dialog to the user.
+
+The standard library [previously
+supported](https://doc.rust-lang.org/1.3.0/std/rt/unwind/fn.register.html) (in
+unstable code) the registration of a set of panic handlers. This API had
+several issues:
+
+* The system supported a fixed but unspecified number of handlers, and a
+  handler could never be unregistered once added.
+* The callbacks were raw function pointers rather than closures.
+* Handlers would be invoked on nested panics, which would result in a stack
+  overflow if a handler itself panicked.
+* The callbacks were specified to take the panic message, file name and line
+  number directly. This would prevent us from adding more functionality in
+  the future, such as access to backtrace information. In addition, the
+  presence of file names and line numbers for all panics causes some amount of
+  binary bloat and we may want to add some avenue to allow for the omission of
+  those values in the future.
+
+# Detailed design
+
+A new module, `std::panic`, will be created with a panic handling API:
+
+```rust
+/// Unregisters the current panic handler, returning it.
+///
+/// If no custom handler is registered, the default handler will be returned.
+///
+/// # Panics
+///
+/// Panics if called from a panicking thread. Note that this will be a nested
+/// panic and therefore abort the process.
+pub fn take_handler() -> Box<Fn(&PanicInfo) + 'static + Sync + Send> { ... }
+
+/// Registers a custom panic handler, replacing any that was previously
+/// registered.
+///
+/// # Panics
+///
+/// Panics if called from a panicking thread. Note that this will be a nested
+/// panic and therefore abort the process.
+pub fn set_handler<F>(handler: F) where F: Fn(&PanicInfo) + 'static + Sync + Send { ... }
+
+/// A struct providing information about a panic.
+pub struct PanicInfo { ... }
+
+impl PanicInfo {
+    /// Returns the payload associated with the panic.
+    ///
+    /// This will commonly, but not always, be a `&'static str` or `String`.
+    pub fn payload(&self) -> &Any + Send { ... }
+
+    /// Returns information about the location from which the panic originated,
+    /// if available.
+    pub fn location(&self) -> Option<Location> { ... }
+}
+
+/// A struct containing information about the location of a panic.
+pub struct Location<'a> { ... }
+
+impl<'a> Location<'a> {
+    /// Returns the name of the source file from which the panic originated.
+    pub fn file(&self) -> &str { ... }
+
+    /// Returns the line number from which the panic originated.
+    pub fn line(&self) -> u32 { ... }
+}
+```
+
+When a panic occurs, but before unwinding begins, the runtime will call the
+registered panic handler. After the handler returns, the runtime will then
+unwind the thread. If a thread panics while panicking (a "double panic"), the
+panic handler will *not* be invoked and the process will abort. Note that the
+thread is considered to be panicking while the panic handler is running, so a
+panic originating from the panic handler will result in a double panic.
+
+The `take_handler` method exists to allow for handlers to "chain" by closing
+over the previous handler and calling into it:
+
+```rust
+let old_handler = panic::take_handler();
+panic::set_handler(move |info| {
+    println!("uh oh!");
+    old_handler(info);
+});
+```
+
+This is obviously a racy operation, but as a single global resource, the global
+panic handler should only be adjusted by applications rather than libraries,
+most likely early in the startup process.
+
+The implementation of `set_handler` and `take_handler` will have to be
+carefully synchronized to ensure that a handler is not replaced while executing
+in another thread. This can be accomplished in a manner similar to [that used
+by the `log`
+crate](https://github.com/rust-lang-nursery/log/blob/aa8618c840dd88b27c487c9fc9571d89751583f3/src/lib.rs).
+`take_handler` and `set_handler` will wait until no other threads are currently
+running the panic handler, at which point they will atomically swap the handler
+out as appropriate.
+
+Note that `location` will always return `Some` in the current implementation.
+It returns an `Option` to hedge against possible future changes to the panic
+system that would allow a crate to be compiled with location metadata removed
+to minimize binary size.
+
+## Prior Art
+
+C++ has a
+[`std::set_terminate`](http://www.cplusplus.com/reference/exception/set_terminate/)
+function which registers a handler for uncaught exceptions, returning the old
+one. The handler takes no arguments.
+
+Python passes uncaught exceptions to the global handler
+[`sys.excepthook`](https://docs.python.org/2/library/sys.html#sys.excepthook)
+which can be set by user code.
+
+In Java, uncaught exceptions [can be
+handled](http://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#setUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler))
+by handlers registered on an individual `Thread`, by the `Thread`'s,
+`ThreadGroup`, and by a handler registered globally. The handlers are provided
+with the `Throwable` that triggered the handler.
+
+# Drawbacks
+
+The more infrastructure we add to interact with panics, the more attractive it
+becomes to use them as a more normal part of control flow.
+
+# Alternatives
+
+Panic handlers could be run after a panicking thread has unwound rather than
+before. This is perhaps a more intuitive arrangement, and allows `catch_panic`
+to prevent panic handlers from running. However, running handlers before
+unwinding allows them access to more context, for example, the ability to take
+a stack trace.
+
+`PanicInfo::location` could be split into `PanicInfo::file` and
+`PanicInfo::line` to cut down on the API size, though that would require
+handlers to deal with weird cases like a line number but no file being
+available.
+
+[RFC 1100](https://github.com/rust-lang/rfcs/pull/1100) proposed an API based
+around thread-local handlers. While there are reasonable use cases for the
+registration of custom handlers on a per-thread basis, most of the common uses
+for custom handlers want to have a single set of behavior cover all threads in
+the process. Being forced to remember to register a handler in every thread
+spawned in a program is tedious and error prone, and not even possible in many
+cases for threads spawned in libraries the author has no control over.
+
+While out of scope for this RFC, a future extension could add thread-local
+handlers on top of the global one proposed here in a straightforward manner.
+
+The implementation could be simplified by altering the API to store, and
+`take_logger` to return, an `Arc<Fn(&PanicInfo) + 'static + Sync + Send>` or
+a bare function pointer. This seems like a somewhat weirder API, however, and
+the implementation proposed above should not end up complex enough to justify
+the change.
+
+# Unresolved questions
+
+None at the moment.


text/0000-global-panic-handler.md
```diff
@@ -0,0 +1,183 @@
...
+
+```rust
+/// Unregisters the current panic handler, returning it.
+///
+/// If no custom handler is registered, the default handler will be returned.
+///
+/// # Panics
+///
+/// Panics if called from a panicking thread. Note that this will be a nested
+/// panic and therefore abort the process.
```
alexcrichton: When a panic handler is invoked, perhaps it could be considered as "being taken"? That would mean that this function would never need to panic as if called while panicking it'd just continue to return the default handler.

sfackler: That could work, but the semantics would be a little weird - the handler can't actually have been taken since it may need to be running on other threads as well, so you'd run into a situation where take_handler doesn't return the thing that's actually registered, and the handler would still be registered

alexcrichton: Oh wait right, this is a global resource, so there can be concurrent panics. Nevermind!

text/0000-global-panic-handler.md
```diff
@@ -0,0 +1,183 @@
...
+
+```rust
+let old_handler = panic::take_handler();
+panic::set_handler(move |info| {
+    println!("uh oh!");
+    old_handler(info);
+});
+```
+
+This is obviously a racy operation, but as a single global resource, the global
```
alexcrichton: I'm a little worried about this sort of restriction, it seems unfortunate today that a library basically can't use this API (as it'll inevitably come up as a use case). I wonder if perhaps `set_handler` could return the previous one, and perhaps also have `take_handler` exist for easier chaining? That way a robust library could use `set_handler` plus its own global state to store the returned handler (still racy, but perhaps "less so"), but I suppose that trading one race for another isn't that great

sfackler: You're envisioning a library that wants to intercept and maybe filter panics? I think if we wanted to support those use cases in a fully race-free way, we'd want a stack of panic handlers that have control over propagation to the next handler. However, I'm not sure that those use cases are common, and that they'd be significantly impacted by the installation race, particularly since no other "equivalent" API in another language behaves like that.

There's really no reason for `set_handler` to not return the old one, so we might as well.

alexcrichton: Yeah I'm not envisioning any particular use case in mind, this just the experience I've had from the standard library itself, specifically with regards to signals. A signal handler is some global state and we as a library want to deal with it sometimes (e.g. provide users nicer abstractions), so we may not be able to assume "only applications will reasonably use this API" is basically all I'm getting at.

Thinking more on this I'm not sure that trading one race for another is worth it, so without going for a stack (which I'd also like to avoid) I think the current API is fine.

sfackler: One other option to actually solve the race would be to wrap the handler in an `RwLock`, and adjust the API here so there's an `update_handler` function that returns a thing holding the write lock and having the `take` and `set` functions.

text/0000-global-panic-handler.md
```diff
@@ -0,0 +1,183 @@
...
+
+impl PanicInfo {
+    /// Returns the payload associated with the panic.
+    ///
+    /// This will commonly, but not always, be a `&'static str` or `String`.
+    pub fn payload(&self) -> &Any + Send { ... }
+
+    /// Returns information about the location from which the panic originated,
+    /// if available.
+    pub fn location(&self) -> Option<Location> { ... }
```
alexcrichton: Ah one thing I meant to ask, is there a reason this returns `Location` instead of `&Location`? In theory a `Location` could be at least `Clone` (storing a `&'static str` internally for the file perhaps) which would allow storing locations elsewhere if necessary.

sfackler: I can't remember why I chose a `Location<'a>` vs a `&Location` off the top of my head, but `&Location` seems fine as well.

bstrie:
Could this mechanism be used to support abort-on-panic semantics without the need for a compiler flag?

sfackler:
Yep, though you'd still have all of the panic infrastructure (landing pads, etc) in place, so we'd probably still want to have the flag. This would cover the use case of "I want to find out about panics via a process abort (maybe with debug data logged or something)", but not "I'm running in an environment that I don't want/can't have stack unwinding".

arielb1:
@sfackler 

`-Z no-landing-pads` causes a very ugly crash on panic with the default (unwinding) panic handler, but otherwise it is kind of orthogonal.

text/0000-global-panic-handler.md
vadimcn: This whole thing reminds me of [Windows Vectored Exception Handlers](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681420%28v=vs.85%29.aspx).  Perhaps we could borrow design from there?

ranma42: It looks like some of the issues that a global panic handler involves (like races) would be trivial to avoid with a per-thread panic handler. The "alternatives" section mentions that it would be a possible extension, but it does not explore the consequences. The first ones that come up in my mind are:
 - these races would not be possible
 - there would be to make some design choice about what handler should be given to new threads
 - the panic handler of each thread is called at most once (except if it is propagated to other threads), hence its signature might be more restrictive

AFAICT these ideas were not explored in #1100 because its main focus was about avoiding the output from the `on_panic` handler, but some more details might be useful in this new RFC in order to better explain and compare the tradeoffs between global & per-thread handlers.

alexcrichton: @vadimcn 

Hm that's a good point! It looks like the two primary things there (if I'm reading it right) are:

* Handlers are stacked, not overwritten. This means that everything registered is called in sequence.
* Handlers all have a token, and the token can be used to unregister the handler.

I think that scheme doesn't have the race problems I'm thinking about here, and it also provides the nice ability to "shut down" something and have handlers in theory scoped for smaller components. Additionally, you don't have to "by default try to be robust" because you don't have to worry about callling someone else's handler.

@sfackler, those semantics sound relatively appealing to me, what do you think?

---

@ranma42 

The major downside of a thread-local handler is that you can't set a handler for yet-to-be-spawned threads. I believe that any sort of global handler scheme can be used to implement a thread-local handler scheme (e.g. with more restrictive interfaces).

nagisa: > The major downside of a thread-local handler is that you can't set a handler for yet-to-be-spawned threads.

That’s not true. We do execute arbitrary code after `clone`-ing and could easily copy handlers over if option is set or it is generally wanted.

alexcrichton: @nagisa perhaps, yeah, but that means that threads not spawned in Rust won't have inheritance I believe?

ranma42: @alexcrichton The suggestion by @nagisa is exactly one of the possible choices for "what handler should be given to new threads". I understand that such approach would lose inheritance for threads spawned by non-Rust code, but currently the focus of the RFC seems to be even more restrictive: "the global panic handler should only be adjusted by applications rather than libraries". Anyway, my point is that it would be desirable to explain in the RFC itself (instead of just here in the discussion) some of these tradeoffs.

aturon:
cc @rust-lang/lang @rust-lang/core -- while technically a "library" issue, this is a pretty central aspect of the language and should get scrutiny across these teams.

aturon:
The API design here seems reasonable for the "global resource" approach to panic handling, and I agree with @sfackler that *as long as panic handlers are an application-level concern*, this probably works just fine. But I have a couple concerns.

### Composability/library use

My biggest worry about the design is potential use within libraries; I agree with @alexcrichton that such use is basically inevitable. Global resources give you poor composability between libraries (think: global state like the current working directory).

As others have mentioned, the per-thread handler model could potentially deal with it. The alternatives section mentioned the per-thread model, but dismisses it essentially due to lack of inheritance. But we should be able to provide inheritance for threads spawned via the standard library at least. So I'd like to discuss the tradeoffs here a bit more deeply.

Relatedly, the RFC claims that it'd be easy to add per-thread handlers on top later. I'm curious exactly what you have in mind -- would it be an entirely separate layer prior to calling the global handler?

### Pre- vs post-unwinding

The proposal here is to invoke the handler immediately upon panic, rather than after unwinding. IIRC this was motivated in part so that the call stack is available. But it's also a bit counterintuitive, and doesn't play well with `recover` -- even in cases that panics are "caught" and otherwise "handled" in some way, the generic global handler will still be run.

Is there some other way we could retain the needed stack information but invoke the handler after unwinding? Or is there rationale, perhaps for providing hooks on both ends of unwinding?

### Abort semantics

Finally, there are two issues around abort that would be good to settle:

- To what degree does this API tie us to the current "abort on double panic" semantics? I know some have argued against that semantics (@nikomatsakis among them, IIRC). I've heard talk of, for example, "bundling together" the payloads from multiple panics. It'd be nice to keep our options open.

- Others have raised the issue of using this mechanism for an abort-on-panic semantics, but @sfackler points out that you probably want to invoke a separate compiler flag as well. It'd be nice to have a relatively clear and simple story about how to get abort-on-panic, and how that might play with the handlers here. (Not a blocker for this RFC, more something to discuss a bit just to be sure there's a reasonable plan.)

aturon:
Follow-up: my previous comment sort of implied that inherited, thread-local handlers "solve" composability, but I think that's not really true; they just mitigate it. It's still possible to use multiple libraries that want to frob the thread-local handler, and they have to play together somehow.

arielb1:
@aturon 

Panic handlers are intended to work even without unwinding, and therefore they should run *before* it. If you want to ignore panics that are caught by `recover`, you should do your own `recover` at the event loop.

The interaction with double panics is annoying. I would prefer a version that has `panic!("foo")` being equivalent to
```Rust
    let panic_info = PanicInfo {
        location: current_location!(),
        info: box "foo"
    };
    GLOBAL_PANIC_HANDLER.borrow()(&panic_info);
    unwind(panic_info);
```

In that case, a panic in a panic handler would lead to a safe death via stack overflow. If this is not desirable, a panic handler could have a recursive lock (we need to have some way to handle concurrent panics anyway).

alexcrichton:

Note that this entering FCP at the same time as #1100 and the two will likely be decided upon as a unit.

yberreby:
This is a more polished design than the one I proposed and didn't have time to revise, and supports adding thread-local handlers in the future.

:+1:

sfackler:
@aturon 

I am not a fan of an inheritance based situation for a couple of reasons:

* Many programming languages provide a global exception handler - C++, Java, Python, etc. As far as I can tell, zero languages provide this kind of handler inheritance scheme.
* Threads not spawned via `std::thread::spawn` are left in some kind of limbo state with respect to panic handling. We could have some function you could call to register a handler after the fact, but that makes writing code that interoperates with other languages way weirder.

It's not totally clear to me what use cases you see for libraries wanting to mess with a global handler. From my experience in Java, the only libraries that mess with it are the ones that explicitly advertise that fact and take over the handler to e.g. forward uncaught exceptions from an Android app back out to some server for tracking. I can imagine we'd update the log crate to have a function that will replace the global handler with one that logs errors for people that want to do that.

Libraries certainly might find it convenient to set custom handlers on specific threads that they create and maintain but that doesn't seem like a necessity - `panic::recover` is a better bet for thread pools and Java is the only language I've looked at that supports per-thread uncaught exception handlers.

Running handlers post-unwinding does seem like a better bet if we can figure out what to do with backtraces. @alexcrichton, is the stack walking part of backtrace generation fast enough that we can reasonably run it for every panic? If not, we might need a separate method to indicate that the handler is interested in backtrace info or something like that.

I don't think this ties us to abort on double panic any more than the Err value for joining a thread does - in both cases we have a `Box<Any + Send>` which right now implies one payload, though the thing inside of it could of course in the future be something wrapping a `Vec<Box<Any + Send>>` or whatever.

The reentrancy of the handler itself is a bit more subtle, but I don't feel super strongly about what the behavior is there. It could be treated as a double panic -> abort, or recursively call the handler, or skip the handler the second time around.

arielb1:
@sfackler 

From my point of view this is not supposed to be an *unhandled exception handler* because Rust does not support *handled exceptions*, it just supports recovery from exceptions. It is just the panic handler, and it should run when panic is called (for example, unwinding may not actually be supported, and even if it is, it may call undesirable destructors).

alexcrichton:
In terms of composability I think I prefer the ability to only add a handler and unregister *just that* handler (e.g. the Windows-like style), but in terms of implementation what's proposed here is more general (e.g. you can build the Windows style on this one). As a result, plus some of the comments @sfackler has made, I think I'd lean in favor of this strategy, acknowledging that the composability just isn't that great and more composable solutions can be built up externally. The upside of this proposal is that it's easy to understand and implement and provides at least the foundation for doing more pieces later.

> Pre- vs post-unwinding

@aturon, this is an interesting point I hadn't considered before, yeah. I wouldn't necessarily say that there's a hard requirement that it runs pre-unwinding, it's just convenient today that the backtrace is available. That being said, it may not be totally unreasonable to have a scheme along the lines of the outermost `recover` block is the only one that will run the global handler. For example if you have nested `recover` blocks on one thread you'll never invoke the global handler until the outermost one catches the panic. We could easily set this up with TLS, and it means that FFI boundaries and thread boundaries would run the global handler, but manual calls to `panic::recover` wouldn't.

Along those lines, it may not be so critical that we distinguish pre vs post unwinding.

> Abort semantics

I agree with @sfackler that I don't think this really changes our story here much. Even if we have to special case the panic-handler panicking that seems like it's not the end of the world, the desire to handle double panics I believe is motivated to work with "most code" rather than *all* code.

> is the stack walking part of backtrace generation fast enough that we can reasonably run it for every panic?

@sfackler, I'm not actually quite sure, but panicking is so slow already it probably wouldn't matter if we just generated a backtrace. That being said I still think we should always have backtraces be optional because including libbacktrace is a nontrivial dependency of the standard library that may not always be desired.

sfackler:
Yep, I figure we'd make any backtrace accessor return an `Option<Whatever>` like `PanicInfo::location` does in this proposal.

aturon:
@sfackler Thanks for the detailed reply!

I think I'm persuaded on the global vs. thread-local issue. In favor of global:

- the most common way for libraries to interact will be to simply chain handlers;
- in principle a single global handler is enough to bootstrap other means of coordinating handlers;
- it's dirt simple;
- everybody does it.

And it's not like the thread-local version completely solves all composability problems.

There's some slight chance that there will arise multiple competing frameworks on top for composing handlers, which would be a shame, but even if that happened we could probably push to standardize on one, so I'm not too worried.

re: pre/post-unwinding, I'm coming around to @arielb1's perspective that this is more like an "on panic hook" rather than a "panic handler". If we're clear enough in the naming and docs, we can probably avoid confusion around when the callback is triggered, and it seems like all of the technical advantages are on the side of doing this pre-unwinding.

So at this point I'm pretty much :+1:. Thanks again for the replies!

arielb1:
@alexcrichton 

I am not particularly sure about the behavior with `panic::recover`. The intended uses for it are
* An FFI handler that re-packages panics across an FFI boundary.
* A "worker thread" top-level handler to allow recycling the worker thread when a panic occurs.

In the first case, the panic handler should definitely run, as the intended behavior is that of a normal panic. In the second case, the panic handler should be cooperating with the top-level handler, so there should be no problem.

Anyway, the handler really should be run before the stack is unwound - a "send backtrace to a server" handler may want to use a logger allocated on the stack, and destroying it before the handler is run would be sad.

alexcrichton:
@arielb1 ah yeah that's a good point, I always seem to forget about the worker-thread-like situation!

nikomatsakis:
I feel uncomfortable with this RFC casually stating that a double panic leads to abort. I'd prefer to leave the proper handling of this case as an Unresolved Question, since I think that aspect of the language semantics is not entirely nailed down. That said, I don't think it really introduces a lot of complications -- I think of this as basically inserting a destructor that runs at the point of panic, before other destructors, and hence the semantics of a panic in that context seems to be no different from the semantics of a panic occurring in the middle of any other destructor. Currently, this aborts, and it may be that we cannot in fact change this (or find a more satisfactory semantics).

Just in general, I think that aborting in some situations seems like it *might* be the right thing. There is certainly precedent: it's quite close to what C++ does. If we are to embrace the abort, I could even imagine going a bit further, and adding the concept of "nounwind" functions, which will abort if unwinding passes through them. All destructors could then be marked nounwind, so that we wind up with a simple rule -- do not panic in a destructor (whether or not a panic has already occurred). 

However, there are definitely use cases where an abort is kind of "game over", so I still find the idea that we can't recover in *some* more graceful manner unfortunate. It does seem like we won't ever be able to guarantee a "complete" recovery though (because the recovery process itself is not completing), so perhaps that would have to be something that the user opts into -- i.e., saying "it's ok if some destructors never complete (or never execute), please don't abort". (Also, this line of thinking argues against making unwinding during destructors a complete no-no.)

In this sense, executing the panic handler immediately on panic is good, because it gives us the opportunity to affect (and maybe configure, someday) how unwinding will proceed. 

So I guess, all in all, I feel pretty comfortable with this RFC as a starting point.

text/0000-global-panic-handler.md
```diff
@@ -0,0 +1,183 @@
...
+```rust
+/// Unregisters the current panic handler, returning it.
+///
+/// If no custom handler is registered, the default handler will be returned.
+///
+/// # Panics
+///
+/// Panics if called from a panicking thread. Note that this will be a nested
+/// panic and therefore abort the process.
+pub fn take_handler() -> Box<Fn(&PanicInfo) + 'static + Sync + Send> { ... }
```
diwic: Do you expect `std::panic` to be part of core and if so how can there be a `Box` here (which is in the alloc crate)? Or if not; how will this work in a no_std scenario?

sfackler: The panicking infrastructure already allocates and is already in libstd.

`no_std` executables need to define some lang items to implement the functionality for panicking.

alexcrichton:
The libs team discussed this RFC during triage yesterday, and the conclusion was to merge. It looks like there is broad consensus around this strategy (especially as a first-but-unstable pass) modulo some careful wording about what it means to double-panic.

[Tracking issue](https://github.com/rust-lang/rust/issues/30449)

PR #1370, opened 2015-11-14
RFC: Transparently use verbatim paths on Windows

tbu-:
[(rendered)](https://github.com/tbu-/rust-rfcs/blob/rfc_windows_verbatim_paths/0000-windows-verbatim-paths.md)

### DIFF
--- a/0000-windows-verbatim-paths.md
+++ b/0000-windows-verbatim-paths.md
@@ -1,0 +1,79 @@
+- Feature Name: windows_verbatim_paths
+- Start Date: 2015-11-14
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Transparently use the long pathname API in Windows, allowing Rust programs to
+deal with paths longer than 260 characters without having to implement their
+own pathname handling.
+
+# Motivation
+[motivation]: #motivation
+
+Windows has multiple ways to handle paths:
+
+1. ANSI paths, as in `CreateFileA`, restricted to one code page with 256
+   characters and paths of 260 characters: `r"C:\"`. Rust doesn't use these.
+2. Traditional Unicode paths, as in `CreateFileW`, accepts a superset of
+   UTF-16, as it allows for unmatched surrogates, maximum of 260 UTF-16 code
+   units: `r"C:\"`. Rust currently uses these.
+3. Extended Unicode paths, also accepted by `CreateFileW`, accepts the same
+   superset of UTF-16 but allows for roughly 32768 code units.
+
+Rust currently uses the second path style, unless the application explicitly
+chooses to implement the path handling needed for 3. Since paths of the other
+forms can be converted into the third form, Rust could implement this in the
+standard library. Rust applications would consistently and robustly support
+long path names.
+
+# Detailed design
+[design]: #detailed-design
+
+The path translation will happen directly above the WinAPI layer, when the
+`OsString`s are converted to arrays of 16-bit integers (UTF-16 code units).
+
+- `C:\foo\bar` to `\\?\C:\foo\bar`.
+- `\foo\bar` to the current working directory (CWD) joined with `\foo\bar`.
+- `foo\bar` to CWD joined with `foo\bar`.
+- `C:foo\bar`. This depends if on the current working directory. If it is on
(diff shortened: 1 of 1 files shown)


dgrunwald:
+1 for the general idea; but we have to be very careful how we normalize the paths. Windows does more than the transformations listed in the RFC.

For instance, Windows strips trailing dots and spaces (in each path component): `C:\a. .` is the same as `\\?\C:\a`, which is a different file than `\\?\C:\a. .`.

And then there's also special files like `CON` and `PRN`. `CreateFileW(L"c:\\con", GENERIC_WRITE, ...)` actually opens a file handle for stdout; but `\\?\c:\\con` is a file on disk. It's debatable whether we want the rust stdlib to be able to open these (probably not, as these are just a relict from the DOS days...).

Not sure if there are any additional caveats. Do 8.3 short names interact with file name normalization, or are they handled completely in the file system driver?

alexcrichton:
I would personally be quite wary of doing this sort of manipulation in the standard library. It should **never** be the case that someone wants to subvert this behavior, because the purpose of the standard library is to provide the safe interface here so applications don't have to bind it themselves.

Along those lines I am not confident that we can convert to verbatim paths and continue to satisfy 100% of use cases of these APIs. I'd be more comfortable if there were other sources cited here such as:

* MSDN articles
* Official Windows documentation about this sort of transformation
* Implementations in other languages or libraries

Zoxc:
I'm supportive of this. I also did this for my Ruby implementation. It's annoying not being able to deal with and those silly special cases long paths in a portable way. I also hate applications telling me the paths it uses are too long.

@alexcrichton I don't understand your comment. Could you be a bit clearer?

For reference: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

0000-windows-verbatim-paths.md
```diff
@@ -0,0 +1,79 @@
...
+# Detailed design
+[design]: #detailed-design
+
+The path translation will happen directly above the WinAPI layer, when the
+`OsString`s are converted to arrays of 16-bit integers (UTF-16 code units).
+
+- `C:\foo\bar` to `\\?\C:\foo\bar`.
+- `\foo\bar` to the current working directory (CWD) joined with `\foo\bar`.
+- `foo\bar` to CWD joined with `foo\bar`.
+- `C:foo\bar`. This depends if on the current working directory. If it is on
```
retep998: I'd like to point out that each drive does store its own current working directory. I tested this before, so we cannot assume the non-current drive's current working directory is the root.

retep998:
I'd personally prefer if there was first a crate on crates.io that provided path normalization so it could be carefully tested.

0000-windows-verbatim-paths.md
tbu-: The [Old New Thing blog post](http://blogs.msdn.com/b/oldnewthing/archive/2010/10/11/10073890.aspx) on MSDN says something different.

0000-windows-verbatim-paths.md
```diff
@@ -0,0 +1,79 @@
...
+
+Windows has multiple ways to handle paths:
+
+1. ANSI paths, as in `CreateFileA`, restricted to one code page with 256
+   characters and paths of 260 characters: `r"C:\"`. Rust doesn't use these.
+2. Traditional Unicode paths, as in `CreateFileW`, accepts a superset of
+   UTF-16, as it allows for unmatched surrogates, maximum of 260 UTF-16 code
+   units: `r"C:\"`. Rust currently uses these.
+3. Extended Unicode paths, also accepted by `CreateFileW`, accepts the same
+   superset of UTF-16 but allows for roughly 32768 code units.
```
ruud-v-a: I think you are missing an example here.

tbu-: Yup. E:g. `r"\\?\C:\"`.

0000-windows-verbatim-paths.md
retep998: I believe you can use `GetFullPathName` with something like `C:` to get the current working directory for that drive.

tbu-:
I currently don't have time to update this.

PR #1375, opened 2015-11-20
Always-true Equality for PhantomData<T> to PhantomData<T>

samdoiron:


Ticki:
:+1:

jonas-schievink:
`PhantomData` already implements `PartialEq` and `Eq`, as can be seen [here](http://manishearth.github.io/rust-internals-docs/src/core/marker.rs.html#399) (the `impls!` macro generates [this](http://manishearth.github.io/rust-internals-docs/src/core/marker.rs.html#246-255)).

The reason that `derive` doesn't work on the struct is probably due to rust-lang/rust#26925

samdoiron:
Looks like it to me! Closing in favour of that issue.

PR #1394, opened 2015-12-05
Add `maybe!` macro equivalent to `try!`

hauleth:
[Rendered](https://github.com/hauleth/rfcs/blob/try-some-macro/text/0000-try-some-macro.md)

### DIFF
--- a/text/0000-try-some-macro.md
+++ b/text/0000-try-some-macro.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-Add `try_some!` macro for `Option` equivalent to `try!` for `Result`.
+Add `maybe!` macro for `Option` equivalent to `try!` for `Result`.
 
 # Motivation
 [motivation]: #motivation
@@ -19,7 +19,7 @@ It would simplify some `Option` related functions.
 Just simple macro that would be almost the same as `try!` macro:
 
 ```rust
-macro_rules! try_some {
+macro_rules! maybe {
     ($expr:expr) => (match $expr {
         Some(val) => val,
         None => return None,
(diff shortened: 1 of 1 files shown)


cassiersg:
For reference, there also exists a [try_opt crate](https://crates.io/crates/try_opt) that does the same thing as soma.

hauleth:
@cassiersg I wasn't aware that this already exist. :+1:

Ticki:
See also #243

4xrsJCr9:
Please consider renaming this macro to 'maybe'.

reem:
Some alternatives: `Option::and_then` or `try!` with `Option::ok_or`/`Option::ok_or_else` depending on what's appropriate.

alexcrichton:

The libs team feels that we are likely to not merge this for now in light of [trait-based exception handling](https://github.com/rust-lang/rfcs/blob/master/text/0243-trait-based-exception-handling.md) having recently landed. We're likely to change `try!` to simply become `?` in which case this may mesh more nicely with `catch { ... }` as proposed there.

daboross:
I feel like this doesn't really need to be a part of the standard library, as it has a lot smaller use case than try! does. I mean it wouldn't be *too* bad to add it, but it might just be extra bloat.

skade:
I would strongly avoid naming the macro `maybe!` for similarity with the type of the same name in other programming languages. Rust uses `Option` instead of `Maybe`, `maybe!` sounds like diverging from that.

k3d3:
Having something that deals with Options being called `maybe!` is confusing IMO - it should be called `some!`, `option!`, or `opt!`.

nrc:
I strongly want a facility for doing this and in the past I've argued for a macro (so many projects have their own for doing this!), however, it seems the best way forward now is to make `?` work with `Option`, so I think we should not accept this RFC (I also don't like `maybe` as a name, fwiw) and push on getting `?` to support `Option` asap.

alexcrichton:
The libs team discussed this RFC during triage yesterday and the conclusion was to close (in the face of trait-based exception handling). Thanks again though for the RFC @hauleth!

PR #1399, opened 2015-12-07
RFC: Add #[repr(pack = "N")]

retep998:
[Rendered](https://github.com/retep998/rfcs/blob/optimal-rabbit-packing/text/0000-repr-pack.md)

### DIFF
--- a/text/0000-repr-pack.md
+++ b/text/0000-repr-pack.md
@@ -20,7 +20,10 @@ writing Rust libraries such as `winapi` a nightmare).
 
 At the moment the only way to work around the lack of a proper
 `#[repr(pack = "N")]` attribute is to use `#[repr(packed)]` and then manually
-fill in padding which is a burdensome task.
+fill in padding which is a burdensome task. Even then that isn't quite right
+because the overall alignment of the struct would end up as 1 even though it
+needs to be N (or the default if that is smaller than N), so this fills in a gap
+which is basically impossible to do in Rust at the moment.
 
 # Detailed design
 [design]: #detailed-design
--- a/text/0000-repr-pack.md
+++ b/text/0000-repr-pack.md
@@ -52,12 +52,11 @@ with the accepted values are:
 * Packing values must be a power of two.
 
 By specifying this attribute, the alignment of the struct would be the smaller
-of the specified packing and the default alignment of the struct otherwise. The
-alignments of each struct field for the purpose of positioning fields would also
-be the smaller of the specified packing and the alignment of the type of that
-field. If the specified packing is greater than or equal to the default
-alignment of the struct, then the alignment and layout of the struct should be
-unaffected.
+of the specified packing and the default alignment of the struct. The alignments
+of each struct field for the purpose of positioning fields would also be the
+smaller of the specified packing and the alignment of the type of that field. If
+the specified packing is greater than or equal to the default alignment of the
+struct, then the alignment and layout of the struct should be unaffected.
 
 When combined with `#[repr(C)]` the size alignment and layout of the struct
 should match the equivalent struct in C.
--- a/text/0000-repr-pack.md
+++ b/text/0000-repr-pack.md
@@ -78,6 +78,8 @@ Windows API.
 * The alternative is not doing this and forcing people to continue using
   `#[repr(packed)]` with manual padding.
 * Alternatively a new attribute could be used such as `#[pack]`.
+* `#[repr(packed)]` could be extended as either `#[repr(packed(N))]` or
+  `#[repr(packed = "N")]`.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-repr-pack.md
+++ b/text/0000-repr-pack.md
@@ -64,15 +64,21 @@ should match the equivalent struct in C.
 `#[repr(packed)]` and `#[repr(pack = "1")]` should have identical behavior.
 
 Because this lowers the effective alignment of fields in the same way that
-`#[repr(packed)]` does (which caused https://github.com/rust-lang/rust/issues/27060 ),
-while accessing a field should be safe, borrowing a field should be unsafe.
+`#[repr(packed)]` does (which caused [issue #27060][gh27060]), while accessing a
+field should be safe, borrowing a field should be unsafe.
+
+Specifying `#[repr(packed)]` and `#[repr(pack = "N")]` where N is not 1 should
+result in an error.
+
+Specifying `#[repr(pack = "A")]` and `#[repr(align = "B")]` should still pack
+together fields with the packing specified, but then increase the overall
+alignment to the alignment specified. Depends on [RFC #1358][rfc1358] landing.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-This would unfortunately make my life easier even though one of the unstated
-goals of Rust is to make my life as difficult as possible when doing FFI with
-Windows API.
+Duplication in the language where `#[repr(packed)]` and `#[repr(pack = "1")]`
+have identical behavior.
 
 # Alternatives
 [alternatives]: #alternatives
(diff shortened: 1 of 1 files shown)


text/0000-repr-pack.md
```diff
@@ -0,0 +1,93 @@
...
+parsed as a `u64`. The restrictions on where this attribute can be placed along
+with the accepted values are:
+
+* Custom packing can only be specified on `struct` declarations for now.
+  Specifying a different packing on perhaps `enum` or `type` definitions should
+  be a backwards-compatible extension.
+* Packing values must be a power of two.
+
+By specifying this attribute, the alignment of the struct would be the smaller
+of the specified packing and the default alignment of the struct otherwise. The
```
cmr: s/otherwise//

text/0000-repr-pack.md
```diff
@@ -0,0 +1,94 @@
...
+`#[repr(packed)]` and `#[repr(pack = "1")]` should have identical behavior.
+
+Because this lowers the effective alignment of fields in the same way that
+`#[repr(packed)]` does (which caused https://github.com/rust-lang/rust/issues/27060 ),
+while accessing a field should be safe, borrowing a field should be unsafe.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This would unfortunately make my life easier even though one of the unstated
```
cmr: Please remove this paragraph.

text/0000-repr-pack.md
```diff
@@ -0,0 +1,94 @@
...
+  `#[repr(packed)]` with manual padding, although such structs would always have
+  an alignment of 1 which is often wrong.
+* Alternatively a new attribute could be used such as `#[pack]`.
+* `#[repr(packed)]` could be extended as either `#[repr(packed(N))]` or
+  `#[repr(packed = "N")]`.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* The behavior specified here should match the behavior of MSVC at least. Does
```
cmr: This must be resolved. Verify against the behavior of clang (or gcc).

text/0000-repr-pack.md
```diff
@@ -0,0 +1,94 @@
...
+  it match the behavior of other C/C++ compilers as well?
+* Should it still be safe to borrow fields whose alignment is less than or equal
```
cmr: It should have the same behavior of `#[repr(packed)]` I think.

cmr:
The feature is necessary. If this behavior is in line with common C compilers, :shipit:

text/0000-repr-pack.md
```diff
@@ -0,0 +1,93 @@
...
+[summary]: #summary
+
+Extend the existing `#[repr]` attribute on structs with a `pack = "N"` option to
+specify a custom packing for `struct` types.
+
+# Motivation
+[motivation]: #motivation
+
+Many C/C++ compilers allow a packing to be specified for structs which
+effectivally lowers the alignment for a struct and its fields (for example with
```
cmr: s/effectivally/effectively/

text/0000-repr-pack.md
```diff
@@ -0,0 +1,93 @@
...
+MSVC there is `#pragma pack(N)`). Such packing is used extensively in certain
+C/C++ libraries (such as Windows API which uses it all over the place making
+writing Rust libraries such as `winapi` a nightmare).
```
cmr: s/a nightmare/challenging/

text/0000-repr-pack.md
```diff
@@ -0,0 +1,93 @@
...
+Extend the existing `#[repr]` attribute on structs with a `pack = "N"` option to
+specify a custom packing for `struct` types.
+
+# Motivation
+[motivation]: #motivation
+
+Many C/C++ compilers allow a packing to be specified for structs which
+effectivally lowers the alignment for a struct and its fields (for example with
+MSVC there is `#pragma pack(N)`). Such packing is used extensively in certain
+C/C++ libraries (such as Windows API which uses it all over the place making
```
cmr: s/all over the place/pervasively/

text/0000-repr-pack.md
```diff
@@ -0,0 +1,93 @@
...
+writing Rust libraries such as `winapi` a nightmare).
+
+At the moment the only way to work around the lack of a proper
+`#[repr(pack = "N")]` attribute is to use `#[repr(packed)]` and then manually
+fill in padding which is a burdensome task. Even then that isn't quite right
+because the overall alignment of the struct would end up as 1 even though it
+needs to be N (or the default if that is smaller than N), so this fills in a gap
+which is basically impossible to do in Rust at the moment.
```
cmr: s/basically //

text/0000-repr-pack.md
retep998: Specifically clang or gcc on a platform other than Windows. On Windows they are supposed to match the msvc ABI.

DanielKeep:
For reference, here's [an example of Clang doing this](http://goo.gl/PeYrl5).

retep998:
See [his comment](https://github.com/rust-lang/rfcs/pull/1358#issuecomment-162403439) for an interesting interaction edge case with the alignment RFC regarding required alignment.

nikomatsakis:

ubsan:
This seems right to have... although, I think it should be `#[repr(packed = "N")]`. So the default of `#[repr(packed)]` would be `#[repr(packed = "1")]`.

tsion:
@ubsan `#[repr(packed(1))]` is not currently valid attribute syntax (but `#[repr(packed = "1")]` is). I guess the RFC mention this, so cc @retep998 as well.

It could become valid with https://github.com/rust-lang/rfcs/pull/1559, though, or as @nrc mentions in a comment, the procedural macros RFC.

retep998:
I don't mind the syntax either way, as long as I can have this attribute at all.

starkat99:
I think it should just extend `#[repr(packed)]` rather than adding a new `pack` syntax. The duplication is confusing otherwise. Just have `#[repr(packed)]` be default to `#[repr(packed = "1")]`.

nikomatsakis:
Hmm, I think I too prefer `#[repr(packed = "N")]`. Otherwise, the RFC seems unobjectionable to me. It'd also be nice to approve https://github.com/rust-lang/rfcs/pull/1358, which covers very similar subject matter.

nikomatsakis:
One area the RFC did not (explicitly) address -- or I missed it -- is what to do if there are multiple `packed` specifiers. It did say that `#[repr(pack = "2")]` and `#[repr(packed)]` cannot be combined, but it did not say anything about `#[repr(pack = "2")]` and `#[repr(pack = "4")]`. Presumably this is an error as well? (Unlike the proposed align RFC?)

Is there any reason one might want multiple specifiers? Like, independent macros might add tighter and tighter requirements? (However, unlike with alignment, it's not obvious to me whether you would want the max or min packing that was specified.)

retep998:
I think if we did allow specifying the packing multiple times, that the right thing to do would be to use the smallest packing specified. The alignment of any given field is `min(default, packing)` so extending packing to be `min(packing1, packing2)` seems fairly natural.

As for the align RFC, there'd need to be a decision on how it interacts with packing. Whether it does the msvc style `__declspec(align(N))` with required alignment that cannot be lowered by packing, or the gcc style alignment attribute that can be lowered by packing.

nikomatsakis:
**Huzzah!** The @rust-lang/lang team has decided **to accept** this RFC, with the caveat that the attribute be renamed to `packed` (and not `pack`). I will make this change when merging (I assume @retep998 you do not object.)

We discussed a few other things:

- We should (obviously) clarify the relationship to an align attribute. We're going to bring the align RFC into FCP, discussion can happen there. But in the @rust-lang/lang meeting, I think the general feeling was that if both attributes are present, then the specified align would overrule the alignment derived from packing (in other words, we would match the MSVC style). Another option is to forbid both options from being used simultaneously, but that seems unnecessary to me.

nikomatsakis:
Tracking issue: https://github.com/rust-lang/rust/issues/33158

PR #1408, opened 2015-12-14
Delayed resolution

mahkoh:
Generalize the delayed resolution of language items to arbitrary items.

### DIFF
--- a/text/0000-delayed-relosution.md
+++ b/text/0000-delayed-relosution.md
@@ -61,14 +61,14 @@ static S: Type;
 const C: Type;
 ```
 
-where `Bounds` is a collection of trait and lifetime bounds. Users of `T`
-definition can only use the properties exposed by the bounds. And where `Type`
-is an arbitrary (possibly extern) type.
+where `Bounds` is a collection of trait and lifetime bounds. Users of `T` can
+only use the properties exposed by the bounds. And where `Type` is an arbitrary
+(possibly extern) type.
 
 ## Defining extern items
 
 The `#[extern]` attribute can be applied in the `#[extern = path]` form to
-define an extern items. For example
+define an extern item. For example
 
 ```rust
 #[extern = "alloc::Default"]
--- a/text/0000-delayed-relosution.md
+++ b/text/0000-delayed-relosution.md
@@ -1,0 +1,104 @@
+- Feature Name: extern_item
+- Start Date: 2015-12-14
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Generalize the delayed resolution of language items to arbitrary items.
+
+# Motivation
+[motivation]: #motivation
+
+Consider an allocator crate that defines a default allocator:
+
+```rust
+type Default = Jemalloc;
+```
+
+At the point in the dependency tree where the allocator crate is defined, the
+default allocator might not yet exist or downstream crates might want to define
+their own allocator. Users of the allocator crate will, however, want to use the
+default allocator as a default type parameter in their allocating types:
+
+```rust
+struct Vec<T, A: Allocator = alloc::Default>
+```
+
+In order to satisfy both constraints, a delayed resolution of items is added:
+
+```rust
+#[extern]
+type Default: Allocator;
+```
+
+# Detailed design
+[design]: #detailed-design
+
+## Declaring extern items
+
+An `#[extern]` attribute is added. This attribute can be applied to
+
+* type definitions,
+* functions,
+* statics, and
+* constants.
+
+For example
+
+```rust
+#[extern]
+type T: Bounds;
+
+#[extern]
+fn f();
+
+#[extern]
+static S: Type;
+
+#[extern]
+const C: Type;
+```
+
+where `Bounds` is a collection of trait and lifetime bounds. Users of `T`
+definition can only use the properties exposed by the bounds. And where `Type`
+is an arbitrary (possibly extern) type.
+
+## Defining extern items
+
+The `#[extern]` attribute can be applied in the `#[extern = path]` form to
+define an extern items. For example
+
+```rust
+#[extern = "alloc::Default"]
+struct X;
+
+impl Allocator for X { /* ... */ }
+```
+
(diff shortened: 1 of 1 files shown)


jonas-schievink:
`#[extern]` currently doesn't parse as an attribute because `extern` is a keyword

mahkoh:
Another application:

An rng that needs to be initialized from another rng. In a freestanding environment there might be no standard rng available and so the first rng cannot be initialized from a default source and thus it cannot implement Default and thus it cannot be used as a default argument for a hashmap in a freestanding environment, etc., etc.

Ericson2314:
liblog could also really use this. It currently gets around this with boxing, trait objects, and `at_exit`; with this it could just use core https://github.com/rust-lang-nursery/log/issues/13 properly.

Should it work on functions too so we can replace more lang items with this?

mahkoh:
>Should it work on functions too so we can replace more lang items with this?

Everything can be simulated with types, static functions on those types, associated constants and statics or const fns and functions returning mutable references.

>If A declares a public #[extern] type, B defines it (and depends on it of course), does B see the concrete type when it looks at A

```rust
#[extern]
type T: ToString;

#[extern = "T"]
type _Dummy = String;
```

When `B` looks at `_Dummy` it sees `String`. When it looks at `T` it sees some type that implements `ToString`.

>Can #[extern]-declared types be used in the definition of #[extern = ...] definitions?

I don't see why not. At worst the compiler might crash.

jethrogb:
What about specifying defaults? For example in the case of the allocator, I would expect most dependants don't really care about what is chosen.

nrc:

Summary: the RFC proposes a mechanism for abstracting (at compile-time) the values of global variables, type aliases, and functions. Where these items are declared, they are only given bounds, rather than a concrete value and type checking, etc. is performed using these bounds. A downstream crate must provide the concrete value of the item, which is checked against the bounds. The motivation is as a more general alternative for mechanisms for supplying allocators, etc.

There has not been much discussion, you can probably just read it, rather than me summarising.

My opinion is that there is the seed of a good idea here, but the motivation is not strong enough for such a big feature, nor has there been enough of a positive reception to warrant further investigation for now. Therefore, I would like to propose that we move to FCP with a view to close this PR.

- [ ] @nikomatsakis
- [x] @nrc 
- [ ] @aturon 
- [ ] @eddyb 
- [ ] @pnkfelix (on vacation)

Ericson2314:
I recall this RFC being a bit underspecified, but I think the motivation is quite good. In various fora it's been discussed how at least log, panicking, compiler-rt/builtins, and allocators could benefit from this. Thats quite the list.

nikomatsakis:
I agree that we have evolved a number of systems that seem to have this kind of "needs-require" relationship. However, I think that before accepting an RFC around this feature we'd want to address a number of interactions.

- How/if this can be used to subsume allocators, panicking, and builtins.
- A bit more thought into how this relates to the trait system. These items are basically associated items from some global trait -- maybe we can make that link more explicit? If nothing else, giving a "desugaring" to a trait mechanism might be nice.

In general, this direction of "global needs/requires" worries me, because it is so clearly non-composable. (The same holds for the existing instances.) In other words, if I have a library that supplies type `X`, I cannot link it with another library that does the same. Do we want to mitigate that risk in the design somehow? I think we should also discuss how we will describe the purpose of this in order to clarify that, basically, people shouldn't be using it very often.

Along those lines, I am very interested in pursing a kind of "functors-lite" approach that addresses similar problems (and maybe even serves as an alternative). The basic idea is that you can parameterize modules and crates with type/lifetime parameters; this could effectively be desugared to threading those parameters around on all the modules contained within, so it is not equivalent to (e.g.) ML functors, and in particular not "generative functors". (I should sketch out my preliminary thoughts here, I keep alluding to them; ah well, I'll do it again for now.)

An advantage of this trait-based approach would be better composability. A disadvantage would be that it doesn't necessarily apply to panic handling and other global codegen properties. Anyway, even if we wanted both, I think I would prefer to have some kind of appealing trait-based alternative in place before we offer these convenient, but non-compsable, global items as an alternative.

golddranks:
@nikomatsakis A side note regarding a "needs/requires" -kind of system: A way to globally have orphan impls without coherency issues by using such a system was also proposed: https://internals.rust-lang.org/t/another-proposal-for-overcoming-orphan-impl-limitations/3863 However, as you say, the globality makes such schemes non-composable.

Ericson2314:
Ah another usecase I forgot to enumerate a better foundation of main and entry points---libraries (std, "frameworks") can declare the entry point and then call a "needed" function, or provide that needed function and need another.

Ericson2314:
@nikomatsakis I love applicative functors too! We could still handle the truely global things by making some crates only includable once in the dependencies dag, even when applied to different parameters. My biggest worry is this will require a lot of Cargo work to get right---though even plain needs-provides still needs some Cargo work.

pnkfelix:
In addition to the other points made in comments above, relying on the attribute system to express (path-based) references a la `#[extern = "alloc::Default"]`  is going to run afoul of the same problems noted in other RFCs like https://github.com/rust-lang/rfcs/pull/1605#issuecomment-229434796 and https://github.com/rust-lang/rfcs/pull/1327#issuecomment-149302743 , right?

 * The cases are slightly different since here I *think* that `#[extern = Foo]` is mean to be *providing* the definition for `Foo`, while in the other two cases I listed above, the attributes are *referencing* paths that are defined elsewhere. That actually might mean that e.g. hygiene is not an issue? I am not sure.

aturon:
🔔 **This RFC is entering its final comment period**. 🔔 

The lang team's [disposition is to close](https://github.com/rust-lang/rfcs/pull/1408#issuecomment-242997097).

nikomatsakis:
No comments since the end of the FCP. Per the [summary comment](https://github.com/rust-lang/rfcs/pull/1408#issuecomment-242997097), the feeling is that there is a good idea here, but that idea would require more elaboration before we could accept it (as described in the comments that come after that summary comment).

Ericson2314:
https://github.com/ghc-proposals/ghc-proposals/pull/5 is highly relevant for the applicative functors approach.

PR #1419, opened 2015-12-20
Safe memcpy for slices ([T]::copy_from_slice)

ubsan:


Aatch:
Looks good to me. I'd personally prefer only panicking on `src.len() > dst.len()` but slicing `dst` to the appropriate length isn't difficult, so I don't feel strongly about it.

bluss:
I think it's a good call to have a free function for `copy`, the two parameters have equal importance.

ubsan:
@Aatch I also don't feel too strongly about this, but the current design does make this specific use nice:

```rust
// let src;
let dst = unsafe { std::mem::uninitialized() };
std::slice::copy(src, &mut dst); // You know that dst is fully initialized from here on
```

Dr-Emann:
As a note, dlang has special syntax for set and copy, and it only allows arrays of equal size.

e.g.
```d
auto x = new int[10];
auto y = new int[11];
x[] = 5; // fill x with 5's
y[] = x[]; // Array lengths don't match for copy: 10 != 11
```

text/0000-slice-copy-set.md
```diff
@@ -0,0 +1,68 @@
...
+
+We could name these functions something different.
+
+`memcpy` is also pretty weird, here. Panicking if the lengths differ is
+different from what came before; I believe it to be the safest path, because I
+think I'd want to know, personally, if I'm passing the wrong lengths to copy.
+However, `std::slice::bytes::copy_memory`, the function I'm basing this on, only
+panics if `dst.len() < src.len()`. So... room for discussion, here.
+
+These are necessary functions, in the opinion of the author.
```
eddyb: They could be methods on slices, not necessarily free functions.

Amanieu:
I think `fill` would be a better name than `set`. This matches the C++ `std::fill` function which fills a range with copies of a single.

eddyb:
@Amanieu I definitely prefer `xs.fill(x)`. Not sure if there are actual issues around adding methods to slice types.

ubsan:
@eddyb I don't think there are issues, but I don't personally like having these functions as methods. These feel "right" as free functions (at least, `copy` does). I can see `set` as a method called `fill` however.

eddyb:
@ubsan The main reason for `std::slice` and `std::str` (and all of the modules for integers and floats, but I digress) is that once upon a time, there were no inherent `impl`s for primitives, so you had to choose between free functions and traits.
And free functions won for a long time.

I can see how `copy` looks cleaner as a free function, but I think assignment syntax is optimal there.

ubsan:
@eddyb true, for `set`. I think I'll change the RFC to be `fn set -> fn [T]::fill`. I still think copy looks better as a free function, however.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,82 @@
...
+`std::slice::bytes`, which means we can remove this deprecated and useless
+module.
+
+# Detailed design
+[design]: #detailed-design
+
+Add one function to `std::slice`.
+
+```rust
+pub fn copy<T: Copy>(src: &[T], dst: &mut [T]);
```
Manishearth: Could this cause issue when T contains &U? We have to be careful to avoid lifetime lengthening. But I guess variance will fix it.

Manishearth:
Note that free functions are harder to discover doc-wise. The general rust community seems to avoid them in code; though I'm not certain of this.

ubsan:
@Manishearth `copy`, imho, is better as a free function. `copy_from` and `copy_to` are both valid ways of writing `copy` as a method, and are both fairly ugly to write in regular code, imo.

```rust
slice.copy(slice2); // which is src? which is dst?
slice.copy_to(slice2); // okay, it's obvious here, but it's kind of ugly
slice.copy_from(slice2); // same as ^
std::slice::copy(slice, slice2); // nice! also, a connection to std::ptr::copy_nonoverlapping here.
```

Manishearth:
I think `copy_to` or `copy_from` are better (I like `copy_to` better of these, but not for any real reason). They're clearer, don't obfuscate the src/dst, and are methods instead of free fns. `copy(src, dst)` will invariably have you looking up the docs all the time. The `&[T]`/`&mut [T]` distinction helps, but in case both happen to be mutable you're stuck again.

jFransham:
@ubsan equally to your first example:
```rust
std::slice::copy(slice, slice2); // which is src? which is dst?
```
I think this is what @Manishearth was talking about when he mentioned looking up the docs. It's obvious if you've used memcpy for years but for people new to systems programming it will be confusing.

Manishearth:
>  It's obvious if you've used memcpy for years

Not even that, because that convention isn't uniformly followed across programming languages. I think even rust had a deliberately "backwards" memcpy ordering in the past (I think it got fixed in the rush for 1.0  or something)

tomjakubowski:
> @ubsan equally to your first example:
> ```rust
> std::slice::copy(slice, slice2); // which is src? which is dst?
> ```
> I think this is what @Manishearth was talking about when he mentioned looking up the docs. It's obvious if you've used memcpy for years but for people new to systems programming it will be confusing.

The order proposed in the RFC is the opposite of `memcpy`'s, in fact:

```c
void *memcpy(void *restrict dst, const void *restrict src, size_t n);
```

Anyway, in many cases the Rust would look more like:

```rust
std::slice::copy(&slice, &mut slice2);
```

which makes it pretty clear which argument is source and which is destination.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,82 @@
...
+and one function to Primitive Type `slice`.
+
+```rust
+impl<T> [T] where T: Copy {
+    pub fn fill(&mut self, value: T);
+}
+```
+
+`fill` loops through slice, setting each member to value. This will lower to a
+memset in all possible cases. It is defined to call `fill` on a slice which has
```
tomjakubowski: it might be worth mentioning how the behavior is defined :)

ubsan: "defined" meaning "will not exhibit undefined behavior". :P

ubsan:
@Manishearth @tomjakubowski I polled some people in the rust community, and it seems that generally people don't like my free functions, and they've convinced me, so I've decided to switch to methods. This also means that I've switched to the name `copy_from`. Is there anything else that you see?

Amanieu:
Some alternatives that might be worth considering:

1) Allow direct assignment of slices with the same panic semantics as the proposed `copy` function:
```rust
a[..] = b[..];
```
This is currently disallowed because [T] is an unsized type.

2) Make `fill` an iterator method. This would allow it to be used on any collection type instead of just slices. This method would only be available when `Iter::Item` is `&mut T`.

3) Allow any `Clone` type for `copy` and `fill` instead of only `Copy` types. This would make the functions more useful for generic code.

4) Maybe consider the general concept of copying data from one iterator to another. This would allow values to be modified prior to copying by using `map` for example:
```rust
iter::copy(b.iter(), a.iter_mut());
iter::copy(b.iter().map(|x| x * 2), a.iter_mut());
```
The downside is that this might be getting too generic, and it wouldn't make sense to have the proposed panic semantics when the iterator lengths don't match.

Manishearth:
As far as iterators go there's not much value (relatively) there because you can write this as a function already.

The reason this RfC is important is that currently there's no way to copy/fill slices in safe Rust efficiently. You have to use a for loop and hope the optimizer magicks it. This RfC provides a way to do that. For iterators, a one-line for loop just works and can't be optimized further, so there's much less value there.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+Two new methods on `slice`. `[T]::fill` *will not* be lowered to a `memset` in
+any case where the bytes of `value` are not all the same, as in
+
+```rust
+// let points: [f32; 16];
+points.fill(1.0); // This is not lowered to a memset (However, it is lowered to
+                  // a simd loop, which is what a memset is, in reality)
+```
+
+Also, `copy_from` has it's arguments in a different order from it's most similar
```
Manishearth: This can be solved by using copy_to instead. Though I don't feel that this is much of an issue.

Manishearth:
@ubsan mostly looks good, :+1: for this rfc.

Small nit: Perhaps `fill()` should also have a descriptive name (`fill_from` in its current form)?

ubsan:
@Manishearth Hmm, I can see your argument... I do like @eddyb's suggestion of `fill_with`. Are there any other suggestions?

eddyb:
I do think `fill_with` is better than `fill_from` but I'd use `fill_with` for a method taking a closure and calling that closure for each element (which isn't really something we really need, iterators work fine there).
But for the record, I still prefer plain `fill`. I don't see how it could be mistaken for anything else.

ubsan:
@eddyb I prefer `fill_with`, as I think it reads better as a method name. `s1.fill(0)` just doesn't look right to me.

eddyb:
@ubsan It doesn't have to read like English.

ubsan:
@eddyb no, but it sounds better to me as `fill_with`, and so I decided to do it that way. If people really do prefer it as `fill`, I can revert the change, I'm not super attached.

mahkoh:
Some advice from experience with a codebase that uses such a function extensively:

```
~/lrs/lib/src$ ack "mem::copy\b" | wc -l  
38
```

Not a single time do I want it to abort if the lengths don't agree. It is almost always used to copy elements from an initial subset of the source to an initial subset of the destination. Some examples:

```rust
pub fn validate(bytes: &[u8]) -> Result<usize> {
    if bytes.len() < IPV6_SOCK_ADDR_SIZE {
        return Err(error::InvalidArgument);
    }
    let mut family: sa_family_t = 0;
    mem::copy(family.as_mut(), bytes);
    match family as c_int {
        AF_INET6 => Ok(IPV6_SOCK_ADDR_SIZE),
        _ => Err(error::InvalidArgument),
    }
}
```

```rust
impl<'a> Read for &'a [u8] {
    fn read(&mut self, buf: &mut [d8]) -> Result<usize> {
        let n = mem::copy(buf, (**self).as_ref());
        *self = &self[n..];
        Ok(n)
    }
```

ubsan:
@mahkoh it's easy enough to cut down `src` and `dst` to size, and I like the guarantee that "`self` is guaranteed to be fully filled"

ubsan:
Also, after sleeping on it, I have started to agree with @eddyb. `fill` is my favorite name for the `memset`function.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+`copy` panics if `src.len() != self.len()`, then `memcpy`s the members into 
+`self` from `src`. Calling `copy_from` is semantically equivalent to a `memcpy`;
+`self` can have uninitialized members, and `self` is guaranteed to be fully filled
+afterwards. This means, for example, that the following is fully defined:
+
+```rust
+let s1: [u8; 16] = unsafe { std::mem::uninitialized() };
+let s2: [u8; 16] = unsafe { std::mem::uninitialized() };
+s1.fill(42);
+s2.copy_from(s1);
```
bluss: should be `&s1` on this line.

ollie27:
Do we really need both `copy_from` and `clone_from_slice`? Surely we should have one and rely on the compiler to use `memcpy` if it can.

`fill` should definitely take `T: Clone` to make it more generic, I expect the compiler to optimise it to a `memset` if possible.

bluss:
@ollie27 The compiler can optimize down to memcpy and memset, but only in simple cases. There's an example of a case where it doesn't optimize to memcpy here https://github.com/rust-lang/rust/issues/27750#issuecomment-162710779 (function copy_composite).

We could have `clone_from_slice` but special case it to memcpy if the type is actually copy, using specialization or a compiler hack (similar to needs_drop). Most seemed dismayed that I would mention such a thing.. :smile: but I just like stating options. We're a low level programming language after all, we have plenty of pragmatism to do.

It seems useful to have a function that is actually memcpy, but in high level safe rust. Lots have expressed concern that they want guarantees, not having to rely on a sufficient smart compiler.

whataloadofwhat:
I think there's use for a `memcpy` function which copies as much as possible from `src` to `dst`, rather than panicking if they aren't the same size. Sure it's easy enough to make them the same size, but in safe code that would mean unnecessary bounds checking (as far as I know?). Maybe another method?

```rust
pub fn copy_from_???(&mut self, src: &[T]) {
    unsafe {
        ptr::copy_nonoverlapping(src.as_ptr(), self.as_mut_ptr(), std::min(self.len(), src.len()));
    }
}
```

I don't know what it could be called though.

eddyb:
@whataloadofwhat slicing the slices before calling `copy_from` should optimize to the same thing.

briansmith:
> Most seemed dismayed that I would mention such a thing

Actually, AFAICT only a very small number of people expressed concern about it. I think ad-hoc specialization for this case is much better than having a bad API, because it can be replaced with the "good" kind of specialization later.

> I think there's use for a memcpy function which copies as much as possible from src to dst

Looking at my own code, I agree. But, usually I've already split the input slices into parts so that I can process the remainder that isn't copied, or that wasn't overwritten. There's no one function that I could write that captures all the different cases without also computing unnecessary results.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+    pub fn copy_from(&mut self, src: &[T]);
+}
+```
+
+`fill` loops through slice, setting each member to value. This will lower to a
+memset in all possible cases. It is defined behavior to call `fill` on a slice
+which has uninitialized members, and `self` is guaranteed to be fully filled
+afterwards.
+
+`copy` panics if `src.len() != self.len()`, then `memcpy`s the members into
```
nagisa: `copy_from`.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+
+```rust
+let s1: [u8; 16] = unsafe { std::mem::uninitialized() };
+let s2: [u8; 16] = unsafe { std::mem::uninitialized() };
+s1.fill(42);
+s2.copy_from(&s1);
+println!("{}", s2);
+```
+
+And the program will print 16 '8's.
```
nagisa: What? You filled one array of `42: u8`, copied it to another array of the same amount of `u8`s and now it prints `8` 16 times?

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+Two new methods on `slice`. `[T]::fill` *will not* be lowered to a `memset` in
+any case where the bytes of `value` are not all the same, as in
+
+```rust
+// let points: [f32; 16];
+points.fill(1.0); // This is not lowered to a memset (However, it is lowered to
+                       // a simd loop, which is what a memset is, in reality)
+```
+
+Also, `copy_from` has it's arguments in a different order from it's most similar
```
nagisa: This whole paragraph seems irrelevant?

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
+- Feature Name: std::slice::copy, slice::fill
```
nagisa: s/std::slice::copy, slice::fill/slice_copy_fill/ or something along the lines. Should be something you could use in `#[feature(..)]`.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+`copy_nonoverlapping`, and the design decision should not be extended to
+`copy_from`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+We could name these functions something else. `fill`, for example, could be
+called `set`, `fill_from`, or `fill_with`.
+
+`copy_from` could be called `copy_to`, and have the order of the arguments
```
nagisa: Seems irrelevant now?

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+different from what came before; I believe it to be the safest path, because I
+think I'd want to know, personally, if I'm passing the wrong lengths to copy.
+However, `std::slice::bytes::copy_memory`, the function I'm basing this on, only
+panics if `dst.len() < src.len()`. So... room for discussion, here.
+
+`fill` and `copy_from` could both be free functions, and were in the
+original draft of this document. However, overwhelming support for these as
+methods has meant that these have become methods.
+
+These are necessary, in my opinion. Much unsafe code has been written because
```
nagisa: That’s not an alternative.

nagisa:
All in all the idea sounds good to me. Would be nice to have.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,99 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Two new methods on `slice`. `[T]::fill` *will not* be lowered to a `memset` in
+any case where the bytes of `value` are not all the same, as in
+
+```rust
+// let points: [f32; 16];
+points.fill(1.0); // This is not lowered to a memset (However, it is lowered to
+                       // a simd loop, which is what a memset is, in reality)
```
nagisa: I think the best we can guarantee here is an equivalent to `rep movsd`, which is what i386 (the common denominator, and AFAIR, the default target for 32bit x86 compilations) supports.

text/0000-slice-copy-fill.md
ubsan: "`self` is guaranteed to be fully filled afterwards". This is important for unsafe code. It is easy to slice down `src` in client code if necessary.

text/0000-slice-copy-fill.md
ubsan: This was an example, showing that something is fully defined even with uninitialized memory.

text/0000-slice-copy-fill.md
ubsan: I can take it out, or put it in alternatives.

text/0000-slice-copy-fill.md
ubsan: Will change.

text/0000-slice-copy-fill.md
ubsan: This would follow `ptr::copy_nonoverlapping` ordering.

text/0000-slice-copy-fill.md
ubsan: `rep movsd` is `memcpy`, not `memset`.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,88 @@
...
+
+```rust
+impl<T> [T] where T: Copy {
+    pub fn fill(&mut self, value: T);
+    pub fn copy_from(&mut self, src: &[T]);
+}
+```
+
+`fill` loops through slice, setting each member to value. This will lower to a
+memset in all possible cases. It is defined behavior to call `fill` on a slice
```
bluss: I would tone down the guarantee about lowering to memset. llvm does nice optimizations and for example lowers `for elt in slice { *elt = None::<i32>; }` to memset, but that's only with optimizations enabled. It's not trivial to replicate or guarantee in all compilation modes.

ubsan: agreed. I was a little hasty.

text/0000-slice-copy-fill.md
ranma42: I think @nagisa meant that it would be quite surprising to see such program printing `8` instead of `42` (in fact it would definitely make me think that something was not initialised properly :P ).

bluss:
I think the name needs to change to `copy_from_slice` to be consistent with `clone_from_slice` in libstd.

ubsan:
@bluss I don't think so. I'd assert that `copy_from` is a common enough operation that it's really, really annoying to type out `copy_from_slice`. It's already annoying to type out `ptr::copy_nonoverlapping`.

I also have no idea if it should be a free function. The issue I've found with that is it's difficult to tell which is src and which is dst.

bluss:
@ubsan The alternative is very inconsistent though? How can we fail to achieve consistency when these two things are decided so closely together?

ubsan:
@bluss I'm personally not a fan of `clone_from_slice`, and believe it should be called `clone_from`. It's exactly the same operation as `Clone::clone_from`, it should be exactly the same name.

alexcrichton:
Thanks again for the RFC @ubsan! I like the thrust here and I think these have a good chance of eventually becoming stable. Some tweaks I might make to the APIs are:

* As @bluss mentioned, I'd recommend `copy_from` to be renamed to `copy_from_slice` to match its [recent stabilization](https://github.com/rust-lang/rust/pull/30943). I think it's important to keep these two functions consistent in naming.
* In addition to being consistent in naming, I think the `copy_from_slice` method should panic if the slices are not of equal lengths (as `clone_from_slice` has now been stabilized to do).
* It may be useful to point out that `copy_from_slice` can (and probably should) be explicitly documented as lowering to a `memcpy`. This has been requested for debug builds in the past, and this way it would always happen regardless of optimization level. It also provides a convincing case for it to exist in the face of `clone_from_slice` (which is otherwise more general).
* I'd change the argument to `fill` from `T` to `&T` and change the bound from `Copy` to `Clone`. I think the main reason to be `Copy` instead of `Clone` would be to **guarantee** a `memset` but we can't really do that, so there may not be much to gain over a more restrictive bound. In using `Clone` you can then take advantage of `clone_from` which means that ownership of the argument is never needed. I also believe this will optimize to `memset` as soon as we hit LLVM optimizations.

What do you think of those ideas?

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,83 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently, the only way to quickly copy from one non-`u8` slice to another is to
+use a loop, or unsafe methods like `std::ptr::copy_nonoverlapping`. This allows
+us to guarantee a `memcpy` for `Copy` types, and is safe. The only way to
+`memset` a slice, currently, is a loop, and we should expose a method to allow
+people to do this. This also completely gets rid of the point of
+`std::slice::bytes`, which means we can remove this deprecated and useless
+module.
```
alexcrichton: This motivation may want to be update now that `std::slice::bytes` has been entirely deprecated (this was a recent change though)

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,83 @@
...
+[drawbacks]: #drawbacks
+
+Two new methods on `slice`. `[T]::fill` *will not* be lowered to a `memset` in
+all cases.
+
+# Alternatives
+[alternatives]: #alternatives
+
+We could name these functions something else. `fill`, for example, could be
+called `set`, `fill_from`, or `fill_with`.
```
alexcrichton: It might be useful to cite some other languages here (e.g. C++) which I believe have similarly functioning methods called `fill`

text/0000-slice-copy-fill.md
alexcrichton: The ordering of `copy_nonoverlapping` primarily exists because it's a free function and there is no clear "receiver" as there is with a method call. We ended up stabilizing `(src, dst)` orderings (e.g. for pointer copies, I/O copies, etc). If there is a clear receiver, though (as there is with `self`) the convention here doesn't necessarily apply.

text/0000-slice-copy-fill.md
```diff
@@ -0,0 +1,83 @@
...
+
+`copy_from` could be called `copy_to`, and have the order of the arguments
+switched around. This would follow `ptr::copy_nonoverlapping` ordering, and not
+`dst = src` or `.clone_from()` ordering.
+
+`copy_from` could panic only if `dst.len() < src.len()`. This would be the same
+as what came before, but we would also lose the guarantee that an uninitialized
+slice would be fully initialized.
```
alexcrichton: It may be worth pointing out here that if `copy_from` does this it will diverge from the implementation of `clone_from_slice` (which is pretty similar)

briansmith:
> I'd change the argument to fill from T to &T

Seems reasonable.

> and change the bound from Copy to Clone. I think the main reason to be Copy instead of Clone would be to guarantee a memset but we can't really do that, so there may not be much to gain over a more restrictive bound. In using Clone you can then take advantage of clone_from which means that ownership of the argument is never needed. I also believe this will optimize to memset as soon as we hit LLVM optimizations.

`fill` lowering to `memset` is one of the main points of the proposal and it would be bad to lose it. That was the motivation for the initial proposal on the internals message board, as that proposal was derived from https://github.com/briansmith/ring/blob/master/src/polyfill.rs.

alexcrichton:
I do think it's worth having a guaranteed `memset`, but the `fill` as proposed here I think unfortunately isn't an appropriate signature. I think the signature would have to look something like:

```rust
pub fn fill_bytes(&mut self, byte: u8) where T == u8;
```

For now I at least think that we can't guarantee a safe memset on anything other than `[u8]` in the standard library generically (e.g. we can't safely memset for all `Copy` types or for all one-byte types). I'd be open to ideas for various signatures though!

briansmith:
But, why should it be limited to just `u8`? My polyfill is limited to `u8` because that's what I needed to do to get it working decently, but the idea is that the compiler should be smart enough to optimize it to memset or a faster equivalent (e.g. in the case of `u64` some direct SIMD usage).

ubsan:
@alexcrichton
The idea was to implement the memset ourselves, being generic over the size of the type. This is only an initial implementation.

The second and third points are already the case.

As for the name, I'm going to go to the thread to see if I can get `clone_from_slice` renamed `clone_from`. `copy_from_slice` is an overly long name for a simple operation.

alexcrichton:
@briansmith in theory we'd need some sort of trait for "this type can contain an arbitrary bit pattern" but unfortunately we don't have that. That means that we don't necessarily have any way (offhand) to make this function generic. Doing so would require introducing this, and it could definitely be implemented for all primitive types, but the API would be getting significantly more complex than just `fill` proposed here.

@ubsan in this proposal the `fill` function won't be optimized to memset unless the argument is the same byte pattern all-the-way-through. For example LLVM will likely optimize it if it's one byte, and *may* optimize it if it's something like 0x1d1d. We can't generically say "fill this type with these bytes", however, due to what I mentioned above.

I also disagree that you should strive to rename `clone_from_slice` to `clone_from`. That name is already taken by the `Clone` trait and we've had lots of problems with inherent methods sharing the name of trait methods (e.g. `Option::as_ref` and `AsRef::as_ref`). The `clone_from_slice` method also differs from the `clone_from` method because the `Clone` trait is not implemented for slices. 

The naming between this less general (copy) function at the very least should match that of the more general (clone) variant. Pursuing whether the `_slice` suffix is present is somewhat orthogonal.

bluss:
We can implement a better `fill` with `T: Copy` than with `T: Clone`. We can make it more generic later, if / when speciailzation arrives.

alexcrichton:
@bluss can you clarify? I'm not actually sure how `fill` with a `Copy` bound can be better than one with a `Clone` bound?

ubsan:
@alexcrichton as in, optimize it ourselves. We don't exclusively have access to C functions, we can write our own optimized code :P

bluss:
`T` being memcpy-safe means we can unroll, simdify, etc to our heart's content, we're only blitting bytes.

ubsan:
@alexcrichton The argument that I have is that `Clone::clone_from` == `[T]::clone_from_slice`, except that `[T]` is an unsized type, and so we cannot implement `Clone` for it. `Clone::clone_from` is defined as `dst = src.clone()`, while `[T]::clone_from_slice` is `dst[..] = src[..].clone()`.

alexcrichton:
@bluss, @ubsan 

So, to clarify, I believe that @briansmith is thinking that we have a function which is **guaranteed** to be a `memset` in all builds (specifically including debug builds). I claim that if we have `T: Copy` that is not memory safe because we don't know that the type can actually be set to an arbitrary bit pattern.

That means that at this point we can't guarantee that `fill` is a `memset` unless we have a restricted set of types. That could perhaps be done via a new trait, but for now I personally think that's out of the scope of this RFC as `fill` with `Clone` is useful in its own right. Consequently we could write crazy code for `fill` with `T: Copy` that works element-by-element calling `memcpy` or using SIMD craziness to ensure good performance in debug builds, but I think that this may be diverging from the original motivation of a fast `fill` in debug builds. Any implementation that operates one-element-at-a-time (which it must do so if generic) will be slow for types like `&mut [u8]` that are megabytes in size. 

So overall while we *may* be able to write a faster `fill` for `T: Copy` in debug mode, I'm not sure that it would be worth it.

Does that make sense? Do you think that there are tangible benefits of restricting to `Copy` today?

ubsan:
@alexcrichton A `memset` *is* SIMD craziness. The argument is that we make `fill` a crazy SIMD-ized function, our `memset`, as it were, except generic over size.

No, it wouldn't have benefits today, but it would have major benefits in the future because we could write that crazy SIMD-ized code.

bluss:
@alexcrichton It's as unsound as it is to memcpy arbitrary `T: Copy`. i.e. it's ok. Also, we're not going to guarantee a call to memset. The point is that we can improve something like "extended memset" ourselves.

alexcrichton:
@ubsan, @bluss If we don't guarantee `memset`, then why do we want to hinder ourselves? Surely LLVM can do better than we at the source level can ever do because of monomorphization?

I guess what I don't understand is that if we don't guarantee `memset`, then why are we hamstringing ourselves to `Copy` when `Clone` is the more general bound here? If we're talking about speed then we're only talking about optimized code, and do you think that we can write a *generic* `fill` that is *faster* than what LLVM would already produce?

bluss:
`T: Copy` is not a hinderance but a boon to the implementor, i.e. more freedom for them.

`slice.fill(x)` with `T: Clone` would be equal with `for elt in slice { *elt = x.clone(); }`, so there's not much to gain by putting that into a function.

With `T: Copy` we *can* provide something of better quality and better performance, even in debug builds.

Specialization will change the parameters we're looking at here, but we don't know if / when that arrives.

bluss:
The `T: Copy` version can call memset if size of T is 1, so already that is an advantage for debug build performance. No guarantee for T of all sizes or so, but it's still a big improvement.

alexcrichton:
@bluss 

> so there's not much to gain by putting that into a function.

I somewhat disagree with this personally because I think the same argument could be made about `clone_from_slice`, e.g. it's just a 

```
for (a, b) in dst.iter_mut().zip(src) { *a = b.clone(); }
```

These kinds of conveniences on core types (like slices) feel to me like they're often useful enough to provide.

Your point about guaranteeing a `memset` if the size of `T` is one seems pretty nice though! To me that's a pretty convincing argument for having *some* variant with `T: Copy` as it'll cover all the easy use cases.

That means in terms of naming, however, that we may want both a `Copy` and a `Clone` variant for filling a slice as we do with copying a slice (e.g. this is proposing a `Copy` alternative to the `clone_from_slice` method). Ideally both filling via clone and copy would both exist and have a similar naming scheme. I'd be fine with just adding the `Copy` fill for now so long as we ensure there's a gap for fill-via-`Clone` to exist.

bluss:
The naive version of the "clone_from_slice" loop produces terrible code and the naive fill loop produces good code. It's good that we have clone_from_slice so that the right way to tickle llvm for an optimized copying is abstracted out into a method. I don't think the pragmatic argument is the only one for clone_from_slice though.

I think .fill(x) is a nice API, but it's not as important as .copy_from().

ubsan:
Agreed, it isn't as important. The only reason I also added `fill` is because `memset` and `memcpy` kinda go together. `copy_from` is the important part of this RFC.

alexcrichton:
In the case that `fill` was a bit of an addition, perhaps that API can be left to its own RFC? (or perhaps an RFC issue for now) It sounds like there's some outstanding questions around it and otherwise `copy_from` seems fine modulo naming.

ubsan:
@alexcrichton I would be fine with that. I don't personally see any pressing need for it.

ubsan:
Yeah, I'm taking out `[T]::fill`. It can be decided upon later.

BurntSushi:
:+1: to this RFC. Guaranteed `memcpy` is good. I do also think it should be named `copy_from_slice` to be consistent with `clone_from_slice`.

alexcrichton:

---

My personal opinion is that this is good to go with the name `copy_from_slice`, the semantics all seem good to me though!

ubsan:
@alexcrichton I don't personally like the name, because it's a very common function (at least in my personal code) to have such a long name. However, I do see the benefit of it in terms of consistency, so I am not opposed to it. I'd just prefer a shorter name.

bluss:
Let's get this in. Thanks for pushing it @ubsan (and I like your new avatar).

Using this to replace most uses of `clone_from_slice` in libstd will improve performance for debug builds by using memcpy directly, so we benefit immediately.

ubsan:
@bluss thanks :) I needed a new avatar very badly...

aturon:
A couple thoughts:

- To clarify some points that @bluss made: the compiler and lang teams were against using a *compiler hack* to do custom specialization for `clone_from_slice`. But the possibility of doing so through [specialization proper](https://github.com/rust-lang/rfcs/pull/1210) is definitely open (and specialization is likely to land in the near future).
  - The main question in actually specializing `clone_from_slice` for `Copy` data is the [divergence](https://internals.rust-lang.org/t/assuming-clone-clone-is-equivalent-to-copy-semantics-for-all-instances-of-copy/3148/) currently possible -- the fact that you can write custom `Clone` impls for `Copy` data. This was necessary for various reasons, and would be difficult to change backwards-compatibly. But we could viably consider this a contract violation, and still give `clone_from_slice` memcpy semantics on `Copy` data.
- On the other hand, there is perhaps some value to having a separate method which is purely a safe memcpy, and in particular asserts the `Copy`-ness of the data -- even if we *also* specialize `clone_from_slice`. I personally could go either way on that question.
- Finally, I agree with the general sentiment that the name of this method should match the idiom for `clone_from_slice`, which was recently stabilized.

alexcrichton:
The libs team discussed this RFC during triage yesterday, and the conclusion was to merge this RFC. We discussed the possibility of having `clone_from_slice` guarantee the same semantics as `copy_from_slice` here if `T` is `Copy` (disregarding possible diverging implementations of copy/clone), but our conclusion was that even if we had that behavior we would *still* want to have `copy_from_slice`. This method is a good static assertion that the operation is indeed a memcpy and we're also not sure how comfortable we would be relying on specialization for semantics like that.

Anyway, I shall merge soon. Thanks again for the RFC @ubsan!

lgvz:
FWIW at this point, I'd most *definitely* prefer the shorter name `copy_from` despite inconsistency with `clone_from_slice`.

PR #1426, opened 2015-12-23
Placement in/box refinement

Stebalien:
Supersedes #1401

[rendered](https://github.com/Stebalien/rfcs/blob/placement-traits/text/0000-placement-traits.md)

### DIFF
--- a/text/0000-placement-traits.md
+++ b/text/0000-placement-traits.md
@@ -1,0 +1,312 @@
+- Feature Name: placement_traits
+- Start Date: 2015-12-21
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC amends RFC #0809 to reduce the number of traits involved, take
+allocators into account, and pin down the story on DST placement.
+
+## Traits
+
+There are now three traits:
+
+1. `Placer` -- Placement in syntax:  `let owner = PLACE <- value`.
+2. `Boxer` -- Box syntax: `let boxed: Box<_> = box value`.
+3. `Place` -- An "out" pointer.
+
+## Allocators
+
+Boxer's now make Places from Allocators. This means that any type implementing
+`Boxer` can be allocated with any Allocator using placement in syntax (see the
+detailed design for more).
+
+## DSTs
+
+`Boxer::make_place` and `Placer::make_place` are bounded by `Data: Sized` to
+future proof against DST placement.
+
+Furthermore, this RFC explicitly defines the guarantees of when/where
+`Placer::make_place` will be called.
+
+# Detailed design
+[design]: #detailed-design
+
+The new trait hierarchy is as follows:
+
+```rust
+/// Interface to implementations of  `PLACE <- EXPR`.
+///
+/// `PLACE <- EXPR` effectively desugars into:
+///
+/// ```rust,ignore
+/// let p = PLACE;
+/// let mut place = Placer::make_place(p);
+/// let raw_place = Place::pointer(&mut place);
+/// let value = EXPR;
+/// unsafe {
+///     std::ptr::write(raw_place, value);
+///     Place::finalize(place)
+/// }
+/// ```
+///
+/// The type of `PLACE <- EXPR` is derived from the type of `PLACE` and the
+/// context. If the type of `PLACE` is `P`, then the final type of the
+/// expression is some owner such that `P` implements `Placer<typeof(EXPR), Owner>`.
+///
+/// Values for types implementing this trait usually are transient
+/// intermediate values (e.g. the return value of `Vec::back`)
+/// or `Copy`, since the `make_place` method takes `self` by value.
+pub trait Placer<Data: ?Sized, Owner> {
+    /// `Place` is the intermedate agent guarding the
+    /// uninitialized state for `Data`.
+    type Place: Place<Data, Owner=Owner>;
+
+    /// Creates a fresh place from `self`.
+    fn make_place(self) -> Self::Place
+        where Data: Sized;
+}
+
+/// Core trait for the `box EXPR` form.
+///
+/// `box EXPR` effectively desugars into:
+///
+/// ```rust,ignore
+/// let mut place = Boxer::make_place(Default::default());
+/// let raw_place = Place::pointer(&mut place);
+/// let value = EXPR;
+/// unsafe {
+///     ::std::ptr::write(raw_place, value);
+///     Place::finalize(place)
+/// }
+/// ```
+///
+/// The type of `box EXPR` is supplied from its surrounding
+/// context; in the above expansion, the result type `T` is used
+/// to determine which implementation of `Boxer` to use, and that
+/// `<T as Boxer>` in turn dictates determines both which `Allocator`
+/// to use and which implementation of `Place` to use.
+pub trait Boxer<Data: ?Sized, A>: Sized
+    where A: Allocator
+{
+    /// The place that will negotiate the storage of the data.
+    type Place: Place<Data, Owner=Self>;
+
+    /// Creates a globally fresh place from a given allocator.
+    fn make_place(allocator: A) -> Self::Place
+        where Data: Sized;
+}
+
+/// Both `PLACE <- EXPR` and `box EXPR` desugar into expressions
+/// that allocate an intermediate "place" that holds uninitialized
+/// state.  The desugaring evaluates EXPR, and writes the result at
+/// the address returned by the `pointer` method of this trait.
+///
+/// A `Place` can be thought of as a special representation for a
+/// hypothetical `&uninit` reference (which Rust cannot currently
+/// express directly). That is, it represents a pointer to
+/// uninitialized storage.
+///
+/// The client is responsible for two steps: First, initializing the
+/// payload (it can access its address via `pointer`). Second,
+/// converting the agent to an instance of the owning pointer, via the
+/// appropriate `finalize` method.
+///
+/// If evaluating EXPR fails, then the destructor for the
+/// implementation of Place is run to clean up any intermediate state
+/// (e.g. deallocate box storage, pop a stack, etc).
+pub unsafe trait Place<Data: ?Sized> {
+    /// `Owner` is the type of the end value of both `PLACE <- EXPR` and
+    /// `box EXPR`.
+    ///
+    /// Note that when `PLACE <- EXPR` is solely used for side-effecting an
+    /// existing data-structure, e.g. `Vec::back`, then `Owner` need not carry
+    /// any information at all (e.g. it can be the unit type `()` in that
+    /// case).
+    type Owner;
+
+    /// Returns the address where the input value will be written.
+    /// Note that the data at this address is generally uninitialized,
+    /// and thus one should use `ptr::write` for initializing it.
+    fn pointer(&mut self) -> *mut Data;
+
+    /// Converts self into the final value, shifting deallocation/cleanup
+    /// responsibilities (if any remain), over to the returned instance of
+    /// `Owner` and forgetting self.
+    unsafe fn finalize(self) -> Self::Owner;
+}
+
+```
+
+First, `box` desugaring constructs the allocator with `Default::default()`. This
+means that `let boxed: Box<_, A> = box value;` works for all allocators `A:
+Default`. It's reasonable to construct new allocators on the fly like this
+because allocators are intended to be passed to collection constructors by
+value.
+
+Additionally, we define the following blanket impl that turns every `Allocator`
+into a `Placer` for all `Boxer`s.
+
+```rust
+impl<T, D, B> Placer<D, B> for T
+    where T: Allocator,
+          B: Boxer<D, T>
+{
+    type Place = B::Place;
+
+    fn make_place(self) -> Self::Place
+        where Self: Sized
+    {
+        B::make_place(self)
+    }
+}
+```
+
+This means that `let boxed_thing: Type<_> = HEAP <- thing` works out of the box
+as long as `HEAP` is an `Allocator` and `Type` implements `Boxer`.
+
+Finally, to support DST placement, this RFC explicitly loosens the placement
+protocol guarantees. Specifically, the place in placement in/new is not
+guaranteed to be allocated before the evaluation of the expression on the right
+hand side. DST placement needs this to be able to compute the size of the DST
+before allocating the place. This means that, in the following cases, whether or
+not the `Box` is ever allocated is explicitly undefined:
+
+```rust
+let _: Box<_> = box panic!();
+let _: Box<_> = HEAP <- panic!();
+```
+
+For completeness, I've included the following DST placement traits to
+demonstrate that the current placement traits are compatible with DST placement.
+Note: A concrete design for returning DSTs is well outside the scope of this
+RFC.
+
+```rust
+trait DstPlacer<Data: ?Sized, Placer>: Placer<Data, Output> {
+    fn make_place_dynamic(self, layout: Layout) -> Self::Place;
+}
+
+trait DstBoxer<Data: ?Sized, A>: Placer<Data, A> where A: Allocator {
+    fn make_place_dynamic(allocator: A, layout: Layout) -> Self::Place;
+}
+
+impl<T, D, B> DstPlacer<D, B> for T
+    where T: Allocator,
+          B: DstBoxer<D, T>
+{
+    type Place = B::Place;
+
+    fn make_place_dynamic(self, layout: Layout) -> Self::Place {
+        B::make_place_dynamic(self, layout)
+    }
+}
+
+// Assuming specialization.
+default impl<D, O> Placer<D, O> for T where T: DstPlacer<D, O> {
+    fn make_place(self, layout: Layout) -> Self::Place
+        where Self: Sized,
+    {
+        self.make_place_dynamic(Layout::new::<D>())
+    }
+}
+
+default impl<D, A> Boxer<D, A> for T
+    where T: DstBoxer<D, A>
+          A: Allocator
+{
+    fn make_place(allocator: A) -> Self::Place
+        where Self: Sized,
+    {
+        self.make_place_dynamic(Layout::new::<D>())
+    }
+}
+```
+
+## Choices
+
+### Placer::make_place
+
+`Placer::make_place` takes self by value. Taking self by reference as discussed
+in #1286 would either require an option dance, HKT to properly handle lifetimes
+(`type Place<'a> = ...`), or shenanigans. Furthermore, taking self by value
+ensures that non-copy placers are used only once. This allows panic-free
+allocation using `try!(place) <- thing`.
+
+### Parameterize Placer with Owner
+
+`Owner` is now a type parameter in `Placer`. This allows placement in syntax to
+be used with allocators. That is:
+
+```rust
+let boxed_thing: Box<_> = HEAP <- thing1;
+let rced_thing: Rc<_> = HEAP <- thing2;
+```
+
+If `PLACER <- thing` can have precisely one type as in the original RFC, it
+wouldn't be possible to produce both `Rc`s and `Box`s (see alternatives for an
+alternative solution).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Placer::make_place self by value
+
+Taking self by-value means that `vec <- value` won't work. However, `vec.back()
+<- value` still works just fine so I'm not convinced this is a problem.
+
+The other drawback is that `arena <- value` won't work either. One solution is
+to autoref the placer if necessary but,
+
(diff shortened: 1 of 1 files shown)


text/0000-placement-traits.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+We could then use `let boxed = HEAP.emplace::<Type<_>>() <- value;` to select
+the output type. However, IMO, this is much less ergonomic.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Does `Place::pointer` need to return an `&mut T` (or some other special pointer)
+to ensure RVO? Should it return `std::ptr::Unique`?
```
eddyb: I almost feel like `Unique<T, A>` should be `Place<T, A>`'s "filled" counterpart, if both were structs and `A` described how the allocation should be handled.

text/0000-placement-traits.md
```diff
@@ -0,0 +1,312 @@
...
+    fn make_place(self) -> Self::Place
+        where Self: Sized
+    {
+        B::make_place(self)
+    }
+}
+```
+
+This means that `let boxed_thing: Type<_> = HEAP <- thing` works out of the box
+as long as `HEAP` is an `Allocator` and `Type` implements `Boxer`.
```
eddyb: Why are we still using `HEAP` as an example?
Isn't there a name like `DefaultAllocator` in one of the allocator RFCs we can pick up and move on?

text/0000-placement-traits.md
```diff
@@ -0,0 +1,312 @@
...
+///
+/// ```rust,ignore
+/// let p = PLACE;
+/// let mut place = Placer::make_place(p);
+/// let raw_place = Place::pointer(&mut place);
+/// let value = EXPR;
+/// unsafe {
+///     std::ptr::write(raw_place, value);
+///     Place::finalize(place)
+/// }
```
eddyb: For inference to work, all of this (and the `box` desugaring below) should be nested expressions (no blocks and statements), either with some safe `write_and_finalize` function (harder to enforce optimizations like we have on `std::ptr::write`) or with a special HIR node that is lowered to the appropriate MIR/LLVM IR.

text/0000-placement-traits.md
Stebalien: #39 (closed). It's not mentioned anywhere else.

eddyb:
To expand on the special HIR node idea:
```rust
PLACE <- EXPR // gets desugared into:
ExprEmplace(Placer, PLACE, EXPR)
```
```rust
box EXPR // gets desugared into:
ExprEmplace(Boxer, ExprCall(Default::default, []), EXPR)
```

This way, we could guarantee that 0 unnecessary copies are being made, DST support can be added without having to come up with impossible source desugarings and, probably the best part, we can employ a simpler form of rust-lang/rust#27292 only for `ExprEmplace`:

If we have an expected type, i.e. `ExprEmplace(Placer, _, E): P<A, B, C>`, we check that `P<A, B, C>: Placer<T>`.
In case we do get a match, `T: Sized` must hold and we can use `T` as the expected type for `E`.

Otherwise, it's either a legitimate error, or our `P<A, B, C>` can only be produced by unsizing the result of `ExprEmplace` (e.g. `Box<Trait>`). We can assume it's not an error and attempt to make it compile.

Thankfully, the unsizing coercions relevant here only involve structs (and `Box`, which we can treat as a struct).
This simplifies our job greatly and if more types support unsizing coercions in the future, we can simply choose not to handle them here.

We can take the most conservative stance and replace all type params in `P<A, B, C>` with inference variables to get a condidate for our missing intermediary smart pointer, e.g. `P<X, Y, Z>.

Then, we require that `P<X, Y, Z>: CoerceUnsized<P<A, B, C>> + Placer<T>`.
This will link up all the type params and allow us to type-check the emplace expression.

 E.g. for `Box<Trait, A>`, we require `Box<X, Y>: CoerceUnsized<Box<Trait, A>> + Boxer<T>` which will result in `X = T` and `Y = A` and `T: Trait` obligation.
In the end we get our intermediary smart pointer, `Box<T, A>`, with a `Boxer` impl we can get the associated `Place` from, and a coercion to `Box<Trait, A>`.

Unlike rust-lang/rust#27292, this also works if you replace `Trait` with `[T]`, and does not affect coercions in existing stable code.

mahkoh:
Why would the Allocator be constructed in the desugaring and the Boxer have an Allocator parameter? One can simply

```rust
impl<T, A> Boxer<T> for Box<T, A> where A: Allocator+Default
```

Having the Boxer trait know anything about allocators is completely unnecessary and a huge restriction.

mahkoh:
@eddyb Doesn't move_val_init already do what you want? Except that move_val_init doesn't unconditionally get translated to a memcpy for some reason.

eddyb:
@mahkoh The problem is that you can't desugar to `move_val_init(ptr, EXPR)` because it requires `unsafe` (and you don't want to allow unsafe code in `EXPR`) and it has to be its own statement in a nested block so coercion propagation is pretty much screwed.

Stebalien:
@mahkoh, how is this a restriction? `Boxer

mahkoh:
The Boxer implementation can acquire an allocator (if it even needs one) by means other than `Default`.

Stebalien:
@mahkoh, sorry, hit enter too soon. See my updated post.

mahkoh:
@Stebalien All of this can be done without the restrictions imposed by having Boxer know about allocators. E.g.

````rust
let a: Box<_, A1> = box val; // where A1: Default

let a = Box::with_pool(CustomAllocator::new(options)) <- thing;
```

mahkoh:
The first line is equivalent to

```rust
let a = Box::with_pool(A1::default()) <- val;
```

Stebalien:
@mahkoh, which means that one has to implement a new `Placer`. Regardless, I still don't see how this is a *restriction*. What is the use case for using `box value` without an `Allocator`? Note, the proposed system still supports your syntax regardless.

mahkoh:
>Note, the proposed system still supports your syntax regardless.

It doesn't support `box` without a `Default` allocator.

>What is the use case for using box value without an Allocator?

The object might want to use a globally shared allocator stored in a static variable or might want to construct an allocator with custom options.

mahkoh:
>which means that one has to implement a new Placer. 

And where is the problem with that?

mahkoh:
You also don't need a "new" Placer. See the following pseudocode

```rust
struct Box<T: ?Sized, A>
    where A: Allocator,
{
    alloc: A,
    val: *mut T,
}

struct BoxBuf<T, A>
    where A: Allocator,
{
    alloc: A,
    val: *mut T,
}

impl<T, A> Boxer<T> for Box<T, A>
    where A: Allocator+Default,
{
    type Place = BoxBuf<T, A>
    fn make_place() -> BoxBuf<T, A> {
        let mut alloc = A::default();
        /* ... */
        BoxBuf {
            alloc: alloc,
            val: val,
        }
    }
}

impl<T, A> Placer<T, Box<T, A>> for BoxBuf<T, A>
    where A: Allocator,
{
    type Place = Self;
    fn make_place(self) -> Self { self }
}

impl<T, A> Place<T> for BoxBuf<T, A>
    where A: Allocator,
{
    type Owner = Box<T, A>;
    fn pointer(&mut self) -> *mut T { self.val }
    unsafe fn finalize(self) -> Box<T, A> {
        Box {
            self.alloc,
            self.val,
        }
    }
}
```

It seems that what you actually want is a default implementation of `Placer` for all `Place` implementations.

mahkoh:
To be quite honest, I don't think there is a need for the Placer trait at all. The only use it has is that it can be implemented for Allocators so that you can use the `HEAP <- val` syntax to construct arbitrary containers.

This can also be done with the Place trait alone:

```rust
Box::with_pool(HEAP) <- val
```

So the only advantage seems to be the shorter syntax. The most common cases are

* the `box val` case if there is no need for error handling
* the `try!(xyz) <- val` case if there is a need for error handling

Neither of these cases are served by the Placer trait. There is also the problem with that Allocators have no idea how to construct containers and that, the Placer implementation for Allocators is limited by this. This inversion then leaks into the placement traits by having them suddenly depend on Allocators.

Apart from this forced dependency, allocators and placement traits are orthogonal.

eddyb:
@mahkoh also consider `vec <- x` and `hash_map.entry(k) <- v`.

mahkoh:
@eddyb The first one doesn't work because Placer consumes self. I'll think a bit about the second one.

eddyb:
@mahkoh it would work if `vec: &mut Vec<T>` and/or `<-` does autoref on the LHS.

mahkoh:
@eddyb Is there a reason that `Place::pointer` cannot do the heavy lifting? (Destroying the current entry in the hashmap if it exists, reserving a slot in the vector.)

Stebalien:
@eddyb, note, as of this RFC, it has to be `vec.back() <- x` (unless we get auto referencing).

@mahkoh, FYI, it's `PLACER <- thing;`, not `PLACE <- thing;`.

Re your previous comments, this really just comes down to the common. With the system proposed in the RFC, one would write:

```rust
let boxed: Box<_> = HEAP <- value;
let object = Object::from_object_pool() <- value;
```

With your system, one would write:

```rust
let boxed: Box<_> = Box::with_allocator(HEAP) <- value;
let object: Object = box value;
```

I'll try to come up with a way to make both cases simpler.

eddyb:
@mahkoh The primary concern @pnkfelix had, IIRC, is handling panics.

mahkoh:
@eddyb I see. Well, I'm not bothered much by the Placer trait itself. It's the idea to implement it for Allocators that seems wrong.

Stebalien:
> @eddyb Is there a reason that Place::pointer cannot do the heavy lifting? (Destroying the current entry in the hashmap if it exists, reserving a slot in the vector.)

Also, `Place::pointer` takes self by reference so you'd have to handle the case of user code calling pointer more or less than once.

eddyb:
@Stebalien Well that could be made unsafe and the compiler implementation of emplace can do the right thing (TM).

@mahkoh Another "nice" example: `arena <- value` where `arena: &Arena`.

mahkoh:
@Stebalien 

>Also, Place::pointer takes self by reference so you'd have to handle the case of user code calling pointer more or less than once.

Just go full `unsafe` in all signatures. Does anyone expect that these traits will be used much outside of the expansions and possibly unsafe code?

>FYI, it's PLACER <- thing;, not PLACE <- thing;.

I'm aware.

>With your system, one would write:

In the first case you don't have to write the type on the lhs because `Box::with_allocator` already contains all the type information:

```rust
let boxed = Box::with_allocator(HEAP) <- value;
```

Stebalien:
Sorry, I assumed that we still had a `Placer` to call an unsafe `make_place` on. Assuming we don't have a `Placer` trait and `Vec::back()` returns an `Place`, what happens when the user drops the `Place` without emplacing (`drop(vec.back())`, `back()` is not unsafe)? All `Place`s would have to keep track of whether or not they have allocated or allocate on construction (not necessarily a bad thing...).

---

As for depending on Allocator, I think I see your point. Having a generic operation (box) depend on allocation is unfortunate. However, I do like `HEAP <- stuff` so how about adding an `Allocate` trait? If you implement `Allocate`, you get `allocator <- thing` for all allocators and `box thing` for all allocators implementing `Default`. Alternatively, you can just implement `Boxer` if you don't want to take an allocator.

```rust
trait Allocator {/* Stuff */}

pub unsafe trait Place<Data: ?Sized> {
    type Owner;

    fn pointer(&mut self) -> *mut Data;

    unsafe fn finalize(self) -> Self::Owner;
}

/// Implement on all types that can be allocated.
///
/// This gives you both `let boxed: Self = ALLOC <- thing` and `let boxed: Self = box thing` (for
/// all allocators implementing `Default`.
trait Allocate<Data: ?Sized>: Sized {
    type Place: Place<Data, Owner=Self>;
    type Allocator: Allocator;

    fn allocate(with: Self::Allocator) -> Self::Place
        where Data: Sized;
}

trait Placer<Data: ?Sized, Owner> {
    type Place: Place<Data, Owner=Owner>;
    fn make_place(self) -> Self::Place
        where Data: Sized;
}

trait Boxer<Data: ?Sized>: Sized {
    type Place: Place<Data, Owner=Self>;

    fn make_place() -> Self::Place
        where Data: Sized;
}

impl<T, D, A> Placer<D, A> for T
    where T: Allocator,
          A: Allocate<D, Allocator=T>
{
    type Place = A::Place;

    fn make_place(self) -> Self::Place
        where Self: Sized
    {
        A::allocate(self)
    }
}

impl<A, D> Boxer<D> for A
    where A: Allocate<D>,
          A::Allocator: Default
{
    type Place = A::Place;

    fn make_place() -> Self::Place
        where Self: Sized
    {
        A::allocate(Default::default())
    }
}
```

text/0000-placement-traits.md
```diff
@@ -0,0 +1,312 @@
...
+}
+```
+
+We could then use `let boxed = HEAP.emplace::<Type<_>>() <- value;` to select
+the output type. However, IMO, this is much less ergonomic.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Does `Place::pointer` need to return an `&mut T` (or some other special pointer)
```
pnkfelix: I'm pretty sure it would be a mistake to have it return `&mut T`, since you'd have to add all the caveats like "there isn't *actually* a `T` behind the reference, so don't call methods on it ... or swap it out with another `T` ..."

mahkoh:
Why is `Data` not an associated type of `Place`? The traits could be a lot simpler if it were:

```rust
pub unsafe trait Place {
    type Owner;
    type Data: ?Sized;

    fn pointer(&mut self) -> *mut Self::Data;
    unsafe fn finalize(self) -> Self::Owner;
}

pub trait Placer<P>
    where P: Place,
{
    fn make_place(self) -> P;
}

pub trait Boxer: Sized {
    type Place: Place<Owner = Self>;

    fn make_place() -> Self::Place;
}
```

Stebalien:
@mahkoh, good point about place. However, `trait Placer<P: Place>` will cause inference problems because it would allow multiple implementations of `Placer` with the same input/output (with different  intermediate `Place`s). Also, `Boxer` without a type parameter wouldn't allow things like:

```rust
impl<T> Boxer for OpaqueBox { // T isn't used
    type Place = MyPlace<T>
    fn make_place() -> Self::Place { /* */ }
}
```

Stebalien:
@mahkoh, what do you think about my proposed `Allocate` trait?

text/0000-placement-traits.md
```diff
@@ -0,0 +1,315 @@
...
+
+There are now three traits:
+
+1. `Placer` -- Placement in syntax:  `let owner = PLACE <- value`.
+2. `Boxer` -- Box syntax: `let boxed: Box<_> = box value`.
+3. `Place` -- An "out" pointer.
+
+## Allocators
+
+Boxer's now make Places from Allocators. This means that any type implementing
```
nagisa: s/Boxer’s/Boxers/?

text/0000-placement-traits.md
```diff
@@ -0,0 +1,315 @@
...
+/// uninitialized storage.
+///
+/// The client is responsible for two steps: First, initializing the
+/// payload (it can access its address via `pointer`). Second,
+/// converting the agent to an instance of the owning pointer, via the
+/// appropriate `finalize` method.
+///
+/// If evaluating EXPR fails, then the destructor for the
+/// implementation of Place is run to clean up any intermediate state
+/// (e.g. deallocate box storage, pop a stack, etc).
```
nagisa: pop a stack *frame*

text/0000-placement-traits.md
nagisa: To my knowledge HEAP et al are supposed to be removed ASAP (or never, ever stabilised), therefore this example is not good.

Also, to my knowledge, there should not be things like `HEAP` in the standard library at all, thus making all this `Allocator` thing redundant somewhat?

All in all, to me it seems like this whole section is a bit hasty and we should at least wait for `Allocator` design to get landed first (which doesn’t seem to be finishing anytime soon).

nagisa:
All in all, everything related to `Allocator` looks hasty (since we do not have any design related to them landed yet) but the changes to the placement traits look okay overall.

text/0000-placement-traits.md
Stebalien: The original intention of `PLACE <- thing` was to allow placing into custom allocators. I'm probably going to change this to https://github.com/rust-lang/rfcs/pull/1426#issuecomment-167160921 which can be added after the allocator API is merged (I think).

text/0000-placement-traits.md
Stebalien: Fixed.

Stebalien:
@nagisa 
> All in all, everything related to Allocator looks hasty (since we do not have any design related to them landed yet) but the changes to the placement traits look okay overall.

Replying again here as my comment was swallowed in the push. IIRC, the original motivation of `in (PLACE) box EXPR` (or the new `PLACE <- EXPR`) was placing in a custom allocator so making sure it's compatible with that use case is important. Note: This RFC makes no assumptions about the definition of the `Allocator` trait, just that one exists. Regardless, I'm going to update the RFC to use the traits I proposed in https://github.com/rust-lang/rfcs/pull/1426#issuecomment-167160921 which means that the allocator integration can be implemented later.

Stebalien:
All (specifically, @nagisa),

I've updated the proposal to move the allocator related logic out of the `Boxer` into the `Allocate` trait and addressed the need for a `Placer` trait.

text/0000-placement-traits.md
```diff
@@ -0,0 +1,362 @@
...
+    /// and thus one should use `ptr::write` for initializing it.
+    fn pointer(&mut self) -> *mut Self::Data;
+
+    /// Converts self into the final value, shifting deallocation/cleanup
+    /// responsibilities (if any remain), over to the returned instance of
+    /// `Owner` and forgetting self.
+    unsafe fn finalize(self) -> Self::Owner;
+}
+
+/// All types implemeinting this trait can be constructed using using the `box`
```
pnkfelix: typo: "implemeinting" --> "implementing"

text/0000-placement-traits.md
```diff
@@ -0,0 +1,362 @@
...
+    }
+}
+```
+
+Finally, to support DST placement, this RFC explicitly loosens the placement
+protocol guarantees. Specifically, the place in placement in/new is not
+guaranteed to be allocated before the evaluation of the expression on the right
+hand side. DST placement needs this to be able to compute the size of the DST
+before allocating the place. This means that, in the following cases, whether or
+not the `Box` is ever allocated is explicitly undefined:
```
pnkfelix: hmm i guess i don't have any way to improve upon this, but I find it worrisome.

Stebalien: The place-expression is still evaluated first. That is, given, `PLACER <- THING`: `PLACER` will always be evaluated before `THING`. This just means that the compiler gets to choose when `Placer::make_place(PLACER)` is evaluated.

I don't like it but don't see any alternative that supports DST placement.

pnkfelix: hmm that seems like a subtle distinction you are making versus what the text says ... that is, I read the phrase "the place in placement in/new" as talking about the LHS of `P <- V` ... is there some way to rephrase to make this clearer?

Stebalien: Sorry about that. What I mean and what I say are not always the same thing. Does it make more sense now?

nikomatsakis:
Can someone summarize the state of this RFC with respect to the comment thread? Does the text include any desired updates? Are important "non-desired" updates summarized in the alternatives section? I'd like to make decisions on older RFCs like this one =)

Ericson2314:
I recall there being good ideas here, but it will be a *lot* easier to evaluate once Allocators are implemented, so I suggest waiting for that.

Stebalien:
There are also a few updates I want to make to take advantage of specialization. Should I close this for now and re-open it when allocators land (or are you planning on stabilizing placement new before the allocation API)?

pnkfelix:
Yeah I haven't moved on landing allocators because I was waiting until the pieces were in place to support them on the library types (see [RFC 1327](https://github.com/rust-lang/rfcs/blob/master/text/1327-dropck-param-eyepatch.md))

But now the most fundamental language change for that has landed, so I hope to make forward progress there soon.

I don't think you need to close this RFC in the meantime.

Ericson2314:
Yay!

ubsan:
@nrc @aturon @pnkfelix 

This would be a good RFC for postponement, predicated on getting allocators.

aturon:
@ubsan Agreed -- let's revisit this down the line, when more of the substrate has stabilized.

@rfcbot fcp postpone

aturon:
OK, I'm going to close this as postponed. Work on placement has largely stalled; we should probably revisit the whole question at some point.

PR #1429, opened 2015-12-27
Feature gate extern fn methods

Aatch:
[Rendered](https://github.com/Aatch/rust-rfcs/blob/extern-fn-methods/text/0000-extern-fn-methods.md)

### DIFF
--- a/text/0000-extern-fn-methods.md
+++ b/text/0000-extern-fn-methods.md
@@ -1,0 +1,51 @@
+- Feature Name: gate-extern-fn-methods
+- Start Date: 2015-12-27
+- RFC PR: (leave this empty)
+- Rust Issue: #30235
+
+# Summary
+[summary]: #summary
+
+Restrict the use of non-Rust ABI methods to a feature gate.
+
+# Motivation
+[motivation]: #motivation
+
+Currently we allow this code:
+
+```rust
+trait Foo {
+    extern fn foo(&self);
+}
+```
+
+In the absense of an ABI string, `extern fn` defaults to using "C". This functionality does not
+work well and appears to frequently be a mistake on the part of the user. Since the decision to
+support this feature either way should be subject to discussion, this RFC is only proposing the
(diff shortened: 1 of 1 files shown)


nagisa:
SGTM.

rphmeier:
This seems very reasonable, but I'd be interested in seeing the fallout. Are there a lot of bindings to C libraries which use this feature?

nagisa:
I know we have received at least one bug report about this feature in conjunction with `no_mangle` once our symbol duplication detection has landed.

RalfJung:
Doesn't this RFC effectively break all Rust crates that wrap C libraries, e.g. <https://crates.io/crates/openssl-sys/> and <https://crates.io/crates/libusb-sys/> and [many more](https://crates.io/search?q=-sys)?

remram44:
@RalfJung No it doesn't. This RFC only aim to feature-gate extern *methods*, not functions.

RalfJung:
Oh, I see :)

nikomatsakis:
@Aatch do you have an impl? if you can whip one up, I can do a crater run, feels like necessary data before we make a decision one way or the other. But I'm inclined to :+1: here.

nrc:
Vaguely related, should rustfmt change `extern` to `extern "C"` or vice versa: https://github.com/rust-lang-nursery/rustfmt/issues/451

nrc:
And +1 for a crater run

eddyb:
This is no longer necessary AFAIK, all methods should be compatible with all ABIs now.

nikomatsakis:

I should note that, personally, I'm inclined not to accept, based on @eddyb's comment.

nikomatsakis:
After discussion in the @rust-lang/lang meeting today, we've decided not to accept this RFC. There doesn't seem to be any particular semantic concerns about allowing methods to have their own ABI, and the implementation is even working as far as we know, so retroactively feature gating is not necessary. Thanks @Aatch!

PR #1437, opened 2015-12-29
ESC byte escape

Ticki:
[Rendered](https://github.com/Ticki/rfcs/blob/esc-escape/text/0000-esc-escape.md)

### DIFF
--- a/text/0000-esc-escape.md
+++ b/text/0000-esc-escape.md
@@ -1,0 +1,38 @@
+- Feature Name: esc_escape
+- Start Date: 2015-12-29
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add a byte escape `\e` as shorthand for 0x1B (ESC), similar to GCC's `\e`.
+
+# Motivation
+[motivation]: #motivation
+
+ANSI escape codes are a vital part of CLIs. These are used both in stdin (special keys, control commands, etc.) and stdout (colors, text formatting, change in previous text, movement of the cursor, etc.). Adding a byte escape for it will make it easier to read the source code of and write interactive CLIs.
+
+Many compilers do already have this byte escape, GCC is an example (however, `\e` is not a part of ISO C).
+
+# Detailed design
+[design]: #detailed-design
+
+Add a byte escape `\e` as shorthand for 0x1B (ESC) to characters, strings, byte characters, and byte strings. Furthermore, add it to the escape iterators.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Add a character escape for CSIs
+
+Add a character escape for Control Sequence Introducers (`\x1B[`). This covers the majority of ANSI escape codes and do not require the extra `[` (as `\e` do), but it do not cover _all_ cases (for example, `\e^`, `\e_`, `\eN`, `\eO`, `\eP` and so on).
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None.


Stebalien:
:+1: I don't think we should add too many of these but this one is actually useful.

text/0000-esc-escape.md
```diff
@@ -0,0 +1,38 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Add a byte escape `\e` as shorthand for 0x1B (ESC) to characters, strings, byte characters, and byte strings. Furthermore, add it to the escape iterators.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None.
```
withoutboats: There is a definite drawback here; adding more escape codes increases the incidental complexity of the language (and [graydon cited some other common escapes as something he was glad Rust did not have](﻿﻿https://graydon2.dreamwidth.org/218040.html)).

withoutboats:
I would have obviously made use of this a lot in the terminal emulator I am writing.

Ticki:
Reference implementation: https://github.com/rust-lang/rust/pull/30647

retep998:
Since anyone who is working with the terminal really should be using a library to handle colors for them instead of doing it themselves (that way it might actually work on Windows if the terminal library supports the console API there), that means most Rust code won't have to deal with inserting escape codes directly.  The few number of libraries that would benefit from this combined with the fact that this merely saves a couple characters going from `\x1b` to `\e` means I feel that this addition is not worth the complexity. Therefore I am opposed to this RFC.

quodlibetor:
This seems like it would cause silent (non-compile) breakage? Anyone who had a literal `\e` in any string would suddenly have an `ESC` byte instead. It seems unlikely that that would be common, but it also seems plausible that it would happen, and crater couldn't catch it.

sfackler:
@quodlibetor Use of `\e` is currently a compilation error: http://is.gd/TiW8bS

Ericson2314:
-1. All our extant single-character escapes map to printable characters (and ones that are far more commonly used). It seems inconsistent to give some control characters a shorthand but not others.

Ticki:
@Ericson2314 We already got carriage return (`\r`), which is not printable, so adding `\e` is no exception.

petrochenkov:
The current escapes are a minimal set of universally useful characters, everything else is either ancient or niche (`\e` is in this category), so I'm mildly against any extensions to this set.

retep998:
@Ticki `\r` goes hand in hand with `\n` and they are very much printable in that they create a new line. It's not just Windows that uses `\r\n`, many protocols use it including HTTP and IRC.

Ticki:
> either ancient or niche

I disagree. `\e` is neither ancient or niche. It's a very useful escape, since any modern terminal support ANSI escape codes. They're the universal standard for interacting non-trivially with the terminal.

Ticki:
@retep998 That's right, but what I'm trying to say is that it's not printable at its own. It's there for convinience, just like this proposal.

withoutboats:
Its bad practice to print ANSI code directly instead of using a library that properly evaluates terminfo sheets to identify the correct code. Obviously there are exceptions to this rule, but maybe Rust shouldn't be promoting this practice.

mahkoh:
>Its bad practice, even in UNIX, to print ANSI code directly instead of using a library that properly evaluates terminfo sheets to identify the correct code.

ANSI codes are independent of the terminal.

notriddle:
:thumbsdown:

DanielKeep:
> They're the **universal standard** for interacting non-trivially with the terminal.

They do *nothing* on Windows.  I've actually had problems with crates that blindly spew out ANSI escape codes, generating garbage on Windows.  I don't see any reason to *encourage* this further.

If Rust had a standard ANSI emulator in std, I'd probably feel differently.  Given that they should really only be used in unusual cases, or in a terminal library, it seems a little niche.

Then again, it's not like we're *hurting* for escapes, and it's not an unprecedented addition.  Mostly, I don't like these "they're universally useful" arguments.

mahkoh:
>They do nothing on Windows. I've actually had problems with crates that blindly spew out ANSI escape codes, generating garbage on Windows. I don't see any reason to encourage this further.

Then maybe it is the other way around. Since the windows terminal is breaking the universal standards, the use of windows shouldn't be encouraged further. It probably depends on whether you consider terminals or windows more important. We're probably lucky that windows treats "\n" correctly. Otherwise people would be arguing that the "\n" escape shouldn't be allowed either.

notriddle:
> Since the windows terminal is breaking the universal standards, the use of windows shouldn't be encouraged further.

1. Yeah, good luck with that.
2. Even on POSIXy platforms, you don't want to dump escape sequences down pipes; a library will identify the terminal type, but println itself won't.

mahkoh:
>Yeah, good luck with that.

Luck has nothing to do with this. Only the conscious not to have your [freedoms](http://www.gnu.org/philosophy/free-sw.en.html) infringed by low quality proprietary software (e.g. the windows "terminal".)

>Even on POSIXy platforms, you don't want to dump escape sequences down pipes; a library will identify the terminal type

A program that wants to print escape codes can do the same thing by calling isatty or similar.

```
~/c/coreutils-8.24/src$ ack isatty | wc -l
32
```

This is well known.

Ticki:
> They do nothing on Windows. I've actually had problems with crates that blindly spew out ANSI escape codes, generating garbage on Windows. I don't see any reason to encourage this further.

That's true, but they're still the universal standard. That Windows breaks it doesn't change that.

> Then maybe it is the other way around. Since the windows terminal is breaking the universal standards, the use of windows shouldn't be encouraged further. It depends on whether you consider terminals or windows more important

That's a very bad idea. I agree that it's a terrible design choice by Windows, but it wouldn't make sense to give less support for Windows terminals, since they widely used, and Rust already has pretty good support of Windows.

Ticki:
> Luck has nothing to do with this. Only the conscious decision not to have your freedoms infringed by low quality proprietary software (e.g. the windows "terminal".) If the windows terminal respected your freedoms, the free software community would have improved it and added support for the ANSI standard a long time ago.

This is non sense. Windows is very widely used, and that's a fact, whether you like it or not. Ultimately, we should support Windows just as much as Unix.

notriddle:
Windows is a tier 1 platform, and if that changes it will be because of a lack of demand, not because it's a proprietary platform or because it doesn't adhere to POSIX.

Rust is not a GNU project. It's goal is to allow people to write safe, low-level code. It's goal is not to spread the use of free software. Notice that it's licensed under a non-copyleft license.

Ticki:
I agree with @notriddle. This is completely out of the scope of this RFC. Support for Windows is a choice made for Rust, changing this is very unlikely.

Ticki:
On the matter of terminals: Only a very few (at least that I know of) Windows users use terminals. Often these types of people are using Unix. So I believe the demand for terminal support of Windows is very low (correct me if I'm wrong).

mahkoh:
>Ultimately, we should support Windows just as much as Unix.

Who is this "we"? Is every user of the rustc compiler forced to only write software that runs on windows? What is nonsense is saying "the windows terminal is broken and everyone has to write windows compatible software and for that reason '\e' will not be added". By the same reasoning '\n' should be removed as the windows editor is also broken.

>It's goal is to allow people to write safe, low-level code.

Exactly. Cross platform and windows compatible are not among the requirements to be forced upon the users.

>Notice that it's licensed under a non-copyleft license.

The licenses are not optimal but rustc is certainly free software.

>This is completely out of the scope of this RFC. Support for Windows is a choice made for Rust, changing this is very unlikely.

Rust does not support windows because rust is a programming language. Rustc, which is a compiler, supports windows. And the libraries distributed with this compiler support windows. "C" does not support linux or openbsd or hurt. gcc does that.

The decision to add a feature to a programming language is in most cases completely independent of particular implementations of the language as long as said feature doesn't restrict implementations. And the feature discussed in this RFC certainly does not.

notriddle:
PowerShell?

Ticki:
> Who is this "we"? Is every user of the rustc compiler forced to only write software that runs on windows?

No, that's not what I'm saying, I'm just stating that Rust is supporting Windows, and that's an explicit goal of language.

>  Rust does not support windows because rust is a programming language. Rustc, which is a compiler, supports windows. And the libraries distributed with this compiler support windows. C does not support linux or openbsd or hurd. gcc does that.

"Rust" as rustc + rust + libstd does support Windows.

mbrubeck:
**Moderator note:** This discussion is straying pretty far from the specific benefits or costs of this RFC.  I think the point has been made that this feature may not be useful (or even detrimental) for some cross-platform software, and on the other hand that it may be useful for software targeting certain platforms.  Let's not let this get too abstract or grow into a flame war.

Ticki:
@mbrubeck Agree. This Windows/Unix discussion is off-topic of this RFC.

notriddle:
@mbrubeck Agreed.

Ticki:
@mbrubeck Feel free to delete my comments if you find them inappropriate or off-topic.

withoutboats:
> ANSI codes are independent of the terminal.

Though essentially all modern virtual terminals implement a subset of ECMA-48 ("ANSI" escape codes), as well as their own custom subcommands, which commands they support remains inconsistent between terminals, and it is bad practice to emit a hard-coded escape code instead of using a library like ncurses or Rust's term library.

tbu-:
I must say that I didn't find it hard to just use `\x1b` for the escape character.

Ticki:
@tbu-: The same could be said for \n and other escapes. That it is easy to remember doesn't change the usefulness of the proposal.

BurntSushi:
I'm with @retep998 on this one. :-1:

mahkoh:
>it is bad practice within UNIX-likes to emit a hard-coded escape code instead of using a library 

The very purpose of having a standard is that you can use its contents without having to use a library that detects the terminal type and emits terminal specific codes. But maybe you're trying to say that, since standards cannot be relied upon, it cannot be relied upon that `\e` is the same as `\x1b`?

mahkoh:
>But maybe you're trying to say that, since standards cannot be relied upon, it cannot be relied upon that \e is the same as \x1b?

If the terminal isn't configured to use UTF-8, then that's a much bigger problem . Yet, everyone seems to be fine with `println!` printing UTF-8 to the terminal. Should `println` also be removed in favor of a library that detects the terminal encoding and transforms the output accordingly? This is not a joke. A larger terminal program I've worked on does actually do exactly this itself.

Unlike ANSI not being followed this is a problem that exists in real life. Isn't it hypocritical that you don't want to allow "\e" because "a library should be used" but nobody has a problem with `println!` (without escape code) being used directly?

retep998:
I believe people are okay with `println!` because printing to the console is such a necessary and important thing for so many programs, but at the same time implementing encoding support for all the various encodings out there would take up a huge amount of binary size (except on Windows which only has one wide encoding).

mahkoh:
>but at the same time implementing encoding support for all the various encodings out there would take up a huge amount of binary size 

On POSIX platforms, libc contains these conversion methods.

mahkoh:
No that's not actually correct. I meant the iconv library.

notriddle:
According to @DanielKeep, ANSI not being followed *is* a problem in real life.

mahkoh:
Giving particular weight to a single "terminal" is also hypocritical because nobody is giving particular weight to a single "editor" which cannot display "\n" correctly.

mahkoh:
In fact, more weight should be given to the problems with Notepad than to the problems with the Windows console since most windows users probably use the notepad more often than the console.

tbu-:
Interesting question, do we automatically output `\r\n` line endings on Windows?

tbu-:
@Ticki Mh.. of course it's a tradeoff, how many control characters do we include with their own letter and which we shouldn't. I guess the current ones are a common denominator of what other languages support. If you say that GCC C supports `\e`, then I guess ANSI C doesn't? Quickly checking other languages: Python does not support it, Ruby does, Go doesn't.

Ticki:
ANSI C does not. It's a GCC extension.

notriddle:
Nope. That would require a switchable "binary mode" vs "text mode" (like
Perl has, for example).

withoutboats:
> The very purpose of having a standard is that you can use its contents without having to use a library that detects the terminal type and emits terminal specific codes. But maybe you're trying to say that, since standards cannot be relied upon, it cannot be relied upon that \e is the same as \x1b?

You are misreading me. The ECMA-48 escape protocol standard is not consistently implemented across the virtual terminals used in UNIX systems, and so it is bad practice not to use a library. You cannot be confident that a standardized escape code is actually supported across all the "ANSI" terminals your program may run on. This "standard" is just not much of a standard (fwiw ANSI withdrew their standard in 1997).

Ticki:
This discussion is going completely off the track. If this trend continues, I'll take the freedom to close this issue. It's not very important anyways.

nagisa:
> This discussion is going completely off the track. If this trend continues, I'll take the freedom to close this issue. It's not very important anyways.

Looks on-topic enough to me.

---

I personally haven’t used these codes much and when I needed `\x1b` I just used `\x1b`. 

The problem I see with this proposal is mostly not the escape sequence itself, but the fact that introducing it wouldn’t really make everybody just use `\e`. Most of people would still use `\x1b` and some would use `\e` resulting in some segmentation. I see it as a pretty low-gain addition with significant trade-offs.

text/0000-esc-escape.md
```diff
@@ -0,0 +1,38 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Add a character escape for CSIs
+
+Add a character escape for Control Sequence Introducers (`\x1B[`). This covers the majority of ANSI escape codes and do not require the extra `[` (as `\e` do), but it do not cover _all_ cases (for example, `\e^`, `\e_`, `\eN`, `\eO`, `\eP` and so on).
```
nagisa: This is non-option, because then code like `'\e'` would be invalid (character must be a valid UTF-8 codepoint).

Ticki: s/character escape/string escape/

mbrubeck:
**Moderator note:** Deleted a handful of comments that were either meta-discussion or veering off to discuss features unrelated to the `'\e'` character escape. At this point I'd like to ask people to refrain from commenting unless they have new information or arguments specifically about this RFC. Also keep in mind that use cases that are unimportant to you may be important to other Rust users, and vice-versa. Thanks!

Ticki:
I will close this, since I don't really think it is worth it anymore.

PR #1449, opened 2016-01-07
Add language support for bitfields

mahkoh:
Add support for bitfields via a single per-field attribute.

### DIFF
--- a/text/0000-bitfields.md
+++ b/text/0000-bitfields.md
@@ -1,0 +1,340 @@
+- Feature Name: bitfield
+- Start Date: 2016-01-07
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add support for bitfields via a single per-field attribute.
+
+# Motivation
+[motivation]: #motivation
+
+Bitfields are commonly encountered when interacting with C code. Since many of
+the details of the behavior of bitfields are left to the implementation, it is
+hard to write cross-platform rust code that uses bitfields correctly.
+
+Currently, such code often manually collapses bitfields into a single integer
+field and attempts to extract and set the individual fields via bitshifts. This
+approach works for simple bitfields and when the full set of target platforms is
+known, however, once the bitfields become more complicated or if the code is
+compiled for an unexpected target, such code will silently fail. In particular,
+code that assumes that bitfields are placed into the integer in increasing bit
+order will likely break on big-endian machines.
+
+Appending A contains various examples of the (to some) surprising behavior of
+bitfields on the x86-64 linux platform.
+
+This rfc introduces a single attribute that can be applied to fields inside of
+`#[repr(C)]` structs to turn them into bitfields.
+
+# Detailed design
+[design]: #detailed-design
+
+The attribute `#[bitfield(N)]` is added. Here, `N` can be any constant
+expression that evaluates to a value of type `usize`.
+
+The attribute can be applied to fields inside a `#[repr(C)]` struct which have
+any of following types:
+
+* `u8`, `u16`, `u32`, `u64`, `usize`,
+* `i8`, `i16`, `i32`, `i64`, `isize`.
+
+For example:
+
+```rust
+#[repr(C)]
+pub struct perf_event_attr {
+    /* ... */
+
+    #[bitfield(1)] pub disabled: u64,
+    #[bitfield(1)] pub inherit: u64,
+    #[bitfield(1)] pub pinned: u64,
+    #[bitfield(1)] pub exclusive: u64,
+    #[bitfield(1)] pub exclude_user: u64,
+    #[bitfield(1)] pub exclude_kernel: u64,
+    #[bitfield(1)] pub exclude_hv: u64,
+    #[bitfield(1)] pub exclude_idle: u64,
+    #[bitfield(1)] pub mmap: u64,
+    #[bitfield(1)] pub comm: u64,
+    #[bitfield(1)] pub freq: u64,
+    #[bitfield(1)] pub inherit_stat: u64,
+    #[bitfield(1)] pub enable_on_exec: u64,
+    #[bitfield(1)] pub task: u64,
+    #[bitfield(1)] pub watermark: u64,
+    #[bitfield(1)] pub precise_ip: u64,
+    #[bitfield(1)] pub mmap_data: u64,
+    #[bitfield(1)] pub sample_id_all: u64,
+    #[bitfield(1)] pub exclude_host: u64,
+    #[bitfield(1)] pub exclude_guest: u64,
+    #[bitfield(1)] pub exclude_callchain_kernel: u64,
+    #[bitfield(1)] pub exclude_callchain_user: u64,
+    #[bitfield(1)] pub mmap2: u64,
+    #[bitfield(1)] pub comm_exec: u64,
+    #[bitfield(39)] pub __reserved_1: u64,
+
+    /* ... */
+}
+```
+
+`N` must evaluate to an integer in the range `[0, sizeof(T)]` where `T` is the
+declared type of the field. It specifies the width of the field.
+
+Many details of the behavior of such fields are left to the implementation.
+However, the implementation is encouraged to follow the behavior of the dominant
+C ABI on the target platform.
+
+Those details that are specified are taken from the C11 standard and reproduced
+here for completeness. However, some of the details specified in said standard
+are omitted here since they are implied by the `#[repr(C)]` attribute and thus
+already covered by the previous paragraph.
+
+Concurrent access to different bitfields inside of an object is a data race
+unless,
+
+* none of the accesses mutate the bitfields,
+* the accesses are synchronized,
+* the bitfields are separated (in declaration order) by a non-bitfield, or
+* the bitfields are separated (in declaration order) by a bitfield of size 0.
+
+For example:
+
+```rust
+#[repr(C)]
+struct X {
+    #[bitfield(8)] pub a: u8,
+    #[bitfield(8)] pub b: u8,
+    #[bitfield(0)] pub c: u8,
+    #[bitfield(1)] pub d: u8,
+}
+```
+
+In this structure, `a` and `b` must not be modified concurrently. `a` and `d`
+can be accessed concurrently.
+
+Values stored in a bitfield have a binary representation using the `m` bits it
+occupies in its memory location.
+
+The `&` and `&mut` operators cannot be applied to bitfields.
+
+The type of a bitfield is its declared type. Reading from a bitfield returns a
+value of the declared type. Only values of the declared type can be written to
+a bitfield. The bits of the binary representation of a bitfield are interpreted
+as the bits of a signed or unsigned integer of the width of the bitfield. In
+particular, a signed bitfield of non-zero size contains a sign bit. When a
+zero-sized bitfield is read, `0` is returned.
+
+For the purpose of overflow checking, bitfields are treated as signed or
+unsigned integers of their declared width. In this context, the only value that
+should be written to a bitfield of size `0` is `0`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Some have suggested using attributes at the top of the structure definition as
+follows:
+
+```rust
+#[repr(C; bitfields(foo: 2, bar: 1))]
+struct A {
+    foo: c_uint,
+    bar: c_uint
+}
+```
+
+This is infeasible given the number of bitfields that can be contained in a
+single structure.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None come to mind.
+
+# Appendix A
+
+This appendix contains C code which should be compiled with either clang or gcc
+for the x86-64 linux platform.
+
+The basic setup is as follows:
+
+```c
+#include <stdio.h>
+
+union {
+ 	struct {
+ 		unsigned char a: 1;
+ 		unsigned char b: 1;
+ 		unsigned char :0;
+ 		unsigned char c: 1;
+ 	} bf;
+	unsigned short val[2];
+} x;
+
+int main(void) {
+	x.bf.a = 1;
+	x.bf.b = 1;
+	x.bf.c = 1;
+
+	printf("0x%04x_%04x\n", x.val[1], x.val[0]);
+}
+```
+
+The following cases will modify the definition of the `x.bf` struct and show the
+output of the program.
+
+## Case 1
+
+```c
+ 	struct {
+ 		unsigned char a: 1;
+ 		unsigned char b: 1;
+ 		unsigned char :0;
+ 		unsigned char c: 1;
+ 	} bf;
+```
+
+**Output:** `0x0000_0103`
+
+This shows the behavior of fields of size `0`:
+
+>As a special case, a bit-field structure member with a width of 0 indicates
+>that no further bit-field is to be packed into the unit in which the previous
+>bit-field, if any, was placed.
+
+## Case 2
+
+```c
+	struct {
+		unsigned char a: 4;
+		unsigned char b: 4;
+		unsigned char :0;
+		unsigned char c: 1;
+	} bf;
+```
+
+**Output:** `0x0000_0111`
+
+This is the expected output.
+
+## Case 3
+
+```c
+	struct {
+		unsigned char a: 4;
+		unsigned char b: 5;
+		unsigned char :0;
+		unsigned char c: 1;
+	} bf;
+```
+
+**Output:** `0x0001_0101`
+
+Here we see the behavior if two adjacent bitfields are too large for their
+storage unit. The `b` field is moved entirely to the second byte. However, this
+behavior is not mandated:
+
+>If insufficient space remains, whether a bit-field that does not fit is put
+>into the next unit or overlaps adjacent units is implementation-defined.
+
+## Case 4
+
+```c
+	struct {
+		unsigned char a: 4;
+		unsigned short b: 5;
+		unsigned char :0;
+		unsigned char c: 1;
+	} bf;
+```
+
+**Output:** `0x0001_0011`
+
+Now that we've changed the type of `b` from `char` to `short`, `b` is placed in
+the first byte. It is unexpected that declaring a larger type decreases the
+position of the field in memory. The explanation is that `a` and `b` are
+allocated in a single storage unit of size `2`, the size of the `short` type.
+The size of the storage units allocated for bitfields is implementation defined.
+It is perfectly legal for the following struct to occupy a megabyte of memory:
+
+```c
+struct {
+    unsiged char: 1;
+};
+```
+
+>An implementation may allocate any addressable storage unit large enough to
+>hold a bit-field.
+
+## Case 5
+
+```c
+	struct {
+		unsigned char a: 1;
+		unsigned char b: 1;
+		unsigned char :0;
+		unsigned short c: 8;
+	} bf;
+```
+
+**Output:** `0x0000_0103`
+
+Here we can see that declaring the end of a storage unit with a zero-sized field
+does not imply that the following bitfield is properly aligned for its type.
+
+## Case 6
+
+```c
+	struct {
+		unsigned char a: 1;
+		unsigned char b: 1;
+		unsigned short :0;
+		unsigned char c: 1;
+	} bf;
+```
+
+**Output:** `0x0001_0003`
+
+On the other hand, changing the type of the zero-sized field itself can change
+the alignment of the following fields. This, too, is not required by the
+standard:
+
+>The alignment of the addressable storage unit is unspecified.
+
+## Case 7
+
+```c
+	struct {
+		unsigned char a: 1;
+		unsigned char b: 1;
+		unsigned char :0;
+		unsigned short c: 9;
+	} bf;
+```
+
+**Output:** `0x0001_0003`
+
+This is similar to case 5, but see case 3.
+
+## Case 8
+
+```c
+	struct {
+		unsigned short a: 1;
+		unsigned short b: 1;
+		unsigned char :0;
+		unsigned short c: 1;
+	} bf;
+```
+
+**Output:** `0x0000_0103`
+
+Another consequence of the size of the storage unit being unspecified. Even
+though we request a new storage unit with a zero-sized field, and even though
+all non-zero fields are `short`, the fields on both sides of the break are
+packed into a single `short` storage unit.
--- a/text/0000-bitfields.md
+++ b/text/0000-bitfields.md
@@ -23,7 +23,7 @@ compiled for an unexpected target, such code will silently fail. In particular,
 code that assumes that bitfields are placed into the integer in increasing bit
 order will likely break on big-endian machines.
 
-Appending A contains various examples of the (to some) surprising behavior of
+Appendix A contains various examples of the (to some) surprising behavior of
 bitfields on the x86-64 linux platform.
 
 This rfc introduces a single attribute that can be applied to fields inside of


burdges:
I like bit fields.  Just some additional alternatives:

Idea 1.  Use a bitfield generator that builds an object with methods for manipulating bits from a bitfield specification.  It could be either a macro or an external program, similar to protobuf or captnproto.  As a result, it'd look pretty different from other rust syntax, but maybe that's an advantage. 

Idea 2.  There might be future enhancements to the type system that allow a slightly prettier presentation as type parameters.  I doubt this fits rust system that well, but just mentioning it.  It might look like : 

```
pub struct perf_event_attr {
   pub pub disabled: bitfield_u64<1>,
   ...
}
```

retep998:
@burdges I already use a macro to handle bitfields in winapi. It is a bit more verbose than I'd like due to the lack of ident concatenation but it works fine otherwise https://github.com/retep998/winapi-rs/blob/master/src/usp10.rs#L78-L124

mahkoh:
The rfc was written to be as unintrusive as possible so that adding it to the language proper doesn't add any complexity that could be avoided by using compiler plugins. In particular:

* It doesn't add any new syntax
* The only new feature is that one cannot apply `&` to such fields. But similar things are already implemented for packed structs.
* The attribute can only be used inside of `repr(C)` structs where the behavior is unambiguous (modulo C ABI) so that the design space for bitfields elsewhere is not restricted.

The approach with external tools is problematic:

* Macros can't do it.
* External tools can't do it since you need access to type information (type aliases) to properly calculate the layout.
* If you can write such a plugin that has access to type information (this is not possible at the moment) then this plugin can only do a subset of what this RFC proposes.

In conclusion:

* This RFC adds little to no complexity to the language.
* This RFC cannot (currently) be implemented in any other way.
* Even if it could, the generated APIs would be significantly worse.

mahkoh:
@retep998 It works fine because you only want to use it on windows and windows bitfields are particularly simple. The examples in the RFC already show that it's not at all simple on other platforms and consider the following example:

```c
struct X {
	char a;
	int b: 8;
	short c;
};
```

On x86_64 linux this struct has a size of `4` but on windows I'd assume that it has a size `12`. How can a macro account for these differences?

text/0000-bitfields.md
```diff
@@ -0,0 +1,340 @@
...
+
+    /* ... */
+}
+```
+
+`N` must evaluate to an integer in the range `[0, sizeof(T)]` where `T` is the
+declared type of the field. It specifies the width of the field.
+
+Many details of the behavior of such fields are left to the implementation.
+However, the implementation is encouraged to follow the behavior of the dominant
```
retep998: Not just encouraged, but probably should be required, since if there's no guarantee it becomes useless for FFI which will likely be the primary consumer of bitfields.

retep998:
@mahkoh Yes, it is indeed 12. I am thankful that Windows has simple rules. Anyway, this RFC looks fine to me. The syntax is reasonable. If winapi's minimum rust version ever increases to a version which has feature in stable, then I'll even use it since it'll make bitfield access more convenient than with the inherent methods of the macro solution.

text/0000-bitfields.md
```diff
@@ -0,0 +1,340 @@
...
+bitfields on the x86-64 linux platform.
+
+This rfc introduces a single attribute that can be applied to fields inside of
+`#[repr(C)]` structs to turn them into bitfields.
+
+# Detailed design
+[design]: #detailed-design
+
+The attribute `#[bitfield(N)]` is added. Here, `N` can be any constant
+expression that evaluates to a value of type `usize`.
```
petrochenkov: Constant expressions in attributes look like a sufficiently large new feature.

mahkoh: Eventually such things should be added anyway especially for attributes that calculate the alignment of types. Until then it's fine if this part is only partially implemented.

petrochenkov: FWIW, I'd like the have a solution which can be later extended to support "repr(Rust)" bitfields with specified and standardized ABI (not "implementation defined"), FFI is certainly not the only application of bitfields.
If constant expressions in attributes are indeed implemented, then this seems to be a good enough for native Rust bitfields as well. (Maybe there are better ways, I don't know)

oli-obk:
While I see the necessity for something like this, I'm strongly against going with the easy way (attributes simulating c bitfields). Ada has a different and much more powerful system. It has [representation specifications](http://www.adaic.org/resources/add_content/standards/12rm/html/RM-13-5-1.html) which are completely independent of the regular specification.

Ada's system has two advantages over the one suggested here:

1. Separation of concerns. Methods are already specified separately from the type's definition. Why not also specify the bitwise representation separately? Maybe the order in the bitwise representation should be different from the order in the structural representation?
2. Exact positioning. Instead of just giving fields a size and moving them around and adding some padding fields manually, you can simply move all fields to the correct location by giving a bit offset denoting the start and end of the field inside the struct. While you loose the easy way to read the size of the field, you know exactly which bits in the struct belong to which field. This could be remedied by optionally allowing a length specification for fields so there's some redundancy guarding against bugs).

mahkoh:
How would you write the following bitfield with your idea:

```c
struct X {
    char a;
    int b: 8;
    short c;
};
```

Make up some syntax if necessary.

oli-obk:
Copying Ada's syntax somewhat:

```rust
struct X {
    a: u8,
    b: i32,
    c: i16,
}

for X use {
     a: 0..8,
     b: 8..16,
     c: 16..32,
}
```

mahkoh:
Then those bitfields are not C compatible and require the user to write platform dependent code for every platform (x86_64 linux, x86_64 windows, arm linux, arm windows, etc.) The RFC is supposed to solve this.

petrochenkov:
@oli-obk 
For native Rust bitfields Ada's system looks like an excellent alternative, but for `#[repr(C)]` bitfields you need compatibility with C ABIs in the first place, not precise ranges.

text/0000-bitfields.md
```diff
@@ -0,0 +1,340 @@
...
+The type of a bitfield is its declared type. Reading from a bitfield returns a
+value of the declared type. Only values of the declared type can be written to
+a bitfield. The bits of the binary representation of a bitfield are interpreted
+as the bits of a signed or unsigned integer of the width of the bitfield. In
+particular, a signed bitfield of non-zero size contains a sign bit. When a
+zero-sized bitfield is read, `0` is returned.
+
+For the purpose of overflow checking, bitfields are treated as signed or
+unsigned integers of their declared width. In this context, the only value that
+should be written to a bitfield of size `0` is `0`.
```
oli-obk: So, doing something else should panic? I think this should specify what happens, not what users of the bitfields should do.

oli-obk:
Right, ignore that. Just for C-interop this is fine.

pornel:
Is this really needed? I thought C bitfields are an esoteric feature nobody uses. I've never encountered them in any API.

retep998:
*cough Windows API cough*
There are plenty of C APIs that make liberal usage of bitfields, although Windows at least has very simple rules for bitfields which are trivial to replicate in Rust, unlike some of the weird things that other platforms do.

nrc:
@pornel they seem to be common in very low-level programming - kernels, drivers, etc. That is a space where Rust would be well-suited but is currently a bit weak. Also useful for FFI/interacting with C, which is another slightly weak area. I'd like to see some kind of bitfields to help address those weaknesses.

jmesmon:
@nrc indeed. Though bitfields for platform compat (need to match how the platform lays out bitfields) might not be the same as how we'd want bitfields for drivers/hardware interaction (need to always be laid out the same way, not something platform specific).

drbo:
I am in favor of this RFC.

A syntax for packed bitfields (useful for assemblers, networking, drivers, etc) would also be useful.  I like Ada's syntax for this, but it is rather verbose.

One note is that in most applications of packed bitfields, one does not want any padding, so needing to add padding should be either a compile-time error or a lint that is deny-by-default.

text/0000-bitfields.md
nrc: What happens in non-debug builds where we don't panic on overflow? Is the write just ignored?

text/0000-bitfields.md
```diff
@@ -0,0 +1,340 @@
...
+}
+```
+
+In this structure, `a` and `b` must not be modified concurrently. `a` and `d`
+can be accessed concurrently.
+
+Values stored in a bitfield have a binary representation using the `m` bits it
+occupies in its memory location.
+
+The `&` and `&mut` operators cannot be applied to bitfields.
```
nrc: And presumably implicit borrows too

nrc:
Nominating for discussion at the lang meeting.

My thoughts:

* I think there is real motivation for this feature. It seems like it would be fairly self-contained and general low impact. It seems like the sort of thing that can only be handled well in the language/compiler (c.f., in a library or tool).
* There's mostly positive feedback in this thread, though not evidence of overwhelming demand.
* It seems worth pointing out the different scenarios of bit-fields as a pure Rust feature vs bit-fields for C compatability (this RFC proposes the latter). For the latter, I like the proposed syntax and semantics. However, it is a bit unsettling that we might go for something else if it were a pure Rust thing.
* I am very worried about having such vague semantics for a feature. In particular the following paragraph is troubling:

> Many details of the behavior of such fields are left to the implementation.
> However, the implementation is encouraged to follow the behavior of the dominant
> C ABI on the target platform.

* Indeed, it is perhaps a show-stopper for such a feature if the only way to ensure multi-platform C compatability is by having a vague and variable specification for the Rust feature. However, it seems that an extra-linguistic solution would have the same (or worse) problems, so if we want to address the problem we might have to swallow this.
* I don't understand the impact of the race safety issues. Do we need to add extra safeguards to the language to keep it safe?
* Will the ban on referencing cause problems? I can't think of anything, but I can't think of anywhere else we do this in Rust.

Ericson2314:
> It seems worth pointing out the different scenarios of bit-fields as a pure Rust feature vs bit-fields for C compatability (this RFC proposes the latter). For the latter, I like the proposed syntax and semantics. However, it is a bit unsettling that we might go for something else if it were a pure Rust thing.

I'm interested in a pure-rust solution for working with hardware (network packets, processor-specific data structures, etc). Quite unlike working with C, one wants representation details mandated ad nauseam for this.

eefriedman:
> Will the ban on referencing cause problems? I can't think of anything, but I can't think of anywhere else we do this in Rust.

We need something like this for https://github.com/rust-lang/rust/issues/27060 anyway.

> Many details of the behavior of such fields are left to the implementation.
> However, the implementation is encouraged to follow the behavior of the dominant
> C ABI on the target platform.

This doesn't need to be quite so general.  Maybe "the representation, padding and alignment of a bitfield field are target-dependent"?

> I don't understand the impact of the race safety issues. Do we need to add extra safeguards to the language to keep it safe?

The restriction that you can't borrow a bitfield is enough to for safety in safe Rust: that makes it impossible to concurrently mutate two bitfields in the same object.  You have to be a little bit careful how exactly you do code generation for bitfields to make sure that writing to a bitfield doesn't overwrite bytes associated with other non-bitfield fields, but that's straightforward.

nagisa:
I still personally feel that this RFC is *undershooting* with ambitions here. I feel like there is a strong potential for a superset of functionality (e.g. arbitrary structure layout) while at the same time solving the issue of bit fields in a more convenient way.

retep998:
>Will the ban on referencing cause problems? I can't think of anything, but I can't think of anywhere else we do this in Rust.

`repr(packed)` is _supposed_ to ban references to its fields in safe code (seriously why hasn't anyone fixed this yet), so having bitfields ban references to its fields is totally fine. Banning references to a struct which contains a bitfield, however, would be much too limiting in my opinion and basically kill the usefulness of this RFC for me.

>I still personally feel that this RFC is undershooting with ambitions here. I feel like there is a strong potential for a superset of functionality (e.g. arbitrary structure layout) while at the same time solving the issue of bit fields in a more convenient way.

I don't really mind if this RFC allows for arbitrary structure layout as long as I still get convenient syntax for defining FFI compatible bitfields that is not significantly worse than C.

nrc:
🔔 This RFC is now entering its **final comment period** 🔔

ubsan:
I think bitfields are fine, but arbitrary struct layout goes too far. The interaction with C code kinda makes the argument for me. Arbitrary struct layout doesn't fit that need.

Also, maybe in the implementation of this, someone can finally make packed-field-ref unsafe 😉

DemiMarie:
I think that there is a VERY simple solution to the problem of C- vs explicit- layout: the `#[repr]` attribute on the type.

- If the type is `#[repr(C)]`, follow C rules.
- If the type is `#[repr(packed)]`, use no padding.
- Otherwise, the layout is unspecified (as for any Rust type with an unspecified layout).  This is not particularly useful (the only possible use I can think of is to get overflow checks in debug builds), and should probably be at least linted against, if not forbidden.

DemiMarie:
@ubsan Arbitrary struct layout would be _extremely_ useful for programs such as assemblers.  Just write the (packed) structure (such as an encoded CPU instruction) into memory and transmute it into a byte array (a transmute which, for packed structs, is guaranteed safe).

ubsan:
@DemiMarie I feel like I haven't seen any good motivating examples.

> Arbitrary struct layout would be extremely useful for programs such as assemblers.

I don't understand where? Could you give an example, please?

> Just write the (packed) structure (such as an encoded CPU instruction) into memory and transmute it into a byte array (a transmute which, for packed structs, is guaranteed safe).

Arbitrary struct layout doesn't seem to get you anything over bitfields in this case?

DemiMarie:
I mixed up "arbitrary struct layout" with "bit fields with specified layout"

Ericson2314:
Once on IRC, I talked about having iN uN for all 0 < N < 64. These would have size rounded up to the nearest power of 2 most of the time (rvalues, normal lvalues), and corresponding alignment, but would be packed to `N` bits in a packed struct. Additionally there would be a rule that no field in a packed struct could span the boundary of it's rounded up size (e.g. no packed `Foo(u3, u7, u5)` because 7 rounds up to 2^3 = 8, and straddles a byte boundary). Finally, there should be "packed tuples" for the ease of 1-off  

Now, this *wouldn't* be useful for C FFI, I assume, because C's weird field-in-field way of doing bitfields causes all sorts of weirdness (especially endianness issue the C library authors probably didn't even anticipate... but I digress).

The consensus on IRC was that this was better suited for a compiler plugin, because operation on these bit-grained integers and packed structs containing them, but I have three rebuttals.

1. We already have precedence for software integers, namely 64 bit ones on < 64-bit architectures.

2. A getter+setter procedural macro only accounts for the packed struct part of this, and doesn't match the ergnomics of working with {i,u]N + packed struct rvalues. After working in Verilog, by no means a nice language, using bit shifts for all this seems extraordinarily primitive.

3. I think there is a serious opportunity for compiler optimizations vastly beating humans here.

  - As far as I can tell, most sub-byte workloads use arithmetic relatively sparingly, which is good because arithmetic seems hardest to optimize (can't fake sub-byte SIMD with normal registers+ALU unless you can prove lack of overflows).

 - Many optimizations are hampered by the extreme cost / inability of working on on multiple functions at one / change interfaces. (e.g. want less indirection in that function's argument? Oh wait, its used as a function pointer.) This sort of packing optimization however is neatly confined to working within functions, there is no inter-procedural extension to be desired.

 - Packing values by hand seems hard to do optimally by hand (has anyone ever tried?) And makes bit-shifting soup even harder to read.

-----

So I'd like to see an RFC with such a plan, as a counter point to both this C-centric approach and the procedural macro approach. I want this to be the standard for performance and legibility that other approaches are judged by.

ubsan:
@Ericson2314 I... do not want to deal with that in order to interop with C. I just want to interop with C. Let me interop with C, please.

Ericson2314:
@ubsan I...do not want to deal with endianness ambiguities to interopt with hardware. I just want to interopt with hardware [Ok, and have good ergonomics]. Let me interopt with hardware, please.

petrochenkov:
>I... do not want to deal with that in order to interop with C. I just want to interop with C. 
>I...do not want to deal with endianness ambiguities to interopt with hardware. I just want to interopt with hardware

That's why these two are totally different features!
As a practical example, qemu is wholly about hardware and precise layouts, from ergonomics point of view it would benefit immensely from using bitfields. qemu is written in C and C has bitfields, however qemu very rarely uses them due to their unpredictability and non-portability and does almost everything with masking and shifting. I'll try to give some concrete example of how an MMIO-based hardware register could look in Rust with precise-layout bitfield support and without it, I only need to get back from vacation.

Ericson2314:
https://github.com/gz/rust-x86/commit/59db19be0ce5308595c2b52e455803bbe9da8ec9 is a good example, I have wrote a bunch of enums allowing for ergnomic representation of some aspects x86 descriptors, and a bitflags for the serialized forms. With what I propose (and also packed enums with 1-bit tags omg yes), I could combine the exact and ergnomic forms.

parched:
> Once on IRC, I talked about having iN uN for all 0 < N < 64. These would have size rounded up to the nearest power of 2 most of the time (rvalues, normal lvalues), and corresponding alignment, but would be packed to N bits in a packed struct. Additionally there would be a rule that no field in a packed struct could span the boundary of it's rounded up size (e.g. no packed Foo(u3, u7, u5) because 7 rounds up to 2^3 = 8, and straddles a byte boundary). Finally, there should be "packed tuples" for the ease of 1-off

I think this idea sounds good, but why the restriction on spanning the boundary? There will be cases where that will be needed.

Ericson2314:
@parched It was a fairly last-minute thought I'll admit. In the cases I've worked with (IP, TCP, x86 stuff) it is upheld. Also, I wanted to ensure that the bit-offset did not effect the number of writes needed to update the field in a platform-agnostic manner. The idea being that if your hardware supports writes of size `2^n`, but only aligned to `2^n`, then anything between 2^(n-1) and 2^n can still be written in one write. I don't know how practical this is, but it seemed conservative.

Maybe it's better to say just enforce that smaller than a byte must not overlap a byte-boundary.

parched:
@Ericson2314 Yes I imagine it is, personally I'm interested in this for ARM system registers, so I had a quick browse and they seem to respect it too. I did find some of the instruction encodings don't, but that's not something that matters for me. On the odd case something does span the boundary isn't better that it does extra writes than having to manually mask and shift again?

ubsan:
@Ericson2314 I'm not sure why `#[repr(packed)]` bitfields are worse than your solution.

Ericson2314:
@ubsan there is no capability for abstraction. I should have added that I'd expect one to be able to nest packed structs without wasting bits.

nikomatsakis:
Bitfields are one of those features that seems to come up again and again. It's clear that they are somewhat "niche" in the sense that most people don't want them, but it's also clear that there is a sizable set of people that would really like them. Adding something in this direction makes sense to me.

That said, I think this RFC is not quite there yet. First, some of the language is rather imprecise:

- I didn't quite understand the language about the interaction with overflow. When are overflow checks performed precisely? What is the "fallback behavior" when overflow checks are disabled?
- We don't have any attributes that have "constant expressions" anywhere else -- though there is a proposal to allow arbitrary literals. It seems like limiting to literals would be fine.
- Avoiding references seems like an ok thing; should be relatively easy to enforce if we check it in the MIR.

It is certainly true that bitfields have many constituencies (C interop, network, etc) that seem to demand different layouts; perhaps indeed `repr` is a good way to bridge those use cases, but I don't think this is spelled out in the RFC (or did I miss it). Or maybe the "precise layout" croud would prefer a distinct feature that gives even more control? I guess this is the key question.

I do also wonder if we can do something to make bitfields feel a bit more... slick. I worry about the `#[bitwidth]` attribute feeling clunky. But perhaps that is silly, since -- as I said in the beginning -- bitfields are not something that most people will interact with.

nikomatsakis:
We discussed this in the @rust-lang/lang meeting. I think the general feeling was that there is much to like about this RFC. In particular, the motivation of easing C interop is a good one, and if that's the target, adding an attribute is likely the most obvious way forward.

However, we didn't feel like we could accept it in its present form, because there are some issues that would need to be cleaned up:

- [ ] clarifying overflow behavior
- [ ] restricting to literals, I suspect, or at least deciding if we really need constant expressions
- [ ] the interaction with `repr` 

@mahkoh, this thread has been quiet lately. Are you interested in revising the RFC to try and address these points and push it over the finish line? Otherwise, we'll probably postpone the RFC, but with the hope that someone will submit a revised version.

retep998:
Also just to provide some numbers in favor of bitfield support.
```bash
# In the Windows 10 SDK 10.0.10586.0
$ grep -Er ': *[0-9]+ *;' . | wc -l
11902
```

petrochenkov:
>Since they have to be literals in C anyway.

From C99/C11 standards:

> The expression that specifies the width of a bit-field shall be an **integer constant
expression** with a nonnegative value that does not exceed the width of an object of the
type that would be specified were the colon and expression omitted.

retep998:
@petrochenkov Aw damn, you're right. After doing some very intense grepping to find cases where that was the case, I found some. For example:

```C
typedef struct _PERFINFO_CCSWAP_LITE {
    ULONG   DataType            : PERFINFO_CCSWAP_BIT_TYPE;
    ULONG   OldThreadIdIndex    : PERFINFO_CCSWAP_BIT_TID;
    ULONG   OldThreadPriInc     : PERFINFO_CCSWAP_BIT_PRI_INC;
    ULONG   OldThreadStateWr    : PERFINFO_CCSWAP_BIT_STATE_WR;
    ULONG   TimeDelta           : PERFINFO_CCSWAP_BIT_SMALL_TS;
} PERFINFO_CCSWAP_LITE, *PPERFINFO_CCSWAP_LITE;
```

Therefore I am changing my view. Support for constant expressions is needed.

nikomatsakis:
It seems like using a constant expression is not a huge problem, but it would require either an extension to attribute syntax, or else something more special purpose.

DemiMarie:
My thoughts on repr: bitfields should behave like any other type. repr(C)
should get C-like behavior,  repr(rust) should be unspecified, and
repr(packed) should be packed.  If a repr(packed) struct cannot be packed
it is a compile-time error.

On Jul 22, 2016 4:33 PM, "Niko Matsakis" <notifications@github.com> wrote:

It seems like using a constant expression is not a huge problem, but it
would require either an extension to attribute syntax, or else something
more special purpose.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/rust-lang/rfcs/pull/1449#issuecomment-234648906>, or mute
the thread

DemiMarie:
Overflow: same as with any other Rust type: panics in debug builds only.

nikomatsakis:
@DemiMarie 

> My thoughts on repr: bitfields should behave like any other type. repr(C) should get C-like behavior, repr(rust) should be unspecified, and repr(packed) should be packed. If a repr(packed) struct cannot be packed it is a compile-time error. 

Seems reasonable.

> Overflow: same as with any other Rust type: panics in debug builds only.

I agree, though we must also define the fallback behavior. Seems like masking out the extra bits is the obvious thing.

(Just to be clear, though: I did not mean to imply that these behaviors were necessarily controversial, or even all that complex, just that they should be spelled out better in the RFC text.)

retep998:
@nikomatsakis The behavior of overflow in C bitfields on Windows is wrapping overflow, aka masking it and dropping the extra bits.

joshtriplett:
One more aspect that needs clear definition: does accessing a bitfield do a read-modify-write that affects adjacent fields?  Even if accessing any field in a struct containing a bitfield requires an `unsafe` block, it'd be helpful to have more precise semantics there.  GCC and LLVM sometimes generate very strange code with bitfields, which can lead to both semantic correctness issues and strange performance issues.

eefriedman:
> One more aspect that needs clear definition: does accessing a bitfield do a read-modify-write that affects adjacent fields?

I think this is addressed by the RFC ("Concurrent access to different bitfields...").  It does a read-modify-write on adjacent bitfields, but not on adjacent non-bitfield fields.  Maybe it could be clarified a bit.

joshtriplett:
@eefriedman "Not on adjacent non-bitfield fields" may not be trivial to guarantee on all platforms; do LLVM primitives make that straightforward?

eefriedman:
See https://github.com/llvm-mirror/clang/commit/72d2dab6058467036df73a5f668036a519043e5b if you want some idea of how clang generates LLVM IR for bitfields; it's not too difficult.

nikomatsakis:
Given that there hasn't been a response to [my prior comment](https://github.com/rust-lang/rfcs/pull/1449#issuecomment-234608279), we decided in the @rust-lang/lang to **postpone** this RFC under (existing) issue #314. As noted earlier, a follow-up that takes more-or-less the same approach but addresses the various ambiguities would be welcome. (@mahkoh, also feel free to re-open this PR with edits.)

blanham:
@Ericson2314 Do you have a copy of the logs from your discussion on IRC about uN and iN integer types? I agree that having such a feature, along with pattern matching, would be amazing for embedded work and I'm considering working on an RFC or compiler extension to make it happen

Luke-Nukem:
@nikomatsakis is there a guide anywhere for people willing to adopt an RFC to help finish it? I'd be willing to take on this one as a learning process.

aturon:
@Luke-Nukem There's not a guide; generally, folks from the @rust-lang/lang team (including shepherds like @joshtriplett, @solson, @scottmcm, @cramertj, and @whitequark) are happy to mentor RFCs. Any of you fine folks want to volunteer?

FWIW, this issue falls into the "areas of exploration" for the [roadmap](https://github.com/aturon/rust-roadmap/), and thus we're happy to provide support for seeing it through.

retep998:
If we use iN and uN, then how would the alignment be calculated? A bitfield derived from a `u32` obviously has the alignment of `u32` but if you have a pile of `u1` and `u2` how do you know what the alignment should be? This is an *essential* detail for FFI.

whitequark:
As a shepherd for Rust on embedded devices, I am opposed to this RFC. When accessing memory-mapped registers of a peripheral, you have to keep several things in mind:

  * Writes to an MMIO register are side-effectful (of course), but often, so are reads.
  * Both reads and writes to an MMIO register must have the correct size and alignment. Accessing just the 1st or 2nd byte of a 4-byte register could return nonsense, raise an exception, or even hang.
  * All read-modify-write cycles involving MMIO registers must be explicit. A typical access pattern would be: read a register, look at bit #X, if it's set, then also set bits #Y and #Z, write the result back. Performing more than one read or write here could easily have different, undesirable semantics.

Because of this, one of two things must happen:
  * Either Rust gains a lot of `repr` attributes specifying that a struct must have volatile accesses, specific size, alignment, etc.
  * Or Rust bitfields are unusable for directly representing MMIO registers, where under "directly" I mean such that a pointer to an MMIO register that has the type of a struct with bitfields cannot be used with field access syntax.

Even in the former case, a read register, read bitfield, write bitfield, write register sequence requires explicitly reading the bitfield-containing struct out of a register, manipulating it, and writing it back.

C bitfields are similar to the latter case, with the added complication that the standard does not specify any equivalent to `#[repr(packed)]`, and so you're stuck with implementation-specific syntax for even more basic things.

I do not believe that MMIO registers warrant adding language support for bitfields. Automatically generating accessors using a builder pattern, like [svd2rust](https://github.com/japaric/svd2rust) does, results in perhaps slightly unwieldy implementation-wise, but perfectly usable code on today's Rust, allow for very compact and readable code without loss of flexibility, does not create a burden on the core language, and does not prompt language changes when Rust is ported to an architecture weirder than anything seen before.

As a systems Rust programmer excluding embedded contexts, I think that bitfields are marginally useful but not encountered nearly often enough to be included in the core language. The bitfield crate is just fine.

Ericson2314:
I think integer types for arbitrary bit widths would be amazing, but agree with @whitequark regarding language changes less ambitious than that---they are just not worth it.

whitequark:
I'm not really sure what is amazing in it. They are needed very rarely, and you can get *almost* all of the functionality using const generics with redefined operators anyway.

solson:
Rather than having bitfields as types or part of struct syntax, I think it would be more interesting to consider [bit-matching patterns and bit string expressions](http://erlang.org/doc/programming_examples/bit_syntax.html) in the style of Erlang. 

This might be doable as a library, but if they really were first-class patterns *and* we had pattern aliases, you could have a regular struct plus a collection of useful pattern aliases defined for matching on the bitfields. I believe this approach avoids many of the problems @whitequark brought up since it's separate from the field/memory access.

Additionally, this could be more generally useful for packing/unpacking binary protocols of any kind. For example, check out the `Dgram` example in my link above.

Ericson2314:
@solson I've never used erlang, but bit-grained integers / const-generic fixed-sized bitvectors + packed tuples seemed to me like the fully-developed version of that feature (in a typed language).

@whitequark My experience with this is indirect via http://www.clash-lang.org/---it's not rust and it's not on a CPU but it has arbitrary number of bit integers and bitvectors, and compound types are effectively packed. And it does indeed make just about any code marshaling across some binary interface *far* more agronomic. I never want to manually write a bit-shift outside of assembly ever again.

whitequark:
> My experience with this is indirect via http://www.clash-lang.org/ for FPGAs

Sure, arbitrary width integers make a *lot* of sense as a core construct in a language for FPGAs (I've worked with Verilog and Migen, personally), but not in the domain Rust targets.

> Rather than having bitfields as types or part of struct syntax, I think it would be more interesting to consider bit-matching patterns and bit string expressions in the style of Erlang.

> I believe this approach avoids many of the problems @whitequark brought up since it's separate from the field/memory access.

I agree that Erlang's bit vectors and patterns are amazing and hard to beat for parsing and serializing data but the case for MMIO is harder to justify. If I understand the involved issues correctly, I'd rather have [Habit]'s monadic approach (which is similar to what svd2rust currently generates anyway) than Erlang's matching and construction approach.

[habit]: http://lambda-the-ultimate.org/node/4205

Ericson2314:
@whitequark IIUC, to prevent the illusion of sub-register access, wouldn't a wrapper struct for the MMIO register static suffice? The wrapper has read and write methods and a single private field, but the type of the value returned would be something like `struct MyRegContents(foo: u5, bar: u9, baz: u3)`, which makes constructing destructuring those (separately from I/O) much more elegant.

whitequark:
> which makes constructing destructuring those (separately from I/O) much more elegant.

not really? svd2rust-generated accessors for RMW cycles look like...

```rust
periph.reg.update(|r, w| w.foo(r.bar() + 1))
```

which is actually shorter than...

```rust
let reg = periph.reg.read();
periph.reg.write({ foo: reg.bar + 1, ..reg });
```

and about as long as a less primitive...

```rust
periph.reg.update(|reg| { foo: reg.bar, ..reg })
```

Ericson2314:
Err the wrapper method is orthogonal to the payload type.

I'm a bit confused when it needs to take a `&mut` pointer back, so what about
```rust
periph.reg.update(|r, w| w.foo = r.bar + 1))
```

Even if the API is the same, it's still easier to generate code that can use this feature (resulting code is simpler).

whitequark:
@Ericson2314 Oh you're absolutely right, disregard what I said.

joshtriplett:
@whitequark So, I'd like to disagree with your conclusion here. I *do* agree that C-style bitfields would not be sufficient for MMIO usage. However, my conclusion there is instead "we should do better than C", rather than "we should give up".

C bitfields have multiple problems, most notably a lack of any way to specify the granularity of the operation. I think we can fix that, and I don't think it'll require *extensive* attributes to do so.

Beyond that, I would also like to see this to simplify interoperability with C via FFI (and in particular, `repr(C)` bitfields should default to matching the behavior of C on the target platform).

joshtriplett:
@whitequark I understood your argument, but I disagree; I think this is important to get right and to make simple.

Ericson2314:
@joshtriplett For me, based on what I've been proposing, "doing the right thing for hardware" and "doing the right thing for C interoperability" are completely unrelated extensions to the language. Trying to use the same set of features for both would I think inevitable result in compromise or way more complexity.

joshtriplett:
@Ericson2314 That's entirely fair. And I'd love to see an extension that's good enough for C FFI interoperability, with some thoughts on forward-compatibility with hardware MMIO.

Ericson2314:
Err because approaches to each are so different, I don't think there's much forward compatibility to worry about. I'd start with hardware/marshalling, because `uN` and and friends are a really cool opportunity to beat just about every other language in a huge and easy to convey (with code examples, harder to explain before) way.

C bitfields I agree are basically a misfeature in that they are too complex and intuitive, and anything feature that wants to interropt with them basically has to mimicked them exactly---there's no way to trim the fat without loosing the flavor. Maybe it would be very useful, but replicating the ugliness sounds like a boring chore, and I wouldn't want developers of new non-C-interfacing code to be tempted to use them because of the absence of `uN`.

Luke-Nukem:
Err, wow. An explosion of discussion here. You'll have to excuse me as some of it goes over my head and beyond my experience.

I do need to state what I thought I understood the purpose of this to be, and that is solely to enable use of C bitfields over FFI. This is the last missing piece of the puzzle for quite a few bindings, especially in the GNOME world with such things as [`_GIOChannel` which has bitfields in the middle of a struct](https://github.com/GNOME/glib/blob/6b948d9613f2dcdec72828f3429ba01403063ff2/glib/giochannel.h#L97-L126);

```C
 struct _GIOChannel
{
  /*< private >*/
  gint ref_count;
  GIOFuncs *funcs;

  gchar *encoding;
  GIConv read_cd;
  GIConv write_cd;
  gchar *line_term;		/* String which indicates the end of a line of text */
  guint line_term_len;		/* So we can have null in the line term */

  gsize buf_size;
  GString *read_buf;		/* Raw data from the channel */
  GString *encoded_read_buf;    /* Channel data converted to UTF-8 */
  GString *write_buf;		/* Data ready to be written to the file */
  gchar partial_write_buf[6];	/* UTF-8 partial characters, null terminated */

  /* Group the flags together, immediately after partial_write_buf, to save memory */

  guint use_buffer     : 1;	/* The encoding uses the buffers */
  guint do_encode      : 1;	/* The encoding uses the GIConv coverters */
  guint close_on_unref : 1;	/* Close the channel on final unref */
  guint is_readable    : 1;	/* Cached GIOFlag */
  guint is_writeable   : 1;	/* ditto */
  guint is_seekable    : 1;	/* ditto */

  gpointer reserved1;	
  gpointer reserved2;	
};
```

If there *is* a way to do bitfields better (and it certainly looks like there is), can this be relegated to another RFC in the future?

To restate from this RFC;
> This rfc introduces a single attribute that can be applied to fields inside of #[repr(C)] structs to turn them into bitfields.
> The attribute #[bitfield(N)] is added. Here, N can be any constant expression that evaluates to a value of type usize.
> 
> The attribute can be applied to fields inside a #[repr(C)] struct which have any of following types:
> 
>     u8, u16, u32, u64, usize,
>     i8, i16, i32, i64, isize.

Thus it is my hope to have the `_GIOChannel` example reduced to;

```rust
#[repr(C)]
pub struct GIOChannel {
    pub ref_count: c_int,
    pub funcs: *mut GIOFuncs,
    pub encoding: *mut c_char,
    pub read_cd: GIConv,
    pub write_cd: GIConv,
    pub line_term: *mut c_char,
    pub line_term_len: c_uint,
    pub buf_size: size_t,
    pub read_buf: *mut GString,
    pub encoded_read_buf: *mut GString,
    pub write_buf: *mut GString,
    pub partial_write_buf: [c_char; 6],
    #[bitfield(1)] pub use_buffer: u8,
    #[bitfield(1)] pub do_encode: u8,
    #[bitfield(1)] pub close_on_unref: u8,
    #[bitfield(1)] pub is_readable: u8,
    #[bitfield(1)] pub is_writeable: u8,
    #[bitfield(1)] pub is_seekable: u8,
    pub reserved1: gpointer,
    pub reserved2: gpointer,
}
```

I think @mahkoh was on the right track here. I don't see any other way to do this without adding a large amount of complexity to written code.

And please, the purpose of the RFC is for FFI with C, nothing more or less than that. If we can get this done, then it opens up a world of possibilities to migrate a lot of things to Rust that rely on the use of such structs as shown above.

Luke-Nukem:
@whitequark if this RFC is specifically for C style bitfields (and largely only for FFI), will that affect your use cases for embedded? This is an area I am interested in personally.

whitequark:
> @whitequark if this RFC is specifically for C style bitfields (and largely only for FFI), will that affect your use cases for embedded? This is an area I am interested in personally.

Yes. C style bitfields are very unergonomic for embedded dev. For example, a baseline requirement for such an interface, in my view, is an ability to define a field as spanning bits n...n+m, regardless of where other fields are laid out, so that I can add, remove, or even `#[cfg]` out fields to account for device variation, without worrying about inserting the precise amount of reserved bits.

joshtriplett:
I'd very much like to see this RFC clarified and the alternatives and tradeoffs filled out. I see *no* problem with having an RFC specifically for C-FFI-compatible bitfields, with the *explicit* acknowledgement that more work would be needed to support hardware bitfields.

solson:
@whitequark Are you familiar with Ada [record representation clauses](https://en.wikibooks.org/wiki/Ada_Programming/Keywords/use#For_records)? Specifically:

```ada
for  Device_Register use
    record
       Ready at 0 range  0 .. 0;
       Error at 0 range  1 .. 1;
       -- Reserved bits
       Data  at 0 range 16 .. 31;
    end record;
```

(Read `<field name> at <start byte> range <start bit> .. <end bit>;`.)

It sounds a lot like what you want, since the padding bits are completely implied here. It would be cool if Rust had a general system like this for those cases where you want total layout control.

Luke-Nukem:
@whitequark I'm not sure how you mean with that - is there an example I can look at? If there were to a come a future RFC geared towards the hardware issues, how do you think the syntax `#[bitfield(1)] pub use_buffer: u8,` would affect it? Keeping in mind that this particular RFC is for a singular task, is there a particular way you would like to see the bitfields written (in visible rust code)?

@joshtriplett I am collating much of the discussion here in to my notes and seeing what I need to do. May be a few days since I also have to study (and do my GSoC project work), but I am definitely leaning towards being specifically C-FFI only for this RFC. It would be nice to find a compromise but that doesn't look likely.

whitequark:
> If there were to a come a future RFC geared towards the hardware issues, how do you think the syntax #[bitfield(1)] pub use_buffer: u8, would affect it?

This syntax would be completely orthogonal and not used in interfacing with hardware.

> Are you familiar with Ada record representation clauses?

No, but what you describe is a very good start. There are some subtleties, e.g. it is perfectly reasonable for fields to overlap, even fully.

solson:
> it is perfectly reasonable for fields to overlap, even fully.

Yeah, for sure. I imagine having a completely general layout toolkit that could simulate `union`, `struct`, bitfields, and whatever weird hybrid is necessary for your specific usecase.

solson:
The split between C-FFI bitfields and layout control for hardware or other reasons seems pretty irreconcilable. The C-FFI notion of bitfields is about a layout the programmer can't know, since C bitfield representation varies across platforms, whereas @whitequark's notion is for when the programmer has very specific knowledge of the exact layout they need or want.

(I mean, technically the programmer could try to match C layout with an Ada-like system and their own collection of `cfg`s, but the purpose of the RFC is to put that burden of platform knowledge on `rustc` itself.)

There might still be some room for overlapping syntax, but it seems like it might be more confusing than helpful. So @Luke-Nukem's idea of splitting into separate RFCs makes sense to me.

joshtriplett:
This discussion actually seems very familiar. When RFC 1444 discussed unions, there were suggestion for complete structure layout control, allowing the arbitrary placement of fields (including overlapping, which would allow for unions).

In my opinion, I think the answer there applies here as well: I think we want both of these things, but separately. I would propose the following: for *this* RFC, let's focus on C FFI bitfields, for which we just need an updated RFC that discusses alternatives/documentation/etc. And someone interested in modeling hardware registers may wish to propose an additional RFC with the level of control they'd propose as sufficient for that task.

Luke-Nukem:
Thanks for the input so far everyone (giving me a lot to digest).
Okay so we're all in agreement with this RFC specialising to C-FFI bifields and not trying to general purpose it? I'll get started on clarification of the raised issues.

Anyone willing to mentor me through this? My time is a little restricted due to study and GSoC, but we could easily do this through email or even matrix clients.

aturon:
> This discussion actually seems very familiar. When RFC 1444 discussed unions, there were suggestion for complete structure layout control, allowing the arbitrary placement of fields (including overlapping, which would allow for unions).

> In my opinion, I think the answer there applies here as well: I think we want both of these things, but separately. I would propose the following: for *this* RFC, let's focus on C FFI bitfields, for which we just need an updated RFC that discusses alternatives/documentation/etc. And someone interested in modeling hardware registers may wish to propose an additional RFC with the level of control they'd propose as sufficient for that task.

I haven't followed the thread in detail, but I wanted to weigh in here. We've had a lot of success having a highly expressive underlying mechanism, and then providing more "sugary" high-level forms that can be understood as a kind of sugar using that. And I agree with @joshtriplett's characterization that we resolved similar issues with unions in this way (though the "underlying mechanism" is still waiting to be proposed.)

There are a few dangers to starting with the sugar, though:

- You can't be *sure* in advance that the expressive underlying mechanism will come into being, or be fully compatible with the sugar.

- You may be wrong about the "defaults" to give a sugary treatment to. This may only be apparent after experience with the more expressive mechanism.

These kinds of concerns ended up shifting our strategy with `impl Trait`, to shift gears and tackle the [fully general mechanism](https://github.com/rust-lang/rfcs/pull/2071).

In any case, I raise these concerns not to block the RFC, but just as something we need to keep in mind prior to *stabilization*. If there's clarity that we want to do *something* in this area, and there's a well-understood, sugary subset, *experimenting* with that subset seems wholly beneficial. But we need to get on the same page with the endpoint before stabilization.

solson:
> We've had a lot of success having a highly expressive underlying mechanism, and then providing more "sugary" high-level forms that can be understood as a kind of sugar using that.

@aturon I agree with this principle, but in this case I don't think it applies in the same way. You could take the same view with existing `#[repr(C)]` structs: they're just sugar for the upcoming fully general layout feature, specifying alignment/placement of each fields. For both `#[repr(C)]` structs and C-FFI bitfields, you can *for a specific C target*, say how they would be laid out, but all you really need to know is that rustc guarantees they will be compatible with C for any supported target.

I tried to articulate this difference in an earlier comment:

> The split between C-FFI bitfields and layout control for hardware or other reasons seems pretty irreconcilable. The C-FFI notion of bitfields is about a layout the programmer can't know, since C bitfield representation varies across platforms, whereas whitequark's notion is for when the programmer has very specific knowledge of the exact layout they need or want.

In other words, both features provide something the other doesn't - in one, target knowledge, in the other, control - and neither is a subset of the other.

> In any case, I raise these concerns not to block the RFC, but just as something we need to keep in mind prior to stabilization. [...] But we need to get on the same page with the endpoint before stabilization.

I agree we shouldn't block, and the primary thing to worry about before stabilization is that the two syntaxes jive with each other (i.e. they should seem like they belong in the same language 😃). Also, I suppose, that they build on top of the same rustc implementation work as much as possible, to avoid duplicated effort.

Ericson2314:
I'd imagine that we'd still want the C version's fields to be properly-sized `uN`, even if the declaration looks for C-like---they're just easier to work with. For this reason I rather start with the hardware feature than the C FFI one.

retep998:
I think the question really boils down to whether a given bitfield just has the given type with special behavior on read/write to only read/write that many bits (with some sort of overflow checking in debug), or whether they have special `uN`/`iN` types that infect the type system and add a lot of *fun* casting.

Ericson2314:
> a lot of fun casting

Hmm? If, given the other motivation I mention above we get those types anyways, I don't see much casting involved using them for this?

mahkoh:
New RFC: https://github.com/rust-lang/rfcs/pull/3064

MauriceKayser:
> I do need to state what I thought I understood the purpose of this to be, and that is solely to enable use of C bitfields over FFI. This is the last missing piece of the puzzle for quite a few bindings, especially in the GNOME world with such things as [`_GIOChannel` which has bitfields in the middle of a struct](https://github.com/GNOME/glib/blob/6b948d9613f2dcdec72828f3429ba01403063ff2/glib/giochannel.h#L97-L126);

@flukejones I was in need of something similar for my Windows related crate, as Windows extensively uses bit fields in API parameters and structures. I added an [example of your use case](https://github.com/MauriceKayser/rs-bitfield/blob/master/examples/gnu_io_channel.rs) to my [bitfield crate](https://github.com/MauriceKayser/rs-bitfield/), which supports:
- `bool`s and C-like enums as bit flags + enumerability over flags, if C-like enums are used
- Primitive types and C-like enums as multi-bit fields
- Explicit and implicit positioning and sizing of fields and flags
- Optional `core::fmt::Debug` and `core::fmt::Display` implementations
- Compile-time overlap and boundary checking.

PR #1461, opened 2016-01-13
Move some net2 functionality into libstd

sfackler:
[Rendered](https://github.com/sfackler/rfcs/blob/net2/text/0000-net2-mutators.md)

### DIFF
--- a/text/0000-net2-mutators.md
+++ b/text/0000-net2-mutators.md
@@ -1,0 +1,129 @@
+- Feature Name: net2_mutators
+- Start Date: 2016-01-12
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+[RFC 1158](https://github.com/rust-lang/rfcs/pull/1158) proposed the addition
+of more functionality for the `TcpStream`, `TcpListener` and `UdpSocket` types,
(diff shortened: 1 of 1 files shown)


alexcrichton:

Please don't everyone comment all at once!

More seriously, though, the libs team is interested specifically in diving into the landing strategy for this in libstd, either:

* This new functionality lands insta-stable
* The new functionality lands as an unstable extension trait

---

My own personal opinion is that we should land these as insta-stable. The APIs have been vetted in net2-rs for quite some time now, and I don't think we'd really gain all that much more use through another round of FCP for the APIs.

Curious what others think though!

briansmith:
I don't think these should be added via traits. Over time, the standard library would become a ridiculous mess of such traits as new functionality is added, if it were to follow that pattern.

sfackler:
The traits would only exist until the methods were stabilized.

alexcrichton:
The libs team discussed this during triage yesterday and the conclusion was to merge this RFC adding these methods as insta-stable. Thanks for the RFC @sfackler!

sfackler:
We're going to hold off until just after the next release to allow for the longest possible time before these hit a stable release and are locked in forever.

tshepang:
How about simply not making them insta-stable?

PR #1471, opened 2016-01-22
Stable compiler features

hauleth:


main--:
With [all](https://github.com/rust-lang/rust/issues/16337) [the](https://github.com/rust-lang/rust/issues/16383) [open](https://github.com/rust-lang/rust/issues/13366) [bugs](https://github.com/rust-lang/rust/issues/13368) [and](https://github.com/rust-lang/rust/issues/28587) [instabilities](https://github.com/rust-lang/rust/issues/27333) [surrounding](https://github.com/rust-lang/rust/issues/13370) [`asm!()`](https://github.com/rust-lang/rust/issues/13520) (there's [a lot](https://github.com/rust-lang/rust/labels/A-inline-assembly)), I really don't think it's ready for stabilization - even though I'd love to have stable inline asm in Rust.

We should also discuss whether today's `asm!()` really is the best solution or if something along the lines of RFC #129 or even [D](http://dlang.org/spec/iasm.html) would be better. One important point to consider here is that `asm()` does *not* support the same set of constraints as gcc. Therefore, we can either:

* Stick to the LLVM behavior and write docs for that (because I've been unable to find any). Nice because it avoids complexity in rustc. Bad because it will confuse programmers coming from C/C++ and because some constraints might be hard to emulate in Rust code.
* Emulate gcc and just link to [their docs](https://gcc.gnu.org/onlinedocs/gcc/Using-Assembly-Language-with-C.html#Using-Assembly-Language-with-C): Nice because many programmers already know this and there's plenty of examples one can just copy-paste with little modifications. Bad because it's a nontrivial extension to the compiler.
* Do something else (like D does): A lot of work that may or may not pay off. If done right, this could be vastly superior to gcc-style in terms of ergonomics while possibly integrating more nicely with language and compiler than just an opaque blob (lots of handwaving here as I'm not familiar enough with compiler internals to assess this).

Finally, another thing to consider is #1201 which in its current design (I think) depends quite heavily on inline asm - or inline asm *done right*, for that matter.

Ticki:
I agree with @main-- `asm!` is in no way ready for stability.

aturon:
Thanks for the RFC, @hauleth. However, our basic [stability story](http://blog.rust-lang.org/2014/10/30/Stability.html) is not open for change at this time. I'd encourage you to instead focus discussion on the [tracking issues](https://github.com/rust-lang/rust/issues/29722) to work out a path toward eventual stabilization.

hauleth:
Maybe `asm!` stabilisation is too much, but overall idea of "compiler dependant stable feature flags" seems worth reconsideration (like LLVM intrinsics). It is like compiler wide `unsafe` flag for features.

PR #1479, opened 2016-01-26
Unix socket support in the standard library

sfackler:
[Rendered](https://github.com/sfackler/rfcs/blob/unix-socket/text/0000-unix-socket.md)

### DIFF
--- a/text/0000-unix-socket.md
+++ b/text/0000-unix-socket.md
@@ -38,7 +38,10 @@ Postgres server will listen on a Unix socket located at
 `/run/postgresql/.s.PGSQL.5432` in some configurations. However, the
 `socketpair` function can make a pair of *unnamed* connected Unix sockets not
 associated with a filesystem path. In addition, Linux provides a separate
-*abstract* namespace not associated with the filesystem.
+*abstract* namespace not associated with the filesystem, indicated by a leading
+null byte in the address. In the initial implementation, the abstract namespace
+will not be supported - the various socket constructors will check for and
+reject addresses with interior null bytes.
 
 A `std::os::unix::net` module will be created with the following contents:
 
@@ -51,11 +54,7 @@ pub struct UnixStream {
 impl UnixStream {
     /// Connects to the socket named by `path`.
     ///
-    /// Linux provides, as a nonportable extension, a separate "abstract"
-    /// address namespace as opposed to filesystem-based addressing. If `path`
-    /// begins with a null byte, it will be interpreted as an "abstract"
-    /// address. Otherwise, it will be interpreted as a "pathname" address,
-    /// corresponding to a path on the filesystem.
+    /// `path` may not contain any null bytes.
     pub fn connect<P: AsRef<Path>>(path: P) -> io::Result<UnixStream> {
         ...
     }
@@ -196,15 +195,6 @@ impl SocketAddr {
 }
 ```
 
-A Linux-specific extension trait is provided for the abstract namespace:
-```rust
-pub trait SocketAddrExt {
-    /// Returns the contents of this address (without the leading null byte) if
-    /// it is an abstract address.
-    fn as_abstract(&self) -> Option<&[u8]>
-}
-```
-
 The `UnixListener` type mirrors the `TcpListener` type:
 ```rust
 pub struct UnixListener {
@@ -214,11 +204,7 @@ pub struct UnixListener {
 impl UnixListener {
     /// Creates a new `UnixListener` bound to the specified socket.
     ///
-    /// Linux provides, as a nonportable extension, a separate "abstract"
-    /// address namespace as opposed to filesystem-based addressing. If `path`
-    /// begins with a null byte, it will be interpreted as an "abstract"
-    /// address. Otherwise, it will be interpreted as a "pathname" address,
-    /// corresponding to a path on the filesystem.
+    /// `path` may not contain any null bytes.
     pub fn bind<P: AsRef<Path>>(path: P) -> io::Result<UnixListener> {
         ...
     }
@@ -294,11 +280,7 @@ pub struct UnixDatagram {
 impl UnixDatagram {
     /// Creates a Unix datagram socket bound to the given path.
     ///
-    /// Linux provides, as a nonportable extension, a separate "abstract"
-    /// address namespace as opposed to filesystem-based addressing. If `path`
-    /// begins with a null byte, it will be interpreted as an "abstract"
-    /// address. Otherwise, it will be interpreted as a "pathname" address,
-    /// corresponding to a path on the filesystem.
+    /// `path` may not contain any null bytes.
     pub fn bind<P: AsRef<Path>>(path: P) -> io::Result<UnixDatagram> {
         ...
     }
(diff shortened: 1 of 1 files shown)


text/0000-unix-socket.md
```diff
@@ -0,0 +1,491 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Is `std::os::unix::net` the right name for this module? It's not strictly
+"networking" as all communication is local to one machine. `std::os::unix::unix`
+is more accurate but weirdly repetitive and the extension trait module
+`std::os::linux::unix` is even weirder. `std::os::unix::socket` is an option,
+but seems like too general of a name for specifically `AF_UNIX` sockets as
+opposed to *all* sockets.
```
alexcrichton: Perhaps another unresolved question for now could be whether Windows named pipes should be added in parallel? Or something along the lines of if we should hold off on stabilization until there's a suitable Windows alternative as well.

text/0000-unix-socket.md
```diff
@@ -0,0 +1,491 @@
...
+    pub fn take_error(&self) -> io::Result<Option<io::Error>> {
+        ...
+    }
+
+    /// Shut down the read, write, or both halves of this connection.
+    ///
+    /// This function will cause all pending and future I/O calls on the
+    /// specified portions to immediately return with an appropriate value
+    /// (see the documentation of `Shutdown`).
+    pub fn shutdown(&self, how: Shutdown) -> io::Result<()> {
```
alexcrichton: This function isn't currently on `UdpSocket`, which I seem to recall is because it doesn't do much for datagram-oriented services, but do you know if it works well with Unix datagrams?

text/0000-unix-socket.md
```diff
@@ -0,0 +1,491 @@
...
+
+impl IntoRawFd for UnixDatagram {
+    ...
+}
+```
+
+Differences from `UdpSocket`:
+* `bind` takes an `AsRef<Path>` rather than a `ToSocketAddrs`.
+* The `unbound` method creates an unbound socket, as a Unix socket does not need
+  to be bound to send messages.
```
alexcrichton: This method differs somewhat from the planned strategy of `UdpSocket` which is to perhaps introduce a `UdpBuilder`, but does this mean that an unbound `UnixDatagram` is useful in its own right for sending/receiving data?

text/0000-unix-socket.md
```diff
@@ -0,0 +1,491 @@
...
+
+These features can bake in the `rust-lang-nursery/unix-socket` as they're
+developed.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+While there is precedent for platform specific components in the standard
+library, this will be the by far the largest platform specific addition.
+
```
alexcrichton: One drawback that may want to be mentioned here is the lack of a smooth migration path from `unix-socket` to the standard library. For example users of Unix sockets will currently have to decide whether to support older versions of Rust (and use `unix-socket`) or newer versions (use `std::os::unix::net`). Additionally, we'll have to duplicate the code when `unix-socket` is compiled against a standard library that also has Unix sockets.

We have ideas for implementing such a migration path, but none of these ideas have an RFC or have been implemented.

text/0000-unix-socket.md
sfackler: I think I probably just added it for completeness, but I'd expect it to bounce any pending `recv` calls at least. I'll see if it actually does anything.

text/0000-unix-socket.md
sfackler: It can't receive any data since you can't address it, but it can send: https://github.com/rust-lang-nursery/unix-socket/blob/master/src/lib.rs#L1159-L1173. This was added explicitly in a PR, so I believe it's a thing that's used in practice: https://github.com/rust-lang-nursery/unix-socket/pull/14

text/0000-unix-socket.md
sfackler: I don't think that blocking this feature on named pipes makes a ton of sense. Named pipes [do not appear to be used heavily](https://crates.io/crates/named_pipe) in the Rust ecosystem, and I don't really want us to get into a habit of wanting to add platform specific feature X to the standard library and then hunting for a roughly comparable other-platform specific feature Y to implement along side it for "symmetry".

text/0000-unix-socket.md
sfackler: It looks like shutdown does do what you'd expect for a datagram socket: https://github.com/rust-lang-nursery/unix-socket/commit/63650be391e5a12a0dba15e0c7d336926af863c6#diff-b4aea3e418ccdb71239b96952d9cddb6R1252

text/0000-unix-socket.md
sfackler: I'm not particularly worried about this specific case - there's not all that much code and these kinds of objects aren't typically going to be passed around between crates like `Duration`s might be.

The migration path ideas we've had seem to me to be a bit like the ideas to store fully elaborated source on crates.io - probably never going to happen :).

text/0000-unix-socket.md
alexcrichton: Hm ok, but I do also want to not tip the scales too far in favor of Unix as there are lots of abstractions on Windows we could reasonably provide (e.g. named pipes, IOCP bindings, etc). While crates today may not use these much it may just be because the Windows community is a little smaller.

These are mostly just musings, though, I'd be ok if the RFC doesn't explicitly mention named pipes.

text/0000-unix-socket.md
alexcrichton: Hm ok, seems reasonable to at least come in as unstable-to-start-out-with

text/0000-unix-socket.md
```diff
@@ -0,0 +1,501 @@
...
+Windows, their use is sufficiently widespread to warrant inclusion in the
+platform-specific sections of the standard library.
+
+# Detailed design
+[design]: #detailed-design
+
+Unix sockets can be configured with the `SOCK_STREAM`, `SOCK_DGRAM`, and
+`SOCK_SEQPACKET` types. `SOCK_STREAM` creates a connection-oriented socket that
+behaves like a TCP socket, `SOCK_DGRAM` creates a packet-oriented socket that
+behaves like a UDP socket, and `SOCK_SEQPACKET` provides something of a hybrid
```
drbo: `SOCK_DGRAM` Unix sockets are reliable.

alexcrichton:

alexcrichton:
Also from discussion in triage yesterday, the libs team concluded that a good way forward for now with abstract sockets is to do something like:

* Validate no nul bytes are found in strings
* Perhaps eventually add a linux-specific extension trait to manage abstract sockets.

It at least errs on the side of being conservative for now!

sfackler:
Updated to disallow use of the abstract namespace.

gkoz:
Shouldn't there be a way to set permissions on the socket before starting to listen to it?

sfackler:
What libc API would that use? According to the man page, the socket will inherit the permissions of the directory it's defined in on Linux, but filesystem permissions are ignored on BSD: http://man7.org/linux/man-pages/man7/unix.7.html#NOTES

gkoz:
I had `chmod` in mind ;) Okay, while I doubt any modern BSDs ignore "file" permissions I totally missed that the portable way is to rely on directory-level permissions.

alexcrichton:
The libs team discussed this RFC during triage yesterday, an the decision was to **merge**. This was the original intention for how Unix sockets would reenter the standard library and the `unix-socket` crate is doing well enough that moving it in almost wholesale seems good at this time (it's quickly becoming a vetted API).

Tracking issue: https://github.com/rust-lang/rust/issues/32312

alexcrichton:
Ah, also, the initial implementation: https://github.com/rust-lang/rust/pull/32302

PR #1480, opened 2016-01-26
RFC proposal for thread affinity

jamperry:


text/0000-thread-affinity.md
```diff
@@ -0,0 +1,133 @@
...
+
+The user may want more fine grained control of which core it should be locked to.
+
+### Locking to a specific CPU core
+
+```
+// locks thread to cpu core 2
+let t = thread::spawn(move || {
+    let cpu_no = 2;
+    let cpu_lock = CpuLock::lock_on(cpu_no).unwrap(); // the current thread is now locked to an arbitrary free CPU core
```
durka: this comment seems wrong

text/0000-thread-affinity.md
```diff
@@ -0,0 +1,133 @@
...
+    // thread logic goes here
+    // the cpu_lock will unlock from the core once out of scope
+});
+```
+
+```
+// locks thread to cpu core 2
+let cpu_no = 2;
+let t = thread::lock_on(2,move || {
+    let cpu_lock = CpuLock::lock_on(cpu_no).unwrap(); // the current thread is now locked to an arbitrary free CPU core
```
durka: if `lock_on` is analogous to `lock`, this line and the one after it shouldn't be here

text/0000-thread-affinity.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. This RFC does not provide any empirical evidence of the computational benefits. I can spend time to provide benchmarks from a C/C++ benchmark.
+2. Thread affinity can have the adverse effect - degrading performance if it interferes with the OS's scheduler if used in the wrong way.
+3. My lack of experience of Rust could be prohibiting an optimal design.
+4. Could be regarded as adding complexity to the standard library.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Calling unsafe OS-specific system calls to create threads and changing the affinity mask because `std::thread::Thread` do not expose the underling OS-specific thread id (AFIAW).
```
retep998: `JoinHandle` is what actually contains the OS-specific thread handle. On Windows `JoinHandle` implements `AsRawHandle` and `IntoRawHandle` and on non-windows it implements `JoinHandleExt`. Also getting the handle to the current thread is pretty simple (`GetCurrentThread` or `pthread_self` depending on your platform).

https://github.com/rust-lang/rust/pull/30177

text/0000-thread-affinity.md
```diff
@@ -0,0 +1,133 @@
...
+The proposed feature provides a performance optimization for servers. It would also be useful for MIO and similar frameworks where the main event-loop is locked to one core and passes connections to workers that are also locked to the remaining cores so it takes advantage of cache-locality.
+
+# Detailed design
+[design]: #detailed-design
+
+I strongly encourage others to improve my RFC since I am not a Rust expert and only serves to start a dialog on a proposed API.
+
+I split it into two parts: *User API* and *Implementation*.
+
+### Locking to an arbitrary free CPU core
```
retep998: There's a lot of room for detail on how exactly it locks to an arbitrary free CPU core. Does it determine what CPU core the thread is currently executing on and then lock to that? Does it pick a random CPU core?

retep998:
Relevant discussion on Rust internals https://internals.rust-lang.org/t/pre-rfc-thread-affinity/3117/9

text/0000-thread-affinity.md
```diff
@@ -0,0 +1,133 @@
...
+Calling unsafe OS-specific system calls to create threads and changing the affinity mask because `std::thread::Thread` do not expose the underling OS-specific thread id (AFIAW).
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+No research has been conducted into the following hypotheses:
+
+  *  Is it possible for consistent & symmetric semantics across all supported operating systems?
+
+  * Is it better to be an external library rather than in std?
```
retep998: I personally think this should start as an external library.

peterhj:
Thread affinity is something I'm sorely needing in Rust today as well. Some immediate comments on the RFC:

1. Rather than have just a lock, probably want a set, which maps better to the underlying thread affinity implementations (e.g., `cpu_set_t`), and which better reflects some real cases of "don't care" among threads on the same physical processor/socket.

2. At least for the pthread implementation of threads, there is a builder that initializes its `pthread_attr_t` _before_ the thread is even spawned. For example, that is how the thread stack size is set (see libstd/sys/unix/thread.rs). Logically, it would make sense to allow setting the affinity through the thread builder as well, using `pthread_attr_setaffinity_np` under the hood. This could be in addition to a guard-based mechanism as you described.

There are also general problems which I think concern the affinity implementations for any language's standard library. Dumping some here off the top of my head 

- How are processors/sockets, cores, and threads numbered? Typically the "system" numbering convention can be arbitrary or insane.
- How to query for all threads belonging to a physical core or a physical processor? I'm aware of the C libraries libnuma and hwloc which address this, neither of which are particularly standard.
- What numbering convention to accept through your affinity interface? My feeling is that the convention should be to just interpret them as the "system" numbering scheme, and leave the finesse to another library.

It could be interesting if libstd received a standard way of querying CPU topology. But more likely it'll start off as just another library, which is just fine if it just works.

alexcrichton:
Thanks for the RFC @jamperry! The discussion here and on the internals forum leads me to think that this may best be served by starting as an external library? It sounds like there's a lot of competing possible strategies and there's a lot of concerns to take into account. Taking some time to flesh out the library externally, get some experience, and toy with an API may be the best step forward here?

alexcrichton:

aturon:
I agree with @retep998, @alexcrichton and others that this work should start in an external library and be considered for addition to `std` at a later time, after we've gained practical experience with it.

diwic:
If implemented, how about using the [thread Builder](https://doc.rust-lang.org/std/thread/struct.Builder.html) like this: `thread::Builder::new().affinity(0x3).spawn(move || { /* ... */ })` ...where `0x3` is a bitmask of allowed cores.

Florob:
Like others I think this should (at least initially) live in an external library.
Apart from that I think the term "lock(ing)" is unfortunate in this context, as it is already used with different meaning(s) (cf. mutexes, and x86's lock prefix). I would prefer pinning, binding, or affinity.

alexcrichton:
The libs team discussed this RFC at triage today and the decision was to close for now. There doesn't appear to be a fundamental reason why this needs to be in the standard library right now, and otherwise it seems like this can be developed externally on crates.io

Thanks again though for the RFC @jamperry!

PR #1482, opened 2016-01-29
Add extension ".md" to text/1361-cargo-cfg-dependencies

petrochenkov:
Otherwise GitHub doesn't render it properly

r? @alexcrichton


alexcrichton:
Gah oops!

PR #1487, opened 2016-02-03
Relax return type constraints for match and if expressions

ConnyOnny:
[full proposal](https://github.com/ConnyOnny/rfcs/blob/master/text/0000-relax-match-return-type.md)

### DIFF
--- a/text/0000-relax-match-return-type.md
+++ b/text/0000-relax-match-return-type.md
@@ -1,0 +1,88 @@
+- Feature Name: relax_return_type_constraints_for_match_and_if_expressions
+- Start Date: 2016-02-03
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+Branches of match and if expressions are required to return a value of exactly the same type.
+This RFC suggests relaxing this constraint to only require all branches to return a value of a type which is usable by a surrounding function call.
+
+# Motivation
+[motivation]: #motivation
+
+When we want to use the outcome of a match for a sufficiently generic function,
+we currently have to duplicate the function call in each of the branches.
+
+```rust
+let file: Box<Read> = match fname {
+	"-" => Box::new(stdin.lock()),
+	_ => Box::new(fs::File::open(fname).unwrap())
+};
+```
+
+This is shorter and more readable:
+
+```rust
+let file: Box<Read> = Box::new(match fname {
+	"-" => stdin.lock(),
+	_ => fs::File::open(fname).unwrap()
+});
+```
+
+# Detailed design
+[design]: #detailed-design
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Poor little type system :(

Personally I feel like the solved issue is a minor annoyance at worst (since you have to type a little bit more), and doesn't justify punching such deep holes into Rust's static type system. I think this is also a potential footgun, since this will probably confuse quite a few people (this looks like it would easily be used by accident).

oli-obk:
You can implement it explicitly with a macro: http://is.gd/szWfuA

That way you explicitly opt-in, instead of having some magic happen in the background.

ConnyOnny:
@jonas-schievink Interesting. I see it more as strengthening the type system than "punching deep holes" in it. In a way it is just trait information propagation, or did I miss something important?
As for the footgun: I think this is only one if there is a generic function which does utterly different things for different input types. But in that case we are already in trouble.
As for the accidental use: Yes, this might be used without much thought, but if the surrounding function is reasonable, then this is probably exactly what the programmer wanted: The correct version of the function for each branch.
@oli-obk As this is just a small convenience improvement, having to write (and for reading look up) a macro for it may defy the purpose of simplifying it.

Another use case for this could be the `from` function.

Of course if you say this is an introduction of too much "magic" for a too little convenience it adds, I see your point. That estimate is up to you. After all I was only suggesting ;-)

jonas-schievink:
Well, I tend to dislike unintuitive type systems based on too many subtle rules. All static type systems I know of have a very simple and helpful rule: The type of all expressions is known statically. This proposal would punch a hole in that rule, in that the type of the `match` inside the method call no longer has a single type.

The fact that this is even more special cased so it can only happen when the match is used as a function call argument makes reasoning about this even harder (and you may wonder why your code no longer works when you assign the match result to a temporary before passing that to the function).

What's even more surprising is that the method you call this way is actually a lie: A different monomorphization is called depending on which match arm is executed. You don't actually have `Box::<_>::new(match ... {})`, where `_` is an inferred type, but something completely different. AFAIK this kind of "I don't know which method I call" currently only happens with trait objects (well, or function references of course), which are explicitly type-erased, which this isn't (this rule is activated automatically, *on type mismatch*).

arielb1:
Even with trait objects you call an explicit method, e.g. `<fmt::Debug as fmt::Debug>::fmt` - it is just that it can call an unknown function through the vtable.

ConnyOnny:
> The type of all expressions is known statically.

Oh. I didn't think about that. I'm convinced. As nobody else was in favor, I'll close the PR.
Thank you for your time and kind explanations.

PR #1492, opened 2016-02-06
RFC: `..` in patterns

petrochenkov:
Permit the `..` pattern fragment in more contexts.

[Rendered](https://github.com/petrochenkov/rfcs/blob/patdotdot/text/0000-dotdot-in-patterns.md)

### DIFF
--- a/text/0000-dotdot-in-patterns.md
+++ b/text/0000-dotdot-in-patterns.md
@@ -79,6 +79,8 @@ S(.., subpatN-1, subpatN)
 S(subpat1, .., subpatN)
 // **NEW**: Tuple struct patterns, any position with a sublist binding.
 // The binding has a tuple type.
+// By ref bindings are not allowed, because layouts of S(A, B, C, D) and (B, C) are not necessarily
+// compatible (e.g. field reordering is possible).
 S(subpat1, binding.., subpatN)
 
 // **NEW**: Tuple patterns, any position.
@@ -87,6 +89,8 @@ S(subpat1, binding.., subpatN)
 (subpat1, .., subpatN)
 // **NEW**: Tuple patterns, any position with a sublist binding.
 // The binding has a tuple type.
+// By ref bindings are not allowed, because layouts of (A, B, C, D) and (B, C) are not necessarily
+// compatible (e.g. field reordering is possible).
 (subpat1, binding.., subpatN)
 
 // Slice patterns, the last position.
@@ -96,6 +100,7 @@ S(subpat1, binding.., subpatN)
 // Slice patterns, any other position.
 [subpat1, .., subpatN]
 // Slice patterns, any of the above with a subslice binding.
+// By ref bindings are allowed, slices and subslices always have compatible layouts.
 [subpat1, binding.., subpatN]
 ```
 
@@ -113,7 +118,7 @@ None.
 # Alternatives
 [alternatives]: #alternatives
 
-None.
+Do not permit sublist bindings in tuples and tuple structs at all.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-dotdot-in-patterns.md
+++ b/text/0000-dotdot-in-patterns.md
@@ -1,0 +1,124 @@
+- Feature Name: dotdot_in_patterns
+- Start Date: 2016-02-06
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Permit the `..` pattern fragment in more contexts.
+
+# Motivation
+[motivation]: #motivation
+
+The pattern fragment `..` can be used in some patterns to denote several elements in list contexts.
+However, it doesn't always compiles when used in such contexts.
+One can expect the ability to match tuple variants like `V(u8, u8, u8)` with patterns like
+`V(x, ..)` or `V(.., z)`, but the compiler rejects such patterns currently despite accepting
+very similar `V(..)`.
+
+This RFC is intended to "complete" the feature and make it work in all possible list contexts,
+making the language a bit more convenient and consistent.
+
+# Detailed design
+[design]: #detailed-design
+
+Let's list all the patterns currently existing in the language, that contain lists of subpatterns:
+
+```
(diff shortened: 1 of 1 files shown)


petrochenkov:
cc https://github.com/rust-lang/rust/issues/31443

text/0000-dotdot-in-patterns.md
```diff
@@ -0,0 +1,124 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+None.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Sublist binding syntax conflicts with possible exclusive range patterns
+`begin .. end`/`begin..`/`..end`. This problem already exists for slice patterns and has to be
```
oli-obk: Slice patterns are unstable (and buggy), and thus can easily be changed.

text/0000-dotdot-in-patterns.md
```diff
@@ -0,0 +1,124 @@
...
+
+// Tuple struct patterns, the last and the only position, no extra subpatterns allowed.
+S(..)
+// **NEW**: Tuple struct patterns, any position.
+S(subpat1, subpat2, ..)
+S(.., subpatN-1, subpatN)
+S(subpat1, .., subpatN)
+// **NEW**: Tuple struct patterns, any position with a sublist binding.
+// The binding has a tuple type.
+S(subpat1, binding.., subpatN)
```
oli-obk: these are somewhat problematic. Afaik there's no guarantee that `(B, C)` has the same memory layout as the last two parts of `(A, B, C)`, so taking a slice into a tuple might not be so trivial.

oli-obk:
Should `S(a, b, ..)` be allowed for `struct S(A, B);` ? For structs it currently is: http://is.gd/TsMz6N .

text/0000-dotdot-in-patterns.md
petrochenkov: Extensions proposed in this RFC are going to be unstable too and are supposed to be changed together with slice patterns if changed at all.

petrochenkov:
>Should S(a, b, ..) be allowed for struct S(A, B); ? 

Yes, `..` means 0 or more subpatterns.

text/0000-dotdot-in-patterns.md
petrochenkov: Hmm, this is an interesting observation.
It means that at least by ref bindings have to be restricted.
I'll update the RFC.

nikomatsakis:
This was pretty Always The Plan -- but we never did it because it involved work. I'm basically :+1: though I've not read the details of the RFC.

petrochenkov:
Implementation (without  sub-list bindings): https://github.com/rust-lang/rust/pull/32079

nikomatsakis:
Nominating for FCP.

petrochenkov:
I should probably strike out the sub-tuple bindings `(a, b, c..)` from this RFC given the discussion in https://github.com/rust-lang/rfcs/pull/1582 and https://github.com/rust-lang/rfcs/issues/376, at least for now.

nikomatsakis:

nikomatsakis:
@petrochenkov very well... to be honest I've not read this closely, though I'm in favor of the spirit. We had decided to put this into FCP, but would you rather we wait? (We can keep your proposed edit in mind, I imagine.)

text/0000-dotdot-in-patterns.md
```diff
@@ -0,0 +1,129 @@
...
+
+// Slice patterns, the last position.
+[subpat1, subpat2, ..]
+// Slice patterns, the first position.
+[.., subpatN-1, subpatN]
+// Slice patterns, any other position.
+[subpat1, .., subpatN]
+// Slice patterns, any of the above with a subslice binding.
+// By ref bindings are allowed, slices and subslices always have compatible layouts.
+[subpat1, binding.., subpatN]
```
nikomatsakis: What does this `binding` mean? This feels like it overlaps with slice patterns, which (I think) still require some further RFC to "complete"? I can't remember if we postponed that work or not. :)

nikomatsakis:
I'd be (more) in favor of a variant of this that excludes all "bindings". In other words, allowing:

```
(pat*, .., pat*)
Foo(pat*, .., pat*)
```

and I'd probably prefer to exclude `[]` patterns which I think ought to be handled separately. 

The binding forms introduce complicated semantics.

text/0000-dotdot-in-patterns.md
petrochenkov: >What does this binding mean?

A subslice pattern, see above - "The binding is not actually a binding, but one more pattern ..."

petrochenkov:
@nikomatsakis 
>I'd probably prefer to exclude [] patterns

This RFC doesn't propose any changes to slice patterns, it only uses them for analogy.

>I'd be (more) in favor of a variant of this that excludes all "bindings".

With this I [agree](https://github.com/rust-lang/rfcs/pull/1492#issuecomment-215266794).

nikomatsakis:
@petrochenkov 

> This RFC doesn't propose any changes to slice patterns, it only uses them for analogy.

Huh. OK, I must have misread it. So much the better.

bjz:
I missed that subtlety too - maybe removing the slice examples might make it clearer?

nikomatsakis:
**Huzzah!** The @rust-lang/lang team has decided **to accept** this RFC. Specifically:

- no bindings
- no slice patterns
- basically the subset described [in this comment](https://github.com/rust-lang/rfcs/pull/1492#issuecomment-218855413), which @petrochenkov seems to agree with.

nikomatsakis:
Tracking issue: https://github.com/rust-lang/rust/issues/33627

RaldyDaldy:
> cc [rust-lang/rust#31443](https://github.com/rust-lang/rust/issues/31443)

RaldyDaldy:
Reply

Op vr 3 apr. 2020 18:55 schreef robbieS <notifications@github.com>:

RaldyDaldy:
> > cc [rust-lang/rust#31443](https://github.com/rust-lang/rust/issues/31443)

Reply

PR #1504, opened 2016-02-21
Add support for 128-bit integers

Amanieu:
Relevant issue: #521

[Rendered](https://github.com/Amanieu/rfcs/blob/int128/text/0000-int128.md)

### DIFF
--- a/text/0000-int128.md
+++ b/text/0000-int128.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-This RFC adds the `i128` and `u128` types to Rust. Because these types are not available on all platforms, a new target flag (`target_has_int128`) is added to allow users to check whether 128-bit integers are supported. The `i128` and `u128` are not added to the prelude, and must instead be explicitly imported with `use core::{i128, u128}`.
+This RFC adds the `i128` and `u128` types to Rust. The `i128` and `u128` are not added to the prelude, and must instead be explicitly imported with `use core::{i128, u128}`.
 
 # Motivation
 [motivation]: #motivation
@@ -16,20 +16,56 @@ Some algorithms need to work with very large numbers that don't fit in 64 bits, 
 # Detailed design
 [design]: #detailed-design
 
-From a quick look at Clang's source, 128-bit integers are supported on all 64-bit platforms and a few 32-bit ones (those with 64-bit registers: x32 and MIPS n32). To allow users to determine whether 128-bit integers are available, a `target_has_int128` cfg is added. The `i128` and `u128` types are only available when this flag is set.
+The `i128` and `u128` types are not added to the Rust prelude since that would break compatibility. Instead they must be explicitly imported with `use core::{i128, u128}` or `use std::{i128, u128}`.
 
-The actual `i128` and `u128` types are not added to the Rust prelude since that would break compatibility. Instead they must be explicitly imported with `use core::{i128, u128}` or `use std::{i128, u128}`. This will also catch attempts to use 128-bit integers when they are not supported by the underlying platform since the import will fail if `target_has_int128` is not defined.
+Implementation-wise, this should just be a matter of adding a new primitive type to the compiler and adding trait implementations for `i128`/`u128` in libcore. Literals will need to be extended to support `i128`/`u128`.
 
-Implementation-wise, this should just be a matter of adding a new primitive type to the compiler and adding trait implementations for `i128`/`u128` in libcore. A new entry will need to be added to target specifications to specify whether the target supports 128-bit integers.
+LLVM fully supports 128-bit integers on all architectures, however it will emit calls to functions in `compiler-rt` for many operations such as multiplication and division (addition and subtraction are implemented natively). However, `compiler-rt` only provides the functions for 128-bit integers on 64-bit platforms (`#ifdef __LP64__`). We will need to provide our own implementations of the following functions to allow `i128`/`u128` to be available on all architectures:
 
-One possible complication is that primitive types aren't currently part of the prelude, instead they are directly added to the global namespace by the compiler. The new `i128` and `u128` types will behave differently and will need to be explicitly imported.
-
-Another possible issue is that a `u128` can hold a very large number that doesn't fit in a `f32`. We need to make sure this doesn't lead to any `undef`s from LLVM.
+```c
+// si_int = i32
+// su_int = u32
+// ti_int = i128
+// tu_int = u128
+ti_int __absvti2(ti_int a);
+ti_int __addvti3(ti_int a, ti_int b);
+ti_int __ashlti3(ti_int a, si_int b);
+ti_int __ashrti3(ti_int a, si_int b);
+si_int __clzti2(ti_int a);
+si_int __cmpti2(ti_int a, ti_int b);
+si_int __ctzti2(ti_int a);
+ti_int __divti3(ti_int a, ti_int b);
+si_int __ffsti2(ti_int a);
+ti_int __fixdfti(double a);
+ti_int __fixsfti(float a);
+tu_int __fixunsdfti(double a);
+tu_int __fixunssfti(float a);
+double __floattidf(ti_int a);
+float __floattisf(ti_int a);
+double __floatuntidf(tu_int a);
+float __floatuntisf(tu_int a);
+ti_int __lshrti3(ti_int a, si_int b);
+ti_int __modti3(ti_int a, ti_int b);
+ti_int __muloti4(ti_int a, ti_int b, int* overflow);
+ti_int __multi3(ti_int a, ti_int b);
+ti_int __mulvti3(ti_int a, ti_int b);
+ti_int __negti2(ti_int a);
+ti_int __negvti2(ti_int a);
+si_int __parityti2(ti_int a);
+si_int __popcountti2(ti_int a);
+ti_int __subvti3(ti_int a, ti_int b);
+si_int __ucmpti2(tu_int a, tu_int b);
+tu_int __udivmodti4(tu_int a, tu_int b, tu_int* rem);
+tu_int __udivti3(tu_int a, tu_int b);
+tu_int __umodti3(tu_int a, tu_int b);
+```
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-It adds a type to the language that may or may not be present depending on the target architecture.
+One possible complication is that primitive types aren't currently part of the prelude, instead they are directly added to the global namespace by the compiler. The new `i128` and `u128` types will behave differently and will need to be explicitly imported.
+
+Another possible issue is that a `u128` can hold a very large number that doesn't fit in a `f32`. We need to make sure this doesn't lead to any `undef`s from LLVM. See [this comment](https://github.com/rust-lang/rust/issues/10185#issuecomment-110955148), and [this example code](https://gist.github.com/Amanieu/f87da5f0599b343c5500).
 
 # Alternatives
 [alternatives]: #alternatives
@@ -39,4 +75,4 @@ There have been several attempts to create `u128`/`i128` wrappers based on two `
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-How should 128-bit literals be handled? The easiest solution would be to limit integer literals to 64 bits, which is what GCC does (no support for `__int128` literals).
+None
--- a/text/0000-int128.md
+++ b/text/0000-int128.md
@@ -27,15 +27,9 @@ LLVM fully supports 128-bit integers on all architectures, however it will emit 
 // su_int = u32
 // ti_int = i128
 // tu_int = u128
-ti_int __absvti2(ti_int a);
-ti_int __addvti3(ti_int a, ti_int b);
 ti_int __ashlti3(ti_int a, si_int b);
 ti_int __ashrti3(ti_int a, si_int b);
-si_int __clzti2(ti_int a);
-si_int __cmpti2(ti_int a, ti_int b);
-si_int __ctzti2(ti_int a);
 ti_int __divti3(ti_int a, ti_int b);
-si_int __ffsti2(ti_int a);
 ti_int __fixdfti(double a);
 ti_int __fixsfti(float a);
 tu_int __fixunsdfti(double a);
@@ -48,14 +42,6 @@ ti_int __lshrti3(ti_int a, si_int b);
 ti_int __modti3(ti_int a, ti_int b);
 ti_int __muloti4(ti_int a, ti_int b, int* overflow);
 ti_int __multi3(ti_int a, ti_int b);
-ti_int __mulvti3(ti_int a, ti_int b);
-ti_int __negti2(ti_int a);
-ti_int __negvti2(ti_int a);
-si_int __parityti2(ti_int a);
-si_int __popcountti2(ti_int a);
-ti_int __subvti3(ti_int a, ti_int b);
-si_int __ucmpti2(tu_int a, tu_int b);
-tu_int __udivmodti4(tu_int a, tu_int b, tu_int* rem);
 tu_int __udivti3(tu_int a, tu_int b);
 tu_int __umodti3(tu_int a, tu_int b);
 ```
--- a/text/0000-int128.md
+++ b/text/0000-int128.md
@@ -1,0 +1,42 @@
+- Feature Name: int128
+- Start Date: 21-02-2016
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC adds the `i128` and `u128` types to Rust. Because these types are not available on all platforms, a new target flag (`target_has_int128`) is added to allow users to check whether 128-bit integers are supported. The `i128` and `u128` are not added to the prelude, and must instead be explicitly imported with `use core::{i128, u128}`.
+
+# Motivation
+[motivation]: #motivation
+
+Some algorithms need to work with very large numbers that don't fit in 64 bits, such as certain cryptographic algorithms. One possibility would be to use a BigNum library, but these use heap allocation and tend to have high overhead. LLVM has support for very efficient 128-bit integers, which are exposed by Clang in C as the `__int128` type.
+
+# Detailed design
+[design]: #detailed-design
+
+From a quick look at Clang's source, 128-bit integers are supported on all 64-bit platforms and a few 32-bit ones (those with 64-bit registers: x32 and MIPS n32). To allow users to determine whether 128-bit integers are available, a `target_has_int128` cfg is added. The `i128` and `u128` types are only available when this flag is set.
+
+The actual `i128` and `u128` types are not added to the Rust prelude since that would break compatibility. Instead they must be explicitly imported with `use core::{i128, u128}` or `use std::{i128, u128}`. This will also catch attempts to use 128-bit integers when they are not supported by the underlying platform since the import will fail if `target_has_int128` is not defined.
+
+Implementation-wise, this should just be a matter of adding a new primitive type to the compiler and adding trait implementations for `i128`/`u128` in libcore. A new entry will need to be added to target specifications to specify whether the target supports 128-bit integers.
+
+One possible complication is that primitive types aren't currently part of the prelude, instead they are directly added to the global namespace by the compiler. The new `i128` and `u128` types will behave differently and will need to be explicitly imported.
+
+Another possible issue is that a `u128` can hold a very large number that doesn't fit in a `f32`. We need to make sure this doesn't lead to any `undef`s from LLVM.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It adds a type to the language that may or may not be present depending on the target architecture.
+
+# Alternatives
+[alternatives]: #alternatives
+
+There have been several attempts to create `u128`/`i128` wrappers based on two `u64` values, but these can't match the performance of LLVM's native 128-bit integers.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How should 128-bit literals be handled? The easiest solution would be to limit integer literals to 64 bits, which is what GCC does (no support for `__int128` literals).
--- a/text/0000-int128.md
+++ b/text/0000-int128.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-This RFC adds the `i128` and `u128` types to Rust. The `i128` and `u128` are not added to the prelude, and must instead be explicitly imported with `use core::{i128, u128}`.
+This RFC adds the `i128` and `u128` primitive types to Rust.
 
 # Motivation
 [motivation]: #motivation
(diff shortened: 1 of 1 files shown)


Ticki:
:+1:

They were originally removed citing the lack of compiler backend support, however this was incorrect, they are completely supported by LLVM, so there should not be anything blocking it.

nagisa:
> they are completely supported by LLVM

AFAIR, the actual (underlying) reason was LLVM’s support for operations on f128 being slightly(?) broken. Since the 128 floats were being removed, 128 integers went along with it.

Amanieu:
I don't think Rust has ever had support for 128-bit integers. The f128 stuff was completely separate.

Ticki:
Right, I misread the title as 128-bit floats.

text/0000-int128.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Some algorithms need to work with very large numbers that don't fit in 64 bits, such as certain cryptographic algorithms. One possibility would be to use a BigNum library, but these use heap allocation and tend to have high overhead. LLVM has support for very efficient 128-bit integers, which are exposed by Clang in C as the `__int128` type.
+
+# Detailed design
+[design]: #detailed-design
+
+From a quick look at Clang's source, 128-bit integers are supported on all 64-bit platforms and a few 32-bit ones (those with 64-bit registers: x32 and MIPS n32). To allow users to determine whether 128-bit integers are available, a `target_has_int128` cfg is added. The `i128` and `u128` types are only available when this flag is set.
```
nagisa: LLVM supports arbitrary sized integers (and operations on them) and will software-emulate operations on them if necessary – as inefficient as they get. LLVM produces perfectly correct assembly even on targets where clang claims there’s no support, so looking at what clang does is not the way to go here IMO.

Ticki: Furthermore, the emulation (which might or might not happen) is often performing pretty decent.

text/0000-int128.md
```diff
@@ -0,0 +1,42 @@
...
+
+The actual `i128` and `u128` types are not added to the Rust prelude since that would break compatibility. Instead they must be explicitly imported with `use core::{i128, u128}` or `use std::{i128, u128}`. This will also catch attempts to use 128-bit integers when they are not supported by the underlying platform since the import will fail if `target_has_int128` is not defined.
+
+Implementation-wise, this should just be a matter of adding a new primitive type to the compiler and adding trait implementations for `i128`/`u128` in libcore. A new entry will need to be added to target specifications to specify whether the target supports 128-bit integers.
+
+One possible complication is that primitive types aren't currently part of the prelude, instead they are directly added to the global namespace by the compiler. The new `i128` and `u128` types will behave differently and will need to be explicitly imported.
+
+Another possible issue is that a `u128` can hold a very large number that doesn't fit in a `f32`. We need to make sure this doesn't lead to any `undef`s from LLVM.
```
nagisa: I don’t really see how that could be a problem in practice. If a u128 doesn’t fit into f32, a +∞ will happen which is a perfectly valid value?

text/0000-int128.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It adds a type to the language that may or may not be present depending on the target architecture. This could lead to surprises, bu
```
nagisa: “bu”…?

text/0000-int128.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+There have been several attempts to create `u128`/`i128` wrappers based on two `u64` values, but these can't match the performance of LLVM's native 128-bit integers.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How should 128-bit literals be handled? The easiest solution would be to limit integer literals to 64 bits, which is what GCC does (no support for `__int128` literals).
```
nagisa: If we add support for i128, we must, in some way, introduce some support for i128 calculations in the compiler anyway for cases like `const X: i128 = !0`, so it seems to me like supporting long literals is the smallest of all implementation woes in this regard.

text/0000-int128.md
Amanieu: I just ran a few quick tests on LLVM bitcode with `i128` and `i192` ([test.ll](https://gist.github.com/023335c64d219a509ec4)), here are my results (LLVM 3.7.1):
- `add` and `sub` are compiled down to a sequence of add with carry instructions for both `i128` and `i192`.
- `mul` and `udiv` are compiled to a call to `__multi3` and `__udivti3` for `i128`. These functions are implemented in `compiler-rt`.
- `mul` and `udiv` generate an LLVM error for `i192`: `LLVM ERROR: Unsupported library call operation!`
- `compiler-rt` contains variations of each function for 32, 64 and 128-bit integers (`__mulsi3`, `__muldi3`, `__multi3`). However, the 128-bit versions are only present on 64-bit platforms (`#ifdef __LP64__`).

In conclusion, LLVM doesn't support many operations on arbitrary sized integers. While we could in theory support 128-bit integers on all architectures, this would require extending `compiler-rt` on 32-bit architectures.

nagisa:
All in all, I’m all +1 for the general idea, but I’m not so keen on following what clang does. The RFC states the integers are not (can’t be?) supported on certain platforms. If there’s specific examples of such platforms, it could be good to know some, because otherwise it seems like LLVM’s emulation covers the ground.

text/0000-int128.md
Amanieu: See [this comment](https://github.com/rust-lang/rust/issues/10185#issuecomment-110955148).

text/0000-int128.md
nagisa: Aha! I didn’t think of multiplication. How silly of me. I’m not too surprised about `i192` – it’s not technically a power-of-two-bytes, but it certainly comes across as strange to me that there’s no generic-implementation-in-C (even if not fast at all) for more complex operations like multiplication for at least i128.

Amanieu:
As I said in [my comment above](https://github.com/rust-lang/rfcs/pull/1504#discussion_r53571682), it seems that LLVM's emulation does *not* cover all the needed operations.

text/0000-int128.md
nagisa: Right, [in my tests so far](https://gist.github.com/nagisa/a9df2bf778e2583b1945) an `inf` will get printed for the specified target. Here’s my thoughts:

I want to require this to reliably result in ∞ on “overflow” regardless of what LLVM thinks the right behaviour should be (apparently UV). I’m inclined to think this is in no way related (or specified) in LLVM itself, but rather a quirk in `compiler-rt`.

Uninteresting-ly enough, compiling and linking the aforementioned file for i386 target fails with `undefined reference to '__floatuntisf'`. I don’t want this feature landing in a “works here, but not there” state, when it is by all means technically possible to make it work regardless of target capabilities, by employing some more or less expensive software emulation. If `compiler-rt` does not support it we can roll our own compiler-rt which supports our [tier one-to-three](https://doc.rust-lang.org/book/getting-started.html#platform-support) platforms for all necessary operations or wait/implement the necessary support in the compiler-rt.

Amanieu: The `undef` only occurs when the input value is a constant: https://gist.github.com/Amanieu/f87da5f0599b343c5500

Regarding supporting `i128` on all architectures, I agree with you. LLVM already has full support for this, it's just `compiler-rt` that needs to be modified to support 128-bit integer operations on 32-bit systems. I'll update the RFC accordingly.

text/0000-int128.md
Aatch: LLVM has some `APInt` and `APFloat` types we could use. Though we could also make our own.

Amanieu:
I've updated the RFC to support 128-bit integers on all architectures, even 32-bit ones. This will require implementing some functions that `compiler-rt` doesn't provide, but this shouldn't be too much trouble.

text/0000-int128.md
```diff
@@ -0,0 +1,78 @@
...
+si_int __ucmpti2(tu_int a, tu_int b);
+tu_int __udivmodti4(tu_int a, tu_int b, tu_int* rem);
+tu_int __udivti3(tu_int a, tu_int b);
+tu_int __umodti3(tu_int a, tu_int b);
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One possible complication is that primitive types aren't currently part of the prelude, instead they are directly added to the global namespace by the compiler. The new `i128` and `u128` types will behave differently and will need to be explicitly imported.
```
petrochenkov: It's still possible to improve name resolution for primitive types and prelude backward compatibly.
I'd define the new `i128`/`u128` in some dedicated module like `core::primitives`, there's a good chance other primitive types will be defined there as well in some time.

Ticki: Adding primitive types is a backwards compatible thing to do (for 99.999% of all code).

petrochenkov: It'll break all the `i128`/`u128` emulations using the same names, like https://crates.io/crates/extprim, and their dependencies. (Also, using playpen often prevents posting misinformation.)

Ticki: u128, i128 are not keywords, so that one is ruled out. However, defining types shadowing primitive types results in E0317, however I think this is justified, since the crates depending on crates using these emulations can easily drop the dependencies, without changing any code (except `use` and `extern crate`).

yberreby:
:+1:

retep998:
Yes please, I'd love to have 128 bit integers.

hauleth:
It could be behind stable flag that would give user more descriptive output when compiling on unsupported platform.

Amanieu:
> It could be behind stable feature gate that would give user more descriptive output when compiling on unsupported platform.
That shouldn't be needed, this RFC proposes adding 128-bit integers for *all* architectures. LLVM is capable of doing this if we implement some runtime library functions.

hauleth:
Oh. I didn't see that. Too late, too old, don't see things…

ubsan:
Seems good to me. I'd love to have 128 bit integers in Rust (if we're sure we're able to support it on all platforms).

gnzlbg:
what are the advantages of this RFC over the emulations we currently have?

Ticki:
@gnzlbg Performance on supported platforms. It will be an order of magnitude faster in some cases.

hauleth:
That x86-64 and some other CPUs already have native support for 128-bit integers. IIRC Intel provides support for 256 and 512-bit integers in their new procesors. 

--
Łukasz Jan Niemier

Dnia 23 lut 2016 o godz. 08:45 gnzlbg <notifications@github.com> napisał(a):

nagisa:
> IIRC Intel provides support for 256 and 512-bit integers in their new procesors.

Aren’t you confusing it with the AVX/SSE? I’m not aware of any features/instructions even in the newest intel processors which would allow for operations on 128/256/512-bit integers.

sdroege:
There's at least divq, idivq, mulq, imulq on all x86-64 for 128 bit integer multiplication/division with 64 bit operands

nagisa:
@sdroege Right, x86 does multiplication and division on double-word-sized operands to allow for all results to fit. This is true all the way back to the 8086 where multiplication result was `dx:ax` AFAIR.

What you cannot do, though, is given a 128bit number in `rdx:rax` add it to another 128 bit immediate or shift `rdx:rax` by 76 or any other common operation in a single instruction. This is what I’d consider a native support for 128-bit integers.

oli-obk:
note that casting `i128` or `u128` to `f32` can result in UB: https://github.com/rust-lang/rust/issues/10185#issuecomment-110955148

sdroege:
@nagisa True (well, it's not available in old 32 bit x86, is it?). Fortunately that's all I need for my own use case here https://github.com/sdroege/rust-muldiv

But as long as LLVM emulates i128/u128 and/or all (other) operations are implemented in Rust for now, this seems like a useful feature to have

gnzlbg:
@Ticki wrote:

> @gnzlbg Performance on supported platforms. It will be an order of magnitude faster in some cases.

Why can't the current crates call the llvm intrinsics directly?

nagisa:
@gnzlbg because `add i128* %x, i128 123` is not an intrinsic in the first place.

gnzlbg:
Oh, I thought llvm would offer intrinsics for 128bit integers in the same way it does offer intrinsics for SIMD.

eternaleye:
I'll also note that RISC-V defines a 128-bit-pointer-and-register variant of the architecture, and there seems to be some interest in that for datacenter use cases. In addition, HP's "Machine" trends in the same direction. Sooner or later, this is likely to need done regardless - doing so sooner means less time for code that uses those identifiers to crop up.

Aatch:
@sdroege The "128-bit" multiplication and division instructions aren't really 128-bit, if they were, I'd be using them in Ramp. Instead the division instructions do two-word by one-word division and multiplication only has a two-word result. Furthermore, for division, the quotient and remainder have to fit into the registers otherwise it triggers a CPU fault, this means that doing the equivalent of `50 / 2` wouldn't work if you just used `div`/`idiv`.

sdroege:
@Aatch I'm aware of that :) It could still be used as part of a proper 128 bit integer emulation though

oli-obk:
Note that rustc currently disallows integral literals larger than `2^64`. So the larger literals should be allowed in tandem with introducing 128bit types.

arthurprs:
Although it increases rust surface area it's very useful in a number of cases. :+1:

rkruppe:
:+1: from me as well. If the runtime library support is well optimized, this might give a virtually free performance boost to numerous applications, the most obvious being bignum implementations. We have one of those in libcore so we'd be dogfooding. Plus the point about future architectures @eternaleye &mdash; it's good precedent if we manage to add new primitive types without too much trouble.

alexcrichton:
Thanks for the RFC @Amanieu! I've got a few questions and thoughts, and I have a feeling you can help me clarify things as well :)

* First off, the RFC states that we're going to add two new primitive types, but they need to be imported. This seems odd to me as you don't need to import any other primitive types, and otherwise if you need to import them it seems like they aren't necessarily a primitive type? I'd be curious to dive into two aspects here:
  * Why is adding new types to the default namespace backwards incompatible? Can you elaborate on this? Perhaps this isn't so bad in the face of https://github.com/rust-lang/rfcs/pull/1560?
  * If we do go with a you-must-import-this design, what would that actually look like? Apart from the oddness of having to import a primitive, you'd also look like you were importing a module (like `core::u32`) when it actually brings a type into scope. What does the source code for libcore look like after this change?
* This RFC claims "LLVM fully supports 128-bit integers on all architectures", but then it says that lots of calls are made to compiler-rt which is in turn only defined for 64-bit platforms? That seemed to very quickly go from "full support" to "probably works on 64-bit". To me this is very worrisome where if we start going *beyond* what LLVM currently does we're likely to hit tons of bugs pretty quickly. You've got a list of functions that we have to implement, but are we sure that it's the list of *all possible* functions we'd have to implement.

  Put another way, what are we really getting from using a native LLVM `i128` type? Why is that superior to, say, `struct i128([i64; 2])` with a bunch of trait impls with concrete code (where we perhaps call out to compiler-rt ourselves in some cases). Are there some operations that LLVM can lower, like I guess addition, to a few instructions which we really need to have LLVM do for the performance?

  I guess all that put another way, I would *love* to be able to just add a handful of intrinsics to start playing around with 128 bit types. If we could add an intrinsic for whatever compiler-operations we must have, and then manually implement other operations as we fill them out, it sounds like that would give us a much clearer platform compatibility story and leeway to test out the implementation, find bugs in LLVM/etc, and prep for adding a full-fledged primitive type to the language.
* The RFC also says that "literals will need to be extended to support i128/u128", which to me implies that we'll need to add parsing routines to libcore as well as formatting routines. Are these both straightforward to implement and they don't require a bunch of extra runtime support to be added to libcore?

I guess overall this RFC feels a little light to me. For an RFC to **add a new primitive type to the language** I would personally expect an existing proof-of-concept implementation, some thorough testing to show that it works well across platforms, initial library routines to work with the types, as well as benchmarks/numbers to show the benefit of using the new primitive type over what we have today. I feel that this only barely dives into the details of adding a new primitive to the language and would love to see the RFC expanded to have some more information like this.

text/0000-int128.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Some algorithms need to work with very large numbers that don't fit in 64 bits, such as certain cryptographic algorithms. One possibility would be to use a BigNum library, but these use heap allocation and tend to have high overhead. LLVM has support for very efficient 128-bit integers, which are exposed by Clang in C as the `__int128` type.
+
+# Detailed design
+[design]: #detailed-design
+
+The `i128` and `u128` types are not added to the Rust prelude since that would break compatibility. Instead they must be explicitly imported with `use core::{i128, u128}` or `use std::{i128, u128}`.
```
petrochenkov: After https://github.com/rust-lang/rust/pull/32131 addition of new primitive types is not a breaking change, so I think everything related to name resolution can be removed from this RFC and `u128`/`i128` can be resolved as all other primitive types.

nagisa:
> Put another way, what are we really getting from using a native LLVM i128 type? Why is that superior to, say, struct i128([i64; 2]) with a bunch of trait impls with concrete code (where we perhaps call out to compiler-rt ourselves in some cases). Are there some operations that LLVM can lower, like I guess addition, to a few instructions which we really need to have LLVM do for the performance?

Yes, LLVM will lower `add i128 %1, i128 %2` to something like following two instrunctions on x86_64

```asm
add num1_low, num2_low
adc num1_high, num2_high
```

and, I believe, similarly good code on many other targets. While we could implement something as efficient ourselves, I’m not sure there’s anything to gain over just letting LLVM to generate such code for all targets it supports.

alexcrichton:
@nagisa ah yeah that's what I would expect, thanks! I wonder though if LLVM basically only implements "nice codegen" for simple operations like addition and subtraction? (maybe shifts too?) If it ends up being that 90% of it is a library implementation then it seems a waste to add a whole primitive for now when we could have a few choice intrinsics perhaps. If, however, the it's more of a 80% compiler magic and 20% library implementation situation then it would make more sense to add a whole new primitive rather than rely so much on libraries.

nagisa:
@alexcrichton It could very well be that sets of instructions that directly lower differ between the targets greatly all up to the potential targets which support 128 bit operations natively.

I did some tests and targeting x86_64 and it compiled down following things:

* add, sub: 2 insns
* mul: 5 insns
* shift: 6 insns (including some overflow checks)
* and/or/xor: 2/3 insns
* div/rem: **a call**

i686:

* add, sub: 4 insns
* mul/div/rem: **a call**
* shift: a *lot* of insns
* and/or/xor: 2/3 insns(?)

aarch64: 

* mul: insns
* div/rem: **a call**

armv7:

* mul/div/rem: **a call**

So the observations match pretty closely to the list of the functions provided by the RFC. My point here being that LLVM will avoid using these as much as possible and lower directly when possible. Moreover, using LLVM’s native support also implies LLVM understands the operations and can optimise them, whereas something like `[i64; 2]` might be more tricky.

Amanieu:
> - First off, the RFC states that we're going to add two new primitive types, but they need to be imported. This seems odd to me as you don't need to import any other primitive types, and otherwise if you need to import them it seems like they aren't necessarily a primitive type? I'd be curious to dive into two aspects here:
>  - Why is adding new types to the default namespace backwards incompatible? Can you elaborate on this? Perhaps this isn't so bad in the face of #1560?
>  - If we do go with a you-must-import-this design, what would that actually look like? Apart from the oddness of having to import a primitive, you'd also look like you were importing a module (like core::u32) when it actually brings a type into scope. What does the source code for libcore look like after this change?

The original reason for this was to avoid break crates like [extprim](https://github.com/kennytm/extprim) which provide library implementations of `u128` and `i128`. However this is no longer necessary now that rust-lang/rust#32131 has landed, since it removes restrictions on shadowing primitive types. I will amend the RFC to have `i128` and `u128` treated like normal primitive types.

> This RFC claims "LLVM fully supports 128-bit integers on all architectures", but then it says that lots of calls are made to compiler-rt which is in turn only defined for 64-bit platforms? That seemed to very quickly go from "full support" to "probably works on 64-bit".

I meant that the LLVM code generator fully supports 128-bit integers, but will fall back to compiler-rt for some operations. This is exactly the same situation that we currently have with `u64` and `i64` on 32-bit platforms: operations such as multiplication and division are lowered to compiler-rt calls.

> You've got a list of functions that we have to implement, but are we sure that it's the list of all possible functions we'd have to implement.

I double-checked the LLVM source code while generating the list and these are the only functions that LLVM will emit calls to (that are related to 128-bit integers).

> - The RFC also says that "literals will need to be extended to support i128/u128", which to me implies that we'll need to add parsing routines to libcore as well as formatting routines. Are these both straightforward to implement and they don't require a bunch of extra runtime support to be added to libcore?

The formatting and parsing code is a fairly trivial extension of the existing code that handles `i64` and `u64`. I'm more worried about how a stage0 compiler is going to handle 128-bit literals, but then again I'm not too familiar with that part of the compiler.

> I wonder though if LLVM basically only implements "nice codegen" for simple operations like addition and subtraction? (maybe shifts too?) If it ends up being that 90% of it is a library implementation then it seems a waste to add a whole primitive for now when we could have a few choice intrinsics perhaps. If, however, the it's more of a 80% compiler magic and 20% library implementation situation then it would make more sense to add a whole new primitive rather than rely so much on libraries.

Here's a rough breakdown of what operations are handled natively vs using a library call:
- add/sub/neg: native, including checked overflow variants
- compare (eq/ne/gt/ge/lt/le): native
- bitwise and/or/xor/not: native
- shift left/right: native on most architectures (some use libcalls, but x86 and arm have native support)
- bit counting, parity, leading/trailing ones/zeroes: native
- byte swapping: native
- mul/div/mod: libcall (including mul_with_overflow)
- conversion to/from f32/f64: libcall

So basically it's just multiplications, divisions and float conversions that rely on library calls, everything else is handled natively.

> I guess overall this RFC feels a little light to me. For an RFC to add a new primitive type to the language I would personally expect an existing proof-of-concept implementation, some thorough testing to show that it works well across platforms, initial library routines to work with the types, as well as benchmarks/numbers to show the benefit of using the new primitive type over what we have today. I feel that this only barely dives into the details of adding a new primitive to the language and would love to see the RFC expanded to have some more information like this.

One issue is that we need `i128` support in the language first before we are able to implement these library functions because it is needed to match the calling convention expected by LLVM. It is not possible to correctly represent these function prototypes without a `i128` type. And writing these in C is also not an option because `__int128` is not supported on 32-bit architectures.

Amanieu:
@alexcrichton 
I updated the RFC with your feedback. It should now explain better what the implementation plan for this feature is.

alexcrichton:
It seems that we'll have to implement the library routines in Rust for non-64-bit platforms, right? In that case, is there an implementation of these functions that we can use already or are we gonna be forging new territory here?

I guess one question I might also have is why does C not support int128 on 32-bit platforms? Maybe that's a reason we should be wary of such as well?

Amanieu:
> It seems that we'll have to implement the library routines in Rust for non-64-bit platforms, right? In that case, is there an implementation of these functions that we can use already or are we gonna be forging new territory here?

 We can just copy/port the existing implementations in compiler-rt.

> I guess one question I might also have is why does C not support int128 on 32-bit platforms? Maybe that's a reason we should be wary of such as well?

I'm not quite sure, but my best guess would be that it was done this way in Clang for compatibility with GCC.

jmesmon:
Regarding common C impls not having 128 bit integers when targeting 32bit machines, it might be a gambit to avoid enlarging the size of `uintmax_t` and `intmax_t` (which would potentially slow down operations on those types).

Amanieu:
> Regarding common C impls not having 128 bit integers when targeting 32bit machines, it might be a gambit to avoid enlarging the size of uintmax_t and intmax_t (which would potentially slow down operations on those types). It could also be a compatibility issue around users of those types in public APIs of libraries (though I'd really hope no one had done that, it certainly is possible).

That's not it since uintmax_t is a 64-bit type on all systems (at least for GCC and LLVM). My guess is that GCC just felt that supporting 128-bit integers on 32-bit systems was too much trouble.

jmesmon:
@Amanieu huh, I didn't expect that. Thinking it was too much trouble sounds like a reasonable guess. Perhaps they were worried someone might then want 256-bit integers on 64 bit machines? :)

Ericson2314:
The linked atomic thread explains this in detail, but this currently blocks `Atomic{U,I}128` (which has native support on x86_64 btw). Now the atomic intrinsics could be hacked up to avoid depending on this in the first place, but that defeats the cleanliness of a pure-rust software `{u,i}128`. So I vote +1 on the RFC.

aturon:
This RFC is now entering is **final comment period**.

alexcrichton:
I personally feel that the addition of `i128` and `u128` are inevitable in the sense that they're relatively-well supported in LLVM, platforms have enough native support to motivate the speedup over other implementations, and they're a natural extension of the existing primitive types. In that sense this seems reasonable to me *as an idea* to have new primitive types in the language (pending the shadowing support in resolve).

Implementation-wise I would predict that this is going to take *a lot* of work to do well. Much of this RFC leans on "compiler-rt will solve our problems" but historically compiler-rt has been very very difficult to work with, especially when it comes to portability. It seems that the correctness of a number of operations (like multiplication) would rely on the implementations in compiler-rt, which I'm not sure how well battle-tested they are in LLVM.

That being said, I don't personally feel that it should necessarily block this RFC per se. I would still prefer to have a relatively high quality implementation before merging (as opposed to just adding a bunch of "if bits == 128" cases in the compiler), but that's a matter for another time!

aturon:
I'm in agreement with @alexcrichton. To clarify the process here, I think we can accept the RFC but with some specific stipulations about the implementation, which perhaps @alexcrichton could elaborate.

Amanieu:
Note that as it is currently written, this RFC does not propose modifying the compiler-rt code. Instead some functions will need to be implemented *in Rust* because they are required by LLVM on 32-bit platforms and not provided by compiler-rt on those platforms.

rkruppe:
However, this meshes well with the plans discussed (though by no means made official, IIUC) in #34400: Eventually replacing compiler-rt with pure Rust code.

alexcrichton:
My threshold for "this is feature-complete enough to merge as unstable" would be to get a program like past our CI (x86/x86_64 on three major platforms, arm Android on one):

```rust
fn main() {
    let a: i128 = 1;
    println!("{}", a);
    println!("{}", a + 2);
    println!("{}", a + a);
    println!("{}", a * a);
    println!("{}", a / 5);
    println!("{}", a / -1);
    println!("{}", a << 127);
    assert_eq!(a as u64, 1);
    assert_eq!(a, 1u32 as i128);
    let a: u128 = 147573952589676412928; // 2^67
    println!("{}", a);
    assert_eq!(a, a);
}
```

Basically:

* Literals work. Literals that don't fit into u64 would be a bonus but I could see how they may not be required in the first implementation pass.
* All basic arithmetic operations work and produce expected results.
* The types have all common traits: Clone, Display, Debug, Default, Eq, Ord, etc
* Conversions to/from other integral types works
* The perf wins you'd *expect* to see are all visible, that is, u128 is much faster than `[u64;2]`

nikomatsakis:
I agree with @alexcrichton's comments; and of course we can also use the stability bar to gate if we feel there are flaws in the implementation, I imagine.

Amanieu:
I started working on an implementation of this, but progress has stalled since I'm going to be very busy finishing my thesis in the next few months. In any case the work didn't seem too complicated, it's just a matter of replacing most uses of u64 in the compiler with u128 and providing a emulated u128/i128 types for stage0.

nikomatsakis:

nikomatsakis:
Tracking issue: https://github.com/rust-lang/rust/issues/35118

PR #1520, opened 2016-03-01
RFC: Generic const and static

comex:
As it says on the tin.

[Rendered](https://github.com/comex/rfcs/blob/master/text/0000-generic-const-static.md)

### DIFF
--- a/text/0000-generic-const-static.md
+++ b/text/0000-generic-const-static.md
@@ -65,7 +65,7 @@ static DEFAULT<T>: T where T: ConstDefault = T::DEFAULT;
 
 - Both the type and the initializer can depend on generic parameters.  As usual, they are resolved and typechecked before generic substitution based on the constraints present.
 
-- Generic `static`s are monomorphized to a separate static allocation per specialization.  (They are guaranteed to have different addresses iff separately written static declarations are so guaranteed; I'm not sure if Rust plans to support coalescing identical truly-immutable statics.)
+- Generic `static`s are monomorphized on demand to a separate static allocation per specialization.  (They are guaranteed to have different addresses iff separately written static declarations are so guaranteed; I'm not sure if Rust plans to support coalescing identical truly-immutable statics.)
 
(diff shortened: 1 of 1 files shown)


arielb1:
I'm against generic `static`s because it is not obvious how they are monomorphized: currently, monomorphized items always have internal linkage, and monomorphizing them from 2 different crates creates 2 different copies. Doing this in a `static` could cause some confusion.

comex:
I didn't think of that, but AFAIK C++ has always had that issue (it supports both static variables in generic functions and classes, and as of C++14, generic variables - plus I believe it guarantees uniqueness for pointers to specialized functions), and solves it by marking those symbols as weak, so the dynamic linker picks one for the whole program.  Weak symbols have some performance impact, but for a small number of them it shouldn't matter.  The number could be further whittled down by detecting truly immutable objects and going back to internal linkage for those - after all, if Rust doesn't guarantee uniqueness for `fn` instantiations, that line has already been crossed, so there's little harm extending the same minor pitfall to variables.

aturon:
cc @eddyb @nikomatsakis

eddyb:
IIRC we never considered associated statics because maintaining uniqueness was deemed unreasonably difficult/impossible (cc @alexcrichto) and plans for generic consts were supersceded by `const fn`.

nikomatsakis:
We discussed this in the @rust-lang/lang meeting recently. I think that, at least for constants, we are all generally sympathetic to this goal, but we feel it's a bit premature to dive into it right now. The story with constants -- and especially constants and generics! -- is quite up in the air. We have const fns, associated constants, and the MIR transition, all of which affect how constants are implemented in the compiler and handled in the type system. Moreover, there is an obvious demand for constant-parameterized items (think `fn foo<const N: usize>()`). It seems most important to press on that last point, and MIR, and to some extent associated constants (which are entangled), and then return to this topic of generic constants.

(That said, I think one of the biggest points of tension here is between `const fn` and generic constants, since there seems to be quite some overlap between the two, though that is not to say we could not have both of them; perhaps they are both useful.)

nikomatsakis:
Postponed under https://github.com/rust-lang/rfcs/issues/1557

PR #1521, opened 2016-03-01
Copy clone semantics

ubsan:
Specifying that `<T as Clone>::clone(&t)` where `T: Copy` should be equivalent to `ptr::read(&t)`.

### DIFF
--- a/text/0000-copy-clone-semantics.md
+++ b/text/0000-copy-clone-semantics.md
@@ -1,0 +1,52 @@
+- Feature Name: N/A
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+With specialization on the way, we need to talk about the semantics of
+`<T as Clone>::clone() where T: Copy`.
+
+It's generally been an unspoken rule of Rust that a `clone` of a `Copy` type is
+equivalent to a `memcpy` of that type; however, that fact is not documented
+anywhere. This fact should be in the documentation for the `Clone` trait, just
+like the fact that `T: Eq` should implement `a == b == c == a` rules.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, `Vec::clone()` is implemented by creating a new `Vec`, and then
+cloning all of the elements from one into the other. This is slow in debug mode,
+and may not always be optimized (although it often will be). Specialization
+would allow us to simply `memcpy` the values from the old `Vec` to the new
+`Vec`. However, if we don't actually specify this, we will not be able to do
+this.
+
+# Detailed design
+[design]: #detailed-design
+
+Simply add something like the following sentence to the documentation for the
+`Clone` trait:
+
+"If `T: Copy`, `x: T`, and `y: &T`, then `let x = y.clone()` is equivalent to
+`let x = *y`;"
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is a breaking change, technically, although it breaks code that was
+malformed in the first place.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternative is that, for each type and function we would like to specialize
+in this way, we document this separately. This is how we started off with
+`clone_from_slice`.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What the exact sentence should be.
--- a/text/0000-copy-clone-semantics.md
+++ b/text/0000-copy-clone-semantics.md
@@ -28,7 +28,8 @@ be able to, and we will be stuck looping over every value.
 [design]: #detailed-design
 
 Specify that `<T as Clone>::clone(t)` shall be equivalent to `ptr::read(t)`
-where `T: Copy, t: &T`.
+where `T: Copy, t: &T`. An implementation that does not uphold this *shall not*
+result in undefined behavior; `Clone` is not an `unsafe trait`.
 
 Also add something like the following sentence to the documentation for the
 `Clone` trait:
@@ -52,4 +53,4 @@ in this way, we document this separately. This is how we started off with
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-What the exact sentence should be.
+What the exact wording should be.
--- a/text/0000-copy-clone-semantics.md
+++ b/text/0000-copy-clone-semantics.md
@@ -27,7 +27,10 @@ be able to, and we will be stuck looping over every value.
 # Detailed design
 [design]: #detailed-design
 
-Simply add something like the following sentence to the documentation for the
+Specify that `<T as Clone>::clone(t)` shall be equivalent to `ptr::read(t)`
+where `T: Copy, t: &T`.
+
+Also add something like the following sentence to the documentation for the
 `Clone` trait:
 
 "If `T: Copy`, `x: T`, and `y: &T`, then `let x = y.clone();` is equivalent to
--- a/text/0000-copy-clone-semantics.md
+++ b/text/0000-copy-clone-semantics.md
@@ -21,8 +21,8 @@ Currently, `Vec::clone()` is implemented by creating a new `Vec`, and then
 cloning all of the elements from one into the other. This is slow in debug mode,
 and may not always be optimized (although it often will be). Specialization
 would allow us to simply `memcpy` the values from the old `Vec` to the new
-`Vec`. However, if we don't actually specify this, we will not be able to do
-this.
+`Vec` in the case of `T: Copy`. However, if we don't specify this, we will not
+be able to, and we will be stuck looping over every value.
 
 # Detailed design
 [design]: #detailed-design
@@ -30,8 +30,8 @@ this.
 Simply add something like the following sentence to the documentation for the
 `Clone` trait:
 
-"If `T: Copy`, `x: T`, and `y: &T`, then `let x = y.clone()` is equivalent to
-`let x = *y`;"
+"If `T: Copy`, `x: T`, and `y: &T`, then `let x = y.clone();` is equivalent to
+`let x = *y;`. Manual implementations must be careful to uphold this."
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-copy-clone-semantics.md
+++ b/text/0000-copy-clone-semantics.md
@@ -24,6 +24,13 @@ would allow us to simply `memcpy` the values from the old `Vec` to the new
 `Vec` in the case of `T: Copy`. However, if we don't specify this, we will not
 be able to, and we will be stuck looping over every value.
 
+It's always been the intention that `Clone::clone == ptr::read for T: Copy`; see 
+[issue #23790][issue-copy]: "It really makes sense for `Clone` to be a
+supertrait of `Copy` -- `Copy` is a refinement of `Clone` where `memcpy`
+suffices, basically." This idea was also implicit in accepting
+[rfc #0839][rfc-extend] where "[B]ecause Copy: Clone, it would be backwards
+compatible to upgrade to Clone in the future if demand is high enough."
+
 # Detailed design
 [design]: #detailed-design
 
@@ -54,3 +61,6 @@ in this way, we document this separately. This is how we started off with
 [unresolved]: #unresolved-questions
 
 What the exact wording should be.
+
+[issue-copy]: https://github.com/rust-lang/rust/issues/23790
+[rfc-extend]: https://github.com/rust-lang/rfcs/blob/master/text/0839-embrace-extend-extinguish.md
--- a/text/0000-copy-clone-semantics.md
+++ b/text/0000-copy-clone-semantics.md
@@ -1,5 +1,5 @@
 - Feature Name: N/A
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Start Date: 01 March, 2016
 - RFC PR: (leave this empty)
 - Rust Issue: (leave this empty)
 


ubsan:
Note that these comments were originally line comments, but that fact seems to have been lost by github.

briansmith:
I originally thought this was a good idea, and even suggested something very much like this on the internals mailing list, but I've changed my mind. Every type that implements `Copy` and `Clone` should derive `Clone` and the derived implementation of `Clone` should be implemented in terms of `Copy`. Further, the optimizer should be able to notice that a hierarchy of copies is equivalent to a single flattened copy and optimize it to a memcpy. If the optimizer doesn't do that, then it is going to generate terrible code for lots of common cases that aren't calls to `Clone::clone` such as record updates that modify a single field.

ubsan:
@briansmith The issue isn't derived `Clone` implementations, it's that we allow you to not derive `Clone`. If your `T: Copy`, and you manually implement `Clone`, this would make it invalid to not make `Clone::clone(&t)` and `ptr::read(&t)` equivalent.

The optimizer is also not turned on in debug mode. Allowing us to turn `Vec::clone` into a `memcpy`, even in debug mode, would make for much faster code.

ubsan:
Note, we've already made breaking changes based on this assumption.

The change in `#[derive(Copy, Clone)]` to `fn clone(&self) { *self }` would break any code that relies on the fact that each of the inner `Clone::clone`s were called.

briansmith:
> it's that we allow you to not derive Clone

IMO, it is OK to say that if you want `Clone` to be fast for your `Copy` type, you need to derive `Clone`.

> The change in #[derive(Copy, Clone)] to fn clone(&self) -> Self { *self } would break any code that relies on the fact that each of the inner Clone::clones were called.

Yes, but IMO that was also a short-term mistake. Once the optimizer is doing the right thing, that shortcut won't be necessary and we can return to the simpler semantics.

In particular, the optimizer can already notice that a loop is equivalent to memcpy. It just needs to get smarter to recognize non-loop equivalents, including recursive invocations of cone/copy, are equivalent to memcpy are equivalent to memcpy too.

Consider `let x = y { u32_field: new_value }`. How would your proposed optimization improve this situation? It seems it wouldn't, but this kind of pattern is also important to optimize to memcpy + 32-bit memory write.

ubsan:
@briansmith This isn't an optimization. This is specification. It's saying that the standard library relies on the fact that a properly implemented `Clone::clone` for `T: Copy` == `ptr::read`. This *allows* for optimization through specialization, for example, the `#[derive(Copy, Clone)]` thing, and many other optimizations based around `memcpy`.

briansmith:
Yes, but I'm saying that if the optimizer is working well, then you don't need to specify anything special here. It would be useful to see an example where this rule is needed that can't be done without the rule by simply improving the optimizer.

Amanieu:
The RFC should specify what the consequences are for types that violate this rule. In particular this must not result in memory unsafety.

ubsan:
@briansmith The issue isn't where the optimizer is working. It's where it isn't. Optimization is not always on, like in debug mode. Also, building a language for a sufficiently smart compiler doesn't seem like the correct way of doing things. Often we will have to specifically specialize.

ubsan:
@Amanieu yes, that is the intention, I will write it down.

briansmith:
Why not specify that even in debug mode, the copy coalescing optimizations are done? Then there is no semantic change to the language needed and debug build results are faster.

ubsan:
@briansmith The semantics would still change, and then you would have different semantics between debug and release.

briansmith:
To be clear, there's nothing that would change semantically in my definition. The only thing that would change is that debug builds would be specified to have an optimization pass enabled, for at least `Clone::clone` that they currently don't have enabled, and that optimization pass would be made smarter. The language definition would stay exactly the same.

ubsan:
@briansmith Ah. You can't actually do that. The necessary optimizations to realize that:

```
assert!(src.len() == dst.len());
let len = dst.len();
let src = &src[..len];
for i in 0..len {
    self[i].clone_from(&src[i]);
}
```
is equivalent to 
```
assert!(src.len() == dst.len());
memcpy(src.as_ptr(), dst.as_mut_ptr(), dst.len());
```
is... most of them.

text/0000-copy-clone-semantics.md
```diff
@@ -0,0 +1,56 @@
...
+[alternatives]: #alternatives
+
+The alternative is that, for each type and function we would like to specialize
+in this way, we document this separately. This is how we started off with
+`clone_from_slice`.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What the exact wording should be.
```
petrochenkov: Alternative wording: "When `Copy` type has user-defined implementation of `Clone` it's unspecified whether `clone` or `memcpy` is called when cloning a value of this type." or (shamelessly stolen from somewhere) "An implementation is allowed to omit the call to `clone` even if is has side eﬀects. Such elision of calls to `clone` is called *clone elision*."

ubsan: @petrochenkov it's not that it's unspecified whether `clone` or `ptr::read` is called when you actually call `.clone()`. It's just that libraries (including the standard library) are able to assume that `Clone::clone` == `ptr::read` for `T: Copy`; and, if that requirement is not met, libraries are allowed to exhibit unexpected (but defined) behavior.

petrochenkov: > it's not that it's unspecified whether clone or ptr::read is called when you actually call .clone()

Why not? If we go this route anyway, let's give the compiler/libraries all the freedom.
I'd even write "user-defined implementations of `Clone` for `Copy` types are ignored", but I'm not sure about generics:
```
fn f<T: Clone>(arg: T) -> T { arg.clone() } // It's not known if `T` is `Copy` or not and I'm not sure it's convenient to turn `clone`s into `memcpy`s during monomorphisation.
```

petrochenkov: >It's just that libraries (including the standard library) are able to assume that Clone::clone == ptr::read for T: Copy

`derive` is not a not a library, so the language already can omit `clone`s as well.

petrochenkov: Also, "`clone` is always ignored" is better than "`clone` is sometimes ignored" from pedagogical point of view.

glaebhoerl:
Is there *any* reason to write a manual `Clone` impl for a type that's `Copy` in the first place, instead of deriving it? If we want to depend on the two being equivalent at the library level (as proposed), we should lint against it. (If we were to want to depend on it at the language level (as *not* proposed), then we should just make it an error outright.)

comex:
@glaebhoerl Unfortunately, derived impls on generic structs [sometimes have the wrong trait bounds](https://github.com/rust-lang/rust/issues/26925), so you have to implement them manually.

oli-obk:
@comex: isn't that "just a bug"?

ubsan:
Also, it's possible to implement manually, so it's possible to implement manually wrong. This RFC isn't about all those people who `#[derive(Copy, Clone)]`. It's about anyone who manually implements `Clone`, for whatever reason (including the reasons @comex  stated, or just reasons of style, or even really silly reasons).

retep998:
@ubsan I manually implement `Clone` in `winapi` to force the implementation to rely on `Copy` and thereby improve compile times.

ubsan:
Some people do have to implement manually, like @retep998.

text/0000-copy-clone-semantics.md
ubsan: hmm...

At first I thought this was a bad idea, but then I started thinking. Not "always ignored", but if your `clone` has different semantics from a `ptr::read` then it should be implementation defined whether it is called.

arielb1: I really don't like the "optimizer magic overriding of `clone`".

arielb1:
I am not sure how much teeth this RFC has.

We already have both `Clone::clone` and `Clone::clone_from`, which may be implemented separately (and allocate differently, even). I don't think that we ever intended to specify which of them is called by standard library functions.

ubsan:
@arielb1 Yes, because it's specified that `Clone::clone_from(x, y)` shall be equivalent to `*x = Clone::clone(y)`.

arielb1:
Right. I think the "equivalent in functionally" language fits this case quite well.

ubsan:
I think that the RFC as written follows what you're saying?

durka:
We did _not_ yet make the `#[derive(Copy, Clone)]` change cited above, this potential behavior change being one of the reasons. If this language goes in then that obstacle would presumably be removed. However, I have an idea for how to do that optimization better anyway, after specialization stabilizes :)

ubsan:
@durka Well, I take that argument back, then. However, it's not important. This still allows for nice optimizations and, importantly, makes sense. Nobody *should* be implementing `Clone::clone` != `ptr::read` when `T: Copy`, and not specifying that would be a very big burden on the language for optimization around generics.

text/0000-copy-clone-semantics.md
```diff
@@ -0,0 +1,56 @@
...
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+With specialization on the way, we need to talk about the semantics of
+`<T as Clone>::clone() where T: Copy`.
+
+It's generally been an unspoken rule of Rust that a `clone` of a `Copy` type is
```
nikomatsakis: You may want to cite https://github.com/rust-lang/rust/issues/23790, which initially proposed making `Clone` a supertrait of `Copy`, with the description:

> "Copy is a refinement of Clone where memcpy suffices"

The assumption that `Copy` and `Clone` would be compatible was also implicitly used in accepting [RFC 839](https://github.com/rust-lang/rfcs/blob/master/text/0839-embrace-extend-extinguish.md), as a premise thereof was the iea that it would be compatible to relax a `T: Copy` bound to a `T: Clone` bound later on.

In other words, it's been more than an unspoken rule that clone and copy must be compatible, even if the rule seems not to have made its way onto the documentation of the `Clone` trait itself.

nikomatsakis: (The example of specializing `#[derive(Copy, Clone)]` that was brought up elsewhere in the thread also seems relevant.)

nikomatsakis:
@briansmith 

> Yes, but IMO that was also a short-term mistake. Once the optimizer is doing the right thing, that shortcut won't be necessary and we can return to the simpler semantics.

Optimizers are great, but the whole point of specialization is to give the power to optimize into the hands of library authors -- this is particularly useful in cases where the compiler cannot necessarily see that two pieces of code are functionally equivalent. Clone and copy impls seem like an obvious example of such a thing (other examples might be more extended mathematical identities, or the use of extended functionality, such as rewriting a series of `read` calls into one `skip` call).

text/0000-copy-clone-semantics.md
ubsan: @arielb1 Yeah, I guess. I do think it would be better if it's just kept to libraries and language constructs like `#[derive]`

Kimundi:
I'm very much in favor of this RFC, but also think the wording can be relaxed a bit: 

- It would not be wrong to implement `Clone` manually without doing a copy operation.
- It would just be considered wrong to have different, or additional behavior in a `clone()`.
- It would always be legal to replace a `clone()` call with a copy operation if the type is `Copy` - whether it will be done so and to what extend and by what mechanism is a different question.

I'm not worried about issues of not knowing whether a `clone()` call and/or `Clone` bound will actually call the `clone()` method body, since if you could observe any difference at all, it would be due to a bug in your manual `Clone` impl.

ubsan:
@Kimundi I believe that a `(t: T).clone()` call should always be desugared to `<T as Clone>::clone(&t)` (assuming no implicit methods). `Clone` should not be specially treated by the compiler. It is only that, where `T: Clone + Copy`, `<T as Clone>::clone` may be assumed by libraries and by the `derive` mechanism to be equivalent to `ptr::read::<T>`. This is not a change in the definition of the language, only a clarification of the definition of the `Clone` trait.

text/0000-copy-clone-semantics.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently, `Vec::clone()` is implemented by creating a new `Vec`, and then
+cloning all of the elements from one into the other. This is slow in debug mode,
+and may not always be optimized (although it often will be). Specialization
+would allow us to simply `memcpy` the values from the old `Vec` to the new
+`Vec` in the case of `T: Copy`. However, if we don't specify this, we will not
+be able to, and we will be stuck looping over every value.
```
bluss: We will be stuck relying on the optimizer to figure it out. Which it can do and insert `memcpy`, typically for `Copy` values of 8 bytes or smaller, it looks like.

ubsan: @bluss exactly.

bluss:
What the memcpy equivalent exactly is, is not specified. A derived `Clone` never copies struct padding, while `Copy` and memcpy do, for example.

I think I agree with the intention of this RFC build I would like it to simply say two things:

- If a type implements `Copy`, it is always ok to replace general `Clone`s with copies.
- Code that relies on specific behavior of `Clone` while implementing `Copy` is buggy; the trait implementation are simply wrong, just like if you implemented `Ord` incorrectly.

ubsan:
@bluss `ptr::read` is specified in the document. It's not defined whether `memcpy` copies padding bytes.

Kimundi:
@ubsan My point was not that there _should_ be `Clone` specific compiler optimizations, but that there _could_ be. 

As in, if `Clone` is specified to not do more than `Copy`, then it would not matter if the compiler did this since there would be no observable behavior difference.

arielb1:
@kimundi

I see this as pretty similar to the situation with `PartialEq` and co. The compiler *will not* assume that your implementation of `PartialEq` is symmetric in any way. However, library functions might get confused.

alexcrichton:

e-oz:
Hate breaking changes. Have you checked how many cargo crates will be affected? If yes - please share the numbers. From discussion it looks like breaking change just to make debug mode faster, but I think it can't be true.

retep998:
@e-oz This won't cause anything to stop compiling. This merely makes an assumption that the standard library and compiler can optimize based on. If you can find any crates in the wild that would be affected negatively by this change, please do bring them up as there isn't really any way to find out in an automated fashion.

ticki:
I think it is very unlikely that such a change will break anything at all.

nikomatsakis:
(I also don't consider this a change, merely formally stating an existing assumption.)

e-oz:
Committed code contains words "this is a breaking change", that's why I asked. Sorry for being suspicious:)

aturon:
I'm in favor of documenting these expectations for the semantics of `Clone` -- clearly, any other behavior for `Copy` data would be incredibly surprising and likely to lead to bugs.

nikomatsakis:
**Huzzah!** The @rust-lang/lang team has decided **to accept** this RFC. However, since the libs team also claims jurisdiction, I'll let them merge it. :)

alexcrichton:
The libs team discussed this RFC at triage today and the decision was to **merge**. It was brought up that the user-facing documentation probably wants to indicate `*self` instead of `ptr::read` as it's a bit more understandable, but that doesn't need to change the body of this RFC.

Thanks again for the RFC @ubsan!

Tracking issue: https://github.com/rust-lang/rust/issues/33416

bluss:
Vec::clone specialization for T: Copy is realized since rust-lang/rust/pull/38182 (Through `extend_from_slice`).

PR #883, opened 2015-02-17
Make function pointer types look like borrowed pointer types for forwards compatability

Ericson2314:
[Rendered](https://github.com/Ericson2314/rfcs/blob/fun-vs-fun-ptr/text/0000-fun-vs-fun-ptr.md)

PR #1524, opened 2016-03-02
Custom Dynamically Sized Types for Rust

ubsan:
I believe this fixes #813, and is a nicer, and far more powerful, solution than #709.

### DIFF
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -223,8 +223,8 @@ slice-like types. However, this is not necessary for the actual RFC.
 
 ```rust
 #[lang = "unsize"]
-unsafe trait Unsize<T: Dst + ?Sized> {
-    const fn unsize(&self) -> &T;
+trait Unsize<T: Dst + ?Sized> {
+    const fn unsize(&self) -> T::Meta;
 }
 ```
 
@@ -247,12 +247,9 @@ struct Slice<T>[const N: usize]([T; N]);
 
 unsafe impl<T> Unsize<[T]> for [T; N] {
     const fn unsize(&self) -> &[T] {
-        unsafe {
-            std::ptr::make_fat_ptr(self, N)
-        }
+        N
     }
 }
-
 ```
 
 # Drawbacks
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -35,11 +35,11 @@ unsafe trait Dst {
 
 #[lang = "unsize"]
 unsafe trait Unsize<T: Dst + ?Sized> {
-    const fn unsize(&self) -> &Self::Output;
+    const fn unsize(&self) -> &T;
 }
 ```
 
-This trait should be unsafe, because it's very easy to give an incorrect
+The `Dst` trait should be unsafe, because it's very easy to give an incorrect
 implementation of `size_of_val`, and safe and unsafe code *must* rely on it
 being correct to not create undefined behavior.
 
@@ -47,6 +47,8 @@ Note that it is specified that, if `size_of::<<T as Dst>::Meta>() == 0`, then
 `size_of::<*const T>()` == `size_of::<*const ()>`. This is really important for
 C compatibility, because our current compatibility with flexible arrays is
 really terrible.
+
+The `Unsize` trait must be unsafe for much the same reasons.
 
 To actually get use of the improved fat pointers, we must add four new
 intrinsics (the names are up for bikeshedding):
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -60,8 +60,12 @@ intrinsics (the names are up for bikeshedding):
 mod ptr {
     pub use intrinsics::{fat_ptr_meta, make_fat_ptr, make_fat_ptr_mut};
 }
+mod mem {
+    pub use intrinsics::size_of_prelude;
+}
 mod intrinsics {
     extern "rust-intrinsic" {
+        fn size_of_prelude<T: ?Sized>() -> usize;
         fn fat_ptr_meta<T: ?Sized + Dst>(ptr: *const T) -> T::Meta;
         fn make_fat_ptr<T: ?Sized + Dst>(data: *const (), meta: T::Meta) -> *const T;
         fn make_fat_ptr_mut<T: ?Sized + Dst>(data: *mut (), meta: T::Meta) -> *mut T;
@@ -163,7 +167,7 @@ impl Dst for PascalStr {
 }
 unsafe impl Sizeable for PascalStr {
     fn size_of_val(&self) -> usize {
-        std::mem::size_of::<usize>() + self.len
+        std::mem::size_of_prelude::<Self>() + self.len
     }
 }
 
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -222,19 +222,6 @@ impl<const N: usize> PascalStr[N] {
 }
 ```
 
-Any type which is, currently, something like
-
-
-```rust
-struct CStr([u8]);
-// or
-struct Wrapper(u16, PascalStr);
-```
-
-will be treated just as they are today. A pointer to a `CStr` will be a 
-`{ptr, usize}`, and a pointer to a `Wrapper` will be a `{ptr, ()}` (although it
-will still be very difficult to create a `Wrapper`).
-
 One more type which is always brought up when one talks about DSTs is the trait
 object; this won't allow us to write trait objects in the standard library, or
 anything, but it is interesting to look at how the compiler will see trait
@@ -296,5 +283,5 @@ Index traits.
 
 How should these fat pointers be passed, if they are larger than two pointers?
 
-We likely need an equivalent to the C `sizeof(unsized)`; what should that
-function be?
+Should you be able to implement a `Dst` without an `Unsize` coercion? My guts on
+yes.
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -188,14 +188,19 @@ compatibility with C libraries that use flexible length arrays):
 
 ```rust
 #[repr(C)]
-struct PascalStr {
+struct PascalStr[const N: usize] {
     len: usize,
-    buffer: [u8],
+    buffer: [u8; N],
+}
+unsafe impl<const N: usize> Unsize<PascalStr> for PascalStr[N] {
+    const fn unsize(&self) -> &PascalStr {
+        unsafe {
+            std::ptr::make_fat_ptr(self, ())
+        }
+    }
 }
 impl Dst for PascalStr {
     type Meta = ();
-}
-unsafe impl Sizeable for PascalStr {
     fn size_of_val(&self) -> usize {
         std::mem::size_of_prelude::<Self>() + self.len
     }
@@ -205,16 +210,14 @@ impl PascalStr {
     pub fn len(&self) {
         self.len
     }
-    pub fn from_str(s: &str) -> Box<PascalStr> {
-        let backing_store = vec![0; std::mem::size_of::<usize>() + s.len()];
-        let ptr = Box::into_raw(backing_store.into_boxed_slice()) as *mut ();
-        let ps = Box::from_raw(std::ptr::make_fat_ptr_mut(ptr as *mut (), ()));
-        ps.len = s.len();
-        unsafe {
-            std::ptr::copy_nonoverlapping(s, ps.buffer.as_mut_ptr(), ps.len)
-        }
+}
 
-        ps
+impl<const N: usize> PascalStr[N] {
+    pub fn from_bstr(s: &[u8; N]) -> PascalStr[N] {
+        PascalStr {
+            len: N,
+            buf: *s,
+        }
     }
 }
 ```
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -1,5 +1,5 @@
 - Feature Name: custom\_dst
-- Start Date: 01 March, 2016
+- Start Date: 02 March, 2016
 - RFC PR: (leave this empty)
 - Rust Issue: (leave this empty)
 
@@ -20,37 +20,33 @@ pixel buffers. In many places, something other than the Rust slice type is
 needed (if you have both a width and a height, or if you have a width, height,
 and stride, or even further).
 
-This doesn't make it very *nice* to work with custom DSTs; it makes it possible,
-and makes sure there's a strong base so that's there's a lot of room to grow.
-
 # Detailed design
 [design]: #detailed-design
 
-The most important addition to the language shall be two new traits:
+The most important addition to the language shall be one new trait, a
+modification of the Unsize trait, and a bit of new syntax (introduced later):
 
 ```rust
 #[lang = "dst"]
-trait Dst {
+unsafe trait Dst {
     type Meta: Copy;
-}
-#[lang = "sizeable"]
-pub unsafe trait Sizeable {
     fn size_of_val(&self) -> usize;
+}
+
+#[lang = "unsize"]
+unsafe trait Unsize<T: Dst + ?Sized> {
+    const fn unsize(&self) -> &Self::Output;
 }
 ```
 
-Sizeable should be unsafe, because it's very easy to give an incorrect
-implementation, and safe and unsafe code *must* rely on it being correct to not
-create undefined behavior.
+This trait should be unsafe, because it's very easy to give an incorrect
+implementation of `size_of_val`, and safe and unsafe code *must* rely on it
+being correct to not create undefined behavior.
 
 Note that it is specified that, if `size_of::<<T as Dst>::Meta>() == 0`, then
 `size_of::<*const T>()` == `size_of::<*const ()>`. This is really important for
 C compatibility, because our current compatibility with flexible arrays is
 really terrible.
-
-`Sizeable` *must* be implemented for all types, and shall be automatically
-implemented for any type which does not implement `Dst`, and will be defined
-as the `size_of_val` of each of the members.
 
 To actually get use of the improved fat pointers, we must add four new
 intrinsics (the names are up for bikeshedding):
@@ -67,22 +63,35 @@ mod intrinsics {
     extern "rust-intrinsic" {
         fn size_of_prelude<T: ?Sized>() -> usize;
         fn fat_ptr_meta<T: ?Sized + Dst>(ptr: *const T) -> T::Meta;
-        fn make_fat_ptr<T: ?Sized + Dst>(data: *const (), meta: T::Meta) -> *const T;
-        fn make_fat_ptr_mut<T: ?Sized + Dst>(data: *mut (), meta: T::Meta) -> *mut T;
+        fn make_fat_ptr<U: Dst + ?Sized, T: Unsize<U>>(data: *const T,
+            meta: U::::Meta) -> *const U;
+        fn make_fat_ptr_mut<U: Dst + ?Sized, T: Unsize<U>>(data: *mut T,
+            meta: U::::Meta) -> *mut U;
     }
 }
 ```
 
-The following is an example implementation of `[T]`:
+These shall, very importantly, be closely tied to integer generics, and will not
+be viable until integer generics are in the language. Fortunately, integer
+generics are coming soon™.
 
 ```rust
-// The last field of a type that implements Dst must be a 0 sized array
+// The [const N] means that the N parameter is optional, and if left off,
+// unsizing coercion shall be done
 #[lang = "slice"]
-struct Slice<T>([T]);
-impl<T> Dst for [T] {
-    type Meta = usize;
+struct Slice<T>[const N: usize]([T; N]);
+// The last member of a struct like this must be Unsize
+
+unsafe impl<T, const N: usize> Unsize<[T]> for [T; N] {
+    const fn unsize(&self) -> &[T] {
+        unsafe {
+            std::ptr::make_fat_ptr(self, N)
+        }
+    }
 }
-unsafe impl<T> Sizeable for [T] {
+
+unsafe impl<T> Dst for Slice<T> {
+    type Meta = usize;
     fn size_of_val(&self) -> usize {
         if self.len() > 0 {
             core::mem::size_of_val(&self[0]) * self.len()
@@ -92,6 +101,7 @@ unsafe impl<T> Sizeable for [T] {
     }
 }
 
+// impl<T> Slice<T>
 impl<T> [T] {
     pub fn len(&self) -> usize {
         unsafe {
@@ -102,12 +112,14 @@ impl<T> [T] {
     pub fn as_ptr(&self) -> *const T {
         unsafe {
             &self.0 as *const [T] as *const T
+            // importantly, you still have access to the last member of the
+            // struct; it is the Output of Unsize on the last member
         }
     }
 
     pub unsafe fn from_raw_parts(buf: *const T, len: usize) -> &[T] {
         unsafe {
-            &*core::intrinsics::make_fat_ptr(buf as *const (), len)
+            &*core::intrinsics::make_fat_ptr(buf as *const [T; 1], len)
         }
     }
 }
@@ -125,28 +137,29 @@ And the following, an example implementation of a pixel buffer; this is the real
 meat of the RFC, custom fat pointers for libraries:
 
 ```rust
-struct PixelBuffer([f32]);
+struct PixelBuffer[const W: usize, const H: usize]([f32; W * H]);
+
+unsafe impl<const W: usize, const H: usize> Unsize<PixelBuffer> for PixelBuffer[W, H] {
+    const fn unsize(&self) -> &PixelBuffer {
+        unsafe {
+            std::ptr::make_fat_ptr(self, (W, H))
+        }
+    }
+}
 impl Dst for PixelBuffer {
     type Meta = (usize, usize);
-}
-unsafe impl Sizeable for PixelBuffer {
     fn size_of_val(&self) -> usize {
         std::mem::size_of::<f32>() * self.width() * self.height()
     }
 }
-impl PixelBuffer {
-    pub fn new_zeroed(width: usize, height: usize) -> Box<PixelBuffer> {
-        if height != 0 {
-            assert!(usize::max_value() / height > width);
-        }
-        let backing_store = vec![0.0; height * width];
-        let ptr = Box::into_raw(backing_store.into_boxed_slice()) as *mut ();
-        unsafe {
-            Box::from_raw(
-                std::ptr::make_fat_ptr_mut(ptr, (width, height)))
-        }
-    }
 
+impl<const W: usize, const H: usize> PixelBuffer[W, H] {
+    const fn from_array(arr: [[f32; W]; H]) -> Self {
+        Self(arr)
+    }
+}
+
+impl PixelBuffer {
     pub fn width(&self) -> usize {
         unsafe {
             std::ptr::fat_ptr_meta(self).0
@@ -159,6 +172,13 @@ impl PixelBuffer {
         }
     }
 }
+
+// ---
+
+fn main() {
+    let pixels: Box<PixelBuffer> = Box::new(PixelBuffer::from_array([[0.0; 1280]; 960);
+}
+
 ```
 
 And now, an example implementation of a Pascal string type (this is good for
@@ -211,20 +231,36 @@ will be treated just as they are today. A pointer to a `CStr` will be a
 will still be very difficult to create a `Wrapper`).
 
 One more type which is always brought up when one talks about DSTs is the trait
-object. Unfortunately, this change won't help trait objects that much; it will
-allow us to get rid of `std::repr` for them, but most of it is compiler magic.
-The implementation of these DST traits should looks something like the
-following, however (unfortunately, inside the compiler because we can't impl
-across all trait objects):
+object; this won't allow us to write trait objects in the standard library, or
+anything, but it is interesting to look at how the compiler will see trait
+objects in the context of wider DSTs:
 
 ```rust
-// trait is a separate type
-impl Dst for Trait { 
-    type Meta = TraitVtable;
+trait Trait {
+    fn func1();
 }
-unsafe impl Sizeable for Trait {
+struct TraitVtable {
+    size: usize,
+    func1: fn(),
+}
+unsafe impl<T: Trait> Unsize<Trait> for T {
+    const fn unsize(&self) -> &Trait {
+        static vt = TraitVtable {
+            size: size_of::<T>(),
+            func1: Self::func1
+        };
+        unsafe {
+            std::ptr::make_fat_ptr(self, vt)
+        }
+    }
+}
+
+unsafe impl Dst for Trait { 
+    type Meta = TraitVtable;
     fn size_of_val(&self) -> {
-        // intrinsic
+        unsafe {
+            std::ptr::fat_ptr_meta(self).size
+        }
     }
 }
 ```
@@ -241,9 +277,9 @@ More complication in the language.
 which unfortunately doesn't give us things that can't be represented with
 existentials, like a rectangular window into one of those `PixelBuffer`s.
 
-My original idea had two types in `Dst`, Meta, and Data, where Data was similar
-to the `[T]` at the end of the struct. This isn't nice for types like
-`PascalStr`.
+I originally had an idea where the DSTs were the first class types, because I
+was afraid of putting integer generics into my proposal. The current proposal
+is, in my opinion, far nicer.
 
 You could define a struct type
 `PixelBuffer<'a> { width: usize, height: usize, ptr: &'a f32 }`, and a similar
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -76,7 +76,7 @@ mod intrinsics {
 The following is an example implementation of `[T]`:
 
 ```rust
-// The last field of a type that implements Dst must be a slice
+// The last field of a type that implements Dst must be a 0 sized array
 #[lang = "slice"]
 struct Slice<T>([T]);
 impl<T> Dst for [T] {
@@ -108,6 +108,14 @@ impl<T> [T] {
     pub unsafe fn from_raw_parts(buf: *const T, len: usize) -> &[T] {
         unsafe {
             &*core::intrinsics::make_fat_ptr(buf as *const (), len)
+        }
+    }
+}
+
+impl<T: Drop> Drop for [T] {
+    fn drop(&mut self) {
+        for el in self {
+            drop_in_place(el)
         }
     }
 }
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -1,0 +1,247 @@
+- Feature Name: custom\_dst
+- Start Date: 01 March, 2016
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow Rust code to define custom fat pointers, and define slice code in terms of
+these new operators (instead of the current `std::raw::Repr`).
+
+# Motivation
+[motivation]: #motivation
+
+One major missing feature of Rust is custom dynamically sized types; it's
+impossible to implement a slice type in Rust, for example. This has led to
+things like the `std::raw::Repr` trait, a stopgap measure. It's also very
+difficult to represent things like views into 2d slices; for example,
+pixel buffers. In many places, something other than the Rust slice type is
+needed (if you have both a width and a height, or if you have a width, height,
+and stride, or even further).
+
+This doesn't make it very *nice* to work with custom DSTs; it makes it possible,
+and makes sure there's a strong base so that's there's a lot of room to grow.
+
+# Detailed design
+[design]: #detailed-design
+
+The most important addition to the language shall be two new traits:
+
+```rust
+#[lang = "dst"]
+trait Dst {
+    type Meta: Copy;
+}
+#[lang = "sizeable"]
+pub unsafe trait Sizeable {
+    fn size_of_val(&self) -> usize;
+}
+```
+
+Sizeable should be unsafe, because it's very easy to give an incorrect
+implementation, and safe and unsafe code *must* rely on it being correct to not
+create undefined behavior.
+
+Note that it is specified that, if `size_of::<<T as Dst>::Meta>() == 0`, then
+`size_of::<*const T>()` == `size_of::<*const ()>`. This is really important for
+C compatibility, because our current compatibility with flexible arrays is
+really terrible.
+
+`Sizeable` *must* be implemented for all types, and shall be automatically
+implemented for any type which does not implement `Dst`, and will be defined
+as the `size_of_val` of each of the members.
+
+To actually get use of the improved fat pointers, we must add four new
+intrinsics (the names are up for bikeshedding):
+
+```rust
+// core
+mod ptr {
+    pub use intrinsics::{fat_ptr_meta, make_fat_ptr, make_fat_ptr_mut};
+}
+mod intrinsics {
+    extern "rust-intrinsic" {
+        fn fat_ptr_meta<T: ?Sized + Dst>(ptr: *const T) -> T::Meta;
+        fn make_fat_ptr<T: ?Sized + Dst>(data: *const (), meta: T::Meta) -> *const T;
+        fn make_fat_ptr_mut<T: ?Sized + Dst>(data: *mut (), meta: T::Meta) -> *mut T;
+    }
+}
+```
+
+The following is an example implementation of `[T]`:
+
+```rust
+// The last field of a type that implements Dst must be a slice
+#[lang = "slice"]
+struct Slice<T>([T]);
+impl<T> Dst for [T] {
+    type Meta = usize;
+}
+unsafe impl<T> Sizeable for [T] {
+    fn size_of_val(&self) -> usize {
+        if self.len() > 0 {
+            core::mem::size_of_val(&self[0]) * self.len()
+        } else {
+            0
+        }
+    }
+}
+
+impl<T> [T] {
+    pub fn len(&self) -> usize {
+        unsafe {
+            core::intrinsics::fat_ptr_meta(self)
+        }
+    }
+
+    pub fn as_ptr(&self) -> *const T {
+        unsafe {
+            &self.0 as *const [T] as *const T
+        }
+    }
+
+    pub unsafe fn from_raw_parts(buf: *const T, len: usize) -> &[T] {
+        unsafe {
+            &*core::intrinsics::make_fat_ptr(buf as *const (), len)
+        }
+    }
+}
+```
+
+And the following, an example implementation of a pixel buffer; this is the real
+meat of the RFC, custom fat pointers for libraries:
+
+```rust
+struct PixelBuffer([f32]);
+impl Dst for PixelBuffer {
+    type Meta = (usize, usize);
+}
+unsafe impl Sizeable for PixelBuffer {
+    fn size_of_val(&self) -> usize {
+        std::mem::size_of::<f32>() * self.width() * self.height()
+    }
+}
+impl PixelBuffer {
+    pub fn new_zeroed(width: usize, height: usize) -> Box<PixelBuffer> {
+        if height != 0 {
+            assert!(usize::max_value() / height > width);
+        }
+        let backing_store = vec![0.0; height * width];
+        let ptr = Box::into_raw(backing_store.into_boxed_slice()) as *mut ();
+        unsafe {
+            Box::from_raw(
+                std::ptr::make_fat_ptr_mut(ptr, (width, height)))
+        }
+    }
+
+    pub fn width(&self) -> usize {
+        unsafe {
+            std::ptr::fat_ptr_meta(self).0
+        }
+    }
+
+    pub fn height(&self) -> usize {
+        unsafe {
+            std::ptr::fat_ptr_meta(self).1
+        }
+    }
+}
+```
+
+And now, an example implementation of a Pascal string type (this is good for
+compatibility with C libraries that use flexible length arrays):
+
+```rust
+#[repr(C)]
+struct PascalStr {
+    len: usize,
+    buffer: [u8],
+}
+impl Dst for PascalStr {
+    type Meta = ();
+}
+unsafe impl Sizeable for PascalStr {
+    fn size_of_val(&self) -> usize {
+        std::mem::size_of::<usize>() + self.len
+    }
+}
+
+impl PascalStr {
+    pub fn len(&self) {
+        self.len
+    }
+    pub fn from_str(s: &str) -> Box<PascalStr> {
+        let backing_store = vec![0; std::mem::size_of::<usize>() + s.len()];
+        let ptr = Box::into_raw(backing_store.into_boxed_slice()) as *mut ();
+        let ps = Box::from_raw(std::ptr::make_fat_ptr_mut(ptr as *mut (), ()));
+        ps.len = s.len();
+        unsafe {
+            std::ptr::copy_nonoverlapping(s, ps.buffer.as_mut_ptr(), ps.len)
+        }
+
+        ps
+    }
+}
+```
+
+Any type which is, currently, something like
+
+
+```rust
+struct CStr([u8]);
+// or
+struct Wrapper(u16, PascalStr);
+```
+
+will be treated just as they are today. A pointer to a `CStr` will be a 
+`{ptr, usize}`, and a pointer to a `Wrapper` will be a `{ptr, ()}` (although it
+will still be very difficult to create a `Wrapper`).
+
+One more type which is always brought up when one talks about DSTs is the trait
+object. Unfortunately, this change won't help trait objects that much; it will
+allow us to get rid of `std::repr` for them, but most of it is compiler magic.
+The implementation of these DST traits should looks something like the
+following, however (unfortunately, inside the compiler because we can't impl
+across all trait objects):
+
+```rust
+// trait is a separate type
+impl Dst for Trait { 
+    type Meta = TraitVtable;
+}
+unsafe impl Sizeable for Trait {
+    fn size_of_val(&self) -> {
+        // intrinsic
+    }
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+More complication in the language.
+
+# Alternatives
+[alternatives]: #alternatives
+
+@eddyb's original idea of existentials `struct [T] = for<N: usize> [T; N];`,
+which unfortunately doesn't give us things that can't be represented with
+existentials, like a rectangular window into one of those `PixelBuffer`s.
+
+My original idea had two types in `Dst`, Meta, and Data, where Data was similar
+to the `[T]` at the end of the struct. This isn't nice for types like
+`PascalStr`.
+
+You could define a struct type
+`PixelBuffer<'a> { width: usize, height: usize, ptr: &'a f32 }`, and a similar
+PixelBufferMut struct, but this is *very* annoying, and cannot be returned by
+Index traits.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How should these fat pointers be passed, if they are larger than two pointers?
+
+We likely need an equivalent to the C `sizeof(unsized)`; what should that
+function be?
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -56,7 +56,11 @@ mod ptr {
     }
 }
 mod mem {
-    pub use intrinsics::size_of_prelude;
+    pub fn size_of_prelude<T: ?Sized>() -> usize {
+        unsafe {
+            core::intrinsics::size_of_prelude::<T>()
+        }
+    }
 }
 mod intrinsics {
     extern "rust-intrinsic" {
@@ -276,6 +280,3 @@ Index traits.
 [unresolved]: #unresolved-questions
 
 How should these fat pointers be passed, if they are larger than two pointers?
-
-Should you be able to implement a `Dst` without an `Unsize` coercion? My guts on
-yes.
--- a/text/0000-custom-dst.md
+++ b/text/0000-custom-dst.md
@@ -23,19 +23,13 @@ and stride, or even further).
 # Detailed design
 [design]: #detailed-design
 
-The most important addition to the language shall be one new trait, a
-modification of the Unsize trait, and a bit of new syntax (introduced later):
+The most important addition to the language shall be one new trait:
 
 ```rust
 #[lang = "dst"]
 unsafe trait Dst {
     type Meta: Copy;
     fn size_of_val(&self) -> usize;
-}
-
-#[lang = "unsize"]
-unsafe trait Unsize<T: Dst + ?Sized> {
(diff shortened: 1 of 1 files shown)


dgrunwald:
`&CStr` was originally supposed to be compatible with `*mut c_char` for FFI purposes. (with `CStr` being an [unsized type](https://github.com/rust-lang/rfcs/issues/813))
It's only a fat pointer because of Rust's current limitations regarding DSTs.

ubsan:
@dgrunwald Ah, well, this RFC does support that.

```rust
struct CStr([libc::c_char]);
impl Dst for CStr {
    type Meta = ();
}
impl Sizeable for CStr {
    fn size_of_val(&self) -> usize {
        libc::strlen(self.as_ptr())
    }
}
impl CStr {
    pub unsafe fn from_ptr(p: *const libc::c_char) -> &CStr {
        // either
        std::mem::transmute::<*const libc::c_char, &CStr>(p)
        // or
        std::ptr::make_fat_ptr(p as *const (), ())
    }
}

**Comment by Aatch** [comment]
I'm not sure about this. My main concern is around how actually useful this is compared to the complexity. There is a lot of code that assumes fat pointers are two words and two words only. This would completely change that. It's a very large increase in complexity.

Another issue is how this interacts with generics and unsizing. Currently `Foo<[T; N]>` will coerce into `Foo<[T]>` and everything will work properly because we know how to do that. There's also the case of people trying to use a generic `Meta` type, it's unclear how that might work.

Right now, my thought is that it's a nice feature (like most that get proposed), but not worth the increased complexity.

**Comment by ubsan** [comment]
@Aatch We need something like this. Look at the documentation for `CStr`:

> Note: This operation is intended to be a 0-cost cast but it is currently implemented with an up-front calculation of the length of the string.

This is just in the standard library. Out there in real code, we have to deal with C interop, where stuff like 

```C
typedef struct _TOKEN_GROUPS {
  DWORD              GroupCount;
  SID_AND_ATTRIBUTES Groups[ANYSIZE_ARRAY];
} TOKEN_GROUPS, *PTOKEN_GROUPS;
```

has to be translated into

```rust
struct TokenGroups {
    group_count: u32,
    groups: [SID_AND_ATTRIBUTES; 0],
}
```

And traits like `Index` and `IndexMut` only return references, so what if you have a 3D world which wants to return 2D slices into the world? You can't. You just can't. You have to define getters which return a

```rust
struct 2DSlice<'a, T> {
    width: usize,
    height: usize,
    ptr: *const T,
    lifetime: PhantomData<&'a [T]>,
}
struct 2DSliceMut<'a, T> {
    width: usize,
    height: usize,
    ptr: *mut T,
    lifetime: PhantomData<&'a mut [T]>,
}
```

ubsan:
@Aatch Unsizing isn't touched in this RFC. That's a magic trait that has to have integer generics to actually do stuff with.

ubsan:
@Aatch A generic `Meta` type isn't a problem, because of this error: "the type parameter `T` is not constrained by the impl trait, self type, or predicates".

reem:
@ubsan `struct Whatever<T> { inner: Vec<T>> } impl<T> Whatever<T> Dst { type Meta = T; }`?

ubsan:
@reem 1) That's not an unsized type, 2) that wouldn't actually be a problem because you know what Meta is.

bluss:
@japaric Didn't you work on user-defined DSTs before? Do you have a link to your old effort?

thepowersgang:
How would indexing the inner `[T]` of a custom DST be handled? It doesn't have the element count metadata from a normal slice. Would it be an unsafe operation? Or would the normal `impl [T]` methods be removed/hidden by the compiler when the slice is accessed via a custom DST pointer.

Idea: Add `fn element_count(&self) -> usize` to the `Dst` trait

ubsan:
@thepowersgang It's an unsafe operation. The inner `[T]` isn't "real"; it's only there to get the base pointer of the unsized data.

eddyb:
I originally proposed using `[T; 0]` for anything slice-like because it has the right alignment and can serve as a base pointer.

@Aatch Fat pointers being hardcoded to one pointer-sized piece of metadata is a mistake. There are important extensions to slice and traits we need to support and that hardcoding has to go (which is easier with the MIR).
cc @nikomatsakis about his desire to have `TraitA+TraitB` work as a trait object with 2 vtable pointers.

ubsan:
I liked `[T]` because it seemed nicer; this should be up for bikeshedding.

japaric:
> Didn't you work on user-defined DSTs before? Do you have a link to your old effort? 

Yes, I was working on it last year. I don't have time right now to analyze/comment on the design proposed in this RFC. So I'm just going to drop a few links here:

- [My draft RFC](https://github.com/japaric/rfcs/blob/unsized2/text/0000-unsized-types.md)
- [My old implementation](https://github.com/japaric/rust/blob/unsized/src/libcore/marker.rs#L56-L73)
- [An example of what you can do with this feature: slices/views on dense matrices](https://github.com/japaric/rusty-edge/tree/master/unsized_types/dense)

Aatch:
@thepowersgang actually brings up a good point, if you can get at the `[T]`, then you'd expect to be able to work with it like a regular `[T]`. I'd prefer a way of indicating "this is unsized" that doesn't use `[T]` in that case.

Either use a `[T;0]` or something else that isn't already a DST. I don't think this really needs a marker for the "unsized portion" anyway, since a more trait-like DST would probably want to use `()`. The trait itself should be enough. It might be possible to treat an implementation of `Dst` as an implied `impl !Sized`, which would handle most of the issues I can think of.

ubsan:
Hmm, yeah, I do think [T; 0] is better now that it's been discussed.

ubsan:
Although, I'm thinking about how it would work with integer generics and Unsizing coercions... I think that `[T]` is the best way, because we would have some sort of way of saying "This PascalStr is a Sized PascalStr (which can be unsized)", or whatever, which would look something like

```rust
struct PascalStr {
    len: usize,
    buf: [u8],
}
// &PascalStr == ptr -> { len, [buf; len] }
// &PascalStr<N> == ptr -> { N, [buf; N] }
// PascalStr<N> == { N, [buf; N] }
```

Aatch:
@ubsan eh, I'm not sure that's really worth it here. My main concern is that the compiler now has to be aware of a lot more context when handling field access. The main advantage of `[T; 0]` (or a semantically-similar marker type) is that you can't safely access any data through a pointer to it, but it has all the relevant properties of `T`. I'm not sure I like the idea of `&foo.bar` being invalid just because `foo` implements `Dst`.

Unsizing is probably better handed via separate types. This is already the case, `[T; N]` might look similar to `[T]`, but they're different types completely. It's more obvious when you consider trait object unsizing. I think the same is going to be true here, if you want unsizing, you're gonna have to have a separate `Sized` type like everybody else.

ubsan:
@Aatch but they're not completely different types. You need them to be exactly the same type, in fact, behind the pointer, for unsizing coercions to work.

Aatch:
@ubsan no, they're the same representation, not the same type.

eddyb:
```rust
struct PascalStrBuffer<B> {
    len: usize,
    buf: B
}
type PascalStrArray<const N: usize> = PascalStrBuffer<[u8; N]>;
type PascalStr = PascalStrBuffer<[u8]>;
```
This might be an okay compromise of sorts.

I've also considered having a single optional integer parameter, which denotes the minimal contained size and defaults to `0` (which would mean `[u8; 0]`, i.e. "any size").
The problem with that, though, is you lose the sized aspect for which you wanted the flexibility in the first place, so it's probably pointless.

ubsan:
@Aatch fine; you can argue whether they're the same type or not. They are the same representation, definitely, and if you had to write separate Sized versions of each type, imagine the really big types:

```rust
struct CLAIM_SECURITY_ATTRIBUTE_RELATIVE_V1 {
    Name: DWORD,
    ValueType: WORD,
    Reserved: WORD,
    Flags: DWORD,
    ValueCount: DWORD,
    union Values {
        pInt64: [DWORD],
        pUint64: [DWORD],
        ppString: [DWORD],
        pFqbn: [DWORD],
        pOctetString: [DWORD],
    }
}
```

imagine writing two of these types, and now for each of the Windows flexible array structs. It gets ridiculous.

eddyb:
It would be almost impossible to ensure the same layout for two different definitions though.
And I'm not sure how well `#[repr(C)]` would work with the DST stuff, *although* I reckon that it seems useful to have C interop defined for non-trait DSTs *anyway*.

Aatch:
In that case how about we restrict this to slice-like DSTs and use @thepowersgang's suggestion of having an `element_count` method?

ubsan:
@Aatch that seems overly restrictive. For example, if you have a

```rust
struct 2dSlice<T>([T]);
impl<T> Dst for 2dSlice<T> {
    type Meta = (usize, usize);
    fn element_count(&self) -> usize {
         // (meta.0 * meta.1)
    }
}
impl<T> Sizeable for 2dSlice<T> {...}

fn main() { 
    let x = 2dSlice::<(), (std::usize::MAX, std::usize::MAX)>::new(); // not possible with `element_count`
}

**Comment by Aatch** [comment]
@ubsan that's not possible anyway! You can't have something that large. And it's not like the same reasoning doesn't apply to the `size_of_val` case.

Oh, and something I just thought of: destructors/dropping. It's all well and good having something that only contains Copy` types, but what if you have something non-Copy in your custom DST?

**Comment by ubsan** [comment]
@Aatch yes you can :) look at the type inside!

You only have a pointer to the values, and aren't the owner. You could `impl Drop` to drop the inside values, and that would work for something like `Box`.

**Comment by Aatch** [comment]
@ubsan "not the owner"? Then who is? `[T]` owns the `T`s it contains, sure, if you have a `&[T]` you don't own the `[T]` itself, but that's not the same.

Do I now have to implement`explicitly implement `Drop` to get the correct behaviour for my custom DST? I mean, fine, if that's the case, that's the case, but it should be in the RFC.

So far everything has been focused on custom strings, N-dimensional arrays of `Copy` types and N-dimensional slices from... somewhere. I'd like to see this actually all worked out, you may have your use-cases, but you need to look at how this *can* be used, not just how you would use it.

**Comment by ubsan** [comment]
@Aatch right, will look at it.

**Comment by ubsan** [comment]
I've been thinking about this further, and this really requires Sized types to go along with the !Sized types in order to be nice. I was far too scared about integer generics in my initial implementation, and so I did not include them. I think this is far closer to the correct implementation (although it's a bit farther off from being implemented). @eddyb has told me that they are working on integer generics, and this solves most problems (although a manual implementation of `Drop` is still required).

**Comment by gnzlbg** [comment]
Is there an implementation of this somewhere?

**Comment by ubsan** [comment]
@gnzlbg No. It's blocked on integer generics.

**Comment by arielb1** [comment]
@ubsan

Why are integer generics needed? Currently `[T]` is completely distinct from `[T; n]`. We could keep the compiler-generated `Unsize` impl for them. Therefore, I don't understand why there is punning between `SliceBuffer[W,H]` and `SliceBuffer` the DST - these types should be distinct.

You need `transmute` (or dependent types) anyway if you want to create DSTs with a dynamic length. If we want a nice way to do it, it should apply to both static and dynamic DSTs.

I also don't really understand why `Unsize` is an unsafe trait - isn't it basically an operator trait? Is it because of the `&`-`&mut` deduplication?

**Comment by ubsan** [comment]
@arielb1 It doesn't require integer generics to work. It requires them for on-stack DSTs, and I believe that those are important. Otherwise, you need to create them manually, on the heap. The current implementation works without integer generics, I just think that having unsizeable types like this is a really nice way of allowing this pattern.

I'm not sure why Unsize is an unsafe trait. I assume I had a reason at some point.

**Comment by ubsan** [comment]
This did kinda evolve into two RFCs. I will devolve it.

**Comment by ubsan** [comment]
@arielb1 The reason that they're not separate is that for a struct like `CLAIM_SECURITY_ATTRIBUTE_RELATIVE_V1`, the person implementing the struct doesn't have to make certain that the unsized version and the sized version will be the same representation.

**Comment by chris-morgan** [comment]
@eddyb @Aatch The main reason hard-coding of fat pointer sizes exists is because of the lack of generic integers—if you’re needing to work with fat pointers, you have no alternative but to assume that they are the same length as `[usize; 2]`. Sensible practitioners put in a `debug_assert!` to that effect, verifying the length. If we can have a proper way of *knowing* the size of the fat pointers, great! We’ll fix our libraries with their unsafe code that we knew from the start was painful to behold and replace it with a proper approach. This sort of RFC is about improving the state of things so that we don’t *need* to make such assumptions—that we had to is the only reason we have.

**Comment by eddyb** [comment]
@chris-morgan I definitely agree, although I was originally thinking of the compiler, which has deeply rooted assumptions that are harder to fix.

File: text/0000-custom-dst.md [hunk 1]
```diff
@@ -0,0 +1,279 @@
...
+# Detailed design
+[design]: #detailed-design
+
+The most important addition to the language shall be one new trait:
+
+```rust
+#[lang = "dst"]
+unsafe trait Dst {
+    type Meta: Copy;
+    fn size_of_val(&self) -> usize;
```
pnkfelix: please add at least a sketches of comments describing what these associated items are supposed to do.

Here is an example of what I am thinking of (just to illustrate that the descriptions can be succinct).

```rust
/// The `Meta` for a type `T` is extra information attached to any pointer or reference to a `T`.
/// In other words, a non-zero sized `T::Meta` means that `&T` is represented as a so-called "fat pointer."
type Meta: Copy;

/// Returns the size, in bytes, of the `T` according to the `Meta` value attached to `self`.
fn size_of_val(&self) -> usize;
```

(I grok that the rest of the RFC is sort of filling in this gap via examples, but as a reader it is a bit frustrating to have no specification, and then have the following paragraph warn about how error-prone implementing this trait is.)

text/0000-custom-dst.md
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+}
+mod intrinsics {
+    extern "rust-intrinsic" {
+        fn size_of_prelude<T: ?Sized>() -> usize;
+        fn fat_ptr_meta<T: ?Sized + Dst>(ptr: *const T) -> T::Meta;
+        fn make_fat_ptr<T: Dst + ?Sized>(data: *const (),
+            meta: U::::Meta) -> *const T;
+        fn make_fat_ptr_mut<T: Dst + ?Sized>(data: *mut (),
```
pnkfelix: can you add a note explaining why two distinct intrinsics that differ only in whether they return `*mut T` or `*const T` is necessary?

(It seems like one could easily make one a wrapper around the other? Is there some issue I'm missing with low-level interaction with the compiler backend?)

text/0000-custom-dst.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+And the following, an example implementation of a pixel buffer; this is the real
+meat of the RFC, custom fat pointers for libraries:
+
+```rust
+struct PixelBuffer([f32]);
+
+unsafe impl Dst for PixelBuffer {
+    type Meta = (usize, usize);
```
pnkfelix: nit: is this supposed to be a fully-general two-dimensional slice, allowing arbitrary subrectangles as subslices?  If so, don't you need a stride in a addition to the `width` and `height` in the `Meta`?

Ericson2314:
I'd like `DST` to be an bounded-by-default trait like `Sized`, and furthermore `Sized: DST` if we can simultaneously do `impl<T: Sized> DST for T { .. }` with that. Truly unsized types should be quantified with `T: ?DST`, and custom dynamically sized can be created with `unsafe impl !Sized for MyDST { ..}`.

Oh and I think `DynSized` is a better name than `DST`. The "T" is redundant for a trait (of course it describes a type), and `DS` is too short and obscure.

text/0000-custom-dst.md
ubsan: I'm not sure why I never replied to this. Sorry, will do.

text/0000-custom-dst.md
ubsan: Yeah, you probably could. It's all design. I personally just thought it felt more "rusty".

text/0000-custom-dst.md
ubsan: Yes, you're right.

eddyb:
@Ericson2314 I would prefer to stick with "unsized" being synonymous to "dynamically sized", use `Sizeable` for the trait holding `{align,size}_of_val` and maybe then call types that don't even implement `Sizeable` "unsizeable".
Except, that could be confused for "can be unsized" which is not the intention. Whereas "truly unsized" doesn't seem different from "unsized" to me. *sigh* words

Ericson2314:
Notice here I didn't use "unsized" for anything :). The choices as would be written are `Sized`, `?Sized`, and `?DynSize`. We are still free to "unsized" for whatever we want.

[In the other thread I wrote "truely unsized" just because that is what the issue was named.]

Ericson2314:
A question is whether we need a third trait for the `meta`. On one hand, for types which are neither `Sized` nor merely `DynSized`, the meta cannot contribute anything. On the other hand it's probably saner to say "pointers to `T`  can only be created if `T` implements this trait", than say "pointers to `T` are different based one which traits `T` implements".

Maybe specialization could someday make the latter more normal, but I don't believe we specialize (associated) types just yet.

Ericson2314:
Oh, and if we have a third trait it should definitely be called `Refer` :).

Ericson2314:
@ubsan I still like my `DynSized` :) but connecting this directly to `Sized` is definitely a step in the right direction.

bugaevc:
This is probably not the best place to discuss an alternative proposal, but nevertheless.

When I think about custom DSTs, I imagine something like this:
```rust
// This declares Slice<T> as a DST.
// It can't be used directly, thus talking about
// the contents or representation of the type itself
// makes no sence
// Instead, we specify what references/pointers to Slice<T> are:
#[lang = "slice"]
struct &Slice<T> {
    ptr: *mut T,
    len: usize,
}
```
[Some more mock-up code](https://is.gd/r2SJ8U).

Can someone explain why this is a bad idea and why we need the complexity of this RFC?

eddyb:
@bugaevc Any pointers to DSTs must be castable to a thin pointer (think of `impl Drop for Box`) and thus have to guarantee there's a pointer field.
But even then, `Rc<RefCell<Slice<T>>>` *must* work yet you didn't specify the layout of that, not even of `Rc<Slice<T>>` but only of pointers to `Slice<T>` directly.
Keeping the pointer and the metadata separate is needed to get that orthogonality.

bugaevc:
@eddyb Thanks, that makes sense. So let's require all DSTs to implement `.size_of_val()` and `.ptr_of_val()`, preferably `const fn`s, like in the RFC. A struct containing a DST field then would just offset the field's result by a constant, a reference to the struct being just a wrapper for a reference to the field.

Admittedly, this makes it more complex; but the complexity is in the compiler, a person implementing a new DST wouldn't need to worry.

AtheMathmo:
Can I ask what is holding up this RFC? Or maybe what people's biggest concerns are with this?

It looks like it would add some really valuable functionality (that I think many people could take advantage of).

nikomatsakis:
@AtheMathmo To be perfectly honest, I don't think many of us on the @rust-lang/lang team have read it. This is both because of general lack of time but also because we've been pretty focused on closing out existing priorities and features aimed at growing adoption. That said, I think there is a general recognition that a lot of good ideas (like, maybe, this one) have been sitting around without progress, so we've been making an effort to try to go through older RFCs, catch up on the conversation, and try to give our two cents. I know that this one has been on my shortlist. 

(Along those lines, we've also been thinking about ways to revise the RFC process to help focus feedback early, make conversations more structured, and so forth. See @aturon's [excellent blog post](https://aturon.github.io/blog/2016/07/05/rfc-refinement/) for some summary of the ideas and links to threads where we've been hashing those thoughts out further.)

nikomatsakis:
OK, so, I read the RFC. :) I think this is pretty nifty stuff, and I imagine it could be a good addition to the language. I'm going to give some scattershot responses. The TL;DR: I think the ideas are pretty solid. I'd like to mix this up with japaric's and see if we can't make the code feel a bit more elegant, and I'd like to elaborate out some more examples, but overall, I like it.

----

First, I want to think a bit about the motivation. There are a couple of problems in this general area of "working with unsized types" and I want to be thoughtful about which ones we are trying to solve and which ones we are not. Here are some example problems around DST that come to mind, starting from the one that the RFC clearly aims at and progressing more towards unlikely:

- Enable more complex notions of slices, notably two- or multi-dimensional slicing. So e.g. being able to make a "submatrix" of a matrix. This RFC seems to solve that problem pretty well. I like the way we keep the existing `Index` traits but basically allow them to return additional "info" for their fat pointers.
    - That said, it's worth digging a bit more into this I think. Do we aim to support N-dimensional slices, where the N is a dynamic value? If so, I guess that the info would have to have a `Vec<T>` for the slice boundaries, which isn't `Copy`, and hence wouldn't work with this trait.
- What about other kinds of unsized types? For example, trying to model something more like trait objects? Obviously we can't do the full method dispatch magic, but trait objects have other properties (e.g., a dynamically computed alignment) that `[T]` don't have, so it's worth thinking them over.
- Creating a kind of "newtype'd slice", such as `Path` and `PathBuf`, requires some tricky unsafe code right now. It seems like this is a kind of "simple case" and it might be nice if it could be done w/ safe code. It seems clear though that this RFC doesn't address that problem (and maybe it shouldn't).
    - Would the mechanisms offered in this RFC be a nicer way to solve that problem?
- It has been proposed that could try to support "compound" unsized types, such as `[Trait]` (which would be an unsized version of `[T; N]` where `T: Trait`) or `[[T]]` (which would be a two-dimensional slice). Supporting types like these seems slick but I'm not sure it really works out; for example, the existential quantifier doesn't nest in a trivial way (e.g., `[T]` is `exists N. [T;N]` but `[[T]]` is `exists N. exists M. [[T; M]; N]`, not `exists N. [exists M. [T; M]; N]`). This seems to suggest that there will be problems around substitution. I think the RFC does not (and righly so) aim to address this.
- Other possible applications?

----

OK, next thing, I had a few nits and questions about the text itself. For example, in the final example of a pixel buffer, I think there are some errors/inconsistencies:

```rust
impl Index<Range<usize>> for Pixels {
  type Output = PixelBuffer; // <-- I think this should be Pixels?
  fn index(&self, idx: Range<usize>) -> &Pixels {
     ... // various assertions 
     unsafe {
       Pixels::from_raw_parts(
         self.as_ptr().offset(index.start as isize), // <-- where did this `as_ptr` method come from?
         index.end - index.start, // <-- also, s/index/idx/?
         self.stride, // <-- this field `stride` is declared on the auxiliary, perhaps `self.stride()`?
         self.height) // <-- same
     }
  }
}
```

I'm not trying to nitpick, just to make sure I understand. In particular, in order to access the meta data, you can't just use field access like `self.stride` (I assume), but rather have to do `ptr::fat_ptr_meta(self).stride`, right? And similarly the `as_ptr` method isn't something you would get automatically, but rather something you would write yourself (much like `from_raw_parts`), right?

---

OK, third part. If there is a flaw in this RFC, I think it might be that it's not quite ambitious enough. I don't necessarily mean in the *motivation* (though if we could solve the second problem, making it easy to make newtype'd slices, that'd be great) but the mechanism. Normally I'm kind of against new syntax, especially for wizardry :), but I feel like the code resulting from this proposal feels like we could make it more elegant.

Interestingly, it seems like @japaric and I were thinking along similar lines. When I came back and read his [older RFC](https://github.com/japaric/rfcs/blob/unsized2/text/0000-unsized-types.md), after reading this one, I found it had a lot of the initial ideas I had been tossing around.

Let me try and elaborate a bit on what didn't feel great to me. Starting point is the definition of `Pixels` itself:

```rust
struct Pixels([f32; 0])
```

This declaration *looks* deceptively normal, but of course it's kind of misleading. The "zero-sized field at the end" has many desirable properties, but it seems to make assumptions about our representation and layout that we don't normally guarantee (for example, one might assume you can take the address of this field to get at the extra data, but that isn't necessarily true since we could opt to reorder fields, right?). I guess I would prefer some kind of more *declarative* syntax for what is happening here, or perhaps a bit more magic.

I sort of liked how @japaric's RFC handled it. Here is a kind of "from memory, hybridized" variant:

```rust
unsized type Pixels; // basically a placeholder
impl Unsized for Pixels {
    type Data = f32; // what the (sized) pointer will point at
    type Meta = (usize, usize);
    
    fn size_of_val(&self) -> usize { size_of::<f32>() * meta(self).0 * meta(self).1 }
    fn align_of_val(&self) -> usize { align_of::<f32>() } // I think we need this too =)
}
```

(Oh, btw, I think that `impl !Sized` feels really weird to me, particularly if we add some more general form of `!` notation. I think I prefer just an `Unsized` lang-item-trait.)

Anyway, as I wrote initially, I'm largely bikeshedding here. I do think we can make it "feel" slicker, but I don't think it's a big change to the core ideas involved.

Ericson2314:
@nikomatsakis thinking long term, I think we want to decouple the DST part and existential part of trait objects (think @thepowersgang's https://github.com/thepowersgang/stack_dst-rs). In hypothetical notation:
```rust
type TraitObject<Tr: trait> = exists T: ?Sized. TraitObjectInner<T, Tr>;
unsized type TraitObjectInner<T: ?Sized, Tr: trait>;
impl<T: ?Sized, Tr: trait> DynSized for TraitObject<Tr, T> {
    type Meta = Dict<Tr, T>;
    fn size_of_val(&self) -> usize { Dict::size(meta(self)) }
    fn size_of_val(&self) -> usize { Dict::align(meta(self)) }
}
```

ubsan:
@nikomatsakis 

So, an answer to your points:

 * I don't think we do want to support N-dimensional slices. We really do want `&T` to be copyable, which would not work well with a `Data` that isn't `Copy`.
* Trying to model something like trait objects is more difficult, but is supported.

```rust
// NOTE: the reason we have the [u64; 0] is for alignment. I'm not sure it's necessary...
// Let's make some changes. How I'd do it today.
trait Foo {
    fn foo(&self);
}

impl Foo for i32 {
    fn foo(&self) {
        println!("foo called on type i32: {}", self);
    }
}

// Now how you'd implement a Foo trait object with this (with some changes)

struct FooVt {
    __size: usize,
    __align: usize,
    foo: fn(_: *const ()),
}

struct FooT;
unsafe impl !Sized for FooT {
    type Data = ();
    type Meta = FooVt;
    fn size_of_val(&self) -> usize { self.size }
    fn align_of_val(&self) -> usize { self.alignment }
}

impl FooT {
    fn new_ref<T: Foo>(t: &T) -> &FooT {
        &*ptr::make_fat_ptr(
            t as *const T as *const (),
            FooVt { 
                __size: mem::size_of::<T>(),
                __align: mem::align_of::<T>(),
                foo: mem::transmute::<fn(&T), fn(*const ())>(T::foo)
            }
        ) 
    }
}
```

* There *are* nice ways to solve newtyped slices, with this, but it requires more design, as well as integer generics, imho
* It actually originally did solve `[[T]]`. I did pull it down a bit. You can look through the history for the solution. This can definitely solve that.

The RFC text probably needs to be rewritten, it was written and rewritten a long time ago, and I was not a good writer then :)

The zero sized field at the end was intended to deal with alignment, and I agree, I'd rather do `Data`. That example looks exactly like my early example :P, except for the `unsized type Pixels`; one important thing is that unsized types can have associated data behind the pointer.

Ericson2314:
> The zero sized field at the end was intended to deal with alignment, and I agree, I'd rather do Data. That example looks exactly like my early example :P, except for the unsized type Pixels; one important thing is that unsized types can have associated data behind the pointer.

What's the point of `Data`? Not all DSTs are container-ish things, and as far as I can tell it's mainly used for the default definition of `align_of_value`. I want to resurrect https://github.com/rust-lang/rfcs/pull/883 with this, but machine code procedures have no sensible `Data`.

gnzlbg:
@nikomatsakis 

> That said, it's worth digging a bit more into this I think. Do we aim to support N-dimensional slices, where the N is a dynamic value? If so, I guess that the info would have to have a Vec<T> for the slice boundaries, which isn't Copy, and hence wouldn't work with this trait.

With type-level integers that could be an array [T; N] for the boundaries. 

My experience tells me that, in practice, to slice an R-rank tensor, you need 2x [T; R] for the begin and end indices. That covers "some" use cases. To cover "most" use cases you also need a [T; R] for the strides (lots of use cases have a stride of 1, but some important use cases have other strides). That basically covers 99% of the slicing an R-rank tensor use cases. One can think of more complicated strides (e.g. following a geometric progression) but I in my domain those aren't common (they might be incredibly common in other domains though).

nikomatsakis:
@ubsan 

> I don't think we do want to support N-dimensional slices. We really do want &T to be copyable, which would not work well with a Data that isn't Copy.

A very good point. I'm convinced.

> There are nice ways to solve newtyped slices, with this, but it requires more design, as well as integer generics, imho

It is probably a separate RFC. I don't think we need integer generics, though. It seems like the minimum thing that is needed is to be able take a `&[T]` and get a `&Foo<T>` where `struct Foo<T>([T])` without doing a transmute. But I'm not really sure the best way to do that. =) 

Perhaps just allowing `p as &Foo<T>` would be ok (respecting the privacy of the field). That is, modeling it as a kind of coercion. (Or, if we adopted your idea of having specialized casting intrinsics, which I prefer, then adding one of those.)

-----

@Ericson2314 

> What's the point of Data?

Well, if you have a `&Foo` where `Foo` is an unsized type, it tells you what you get when you dereference the `Foo` (or at least a prefix thereof). It's true that you can casting and coercion to do everything, but why force that on people in cases where there is a simpler option?

-----

@gnzlbg 

> With type-level integers that could be an array [T; N] for the boundaries. 

Hmm, good point, I am comforted to think that with constant generics we might be able to scale up to higher -- but still compile-time constant -- dimensions.

Ericson2314:
> Well, if you have a `&Foo` where `Foo` is an unsized type, it tells you what you get when you dereference the `Foo` (or at least a prefix thereof). It's true that you can casting and coercion to do everything, but why force that on people in cases where there is a simpler option?

This just sounds to narrow-minded for me. Even if primary usecase of this is array-like things, I want this to be the foundation of all DSTs. You mention "casting and coercion", does this have something to do with our current Unsize trait that I am missing?

eddyb:
@Ericson2314 You simply *cannot* avoid supporting a prefix, because, e.g. `RefCell<Trait>` (or the heap part of `Rc<T>`) already makes good use of it - the mechanism involved is *actually* in stable Rust.

nikomatsakis:
@Ericson2314 

> This just sounds narrow-minded for me. Even if primary usecase of this is array-like things, I want this to be the foundation of all DSTs. 

I don't follow you at all, I have to admit. Perhaps we view the significance of `Data` differently. I am assuming that a `Data` type is mostly a convenience: if you construct a fat pointer, it tells you what type the "main pointer" should have, but it's not required to be a complete type, just a prefix. So for example you could just give `*mut ()` -- an empty prefix -- which would mean that the contents are completely unknown statically (this would be suitable for a trait object). But sometimes you can say more, and it seems useful to be able to do so.

ubsan:
@eddyb wanted to see a C++ vtable implementation:

```rust
trait Foo {
    fn foo(&self);
}

impl Foo for i32 {
    fn foo(&self) {
        println!("foo called on type i32: {}", self);
    }
}

// Now how you'd implement a Foo trait object with this

struct FooVt {
    __ptr: *const (),
    __size: usize,
    __align: usize,
    foo: fn(_: *const ()),
}

pub struct FooT;
unsafe impl !Sized for FooT {
    type Data = FooVt;
    type Meta = ();
    fn size_of_val(&self) -> usize { self.vtable().__size }
    fn align_of_val(&self) -> usize { self.vtable().__align }
}

impl FooT {
    pub fn new_ref<T: Foo>(t: &T) -> &FooT {
        &*ptr::make_fat_ptr(
            t as *const T as *const (),
            FooVt { 
                __size: mem::size_of::<T>(),
                __align: mem::align_of::<T>(),
                foo: mem::transmute::<fn(&T), fn(*const ())>(T::foo)
            }
        ) 
    }

    fn vtable(&self) -> &FooVt {
        unsafe { &*(self as *const Self as *const FooVt) }
    }
}

impl Foo for FooT {
    fn foo(&self) {
        (self.vtable().foo)(self.vtable.__ptr)
    }
}
```

ubsan:
@eddyb wanted us to show off a C++-style "thin trait" thing.

```rust
trait Foo {
    fn foo(&self);
}

// I'm sad that this can't be called "Foo"
struct FooT {
    __align: usize,
    __size: usize,
    foo: fn(*const ()),
}

impl !Sized for FooT {
    type Data = ();
    type Meta = ();

    fn size_of_val(&self) -> usize { self.__size }
    fn align_of_val(&self) -> usize { self.__align }
}

struct Bar {
    __foo_vt: FooVt,
    first: u32,
    second: u32,    
}

impl Foo for Bar {
    fn foo(&self) {
         println!("called Foo::foo with Bar {{ first: {}, second: {} }}", self.first, self.second);
    }
}

impl Foo {
    fn downcast(&self) -> &FooT {
        unsafe {
            &*std::ptr::make_fat_ptr(self as *const _ as *const (), ())
        }
    }
}
```

japaric:
> unsized type Pixels;

If we go this route, this RFC should specify how the variance of these unsized types is determined. For example in this case:

``` rust
enum ColumnMajor {}
enum RowMajor {}

// `Order` is a phantom type that can be `RowMajor` or `ColumnMajor`
unsized type Matrix<T, Order>;

impl<T, Order> Unsized for Matrix<T, Order> {
    type Data = T;
    type Meta = (u32, u32); // rows x columns

    // ...
}
```

How `Order` affects the variance of `Matrix`. Is it required to have a `PhantomData<Order>` field in the `Meta` information of the unsized type? etc.

ubsan:
@japaric I wasn't planning on going that way...

ubsan:
The thin-pointer vtable example under this syntax would be as follows:

```rust
pub trait Foo {
    fn foo(&self);
}

struct FooVt {
    __align: usize,
    __size: usize,
    foo: fn(*const ()),
}

pub struct FooT {
    __foo_vt: &'static FooVt,
    unsized ()
}

impl !Sized for FooT {
    type Meta = ();
    fn size_of_val(&self) -> usize { self.__foo_vt.__size }
    fn align_of_val(&self) -> usize { self.__foo_vt.__align }
}

pub struct Bar {
    __foo_vt: FooVt,
    pub first: u32,
    pub second: u32,    
}

impl Foo for Bar {
    fn foo(&self) {
         println!("called Foo::foo with Bar {{ first: {}, second: {} }}", self.first, self.second);
    }
}

static __BAR_FOO: FooVt = FooVt {
    __align: std::mem::align_of::<Bar>(),
    __size: std::mem::size_of::<Bar>(),
    foo: Bar::foo,
};

impl Foo {
    pub fn new(first: u32, second: u32) -> Foo {
        Foo {
            __foo_vt: __BAR_FOO,
            first: first,
            second: second,
        }
    }

    pub fn downcast(&self) -> &FooT {
        unsafe {
            &*std::ptr::make_fat_ptr(self as *const _ as *const (), ())
        }
    }
}
```

nikomatsakis:
@japaric 

> If we go this route, this RFC should specify how the variance of these unsized types is determined.

Yes, a solid point. I suspect the rule would be that the variance is determined by the position of said parameters in `Data` -- and all parameters must be used. 

One question is whether having "extra syntax" like `unsized type` here is all that helpful, as well. Somehow I sort of feel like it is, though. I feel weird about the fact that an impl changes the "character" of a type so drastically -- an "inherently unsized type" like `[T]`, `Trait`, or `Path` feels very different to me.

Similar to @ubsan's thoughts, another syntax I considered was to use `..` inside a struct declaration to create an "unsized" type. The idea would be that the fields in the struct constitute a "prefix". Perhaps we could also have you give a name and type to the `..` so you can take the address of it and get the start of the "unsized" part of data:

```rust
struct Slice<T> {
     ..data: T // maybe?
}
```

eddyb:
@nikomatsakis At that point, what's the advantage over `data: [T; 0]`?

nikomatsakis:
@eddyb to me the improvement is mostly that the struct declaration, seen in isolation, is not misleading. If I see `[f64; 0]`, I have to find the `impl` to understand what is going on with this struct.

nikomatsakis:
Question: what are some use cases for custom DST where the fat pointer is not a matrix? (Sorry if there have been many listed and I forgot; in that case please send me a link :)

eddyb:
@nikomatsakis One example is [thin pointers to C++-like objects](https://github.com/rust-lang/rfcs/pull/1524#issuecomment-240922698) (vtable pointer at the start of the object).
You can also represent any trait object, including those with multiple vtable pointers (but the built-in ones would only benefit from this kind of abstraction via reflection, a library can't do something for all traits).

nikomatsakis:
@eddyb ah, ok, yes, the thin-ptr vtable is another example, though not -- I think -- the way I would imagine supporting such a feature. That is, I still think we would want something nice like `#[repr(thin)]` that one can apply to an existing trait as an optimization. =) 

It is nice to think about having ways to expose vtables (in general) as structs full of pointers and tie them into this metadata though, so that one can work with trait objects that way. (And we probably ought to write an rfc for general `Trait1+Trait2` types at some point.)

Still, those examples don't feel totally compelling to me: not because they don't work, but because it feels like something that we'd ultimately want to built into the trait system or expose with some other desugarings on top.

nikomatsakis:
So @nrc points out that this RFC would interact with unsized coercions. Has there been some discussion about how to handle this? Currently, the way unsized coercions basically work is that the compiler has some builtin rules about things it knows how to coerce, and you impl a trait in order to authorize it to use those rules. It seems like custom DST would require either that the trait be extended with some user-defined code or else that unsized type defines a way to synthesize its metadata from other inputs, right?

eddyb:
@nikomatsakis Then if we ignore vtables, slices with length embedded, before the data, would also be able to be thin pointers instead of fat ones.
This can be taken further, e.g., `CStr` could have an `O(n)` `size_of_val` that literally calls `strlen`.

nikomatsakis:
@eddyb 

> Then if we ignore vtables, slices with length embedded, before the data, would also be able to be thin pointers instead of fat ones.

That doesn't work, unless you can't take sub-slices. (If you do `&x[1..]`, then you need to remember both the offset (`1`) and the data pointer.)

>  This can be taken further, e.g., CStr could have an O(n) size_of_val that literally calls strlen.

This seems interesting to pursue though. (And you could do some amount of sub-slicing, in theory.)

Basically I'd like to push a bit more on using this design for different patterns to see what kinds of hazards come up. I was thinking the other day (for example) whether it might ever make sense to have auxiliary data that includes a region, and what kinds of restrictions we need in such a case (seems like it could, and we would need the regions to be constrained by the "unsized type").

ubsan:
@nikomatsakis No, you can't subslice such an object.

A real CStr is definitely a thing which I'd like to have.

ubsan:
ping @nikomatsakis 

status?

nikomatsakis:
### On priorities

So @ubsan and I had a good chat about this RFC the other day. Let me lay out my personal perspective. I think that the core ideas of this RFC are really solid. I also think this RFC pretty clearly doesn't fit well into the roadmap priorities. What I would like to see happen is something where we can accept this RFC on a highly provisional basis -- basically, for experimentation in tree, since there is no way to experiment on this out of tree. We don't really have a good process for experimentation right now. We have accepted some RFCs in this kind of "very tentative" fashion (e.g., naked fns), but it never feels like a 'fully proper' thing to me. I have some thoughts on this point in particular (trying to adjust the process to make a different pathway for "explicitly experimental"), but I think this PR isn't the proper place for them. I'll try to write up something to internals. So, let's presume we work that out, and decide to accept this RFC on some kind of provisional basis. (I'm honestly not sure what the rest of @rust-lang/lang thinks about this whole obsession of mine.)

One thing I worry about is that even reviewing and landing PRs related to this RFC might be a lot of work. The changes to the compiler will cause churn, merge conficts, and accidental regressions, all of which take a toll. But I'm not sure how much in practice.

### On the RFC itself

But speaking more to the RFC itself: fundamentally, I think that the motivation is pretty strong. As @ubsan pointed out to me, the `Deref` and `Index` traits bake in references into the trait itself. In some ways, this seems like a good thing to me. It ensures that we can syntactically recognize lvalues like `*foo` and `foo[10]` -- in contrast, if those sometimes generated wholly new values that had ownership (e.g. `IndexGet`, `DerefGet`), we'd have to introduce temporaries to take ownership, which would affect the lifetimes, and you'd only know this by evaluating the various traits. So, I find the idea of custom DST to be an appealing middle road here. It preserves the fact that `Deref`/`Index` always returns a borrow, while allowing for enough flexibility to enable a lot of useful use cases that don't work today:

- subslices of multidimensional arrays;
- modeling C and Pascal strings;
- are there others? I feel confident people will come up with some. =)

The RFC's design itself seems about right. Whatever we do, it's going to look something like this: We need a way to declare an unsized type that combines "data" and "meta" parts. The meta part must be copy so that `&T: Copy` continues to hold. We need some intrinsics or lang items or something for breaking apart and reassembling fat pointers. I see three overall areas of uncertainty here:

1. How hard will this be to implement? Will that reveal horrible flaws?
1. Will we be able to write matrix libraries? Model CStr? Perhaps do some other clever things?
1. What syntax should we use (more on this later).

The first two seem like things that will be hard to answer without implementing this and giving it a spin. Given that this is not really part of the roadmap, I think this would have to be something that is done done by motivated people in the community. It's a good way to learn about the Rust compiler. The role of the teams would ideally be a mentorship role. @nrc can tell you how awesome DST is as a way to learn everything you might want to know about trans. :P

I also expect that the RFC will likely change shape over the course of implementation. That's fine and expected.

That said, there are some particulars I'd still like to change in the RFC. To some extent these are "surface" syntactic issues, but I think they also get at the mental model people should have. I think we should move away from an `Unsized` or `!Sized` trait, which seems to pertain only to unsized types, and instead think of a `Referent` trait, which pertains to all types. It defines what it means for a type to be a reference. This would also move away from the idea of "fat and thin" pointers, and instead say that all references have a "data" and a "meta" part. But for "thin" pointers (i.e., reference to sized types) the `Meta` is always `()`.

Here is the trait I propose, roughly. You will notice a certain resemblance to the `!Sized` trait:

```rust
pub trait Referent {
    type Data;
    type Meta: Copy;

    /// Compute size/alignment of the data behind a reference.
    fn size_of_val(&self) -> usize; 
    fn align_of_val(&self) -> usize;
    
    /// Make a reference from its constituent parts.
    fn assemble(data: *const Self::Data, meta: Self::Meta) -> *const Self;

    /// Break a reference down into its constituent parts.
    fn disassemble(data: *const Self) -> (*const Self, Self::Meta);

    /// Convenience functions.
    fn data(r: &Self) -> &Self::Data { /* defined in terms of disassemble */ }
    fn meta(r: &Self) -> Self::Meta { let (_, meta) = Self::disassemble(r); meta }
}
```

This trait would be deeply built into the compiler, so it wouldn't be something you can implement by hand. For all existing types, the compiler would provide the implementation. For illustrative purposes, you might imagine that the impl for all sized types would look something like this:

```rust
impl<T> Referent for T {
    type Data = T;
    type Meta: Copy = ();

    fn size_of_val(&self) -> usize { size_of::<T>() }
    fn align_of_val(&self) -> usize { align_of::<T>() }
}
```

Ok, so how do we declare custom DST? The current RFC sort of "puns", using a single struct to represent both the unsized type *and* a prefix of the data that it points at, and pairing that with an impl:

```rust
// Current RFC
struct Pixels([f32; 0]);
struct CStr([u8; 0]);
impl !Sized for CStr { ... }
```

The unfortunate part of this pun is that it means that a raw pointer to the unsized type is a fat pointer. So, e.g., `*mut PascalStr` would be a fat pointer (admittedly, one with a `()` metadata). We now have no good way to refer to *just* the data pointer, devoid of its metadata. The only way we can do it is with a `*const ()`, effectively. 

In contrast, if we declare the unsized type explicitly, and then connect it to a separate data/meta type, we have a name for each constituent part. Here I'm going to go wild using some syntax. The idea is that you declare an `unsized type` and, together with it, you declare some parts of the `Referent` trait:

```rust
/// Declare an unsized type and some of its parameters.
/// All of these members are required.
unsized type Pixels {
    type Data = f32; // data pointer is a `*const f32`
    type Meta = [usize; 2]; // meta is two `usize` values
    fn size_of_val(&self) -> usize { .. }
    fn align_of_val(&self) -> usize { .. }
}
```

Here, the `Data` field basically tells us what type the data has. It is really a convenience for users, to help them write more type-safe code. It might describe some prefix of the data, or, as with `Pixels`, it describes the type of a repeating segment (but not the bounds).

### Concerns

Finally, some of my concerns:

- If we accept this, and implement it, will it just linger in a quasi-supported, perma-unstable state? (This applies equally well to many other experiments, e.g. naked fns)
    - A counter-pressure here might come from libraries like ndarray, if they decided they could make good use of this feature.
- What about unsize coercions? How do they fit in here? Perhaps they just come naturally from deref coercions?
- Does all the complexity and mechanism here pay for itself? I don't think you can really argue that this makes language *simpler*, even if it does expose some previously baked-in parts to customization.
- Even reviewing and landing PRs related to this RFC might be a lot of work, which might delay other things (as I mentioned earlier).
- We might find we want some more general descriptive power. This RFC cannot (by itself) describe thin trait objects, for example, they would need some reflective mechanism. That's probably an orthogonal concern.

Ericson2314:
Cross-posting from https://internals.rust-lang.org/t/recent-change-to-make-exhaustiveness-and-uninhabited-types-play-nicer-together/4602/56?u=ericson2314

Earlier in this thread I mentioned a `DynSize` trait to be used in a `Sized: DynSized: Referent` hierarchy. The point of this is some DSTs have unknown size at runtime too. The classic example of this are:

1. Opaque structs in C. These are, quite operational, unknown types of unknown size.

2. The `void` in C's `void*`. This is *not* the void for zero arguments or an () return! This is in fact exactly like the above, except for C's coercion rules, which make it act like a supertype of many types.

3. Functions *themselves* (see https://github.com/rust-lang/rfcs/pull/883). Then we can use `Option<&'static fn-itself-type>`, `*const fun-itself type`, and `&'a fn-itself-type` for weak linking and run-time (not init-time), dynamic linking, and JIT compilers.

ubsan:
I can no longer keep this open in good faith. Someone else can take it from me, but ... personal issues.

nikomatsakis:
So, for some time now I've been wanting to write a followup to my previous comment sharing my thoughts about this work. Even though @ubsan has closed the RFC, I figured I should leave it anyway. Hopefully it is clear from my previous comments that I think the ideas in here are good in a technical sense, and I'd like to do some experimentation with them. However, it's probably also clear that I feel somewhat hestitant about this. My concerns here are not technical, but rather all about **prioritization and motiviation**.

Rust is an ambitious language. Making it successul is a big job and there is a lot of work ahead for us. When I ask myself "how would Custom DST help Rust grow over the next year", I don't feel like I hear a convincing answer. The main use case that I see is helping make matrix libraries more ergonomic. Now, admittedly, this is somewhat compelling, but otoh lack of custom dst isn't, I don't think, **blocking** work on Matrix-like things, it's just a way to make them nicer to use. (I'd be interested to hear otherwise; are there any applications where the presence of custom DST would dramatically change the design in some way that it's not even worth doing it otherwise?)

Earlier I had thoughts, which I mentioned briefly in my comment, about the idea of trying to land this work in a very provisional way, to enable hacking. I am still interested in establishing a process for this. I think there are a number of things that might benefit from having a more "experimentation friendly" process (e.g., naked fns, inline asm, and interrupt ABIs come to mind).

But I haven't gone and publicly worked on defining such a process because I also have concerns: the fact is that any time there is significant churn in the codebase, it has the potential to take a lot of time for core developers who are trying to focus on other things. I think the work around `!` has proven this point quite dramatically! By any measure, `!` is not a *core* feature, but working out all the details on it and how it interacts with the unsafe code guidelines, match statements, and so forth has taken and will continue to take quite a lot of time and energy. (I think that much of the work around `!` is work we would have to do anyway, as it happens, but perhaps with less urgency and more at a time of our choosing.) And even the most innocent and orthogonal of feature can cause ICEs and regressions which are a big distraction.

(On the other hand, there are some areas where we *are* adopting a looser process. We did land naked fns in a "provisional state", for example, and that has largely been a non-event. There is some experimentation around the embedded area, particularly with ABIs. But all of these are quite narrow and targeted compared to custom dst, which affects the code that handles every single reference.)

Before @ubsan closed this RFC, I was working up my courage to **move to postpone** (close) -- but I kept hesitating, because I think it's a good idea and I like it. Yet still I have the feeling that on balance it just isn't a good investment of resources. I regret that hesitating perhaps sends a more frustrating message than actually moving to close. So @ubsan, I'm sorry about that.

AtheMathmo:
@nikomatsakis Thank you for your comment.

I agree with what you're saying - there are more pressing things to work on right now. I also wanted to comment on the following:

> Now, admittedly, this is somewhat compelling, but otoh lack of custom dst isn't, I don't think, blocking work on Matrix-like things

I think this is an accurate statement. There are a few things that I cannot do without custom DSTs, some of which don't even have ugly workarounds, but these are only mechanisms to tidy up the code and make things easier for users. Here are some example issues that we had to put off:

https://github.com/AtheMathmo/rulinalg/issues/149 (I _think_ custom DSTs help this)
https://github.com/AtheMathmo/rulinalg/issues/14

It would be really nice to parallel the `Vec<T>`-`&[T]` relationship in the core library but Matrix-like things are still workable without this (to me).

nikomatsakis:
@AtheMathmo interestingly, I was planning to come and post a comment here anyhow just to mention I myself ran into a use-case for this just yesterday when hacking on [my NLL prototype, specifically the bitset code](https://github.com/nikomatsakis/nll/blob/master/graph-algorithms/src/bit_set/mod.rs). Really it's just the "2-D matrix use case" in disguise, but in trying to play with it, I did appreciate that there are things that are hard to do "just with methods". Though maybe I didn't find the best answer. 


In particular, I have a `BitSet` type that supports, for each graph node, a bit set of `N` bits. I wanted the ability to also create a `BitBuf` (one row of N bits) and a `BitSlice` (a view into either `BitSet` or `BitBuf`). This all worked ok but it seemed like I had to copy-and-paste various bits of logic and couldn't easily eliminate the redundancies. I remember thinking that if I could have `Deref` and `DerefMut` impls I would have wound up with less redundancy. Sorry not very specific. I have to revisit it now and try to re-examine.

In any case, I definitely think that this conversation -- to what extent is this "just sugar" vs unlocking some fundamental capability -- is the critical one to deciding how to prioritize this change.

nikomatsakis:
Another thing that I was thinking about was this question: is this the sort of capability that, when added, would mean that the interface for existing libraries would want to be completely reworked to take advantage of it?

AtheMathmo:
>  is this the sort of capability that, when added, would mean that the interface for existing libraries would want to be completely reworked to take advantage of it?

I cannot speak for the others who are looking to make use of this feature, but for me I would be making significant changes.

Right now we have the following types in [rulinalg](https://github.com/AtheMathmo/rulinalg): `Matrix`, `MatrixSlice`, `MatrixSliceMut`. And we have `BaseMatrix` and `BaseMatrixMut` traits to let us be generic over these. If we had custom DSTs then we could use `Deref` and `DerefMut` and (probably) wouldn't want these traits any more - we just implement everything for the slice types and let `Deref` do the work.

Another particularly difficult area is with operator overloading. Right now we have an explosion of overloading implementations for all combinations of matrix types. I think that custom DSTs could help here too but I haven't explored the idea too much. (Hoping `Deref` lets us cover all `&MatType op &OtherMatType` etc. ops).

I'll try to spend some time thinking about whether there is anything else that I can add to this discussion. I've got quite used to working around the lack of custom DSTs and I don't *think* there are many features missing for the user. There are however quite a few ways we could make things nicer - not needing to import traits everywhere, being able to have `&MatrixSlice` return types in their code and with `Index` etc., and others.

nikomatsakis:
@AtheMathmo it seems like it would be a fruitful exercise to try and use e.g. the design [from my comment](https://github.com/rust-lang/rfcs/pull/1524#issuecomment-272020775) and sketch out how the types in your library would work (in contrast to now) and throw it in a gist. (I'd like to see how you envision the operator overloading working, too.) Apologies if you've already done it and I missed it.

AtheMathmo:
@nikomatsakis that sounds like a good idea!

I'm a little busy for the rest of this week but hopefully will be able to sketch something out early next week.

nikomatsakis:
@AtheMathmo great, I'd love to see it. Since this RFC is closed, and it feels like we're still a bit in "design discussion" here, I'm going to open a thread on internals to carry on the conversation: 

https://internals.rust-lang.org/t/custom-dst-discussion/4842

stjepang:
For some more motivation, here's what I need in Crossbeam:

1. Heap-allocated `Array<T>`. The head consists of an `usize` (length), and the tail is `[T]`.
2. Heap-allocated `SkiplistNode<K, V>`. The head consists of `(K, V, usize)` (key, value, height), and the tail is `[AtomicPtr<SkiplistNode<K, V>>]`.
3. Other similar situations with heap-allocated objects. The head is some arbitrary data and the length. The tail is an array of that length.

Pointers to such objects must be thin because atomically manipulating multiple words is a pain to do portably and performantly. Some support from the language for such use of DSTs would be great to have.

SimonSapin:
Heap-allocating `(HeaderIncludingLength, [Item])` (possibly with not all items initialized) is a common pattern that could use some more built-in support. But does it need to be language-level with DSTs, or could it be a library-level?

bergus:
@SimonSapin Looking for exactly that as well. Have you found a suitable workaround yet?

SimonSapin:
@bergus Today, the best you can do is manual memory layout computation, raw allocation, and `unsafe` initialization. See for example:

https://github.com/rust-lang/rust/blob/1.26.2/src/liballoc/rc.rs#L714-L727
https://github.com/rust-lang/rust/blob/1.26.2/src/liballoc/rc.rs#L664-L680

eddyb:
There's also  this pattern for a slice with a header, and `!Sized` thin pointers:
https://github.com/rust-lang/rust/blob/5205ae8bdc2991eecf3bfbb58ed8f56c0673e738/src/librustc/ty/mod.rs#L603-L706

bergus:
@SimonSapin Thanks
@eddyb That's pretty much what I ended up with (needed a bit until I figured out `#[repr(C)]` was necessary…) though without the `OpaqueSliceContents`. I had seen `extern type` mentioned in #813, but the [guide explanation](https://github.com/canndrew/rfcs/blob/extern-types/text/1861-extern-types.md#how-we-teach-this) from the RFC was not exactly helpful. Does this basically prevent us from allocating the dynamically-sized slice on the stack and from statically determining its size, so that we are forced to use it only through pointers?

eddyb:
> Does this basically prevent us from allocating the dynamically-sized slice on the stack and from statically determining its size, so that we are forced to use it only through pointers?

Yes, and if you need mutable references, it prevents someone from (mis-)using `mem::swap`.

velvia:
Hi folks, I'm new to the Rust community, but was really hoping to implement a cross-platform library for high performance compressed vectors using this DST feature.  These vectors would have, for compatibility reasons, a u32 size header and other header fields followed by something like `data: [u8]` (or u64, etc.).  Kind of like what @stjepang and @SimonSapin are asking for.  Having to work around this using unsafe thin pointer coercion and tricks like `[u8; 0]` as markers kind of is discouraging me a bit from porting it from the JVM (though there are certainly other reasons to port it, such as access to SIMD and things like that).

eddyb:
@velvia This was superseded by #2594.

PR #1526, opened 2016-03-04
Remove keepalive methods from net2 RFC

sfackler:
Issues came up in the implementation of these methods that caused us to
back off from implementing keepalive functionality for now.

r? @alexcrichton


alexcrichton:
Yeah these didn't make their way into libstd, and we can continue to work them out in net2 for now, so let's merge and consider re-RFC-ing before adding them to libstd.

PR #1533, opened 2016-03-06
Keyless Entry

gereeter:
Fixes #1203.

### DIFF
--- a/text/0000-keyless-entry.md
+++ b/text/0000-keyless-entry.md
@@ -1,0 +1,150 @@
+- Feature Name: keyless_entry
+- Start Date: 2016-03-06
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add a variation to the entry API for maps that doesn't store the search key in the `Entry`.
+
+# Motivation
+[motivation]: #motivation
+
+The `Entry` API for maps allows users to save time by allowing the user to perform arbitrary modifications
+at a given key dependent upon whether that key was present and if it was, what value was previously there.
+However, although insertion is the only action the user might take that requires a by-value key, there is
+no way to create an `Entry` without a fully owned key. When the user only has a by-reference key on hand,
+this is inefficient at best, requiring an unnecessary `.to_owned` that may involve an expensive allocation
+and/or copy, and unusable at worst, if the key cannot be cloned.
+
+It is a testament to the flexibility and control afforded by the `Entry` API that nearly every method on
+a generic map can be conceptually implemented in terms of `entry`. However, this trick is not used
+and cannot be used in practice because of the same issue - since, e.g., `remove` takes a by-reference key,
+it cannot call `entry` and therefore cannot completely share its implementation with `entry`. To still recover
+as much sharing as possible, map implementations seem to have converged on using an internal variation on the
+`Entry` API, only differing in that the key itself is not stored with the `Entry`. For example, `HashMap` has
+its `FullBucket`s and `VacantEntryState`s, and after rust-lang/rust#32058 will have an `InternalEntry` type,
+while `BTreeMap` has its `Handle`s, and before the rewrite had `SearchStack`s. Making these modified entries
+public gives the user exactly the control they need.
+
+# Detailed design
+[design]: #detailed-design
+
+## API
+
+Two new types are introduced on each map. `VacantKeylessEntry<'a, K, V>` is kept abstract, while
+`KeylessEntry<'a, K, V>` is defined as follows:
+
+```
+pub enum KeylessEntry<'a, K, V> {
+    Occupied(OccupiedEntry<'a, K, V>),
+    Vacant(VacantKeylessEntry<'a, K, V>)
+}
+```
+
+To obtain a `KeylessEntry`, a new method is added to the maps:
+
+```
+impl<K, V> Map<K, V> {
+    /// Gets the given key's corresponding entry in the map for in-place manipulation.
+    pub fn keyless_entry<'a, Q: ?Sized>(&'a mut self, key: &Q) -> KeylessEntry<'a, K, V> where K: Borrow<Q>;
+}
+```
+
+with the appropriate `Ord` or `Eq + Hash` bound on `Q`.
+
+The methods on `KeylessEntry`s mirror those on `Entry`s, only differing in that they take keys when doing
+insertion:
+
+```
+impl<'a, K, V> KeylessEntry<'a, K, V> {
+    /// Ensures a value is in the entry by inserting the default if empty, and returns
+    /// a mutable reference to the value in the entry.
+    pub fn or_insert(self, default_key: K, default: V) -> &'a mut V {
+        match self {
+            Occupied(entry) => entry.into_mut(),
+            Vacant(entry) => entry.insert(default),
+        }
+    }
+
+    /// Ensures a value is in the entry by inserting the result of the default function if empty,
+    /// and returns a mutable reference to the value in the entry.
+    pub fn or_insert_with<F: FnOnce() -> (K, V)>(self, default: F) -> &'a mut V {
+        match self {
+            Occupied(entry) => entry.into_mut(),
+            Vacant(entry) => {
+               let (key, value) = default();
+               entry.insert(key, value)
+            },
+        }
+    }
+}
+
+impl<'a, K: 'a, V: 'a> VacantKeylessEntry<'a, K, V> {
+    /// Sets the value of the entry with the VacantKeylessEntry's key,
+    /// and returns a mutable reference to it
+    pub fn insert(self, key: K, value: V) -> &'a mut V;
+}
+```
+
+`KeylessEntry` and `VacantKeylessEntry` also have methods for converting them into more traditional entries:
+
+```
+impl<'a, K, V> KeylessEntry<'a, K, V> {
+    /// Associates a key with the `KeylessEntry`, making it no longer keyless and just an `Entry`.
+    pub fn with_key(self, key: K) -> Entry<'a, K, V> {
+        match self {
+            Occupied(entry) => Entry::Occupied(entry),
+            Vacant(entry) => Entry::Vacant(entry.with_key(key)),
+        }
+    }
+}
+
+impl<'a, K: 'a, V: 'a> VacantKeylessEntry<'a, K, V> {
+    /// Associates a key with the `VacantKeylessEntry`, making it no longer keyless and just
+    /// a `VacantEntry`.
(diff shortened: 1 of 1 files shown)


killercup:
[Rendered](https://github.com/gereeter/rfcs/blob/keyless-entry/text/0000-keyless-entry.md)

pczarn:
All possible changes have tradeoffs. For me, backwards incompatibility is a non-starter.

Perhaps there is a way of extending the current Entry API with genericity and conversion traits.
```
enum Entry<'a, K, V, Q = K> {
    Occupied(OccupiedEntry<'a, K, V>),
    Vacant(VacantEntry<'a, K, V, Q>),
}
// method on HashMap
fn entry<Q>(&mut self, key: Q) -> Entry<K, V, Q>
    where Q: Borrow<K>
// method on VacantEntry
fn or_insert<W>(self, value: W) -> &'a mut V
    where (K, V): From<(Q, W)>
```
Today's Entry is a special case of this interface with `(K, V) == (Q, W)`. For keyless entry, you'd have a conversion from `(Placeholder, (K, V))` to `(K, V)`, and uglier invocations of `entry.or_insert((k, v))` as the only disadvantage. If trait coherency is a problem, I suppose there could be a new type for key-value pairs. The biggest issue with genericity is always documentation and much more confusing error messages.

pczarn:
Let's reconsider what needs to be done.

The current Entry API is not enough in two scenarios:
* I own a key, and want to retain it if the entry is occupied.
* I have a reference to a key, and I'm not sure if I will need to clone a key to insert it into a vacant entry.

The problem in the first scenario is simple, because it only concerns `OccupiedEntry`.
The problem in the second scenario concerns `VacantEntry` and `fn entry`, so it's basically orthogonal to the first problem.

`OccupiedEntry` doesn't own a key simply because there's no internal need for a retained key. However, the user may want to retain a key, so we modify `OccupiedEntry` and add a method for key recovery. This fixes the first scenario, which I have in my refactoring PR.

```rust
struct NoKey;

struct OccupiedEntry<'a, K, V> {
    key: Option<K>,
    elem: FullBucket<...>
}

impl<'a, K, V> OccupiedEntry<'a, K, V> {
    pub fn take_key(&mut self) -> Option<K> {
        self.key.take()
    }
}
```

Stebalien:
The original purpose of the entry API was to prevent double lookups. This *can't* fix that while `entry_or_clone` can.

gereeter:
> All possible changes have tradeoffs. For me, breaking backwards compatibility is a non-starter.

I completely agree - I included the backwards incompatible alternative just for completeness. The main proposal is not backwards incompatible in any way.

> Perhaps there is a way of extending the current Entry API with genericity and conversion traits like so:

This sort of API would probably be ideal from a usability standpoint, but I don't see any way that it can work. You take `Q` by value to ensure compatibility with passing a by-value key, but this doesn't work with implementations of `Borrow`, as, e.g., `str: Borrow<String>`, not `&str`, so you would have to pass a `str` by value, which is impossible. I don't think that any other trait would fare any better, though I can't completely rule that out.

There may also be compatibility issues with broken type inference, but I don't know how much of a problem that would be.

> The current Entry API is insufficient in two scenarios:
>
> - I own a key, and want to retain that key if the entry is occupied.
> - I have a reference to a key, and I'm not sure if I will need to clone that key to insert it into a vacant entry.

Exactly. This RFC aims to solve only the second problem.

> The original purpose of the entry API was to prevent double lookups. This can't fix that while `entry_or_clone` can.

I'm confused as to why you think that - not only can this API be used to prevent double lookups, but `entry_or_clone` can be implemented in terms of this API:

```rust
fn entry_or_clone(&mut self, key: &K) -> Entry<K, V> where K: Clone {
    match self.keyless_entry(key) {
        KeylessEntry::Occupied(entry) => Entry::Occupied(entry)
        KeylessEntry::Vacant(entry) => Entry::Vacant(entry.with_key(key.clone()))
    }
}
```

Stebalien:
> I'm confused as to why you think that - not only can this API be used to prevent double lookups, but entry_or_clone can be implemented in terms of this API

Sorry, I didn't read the drawbacks. IMO, not being able to guarantee that the key is the same is way too much of a foot gun to be acceptable.

jmesmon:
@Stebalien 

could you clarify what you mean by

> not being able to guarantee that the key is the same

Stebalien:
@jmesmon A user could write:

```rust
self.keyless_entry(&key1).or_insert_with(||(key2, 0)) += 1;
```

and end up with key2 in key1's slot.

comex:
Alternate alternative: combine this proposal and the `entry_or_clone` one.  Use an alternate Entry type (or just extend the existing one to support the new functionality), but have its `insert` method call `to_owned()` on the key rather than allowing it to be specified explicitly, reducing footguniness at the cost of some flexibility.  (`to_owned()` should be preferred over `clone()` because of things like searching a `HashMap<String, X>` with `&str`.)

text/0000-keyless-entry.md
```diff
@@ -0,0 +1,150 @@
...
+by putting in two unequal keys. Although this can mess up the internal structure of the map, it cannot cause
+memory unsafety, as this possibility was already allowed by having a query where `key` and `key.to_owned()`
+are unequal.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Replace the `Entry` API entirely. This would get rid of the inherent duplication of having two `Entry` APIs,
+  but would be backwards incompatible.
+- Create an `OccupiedKeylessEntry` that is separate from `OccupiedEntry` to help keep to two different `Entry`s
```
pczarn: Please add a note saying this alternative is an addition to the main proposal.

pczarn:
I haven't given up on this issue. It can be decomposed into two essential steps.

* I have a borrowed key, and want a keyless entry.
* I have a vacant keyless entry, and want to populate it by computing a key with one of these actions:
 1. I compute and give my owned key to the entry.
 2. I want the entry to use a clone of the key it borrows.
 3. The entry already owns an key, and nothing needs to be done.

Let's focus on the first step.

An API with ideal usability has to take a key that is either borrowed or owned. `Cow` is a natural fit for this purpose and is passed into `entry2` in [hashmap2](https://github.com/contain-rs/hashmap2/). However, we would like to keep backwards compatibility without creating a separate component. Adding `Into` to the mix allows that:

```rust
pub fn entry<'a, Q, B>(&mut self, key: Q) -> Entry<K, V, B>
    where K: Into<Cow<'a, B>>,
          Q: Into<Cow<'a, B>>,
          B: Eq + Hash + ToOwned + 'a,
{
    self.reserve(1);
    let key_cow: Cow<'a, B> = key.into();
    self.search_mut(&key_cow).into_entry(key_cow).expect("unreachable")
}
```

The compiler will complain about `K: Clone` not being satisfied in various places. Yes, we broke backward compatibility. I gave up on solving that. Instead, we can write simpler bounds with `Borrow` in a similar manner. This will require a method with default type parameters that drive inference.

```rust
pub fn entry<'a, Q = K, B = K>(&mut self, key: Q) -> Entry<K, V, Q>
    where Q: Borrow<B>,
          K: Borrow<B>,
          B: Eq + Hash,
{
    self.reserve(1);
    self.search_mut(Borrow::<B>::borrow(&key)).into_entry(key).expect("unreachable")
}
```

These default type parameters are necessary, because some types have two different implementations of `Borrow<…>` and `Borrow<Self>`. We ask the compiler's inference to choose `Borrow<Self>`. This way, backward compatibility is be maintained for the following code:

```rust
fn foo<K>(map: HashMap<String, …>, key: String) {
	map.entry(key);
	// Types can be inferred in two ways:
	//   first
	map.entry::<String, String>(key);
	//   second, due to `String: Borrow<str>`
	map.entry::<String, str>(key);
}

```

As for the second step, the Entry API should be able to provide a function transforming an `Entry<K, V, Q>` into `Entry<K, V, K>`.

cristicbz:
So I've got a suggestion ([playground](https://is.gd/1zEWUB)) which may not require an additional method. The idea is to introduce a new `IntoOwned` trait which unlike `ToOwned` takes self by-value:

```rust
pub trait IntoOwned<T>{
    fn into_owned(self) -> T;
}

impl<T> IntoOwned<T> for T {
    fn into_owned(self) -> T { self }
}

impl<'a, T> IntoOwned<T> for &'a T where T: Clone {
    fn into_owned(self) -> T {
        (*self).clone()
    }
}

impl<'a> IntoOwned<String> for &'a str {
    fn into_owned(self) -> String {
        self.to_owned()
    }
}

/* ... */
```

The `entry` method's signature can then look like this:

```rust
    // Entry would have to gain Q: IntoOwned<K> as a type param with default Q=K.
    fn entry<'a, Q>(&'a mut self, key: Q) -> Entry<'a, K, V, Q>
        where K: Hash + Eq,
              Q: Hash + Eq + IntoOwned<K>
    {
        // Build an `Entry` which stores `key` and calls `into_owned`
        // on vacant insertion only.     
    }
```

The trouble with this is that we'd need to duplicate all the existing `ToOwned` impl-s (with some `&'a`-s in there). This is because

```rust
  impl<'a, T> IntoOwned<T::Owned> for &'a T where T: ToOwned
```

conflicts with `impl<T> IntoOwned<T> for T` due to the `T::Owned == T` possibility. Had we gone this route to begin with (and not have `ToOwned`) this version might have been more appealing.

As it stands, we could keep `IntoOwned` unstable and provide `impl`-s for `T: Clone` and the `std` smart pointers. It'd still greatly improve the situation, but the lack of extensibility for any `Borrow`/`ToOwned` pairs is disappointing.

cristicbz:
Just to bring in some conversation from IRC. The only coherence trouble the impl runs into is of the form `impl<T, U> X<T> for U where T != U`. The typical arguments against negative reasoning don't apply here (at first glance anyway), so assuming that where clause was expressible the following ought to work:

```rust
impl<T> IntoOwned<T> for T {
    fn into_owned(self) -> T { self }
}

impl<'a, T> IntoOwned<T::Owned> for &'a T
    where T: ToOwned,
          T::Owned != &'a T
{
    fn into_owned(self) -> T::Owned {
        (*self).to_owned()
    }
}
```

cristicbz:
I worked around the coherence issues an wrote up an RFC #1769

ubsan:
ping @gereeter @aturon 

Status? Should this be closed, given the existence of #1769?

aturon:
I've just proposed a move to FCP on #1769. I'm going to propose to close this RFC as part of the same decision.

@rfcbot fcp close

gereeter:
I approve of this being closed - I think #1769 is a better solution.

alexcrichton:
Ok, all checkboxes checked, closing in favor of #1769

PR #1542, opened 2016-03-12
Add TryFrom and TryInto traits

sfackler:
[Rendered](https://github.com/sfackler/rfcs/blob/try-from/text/0000-try-from.md)

### DIFF
--- a/text/0000-try-from.md
+++ b/text/0000-try-from.md
@@ -1,0 +1,165 @@
+- Feature Name: try_from
+- Start Date: 2016-03-10
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+The standard library provides the `From` and `Into` traits as standard ways to
+convert between types. However, these traits only support *infallable*
+conversions. This RFC proposes the addition of `TryFrom` and `TryInto` traits
+to support these use cases in a standard way.
+
+# Motivation
+[motivation]: #motivation
+
+Fallible conversions are fairly common, and a collection of ad-hoc traits has
+arisen to support them, both [within the standard library][from-str] and [in
+third party crates][into-connect-params]. A standardized set of traits
+following the pattern set by `From` and `Into` will ease these APIs by
+providing a standardized interface as we expand the set of fallible
+conversions.
+
+One specific avenue of expansion that has been frequently requested is fallible
+integer conversion traits. Conversions between integer types may currently be
+performed with the `as` operator, which will silently truncate the value if it
+is out of bounds of the target type. Code which needs to down-cast values must
+manually check that the cast will succeed, which is both tedious and error
+prone. A fallible conversion trait reduces code like this:
+
+```rust
+let value: isize = ...;
+
+let value: u32 = if value < 0 || value > u32::max_value() as isize {
+    return Err(BogusCast);
+} else {
+    value as u32
+};
+```
+
+to simply:
+
+```rust
+let value: isize = ...;
+let value: u32 = try!(value.try_into());
+```
+
+# Detailed design
+[design]: #detailed-design
+
+Two traits will be added to the `core::convert` module:
+
+```rust
+pub trait TryFrom<T>: Sized {
+    type Err;
+
+    fn try_from(t: T) -> Result<Self, Self::Err>;
+}
+
+pub trait TryInto<T>: Sized {
+    type Err;
+
+    fn try_into(self) -> Result<T, Self::Err>;
+}
+```
+
+In a fashion similar to `From` and `Into`, a blanket implementation of `TryInto`
+is provided for all `TryFrom` implementations:
+
+```rust
+impl<T, U> TryInto<U> for T where U: TryFrom<T> {
+    type Error = U::Err;
+
+    fn try_into(self) -> Result<U, Self::Err> {
+        U::try_from(self)
+    }
+}
+```
+
+In addition, implementations of `TryFrom` will be provided to convert between
+*all combinations* of integer types:
+
+```rust
+#[derive(Debug)]
+pub struct TryFromIntError(());
+
+impl fmt::Display for TryFromIntError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.write_str(self.description())
+    }
+}
+
+impl Error for TryFromIntError {
+    fn description(&self) -> &str {
+        "out of range integral type conversion attempted"
+    }
+}
+
+impl TryFrom<usize> for u8 {
+    type Err = TryFromIntError;
+
+    fn try_from(t: usize) -> Result<u8, TryFromIntError> {
+        // ...
+    }
+}
+
+// ...
+```
+
+This notably includes implementations that are actually infallible, including
+implementations between a type and itself. A common use case for these kinds
+of conversions is when interacting with a C API and converting, for example,
+from a `u64` to a `libc::c_long`. `c_long` may be `u32` on some platforms but
+`u64` on others, so having an `impl TryFrom<u64> for u64` ensures that
+conversions using these traits will compile on all architectures.  Similarly, a
+conversion from `usize` to `u32` may or may not be fallible depending on the
+target architecture.
+
+The standard library provides a reflexive implementation of the `From` trait
+for all types: `impl<T> From<T> for T`. We could similarly provide a "lifting"
+implementation of `TryFrom`:
+
+```rust
+impl<T, U: From<T>> TryFrom<T> for U {
+    type Err = Void;
+
+    fn try_from(t: T) -> Result<U, Void> {
+        Ok(U::from(t))
+    }
+}
+```
+
+However, this implementation would directly conflict with our goal of having
+uniform `TryFrom` implementations between all combinations of integer types. In
+addition, it's not clear what value such an implementation would actually
+provide, so this RFC does *not* propose its addition.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It is unclear if existing fallible conversion traits can backwards-compatibly
+be subsumed into `TryFrom` and `TryInto`, which may result in an awkward mix of
+ad-hoc traits in addition to `TryFrom` and `TryInto`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+We could avoid general traits and continue making distinct conversion traits for
+each use case.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Are `TryFrom` and `TryInto` the right names? There is some precedent for the
+`try_` prefix: `TcpStream::try_clone`, `Mutex::try_lock`, etc.
+
+What should be done about `FromStr`, `ToSocketAddrs`, and other ad-hoc fallible
+conversion traits? An upgrade path may exist in the future with specialization,
+but it is probably too early to say definitively.
+
+Should `TryFrom` and `TryInto` be added to the prelude? This would be the first
+prelude addition since the 1.0 release.
+
+[from-str]: https://doc.rust-lang.org/1.7.0/std/str/trait.FromStr.html
+[into-connect-params]: http://sfackler.github.io/rust-postgres/doc/v0.11.4/postgres/trait.IntoConnectParams.html


seanmonstar:
Additional existing third party trait: hyper::IntoUrl.

comex:
:-1: on implementing it for infallible conversions.

Ticki:
I wonder how this interacts with enum casting? Will it be possible to cast an integral representation _to_ an enum variant?

sfackler:
@comex one of the major motivations for implementing fallible conversion traits for integers is for FFI use cases. Could you clarify how you see ergonomic casts between, for example, `u64` and `c_ulong` working if there is no `impl TryFrom<u64> for u64`?

text/0000-try-from.md
```diff
@@ -0,0 +1,165 @@
...
+} else {
+    value as u32
+};
+```
+
+to simply:
+
+```rust
+let value: isize = ...;
+let value: u32 = try!(value.try_into());
```
petrochenkov: This example is pretty untypical.
These overflows, like arithmetic overflows, are very rarely handled, so most of the time people will have to deal with
```
let value: u32 = value.try_into().unwrap();
```

text/0000-try-from.md
```diff
@@ -0,0 +1,165 @@
...
+We could avoid general traits and continue making distinct conversion traits for
+each use case.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Are `TryFrom` and `TryInto` the right names? There is some precedent for the
+`try_` prefix: `TcpStream::try_clone`, `Mutex::try_lock`, etc.
+
+What should be done about `FromStr`, `ToSocketAddrs`, and other ad-hoc fallible
```
petrochenkov: I dislike the tendency of replacing concrete methods with meaningless generic implementations.
I'd never want `parse(&str) -> Result<i32, Err>` to become `try_into(&str) -> Result<i32, Err>`.
`From/Into` are already overused in this sense, `impl From<Ipv4Addr> for u32` goddamit.

Is anyone going to write generic code which would be applicable to both `FromStr` and `ToSocketAddrs`? I suppose no. It means than `TryFrom`/`TryInto` in this case is a dangerous over-generalization.

I'm not against `TryFrom`/`TryInto` in principle, but I'm pretty sure they will be heavily misused.

SimonSapin: I don’t know what should be concluded from this, but note that `ToSocketAddrs` has an associated type for the `Ok` case.

sfackler: @petrochenkov Why do you feel that `impl FromStr for Foo` is an intrinsically more meaningful implementation than `impl<'a str> TryFrom<&'a str> for Foo`? They seem totally isomorphic to me.'

It's also worth noting that the fate of `parse` is basically unrelated to the fate of `FromStr`.

text/0000-try-from.md
sfackler: It's not all that uncommon in my experience:
https://github.com/sfackler/rust-postgres/blob/master/src/types/mod.rs#L700
https://github.com/sfackler/rust-postgres-large-object/blob/master/src/lib.rs#L220

But in any case, what you do with the `Err` once you get it doesn't really have much to do with this RFC - it's focused on how you get that `Err`.

text/0000-try-from.md
petrochenkov: >It's also worth noting that the fate of parse is basically unrelated to the fate of FromStr.

`parse` itself is not affected, it'll just get a generic duplicate `try_into` if `TryFrom<str>` is implemented for the same types as `FromStr` and `TryInto` gets the blanket impl based on `TryFrom`.
And I'm saying that this duplicate will almost never be used, because `parse` is preferable in concrete code and probably never used in code generic enough to not be able to use `parse`.

>Why do you feel that impl FromStr for Foo is an intrinsically more meaningful implementation than impl<'a str> TryFrom<&'a str> for Foo?

I don't feel that way with respect to `FromStr` in particular, `FromStr` is already meaningless enough :)

jimmycuadra:
From an API discoverability perspective, I think these traits would be useful. I was looking for something that did this a while back and was surprised to find that there was no fallible equivalent of `From` and `Into`. Instead, some kind folks on IRC directed me to the totally non-obvious `parse` and `FromStr`. Once you know about them, it works okay, but the proposed API is much more intuitive to me.

Zoxc:
I would like the signature of the methods to be changed to return `(Result<To, (From, Self::Err)>)` where `From` is the original from value which can be printed on error without requiring `Clone`.

For example in this function:
```rust
fn coerce<B, A: TryInto<B> + Debug>(from: A) -> B {
	match from.try_into() {
		Ok(v) => v,
		Err((f, err)) => {
			panic!("Cannot coerce value {} of type {} into type {} because {}", f, type_name::<A>(), type_name::<B>(), err);
		}
	}
}
```

Alternatively there could be a method on the error giving a way to extract it.

glaebhoerl:
I think that's a good idea. A lot of `try_foo`-style methods that already exist do that because many types don't even *implement* `Clone`. (Or it's just expensive.)

SimonSapin:
To repeat my comment in https://github.com/rust-lang/rust/issues/31405, the naming convention of `try_FOO` for things that return `Result` is consistent with `try_clone`.

The `try!` macro however *takes* a `Result` and does not generally return one. So we may end up with `try!(x.try_clone())` and `try!(y.try_into())` (which sounds repetitive and has two different meaning for the same word) being used a lot… But then if the `?` operator sticks around maybe `try!` will be deprecated or go out of fashion, leaving us with `x.try_clone()?` and `y.try_into()?` which looks nicer to me.

alexcrichton:

alexcrichton:
Overall I like the idea of having these traits in the standard library as they provide a nice vector for us to implement all sorts of conversions in a standardized fashion. I also think that the naming here, (as @SimonSapin [mentioned](https://github.com/rust-lang/rfcs/pull/1542#issuecomment-206804137)) seems good (I'm personally less worried about the `try` stuttering).

On question I'd have is whether we'd want to add these traits to the prelude? One of the reasons `Into` and `From` are so easy to use is because of their prominent placement and lack of a need to import them. I'd be of the opinion that these likely want to go into the prelude at *some point*, but we may not have all of the resolve changes in place to do that without unduly breaking crates. To me though one of the major motivations for this RFC is integer conversions, which I believe aren't gonna work out *unless* they're in the prelude, so I'd be fine adding it sooner rather than later if we can.

Finally, as to some other points brought up on this RFC:

* I like the addition impls of the traits between every combination of primitive types. That's easy to understand, always works, and is predictable.
* I'm ok not having a "lifting" impl from `From` to `TryFrom`. That'd also require the addition of a `Void` type which while it has come up from time to time hasn't ever been clearly well motivated for libstd.
* I don't think we can do anything about other ad-hoc traits unless we don't break code. That may require specialization or some other kinds of trickery that may take awhile to create. That being said while they both exist we should strive to make them interoperable and have the same set of impls for both (at least). Methods like `parse` won't go away, they may just change trait bounds if it's compatible.
* I don't think we should change the `Result` type returned to carry the payload of the original object. Not all implementors may always have access to the original object once a conversion has failed, and the error type can always contain the original object anyway. Additionally, that'd close off the door for an `Error` implementation for the error side of `Result`, so it may not always work with `try!`/`?`

text/0000-try-from.md
```diff
@@ -0,0 +1,165 @@
...
+    fn try_into(self) -> Result<T, Self::Err>;
+}
+```
+
+In a fashion similar to `From` and `Into`, a blanket implementation of `TryInto`
+is provided for all `TryFrom` implementations:
+
+```rust
+impl<T, U> TryInto<U> for T where U: TryFrom<T> {
+    type Error = U::Err;
```
nagisa: Nit: `Err`, not `Error`.

text/0000-try-from.md
```diff
@@ -0,0 +1,165 @@
...
+    fn try_from(t: T) -> Result<U, Void> {
+        Ok(U::from(t))
+    }
+}
+```
+
+However, this implementation would directly conflict with our goal of having
+uniform `TryFrom` implementations between all combinations of integer types. In
+addition, it's not clear what value such an implementation would actually
+provide, so this RFC does *not* propose its addition.
```
nagisa: I think this is a very important implementation to add.

I write code like this *all* the time:

```
fn foo<T: Into<Bar>>(bar: T) -> … {
    let bar = bar.into();
}
```

which then can be used with anything that implements `From`. I would certainly like to be able to do

```
fn foo_2<T: TryInto<Bar>>(bar: T) -> … {
    let bar = bar.try_into().unwrap();
}
```

and be able to pass in anything implementing plain old `From` as well. 

The coherence problem could’ve been a reason for not considering this at the time of writing, but since specialisation has been implemented, perhaps it can be made to work?

sfackler: I do not believe this can be done in the current implementation of specialization.

petrochenkov:
`Into` usually goes hand by hand with `AsRef` and `AsMut`, so I'm surprised nobody mentioned `TryAsRef` and `TryAsMut` yet.
Any fundamental reasons why they would be less useful than their infallible counterparts or `TryInto`?

sfackler:
It seems to me that there would be very few contexts in which you'd actually be able to implement `TryAsRef` or `TryAsMut` since they return references. I can't really think of contexts in which you possibly have a reference to some other type.

nodakai:
On which types are we going to implement the proposed traits?

1. Are they solely for built-in integer types?
2. Float types?
3. Boolean?
4. Pointers?
5. I'd like to hear more discussions on ad-hoc conversions like `&str` <-> `Ipv4Addr`
6. Enums? (Core language change?)

Also, the proposed interface does not obviously eliminate such silly implementations like

```rust
    fn try_from(t: usize) -> Result<u8, TryFromIntError> { Ok(42) }
```

Is it possible to add kind of "axioms" that all implementers should satisfy? For example, for `S: Copy+Eq` and `T:Copy+Eq`, I'd expect this assertion to hold for all implementers:

```rust
    let x0: S = ...;
    let x1: S = ...;
    let y0: T = x0.try_into()?;
    let y1: T = x1.try_into()?;
    assert_eq!(x0 == x1, y0 == y1);
```

text/0000-try-from.md
aturon: Yes, IIRC this requires the "lattice rule". (I expect something with that level of expressiveness to be available eventually, but it will be a while.)

aturon:
I'm in favor of this RFC. To me, the biggest use-case is ergonomic overloaded construction via `try_from`, the same way that we replaced a pile of ad hoc constructors with `from` with the initial conversion traits.

@alexcrichton has already made most of the other points I wanted to, so I'll leave it at :+1:!

sfackler:
@nodakai 

> On which types are we going to implement the proposed traits?
> 
> Are they solely for built-in integer types?
> Float types?
> Boolean?
> Pointers?
> I'd like to hear more discussions on ad-hoc conversions like &str <-> Ipv4Addr
> Enums? (Core language change?)

The traits are for things that can be converted to another type in a potentially fallible way, and as is the case for `From` and `Into`, the set of implementations will expand over time.

Implementations of `FromStr` are the kinds of things that would work as as `TryFrom`/`TryInto` implementations, which includes the `&str -> Ipv4Addr` conversion.

> Also, the proposed interface does not obviously eliminate such silly implementations like
> 
>     fn try_from(t: usize) -> Result<u8, TryFromIntError> { Ok(42) }

In what way is this different from any other interface in the standard library? We can't guard against pathologically malicious implementors.

> Is it possible to add kind of "axioms" that all implementers should satisfy? For example, for S: Copy+Eq and T:Copy+Eq, I'd expect this assertion to hold for all implementers:
> 
>     let x0: S = ...;
>     let x1: S = ...;
>     let y0: T = x0.try_into()?;
>     let y1: T = x1.try_into()?;
>     assert_eq!(x0 == x1, y0 == y1);

I do not see us making any strong guarantees in this sense - we don't for `Into` and `From`, which simply state that they are "conversion that consumes self, which may or may not be expensive". In particular, any guarantees like the one proposed will disallow a `&str -> Ipv4Addr` implementation since it involves DNS lookup that may change over time.

sfackler:
@alexcrichton 

The prelude question is pretty complicated. We certainly can't add these traits to the prelude with the current resolve implementation. The `conv` crate provides `TryFrom` and `TryInto` traits identical to those proposed here, and we wouldn't want to break all crates using those.

I think these will be useful/used in practice for integral conversions because they are already a pain to write. Something like this:

```rust
use std::u32;
// ...
let x = if x < 0 || x > u32::max_value() as i64 {
    return Err(Whatever);
} else {
    x as u32
};
```

will be simplified to

```rust
use std::convert::TryInto;
// ...
let x = try!(x.try_into());
```

In particular, you have to import something either way.

ollie27:
I'm also curious if there are any concrete use cases other than int to int conversions which in my opinion are much better handled by something like #1218?

> I like the addition impls of the traits between every combination of primitive types.

I assume this was meant to say primitive integers as all primitive types doesn't make any sense.

One of the main advantages of `From` is that we have `impl<T> From<T> for T`. If `TryFrom` doesn't have that which types should manually implement `TryFrom` for themselves?

I kind of think fallible conversions can be left a bit ad hoc, for example `Vec<u8>` to `String` has `from_utf8` which presumably is a good use case for this trait but that still leaves `from_utf8_lossy` and `from_utf8_unchecked` as ad hoc methods.

sfackler:
@ollie27 The RFC mentions some in the motivation section. Here's a list of traits that all do this off the top of my head:

* `std::str::FromStr`
* `std::net::ToSocketAddrs`
* `postgres::IntoConnectParams`
* `hyper::IntoUrl`

SimonSapin:
> I'm ok not having a "lifting" impl from From to TryFrom. That'd also require the addition of a Void type which while it has come up from time to time hasn't ever been clearly well motivated for libstd.

FWIW `std` already has at least one `Void`-like type: `std::string::ParseError` is an empty enum used for `<String as FromStr>::Err`.

SimonSapin:
> In particular, any guarantees like the one proposed will disallow a &str -> Ipv4Addr implementation since it involves DNS lookup that may change over time.

(`&str -> Ipv4Addr` doesn’t have to do DNS, it can simply support IPv4 syntax like `Ipv4Addr::from_str` currently does. I’d even argue that any kind of external lookup is *not* appropriate for `TryFrom`/`TryInto`.)

petrochenkov:
@sfackler 
>We certainly can't add these traits to the prelude with the current resolve implementation.

Why exactly?

>The conv crate provides TryFrom and TryInto traits identical to those proposed here, and we wouldn't want to break all crates using those.

Prelude names are shadowable, if code uses `TryFrom` from the `conv` crate, it will continue use it and `prelude::TryFrom` will be shadowed.

sfackler:
Hmm, I may have imagined that there would be more troubles adding this to the prelude than there are.

nodakai:
@sfackler 
     
> The traits are for things that can be converted to another type in a potentially fallible way, and as is the case for `From` and `Into`, the set of implementations will expand over time.

I thought it would be nice to discuss how far it should be expanded in the future at this stage, and in order to do so, we need some guideline on its applicability.

> In what way is this different from any other interface in the standard library? We can't guard against pathologically malicious implementors.

> I do not see us making any strong guarantees in this sense - we don't for `Into` and `From`, which simply state that they are "conversion that consumes self, which may or may not be expensive".

I might argue that's just an unfortunate lack and shouldn't be taken as the norm.

The prominent applications of the proposed crates are built-in integers, and their success/failure will be defined by whether "the value is preserved" or not, right? I was wondering if all the implementers could follow an extended definition of "value-preserving."

> In particular, any guarantees like the one proposed will disallow a `&str -> Ipv4Addr` implementation since it involves DNS lookup that may change over time.

As @SimonSapin pointed out, it can still make sense without DNS, but anyways, I'd argue this is exactly why I'd like to bring up the point at this stage, not at the implementation phase.

alexcrichton:
The libs team discussed this RFC at triage today and the decision was to **merge**. We had some small discussion about the prelude-ness of these traits, but the conclusion was that this likely wants to wait until they're stable in any case.

Tracking issue: https://github.com/rust-lang/rust/issues/33417

Kixunil:
@petrochenkov Regarding `TryAsRef` I know of at least one case in which it would be useful: Json (or some kind of dynamic type). The user of library might want to read some value in Json and return `Err` if the value doesn't exist or has different type.

Kroc:
Just learning Rust, and was surprised this wasn't in stable! I'm writing an assembler and I'm using a data-driven methodology where I'm avoiding any functions; all computing is done by type conversion; e.g. you want to parse a string? Convert it into a TokenStream, the logic is internal. Where this falls apart is fallibility; If I'm trying to convert a string token to a scalar value (such as `parse`), there isn't an easy way to error handle this other than panics.

`TryFrom` solves my issue perfectly!

steveklabnik:
We all are. It’s run into some really unfortunate issues. Until then, there’s the crates.io package.

PR #1558, opened 2016-03-25
RFC: Allow coercing non-capturing closures to function pointers.

archshift:
[Rendered](https://github.com/archshift/rfcs/blob/closure-to-fn-coercion/text/0000-closure-to-fn-coercion.md)

Closes #1555.

### DIFF
--- a/text/0000-closure-to-fn-coercion.md
+++ b/text/0000-closure-to-fn-coercion.md
@@ -20,8 +20,8 @@ closure with a certain type signature.
 It is not possible to define a function while at the same time binding it to a
 function pointer.
 
-This is mainly used for convenience purposes, but in certain situations
-the lack of ability to do so creates a significant amount of boilerplate code.
+This is, admittedly, a convenience-motivated feature, but in certain situations
+the inability to bind code this way creates a significant amount of boilerplate.
 For example, when attempting to create an array of small, simple, but unique functions,
 it would be necessary to pre-define each and every function beforehand:
 
@@ -40,10 +40,10 @@ const foo: [fn(&mut u32); 4] = [
 ```
 
 This is a trivial example, and one that might not seem too consequential, but the
-code doubles with every new item added to the array. With very many elements,
+code doubles with every new item added to the array. With a large amount of elements,
 the duplication begins to seem unwarranted.
 
-Another option, of course, is to use an array of `Fn` instead of `fn`:
+A solution, of course, is to use an array of `Fn` instead of `fn`:
 
 ```rust
 const foo: [&'static Fn(&mut u32); 4] = [
@@ -54,19 +54,28 @@ const foo: [&'static Fn(&mut u32); 4] = [
 ];
 ```
 
-And this seems to fix the problem. Unfortunately, however, looking closely one
-can see that because we use the `Fn` trait, an extra layer of indirection
-is added when attempting to run `foo[n](&mut bar)`.
+And this seems to fix the problem. Unfortunately, however, because we use
+a reference to the `Fn` trait, an extra layer of indirection is added when
+attempting to run `foo[n](&mut bar)`.
 
-Rust must use dynamic dispatch because a closure is secretly a struct that
-contains references to captured variables, and the code within that closure
-must be able to access those references stored in the struct.
+Rust must use dynamic dispatch in this situation; a closure with captures is nothing
+but a struct containing references to captured variables. The code associated with a
+closure must be able to access those references stored in the struct.
 
-In the above example, though, no variables are captured by the closures,
-so in theory nothing would stop the compiler from treating them as anonymous
-functions. By doing so, unnecessary indirection would be avoided. In situations
-where this function pointer array is particularly hot code, the optimization
-would be appreciated.
+In situations where this function pointer array is particularly hot code,
+any optimizations would be appreciated. More generally, it is always preferable
+to avoid unnecessary indirection. And, of course, it is impossible to use this syntax
+when dealing with FFI.
+
+Aside from code-size nits, anonymous functions are legitimately useful for programmers.
+In the case of callback-heavy code, for example, it can be impractical to define functions
+out-of-line, with the requirement of producing confusing (and unnecessary) names for each.
+In the very first example given, `inc_X` names were used for the out-of-line functions, but
+more complicated behavior might not be so easily representable.
+
+Finally, this sort of automatic coercion is simply intuitive to the programmer.
+In the `&Fn` example, no variables are captured by the closures, so the theory is
+that nothing stops the compiler from treating them as anonymous functions.
 
 # Detailed design
 [design]: #detailed-design
@@ -107,17 +116,15 @@ const foo: [fn(&mut u32); 4] = [
 ];
 ```
 
-Note that once explicitly assigned to an `Fn` trait, the closure can no longer be
-coerced into `fn`, even if it has no captures. Just as we cannot do:
+Because there does not exist any item in the language that directly produces
+a `fn` type, even `fn` items must go through the process of reification. To
+perform the coercion, then, rustc must additionally allow the reification of
+unsized closures to `fn` types. The implementation of this is simplified by the
+fact that closures' capture information is recorded on the type-level.
 
-```rust
-let a: u32 = 0; // Coercion
-let b: i32 = a; // Can't re-coerce
-let x: *const u32 = &a; // Coercion
-let y: &u32 = x; // Can't re-coerce
-```
+*Note:* once explicitly assigned to an `Fn` trait, the closure can no longer be
+coerced into `fn`, even if it has no captures.
 
-We can't similarly re-coerce a `Fn` trait.
 ```rust
 let a: &Fn(u32) -> u32 = |foo: u32| { foo + 1 };
 let b: fn(u32) -> u32 = *a; // Can't re-coerce
@@ -127,7 +134,7 @@ let b: fn(u32) -> u32 = *a; // Can't re-coerce
 [drawbacks]: #drawbacks
 
 This proposal could potentially allow Rust users to accidentally constrain their APIs.
-In the case of a crate, a user accidentally returning `fn` instead of `Fn` may find
+In the case of a crate, a user returning `fn` instead of `Fn` may find
 that their code compiles at first, but breaks when the user later needs to capture variables:
 
 ```rust
@@ -158,13 +165,13 @@ fn func_general<'a>(&'a mut self) -> impl FnMut() -> u32 {
 }
 ```
 
-This drawback is probably outweighed by convenience, simplicity, and the potential for optimization
-that comes with the proposed changes, however.
+This aspect is probably outweighed by convenience, simplicity, and the potential for optimization
+that comes with the proposed changes.
 
 # Alternatives
 [alternatives]: #alternatives
 
-## Anonymous function syntax
+## Function literal syntax
 
 With this alternative, Rust users would be able to directly bind a function
 to a variable, without needing to give the function a name.
@@ -184,12 +191,24 @@ const foo: [fn(&mut u32); 4] = [
 ```
 
 This isn't ideal, however, because it would require giving new semantics
-to `fn` syntax.
+to `fn` syntax. Additionally, such syntax would either require explicit return types,
+or additional reasoning about the literal's return type.
+
+```rust
+fn(x: bool) { !x }
+```
+
+The above function literal, at first glance, appears to return `()`. This could be
+potentially misleading, especially in situations where the literal is bound to a
+variable with `let`.
+
+As with all new syntax, this alternative would carry with it a discovery barrier.
+Closure coercion may be preferred due to its intuitiveness.
 
 ## Aggressive optimization
 
 This is possibly unrealistic, but an alternative would be to continue encouraging
-the use of closures with the `Fn` trait, but conduct heavy optimization to determine
+the use of closures with the `Fn` trait, but use static analysis to determine
 when the used closure is "trivial" and does not need indirection.
 
 Of course, this would probably significantly complicate the optimization process, and
@@ -199,4 +218,5 @@ checking the disassembly of their program.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-None
+Should we generalize this behavior in the future, so that any zero-sized type that
+implements `Fn` can be converted into a `fn` pointer?
--- a/text/0000-closure-to-fn-coercion.md
+++ b/text/0000-closure-to-fn-coercion.md
@@ -6,19 +6,19 @@
 # Summary
 [summary]: #summary
 
-A non-capturing (that is, does not `Clone` or `move` any local variables) should be
-coercable to a function pointer (`fn`).
+A non-capturing (that is, does not `Clone` or `move` any local variables) closure
+should be coercable to a function pointer (`fn`).
 
 # Motivation
 [motivation]: #motivation
 
-Currently in rust, it is impossible to bind anything but a pre-defined function
+Currently in Rust, it is impossible to bind anything but a pre-defined function
 as a function pointer. When dealing with closures, one must either rely upon
-rust's type-inference capabilities, or use the `Fn` trait to abstract for any
+Rust's type-inference capabilities, or use the `Fn` trait to abstract for any
 closure with a certain type signature.
 
-What is not possible, though, is to define a function while at the same time
-binding it to a function pointer.
+It is not possible to define a function while at the same time binding it to a
+function pointer.
 
 This is mainly used for convenience purposes, but in certain situations
 the lack of ability to do so creates a significant amount of boilerplate code.
@@ -107,19 +107,66 @@ const foo: [fn(&mut u32); 4] = [
 ];
 ```
 
+Note that once explicitly assigned to an `Fn` trait, the closure can no longer be
+coerced into `fn`, even if it has no captures. Just as we cannot do:
+
+```rust
+let a: u32 = 0; // Coercion
+let b: i32 = a; // Can't re-coerce
+let x: *const u32 = &a; // Coercion
+let y: &u32 = x; // Can't re-coerce
+```
+
+We can't similarly re-coerce a `Fn` trait.
+```rust
+let a: &Fn(u32) -> u32 = |foo: u32| { foo + 1 };
+let b: fn(u32) -> u32 = *a; // Can't re-coerce
+```
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
-To a rust user, there is no drawback to this new coercion from closures to `fn` types.
+This proposal could potentially allow Rust users to accidentally constrain their APIs.
+In the case of a crate, a user accidentally returning `fn` instead of `Fn` may find
+that their code compiles at first, but breaks when the user later needs to capture variables:
 
-The only drawback is that it would add some amount of complexity to the type system.
+```rust
+// The specific syntax is more convenient to use
+fn func_specific(&self) -> (fn() -> u32) {
+  || return 0
+}
+
+fn func_general<'a>(&'a self) -> impl Fn() -> u32 {
+  move || return self.field
+}
+```
+
+In the above example, the API author could start off with the specific version of the function,
+and by circumstance later need to capture a variable. The required change from `fn` to `Fn` could
+be a breaking change.
+
+We do expect crate authors to measure their API's flexibility in other areas, however, as when
+determining whether to take `&self` or `&mut self`. Taking a similar situation to the above: 
+
+```rust
+fn func_specific<'a>(&'a self) -> impl Fn() -> u32 {
+  move || return self.field
+}
+    
+fn func_general<'a>(&'a mut self) -> impl FnMut() -> u32 {
+  move || { self.field += 1; return self.field; }
+}
+```
+
+This drawback is probably outweighed by convenience, simplicity, and the potential for optimization
+that comes with the proposed changes, however.
 
 # Alternatives
 [alternatives]: #alternatives
 
 ## Anonymous function syntax
 
-With this alternative, rust users would be able to directly bind a function
+With this alternative, Rust users would be able to directly bind a function
 to a variable, without needing to give the function a name.
 
 ```rust
--- a/text/0000-closure-to-fn-coercion.md
+++ b/text/0000-closure-to-fn-coercion.md
@@ -1,0 +1,155 @@
+- Feature Name: closure_to_fn_coercion
+- Start Date: 2016-03-25
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+A non-capturing (that is, does not `Clone` or `move` any local variables) should be
+coercable to a function pointer (`fn`).
+
+# Motivation
+[motivation]: #motivation
+
+Currently in rust, it is impossible to bind anything but a pre-defined function
+as a function pointer. When dealing with closures, one must either rely upon
+rust's type-inference capabilities, or use the `Fn` trait to abstract for any
+closure with a certain type signature.
+
+What is not possible, though, is to define a function while at the same time
+binding it to a function pointer.
+
+This is mainly used for convenience purposes, but in certain situations
+the lack of ability to do so creates a significant amount of boilerplate code.
+For example, when attempting to create an array of small, simple, but unique functions,
+it would be necessary to pre-define each and every function beforehand:
+
+```rust
+fn inc_0(var: &mut u32) {}
+fn inc_1(var: &mut u32) { *var += 1; }
+fn inc_2(var: &mut u32) { *var += 2; }
+fn inc_3(var: &mut u32) { *var += 3; }
+
+const foo: [fn(&mut u32); 4] = [
+  inc_0,
+  inc_1,
+  inc_2,
+  inc_3,
+];
+```
+
+This is a trivial example, and one that might not seem too consequential, but the
+code doubles with every new item added to the array. With very many elements,
+the duplication begins to seem unwarranted.
+
+Another option, of course, is to use an array of `Fn` instead of `fn`:
+
+```rust
+const foo: [&'static Fn(&mut u32); 4] = [
+  &|var: &mut u32| {},
+  &|var: &mut u32| *var += 1,
+  &|var: &mut u32| *var += 2,
+  &|var: &mut u32| *var += 3,
+];
+```
+
+And this seems to fix the problem. Unfortunately, however, looking closely one
+can see that because we use the `Fn` trait, an extra layer of indirection
+is added when attempting to run `foo[n](&mut bar)`.
+
+Rust must use dynamic dispatch because a closure is secretly a struct that
+contains references to captured variables, and the code within that closure
+must be able to access those references stored in the struct.
+
+In the above example, though, no variables are captured by the closures,
+so in theory nothing would stop the compiler from treating them as anonymous
+functions. By doing so, unnecessary indirection would be avoided. In situations
+where this function pointer array is particularly hot code, the optimization
+would be appreciated.
+
+# Detailed design
+[design]: #detailed-design
+
+In C++, non-capturing lambdas (the C++ equivalent of closures) "decay" into function pointers
+when they do not need to capture any variables. This is used, for example, to pass a lambda
+into a C function:
+
+```cpp
+void foo(void (*foobar)(void)) {
+    // impl
+}
+void bar() {
+    foo([]() { /* do something */ });
+}
+```
+
+With this proposal, rust users would be able to do the same:
+
+```rust
+fn foo(foobar: fn()) {
+    // impl
+}
+fn bar() {
+    foo(|| { /* do something */ });
+}
+```
+
+Using the examples within ["Motivation"](#motivation), the code array would
+be simplified to no performance detriment:
+
+```rust
+const foo: [fn(&mut u32); 4] = [
+  |var: &mut u32| {},
+  |var: &mut u32| *var += 1,
+  |var: &mut u32| *var += 2,
+  |var: &mut u32| *var += 3,
+];
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+To a rust user, there is no drawback to this new coercion from closures to `fn` types.
+
+The only drawback is that it would add some amount of complexity to the type system.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Anonymous function syntax
+
+With this alternative, rust users would be able to directly bind a function
+to a variable, without needing to give the function a name.
+
+```rust
+let foo = fn() { /* do something */ };
+foo();
+```
+
+```rust
+const foo: [fn(&mut u32); 4] = [
+  fn(var: &mut u32) {},
+  fn(var: &mut u32) { *var += 1 },
+  fn(var: &mut u32) { *var += 2 },
+  fn(var: &mut u32) { *var += 3 },
+];
+```
+
+This isn't ideal, however, because it would require giving new semantics
+to `fn` syntax.
+
+## Aggressive optimization
+
+This is possibly unrealistic, but an alternative would be to continue encouraging
+the use of closures with the `Fn` trait, but conduct heavy optimization to determine
+when the used closure is "trivial" and does not need indirection.
+
+Of course, this would probably significantly complicate the optimization process, and
+would have the detriment of not being easily verifiable by the programmer without
+checking the disassembly of their program.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
--- a/text/0000-closure-to-fn-coercion.md
+++ b/text/0000-closure-to-fn-coercion.md
@@ -6,8 +6,8 @@
 # Summary
 [summary]: #summary
 
-A non-capturing (that is, does not `Clone` or `move` any local variables) closure
-should be coercable to a function pointer (`fn`).
+A closure that does not move, borrow, or otherwise access (capture) local
+variables should be coercable to a function pointer (`fn`).
 
 # Motivation
 [motivation]: #motivation
--- a/text/0401-coercions.md
+++ b/text/0401-coercions.md
@@ -154,6 +154,9 @@ Coercion is allowed between the following types:
 
 * `&mut T` to `*mut T`
 
+* `T` to `fn` if `T` is a closure that does not capture any local variables
+  in its environment.
+
 * `T` to `U` if `T` implements `CoerceUnsized<U>` (see below) and `T = Foo<...>`
   and `U = Foo<...>` (for any `Foo`, when we get HKT I expect this could be a
   constraint on the `CoerceUnsized` trait, rather than being checked here)
@@ -338,7 +341,7 @@ and where unsize_kind(`T`) is the kind of the unsize info
 in `T` - the vtable for a trait definition (e.g. `fmt::Display` or
 `Iterator`, not `Iterator<Item=u8>`) or a length (or `()` if `T: Sized`).
 
-Note that lengths are not adjusted when casting raw slices - 
+Note that lengths are not adjusted when casting raw slices -
 `T: *const [u16] as *const [u8]` creates a slice that only includes
 half of the original memory.
 
@@ -440,5 +443,10 @@ These rules could be tweaked in any number of ways.
 Specifically for the DST custom coercions, the compiler could throw an error if
 it finds a user-supplied implementation of the `Unsize` trait, rather than
 silently ignoring them.
+
+# Amendments
+
+* Updated by [#1558](https://github.com/rust-lang/rfcs/pull/1558), which allows
+  coercions from a non-capturing closure to a function pointer.
 
 # Unresolved questions


archshift:
This is my first RFC — should I modify [0401-coercions](https://github.com/rust-lang/rfcs/blob/master/text/0401-coercions.md) as well?

bstrie:
I much prefer the version that doesn't introduce any new syntax.

How would this interact with the existing closure types? Would the following code compile?
```rust
fn main() {
    let x = |var: &mut u32| {};
    foo(&x);
    bar(x);
}

fn foo(_: &Fn(&mut u32)) {}

fn bar(_: fn(&mut u32)) {}
```
I suppose what this is asking is, is the type of `x` actually a function pointer that just so happens to be using the closure syntax, or is it a closure that just so happens to be coercible to a function pointer? If it's the former, then should we expect to have to use the following syntax, in the same way that trait objects do today?
```rust
fn main() {
    let x: fn(&mut u32) = |var| {};
    // alternatively:
    let x = |var| {} as fn(&mut u32);
    bar(x);
}

fn bar(_: fn(&mut u32)) {}
```

archshift:
Whether a closure's actual type becomes `fn` when it does not capture can theoretically be an implementation detail, because `fn` already satisfies the `Fn` trait anyway. This is, of course, barring the programmer taking any unsafe liberties with the structure of a closure object.

For example, this will compile and run fine:
```rust
fn foobar(_: &mut u32) {}

fn main() {
    let x = foobar;
    foo(&x);
    bar(x);
}

fn foo(_: &Fn(&mut u32)) {}

fn bar(_: fn(&mut u32)) {}
```

Strictly according to the RFP, though, the variable would be a closure that later coerces to a function pointer, which is probably for the better to prevent any breakage of silly unsafe code.

bstrie:
> the variable would be a closure that later coerces to a function pointer

Does the type system need some way to distinguish between normal closures and "bare" closures? Can we achieve all this without runtime checks?

archshift:
Yes, ideally the type system would keep track of closures that do not capture their environment.

Runtime checking is not a solution at all, in my opinion, nor is it necessary.

bstrie:
Am I right in that there would be no semantic difference between the existing `fn` type and a hypothetical "bare closure" type? If that's the case then it seems like the simplest route would just be to have Rust automatically assign zero-size closures a `fn` type (I had forgotten that `fn` satisfies `Fn`). But you say that this might be a problem for unsafe code, can you give an example of what sort of unsafe code you're worried would break under this scheme?

archshift:
After a quick check of rustc source, you're right. There really isn't any way to transmute a non-capturing closure (TIL it's zero-sized!), so no unsafe code could be broken by this.

In that case, there wouldn't be any downside after all to simply using a `fn` type for such "trivial" closures.

blaenk:
I remember I used to talk about this with cc @P1start over a year ago; I think @eddyb too, perhaps regarding bounds-targetting coercions? I don't remember.

It'd be nice to have in some form. I also would prefer it to be automatic if possible.

eddyb:
This can all be done very easily at the type level: allow coercions from closures with no captures (a fact which is recorded in their very type)[https://github.com/rust-lang/rust/blob/1f131fbe7799b9c3d4a775798b90760f7c11ef33/src/librustc/ty/sty.rs#L245-L248).

The coercion you want is called a "reification" coercion and it's what turns functions into `fn` pointers today (you can treat non-closure `fn`s as closures with no captures if you want to).

As for trans, well, the `fn` pointer is just `<typeof(closure) as FnOnce<_>>::call_once` (as zero-sized arguments are always ignored now), so that's a small bit of plumbing.

glaebhoerl:
FWIW I wouldn't mind having `fn` literals - independently of the fate of this RFC. Yeah it's niche and "more syntax", but the `fn` type not having a literal form is kind of a hole anyways.

eddyb:
@glaebhoerl FWIW there are no more items with `fn` types - they can only be produced by reification, which is what this PR would allow for closure expressions too.

gkoz:
It could be useful to specialize for non-capturing closures (together with usual `fn`s) when talking to APIs that don't understand thick pointers. Simply making `|| {}` coercible to `fn()` doesn't seem sufficient though, there needs to be a way to bound on this.

```
trait IntoRaw {
    /// converts into a thin pointer
    fn into_raw(self) -> *mut c_void;
}

impl<F: Fn()> IntoRaw for F {
    default fn into_raw(self) -> *mut c_void {
        // box twice to get a thin pointer
        let b: Box<Box<Fn()>> = Box::new(Box::new(self));
        Box::into_raw(b) as *mut _
    }
}

impl<F: Fn() + Coercible> IntoRaw for F {
    fn into_raw(self) -> *mut c_void {
        // even a closure's code is static right?
        self as fn() as *mut _
    }
}
```

withoutboats:
This RFC is hasn't seen much activity in a long time, so it'd be good to get it resolved. I'm mildly in favor of merging, but I have concerns that are enough that I haven't asked rfcbot to fcp it. Here are my thoughts.

### Breaking changes

My biggest concern is that this could make adding a captured variable to a closure a breaking change in a library; it would be very unfortunate for users to have to determine if subtle changes like that could break a downstream crate, rather than just always knowing that they're fine. To me, if this were the case, this RFC would be a no go.

Fortunately, I can't think of a way to expose a closure to clients except through the `impl Fn` mechanism. It seems obvious to me that an `impl Fn` should not coerce to an `fn`, even if the type it secretly is could coerce, but this should be made explicit in the RFC.

If there are no ways to expose values of the anonymous type to clients, then this should be fine. But I want to make sure we've thought about this.

### Accidental guarantees

A more subtle concern is one in which the user _accidentally_ guarantees this coercion, and then later is forced to break their API to perform the upgrade they intended.

This basically introduces two ways to return closures, one of which is strictly more general than the other:

```rust
fn incr() -> fn(i32) -> i32 {
    |x| x + 1
}

fn incr() -> impl Fn(i32) -> i32 {
    |x| x + 1
}
```

My concern is that the first of these is more natural to write than the second. Users could write functions which return `fn` types, not realizing that they are declaring in their API that the function they return captures no variable from the environment.

This seems a pretty serious trade off to me - we want the language to just enable you to do the right thing. The current behavior is unergonomic and confounding, but I'm not 100% sure that the footguns this behavior would introduce are less of a problem.

glaebhoerl:
@wycats So have `fn` literals instead? :grin:

withoutboats:
@glaebhoerl assuming that @ was at me :wink:, I think the downsides of a new lambda syntax are greater than the downsides of the current behavior or the RFC.

glaebhoerl:
Oh, sorry! And I made the opposite mistake just a couple of hours ago... (but thankfully twitter does the reply-tos automatically;)

arielb1:
> Fortunately, I can't think of a way to expose a closure to clients except through the `impl Fn` mechanism. It seems obvious to me that an `impl Fn` should not coerce to an `fn`, even if the type it secretly is could coerce, but this should be made explicit in the RFC.

`impl Fn` types are hidden just as much as generic type parameters.

archshift:
@withoutboats 
>It seems obvious to me that an impl Fn should not coerce to an fn, even if the type it secretly is could coerce, but this should be made explicit in the RFC.

For what it's worth, I agree, and I can add a comment to the RFC making that explicitly clear.

>A more subtle concern is one in which the user accidentally guarantees this coercion, and then later is forced to break their API to perform the upgrade they intended.

I agree that this could happen, and I'll add this to the drawbacks section of the RFC, but we do already have situations in which a Rust programmer might take the "easy route" and accidentally constrain their own API.

For [a contrived example](https://play.rust-lang.org/?gist=c74a3a5381440ea94eed59f60c0620a6&version=stable&backtrace=0), this is an easy mistake to make with `mut`. (Again, please note that this is a very contrived example.) I do think that to some extent, we expect a Rust programmer to be aware of the limitations they create for themselves.

In this case we're talking about a much more niche feature than `mut`, of course, but I think the benefits to code clarity and brevity outweigh some accidental misuse.

withoutboats:
@rfcbot fcp close

We discussed this at the lang meeting this week. We considered both the coercion option the RFC proposed and the alternative function literal syntax, as well as allowing only explicit `as` casts of closures into fn types, but ultimately we felt that there was not strong enough motivation to adopt any of these.

In general, folks felt most favorably about the function syntax (though speaking personally I think I'm more against it than other people were), but there were concerns about how it would work. If the function syntax infers your arguments (including the return type), it behaves significantly differently from function declarations. If it doesn't, it doesn't save you much in comparison to writing out a literal function.

No one loves the current situation, but we felt that the language as it exists right now does not strongly encourage you to use fn pointers, so its not clear how often some new syntax or coercion would actually be used for its intended purpose.

crumblingstatue:
> If the function syntax infers your types (especially the return type), it behaves significantly differently from named function declarations. If it doesn't, it doesn't save you much in comparison to writing out a named function.

I've been thinking about allowing inference for function-local items (Or at least fns, consts, and statics).
Since inference is done locally to a function, and these items are constrained to the function body they are located in, I see no obvious reason not to allow this. This could make a lot of local code-reuse situations a lot more convenient. (At times I've been using local macros instead of local fns, just so I can avoid having to type out the type signatures).

Of course this would require its own RFC. My point is, this could make the function literal alternative more appealing.

glaebhoerl:
> If the function syntax infers your types (especially the return type), it behaves significantly differently from named function declarations. 

I assume it would just infer everything the same way as closure syntax does, apart from not allowing captures. Is there anything wrong with that? That doesn't sound too surprising to me; in fact it's exactly what I would expect. My sense is that inference for `fn` items is a separate issue. In my mind, the reason why `fn` items require explicit types is because "global items require type annotations" (see also: `const`, `static`) rather than "`fn` types require type annotations".

Meta-level: What are the conditions under which this issue would be revisited? It doesn't seem like there's much scope for new information coming to light or different solutions being thought of. Do issues which aren't annoying *enough* just stay unresolved indefinitely? That sounds like it could lead to "death by a thousand cuts".

(N.B. I don't care that much about the feature itself here. More about the reasoning.)

crumblingstatue:
Assuming the following were meant as replies for [my comment](https://github.com/rust-lang/rfcs/pull/1558#issuecomment-266465834):

> My sense is that inference for fn items is a separate issue.

I do indeed state that this would require its own RFC. I just wanted to promote the idea that this would make inference for a function literal syntax less out of place.

> In my mind, the reason why fn items require explicit types is because "global items require type annotations" (see also: const, static) rather than "fns require type annotations".

Right. To emphasize, my idea is to allow type inference for items __local__ to the function body, as opposed to being global items.

I do not mean to derail this thread, I just want to clarify myself.

glaebhoerl:
It was more in response to @withoutboats (and by extension the lang team) who originally raised the inference issue.

> I just wanted to promote the idea that this would make inference for a function literal syntax less out of place.

I think it wouldn't have any impact on how out of place it would or wouldn't be because the two features are unrelated :o)

> To emphasize, my idea is to allow type inference for items local to the function body, as opposed to being global items.

Right, no misunderstanding there. I think inference for local and/or private items is a potentially reasonable idea, but is logically related to (doing likewise for) `const` and `static` items, and not to `fn` literals in expressions.

crumblingstatue:
> I think it wouldn't have any impact on how out of place it would or wouldn't be because the two features are unrelated :o)

The counter-argument(?) against allowing inference I replied to was "If the function syntax infers your types (especially the return type), it behaves significantly differently from _named function_ declarations."

If types were inferred for _named functions_ __inside__ fn bodies, it wouldn't behave differently from them.
And as far as I'm aware, both features could only appear __inside__ fn bodies, so they would be consistent with one another.

glaebhoerl:
I understand that. I disagree with the assumption that they need to be consistent with one another.

To make it really really clear:

@withoutboats and lang team: "The problem is froop."

You: "We can solve froop with brazzle."

Me: "I don't think froop is a problem."

aturon:
@glaebhoerl 

> I assume it would just infer everything the same way as closure syntax does, apart from not allowing captures. Is there anything wrong with that?

There's no problem in making it work; it's more a question of consistency/understandability.

With closures, we have very distinct expression and "type" forms (the latter using the `Fn*` traits). It's well understood that closure syntax involves a fair amount of machinery and inference, but one could never confuse it for a type signature.

OTOH, if we use a syntax like `fn(x) { x + 1 }` for function literals, it's much closer to the type form, but performs inference. That's potentially confusing -- though, in my opinion, not by itself the problem.

As @withoutboats mentioned, the place this really becomes a concern is with return types, when you start mixing in type information. The closure syntax allows you to optionally specify types for arguments and the return value. Consider:

```rust
|x: bool| !x
```

Even though we've annotated the argument type, because we didn't write a `->`, the return type is still inferred (as `bool`). Consider the same with `fn` syntax:

```rust
fn(x: bool) { !x }
```

That looks dangerously close to a standalone `fn` definition, **where the return type would be understood as `()`**. That's the key issue.

Is this concern an absolute blocker for the feature? Probably not. But it means that the literal syntax is not a slam dunk, and when combined with the somewhat narrow motivation for the feature in the first place, it didn't seem worthwhile to pursue the addition at this time.

> What are the conditions under which this issue would be revisited? It doesn't seem like there's much scope for new information coming to light or different solutions being thought of. Do issues which aren't annoying enough just stay unresolved indefinitely? That sounds like it could lead to "death by a thousand cuts".

IMO, the main thing that's likely to bring this issue to the fore is more evidence that the lack of such notation is painful. I understand the point about papercuts, but OTOH language additions are not free -- there's also the danger of reaching enormous complexity by a thousand small additions.

These considerations are never cut and dried, and we shouldn't expect them to be. But at the moment, given the somewhat fiddly problems with the design, and the lack of widespread pain here, we're leaning toward the "not now" angle.

glaebhoerl:
Thanks. That does sound like a legitimate issue. (Which could be finessed in various imperfect ways, which you've no doubt also thought of, such as requiring the return type to be present if the argument types are, or maybe just linting it.)

> OTOH language additions are not free -- there's also the danger of reaching enormous complexity by a thousand small additions

Indeed. For what it's worth, the (hand-wavy!) way I tend to think about this is that there's a big, imaginary matrix with a row and a column for every "language feature" (in some loose sense), and in each cell at the intersection of a row and a column, the interaction of those two features. In this metaphor, an empty cell means something "should be there" but isn't, and "filling in holes" is a good thing, as it is cleaning up irregularities in the language. On the other hand, adding new rows and columns is bad -- it's a completely new, separate feature for every other feature to potentially interact with. (I think I've heard team members refer to this as "multiplicative complexity".) I guess the common thread here is to minimize the (conceptual) descriptive complexity of the language, which both "negative" special cases (omissions) and "positive" features contribute to.

In case you haven't guessed, I feel like the fact that `fn` types lack a literal form is one of those negative special cases.

(Again: this is not to be taken as me pushing to have the decision changed - I feel like my questions have been answered - just explaining my reasoning.)

aturon:
> I guess the common thread here is to minimize the (conceptual) descriptive complexity of the language, which both "negative" special cases (omissions) and "positive" features contribute to.

Yes, I'm deeply sympathetic to this point of view. If the literal syntax was a more straightforward addition, I'd be more in favor of it for precisely this kind of reason (in conjunction with the fact that it *is* occasionally useful).

archshift:
>  I understand the point about papercuts, but OTOH language additions are not free -- there's also the danger of reaching enormous complexity by a thousand small additions.

Very true, and part of the reason that C++ has become such a monstrosity over time. I figured a feature like this would simply complement the "zero-cost abstraction" focus in Rust.

briansmith:
> I'm not 100% sure that the footguns this behavior would introduce are less of a problem.

The history of the evolution of closures in Rust has been that worries have tended to be overblown. I think that if somebody wants to implement the feature, it should be implemented and then we can experiment with it to see if these worries are really valid.

> the language as it exists right now does not strongly encourage you to use fn pointers

Rust does encourage one to use function pointers if one wants to avoid unnecessary code bloat due to monomorphization, and/or one otherwise wants to avoid forcing callers that forward function pointers to be parameterized. This is why I use function pointers heavily and avoid using closures when practical.

> If the function syntax infers your types (especially the return type), it behaves significantly differently from named function declarations. If it doesn't, it doesn't save you much in comparison to writing out a named function.

Regardless of whether the closure syntax is used or a function literal syntax is used, the feature does offer useful benefits:

* The definition of the function can be inline instead of out-of-line.
* When using macros or other code generation, there's no need to generate a dummy or confusing name for a function that wouldn't otherwise need a name.
* The syntax makes it clear that the function is only called through the function pointer, and never called directly by name, since it has no name.

> There's no problem in making it work; it's more a question of consistency/understandability.

When learning Rust, I found it difficult to understand why a closure anything can't be used where a function pointer is needed.

Regarding the complexity arguments, I feel that, while adding implicit coercion from non-capturing lambda expressions to function pointers technically adds complexity to the implementation, it simplifies the language conceptually. Conceptually, now we can't pass a closure where a function pointer is expected only because the closure might have captured something. But, it's obvious that a closure that doesn't capture anything...doesn't capture anything, so there's no conceptual basis for the limitation in that case.

So, all in all, I think it would be a worthwhile improvement to have non-capturing closures be coercible to function pointers and usable in every context where a function name could be used to reference a named function.

ubsan:
I'd like to give an irc conversation from #rust-beginners as evidence for why this should be merged:

```
sadiq | quick question
sadiq | why does https://play.rust-lang.org/?gist=f42be222188cd1e160a777de64d7c4df&version=nightly&backtrace=0 not compile?
sadiq | but https://play.rust-lang.org/?gist=2f32e5953b7a1cf1ed4cea20030f024c&version=nightly&backtrace=0 does?
sadiq | (as in, why is that closure not Copy?)
ubsandroid | sadiq: because closures can't be clone
ubsandroid | and we didn't want even more ices
sadiq | ubsandroid, oh, I see.
ubsandroid | sadiq: basically, lack of good generics
sadiq | kinda weird that functions are Clone though?
ubsandroid | sadiq: I don't think they are?
sadiq | well, that second playground compiles
ubsandroid | might be up to 8 that are
ubsandroid | or fifteen
sadiq | and there's Copy on the bounds
ubsandroid | playbot: fn foo(_: (), _: (), _: (),_: (), _: (), _: (),_: (), _: (), _: (),_: (), _: (), _: (),_: (), _: (), _: (), _: ()) {} fn check_clone<T: Clone>(_: &T) {} check_clone(&foo)
-playbot- | error[E0277]: the trait bound `fn((), (), (), (), (), (), (), (), (), (), (), (), (), (), (), ()) {main::foo}: std::clone::Clone` is not satisfied
ubsandroid | sadiq: ^
sadiq | ubsandroid, so up to 15 parameters are?
ubsandroid | sadiq: or 8, don't remember and also on my phone
```
*sadiq | guess I'm just not sure what the difference is between the function and closure playgrounds from ^ were, given the closure didn't capture anything from it's environment.*
```
ubsandroid | sadiq: because functions and closures not capturing anything are different, for now
```

Basically, there is confusion leading from the fact that closures that don't capture anything are different from function literals, *and* that Clone/Copy are not implemented for closures.

ubsan:
I've been hearing about this a lot on IRC lately; people do get weirded out by the fact that there's no anonymous lambda syntax:

```
bollu  | so rust has closures, but not "pure lambda" syntax?
ubsan  | bollu: no, but there's an RFC for something like that
Xion   | You mean closures without the closure part?
Ms2ger | What would a "pure lambda" syntax be?
ubsan  | until then, you can write let x = {fn tmp(...) { ... } tmp};
Xion   | I.e. without any environment.
bollu  | Xion: yes
ubsan  |Ms2ger: I assume closure syntax without captures
Ms2ger | So, a function?
bollu  | Xion: just an annonymous function
bollu  | Ms2ger: "annonymous" is what I
bollu  | I am interested in :)
```

ubsan:
So, I'd like to editorialize, from my experiences with people, talking on IRC.

1) This is a feature that can pretty easily be worked around, in a fairly ugly way.

2) People want to use the pretty syntax for non-capturing closures, and it's weird for them that they can't.

3) It's unlikely to, in practice, cause breaking changes when someone returns a `fn()` and needs to return an `impl Fn()`: the beginners likely to make this mistake will mostly be writing binaries or single-user libraries, and by the time a crate gets used enough for it to matter, whether a closure closes over anything is likely to be solved.

4) It's a simple feature to implement, and it fixes something that's missing in the language which confuses people.

aturon:
@ubsan Thanks very much for the input from people learning the language; that's extremely valuable!

Just to clarify: while some of the discussion is about anonymous function syntax, this RFC proposes that the existing closure syntax do double-duty as anonymous function syntax as well. That's different from other primitive types, which have dedicated literal syntaxes; one might expect a literal anonymous `fn` syntax (though there are some [problems with that](https://github.com/rust-lang/rfcs/pull/1558#issuecomment-266802192)).

Can you say a bit more about how you see the tradeoffs around that syntactic choice and the learnability issues?

withoutboats:
I'd like to know more about the ways people are using `fn` pointers right now. Are these situations that we'd say they should be using `impl Fn` if it were stable, or something where `impl Fn` is unsatisfactory?

ubsan:
@aturon I don't feel that there is a significant learnability issue with either; this is not something that's hard to grasp, either with dedicated syntax, or with reusing closure syntax. However, adding new syntax means discoverability issues; it's not a big deal though.

@withoutboats Generally (nits about small sample size aside), people seem to not be returning closures or functions, at least to me. sadiq was interested in something they could copy, i.e., pass to a function that takes `T: FnOnce` and continue to use; bollu was interested in why that syntax doesn't exist, more than any real problem. I am more saying that the overload of the syntax is not confusing to people, and it's at least less confusing than `{ fn tmp

elahn:
Since fn() coerces to Fn(), it's surprising it doesn't go in the opposite direction when nothing's captured.
```rust
fn use_closure<F>(closure: F) where F: Fn(char) -> bool {
    println!("{}", closure('a'));
}

fn main() {
    use_closure(char::is_uppercase);
}
```
https://play.rust-lang.org/?gist=dbe0707259def829c778ee9ba9526ea1&version=stable&backtrace=0

aturon:
In re-reading this thread, I realized that the earlier summary of the lang team discussion was missing some important bits. To try to recap:

You can imagine essentially two ways of solving the problem this RFC is trying to solve. Either creating a dedicated `fn`-literal syntax, or somehow re-using the closure syntax.

- A dedicated `fn`-literal syntax is more consistent with other primitive types, which are equipped with their own literal syntax, but [has problems with type signatures](https://github.com/rust-lang/rfcs/pull/1558#issuecomment-266802192). These may be surmountable problems, but would need some design work.

- Allowing closure syntax to create/coerce to `fn` types, OTOH, suffers from the problems that affect coercions in general: they don't always work and can be a bit fiddly. Common examples include calling generic functions, where the arguments would work if a coercion fired, but the fact that the receiving type is generic means the coercions don't fire.

In thinking about this again, I'm not sure how much the finickiness of coercions would matter; I think it'd be pretty unusual to need a coercion to `fn` to fire in a generic context. It'd be helpful to try to think through some realistic examples where this falls down.

There was another, more vague concern, which is that it's somewhat odd to have a primitive type like `fn` that doesn't have its own literal syntax, but for which we have some *other* special syntax which happens to coerce into that type. A dedicated literal syntax feels more clean and consistent (but suffers from the problem mentioned above). But it's hard to know how much this "inconsistency" would matter in practice, for learnability.

nikomatsakis:
I agree, @aturon, with the sentiment that a dedicated `fn` syntax feels more principled, but in practice has flaws, and I'm not sure how much it will matter. As long as this coercion is limited to using the expected type, this change feels fairly harmless. It's ultimately a kind of niche thing, and the explicit syntax of making a named fn and referencing that name isn't so bad.

arielb1:
@aturon

Function *pointer* types don't really have a literal syntax *at all*. Function literals create fn *items*, which you have to coerce to a function pointer anyway.

In any case a closure is quite like an fn item, and therefore should be similarly coercible to an fn pointer.

aturon:
@arielb1 Yes, I take your point. The `fn` type is a sort of weird hybrid: it looks like a primitive type, but it works a bit more like a trait object.

So in particular, if we had an anonymous `fn` syntax, it'd be a bit of a lie, since it would actually work more like closures under the hood (adding an anonymous `fn` item that can be *coerced* to the `fn` type). That's a pretty decent argument that closure syntax is in fact a good fit.

nikomatsakis:
> That's a pretty decent argument that closure syntax is in fact a good fit.

Agreed.

glaebhoerl:
I think that would be a compelling argument if we had actually gone the route of making `fn`s look like trait objects, by making them unsized types and having `&'static fn` in place of the current `fn`, and so on, as some RFC proposed to do at the time. I don't think it's as compelling in Rust-as-it-is, where `fn` *is* a primitive function pointer type (and `fn` items are weird as stated, but that's kind of neither here nor there to me).

> since it would actually work more like closures under the hood (adding an anonymous `fn` item that can be *coerced* to the `fn` type)

*Would* it work that way? I mean, the choice is up to us.

Ericson2314:
@glaebhoerl https://github.com/rust-lang/rfcs/issues/1037 I want to resurrect that at some point---one of those is my RFC if you remember :).

aturon:
@glaebhoerl 

> *Would* it work that way? I mean, the choice is up to us.

I want to make sure we're not talking past each other, so let me be concrete. Today, `fn` items implicitly declare a unique *type* which implements the closure traits, but which can be *coerced* to the `fn` type. This allow for both static and dynamic dispatch:

```rust
fn static_dispatch<F: FnOnce()>(f: F) { f() }
fn dynamic_dispatch(f: fn()) { f() }

fn myfn() {}

fn main() {
    static_dispatch(myfn);
    dynamic_dispatch(myfn);
}
```

and note that the static dispatch relies in an essential way on the `fn` item producing a unique type -- exactly the same as with closures.

What I'm saying is that if we added an anonymous `fn` literal syntax, I would expect it to work the same way:

```rust
fn static_dispatch<F: FnOnce()>(f: F) { f() }
fn dynamic_dispatch(f: fn()) { f() }

fn main() {
    let myfn = fn() {};
    static_dispatch(myfn);
    dynamic_dispatch(myfn);
}
```

Getting the same behavior here requires the `fn() {}` definition to create a fresh type, so that we can statically dispatch. While we wouldn't *have* to reproduce that behavior, it would be a very odd choice not to!

My follow-up point was that, for anonymous `fn() {}` syntax to work this way, it would behave in most respects identically to closures; they would produce values of some anonymous type, *not* of type `fn` as the syntax would suggest. The only difference being that the resulting object can coerce to the `fn` type.

At that point, I've convinced myself that the RFC as written is the best option: just allow closures to coerce directly.

glaebhoerl:
Yeah, I understand all that completely. Thanks for making sure though!

> While we wouldn't have to reproduce that behavior, it would be a very odd choice not to!

This is the assumption I don't have in common. Since `fn()` types are the type of function pointers, it's completely natural to me that `fn()` literals would be function pointers. And `fn` items are special. Which they are in other ways besides, I believe. Shrug.

It's not like the closure-coercion approach doesn't have advantages. It avoids adding a new syntactic form for people to encounter and to have to learn about, and it'll be more familiar for people coming from C++. I think those are the major points, and what we're discussing right now is small potatoes down in the weeds, relatively speaking.

aturon:
@withoutboats Given the post-FCP discussion, I think we should take the RFC back out of FCP to close and reconsider it. Can you cancel the FCP?

Ericson2314:
@aturon 

Out of curiosity:
```rust
fn static_dispatch<F: FnOnce()>(f: F) { f() }

fn main() {
    fn foo() {};
    let myfn = foo;
    static_dispatch(myfn);
}
```
Is this actually static? Or would this be function -> function pointer which impls closure trait?

Ericson2314:
Also, did anyone bring up capture lists yet? If we had (optional) explicit capture lists, then we could only allow closures with an explicit empty capture list to convert to function pointers. This makes sure that people really opt-in to what's going on.

glaebhoerl:
> Since `fn()` types are the type of function pointers, it's completely natural to me that `fn()` literals would be function pointers.

Let me elaborate. I suspect the reason it feels like "it would be a very odd choice" to do this is due to the intuition that dynamic dispatch in Rust is always opt-in. Wherever possible, Rust tries to dispatch calls statically, and to steer users in that direction as well. So, this is a heuristic. And as far as heuristics go, it's quite a strong one!

But I don't think it leads to an accurate conclusion in this particular case. The use case for `fn()` literals is to conveniently define a function in-place where a value of `fn()` type is expected. Now, the worry might be that if `fn()` literals are always *actually* function pointers, we might inadvertently lose out on opportunities for static dispatch in statically monomorphized contexts -- `fn higher_order<F: Fn()>(f: F)`, and the like. But: there is no actual motivation to use an `fn()` literal in such a context! The normal `||` literals work perfectly. 

The situation where this *would* actually matter is if it were also possible to use `fn()` types as trait bounds -- that is, something like `fn higher_order<F: fn()>(f: F)`. In that case, the `||` literal wouldn't work, and the `fn()` literal would leave performance on the table. *But that situation doesn't exist!* (And I don't believe it would offer any benefits over `F: Fn()` if it did.) The only actually-existing situation where `fn()` literals are relevant is when you want to create a value of an actual, honest-to-God `fn()` pointer type, and in that situation, the fact that an `fn()` literal is *in fact* a function pointer could not be any less surprising. Having `fn()` literals be anonymous types which need an intervening coercion to turn into `fn()` pointers would be a semantic pessimization here, by impairing type inference.

glaebhoerl:
Put another way, `fn` items (statically-dispatched values of anonymous zero-sized types) and `fn` types (dynamically-dispatched values of a concrete non-zero-sized type) are inconsistent with each other *already*, in the language as it is, and there is no reason a priori why we should have to choose for `fn` literals to be consistent with the former rather than with the latter.

nikomatsakis:
@glaebhoerl 

> Put another way, fn items (statically-dispatched values of anonymous zero-sized types) and fn types (dynamically-dispatched values of a concrete non-zero-sized type) are inconsistent with each other already, in the language as it is, and there is no reason a priori why we should have to choose for fn literals to be consistent with the former rather than with the latter.

While I agree, I think it would be the wrong choice for `fn` literals to be consistent with the types and not the items. Here are some reasons:

- Suppose there is an API that is written (today) to take a `fn(...)` but later generalized to take closures. If I am calling it like `foo.(fn() { ... })`, then I would now have less efficient behavior. 
    - The same benefit is true if we use the `||` form, of course, and it avoids the discomfort around the return type.
- To me, there is actually a greater affinity between the item and expression form than between the expression form and the type. It's kind of hard for me to say just why that is -- I guess because having a named and anonymous for that act the same way seems to make sense.

Anyway, I'm pretty convinced by now. I am in favor of this RFC as originally written. Moreover, I could imagine that sometime later we might consider generalizing this to say that "any zero-sized type that implements `Fn` can be converted into a `fn` pointer", but I think that doing that would require us to reify some form of predicate for "zero-sized" (so that you can e.g. write a fn like `fn foo<T: ZeroSized + Fn()>(t: T) -> fn() { t }`).

aturon:
> Moreover, I could imagine that sometime later we might consider generalizing this to say that "any zero-sized type that implements Fn can be converted into a fn pointer"

That's a nice idea.

@archshift, it'd be great if the RFC could be updated to take into account some of this recent discussion, as well as logging some of the downsides we've found with `fn` literal syntax.

text/0000-closure-to-fn-coercion.md
```diff
@@ -0,0 +1,222 @@
...
+unsized closures to `fn` types. The implementation of this is simplified by the
+fact that closures' capture information is recorded on the type-level.
+
+*Note:* once explicitly assigned to an `Fn` trait, the closure can no longer be
+coerced into `fn`, even if it has no captures.
+
+```rust
+let a: &Fn(u32) -> u32 = |foo: u32| { foo + 1 };
+let b: fn(u32) -> u32 = *a; // Can't re-coerce
+```
```
nikomatsakis: This detailed design doesn't seem to mention how this fits into Rust's type checker. My expectation is that it uses the 'expected type' -- meaning that it's a kind of coercion. Basically, when we type-check a closure expression, if the expected type is a `fn()`, we will coerce to a fn pointer, but otherwise we will not. 

This implies the usual limitations that come along with coercions (at least today). These are somewhat stronger than what you mention. For example, once a closure is assigned to a variable, it will no longer be possible to coerce it to a `fn()` pointer:

```rust
fn foo(x: fn());

foo(|| ..); // OK

let x = || ...;
foo(x); // ERROR
```

aturon: Is that stronger limitation fundamental? It seems to me that we could introduce a marker trait, or some other way of knowing that a particular struct is coercible to a `fn` pointer.

withoutboats:
@rfcbot fcp cancel

I hate to say I'm still inclined against this RFC though. I don't feel like function literals are a missing syntax; I feel like closures *are* our function literals. I am more of a fan of the coercion, but I am concerned about the ability to get yourself into a bind with it where you need to add some environment to your function and you can't.

text/0000-closure-to-fn-coercion.md
eddyb: That... limitation strikes me as false. What is the difference between the type of `|| ...` and the type of `x`? The coercion goes from `TyClosure` to `TyFnPtr`, doesn't it?

nikomatsakis:
@withoutboats 

> I am concerned about the ability to get yourself into a bind with it where you need to add some environment to your function and you can't.

Hmm, it seems like this is fundamentally a judgement call. There are definitely legitimate cases where you **actively want** a function pointer and not a closure (e.g., interfacing with C code, as well as wanting to guarantee that there is no environment, though that latter case could be achieved by adding some kind of "zero-sized" trait). But making those cases more ergonomic comes at the cost of making it easier to write the wrong thing in your public APIs.

I have to admit I am not so worried about that. I think beginners probably *will* make mistakes, but as they learn rust, they will learn the better fix, and their APIs will improve and be updated. It seems likely that many early APIs also make dubious choices about whether to take ownership of something or to borrow it, and so forth, as well.

aturon:
@nikomatsakis I agree. I'd also note that `fn` as a type is not something that people get exposed to early on in learning Rust, so I'm not sure how much temptation there would be to use it in APIs before one understands the implications.

I'm going to go ahead and move toward *acceptance* of this RFC, given recent discussion, but of course @withoutboats and others should feel free to raise concerns.

@rfcbot fcp merge

text/0000-closure-to-fn-coercion.md
archshift: Should I make it explicit in this RFC that it describes a coercion between the closure and fn-pointer *types*, rather than an implicit conversion of closure *expressions* to anonymous function pointers?

archshift:
@withoutboats 
Regarding accidentally trapping oneself API-wise, I believe distinguishing between a `fn()` and a closure is akin to distinguishing between `Fn`/`FnMut`/`FnOnce`. (The traits, I believe, are much easier to misuse as well.)

As long as we sufficiently emphasize the fact that function pointers are *completely* self-contained, I don't anticipate this to be a problem.

aturon:
Ping @nrc @pnkfelix @withoutboats

withoutboats:
I've marked this reviewed to merge. I think my biggest concern has to do with stabilizing this before `impl Trait` - people might start moving to `fn` pointers because they don't want to use `Box<Fn>`.

pnkfelix:
I have marked this reviewed to merge, but I have to admit that I was confused when re-reading the Summary sentence:

> A non-capturing (that is, does not `Clone` or `move` any local variables) closure
should be coercable to a function pointer (`fn`).

What's up with that definition of "non-capturing"? I would have thought the definition would be broader, to include *referencing* local variables (even without moving nor cloning them).

(And I'm pretty sure that that broader definition of "non-capturing" must be what everyone else is using, because the only way the proposal makes sense is for the generated code to not have any dynamic environment that it needs to access...)

eddyb:
Yeah, "capture by reference" is pretty standard at this point, the parenthesis is the confusing bit there IMO.

archshift:
Ah, that was an oversight. I'll fix the summary.

aturon:
The final comment period has elapsed without any additional commentary, so at this point we're pretty much ready to merge the RFC.

However, I noticed that @nikomatsakis has an [outstanding review request](https://github.com/rust-lang/rfcs/pull/1558#discussion_r95470740) and I want to make sure he feels the issue has been resolved at this point.

nikomatsakis:
@aturon @eddyb 

> However, I noticed that @nikomatsakis has an outstanding review request and I want to make sure he feels the issue has been resolved at this point.

Ah, hmm, yes I forgot about that. So my concern was that I would expect this coercion (from closure to fn ptr) to be driven by the top-down coercion process. I wrote this:

> My expectation is that it uses the 'expected type' -- meaning that it's a kind of coercion. Basically, when we type-check a closure expression, if the expected type is a fn(), we will coerce to a fn pointer, but otherwise we will not. This implies the usual limitations that come along with coercions (at least today). These are somewhat stronger than what you mention. For example, once a closure is assigned to a variable, it will no longer be possible to coerce it to a fn() pointer.

And @aturon replied: 

> Is that stronger limitation fundamental? It seems to me that we could introduce a marker trait, or some other way of knowing that a particular struct is coercible to a fn pointer.

Then @eddyb wrote:

@eddyb also wrote:

> That... limitation strikes me as false. What is the difference between the type of || ... and the type of x? The coercion goes from TyClosure to TyFnPtr, doesn't it?

I suppose that thinking about it more there is another alternative. We could consider this a variant of upvar inference (the thing that decides, based on how it uses its upvars, when a closure implements `Fn`, `FnMut`, or `FnOnce`). Basically if a local closure type (defined in this fn, which at the moment is always the case I think) is coerced to a fn ptr, then we add a constraint that it can have **no** upvars.

I'd be ok with that, I guess, though I'm not totally psyched about it, just because I feel like it's yet another ad-hoc bit of obligation tracking we'd be doing in the type-checker (rather than having it all consolidated into a new-and-improved trait system sort of thing). Just examining the expected type at the time when we create the closure feels way easier. 

But yeah I can go either way I guess. It's fine.

eddyb:
> then we add a constraint that it can have **no** upvars.

The set of "upvars" is already known before type-checking begins, so that shouldn't be a problem.

aturon:
**Per post-FCP discussion, this RFC has been merged!**

[Tracking issue](https://github.com/rust-lang/rust/issues/39817)

AaronKutch:
The "Rendered" link is broken

shepmaster:
Accepted RFCs are available at https://rust-lang.github.io/rfcs/1558-closure-to-fn-coercion.html

PR #1574, opened 2016-04-06
More api documentation conventions

steveklabnik:
[rendered](https://github.com/steveklabnik/rfcs/blob/more-api-conventions/text/0000-more-api-documentation-conventions.md)


text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+[english]: #english
+
+This section applies to `rustc` and the standard library.
+
+All documentation is standardized on American English, with regards to
+spelling, grammar, and punctuation conventions. Language changes over time,
+so this doesn’t mean that there is always a correct answer to every grammar
+question, but there is often some kind of formal consensus.
+
+One specific rule that comes up often: when quoting something for emphasis,
```
nrc: Is quoting for emphasis something that people do deliberately? I thought it was a mistake. Maybe an english/american thing. Anyway, can we recommend using markdown punctuation for emphasis? Or are we targeting an ascii world? I.e., `*foo*`. And then just quotes for quoting?

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+/// ```
+/// use op;
+///
+/// let s = "foo";
+/// let answer = op::compare(s, |a| a.chars().is_whitespace().all());
+/// ```
+```
+
+For particularly simple APIs, still say “Examples” and “Basic usage:” for
+consistency’s sake.
```
tshepang: I do not like this practice for it is adds noise. Including "Basic usage" is sensible only when there is another example showing some other usage. Also, most of the examples (at least in the stdlib) are simple, so the consistency argument falls away.

jimmycuadra:
I originally used the section headings (e.g. examples, failures, etc.) in my docs but stopped doing it because of how hard it is to read with the current styling of rustdoc pages. The headlines are way too large, and the underline beneath them makes it very hard to see the separations between functions, as the eye is drawn to the underline instead. I'm not sure whether or not it makes sense to take the current usability issues of rustdoc into account when considering this RFC. The idea on its own is good, but I'd hesitate to make the recommendation until rustdoc is easier to read.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+## Form
+[form]: #form
+
+These conventions relate to the formatting of the documentation, how they
+appear in source code.
+
+### Use line comments
+[use-line-comments]: #use-line-comments
+
+Avoid block comments. Use line comments instead:
```
zetok: You got that one wrong.

It should be preferable to avoid line comments, and use block comments whenever possible.

Case: line comments aren't that easy to work with for blind people, like block comments are.
https://github.com/rust-lang/rfcs/pull/1373#issuecomment-205864317

Also note that block comments should be
```
/*
    Some
    multiline
    comment.
*/
```
with no unnecessary asterisks.

steveklabnik: This particular convention has already been in place for years, and is in the original RFC.

zetok: >This particular convention has already been in place for years, and is in the original RFC.

Yes, I know. Just because something has been in place "for years" doesn't mean that it's right.

I just wanted to highlight that this RFC makes same mistake as the original one – hopefully there'll be interest to make corrections :)

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+[link-all-the-things]: #link-all-the-things
+
+A major drawback of Markdown is that it cannot automatically link types in API documentation.
+Do this yourself with the reference-style syntax, for ease of reading:
+
+```
+/// The [`String`] passed in lorum ipsum...
+///
+/// [`String`]: ../string/struct.String.html
+```
```
cmbrandenburg: Manual linking is a lot of work that the computer could be automating for me. I understand that automatic linking is tricky, but manual linking leads to inconsistencies, broken links, omissions, etc. Are there any plans to use automatic linking in the future?

steveklabnik: The issue is that Markdown has no automatic linking. If we can get rustdoc to use commonmark, and commonmark has a way of doing extensions, then that's a path forward, but it's unclear when that will happen.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Summary sentence
+[summary-sentence]: #summary-sentence
+
+In API documentation, the first line should be a single-line short sentence
+providing a summary of the code. This line is used as a summary description
+throughout Rustdoc’s output, so it’s a good idea to keep it short.
+
+The summary line should be written in third person singular present indicative
+form. Basically, this means write “Returns” instead of “Return.”
```
cmbrandenburg: This might be a little off-topic—and maybe nit-picky—but here goes.

One thing that bugs me about Rust documentation—including the documentation I write for my crate—is that the summary line looks ugly. In a module's documentation page, where the types, functions, methods, etc. are neatly aligned in tables, the summary line looks best as a fragment. E.g.,

<table>
<tr><td><code>Foo</code></td><td>Manages a foobar thingamabob.</td></tr>
<tr><td><code>Qux</code></td><td>Wraps a blah blah blah.</td></tr>
</table>

But when I navigate to the documentation page for `Foo` , the same fragment looks _abrupt_ because it's in prose. Instead, I would like the summary to be a complete sentence starting with a noun. E.g., <q>`Foo` manages a foobar thingamabob.</q> 

I wish Rust automatically formatted this for me. That is, I wish I could write a complete sentence for the summary line and have `rustdoc` leave out the noun when generating the one-liner for the module-level table. I suspect `rustdoc` wouldn't need to be very smart to do this. It could recognize that the sentence starts with <q>Foo</q> and, because `Foo` is the thing being documented, elide it from the output and capitalize the following word. E.g., <q>`Foo` manages</q> becomes <q>Manages.</q>

In addition to looking prettier, I think this would also encourage documentation authors to write better summary lines.

text/0000-more-api-documentation-conventions.md
cmbrandenburg: What about `rustdoc` doing some kind of pre-processing on the Markdown to look for things that can be linked? Everyone ends up inventing their own Markdown flavor :).

Another question: Is it possible to define all link targets in one place in my crate and reference those links from different source files? E.g., with `src/lib.rs`:

```
[`Foo`]: struct.Foo.html
```

And then in `src/bar.rs`:

```
Here's my link to [`Foo`], even from another file.
```

I ask this because I don't see myself providing manual links if I must redefine the same links in many different source files. But if I can list all my links in one place, it's at least tractable.

Centril:
@steveklabnik As discussed in IRC, I think the part about **American English** should be **clarified** so that it states that this is preferred/required in the **standard library** (and cargo) and not in *community libraries*.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+* Panics
+* Errors
+* Safety
+* Aborts
+* Undefined Behavior
+
+Even if you only include one example, use the plural form: ‘Examples’ rather
+than ‘Example’. Future tooling is easier this way.
+
+Use graves (`) to denote a code fragment within a sentence.
```
robinst: I think referring to \` as "backticks" is more common than "graves", and also matches the terminology that both [Markdown](http://daringfireball.net/projects/markdown/syntax#code) and [CommonMark](http://spec.commonmark.org/0.25/#code-spans) use.

text/0000-more-api-documentation-conventions.md
robinst: Maybe follow the style guide also in this RFC and use `‘Return’.` instead of `“Return.”`?

text/0000-more-api-documentation-conventions.md
mitaa: > Another question: Is it possible to define all link targets in one place in my crate and reference those links from different source files? E.g., with src/lib.rs:

No, the references need to be in the same Markdown block as they're used in.

starkat99: I think this should be solved in rustdoc tool rather than manually. For instance, it should be trivial in a preprocessing step, assuming there's an index of symbols, to find all `[]` links that do not have a corresponding reference link and look them up in the symbol index for the proper URL.

text/0000-more-api-documentation-conventions.md
starkat99: I agree, accessibility should be enough motivation to change the standard here.

text/0000-more-api-documentation-conventions.md
steveklabnik: > I think this should be solved in rustdoc tool rather than manually.

That's something that we can hopefully do someday, but I don't want future rustdoc features blocking useful work today.

text/0000-more-api-documentation-conventions.md
nixpulvis: I do think it should be a convention to give descriptions to each example *if* there is more than one.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+
+For particularly simple APIs, still say “Examples” and “Basic usage:” for
+consistency’s sake.
+
+### Referring to types
+[referring-to-types]: #referring-to-types
+
+When talking about a type, use its full name. In other words, if the type is generic,
+say `Option<T>`, not `Option`. An exception to this is lengthy bounds. Write `Cow<'a, B>`
+rather than `Cow<'a, B> where B: 'a + ToOwned + ?Sized`.
```
nixpulvis: Where exactly do you draw the line for lengthy?

GuillaumeGomez: I guess it'll be "on case".

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: Agreed.

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: > Also note that block comments should be
>
> ```rust
> /*
>     Some
>     multiline
>     comment.
> */
> ```
>
> with no unnecessary asterisks.

I disagree. I guess it's a personal opinion but I really don't like this writing.

```rust
/*
 * Some
 * multiline
 * comment.
 */
```

Is more pleasant to read. Or:

```Rust
// Some
// multiline
// comment
```

(which is my favourite).

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: @tshepang: It's for consistency. I don't mind having to write it every time, even if there is only one example. People will know the pattern and just get faster to what they're looking for.

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: So for now, let's continue this way and maybe add a processus to check if urls are valid as a test while waiting for rustdoc to generate such urls by himself.

text/0000-more-api-documentation-conventions.md
eternaleye: @GuillaumeGomez That just revives the exact reason @camlorn gave for [preferring block comments for a11y reasons](https://github.com/rust-lang/rfcs/pull/1373#issuecomment-205864317) - specifically, that screen readers then pepper their output with meaningless noise, obscuring the comment itself.

It may be more pleasant to read for you, but I find the two about equal, and for people with a11y concerns your proposal is just as bad as line comments.

camlorn:
Sorry, but I can't work out how to reply to inline comments to the diff.  I'm not actually sure that's possible, but GitHub's interface has enough accessibility problems that I'm not sure.

As long as this remains a recommendation and I'm free to ignore it in my own projects, I'm more indifferent than you'd think.  If I thought we could make the guideline say use block comments, then I'd be all over that.  But people are going to just use line comments all over anyway.  If the response to suggesting line comments for documentation is more than borderline in favor, I'm not going to come out of the woodwork later and try to make us change it.  I do similar ignoring on Python's PEP8.

I'm in favor of removing it or even changing it to block commenting, but I thought it was worth clarifying that I'm not going to be unreasonable about us having it.  My issue with the other one is that it essentially wanted to force it at a compiler level, which means that not only is it a guideline, it's one I can't just ignore for justified reasons in personal projects.  I might end up contributing to the standard library and would be thrilled if it started using block comments as this would involve writing documentation, but I am realistic enough to realize that line comments are overwhelmingly preferred for this purpose and that I have rather unique preferences.  Suffice it to say my commenting style is inconsistent.  There are rules, but I'm not sure I could explain them.

Also, as a sidenote, I'm glad that people actually remember my comments.  The responses to "Hi I'm a blind programmer and here are my problems" in this community have been more positive than every other time I've tried to have conversations about accessibility with people who aren't the devs of my screen reader.  Usually you're just completely ignored as a minority.

eternaleye:
@camlorn Regarding replies to comments on diffs, GitHub encloses diff-comment threads in a `div` with the `discussion-item` class, which contains (near the end, and several layers of nesting inwards) a button with the text "Add a line note" and the class `js-toggle-inline-comment-form`. That button enables a comment editor for that thread, and it may be possible to have a userscript that iterates over the page and enables all such editors.

If that doesn't work, it may also be possible to reply to the appropriate email and have it be placed appropriately, but I'm not especially optimistic.

text/0000-more-api-documentation-conventions.md
peschkaj: "Lengthy" would be, for me, when absolute correctness gets in the way of communicating clearly. This goes along with why we frequently `unwrap()` in documentation, even though we all why we shouldn't do it in production.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+
+This will highlight syntax in places that do not default to ‘rust’, like GitHub.
+
+Rustdoc is able to test all Rust examples embedded inside of documentation, so
+it’s important to mark what is not Rust so your tests don’t fail.
+
+References and citation should be linked ‘reference style.’ Prefer
+
+```
+[Rust website]
```
nikomatsakis: Nit: I would recommend `[Rust website][]`, since this avoids ambiguity in the case of two consecutive comments (like, `[Rust][] [website][]`). 

One question that I had was whether this was intentionally recommending *against* using a shortened thunk like `[Rust website][rw]`? (I don't think we should take a stance about that; let people decide for themselves based on how long the link text is.)

text/0000-more-api-documentation-conventions.md
nikomatsakis: Wait, would you expect to *ever* write bounds? I would not.

text/0000-more-api-documentation-conventions.md
steveklabnik: Yes, it is something that's done deliberately, though I am not aware if it's a regionalism or not.

I've often seemed bolding used when showing off a term for the first time, or to highlight things for other reasons. Maybe this is too specific, though. Right now, all the docs use quoting for emphasis, and use bold very sparingly.

text/0000-more-api-documentation-conventions.md
steveklabnik: I personally feel that it looks much uglier, but you're right that it can remove ambiguity. However, I don't think that putting two links side by side like that is generally a good idea, because then it looks visually ambiguous when rendered; it all looks like one big link.

And yeah, my intention wasn't to ever say never. I'll modify it to make this clearer.

steveklabnik:
I've pushed an update to the structure of the RFC, though it's not quite 100% done yet. Rather than make the whole RFC a new text, I'm going to make the full text an appendix, and only include my changes in the body itself. This will make it more clear what's up for discussion and what isn't.

@camlorn https://github.com/rust-lang/rfcs/pull/1574#issuecomment-216400267 : Yes, this is largely intended to the project itself, as well as giving projects a starting place to work. While I do care a lot about the accessibility case, I'd be interested in hearing more generally across a variety of screenreaders, rather than hastily changing a ton of styles because of one instance of difficulty. Thanks for chiming in here and in the other threads I've seen you in, though: I've done a little bit of accessability work on Rustdoc's output, but we have a long way to go...

@Centril https://github.com/rust-lang/rfcs/pull/1574#issuecomment-209005551: absolutely, this should be taken care of in the commit I just pushed

@jimmycuadra https://github.com/rust-lang/rfcs/pull/1574#issuecomment-207743058 I hear you, but changing Rustdoc's output is a different thing. I'd rather fix that issue than change semantic conventions because the current output is less than than stellar.

text/0000-more-api-documentation-conventions.md
jimmycuadra: Anecdotally, I can't recall ever seeing single quotes used for emphasis except in Rust documentation. I don't think it's standard or common.

jmesmon: for what it's worth, asciidoc uses single quotes to emit emphasized (italicized and differently colored) text. Perhaps rustdoc simply needs to use a similar rule & emphasize single quoted text? Or just convert the existing single-quote-to-emphasize into some other markup that actually applies an emphasis style.

cmbrandenburg: Always worth a laugh: [the <q>blog</q> of <q>unnecessary</q> quotation marks](http://www.unnecessaryquotes.com/).

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
...
+
+    This code does something cool.
+
+    ```rust
+    let x = foo();
+
+    x.bar();
+    ```
+
+When appropriate, make use of Rustdoc’s modifiers. Annotate triple grave blocks with
```
robinst: grave → backtick here too?

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,409 @@
+- Feature Name: More API Documentation Conventions
+- Start Date: 2016-03-31
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+[RFC 505] introduced certain conventions around documenting Rust projects. This
+RFC aguments that one, and a full text of the older one combined with these
```
carols10cents: Typo: s/aguments/augments/

carols10cents:
I just got [a great review on some doc changes](https://github.com/rust-lang/rust/pull/33815), and reading your comment gave me the sheepish feeling that there were some style guides about docs that I should have read. 🐑 😳 

I found this, and it does contain some of the suggestions you made (like a one-line summary at the beginning) but I think it could also use:

- [When referring to functions, include the `()`](https://github.com/rust-lang/rust/pull/33815#discussion_r64243056)
- [Module-level documentation should provide a broad overview; details should be in the struct/function/trait/etc documentation](https://github.com/rust-lang/rust/pull/33815#discussion_r64255823)
- [Strive to make code examples run with out warnings](https://github.com/rust-lang/rust/pull/33815#discussion_r64259119)?? I'm not sure about this one; the stdlib seems inconsistent to me and I'm not sure if this is enforced.
- I think it'd be worth saying, though it sounds super obvious, that code examples should follow the stdlib coding style? Maybe?

text/0000-more-api-documentation-conventions.md
steveklabnik: The new rustbuild process does check links, so this is done :)

steveklabnik:
Okay! Thank you for the delay here, everyone.

I believe I have addressed most/all of the comments here, but in different ways. Some of them have been updated directly, others have not, for a few reasons. 

Basically, it's not my intention with this RFC to re-litigate what was decided in RFC 505. As such, this RFC is now broken out: the contents of the RFC are _just_ the conventions that I'm proposing we add or tweak, and then the full text of RFC 505 + this RFC is added into an appendix. Some of the comments were about stuff that was in RFC 505, and that's out of scope of this RFC.

I would like to specifically call out the block comments vs line comments thing with a personal note. I do not want to change this in this RFC, but I also don't want this to be taken as not caring about the accessibility issue. However, I think that the accessibility issue needs to be dealt with more holistically: is it just this one particular screen reader that has issues, or is it many of them? Is there a way we could improve that software to deal with this particular issue, such that it would help everyone? Even if we did decree that this needed to change in the standard library today, as it is a years-long convention, there is a _lot_ of Rust code that would still be using this style. I think accessibility improvements would be great, generally, but also don't want to block the rest of this RFC on this issue, specifically. And since the proposed *changes* here don't make the situation worse, I feel it's okay to not address them at this particular moment.

A few others were about something we can't really change right now: rustdoc. I would prefer to not have to manually link, but until some hero goes and figures out a markdown extension to auto link types and implements it in Rustdoc, it's not realistic, so it is what it is, for now.



Whew! So with all of that, please re-review the actual Detailed Description. Thank you everyone!

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,726 @@
...
+[module-level-vs-type-level-docs]: #module-level-vs-type-level-docs
+
+There has often been a tension between module-level and type-level
+documentation. For example, in today's standard library, the various
+`*Cell` docs say, in the pages for each type, to "refer to the module-level
+documentation for more details."
+
+Instead, module-level documentation should show a high-level summary of
+everything in the module, and each type should document itself fully. It is
+okay if there is some small amount of duplication here. Module-level
```
GuillaumeGomez: "if there is _a_ some small amount"?

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,726 @@
...
+documentation should be broad, and not go into a lot of detail, which is left
```
GuillaumeGomez: detail__s__?

GuillaumeGomez: And I also think the first comma in the sentence is too much:

"documentation should be broad and not go into a lot of detail"

text/0000-more-api-documentation-conventions.md
tshepang: If someone new sees "Basic usage", they will expect to see something like "Advanced usage", which they won't in **most** examples.

GuillaumeGomez: I don't see why they'd expect an "Advanced usage". The goal of the doc is to help you understand how you can use the API. It's not supposed to cover every case.

tshepang: Seeing "Basic usage" implies "Advanced usage" is coming up, else why even mention it.

GuillaumeGomez: In case we want to add "Advanced usage". But "Basic usage" is a description, not an introduction.

tshepang: If you intend to add "Advanced usage", it's then that you may add "Basic usage". And those aren't so common, so rather leave it out.

GuillaumeGomez: Hum, I see your point. But then it wouldn't be uniformized. :-/

tshepang: I myself would not go as far as placing "Basic usage" text. I'd just:

1. present the (basic example) code block,
2. then describe the (more advanced) example,
3. and then its code block.

text/0000-more-api-documentation-conventions.md
steveklabnik: Both with and without the s work, actually.

text/0000-more-api-documentation-conventions.md
steveklabnik: Nope, it would be "if there is some small amount" or "if there is a small amount" :smile:

text/0000-more-api-documentation-conventions.md
steveklabnik: (the comma can probably be removed)

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: It leaves a lot of places to do whatever you want in here.

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: Damn! Too much "some"! :laughing:

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,724 @@
...
+    // ...
+}
+```
+
+### Using Markdown
+[using-markdown]: #using-markdown
+
+Within doc comments, use Markdown to format your documentation.
+
+Use top level headings # to indicate sections within your comment. Common headings:
```
jbcden: Should the `#` be in the middle of this line? If it is there to give an example of headers, we should do something to make it stand out more.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,724 @@
...
+* Safety
+* Aborts
+* Undefined Behavior
+
+Even if you only include one example, use the plural form: ‘Examples’ rather
+than ‘Example’. Future tooling is easier this way.
+
+Use backticks (`) to denote a code fragment within a sentence.
+
+Use backticks (```) to write longer examples, like this:
```
GuillaumeGomez: Not sure to understand what you mean here.... A one backtick is for "highlighting" a "small" element whereas triple backticks are for a whole code block.

text/0000-more-api-documentation-conventions.md
GuillaumeGomez: Right, a rewording might be needed.

text/0000-more-api-documentation-conventions.md
jmesmon: He's asking that the wording fully describe the thing in parens:

ie: "Use backticks (```)" becomes "Use triple backticks (```)".

steveklabnik:
Hear ye, hear ye! The nanescent docs team has decided to move this RFC into its final comment period.

wycats:
@steveklabnik More consistency for the win!

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,266 @@
...
+
+All documentation is standardized on American English, with regards to
+spelling, grammar, and punctuation conventions. Language changes over time,
+so this doesn’t mean that there is always a correct answer to every grammar
+question, but there is often some kind of formal consensus.
+
+One specific rule that comes up often: when quoting something for emphasis,
+use a single quote, and put punctuation outside the quotes, ‘this’. When
+quoting something at length, “use double quotes and put the punctuation
+inside of the quote.” Most documentation will end up using single quotes,
```
joshtriplett: I don't think code documentation conventions should recommend putting punctuation inside the quotes unless the punctuation itself is part of the quote.  Quoting the Jargon File, chapter 5, "Hacker Writing Style":
> This is incorrect according to standard American usage (which would put the continuation commas and the final period inside the string quotes); however, it is counter-intuitive to hackers to mutilate literal strings with characters that don't belong in them.'

This case will come up often in documentation, and a style guideline should not make a recommendation that on balance will reduce precision in favor of an arbitrary English formatting/typesetting convention that isn't even universal in English text.  (Convention varies: as I understand it, some style conventions for British English also put the punctuation outside the quotes.)

I realize that a style guide is the ultimate example of a bikeshed topic, and for the vast majority of items an arbitrary convention for consistency improves on having no convention at all. But in this particular case, I think this guideline does more harm than good.

text/0000-more-api-documentation-conventions.md
lambda: I agree with @tshepang that the "Basic usage" detracts more than it adds.

It is not currently used consistently in the standard library. And when I look at examples that include it, and examples that don't, I prefer the latter. For example, [`str::len()`] vs. [`Vec::len()`].

Now, if there is any prose discussion needed for an example, it makes sense to have some introductory text, or text describing each example in turn. But the string "Basic usage" just takes up vertical space, without adding much. The `# Examples` heading already sets this off as an example, and most examples can be fairly basic, so it seems like doing this just for the sake of consistency will waste a lot of space.

Even for those examples that do have some introductory text for more specific examples later, I don't think the "Basic usage" adds much. The [`str::chars()`] example would read just as well without it.

I decided to look to see if other style guides addressed examples in doc comments. Not many do; the [JavaDoc style guide] does not, nor does [Python's PEP-257]. The [Python `doctest` documentation] does not separate examples into a separate section, but intersperses them with the rest of the documentation. The [Ruby on Rails API Documentation Guide] recommends putting examples at the end, and makes an "Examples" heading optional.

I think that there are good arguments both ways for the "Examples" heading being optional or mandatory for consistency; I don't see a problem with requiring that. But requiring both an "Examples" heading and "Basic usage" seems like overkill.

[`str::len()`]: http://doc.rust-lang.org/std/primitive.str.html#method.len
[`Vec::len()`]: http://doc.rust-lang.org/collections/vec/struct.Vec.html#method.len
[`str::chars()`]: http://doc.rust-lang.org/std/primitive.str.html#method.chars
[JavaDoc style guide]: http://www.oracle.com/technetwork/articles/java/index-137868.html
[Python's PEP-257]: https://www.python.org/dev/peps/pep-0257/
[Python `doctest` documentation]: https://docs.python.org/2.7/library/doctest.html
[Ruby on Rails API Documentation Guide]: http://guides.rubyonrails.org/api_documentation_guidelines.html#example-code

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,730 @@
...
+‘string’ can refer to a `String` or an `&str`, and ‘an option’ can be ‘an `Option<T>`’.
+
+### Use parentheses for functions
+[use-parentheses-for-functions]: #use-parentheses-for-functions
+
+When referring to function names, include the `()`s after the name. For example, do this:
+
+```rust
+/// This behavior is similar to the way that `mem::replace()` works.
+```
```
lambda: I'm not sure that this should be the suggested convention. This misleadingly suggests that [`mem::replace`] takes no arguments.

I think that the convention should either be to write out names for the arguments when referring to a function, such as `mem::replace(dest, src)`, or leave off the parentheses entirely, as in `mem::replace`.

The [JavaDoc Style Guide] addresses this, and makes the same recommendation; of course, Java has the additional complication of having overloaded methods, but I still think that even without them, it's still misleading to refer to a function or method with an empty parameter list when it takes parameters.

[`mem::replace`]: http://doc.rust-lang.org/std/mem/fn.replace.html
[JavaDoc Style Guide]: http://www.oracle.com/technetwork/articles/java/index-137868.html#styleguide

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,730 @@
...
+
+Not this:
+
+```rust
+/// This behavior is similar to the way that `mem::replace` works.
+```
+
+This helps visually differentiate it in the text.
```
lambda: To address this reason, there is already plenty of visual distinction in both the Markdown, with the backticks and if the other guidelines are followed the link brackets, and in the rendered form, in the monospace font, grey background, and if other guidelines are followed, the link highlighting. I don't know if the additional visual differentiation of the parens is necessary; the only lowercase item in another module you could be referring to is a function, I believe, unless someone violates naming conventions.

jmesmon: One could also be referring to a module.

lambda: Ah, you're right.

text/0000-more-api-documentation-conventions.md
```diff
@@ -0,0 +1,730 @@
...
+
+[website]: http://www.rust-lang.org
+```
+
+But, if the text is very long, it is okay to use this form.
+
+### Examples in API docs
+[examples-in-api-docs]: #examples-in-api-docs
+
+Everything should have examples. Here is an example of how to do examples:
```
lambda: Could this section go into a little more detail? For instance, whether to prefer plain text outside of code blocks over comments within the code blocks. One recent example I've seen is [`str::is_char_boundary`], while I'm wondering if something more like the example in [`str::find`], with plain text explaining the examples, would be more appropriate.

[`str::is_char_boundary`]: http://doc.rust-lang.org/std/primitive.str.html#method.is_char_boundary
[`str::find`]: http://doc.rust-lang.org/nightly/std/primitive.str.html#method.find

tshepang: I think it's not very useful to have very simple API to have examples. An example is `Vec::new()`, which is simple due to it not taking any arguments. I also don't expect anyone looking at that API to get lost because they don't find an example on an API that simple. Let's not add examples for the sake of adding examples (i.e. needless consistency).

GuillaumeGomez: I can't agree on that. Examples are just way too convenient and useful to not add them. Just in the case you didn't understand the function description, you look at the example and it generally helps. Otherwise it's just convenient if you're too lazy to understand each parameter and just copy/paste.

tshepang: You want examples even for stuff like `Vec::new()`?

```rust
let vec = Vec::new()
```

That's?

GuillaumeGomez: Yup.

text/0000-more-api-documentation-conventions.md
jan-hudec: > That's something that we can hopefully do someday, but I don't want future rustdoc features blocking useful work today

But this is not about blocking useful work. It is about adding work that will be useless when the features are implemented.

steveklabnik: It is unclear, if ever, when that is landing. In the meantime, we need to improve the docs today.

jan-hudec: The best way to improve the docs is to _give that feature priority_. Adding the references manually is unqualified work, unlike implementing the feature, but it is a _lot_ of work. And not just for the standard library maintainers, but for maintainers of all the crates out there, if they want to have good documentation.

brson:
lgtm. <3 good docs

steveklabnik:
Hooray! The docs team, alongside the core team, has decided to *accept* this RFC.

There were some questions about a few points, and we decided to remove them all. You can see this in the last three commits I pushed to this branch last week.

Thank you everyone! I will be actually doing the merge later today.

Mahya321123 (approved):

PR #1216, opened 2015-07-19
Promote `!` to a type.

canndrew:
Promote `!` to be a full-fledged type equivalent to an `enum` with no variants.

[Rendered](https://github.com/canndrew/rfcs/blob/bang_type/text/0000-bang-type.md)

PR #1578, opened 2016-04-10
RFC for a Rust Memory Model

ubsan:
with thanks to Amanieu, huonw, durka42, aatch, acrichto, nmatsakis, and anyone
else I might have missed.

Fixes, or at least starts the process of fixing, #1447

### DIFF
--- a/text/0000-memory_model.md
+++ b/text/0000-memory_model.md
@@ -42,7 +42,8 @@ ever; examples are `std::mem::uninitialized()`, and padding bytes.
 are either "Defined" or "Undefined" depending on implementation details, like
 layout of structs. An "Implementation Defined" byte is only legal to read as a
 member of the original type it was a part of, in the original place it was in,
-or through reading fields of the original type.
+or through reading fields of the original type. Otherwise, how they are treated
+in all cases is implementation defined behavior.
 
 A value of type `T` can be in one of two states: "Valid", or "Invalid". Using an
 "Invalid" value is Undefined Behavior. Note that the definition of "Valid" and
@@ -65,7 +66,8 @@ not equal to the null pointer.
 
 Function pointer values will be "Valid" if each byte is "Defined", and are not
 equal to zero. Note that the size of function pointers is implementation
-defined, and not guaranteed equal to `*const ()`.
+defined, and not guaranteed equal to `*const ()`; however, a function pointer
+shall be compatible for FFI with a C function pointer of the same type.
 
 --
 
@@ -308,9 +310,7 @@ writes and reads happened in order.
 ## Typecasting
 
 Typecasting through pointers is fine. The clear example of this is `transmute`.
-However, "Implementation Defined" bytes are only readable as either the source
-type, or in an implementation defined way (see `std::repr`). Otherwise, if the
-type read is valid, then the type read is valid. The following are examples:
+The following are examples:
 
 ```rust
 // The following is completely valid
@@ -385,3 +385,5 @@ Are signaling NaNs "Invalid"?
 
 What's the deal with `UnsafeCell`? Probably something similar to raw pointers,
 except that it only applies to the array of bytes that make up the `UnsafeCell`.
+
+May different types of function pointers have different sizes?
--- a/text/0000-memory_model.md
+++ b/text/0000-memory_model.md
@@ -1,0 +1,387 @@
+- Feature Name: N/A
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Giving Rust a memory model. This allows us to understand what exactly is
+undefined behavior, and what is not, in unsafe code.
+
+# Motivation
+[motivation]: #motivation
+
+To allow unsafe code to be written without worrying about whether the compiler
+will miscompile your code.  Our current system is ill-defined, and far too
+cautious (and in other cases, completely undefined; for example, what are the
+semantics of raw pointer aliasing?).
+
+# Detailed design
+[design]: #detailed-design
+
+This is the complicated part :)
+
+## Using a value
+
+Anything which touches a value, and is not either a move or copy from the value,
+move or copy into an lvalue, or taking a reference to the value, is a use of the
+value. Examples of uses are: arithmetic, match, indexing. Examples of things
+which are not uses are: returning, passing a value to a function, `let x = y`,
+and `let x = &y`. A reborrow is not a use.
+
+## Type Representation
+
+Each type `T` shall be equivalent to a byte array: `[u8; size_of::<T>()]`. Each
+byte in this byte array shall be in one of three states: "Defined", "Undefined",
+or "Implementation Defined". "Defined" bytes are in a defined state at all
+times; they do not depend on which compiler, nor which platform one is on.
+"Undefined" bytes are also easy to understand; they do not have a defined value
+ever; examples are `std::mem::uninitialized()`, and padding bytes.
+"Implementation Defined" bytes are a little more difficult to understand; these
+are either "Defined" or "Undefined" depending on implementation details, like
+layout of structs. An "Implementation Defined" byte is only legal to read as a
+member of the original type it was a part of, in the original place it was in,
+or through reading fields of the original type.
+
+A value of type `T` can be in one of two states: "Valid", or "Invalid". Using an
+"Invalid" value is Undefined Behavior. Note that the definition of "Valid" and
+"Invalid" do not mean that a given value is correct for all uses; only that it
+is representationally valid. One good example of this is references; just
+because they are "Valid" does not mean that you can dereference them, only that
+they are not null.
+
+### Invalid and Valid Values
+
+All integer, floating point, and raw pointer to Sized values will be "Valid" if
+each byte is "Defined".
+
+--
+
+Reference to Sized values will be "Valid" if each byte is "Defined", and are
+not equal to the null pointer.
+
+--
+
+Function pointer values will be "Valid" if each byte is "Defined", and are not
+equal to zero. Note that the size of function pointers is implementation
+defined, and not guaranteed equal to `*const ()`.
+
+--
+
+`bool` values will be "Valid" if the byte is "Defined", and equals either `0x1`,
+or `0x0`.
+
+--
+
+`char` values will be "Valid" if each byte is "Defined", and, if read as a
+`u32`, would be within the range `[0x0,0xD7FF]∪[0xE000,0x10FFFF]`
+
+--
+
+Struct (including tuple) values will be "Valid" if each field is "Valid".
+
+Each field shall be an offset into the byte array which makes up the value of
+the struct.
+
+A rust representation struct will be made of "Implementation Defined" bytes; a C
+representation struct will be made of whatever the inner types are made of, in
+order, and "Undefined" bytes for the padding; a packed struct shall be made of
+whatever the inner types are made of, in order, with no padding.
+
+--
+
+Enum values will be "Valid" if the Discriminant is "Valid", and is one of the
+valid discrimants for the enum, and the discriminated value is also "Valid".
+
+Each discriminated value shall be at an offset into the byte array which makes
+up the value of the enum.
+
+An enum without associated values shall be equivalent to the discriminant, and
+shall have all "Defined" bytes. If the enum has associated values, all bytes
+shall be "Implementation Defined".
+
+--
+
+Union values will be "Valid" if the initialized field is "Valid".
+
+Each field shall be at an offset into the byte array which makes up the value of
+the union.
+
+A rust representation union will be made of "Implementation Defined" bytes; a C
+representation union will follow the C ABI of the platform, using inner bytes
+for where the inner types should go, and "Undefined" bytes where padding should
+go.
+
+--
+
+Pointer to !Sized values will be "Valid" if the pointer part of the !Sized
+pointer is "Valid", and the metadata part is "Valid"
+
+Each byte in a pointer to !Sized value shall be "Implementation Defined".
+
+## Pointer Rules
+
+These are only valid for Sized pointers; !Sized pointers will work the same way
+except that only the pointer part of the !Sized pointer is used.
+
+`ptr::read` and `ptr::write` will be the basis of the Rust pointer rules. They
+are both defined as a use.
+
+To `ptr::read` or `ptr::write` a value of type `T` from or to a raw pointer, the
+alignment of the raw pointer must be greater than or equal to `align_of::<T>()`
+
+To `ptr::read` a value of type `T` from a raw pointer, there must be
+`size_of::<T>()` bytes of storage readable behind the raw pointer
+
+To `ptr::write` a value of type `T` to a raw pointer, there must be
+`size_of::<T>()` bytes of storage writeable behind the raw pointer
+
+### Pointer Write Aliasing
+
+If a pointer refers to a value, then an aliased pointer is one where there is
+overlap in the referred to byte arrays; for example:
+
+```rust
+{
+  let x = [u32; 5];
+  let ref1 = &x[0..3];
+  let ref2 = &x[2..4];
+  // ref1 and ref2 are aliased
+}
+```
+
+A derived pointer shall be an in bounds pointer, calculated as a defined offset
+from another pointer value. Derived pointers shall be a tree; each derived
+pointer D derived from pointer D' shall also be derived from the pointers that
+D' is derived from.
+
+```rust
+// one common way to do this is with a reborrow
+{
+  let mut x = 0;
+  let ref1 = &mut x;
+  let ref2 = &mut *ref1;
+}
+
+// another is with field access
+{
+  let mut x = (0, 1);
+  let ref1 = &mut x;
+  let ref2 = &mut ref1.1;
+}
+
+// this is a tree
+{
+  let mut x = (0, (1, 2));
+  let ref1 = &mut x;
+  let ref2 = &mut ref1.1; // this is a derived pointer of ref1
+  let ref3 = &mut ref2.1; // this is a derived pointer of both ref1 and ref2
+}
+```
+
+A `ptr::read` or `ptr::write` of a reference makes any pointer derived from that
+reference non-derived.
+
+```rust
+{
+  let mut x = 0;
+  let ref1 = &mut x;
+  let ptr = ref1 as *mut i32; // ptr is now a "derived pointer"
+  ptr::write(ptr, 1); // fine, ptr is a derived pointer of ref1
+  ptr::read(ref1) // okay, ptr is no longer "derived", so don't touch it from
+                  // this point on
+}
+```
+
+Any move of a reference is a rederivation.
+
+```rust
+// UB if ref_ aliases ptr
+fn foo(ref_: &mut i32, ptr: *mut i32) -> i32 {
+  *ref_ = 0;
+  *ptr = 1;
+  *ref_
+}
+
+{
+  let mut x = 0;
+  let ref_ = &mut x;
+  let ptr = ref_ as *mut i32;
+  foo(ref_, ptr) // Undefined Behavior!!! ref_ is reborrowed with this move, so
+                 // the reference inside the function call doesn't see ptr as
+                 // derived
+}
+```
+
+To `ptr::read` or `ptr::write` a value of type `T` from or to a reference, in
+addition to following the rules of the raw pointer `ptr::read` or `ptr::write`:
+from the time of the creation of the reference, to when the reference goes out
+of scope, there shall be no aliasing `ptr::write` of any pointer which is not
+derived from that reference. Additionally, there shall be no `ptr::read` of an
+aliased pointer in the case of a `ptr::write`.
+
+```rust
+// the following is defined, as neither ref2 nor ref1 are written through
+{
+  let mut x: i32 = 0;
+  let ref1: &mut i32 = unsafe { &mut *(&mut x as *mut i32) };
+  let ref2: &mut i32 = &mut x;
+  // ref1 and ref2 can be assumed not to alias, but they are treated as &i32s,
+  // and &i32s are allowed to do this
+  ptr::read(ref2);
+  ptr::read(ref1)
+}
+// the following is defined, as ref1 is never touched
+{
+  let mut x: i32 = 0;
+  let ref1: &mut i32 = unsafe { &mut *(&mut x as *mut i32) };
+  let ref2: &mut i32 = &mut x;
+  // ref1 and ref2 can be assumed not to alias, but ref1 isn't ever read through
+  // or written through
+  ptr::write(ref2, 8);
+  ptr::read(ref2)
+}
+// the following is Undefined Behavior, as ref2 is written through *even after
+// ref1 is read from*, before ref1 goes out of scope
+{
+  let mut x: i32 = 0;
+  let ref1: &mut i32 = unsafe { &mut *(&mut x as *mut i32) };
+  let ref2: &mut i32 = &mut x;
+  // UB as ref1 and ref2 can be assumed not to alias
+  let ret = ptr::read(ref1);
+  ptr::write(ref2, 5);
+  ret
+}
+// the following is Undefined Behavior, as both ref1 and ref2 are written
+// through
+{
+  let mut x: i32 = 0;
+  let ref1: &mut i32 = unsafe { &mut *(&mut x as *mut i32) };
+  let ref2: &mut i32 = &mut x;
+  // this is UB as ref1 and ref2 can be assumed not to alias
+  ptr::write(ref2, 3);
+  ptr::write(ref1, 5); // UB
+}
+// the following is Undefined Behavior, as the raw pointer is read in the case
+// of a reference write
+{
+  let mut x: i32 = 0;
+  let ref1: &mut i32 = &mut x;
+  let ptr: *mut i32 = ref1 as *mut i32; // derived from ref1
+  let ref2: &mut i32 = ref1; // ptr is not derived from ref2
+  // This is UB as ref2 and ptr can be assumed not to alias
+  ptr::write(ref2, 15);
+  ptr::read(ptr)
+}
+// the following is defined, as two *mut Ts may alias, and they are both derived
+// from the first &mut T
+{
+  let mut x: i32 = 0;
+  let ref_: &mut i32 = &mut x;
+  let ptr1: *mut i32 = &mut *ref_;
+  let ptr2: *mut i32 = &mut *ref_;
+  ptr::write(ptr1, 3);
+  ptr::write(ptr2, 8);
+  ptr::read(ref_) // defined to return 8
+}
+// the following is defined, for the same reason as above; derived pointers are
+// a tree, remember
+{
+  let mut x: i32 = 0;
+  let ref_: &mut i32 = &mut x;
+  let ptr: *mut i32 = &mut *ref_;
+  let ptr1: *mut i32 = &mut *ptr;
+  let ptr2: *mut i32 = &mut *ptr;
+  ptr::write(ptr1, 3);
+  ptr::write(ptr2, 8);
+  ptr::read(ref_) // defined to return 8
+}
+```
+In other words: references may not observe aliased writes, and a reference only
+observes when it is actually used to `ptr::write` or `ptr::read`.
+
+Raw pointers may observe all aliased writes (assuming single threaded code), and
+it shall have a defined behavior, and the outcome shall be the same as if all
+writes and reads happened in order.
+
+## Typecasting
+
+Typecasting through pointers is fine. The clear example of this is `transmute`.
+However, "Implementation Defined" bytes are only readable as either the source
+type, or in an implementation defined way (see `std::repr`). Otherwise, if the
+type read is valid, then the type read is valid. The following are examples:
+
+```rust
+// The following is completely valid
+{
+  let i32_ptr: *const i32 = &(-5);
+  let u32_ptr = i32_ptr as *const u32;
+  ptr::read(u32_ptr)
+}
+// The following is also completely valid
+{
+  let i32_ptr: *const i32 = &0;
+  let f32_ptr = i32_ptr as *const f32;
+  ptr::read(f32_ptr)
+}
+// The following results in an invalid reference, which will result in UB if
+// used. However, it is fine to return it.
+{
+  let isize_ptr: *const isize = &0;
+  let ref_ptr = isize_ptr as *const &i32;
+  ptr::read(ref_ptr)
+}
+// The following is Undefined Behavior, as the tuple is larger than the original
+// type; in other words, i32_ptr points to 4 bytes of memory, while you are
+// reading at least 8
+{
+  let i32_ptr: *const i32 = &0;
+  let tuple_ptr = i32_ptr as *const (i32, i32);
+  ptr::read(tuple_ptr)
+
+}
+```
+
+`transmute` shall be defined very simply; equivalent to:
+
+```rust
+pub const unsafe fn transmute<T, U>(t: T) -> U
+    where size_of::<T>() == size_of::<U>() {
+    // assuming we get where bounds on values at some point (and const size_of)
+  let u = ptr::read(&t as *const T as *const U);
+  mem::forget(t);
+  u
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+More complicated rules. These are less easy to explain to people, and don't have
+the nice property of being proven (although I believe that they are closer to
+reality).
+
+Threading isn't defined in this document; it's only concerned with single
+threaded code. The current definitions are good enough, as far as I can tell,
+and I don't understand threading well enough to write the standard.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Keeping most unsafe code in the dark; currently, "It is an open question to what
+degree raw pointers have alias semantics. However it is important for these
+definitions to be sound that the existence of a raw pointer does not imply some
+kind of live path." This isn't good enough.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What is the exact definition of using a value?
+
+How do you define a valid discriminant value?
+
+Are signaling NaNs "Invalid"?
+
+What's the deal with `UnsafeCell`? Probably something similar to raw pointers,
+except that it only applies to the array of bytes that make up the `UnsafeCell`.
--- a/text/0000-memory_model.md
+++ b/text/0000-memory_model.md
@@ -45,6 +45,10 @@ member of the original type it was a part of, in the original place it was in,
 or through reading fields of the original type. Otherwise, how they are treated
 in all cases is implementation defined behavior.
 
+A `ptr::read` of type `T` from a pointer where a `T` was not last written shall
+result in an implementation defined value in any place that `T` has
+"Implementation Defined" bytes.
+
 A value of type `T` can be in one of two states: "Valid", or "Invalid". Using an
 "Invalid" value is Undefined Behavior. Note that the definition of "Valid" and
 "Invalid" do not mean that a given value is correct for all uses; only that it


ubsan:
cc @nikomatsakis @RalfJung @arielb1 @pnkfelix @Amanieu

Amanieu:
This opens up the possibility of performing TBAA on types which have implementation-defined bytes since these types are only allowed to be read through a pointer of the correct type. I don't really see this as a big issue since this is only possible from unsafe code and the struct layout isn't guaranteed to match any other type.

ubsan:
It's not that they're only allowed to be read through a pointer of a correct type. It's that if they are then used, it's undefined behavior for them to be aliased.

petrochenkov:
Enthusiasm is commendable, but I'd prefer this particular work to be done by professionals.

Ticki:
I agree with @petrochenkov, there are several shortfallings in this model.

petrochenkov:
@Ticki 
I'm not talking about concrete shortcomings, but about general approach.
I'd expect this to be a published academic paper by @RalfJung or someone else, with level appropriate for such papers, overview of prior art etc.

Anything less, e.g. "obvious" useful guarantees on struct layout or valid values of primitive types, can be specified on case by case basis.

Ticki:
> I'm not talking about concrete shortcomings, but about general approach.

Exactly. It should be formalized (to avoid these ambigiuities). It is very vague for a memory model, as is.

ubsan:
@Ticki @petrochenkov Honestly, this has been simplified and such a lot from how it would appear originally, in order to fit in the RFC format. Being a "professional" also has very little to do with the work here; I've been thinking about this memory model for months, and I'm likely one of the experts on Rust's. People send people to me to explain it. And, I get it; you would be more comfortable if the memory model was written by a professional. But I'm not a professional, and I wrote (part of) a memory model (the important part, the pointer aliasing part). The other things are just a basis for pointer aliasing, which is what really needs to be defined.

Ticki:
@usban Don't get me wrong. I don't doubt that you know a lot about this subject, and have put a lot of effort into it. What I am saying is just that this wasn't really what was inteded with a memory model. What was inteded is a formalized model, instead.

ubsan:
@Ticki well, I am working on formalizing the model now... I just wanted to talk about it first, in public. It's important that we are all able to give feedback on it.

archshift:
The fact is, *anything* is better than *nothing*, and *nothing* is exactly what Rust has right now. People are clueless as to the precise rules when it comes to pointer aliasing, which aren't currently documented anywhere.

dgrunwald:
This RFC reads to me as if there are two different operations that are both called `ptr::read`, but follow different rules (based on whether the argument is a pointer or reference).

In fact, if I see it correctly, replacing `ptr::read(int_ref as *const i32)` with `*int_ref` introduces the side-effect of making other pointers derived from that reference non-derived. Which means one can probably build programs where this harmless-looking code change introduces undefined behavior?

steveklabnik:
I am not gonna weigh in on this RFC very much, but one comment:

> The fact is, anything is better than nothing,

When it comes to guarantees that you're making, nothing can often be better. If we were to adopt a poor model, it could inhibit Rust for the rest of its life. You really want something like this to be absolutely foolproof, or as close to it as possible. Adopting something that's not ready because "something is better than nothing" is not a good idea here.

archshift:
For the moment, then, can we at least have some documentation that pointer aliasing of any form is Undefined Behavior? Because that's what it really is.

ubsan:
@archshift This is, in fact, true currently.

ubsan:
@steveklabnik One thing I found really compelling was something that dikaiosune said in IRC:

"is it just me, or is waiting for 5 or more years for any guarantees about unsafe pointers kinda untenable for rust adoption?"

Currently, this:

```rust
let mut x = 0;
let ptr1 = &mut x as *mut i32;
let ptr2 = &mut x as *mut i32;
*ptr1 = 1;
*ptr2 = 2;
*ptr1
```

is completely undefined. And most unsafe code assumes it is.

RalfJung:
> I'd expect this to be a published academic paper by @RalfJung or someone else, with level appropriate for such papers, overview of prior art etc.

Just to be clear here, I am not right now working on a memory model for Rust. Nobody should feel blocked by me here. For my [Rust formalization](https://www.ralfj.de/blog/2015/10/12/formalizing-rust.html) I decided to shortcut this discussion, and instead define *too many* behaviors. This means that at least I can prove the desired libraries to be safe in my model (hopefully), but they may still turn out to rely on behavior that Rust wishes not to define.

archshift:
@ubsan Even that is unclear, as according to the [Rust Doc](http://doc.rust-lang.org/book/raw-pointers.html):

>`*` pointers are allowed to alias, allowing them to be used to write shared-ownership types, and even thread-safe shared memory types

>[Raw pointers] have no guarantees about aliasing or mutability other than mutation not being allowed directly through a `*const T`.

ubsan:
@archshift Exactly. It's completely undefined.

dikaiosune:
Since @ubsan pinged me here, I'd like to share some of my thoughts on the level of formalism or academic chops present in the RFC.

One of the things that most impresses me about and attracts me to open source in general (and to a greater degree, Rust specifically) is the power that "expert amateurs" can have when they ignore credentials and focus on a substantive problem, emphasizing collaboration and results over social status and signalling. In many cases when one lacks the time, specific knowledge, or some other resource to substantively critique a work, the easiest signal to process can be from a contributor's credentials and prior achievements, or from the format of the presentation. But I think that's a bad trap for an inclusive community like Rust's to fall into.

I have always been impressed by the pragmatism in the other discussions I've read here, and I think it's a distraction to focus (explicitly or implicitly) on ubsan's qualifications or the current level of formalism of the RFC, especially when its author has approached it in this way:

> I am working on formalizing the model now... I just wanted to talk about it first, in public. It's important that we are all able to give feedback on it.

I don't see any reason to reject this out of hand. The product of this work will be very important for Rust's community and its future, and I think it deserves concrete feedback and a substantive appraisal.

Ticki:
I have to disagree with @steveklabnik on this one. In this case something _is_ better than nothing, as long as we make sure that the guarantees are not going to get broken in the future.

Personally, I would prefer an access based model over the one presented here. I am using an access based model in my formal verification of Redox. I have outlined my idea below:

(unfortunately, github doesn't support Latex, so I have rendered it as images)



and formalized through:

steveklabnik:
> I have to disagree with @steveklabnik on this one. In this case something is better than nothing, as long as we make sure that the guarantees are not going to get broken in the future.

I think we're on the same page. It's not so much that it has to be absolutely, fully, 100% perfect at first blush, just that we're not painted into a corner.

Ticki:
@steveklabnik, so what you are saying is that you think it should rather make one promise less than one promise more?

steveklabnik:
@Ticki I'm not familiar with that idiom. But I don't mean "something is worse than nothing" in the sense that it's an all-or-nothing enterprise. But I wouldn't want to rush to get something down just to have some kind of guarantee, only to find out that we're not actually happy with it later.

Anyway, I said I wasn't going to say much, and now I've made three posts, so I'll stop :wink:

comex:
Clearly we must aspire to the level of clarity and precision provided by the gold standard of standards, C. [[1]](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1520.htm) [[2]](http://www.open-std.org/jtc1/sc22/wg14/www/docs/summary.htm#dr_403) [[3]](http://plv.mpi-sws.org/c11comp/popl15.pdf) [[4]](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1637.pdf) [[5]](http://blog.frama-c.com/index.php?post/2013/03/13/indeterminate-undefined) [[6]](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66110) [[7]](http://blog.regehr.org/archives/1307)

arielb1:
```Rust
let mut x = 0;
let ptr1 = &mut x as *mut i32;
let ptr2 = &mut x as *mut i32;
*ptr1 = 1;
*ptr2 = 2;
*ptr1
```

This will actually work on every model that we plan to use. It is undefined in the exact same sense as everything else.

What *is* more dubious, is for example
```Rust
#[derive(Debug)] struct Foo<T>(T);
fn main() {
    let five = 5;
    let five_ref = &five;
    let mut x = Some(Foo(five_ref));
    let x_addr = &mut x as *mut Option<_>;
    match x {
        x_copy => {
            if let Some(ref inner) = x_copy {
                unsafe { *x_addr = None; }
                println!("{:?}", inner);
            }
        }
    }
}
```

This generates code that crashes *on current rustc*, through I don't think it should.

In any case, I don't like the derived pointer model because it is fundamentally based on "confidentality", while Rust basically tries to guarantee "integrity".

arielb1:
Additionally, LLVM likes attributes on loads, so lvalue-to-rvalue-conversions require the rvalue to be valid.

ubsan:
@arielb1 I don't think I understand what you are saying about confidentiality versus integrity. Or about lvalue-to-rvalue conversion attributes. Which specific attributes are you talking about, which would not be allowed?

Connorcpu:
@arielb1 I think crashing makes perfect sense though in this case. Moving a value is only semantic, it doesn't necessarily need to be given a new address, and for performance reasons often shouldn't.

Ericson2314:
Yeah this needs to wait for formal verification. I don't even trust a formal model that doesn't have a proven implementation to go with it. [A formal model is definitely first step to a proven implementation, but absent the proven implementation demonstrating the model's assumptions are valid, there's little use it "ratifying" the formal model.]

eternaleye:
I'll note that I specifically find the approach of specifying it in plain English unconvincing - even aside from the compelling  evidence that doing so is a strategy that consistently fails to provide results that are worth the effort[1][2][3], the simple fact of the matter is that English is a lot easier to mistranslate or misconstrue than a formal spec. "Plain English" is, to be quite honest, anything but - and I say this as a native speaker.

As a result, I have a *very* strong preference for a canonical formal spec paired with an explanatory document, rather than a canonical English spec that will hopefully one day be formalized.


[1]: Java's concurrency model first being unimplementable and then forbidding CSE
[2]: C11's memory model [forbidding common optimizations, amongst other issues](http://plv.mpi-sws.org/c11comp/popl15.pdf)
[3]: The heroic efforts needed to forumulate a [memory model for x86](https://www.cl.cam.ac.uk/~pes20/weakmemory/cacm.pdf) given prior plain-English descriptions

ubsan:
@eternaleye I understand that. I'm working on a formalization currently, but again, I wanted to talk about it in the open first to allow people to comment on it. This is a "Request For Comments".

eternaleye:
@ubsan To clarify, by a formal spec, I mean one suitable for machine-checking - the borrow checker and type checker are essential to safe Rust because humans cannot reliably model all the concurrency *present* in any *given* program; as a memory model must be well-defined for all concurrency *possible* in *any* program, the full verification of the model is in my opinion at least as crucial as borrow and type checking in safe Rust.

text/0000-memory_model.md
```diff
@@ -0,0 +1,393 @@
...
+
+This is the complicated part :)
+
+## Using a value
+
+Anything which touches a value, and is not either a move or copy from the value,
+move or copy into an lvalue, or taking a reference to the value, is a use of the
+value. Examples of uses are: arithmetic, match, indexing. Examples of things
+which are not uses are: returning, passing a value to a function, `let x = y`,
+and `let x = &y`. A reborrow is not a use.
```
Manishearth: A reborrow is not a use of the pointer or the value being pointed to? (best to be clear here, it's probably the latter)

chrisvest:
Would it be possible to have a memory model without Undefined Behaviour?

Manishearth:
Sure. "Exactly what Rust does right now" is a fully-defined memory model, albeit a rather useless one. You could also get rid of UB systematically.

But ultimately you need undefined behavior to give the optimizer room to maneuver. Rust protects you from undefined behavior within safe code, so it's not as much a problem as it is in other languages. You just need to worry about it when writing unsafe code or doing FFI.

Ericson2314:
@chrisvest even without "launch the nukes" undefined behavior, there are unavoidably plenty of non-deterministic machine programs due to the nature of multi processor computing. Yes, languages should be defined independent of implementation, but many features of Rust we care about *only* because they ensure the image of compiling avoids the peskier of those non-deterministic programs. So to prove Rust does we want, we need to model that non-determinism.

text/0000-memory_model.md
```diff
@@ -0,0 +1,393 @@
...
+
+## Type Representation
+
+Each type `T` shall be equivalent to a byte array: `[u8; size_of::<T>()]`. Each
```
petrochenkov: Before making statements like this it would be useful to give some motivation, discuss [alternatives](https://github.com/rust-lang/rfcs/issues/1397), their pros and cons, and why we need to abandon them forever in favor of this approach.
This text contains a lot of statements.

eternaleye:
@Manishearth There's some room to debate [how much UB-based optimization actually benefits optimization by compilers](http://www.complang.tuwien.ac.at/kps2015/proceedings/KPS_2015_submission_29.pdf), especially in the broader context of what changes it forces programmers to make in order to cope.

pnkfelix:
I am quite grateful to @ubsan for writing this up.

I too think that we will require a machine-formalized model of some form.

But we can only benefit from having a variety of well-written documents to draw from in order to understand the design space and trade-offs here.

It is good to have dialogue about the particulars of a model like this.

And it is much easier to read an RFC than to trudge through the various github issues trying to digest the many threads of conversation regarding memory models. (Issue #1447 could be a central repository of knowledge and ideas on the topic, but we will require reference documents that summarize the information there.)

text/0000-memory_model.md
petrochenkov: (This is also the reason why it would be better to discuss things on case by case basis rather than roll it up in one large blob. This RFC is wholly about pedantry and attention to details is important.)

eternaleye:
To summarize some discussion that occurred on #rust-offtopic (@ubsan, @eddyb, correct me if I misconstrue or miss something):

- There exist formalized 'weak' CPU memory models that we can use to bound us on one side
- Can shove libcore and libstd through `miri` and permute by model-permitted transformations to bound on the other side by seeing if anything goes splat

References:

[1]: [A formal hierarchy of weak memory models](http://www0.cs.ucl.ac.uk/staff/J.Alglave/papers/fmsd12-41-2.pdf)
[2]: [Other work by the author](http://www0.cs.ucl.ac.uk/staff/J.Alglave/, "herd" and "goto-instrument" likely to be of particular interest (simulating CPU memory models and a similar approach to the `miri` proposal applied to C, respectively)
[2]: [Software verification for weak memory](http://www.cprover.org/wmm/) (has source for goto-instrument, along with coq proofs, at the bottom)

eddyb:
@eternaleye The advantage of `miri` is that with some FFI support, it won't be limited to `libcore` and `libstd` but could run even (parts of) Servo (the use of IPC should make that easier).

eternaleye:
@eddyb The main reason I'd rather stick to libcore and libstd for now is to try and keep the model small - more guarantees can be added later, but by focusing on something minimally-viable (possibly by also altering libcore and libstd to ask less of the model in tandem) we can have something small and easy to reason about while being able to extend it to be stronger in the future.

arielb1:
@eternaleye 

If we have a model, we want to check all unsafe code we can to verify that it follows the model.

@ubsan 

Maybe I did not understand LLVM metadata correctly, but we occasionally emit assume/metadata on loads that the loaded rvalues are valid. For example:
```Rust
pub unsafe fn foo(zero_ref: &usize) -> &'static str {
    let do_ub = *(zero_ref as *const usize as *const &());
    let do_ub_ref : *const &() = &do_ub;
    if *(do_ub_ref as *const usize) == 0 {
        "zero"
    } else {
        "nonzero"
    }
}

fn main() {
    println!("{}", unsafe { foo(&11) });
    println!("{}", unsafe { foo(&0) });
}
```

text/0000-memory_model.md
ubsan: Not a use of the pointer, this is obviously not clear. I will clarify.

text/0000-memory_model.md
ubsan: I don't understand why having `size_of` return a stride is useful. This is easy to understand, makes sense, and is what C/C++ do. If there's a good argument for why we shouldn't do it this way, please, tell me :)

Amanieu: The main argument is that if we ever decide to do something like #1397 then making `size_of` return a stride would keep all existing code working. A separate function would be provided to get the real size of an object.

ubsan:
@arielb1

Correct. That's a use of an "Invalid" value; when you did `do_ub_ref as *const usize`, and then loaded it, and then did `== 0`, you used a value derived from an "Invalid" value, which is undefined behavior.

I need to specify this further. This is all in my head and not on the page.

@eternaleye I'm using coq right now, although I'm not very far.

arielb1:
LLVM's docs are *really* unclear as for whether loads require that the loaded value is valid. However, they assume that loads of dereferencable things are pure, so I guess that seals the deal.

So we probably can have poison rvalues like LLVM. I don't want to introduce the poison/undef distinction unless I am forced to, because nobody really understands `undef`, but I'm fine with just copying poison values around.

On the other hand, I am not sure how useful is that - what you want to do with poison values is either to have direct `let foo = mem::uninitialized()` or to put them in constructors, but *clearly* you can't put them in enum constructors. It probably makes sense to treat struct constructors as different from enum constructors here.

arielb1:
The reason I don't like derived-based analysis is that if you want to reason in a modular manner you have to play with "can-be-derived", which is very annoying to deal with.

ubsan:
@arielb1 The idea is that there is no can-be-derived. You know exactly which are derived, and which are not derived. It's a strict, at each point in the program, this pointer is either derived from or not derived from another pointer. There is no "can-be-derived", if I understand you.

arielb1:
@ubsan

I understand that at runtime the derived relation is definite. On the other hand, if you want to reason about the validity of optimizations in the presence of external functions, determining whether one pointer is derived from another is non-trivial.

Also, derived pointers are somewhat similar to C++'s consume memory model, which is a thing I want to stay away from (but maybe there is some significant difference?)

ubsan:
@arielb1 Function arguments are one area where derived pointer relations are cut off entirely. A move is counted as a rederivation, and you must move a reference into a function in order to pass the argument; therefore, each new function gets a "clean slate", as it were.

I don't know what you mean by "consume memory model". Could you give an example?

gereeter:
This looks like a very good start, but it is important to note that, so far, this RFC is only a memory model for a single thread. I don't necessarily think that concurrency should be a blocking issue, as even just a single-threaded memory model clears up a lot of questions about the correctness of unsafe code. However, the RFC should explicitly state that it leaves the details with respect to concurrency unspecified, and should eventually be extended to support concurrency.

arielb1:
@ubsan

I meant memory_order_consume.

arielb1:
@ubsan

I meant memory_order_consume.

ubsan:
@arielb1 I don't understand what you are saying. Could you explain it out, please?

arielb1:
I would like to reword the "using a value" paragraph to be more in line with the MIR nomenclature.

I am using pseudo-Rust as my language for data declarations because it is the most familiar for the people reading this.

## Type Representation Kinds

At run-time, every type is eqty to a single unique *concrete type*, that's it, a type without type parameters and projections (*FIXME* projections within binders). Concrete types can be divided as follows:

 * A *primitive* type is a type that is treated as a set of bits. This is integers, floats, and raw pointers.
 * A *scalar* type is a type with an atomic representation - primitive types, plus bool, char, references, function pointers, and `FnDef` markers
 * A *struct-like* type is a struct, tuple, or closure.
 * An *array* type is a slice or array
 * A *enum-like* type is an enum
 * A *universal* type is a trait object

*NOTE*: should C-like enums be scalars?

## Values

MIR Rvalues evaluate to a value, which is defined as follows:

```Rust
pub enum Value {
    Poison,
    Scalar(Vec<u8>),
    Struct(Vec<Value>),
    Array(Vec<Value>),
    Enum(Discr, Vec<Value>),
}

impl Value {
    fn contains_poison(&self) {
        match self {
	    &Poison => true,
	    &Scalar(..) => false, // NOTE: don't look through pointers
	    &Struct(ref vs) | &Array(ref vs) | &Enum(_, ref vs) => {
	        vs.iter().any(|v| v.contains_poison())
	    }
	}
    }
}
```

*NOTE*: Do we want to track pointers as separate from integers? This is
required for const-eval and derivation-path.

Non-Poison values must conform to their type representation. This means that:
 * primitives must be a Scalar of the right size
 * bool and char must be valid Scalars
 * pointers and function pointers must be non-null. [0]
 * enums must have a valid discriminant, and must not contain a Poison.
 * struct-likes, arrays and enums must have the correct number of fields, and they all must all conform to their representation.
 * a NonZero struct must contain a non-zero Scalar

If an operation would create a value that does not conform to its representation, a Poison value is created instead.

When a function returns, its return pointer must not contain a Poison. Also, unless otherwise specified, operations on rvalues invoke undefined behavior if one of their inputs contains a Poison. Notable exceptions are:

 * `StatementKind::Assign`: writes poison bytes to the memory ranges covered by poison.
 * `Rvalue::Aggregate` and `Rvalue::Repeat`: these operate normally, creating a `Value` that contains a `Poison` but that is of the correct constructor. Note that the rule about values having to conform to their representation applies - an `enum` that contains a poison will be immediately converted to a `Poison`.

[0] This prohibits using the `!dereferenceable` or `!align` metadata on loads of pointers. We currently don't do so but we may want to in the future. In any case, we want this pattern to be a GEPi with no extra requirements:
```Rust
fn foo(random: *const Foo) -> *const u8 {
    unsafe { &((&*random).1) }
}
```

or in MIR
```
let arg0: *const Foo;
let tmp0: &Foo;
let tmp1: &u8;

begin:
    tmp0 = &*arg0;
    tmp1 = &(*tmp0).1;
    return = tmp0 as *const u8 (Misc);
    return;
```

nikomatsakis:
Hello all. I want to weigh in briefly. I am still digesting the RFC text and the comment thread, but I want to make some kind of "meta" comments.

#### Generalities

First, like @pnkfelix, I certainly would like to thank @ubsan for putting energy and effort into a "first draft" at some sort of memory model for Rust. Whatever model we eventually adopt, it's great to make forward progress of some kind.

I think the best place for us to focus our efforts at this stage is not on determining the precise model we will use, but rather the **process** by which we will arrive at a model, and the **high-level principles** that will guide us in this effort.

I think the **process** should be as top-down as possible. Basically I'd like to start just by agreeing on the kinds of examples we think are interesting and what makes them interesting. (More on this in a bit.) From there maybe we can come up with high-level rules that "summarize" the key ideas, and then drill down into making them more concrete. One nice part about this is that at each stage of the process we are giving more information to end users about what kinds of code we expect to be legal and illegal. Honestly, I hope that the list of examples alone will be sufficient advice for the most part.

An ideal end-point would be to arrive at something testable, for sure. The idea of an instrumented miri-like interpreter is particularly appealing. I would like to be able to "test" our proposed models by doing "crater runs" across crates.io and seeing what kind of code reports violations. 

Now, as far as guiding principles go, I would like to see a model that aims to **avoid crashes** in the code that **naive people** write. Put another way, I expect that most programmers will either not read the memory at all, or will only read no more than a few paragraphs. I would like that, given that level of innattention, they still mostly do things right when they write unsafe code. Effectively this means the model must be "intuitive"; that is, whatever model we adapt, it has to permit the kinds of tricks that people pull in practice. I think that many models I have seen fail this criteria dramatically. The problem with an "intuitive" (or perhaps "permissive" is a better word) model is that it naturally runs somewhat counter to the ability to optimize. I think that Rust has a leg up here, though, both because of the nature of our type system and because of the safe/unsafe code division.

It is not clear that we can achieve a model that is both permissive *and* permits a sufficient level of optimization. But let's posit that we can for a second (and I do believe it is possible). In that case, it's interesting to note that the urgency of having a formal memory model is *somewhat* reduced, in that we expect most things people do in practice to wind up being sound.

#### Examples and repo

I mentioned that I thought we ought to start by agreeing upon a set of interesting examples and so forth. I've been making an effort to collect such a set, though I think it's still in relatively early stages (and I haven't had much time for this in the last month or so). You can see my [current results][cr] here.

There are still a number of things I want to do:

1. Add some examples about when it is safe to "reuse" a lot; as [@arielb1's example][ex] earlier demonstrates, this is not always clear).
2. Consolidate and clean up the examples, which I think contain duplication now, since I mostly just copied things I found on various threads.
3. Go through unsafe code and extract interesting patterns. I was planning to start with libstd. 

I'd very much welcome contributions.

I've been hoping that once we have a pretty good set of examples etc, we can (a) consolidate and clean them up and then (b) use them as a basis for comparing proposals.

[cr]: https://github.com/nikomatsakis/rust-memory-model
[ex]: https://github.com/rust-lang/rfcs/pull/1578#issuecomment-208582907

#### Specific responses

Sorry this comment doesn't have much in way of specific responses to this RFC or comments in the thread. It's long enough as is, I guess. I've been taking notes on the RFC as I read as well as other comments on the thread, and I'll try to get back and respond to more specific details there as well.

ubsan:
@nikomatsakis 

I've been thinking about this a lot since then. I think this is a good start for a memory model, but I didn't understand optimizations well enough at that point. Yeah, I do believe it's a good idea to use Miri to find undefined behavior.

What I've been thinking about is this:

Unfortunately, I think that any use of an uninitialized value should be undefined behavior. On some architectures, even moving from an uninitialized register to another register will be bad news, due to flags. I think we might have to adopt the C standards ideas of `u8/i8` is special; it can be undef, and nothing else. The only thing you can do with an undef variable is write to it, or take a pointer to it.

Our references are `dereferenceable`, which is good for optimization purposes, but which also means; you can't have a reference to an undef, in case you accidentally read an undef. Perhaps something like, the only thing you can do with a reference to an undef is 1) write through it, or 2) create a pointer from it?

eternaleye:
@ubsan: Doesn't that then require the optimizer to be conservative with any reference it cannot prove is to a defined value? I'd much prefer saying that pointer-to-reference coercions are UB if the pointee is an undefined value, and use `ptr::write()` for the "write through" part.

ubsan:
@eternaleye No, the optimizer assumes that all references it doesn't know are to a defined value, are to a defined value (unless that reference is a reference to a byte, or a byte array). The following are the only times a reference to an undefined is okay, I think:

```rust
let mut x = std::mem::uninitialized();
let ptr = &mut x;
write(ptr, ...);
// or
let mut x = std::mem::uninitialized();
let ptr = &mut x as *mut ...;
```

eternaleye:
@ubsan: Hm, I see. Perhaps a combination, by saying that the only valid op on ref-to-undef is _immediate_ conversion to pointer, something trivially lintable. Note that I would also lint returning the ref.

It can't be an error (back-compat), but IMO ref-to-undef is an unequivocal error, and the first form should habe been disallowed. :(

ubsan:
@eternaleye Perhaps. this is far easier to deal with, however; just moving a reference to undef is undefined behavior.

eternaleye:
@ubsan: If a reference to undef existing any more than ephemerally is UB, all timelines in which such a move occurs already encountered UB at an earlier point.

Consider:
```Rust
unsafe fn foo(x: &mut T) {
    *x = mem::uninitialized();
}
```

This function should lint, IMO, merely by existing. "Escape of ref to undef" permits that.

ubsan:
@eternaleye It makes sense to lint, because you're moving uninitialized data.

```rust
let x = std::mem::uninitialized();
```

is the only way to get uninitialized data, I think... That would be equivalent to something like

```rust
unsafe fn foo(x: &mut T) {
  let uninit = mem::uninitialized();
  *x = uninit;
}
```

ubsan:
Or... `= std::mem::uninitialized()` could be defined as a no-op.

```rust
fn foo(x: &mut T) {
  *x = mem::uninitialized();
}
```
could be equivalent to
```rust
fn foo(x: &mut T) {
}
```

eternaleye:
Mm, right - forgot the initializer/move distinction there. Still, I think losing the `dereferenceable` attribute on references is a bad trade - and if moving undef is UB, then proving refs to undef are immediately converted to pointers is an entirely tractable analysis AFAICT.

ubsan:
@eternaleye The point is there is no losing the `dereferenceable` attribute on references. Once you move a reference to an undef, it's completely `dereferenceable`.

eternaleye:
Mm, I think we may actually being violent agreement, but were separated by terms. As 'escape' requires a move one way or another (assign or return). Sorry about that.

I do think the non-escaped ref-of-undef should only permit pointer converstion and not writes, though - we have `ptr::write()` already.

ubsan:
@eternaleye Yeah, I can agree with that.

I do like the idea of `= std::mem::uninitialized()` being a no-op in the future. Although... we definitely use undefined behavior with our current implementation if there are any trap representations. If we changed `std::mem::uninitialized` to `pub use std::intrinsics::uninit as uninitialized`, and just made any use of `intriniscs::uninit` on the rhs of an = a no-op.

This would make:

`<expr> = intrinsics::uninit();` into `<expr>;`

arielb1:
Does LLVM `dereferencable` require that the value behind it is valid? I thought that it only requires that it can be read without triggering a page fault (i.e., it does not even guarantee that the value is not modified by other threads).

The other kinds of load metadata (e.g. `nonnull`) *seem* to turn invalid data loaded to undef/poison as I specified. It would be better if that was documented somewhere. LLVM has no problem moving undefs and even storing them to memory - it even performs optimizations that rely on that.

ubsan:
@arielb1 The issue is that, at least in my opinion, we shouldn't be using LLVM as a benchmark. I'm of the opinion that we should allow implementations almost as much wiggle room for trap representations as C does, perhaps excluding integer types. But, for example, if a computer traps on passing sNaN, that should be allowed behavior, or if a computer traps if you read from a register which has NaT set.

arielb1:
I *thought* you were *for* allowing the loading of `undef` rvalues?

The issue here is that we want `box mem::uninitialized()` and the likes to work, and that desugars into something of the form
```
        tmp0 = Box::new();
        (*tmp0) = std::mem::uninitialized::<u32>() -> [return: bb3, unwind: bb2];
```

Should we say that "RVO" writes of `undef` are ok?

In any case, I heard that people want to create structs with only *some* of their fields uninitialized, which currently generates e.g.
```
    start:
        tmp0 = Box::<Big>();
        tmp2 = std::mem::uninitialized::<[u32; 65536]>() uwto unwind0
        (*tmp0) = Big { tag: const 42u8, data: tmp2 };
        return = tmp0;
        return;

    unwind0:
        tmp1 = alloc::heap::box_free::<Big>(tmp0);
        resume; // Scope(0) at <anon>:11:1: 13:2
```

arielb1:
Source code for ^:
```Rust
fn make() -> Box<Big> {
    unsafe { box Big { tag: 0, data: mem::uninitialized() } }
}
```

ubsan:
@arielb1 I thought they were. Then I had a big dose of reality with CPUs that weren't x86 or ARM:

https://blogs.msdn.microsoft.com/oldnewthing/20040119-00/?p=41003/

We may have to use (something like) the C/C++ definition of uninitialized:

An uninitialized variable may either be in a valid, but unspecified state, or a trap representation (and it may change whenever you read it). The only thing that's not allowed to have a trap representation is i8/u8. If you do *anything* except write to a variable with trap representation, UB happens.

ubsan:
So, for rustc on x86 and ARM, for example, the integer and pointer types would have no trap representations.

ubsan:
@arielb1 It doesn't really matter how we translate it, just how we define it; `box mem::uninitialized::<T>()` should be equivalent to `Box::from_raw(allocate(size_of::<T>(), align_of::<T>()) as *mut T)`.

eternaleye:
@ubsan: And placement box with `mem::uninitialized::<T>()` would simply skip initialization (i.e. noop)?

ubsan:
@eternaleye That's the idea.

ticki:
@ubsan I'm interested in hearing your response to [the comment I wrote almost a month ago](https://github.com/rust-lang/rfcs/pull/1578#issuecomment-208509258). I wonder what your thoughts are?

ubsan:
@ticki 1) you don't account for shared mutable state, which is allowed for with both raw pointers, and with UnsafeCell, afaict, 2) What can you do with uninitialized variables besides write to them? 3) how is dropping a reference defined? I imagine something like "last use".

Seems good otherwise. Sorry, it kinda got missed.

ticki:
1) I do. See point 1.:

> mutating a **non-aliased** value [..]

Or do you mean mutating while having other (passive) pointers accesible?

2) Well, reading an uninitialized value will always be unsafe, writing or aliasing one is valid of course.

3) See point 3., basically the "hidden counter" is decremented. From a technical point of view, this has the effect of changing the type state, potentially allowing future mutation.

4) Fair point, but the idea is the same: the program's behavior is well defined.

>  Sorry, it kinda got missed.

np.

ubsan:
@ticki 1) Yes, but mutating an aliased value is defined, i.e. in the case of Cell<T>. 2) Reading an uninitialized value may be *unsafe*, but when is it *invalid* (for example, i8/u8 should always be valid, other types may be valid/invalid). 3) *When* is a reference dropped, not what happens when a reference is dropped.

ticki:
Reading an uninitialized u8 is undefined in LLVM. Despite the types having no inconsistent representation, they cannot be read uninitialized, since the behavior would be defined by how the stack frame was layed out.

ticki:
Reading an uninitialized u8 is undefined in LLVM. Despite the types having no inconsistent representation, they cannot be read uninitialized, since the behavior would be defined by how the stack frame was layed out.

ubsan:
@ticki It results in an uninitialized value. It does not result in undefined behavior (and in fact, this is true of all integer types in LLVM).

ticki:
On 3), the reference is dropped when it is inaccesible or dead.

ubsan:
@ticki "Inaccessible" and "dead" are two very different things, and could be defined in multiple ways. Please give examples?

ticki:
Dead means that _it won't be used at some later point in execution_. Inaccessible means that _it is out of scope or otherwise not possible to read or write_.

ubsan:
Got it. I believe we want "dead" as the definition, otherwise our new SEME borrowck would be breaking the rules :P

ubsan:
And I'm still not clear on what aliased writes would mean?

ticki:
An aliased write means that you write while another pointer is active.

ubsan:
That's the definition of an aliased write. What does it mean for the program? How does one define an aliased write in your model? (for example, when using Cell<T>, or using raw pointers)

ubsan:
https://gist.github.com/ubsan/1fb0aa56dbab7ff2cd6713cbb5ae75f0

@nikomatsakis These are a few more litmus tests, and some thoughts on `std::mem::uninitialized`.

nikomatsakis:
I wrote a blog post discussing my thoughts on a high-level approach to the memory model:

http://smallcultfollowing.com/babysteps/blog/2016/05/27/the-tootsie-pop-model-for-unsafe-code/

The heart of the proposal is the intution that:

- when you enter the unsafe boundary, you can rely that the Rust type system invariants hold;
- when you exit the unsafe boundary, you must ensure that the Rust type system invariants are restored;
- in the interim, you can break a lot of rules (though not all the rules).

Discuss thread: http://internals.rust-lang.org/t/tootsie-pop-model-for-unsafe-code/3522

nikomatsakis:
@ubsan 

> These are a few more litmus tests, and some thoughts on std::mem::uninitialized.

Great! I'm looking through now. As always, I apologize for my permanently backlogged state. Most of the initial examples are basically safe code, so it seems indeed pretty hard to imagine them being undefined. 

----

```rust
unsafe fn memcpy(dst: *mut u8, mut src: *const u8, len: usize) {
  for el in std::slice::from_raw_parts_mut(dst, len) {
    *el = *src;
    src = src.offset(1);
  }
}
```

I guess the key point here is the `std::slice::from_raw_parts_mut`, which then produces an `&mut [u8]`? Or is the key point that this is applied to a struct with padding?

-------

```
// dereferencing any non-null pointer to a ZST
fn f(x: *const ()) -> () { if x.is_null() { () } else { unsafe { *x } } }
```

I definitely think that dereferencing a pointer to a ZST is invalid. Put another way, you have the right to read 0 bytes of data through that pointer, which means dereferencing is invalid.

------

```
// Creating a raw pointer to an uninitialized value
let x: u32 = std::mem::uninitialized();
let y = &x as *const u32;
```

What about this example worries you? Is it the transient `&x`? It seems like a "pointer to an uninitialized value" will happen every time you call `free`, so the mere existence of such a pointer had better not be a big problem.

------

```
let x: u32 = std::mem::uninitialized();
let y = x;
``` 

Yeah, this is a good example to chew on. There are also some fine distinctions to be drawn, around questions like padding and so forth. (Like, just what constitutes uninitialized?) I need to read more into that blog post you cited also.

I'm curious what differentiates the `u32` above from the `u8` case you cite later on:

```
// C allows this; should we (I believe so)
let x: u8 = std::mem::uninitialized();
let y = x;
```

> And note, I haven't talked about this, but std::mem::uninitialized is a function, and so we're (semantically, at least) moving uninitialized values when we call it. Why is it special? 

Yes, a good question, I agree! Seems to argue that we should permit accesses from uninitialized memory, at least in certain ways.

-----

As you can see from my blog post, I disagree that this ought to be undefined:

```
// Aliasing writes with &mut
fn f(x: &mut i32, y: &i32) -> i32 { 
  *x = *y + 1;
  *y
}
let x = 10;
let y = &mut x as *mut _;
unsafe { f(&mut *ptr, &*ptr) }
```

glaebhoerl:
> I definitely think that dereferencing a pointer to a ZST is invalid. Put another way, *you have the right to read 0 bytes of data through that pointer*, which means dereferencing is invalid.

(Emphasis mine.)

By dereferencing it, aren't you... doing exactly that? I would expect dereferencing a pointer to type `T` to always read `size_of::<T>()` bytes of memory from the address that it points to. In the case of a zero-sized type, that's 0 bytes. A no-op. Safe as heck.

ticki:
@glaebhoerl Despite being zero sized, it can still be unsafe to deref (e.g., segfaults), unless you want to add a zero-check for unsized types.

ubsan:
@nikomatsakis

Yes, the issue is that you're applying it to a struct with padding, which should be okay with `u8`s.

 "you have the right to read 0 bytes of data through that pointer". Which is what you are doing.

The issue I have with it is that `&T` should always be safe to deref, I guess? They're dereferenceable in LLVM, so... Pointers, I have no problem with, they're not dereferenceable.

What differentiates `u32` from `u8` is that C says that `char` must be allowed to be uninitialized. Therefore, we can assume that, at least, char doesn't have any trap representations.

Assume that f is in a different, completely safe module :P

arielb1:
A problem that has come up in the discussion is assignment temporary elimination.

As you know, assignment of a struct without a destructor creates MIR like `tmp = RVAL; LVAL = tmp`. The question is when we can optimize that to `LVAL = RVAL`.

That is significant, because the computation of `RVAL` can depend on `LVAL`, For example (this compiles today):
```Rust
fn must_work() {
    let mut x = S::new();
    x = foo(&x);
}
```

If we are talking about our optimization, the most simple interesting example is (everywhere `Big` is a struct without a destructor).
```Rust
fn example0(b: &mut Big) {
    println!("{:?}", b);
    *b = Big::new();
}
```

Here, we would like to optimize out the assignment temporary. The problem is, there is no way we can prove the formatting code is not storing a copy of `b` inside some global, which `Big::new` will read after it invalidates it.

If we decide to special-case globals in some way, we still are in some trouble:
```Rust
fn example0(b: &mut Big) {
    let x = foo(&b);
    *b = Big::new(x);
}
```

Here, there are a few principal possibilities.
* `foo` returns a boolean or something. In that case, we would like to avoid generating a temporary if possible.
* `foo` is the identity function. Then the borrow of `b` will be extended to contain the call to `Big::new`, and we have safe code that compiles, and we must generate the temporary.
* `foo` is `<*const Big>::as_ptr`. In unsafe code, we would like that to work just like the previous case. The problem is that we can't distinguish this case from the first one reliably - every return value that contains a raw pointer is possibly suspect.

arielb1:
There is one pattern that is commonly used that @ubsan's definition seems to make make UB:

```Rust
fn whatever(a: &mut u32) -> u32 {
     // either order of assignments
    let ptr0: *const u32 = a;
    let ptr1: *mut u32 = a;
    let result = access(&*a_ptr0); // for some `fn access(&u32) -> u32`
    *a_ptr1 = 0;
    result
}

fn whatever_mut(a: &mut u32) -> u32 {
     // either order of assignments
    let ptr0: *const u32 = a;
    let ptr1: *mut u32 = a;
    manipulate(&mut *a_ptr1); // for some `fn manipulate(&mut u32)`
    *a_ptr0
}
```

For example, `ptr::write` desugars to the second.

ubsan:
@arielb1 Yes, your second example would be UB under my memory model. I have some ideas for solving it though.

arielb1:
## Rust aliasing model

### Base Definitions

References are either immutable, mutable, or moving.
Access kinds are `read` or `write`.

### The aliasing rule

If a reference `r` with lifetime `'r` is created, then afterwards, before `'r` ends but at no particular order
    * Some memory location `l` is accessed through a borrow-chain from `r` - this is the asserting access.
    * The same memory location is accessed in any other way - this is the conflicting access.
    * One of these accesses is a write.
Then an aliasing access has occured.

Aliasing accesses are UB, unless there exists a reference `s` to with lifetime `'s`, such that
    * `s` was created by a borrow-chain from `r`
    * `s` has sufficient permissions for the conflicting access.
    * `'s` is alive at the point of access.
    * `s` is potentially alive. This means either:
        - `s` is an immutable reference 
        - `s` is the reference used for the conflicting access
        - `s` was converted to a raw pointer before the conflicting access

In that case, `s` is called the guaranteeing reference.

### Reference Creation

All references are created by reborrowing, either of a prior reference or of a raw pointer.

A reference is directly reborrowed from another if it is the borrow of some lvalue that came from the original reference. A borrow chain is the reflexive transitive closure of the directly-reborrowed relation.

NOTE: lvalues can contain dereferences of `&mut` - these have the borrow chain of *their parent*. However, dereferences of `&` have the borrow chain of the borrowed reference itself.

NOTE: raw pointers can't take part of a borrow chain.

References created by transmutation *ex nihilo* should be handled as if they were created by reborrowing a raw pointer.

glaebhoerl:
> unless there exists a reference `s` to with lifetime `'s`

Missing word or stray "to"?

arielb1:
### Additional

Does anyone have any other interesting examples for me to analyse? @ubsan @nikomatsakis

nikomatsakis:
@arielb1 I have a number of questions about the rules you propose here. I think a lot of this has to do with definitions.

For example, when you say "reference", do you mean specifically a value of type `&T` or `&mut T` (as opposed to, say, a raw pointer)?

Similarly, when you say "reborrow", what do you mean? I know of at least two definitions for that term (me and @RalfJung seem to have intuited different meanings, for example). I'm not sure how important it is. My definition is basically just a subset of borrows: that is, borrowing a value via some lvalue that involves dereferencing a reference (and thus reaching into content that was already borrowed). So if `x: &i32`, then `&x` is a borrow and `&*x` is a reborrow of the referent.

Certainly the part of this proposal that I find most worrisome was the fact that it relies on specific lifetimes, and in particular the "conservative" rule around reborrows (casts) from `&T` to `*const T`. For example, if I have an argument `p: &'a u32`, and I cast `p` to `*const u32`, then saying that this cast extends for all of `'a` seems maybe plausible. But what if I cast `&*p` to `*const u32`? In that case, the reference I am casting actually has a very short lifetime, but surely we want the same result? 

I've been hoping rather for a model where the legal aliasing doesn't consider the lifetime of the references in question at all. Rather, we can define the memory that a fn has *permissions* to affect and how -- and lifetimes do come into play here, since if you pass in a `&'a i32`, you are granting that fn permission to read from the reference until the end of `'a` (and you are *not* granting permission to mutate). (And when you declare `let x: u32`, you are giving permission to read from that stack slot until the end of the enclosing scope.) But within the fn body itself, the optimizer would not consider lifetimes. (Note that you can have this without going to the full Tootsie Pop model I proposed in my blog post.)

arielb1:
@nikomatsakis 

Sure. Reference = `TyRef` = `&`,`&mut`,`&move`. And in my scheme, all borrows are reborrows because locals are lifted into allocas. 

> Certainly the part of this proposal that I find most worrisome was the fact that it relies on specific lifetimes, and in particular the "conservative" rule around reborrows (casts) from `&T` to `*const T`. For example, if I have an argument `p: &'a u32`, and I cast `p` to `*const u32`, then saying that this cast extends for all of `'a` seems maybe plausible. But what if I cast `&*p` to `*const u32`? In that case, the reference I am casting actually has a very short lifetime, but surely we want the same result? 

And it would *have* the same result. The `&*p` borrow must also be treated conservatively for UB purposes, therefore it would also be extended. What I meant by being conservative is that if there *exists* some way of inferring lifetimes that is not UB, then the code is not UB. Basically, we can't infer lifetimes correctly, so we must assume the worst.

> permissions

I am quite sure that my model can be stated equivalently in terms of capabilities. In fact, that was the first form I had in mind, and I switched to the access-based form for clarity.

> you are granting that fn permission to read from the reference until the end of `'a` (and you are not granting permission to mutate).

See how the lifetimes sneak back in? I don't see any way to have a workable model without lifetimes - for example, the only difference between `rvo_safe` and `no_rvo_safe` (which I added now) is the lifetime bounds in `as_ptr` vs. `id`, yet the functions are supposed to be optimized differently.

nikomatsakis:
@arielb1 

> See how the lifetimes sneak back in? I don't see any way to have a workable model without lifetimes - for example, the only difference between rvo_safe and no_rvo_safe (which I added now) is the lifetime bounds in as_ptr vs. id, yet the functions are supposed to be optimized differently.

Named lifetimes in the fn signature are a totally different beast than random lifetimes that result from inference. I agree you can't "excise" them completely in some sense. They are the source of the permissions -- and of course if you continue to use memory you obtained from a reference after the fn returns, you need some way (e.g., mentioning the named lifetime in your type) to show that you are still within that lifetime.

Ericson2314:
@arielb1 This is such an unimportant nit I almost feel bad bringing it up:

> let local = &move *alloca();.

I believe you want a scratch pointer since `local` is uninitialized at lifetime begin, and must be uninitialized at lifetime end.

ubsan:
@Ericson2314 could you explain further?

Ericson2314:
@ubsan `&move` is used for moving data out, so at the beginning of the lifetime it must point to an initialized location, and at the end of the lifetime, since the data is moved out, it must point to an uninitialized location.

`&uninit` or `&scratch` would be a pointer type where at *both* the beginning and end of the lifetime, the location pointed to must be uninitialized. The name "scratch" invokes the idea that this pointer gives access to scratch space / empty buffer / playpen etc. It is the proper return type for `alloca`, and encapsulated in a newtype for `Drop` to call `free`, it is the proper return type for `malloc`.

For more information see https://github.com/rust-lang/rfcs/pull/1617

ubsan:
@Ericson2314 Ah. We've taken to calling it `&out`.

`&move` can point to uninitialized memory (i.e., moving in and out, like in a local, would be allowed), and `&out`'s semantics are not well known enough.

Ericson2314:
mm I thought the name `&out` referred to the pointer which starts uninitialized and must be initialized before it is dropped / lifetime ends. This one alone of the 4 possible is a linear type, but other than that I think all 4 are pretty understood.

-----

Regardless all for readily fit the proposal of @arielb1 that I was quoting, hence why I felt bad bringing them up.

ubsan:
@Ericson2314 Ah, got it. Yeah, see, `&uninit` is definitely not well understood, I didn't understand what you were talking about ;P.

Anyways, `&move` does fill that hole; it has two states, just like a local variable (at least in my model of `&move`), "moved out", and "moved in". This would just be starting the reference in the "moved out" state.

arielb1:
@ericson2314

This would be a better approximation:
```Rust
let local = &move *alloca();
mem::forget(*local); // mark `*local` as uninitialized
```

However, the presence/absence of the `mem::forget` does not affect the aliasing semantics, so I omitted it.

Ericson2314:
> it has two states, just like a local variable

Well, there needn't be any extra initialized/uninialized "state" because 4 is enough pointer types to make it first-class https://github.com/rust-lang/rfcs/pull/1617#issuecomment-220697500. 

> This would be a better approximation:

With the above, the `forget` call wouldn't even be legal, unless we want "rvalue-initalized-ness" and `forget` to be polymphic over such a thing.

> PS: at the MIR level, &move is just as linear as &out - drops do not insert themselves, after all.

Yes indeed! This symmetry is much nicer, and I love that I can say "well, we *already* have them!" on the next linear types proposal.

Ericson2314:
Back on topic, I'm still confused on ZST-deref being unsafe. IMO, `*x: ()` should be safe, `*Void: ()` should not. This is one of the reasons I want Void-iso types to have size -∞.

arielb1:
> With the above, the `forget` call wouldn't even be legal, unless we want "rvalue-initalized-ness" and forget to be polymphic over such a thing.

I want `Box<T>` to be implementable as `struct Box<T>(&'static move T)` up to it not requiring `T: 'static` due to WF. You *can* `mem::forget` out of a box.

> Back on topic, I'm still confused on ZST-deref being unsafe. IMO, `*x: ()` should be safe, `*Void: ()` should not. This is one of the reasons I want Void-iso types to have size -∞.

The unsafety rules take effect *before* monomorphization, so the size of types is not known yet. I think that deref of a unit type should always be a no-op (never UB) on a non-null pointer/reference, and that a deref of an empty type should be an `intrinsics::unreachable()`.

Ericson2314:
> I want Box<T> to be implementable as struct Box<T>(&'static move T) up to it not requiring T: 'static due to WF. You can mem::forget out of a box.

>That looks like killing a fly with a nuke. We just need some control over the internal drop-flag - a shallow_drop - that could be unsafe too.

I suppose we can add `&move` before the others and then that will make sense. Empty boxes are unsafe, but it's a step in the right direction.

> The unsafety rules take effect before monomorphization, so the size of types is not known yet. I think that deref of a unit type should always be a no-op (never UB) on a non-null pointer/reference, and that a deref of an empty type should be an intrinsics::unreachable().

I think that's in agreement with everything I've said? There many (safe) things we can and should do with generic code that assume >0 size, such as allowing `match x: Option<A> { Some(x) => ..., ...}` even though if `A` was `Void` we'd have a dead match arm.

ubsan:
@arielb1 @Ericson2314 The point, at least for me, of &move is that it gets rid of `Drop::drop` dropping inner fields that it derefs to. See the codegen section of https://gist.github.com/ubsan/9d79ad4299870faf8c7b

arielb1:
@ubsan 

So have drop glue call `DerefMove` to get the interior to drop? That... could work. 

OTOH, you ought to be able to do partial drops, which would make `DerefMove` unsafe:
```Rust
struct Foo(Box<u32>, Box<u32>);
fn main() {
    let x = Box::new(Foo(Box::new(0), Box::new(1)));
    drop(x.1);
    // drop glue:
    let d = DerefMove::deref_move(&mut x); // unsafe! can access `x.1`!
    drop(d.0);
    mem::forget(d.0);
    Drop::drop(&mut x);
}
```

Ericson2314:
@ubsan and I had talked a bit about `&move` on IRC, but I am wary of extending the language if we are *still* lift with (different) unsafety/magic. So yes `&move` alone is still useful (e.g. for drop in place) but we ought to hold of on `DerefMove` until it can be done right.

My "4 types of unique borrowed pointer" proposal is really less about pointers and more about making "initializedness" first-class---part of the type system instead some extra static analysis tacked onto borrow checking. Once you have that, the generalized pointer types are fairly inevitable.

ubsan:
@arielb1 well, in that case, it's not unsafe :)

Notice how you're `deref_move`ing in `drop(x.1)`. That would turn into

```rust
let inner = x.deref_move();
drop(inner.1); 
// drop glue
drop(inner.0);
Drop::drop(&mut x);
```

ubsan:
So there's gotta be some sort of guarantee about only being called once, I'm thinking... something like that.

nikomatsakis:
I have just opened https://github.com/rust-lang/rfcs/pull/1643, which proposes that rather than adopting one RFC, we adopt a "strike-team-based" approach to work these rules out in a more systematic way.

arielb1:
@ubsan 
\
I think we want the `DerefPure` unsafe trait, and to not allow using `DerefMove` unless it is implemented.

ubsan:
@arielb1 

But why? DerefMove seems fine if we just... don't call DerefMove multiple times. I think it's actually a really nice guarantee that DerefMove doesn't get called multiple times, just like Drop.

ubsan:
This allows for things like dropping in DerefMove, since it's only called once. If, for example, you have a cache, and you don't need it once you're moved out of, it would be nice to get rid of it once you're no longer useful; and, I don't see the point of requiring DerefPure. It just makes it more annoying to implement DerefMove for users.

RalfJung:
@Ericson2314 

> Back on topic, I'm still confused on ZST-deref being unsafe. IMO, *x: () should be safe, *Void: () should not. This is one of the reasons I want Void-iso types to have size -∞.

Most of the discussion here is well beyond the scope currently covered by my formalization, but this one part is actually already "in scope" and I have definitions which I think make sense. So I will try to share my formal thoughts on this. Funny enough, the outcome is exactly the opposite if what you suggest.

Let me try to explain the relevant parts of my model of types:
A type is a *set of lists of values*. Values are, for example, integers, addresses (i.e., memory locations), or booleans. These lists describe the layout of the type in memory. This is a very abstract model that does not distinguish integers of different size, or the fact that integers actually overflow. All these issues can be modeled faithfully,  but they are mostly irrelevant for what I am interested in right now, so I am making my life simpler.
For example, the type `i32` is the set of singleton lists whose only element is an integer.
The type `struct { i32, bool }` is the set of two-element lists, the first element being a number and the second being a boolean. The type `()` is the set consisting only of the empty list. The type `Void` is the empty set. Every type has a size, which is the length of the lists in the set. (They all have to have the same length -- no unsized types so far.) Clearly, the size of `()` must be 0. The size of `Void` can be anything, since there is *no* list in that set.

The type `&mut 'a T` is (very roughly speaking) the set of singleton lists whose element is a location `l`. Furthermore, the locations in `[l, l+T.size)` (left-inclusive, right-exclusive) have to be valid, allocated addresses that we can access and mutate for the duration of lifetime `'a`, and the list of values stored at these addresses must be in `T`. (I am using mutable borrows here because they are way simpler than shared borrows. I hope to write a blog post some point on why that is, I "just" need to come up with a nice explanation...)

So, if we consider `&mut 'a ()`, what do we have? We have a location `l` such that `[l, l+0)` are valid addresses, and the list of values stored at these locations is in the set representing `()`. This interval is empty. Hence we do actually not know anything about the location, so in particular, we must not dereference it.

Let's consider `&mut 'a Void`. We have a location `l` such that `[l, l+Void.size)` are valid addresses, *and* the list of values stored at these locations is in the set representing `Void`. That's the empty set. There cannot be anything in the empty set, so we have a contradiction. From this, we can derive anything, and hence in particular, we are allowed to dereference the pointer.
What I am saying here, essentially, is that `&mut 'a Void` is actually *the same type* as `Void` in the sense that there is no possible value of that type, and hence we can always assume during compilation that no value of type `&mut 'a Void` exists. (This is actually assuming that `'a` is an active lifetime.) Because of this, you can do literally anything if you have a value of type around; that's unreachable code.

eternaleye:
@RalfJung But the topic was `*mut Void`, not `&'a mut Void` - does the same reasoning apply? If so, wouldn't that make using it for FFI pointers violently invalid?

eddyb:
@RalfJung I believe there's a distinction to be made here that `&mut Void` is `Void` but `*mut Void` isn't.
Specifically, one can create a `*mut Void` in safe code, but dereferencing it (in unsafe code) creates a value of type `Void`, which can make reachable code behave as if it wasn't (literally "invoking" UB).

About ZSTs, I believe you're talking about memory reads, not *syntactical* dereferences, which do not touch memory at all to "read" a ZST.
Sorry if I sound like a broken record, but this distinction in general can lead to a lot of confusion, because `*ptr` is a memory no-op in C, C++ and Rust (the latter 2 only if not overloaded) in that it converts a pointer rvalue into an lvalue while performing no memory access, but even if, say, `&*ptr == ptr` can be considered to be trivially `true`, AFAIK *syntactical* dereference can be UB in C if `ptr == NULL`.

Ericson2314:
@RalfJung still reading the rest of your post, but btw I totally agree that `&mut` is way easier to understand deeply. In https://internals.rust-lang.org/t/a-stateful-mir-for-rust/3596/15 (which you might be interested in anyways :)) I have a pretty good idea of what I will write for `&mut` but much less so for `&`.

Ericson2314:
Ah OK, so even besides @eddyb's concerns I consider a *load* from `&mut ()` to be safe because that load must have size 0, so the load is actually a no-op. Totally agree `&mut Void` is absurd.

Also I'd like to point out that @RalfJung's model is another good reason why the conflation of ()-isomorphic and Void-isomorphic types as both having size 0 is a dangerous thing to do.

ubsan:
@Ericson2314 To be specific, lvalue->rvalue conversion for a `()` is a no-op, whereas lvalue->rvalue conversion for a `!` is undefined behavior (I believe I can start using `!` now as the `Void` type, it seems that will go through).

RalfJung:
@eternaleye 
> But the topic was *mut Void, not &'a mut Void - does the same reasoning apply? If so, wouldn't that make using it for FFI pointers violently invalid?

Oh, I see. As far as I can tell, Rust attaches absolutely no guarantees to raw pointers, so their dereferencability depends entirely on the meaning you assign to them. If your invariants ensure that the given `*mut T` points to X bytes of valid memory, then sure you can make use of that. Of course, if you put the result into a variable of type `T`, you have to make sure it is a valid `T`. So, you better don't dereference the `*mut Void`, and (naturally) there is a manual proof obligation when creating a `*mut Void` that this pointer actually satisfies the guarantees you'd like to attach to it.

@eddyb 
> Sorry if I sound like a broken record, but this distinction in general can lead to a lot of confusion, because *ptr is a memory no-op in C, C++ and Rust (the latter 2 only if not overloaded) in that it converts a pointer rvalue into an lvalue while performing no memory access, but even if, say, &*ptr == ptr can be considered to be trivially true, AFAIK syntactical dereference can be UB in C if ptr == NULL.

Oh, the good ol' lvalue-rvalue confusion. I am entirely sidestepping that issue in my formalization by not having lvalues, which is why I payed no attention to it. ;-) Anyway, when I wrote about dereferencing pointers above, I was talking about *actual* load operations that happen on the machine.

@Ericson2314 
> Also I'd like to point out that @RalfJung's model is another good reason why the conflation of ()-isomorphic and Void-isomorphic types as both having size 0 is a dangerous thing to do.

I am not sure that is the case. ;-) Actually, I did not even assume any particular size for `!` above -- the reasoning works for any size. Truth is, the size of `!` could be literally anything, it just doesn't matter as the type is not inhabited. I doubt it is worth complicating the algebra of sizes for this corner-case. It would be nice to have things like `Option<!>` having the same representation as `()`, but that won't fall out of any particular size for `!` -- instead, that'd be an extension of the optimizations for `enum` layout that we already have (e.g., `Option<&T>` having the pointer size, without an added discriminant).

@ubsan 
> To be specific, lvalue->rvalue conversion for a () is a no-op, whereas lvalue->rvalue conversion for a ! is undefined behavior (I believe I can start using ! now as the Void type, it seems that will go through).

I would say that lvalue -> rvalue conversion for `!` assumes that we have a `!`, so this cannot even happen. That's not UB, that's just impossible -- the UB happened earlier, when we obtained a value of type `!`. But the consequence is the same, the compiler is free to emit literally any assembly code when anything is done to a `!`.

ubsan:
@RalfJung Ah, that comment wasn't supposed to go through :)

The issue is raw pointers, where `*mut !` is a valid type, despite being an lvalue to an impossible rvalue.

RalfJung:
@ubsan Right, as long as you don't dereference it. Whether `&*x` is already UB, I am not so sure... I would tend to argue it should not be.

glaebhoerl:
Quoting my [comment](https://github.com/rust-lang/rfcs/pull/1216#issuecomment-224880355) from #1216 with respect to using `*mut !` for things like FFI types:

> Given that the contract of `*const T`/`*mut T` is that it shall only be dereferenced when it actually points to valid/live data (the type system can't know this, which is why it's `unsafe`; it's entirely up to the user to determine this): I think `*mut UninhabitedType`, with the understanding that it will never be dereferenced, and `*mut UnitType`, with the understanding that it can be dereferenced whenever but there's no point in doing so, make equal amounts of sense.
>
> (In other words, this is unlike `&mut UninhabitedType`, which is logically equivalent to `UninhabitedType` itself, because `*mut` doesn't imply liveness.)

arielb1:
@RalfJung 

I am not sure about `&*bot`, but a use (aka lvalue-to-rvalue conversion) of a `Void` is UB.

RalfJung:
Agreed. I'd argue that `&*x` is a noop (when translating to an lvalue-free semantics, it compiles to `x`). But if an actual memory access happens, we get UB.

ubsan:
@RalfJung But the having `&!` is UB, so `&*x` is undefined behavior, because it creates an `&!`.

drbo:
Here are some of my own thoughts (not an expert, but am working on a garbage collector):

- The memory model needs to be formal.  Otherwise you can't formally verify unsafe Rust code, which is a use case that (at least) Redox definitely wants.
- We need to avoid falling into the trap that C/C++ fell into with strict aliasing, which as I understand it breaks so much code that major compilers implement `-fno-strict-aliasing` and major projects rely on this (the OCaml runtime, CPython 2.x, and the Linux kernel, as well as the output of an RPC compiler at a minimum), and yet as I understand it has only minimal performance gains in practice for typical code (i.e. not tight numeric loops).
- The memory model needs to be easy for people to understand.  That means that an equally normative, easy to understand prose description of the memory model needs to be present as well – it is, after all, what most people will be programming to.  If the formal model and the prose model diverge, this is a bug in the spec that must be fixed.
- It needs to be possible to do what needs to be done without unnecessary copies.  In C++, the only standards-conforming way to pass the buffer stored in an `std::string` to a function that takes an `unsigned char *` is to _copy the entire buffer!_.  This is clearly absurd.
- The memory model needs to support writing memory allocators, garbage collectors, etc. without too much boilerplate or pitfalls beyond what C would have.  If the Rust version of some code is twice as long as the C version (with `-fno-strict-aliasing`) and is much harder to read, then that is a problem with Rust.
- Aliasing `*mut` pointers should be permitted.  Furthermore, it should be possible for an `&mut` or `&` pointer to alias a `*const` or `*mut` pointer, so long as the pointee does not change (in the case of a `&`-pointer) or is only modified through the reference (in the case of an `&mut`) for as long as the reference is live.
- In multithreaded code, a common type of data race is two threads racing to write the same value to the same address, with no racing reads.  Such races should be allowed, as the operation of writing a value to an address is idempotent.  Similarly, the case where two threads simultaneously read a value from an address, set and/or clear a bit, and then write the new value back should be allowed – again, no reordering that I can imagine would break the code, and I have read about several algorithms (garbage collectors, if I recall correctly) that rely in this to work.

nrc:
Nominating for discussion at lang-team meeting - the unsafe guidelines team exists now, so we should consider moving this PR either somewhere else or just the team tag.

nikomatsakis:
I agree with @nrc that, given that we have accepted https://github.com/rust-lang/rfcs/pull/1643, we should close this RFC, since its contents are subsumed by the discussion taking place at the unsafe guidelines repo.

I reviewing the thread and have tried to extract out the most notable questions that were debated. It would make sense to move many of these to issues on the [rust-memory-model repo](https://github.com/nikomatsakis/rust-memory-model), I think:

- [how to handle uninitialized data](https://github.com/rust-lang/rfcs/pull/1578#issuecomment-217223926) and [whether some CPUs will require loading of undef to be UB](https://github.com/rust-lang/rfcs/pull/1578#issuecomment-217922438)
- [some discussion of temporary elimination in MIR](https://github.com/rust-lang/rfcs/pull/1578#issuecomment-222329531)
- [an alternative aliasing model jointly proposed by @arielb1 and @ubsan](https://github.com/rust-lang/rfcs/pull/1578#issuecomment-222530225)
- [some thoughts on ZST interactions](https://github.com/rust-lang/rfcs/pull/1578#issuecomment-225979604)

The discussion on the RFC has focused on a number of topics, but surprisingly little on the *meat* of the rules here. 
- how much and what to formalize
- how well-specified LLVMs docs are (not much)
- a reframed proposal by @arielb1 and @ubsan 

- [x] @nikomatsakis
- [x] @nrc (I took the liberty of pre-checking, given the previous comment)
- [ ] @aturon 
- [ ] @eddyb 
- [x] @pnkfelix (pre-checked, since he is on vacation)

ubsan:
I'm just going to close it now, since I owns it :P

PR #1590, opened 2016-04-22
Add a `lifetime` specifier to `macro_rules!`

sgrif:
[Rendered](https://github.com/sgrif/rfcs/blob/sg-macro-lifetimes/text/0000-macro-lifetimes.md)

### DIFF
--- a/text/0000-macro-lifetimes.md
+++ b/text/0000-macro-lifetimes.md
@@ -28,9 +28,12 @@ separators.
 [design]: #detailed-design
 
 This RFC proposes adding `lifetime` as an additional specifier to
-`macro_rules!` (alternatively: `life` or `lt`). Since a lifetime acts very much
+`macro_rules!` (alternatively: `life` or `lt`). As it is a single token, it is
+able to be followed by any other specifier. Since a lifetime acts very much
 like an identifier, and can appear in almost as many places, it can be handled
-almost identically. A preliminary implementation can be found at
+almost identically.
+
+A preliminary implementation can be found at
 https://github.com/rust-lang/rust/pull/33135
 
 # Drawbacks
--- a/text/0000-macro-lifetimes.md
+++ b/text/0000-macro-lifetimes.md
@@ -1,0 +1,49 @@
+- Feature Name: Allow `lifetime` specifiers to be passed to macros
+- Start Date: 2016-04-22
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add a `lifetime` specifier for `macro_rules!` patterns, that matches any valid
+lifetime.
+
+# Motivation
+[motivation]: #motivation
+
+Certain classes of macros are completely impossible without the ability to pass
+lifetimes. Specifically, anything that wants to implement a trait from inside of
+a macro is going to need to deal with lifetimes eventually. They're also
+commonly needed for any macros that need to deal with types in a more granular
+way than just `ty`.
+
+Since a lifetime is a single token, there is currently no way to accept one
+without an explicit matcher. Something like `'$lifetime:ident` will fail to
+compile.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposes adding `lifetime` as an additional specifier to
+`macro_rules!` (alternatively: `life` or `lt`). Since a lifetime acts very much
+like an identifier, and can appear in almost as many places, it can be handled
+almost identically. A preliminary implementation can be found at
+https://github.com/rust-lang/rust/pull/33135
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
+
+# Alternatives
+[alternatives]: #alternatives
+
+A more general specifier, such as a "type parameter list", which would roughly
+map to `ast::Generics` would cover most of the cases that matching lifetimes
+individually would cover.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
--- a/text/0000-macro-lifetimes.md
+++ b/text/0000-macro-lifetimes.md
@@ -18,9 +18,11 @@ a macro is going to need to deal with lifetimes eventually. They're also
 commonly needed for any macros that need to deal with types in a more granular
 way than just `ty`.
 
-Since a lifetime is a single token, there is currently no way to accept one
-without an explicit matcher. Something like `'$lifetime:ident` will fail to
-compile.
+Since a lifetime is a single token, the only way to match against a lifetime is
+by capturing it as `tt`. Something like `'$lifetime:ident` would fail to
+compile. This is extremely limiting, as it becomes difficult to sanitize input,
+and `tt` is extremely difficult to use in a sequence without using awkward
+separators.
 
 # Detailed design
 [design]: #detailed-design


text/0000-macro-lifetimes.md
```diff
@@ -0,0 +1,49 @@
...
+
+Certain classes of macros are completely impossible without the ability to pass
+lifetimes. Specifically, anything that wants to implement a trait from inside of
+a macro is going to need to deal with lifetimes eventually. They're also
+commonly needed for any macros that need to deal with types in a more granular
+way than just `ty`.
+
+Since a lifetime is a single token, there is currently no way to accept one
+without an explicit matcher. Something like `'$lifetime:ident` will fail to
+compile.
```
bluss: lifetime tokens (`'xyz`) are matched by `tt`. So just like the literal matcher RFC, this caveat should apply, that it doesn't actually add much power just ease of use to macros by example. rust-lang/rfcs/pull/1576

https://github.com/rust-lang/rfcs/pull/1576#issuecomment-208322492

sgrif: Ah good point. I would argue it does still add a good bit of power in the context of sequences. I will amend the RFC to note this, however.

text/0000-macro-lifetimes.md
```diff
@@ -0,0 +1,51 @@
...
+by capturing it as `tt`. Something like `'$lifetime:ident` would fail to
+compile. This is extremely limiting, as it becomes difficult to sanitize input,
+and `tt` is extremely difficult to use in a sequence without using awkward
+separators.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposes adding `lifetime` as an additional specifier to
+`macro_rules!` (alternatively: `life` or `lt`). Since a lifetime acts very much
```
bjz: `lt` might be confused with '`literal`', if it were ever added.

Also, perhaps these alternatives could be moved to the `# Alternatives` section?

text/0000-macro-lifetimes.md
sgrif: As it turns out, you actually can't use a `tt` in a lifetime position:

```
 error: expected identifier, found `'a`
 impl<$($lifetime),*> SomeTrait
```

text/0000-macro-lifetimes.md
sgrif: Are differing names an alternative to the RFC? Seems like it's a detail of the RFC itself. And yeah, I'm strongly in favor of `lifetime` as the name

text/0000-macro-lifetimes.md
bluss: tts don't expand anywhere without the ast coercion that danielkeep was referring to.

bluss: Don't get me wrong, I think this is a good addition, good RFC. But I don't think it gives macro_rules powers it didn't already have.

bluss: Here's the reparse trick. https://play.rust-lang.org/?gist=3287a8aab3bea913916f56d12c7a00ad&version=stable&backtrace=0

durka:
I know it's in the implementation, but the follow set for the matcher is probably a detail that should be in the RFC (and possibly included as a delta to RFC 550 as well).

text/0000-macro-lifetimes.md
bluss: I agree that the long word `lifetime` seems good. I don't see a reason to save letters here.

text/0000-macro-lifetimes.md
sgrif: Gotcha. Thanks for the link. I'm also interested because I would like to use this in Diesel as a workaround until this (hopefully) is accepted and in stable

Diggsey:
I wrote a PR for this a while back and @alexcrichton mentioned that `lifetime` might become a keyword, which would cause problems. Is that still the case?

durka:
I don't think we can really make new non-contextual keywords anymore. And I don't see the ambiguity anyway if there were to be a matcher with the same name as a keyword.

nikomatsakis:
I'm nominating this for discussion as a "fast track for FCP" thing. However, I also think we should add a note about the follow-set (though it is mostly implied by the comparison to identifier).

sgrif:
Updated to mention the follow-set

pnkfelix:
(another potential way to denote the bound could be `'ident`. As in `$lifetime:'ident`)

nikomatsakis:
We discussed this RFC in the previous @rust-lang/lang meeting. We all agree this is a pretty harmless addition, but for one concern: there remains some lingering concern among us that the term *lifetime* may not have been the best choice. Specifically, there is this ambiguity between the *lifetime of a value* -- i.e., when the destructor will run, which I tend to call the value's *scope* -- and the *lifetime of a reference* -- the region of the code where the reference will be used. To make matters just a bit more confusing, these two notions of lifetime are interrelated -- specifically, you can't make a reference to a value for longer than the value's scope. 

Up till now the term reference, while firmly entrenched in our documentation and blog posts, hasn't been in the *language itself* (though it does appear in the associated items RFC; but that was never implemented -- something we ought to fix, but then of course we'd run into the same problem). That said, this may well be a distinction without a difference.

I have to admit I feel pretty terrible holding up this otherwise unobjectionable RFC on this matter, since it's kind of an amorphous concern with no clear resolution. This is one reason we were brainstorming other possible syntaxes, such as `'ident` (or maybe even `ident`?).

I was thinking now though that there is though another possible way to view things which might make the name *lifetime* perfectly acceptable. That is, one could coin a term representing the region of the code where a reference is used -- candidates might be calling it the reference's region, or extent, or the duration of a borrow. In that case, the term lifetime is kind of a "superset" of these things: a scope is a lifetime, but so is a region. Not sure yet if that's a useful way to think about things.

Anyway, those were our concerns.

durka:
This is about syntax. Everyone calls the `'a` syntax a "lifetime" or a "lifetime annotation". This includes libsyntax :)

`$x:'ident` would be fine too, and avoid naming considerations!

durka:
Amusingly, the macro `macro_rules! m { ($x:'ident) => {} }` parses today, and there's no error until you try to invoke it.

bluss:
Maybe this fact can inspire new names: Loop labels use the same syntax.

sgrif:
I definitely think 'ident would be fine as well, but I'm not sure how
painful that would be for implementation.

bluss:
`lifetime` is good (named for a lifetime parameter), you could also call it just `label` (named for a loop label). I think using `'ident` is too irregular, and would be another oddity in an already odd macro_rules world.

nikomatsakis:
Yes, this was intentional, because we intended to (and I think still
vaguely intend) allow users to manually label blocks as a way to provide
explicit lifetime names. For example:

```rust
let x;
let y;
'a: {
    x = &'a y; // (error)
}
```

durka:
Is there any hope of un-stalling this? Several ideas have been proposed to solve the trivial naming issue which is blocking the RFC (ignore it and use `:lifetime`, use `':ident` instead, make up a new name like `:label` or `:region`). Does the lang team like any of them or shall we continue painting the bike shed?

bluss:
`lifetime` is short for *lifetime parameter*. Does the lang team want to coin a new name for that one? Otherwise i think `lifetime` should just be the name.

pnkfelix:
@durka I talked with @nikomatsakis last week about this. I suspect at this point that we are likely to bite the bullet and just accept `:lifetime` here. I think we agreed that the term "lifetime" is too entrenched at this point for us to expect to actually *change* it across the board.

Renominating for discussion at lang-team mtg.

nikomatsakis:
We discussed in the @rust-lang/lang meeting. We came to the conclusion that we should just go with the term "lifetime". Hence we are nominating this for final comment period.

My personal feeling is that I would like to have three terms:

- "lifetime" -- means some span in the code; a portion of the control-flow-graph;
- "scope" -- the lifetime corresponding to a lexical scope or statement; this corresponds to when things can be named, but also when destructors will run
- "extent" (maybe) -- the lifetime corresponding to where a reference is used. When you borrow some value, that borrow has an extent which is defined by how long the resulting reference (and any references derived from it) wind up being used.

Right now we tend to call all 3 of these lifetimes, so I am roughly proposing keeping that in place, but adding some more specific terms for two distinct concepts. I am not quite sure what that third term ought to be, though: I vacillate between "extent", "region", "duration", and a few other such words. I hope that in most docs and error messages though we can get away with not needing to use it, so maybe it's just a term that we use when trying to explain the type system at a more advanced level.

nikomatsakis:

durka:
FCP should be over soon, yeah?

aturon:
@durka 

Yes -- the lang meeting wound up cancelled last week, but we should have a final decision on Thursday. Sorry or the delay!

durka:
This is getting a bit ridiculous.

nikomatsakis:
Actually, the problem is my fault. The @rust-lang/lang team *did* discuss this and decided to accept it some time back (a week or so) -- but i've been overwhelmed and failed to follow up on my clerical duties! I do apologize

nikomatsakis:

durka:
Woohoo!

PR #1608, opened 2016-05-07
clarify 'root' with 'root crate' or 'root `Cargo.toml`'

liigo:
Currently, it is not clear if the 'root' is a file or directory, and so the meaning of "the `target` directory next to the root" is ambiguous.

Note that 'root crate' is not new words, it appeared in the origianl RFC text.


alexcrichton:
Thanks!

PR #1622, opened 2016-05-19
Add a compiler flag that emits crate dependencies on a best-effort basis

mahkoh:


mahkoh:
The implementation of (a basic version of) this is almost trivial and I've done it myself in my compiler driver.

This can then be used as in the following python script: https://gist.github.com/anonymous/a12206da17ab36f0c30d054ba5b05e89

When a new compilation unit is added, the only thing that has to be done is to add its name to the top of the file. If the dependencies between the units change, nothing has to be done because the script tracks those dependencies itself.

text/0000-emit-crate-deps.md
```diff
@@ -0,0 +1,62 @@
...
+[motivation]: #motivation
+
+When working with projects that consist of dozens or hundreds of compilation
+units, it is useful to have some kind of automatic dependency management in the
+build system. Otherwise such information has to be maintained in at least two
+locations: Once in the compilation unit itself via an `extern crate X` statement
+and once in the build system.
+
+To this end, this RFC proposes a new compiler flag which emits the names of the
+crates a compilation unit depends on to a text file.
```
oli-obk: Since we might want to add more information than just the crate name, I suggest to produce a json file (or any other structured format), which also yields the attributes given to the `extern crate` item. This is important for things like macro imports and compiler plugins.

text/0000-emit-crate-deps.md
```diff
@@ -0,0 +1,62 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If `-Z parse-only` is passed, the created file is, in general, not precise. This
+is because macro expansion can itself create new `extern crate` statements.
+However, macro expansion does, in general, depend on the dependencies having
+already been compiled.
+
+The `-Z parse-only` variant is, however, essential for the first round of
+dependency resolution in the build system.
```
oli-obk: Please elaborate in the detailed design why this is so

text/0000-emit-crate-deps.md
```diff
@@ -0,0 +1,62 @@
...
+
+I assume that the generation of `extern crate` statements is a rarely used
+features. If it is used, the problem can be mitigated by tracking only the
+generated statements separately in the build system.
```
oli-obk: Every macro that takes an expression can potentially contain `extern crate` items in the expression (if it contains a block). for example:

```rust
println!("foo: {:?}", { extern crate chrono; chrono::UTC::now() } );
```

text/0000-emit-crate-deps.md
mahkoh: What's unclear about it?

text/0000-emit-crate-deps.md
mahkoh: Yes, that's what it says in this paragraph.

oli-obk: I would include that in the detailed design, instead of the drawbacks. I don't think your system should be best-effort, but either do this for all possible rust code or not at all

text/0000-emit-crate-deps.md
oli-obk: You are just stating that it is essential, not why.

text/0000-emit-crate-deps.md
oli-obk: This can be used for the first iteration, since you can now check for `#[macro_use]` attributes.

text/0000-emit-crate-deps.md
mahkoh: A build system needs a dependency graph to know where to start compiling.

text/0000-emit-crate-deps.md
mahkoh: That is impossible as I've explained in this section.

text/0000-emit-crate-deps.md
oli-obk: That's what I meant with `add a workflow of calls` (maybe the motivation would be a better place than the design)

1. rustc -Z parse-only -Z emit-crate-deps $file
2. find dependencies, run this entire process on the dependencies
3. rustc -Z no-analysis -Z emit-crate-deps $file
4. find new dependencies, run this entire process on the new dependencies
5. compile $file by passing the dependencies

make the life easier for readers of your RFC

text/0000-emit-crate-deps.md
oli-obk: sure it is. The macro expansion -> new dependency game can't be continued forever, even inside rustc + cargo with explicitly stating the dependencies:

```rust
// main.rs in blub
#[macro_use]
extern crate blarg;
fn main() {
    println!("{:?}", blarg!());
}
```

```rust
// lib.rs in blarg
#[macro_export]
macro_rules! blarg {
    () => {{
        #[macro_use] extern crate foo;
        foo!();
    }}
}
```

```rust
// lib.rs in foo
#[macro_export]
macro_rules! foo {
    () => {5}
}
```

and you get an error while compiling `blub`:

> macro undefined: 'foo!'

mahkoh: 1. Macro expansion is, in general, not deterministic so it's impossible to give precise results.

2. You've already given an example where your algorithm fails for different reasons.

text/0000-emit-crate-deps.md
mahkoh: The process is not part of the RFC.

text/0000-emit-crate-deps.md
oli-obk: > Macro expansion is, in general, not deterministic so it's impossible to give precise results.

Well it is deterministic between runs if there are no compiler plugins involved.

> You've already given an example where your algorithm fails for different reasons.

I have? Everything that works right now would still work with my algorithm, because rustc cannot handle macros that expand to `exter crate` items that add new macros.

text/0000-emit-crate-deps.md
oli-obk: it can be part of the motivation, so the detailed design can be understood.

text/0000-emit-crate-deps.md
mahkoh: >I have?

Yes, steps 3 and 4 must be run in a loop since macro expansion can reveal new dependencies which can allow more macros to expand which reveal new dependencies etc.

I don't think it's worth discussing a more complicated process that allows the compiler to emit such a file once it encounters the first error at this time. I have no idea how such a process would be implemented but I know that the process I described can be implemented in about 100 lines of code. And while such simple code cannot solve the 100% case, it can solve the 99% case and allows the other 1% to be handled downstream.

It's also not productive spending more time writing comments here until those who rule over us have signaled whether they might add this feature (in one form or the other) or not. Like I said, I'm already using this feature in my driver and maintaining 100 loc every two months isn't that much work.

oli-obk: As I showed in my example, rustc doesn't allow macros to add dependencies that add macros, so there's no need to support it in a build system.

mahkoh: This is with the current macro system. Does this still hold with the new one? From what I know, the new macro system will perform macro resolution lazily which would allow macros to add new dependencies.

mahkoh: Buf if you're working under the assumption that it does not, then I don't see the point of your comments since the system suggested in this RFC already supports the 100% case via your algorithm.

mahkoh: The current behavior seems to be considered a bug by some: https://github.com/rust-lang/rust/issues/33936

nikomatsakis:
One obvious thing: the flag should not be `-Z`, but something else. `-Z` is for unstable functionality.

nrc:
This RFC is entering its :bell: final comment period :bell:

nikomatsakis:
Also, I would rather not have a separate `-Z parse-only` flag, but rather combine the two things into one flag.

text/0000-emit-crate-deps.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Add a new compiler flag `-Z emit-crate-deps`. If the flag is passed, the
+compiler emits a file `<crate name>.crate_deps` which contains a list of all
+crates the compilation unit depends on.
+
+If the flag `-Z parse-only` is passed, the file is emitted immediately after
+parsing. Otherwise it's emitted after macro expansion.
```
nrc: Why is this detail important?

nrc: Ah, nevermind, it is explained below.

nrc:
CC @rust-lang/tools 

Nominating for the tools team to have a looks at this.

nrc:
To summarise, this seems like a desirable and small feature to add - emit a list of deps for the current crate. However, the details are a little concerning, in particular the need for a 'best effort' pre-macro expansion version feels bad to me (and @oli-obk by the sounds of it). I'm not sure if there is a good solution for that? What does Cargo do?

brson:
Seems like a useful feature, but as-is the details are a bit light.

I don't understand the need for 'best effort' resolution. The only motivation for it I see is this:

> The -Z parse-only variant is, however, essential for the first round of dependency resolution in the build system.

What does this mean? What are tools going to use pre-expansion resolution for? Why can't they tolerate doing the expansion? Please expand the main text to include this information.

The model for building this is proposing is backwards of Cargo: instead of the build system telling rustc its dependencies, it is asking rustc for its dependencies. This may be important for non-Cargo build systems but they are necessarily going to get less fidelity about the deps (all rustc knows is crate names).

What exactly is in the crate_deps file? The RFC says "a list of all crates". Are these the idents of the crates as contained in the `extern crate` statement? Please make it explicit. For example, it could also conceivably be paths to crate files.

This feature has parallels to `--emit dep-info` so it seems like it probably makes the most sense to call this `--emit crate-deps`. The names are very similar though so it would probably pay to brainstorm more distinct names.

text/0000-emit-crate-deps.md
```diff
@@ -0,0 +1,62 @@
...
+statements. If the statement is of the form `extern crate X as Y`, `X` but not
+`Y` is added to the list.
+
+The file contains the names of the crates separated by `\n`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If `-Z parse-only` is passed, the created file is, in general, not precise. This
+is because macro expansion can itself create new `extern crate` statements.
```
nikomatsakis: Also, macro expansion may remove `extern crate` statements, naturally. (Since it e.g. applies `#[cfg]` directives.)

nikomatsakis:
> The build model this is proposing is backwards of Cargo: instead of the build system telling rustc its dependencies, it is asking rustc for its dependencies. 

I wanted to bring this up. The cargo model has many advantages -- for example, it allows one to process the dependencies without inspecting the source, and sidesteps the problems around macro expansion -- which is why it was adopted in the first place. It does seem a bit odd for build systems to "swim against the grain" (to mix a few metaphors) by extracting "extern crates" rather than adopting a Cargo-like alternative where dependency information is extracted into a manifest (this connects to my desire to allow `extern crate` declarations to be elided entirely when they are supplied on the command line).

Also, it seems like one could literally use `grep` or a regular expression applied across the source files to achieve the goals of the `-Z parse-only` flag, and in fact with better fidelity, since it would cover cases like `foo!({ extern crate bar; })`.

nikomatsakis:
Removed the `T-compiler` tag as we feel this is more of a tools decision. But this has been in FCP for a while!

nrc:
We discussed this at the tools meeting today. We are broadly sympathetic with the goal of interfacing with non-Cargo build systems, and more specifically to exporting dependencies in some form. However, there are worries about this RFC (specifically the bootstrapping part) and its motivation (how exactly, in the bigger picture, should Rust work with other build systems). There does not seem to be a huge desire for the feature right now. Therefore we agreed to close this RFC, however, we may be in favour of something similar at some point in the future, given sufficient demand and a better idea of the big picture.

PR #1624, opened 2016-05-20
loop-break-value (issue #961)

dhardy:
This is the culmination of recent discussion in issue #961, minus discussion about extensions to for/while/while let (for which there is no clear solution).

### DIFF
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -218,16 +218,50 @@ otherwise and allows more flexibility in code layout.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-It would be possible to allow `for`, `while` and `while let` expressions return
-values in a similar way; however, these expressions may also terminate
-"naturally" (not via break), and no consensus has been reached on how the
-result value should be determined in this case, or even the result type.
-It is thus proposed not to change these expressions at this time.
+### Extension to for, while, while let
 
-It should be noted that `for`, `while` and `while let` can all be emulated via
-`loop`, so perhaps allowing the former to return values is less important.
-Alternatively, a keyword such as `else default` could be used to
-specify the other exit value as in:
+A frequently discussed issue is extension of this concept to allow `for`,
+`while` and `while let` expressions to return values in a similar way. There is
+however a complication: these expressions may also terminate "naturally" (not
+via break), and no consensus has been reached on how the result value should
+be determined in this case, or even the result type.
+
+There are three options:
+
+1.  Do not adjust `for`, `while` or `while let` at this time
+2.  Adjust these control structures to return an `Option<T>`, returning `None`
+    in the default case
+3.  Specify the default return value via some extra syntax
+
+#### Via `Option<T>`
+
+Unfortunately, option (2) is not possible to implement cleanly without breaking
+a lot of existing code: many functions use one of these control structures in
+tail position, where the current "value" of the expression, `()`, is implicitly
+used:
+
+```rust
+// function returns `()`
+fn print_my_values(v: &Vec<i32>) {
+    for x in v {
+        println!("Value: {}", x);
+    }
+    // loop exits with `()` which is implicitly "returned" from the function
+}
+```
+
+Two variations of option (2) are possible:
+
+*   Only adjust the control structures where they contain a `break EXPR;` or
+    `break 'label EXPR;` statement. This may work but would necessitate that
+    `break;` and `break ();` mean different things.
+*   As a special case, make `break ();` return `()` instead of `Some(())`,
+    while for other values `break x;` returns `Some(x)`.
+
+#### Via extra syntax for the default value
+
+Several syntaxes have been proposed for how a control structure's default value
+is set. For example:
 
 ```rust
 fn first<T: Copy>(list: Iterator<T>) -> Option<T> {
@@ -239,8 +273,18 @@ fn first<T: Copy>(list: Iterator<T>) -> Option<T> {
 }
 ```
 
-The exact syntax is disputed; (JelteF has some suggestions which should work
-without infinite parser lookahead)
-[https://github.com/rust-lang/rfcs/issues/961#issuecomment-220728894].
-It is suggested that this RFC should not be blocked on this issue since
-loop-break-value can still be implemented in the manner above after this RFC.
+or:
+
+```rust
+let x = for thing in things default "nope" {
+    if thing.valid() { break "found it!"; }
+}
+```
+
+There are two things to bear in mind when considering new syntax:
+
+*   It is undesirable to add a new keyword to the list of Rust's keywords
+*   It is strongly desirable that unbounded lookahead is required while syntax
+    parsing Rust code
+
+For more discussion on this topic, see [issue #961](https://github.com/rust-lang/rfcs/issues/961).
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -6,6 +6,10 @@
 # Summary
 [summary]: #summary
 
+(This is a result of discussion of issue #961 and related to RFCs
+[352](https://github.com/rust-lang/rfcs/pull/352) and
+[955](https://github.com/rust-lang/rfcs/pull/955).)
+
 Let a `loop { ... }` expression return a value via `break my_value;`.
 
 # Motivation
@@ -159,3 +163,5 @@ It is thus proposed not to change these expressions at this time.
 
 It should be noted that `for`, `while` and `while let` can all be emulated via
 `loop`, so perhaps allowing the former to return values is less important.
+
+See discussion of #961 for more on this topic.
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -84,14 +84,17 @@ fn h() -> ! {
 }
 ```
 
-This proposal changes the type to `T`, where:
+This proposal changes the result type of 'loop' to `T`, where:
 
-*   a loop which is never "broken" via `break` has result-type `!` (which is coercible to anything, as of today)
-*   where a loop is "broken" via `break;` or `break 'label;`, its result type is `()`
-*   where a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
-*   a loop's return type may be deduced from its context, e.g. `let x: T = loop { ... };`
+*   if a loop is "broken" via `break;` or `break 'label;`, the loop's result type must be `()`
+*   if a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
+*   as a special case, if a loop is "broken" via `break EXPR;` or `break 'label EXPR;` where `EXPR` evaluates to type `!` (does not return), this does not place a constraint on the type of the loop
+*   if external constaint on the loop's result type exist (e.g. `let x: S = loop { ... };`), then `T` must be coercible to this type
 
-It is an error if these types do not agree. Examples:
+It is an error if these types do not agree or if the compiler's type deduction
+rules do not yield a concrete type.
+
+Examples of errors:
 
 ```rust
 // error: loop type must be () and must be i32
@@ -115,7 +118,7 @@ fn z() -> ! {
 }
 ```
 
-Where a loop does not break, the return type is coercible:
+Examples involving `!`:
 
 ```rust
 fn f() -> () {
@@ -126,28 +129,15 @@ fn g() -> u32 {
     // ! coerces to u32
     loop {}
 }
-```
-
-### Result value
-
-A loop only yields a value if broken via some form of `break ...;` statement,
-in which case it yields the value resulting from the evaulation of the
-statement's expression (`EXPR` above), or `()` if there is no `EXPR`
-expression.
-
-## Examples
-
-```rust
-assert_eq!(loop { break; }, ());
-assert_eq!(loop { break 5; }, 5);
-let x = 'a loop {
-    'b loop {
-        break 'a 1;
+fn z() -> ! {
+    loop {
+        break panic!();
     }
-    break 'a 2;
-};
-assert_eq!(x, 1);
+}
 ```
+
+Example showing the equivalence of `break;` and `break ();`:
+
 ```rust
 fn y() -> () {
     loop {
@@ -159,12 +149,26 @@ fn y() -> () {
     }
 }
 ```
+
+### Result value
+
+A loop only yields a value if broken via some form of `break ...;` statement,
+in which case it yields the value resulting from the evaulation of the
+statement's expression (`EXPR` above), or `()` if there is no `EXPR`
+expression.
+
+Examples:
+
 ```rust
-fn z() -> ! {
-    loop {
-        break panic!();
+assert_eq!(loop { break; }, ());
+assert_eq!(loop { break 5; }, 5);
+let x = 'a loop {
+    'b loop {
+        break 'a 1;
     }
-}
+    break 'a 2;
+};
+assert_eq!(x, 1);
 ```
 
 # Drawbacks
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -54,7 +54,7 @@ Four forms of `break` will be supported:
 3.  `break EXPR;`
 4.  `break 'label EXPR;`
 
-where `'label` is the name of a loop and `EXPR` is an expression.
+where `'label` is the name of a loop and `EXPR` is an converging expression.
 
 ### Result type of loop
 
@@ -88,7 +88,6 @@ This proposal changes the result type of 'loop' to `T`, where:
 
 *   if a loop is "broken" via `break;` or `break 'label;`, the loop's result type must be `()`
 *   if a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
-*   as a special case, if a loop is "broken" via `break EXPR;` or `break 'label EXPR;` where `EXPR` evaluates to type `!` (does not return), this does not place a constraint on the type of the loop
 *   if external constaint on the loop's result type exist (e.g. `let x: S = loop { ... };`), then `T` must be coercible to this type
 
 It is an error if these types do not agree or if the compiler's type deduction
@@ -128,11 +127,6 @@ fn f() -> () {
 fn g() -> u32 {
     // ! coerces to u32
     loop {}
-}
-fn z() -> ! {
-    loop {
-        break panic!();
-    }
 }
 ```
 
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -84,7 +84,7 @@ Four forms of `break` will be supported:
 3.  `break EXPR;`
 4.  `break 'label EXPR;`
 
-where `'label` is the name of a loop and `EXPR` is an converging expression.
+where `'label` is the name of a loop and `EXPR` is an expression.
 
 ### Result type of loop
 
@@ -118,6 +118,7 @@ This proposal changes the result type of 'loop' to `T`, where:
 
 *   if a loop is "broken" via `break;` or `break 'label;`, the loop's result type must be `()`
 *   if a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
+*   as a special case, if a loop is "broken" via `break EXPR;` or `break 'label EXPR;` where `EXPR` evaluates to type `!` (does not return), this does not place a constraint on the type of the loop
 *   if external constaint on the loop's result type exist (e.g. `let x: S = loop { ... };`), then `T` must be coercible to this type
 
 It is an error if these types do not agree or if the compiler's type deduction
@@ -157,6 +158,11 @@ fn f() -> () {
 fn g() -> u32 {
     // ! coerces to u32
     loop {}
+}
+fn z() -> ! {
+    loop {
+        break panic!();
+    }
 }
 ```
 
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -1,0 +1,161 @@
+- Feature Name: loop_break_value
+- Start Date: 2016-05-20
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Let a `loop { ... }` expression return a value via `break my_value;`.
+
+# Motivation
+[motivation]: #motivation
+
+This pattern is currently hard to implement without resorting to a function or
+closure wrapping the loop:
+
+```rust
+fn f() {
+    let outcome = loop {
+        // get and process some input, e.g. from the user or from a list of
+        // files
+        let result = get_result();
+        
+        if successful() {
+            break result;
+        }
+        // otherwise keep trying
+    };
+    
+    use_the_result(outcome);
+}
+```
+
+In some cases, one can simply move `use_the_result(outcome)` into the loop, but
+sometimes this is undesirable and sometimes impossible due to lifetimes.
+
+# Detailed design
+[design]: #detailed-design
+
+This proposal does two things: let `break` take a value, and let `loop` have a
+result type other than `()`.
+
+### break syntax
+
+Four forms of `break` will be supported:
+
+1.  `break;`
+2.  `break 'label;`
+3.  `break EXPR;`
+4.  `break 'label EXPR;`
+
+where `'label` is the name of a looping construct and `EXPR` is an evaluable
+expression.
+
+### result type
+
+Currently the result-type of a 'loop' without 'break' is `!` (never returns),
+which may be coerced to `()`. This is important since a loop may appear as
+the last expression of a function:
+
+```rust
+fn f() {
+    loop {
+        do_something();
+        // never breaks
+    }
+}
+fn g() -> () {
+    loop {
+        do_something();
+        if Q() { break; }
+    }
+}
+fn h() -> ! {
+    loop {
+        do_something();
+        // this loop is not allowed to break due to `!` result type
+    }
+}
+```
+
+This proposal changes the result type to `T`, where:
+
+*   a loop which is never "broken" via `break` has result-type `!` (coercible to `()`)
+*   a loop's return type may be deduced from its context, e.g. `let x: T = loop { ... };`
+*   where a loop is "broken" via `break;` or `break 'label;`, its result type is `()`
+*   where a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
+
+It is an error if these types do not agree. Examples:
+
+```rust
+// error: loop type must be ! or () not i32
+let a: i32 = loop {};
+// error: loop type must be i32 and must be &str
+let b: i32 = loop { break "I am not an integer."; };
+// error: loop type must be Option<_> and must be &str
+let c = loop {
+    if Q() {
+        "answer"
+    } else {
+        None
+    }
+};
+fn z() -> ! {
+    // function does not return
+    // error: loop may break (same behaviour as before)
+    loop {
+        if Q() { break; }
+    }
+}
+```
+
+### result value
+
+A loop only yields a value if broken via some form of `break ...;` statement,
+in which case it yields the value resulting from the evaulation of the
+statement's expression (`EXPR` above), or `()` if there is no `EXPR`
+expression.
+
+Examples:
+
+```rust
+assert_eq!(loop { break; }, ());
+assert_eq!(loop { break 5; }, 5);
+let x = 'a loop {
+    'b loop {
+        break 'a 1;
+    }
+    break 'a 2;
+};
+assert_eq!(x, 1);
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The proposal changes the syntax of `break` statements, requiring updates to
+parsers and possibly syntax highlighters.
+
+The type of `loop` expressions is no longer fixed and cannot be explicitly
+typed.
+
+# Alternatives
+[alternatives]: #alternatives
+
+No alternatives to the design have been suggested. It has been suggested that
+the feature itself is unnecessary, and indeed much Rust code already exists
+without it, however the pattern solves some cases which are difficult to handle
+otherwise and allows more flexibility in code layout.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+It would be possible to allow `for`, `while` and `while let` expressions return
+values in a similar way; however, these expressions may also terminate
+"naturally" (not via break), and no consensus has been reached on how the
+result value should be determined in this case, or even the result type.
+It is thus proposed not to change these expressions at this time.
+
+It should be noted that `for`, `while` and `while let` can all be emulated via
+`loop`, so perhaps allowing the former to return values is less important.
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -16,28 +16,58 @@ Let a `loop { ... }` expression return a value via `break my_value;`.
 # Motivation
 [motivation]: #motivation
 
-This pattern is currently hard to implement without resorting to a function or
-closure wrapping the loop:
+> Rust is an expression-oriented language. Currently loop constructs don't
+> provide any useful value as expressions, they are run only for their
+> side-effects. But there clearly is a "natural-looking", practical case,
+> described in [this thread](https://github.com/rust-lang/rfcs/issues/961)
+> and [this] RFC, where the loop expressions could have
+> meaningful values. I feel that not allowing that case runs against the
+> expression-oriented conciseness of Rust.
+> [comment by golddranks](https://github.com/rust-lang/rfcs/issues/961#issuecomment-220820787)
+
+Some examples which can be much more concisely written with this RFC:
 
 ```rust
-fn f() {
-    let outcome = loop {
-        // get and process some input, e.g. from the user or from a list of
-        // files
-        let result = get_result();
-        
-        if successful() {
-            break result;
+// without break-with-value:
+let x = {
+    let temp_bar;
+    loop {
+        ...
+        if ... {
+            temp_bar = bar;
+            break;
         }
-        // otherwise keep trying
-    };
-    
-    use_the_result(outcome);
-}
-```
+    }
+    foo(temp_bar)
+};
 
-In some cases, one can simply move `use_the_result(outcome)` into the loop, but
-sometimes this is undesirable and sometimes impossible due to lifetimes.
+// with break-with-value:
+let x = foo(loop {
+        ...
+        if ... { break bar; }
+    });
+
+// without break-with-value:
+let computation = {
+    let result;
+    loop {
+        if let Some(r) = self.do_something() {
+            result = r;
+            break;
+        }
+    }
+    result.do_computation()
+};
+self.use(computation);
+
+// with break-with-value:
+let computation = loop {
+        if let Some(r) = self.do_something() {
+            break r;
+        }
+    }.do_computation();
+self.use(computation);
+```
 
 # Detailed design
 [design]: #detailed-design
--- a/text/0000-loop-break-value.md
+++ b/text/0000-loop-break-value.md
@@ -59,8 +59,9 @@ expression.
 
 ### result type
 
-Currently the result-type of a 'loop' without 'break' is `!` (never returns),
-which may be coerced to `()`. This is important since a loop may appear as
+Currently the result-type of a 'loop' without 'break' is `!` (never returns,
+and may be coerced to any type), and the result type of a 'loop' with 'break'
+is `()`. This is important since a loop may appear as
 the last expression of a function:
 
 ```rust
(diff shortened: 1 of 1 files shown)


llogiq:
[Rendered](https://github.com/dhardy/rfcs/blob/master/text/0000-loop-break-value.md)

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,216 @@
...
+[352](https://github.com/rust-lang/rfcs/pull/352) and
+[955](https://github.com/rust-lang/rfcs/pull/955).)
+
+Let a `loop { ... }` expression return a value via `break my_value;`.
+
+# Motivation
+[motivation]: #motivation
+
+This pattern is currently hard to implement without resorting to a function or
+closure wrapping the loop:
```
sfackler: It's not *that* hard:

```rust
fn f() {
    let outcome;
    loop {
        // get and process some input, e.g. from the user or from a list of
        // files
        let result = get_result();
        
        if successful() {
            outcome = result;
            break;
        }
        // otherwise keep trying
    };
    
    use_the_result(outcome);
}
```

dhardy: Yes, you're right. I was able to reduce my problem to this pattern:

```rust
let computation = {
    let result = None;
    loop {
        if let Some(r) = self.do_something() {
            result = Some(s);
            break;
        }
    }
    result.unwrap().do_computation()
};
self.use(computation);
```
Lifetimes make some of these problems horrible to think about.

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,216 @@
...
+        // this loop is not allowed to break due to inferred `!` type
+    }
+}
+```
+
+This proposal changes the result type of 'loop' to `T`, where:
+
+*   if a loop is "broken" via `break;` or `break 'label;`, the loop's result type must be `()`
+*   if a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
+*   as a special case, if a loop is "broken" via `break EXPR;` or `break 'label EXPR;` where `EXPR` evaluates to type `!` (does not return), this does not place a constraint on the type of the loop
```
birkenfeld: I don't see the need for this special case: there's no reason to write `break EXPR;` instead of `EXPR;` when the expr is diverging.

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,216 @@
...
+2.  `break 'label;`
+3.  `break EXPR;`
+4.  `break 'label EXPR;`
+
+where `'label` is the name of a loop and `EXPR` is an expression.
+
+### Result type of loop
+
+Currently the result-type of a 'loop' without 'break' is `!` (never returns),
+which may be coerced to any type), and the result type of a 'loop' with 'break'
```
birkenfeld: consistency nit: "result-type" vs "result type"

text/0000-loop-break-value.md
dhardy: No, and I suppose this is why it's not legal to type `let x: ! = ...;`.

birkenfeld: Well, that's because `!` is not a real type (yet - there is an RFC open for that) and is only allowed in function return type notation.

durka:
Btw, I have a [scary macro](https://crates.io/crates/named-block) that mostly [implements this](https://github.com/durka/named-block/blob/25511679787601e28f8782b22f50963ab80df6e0/src/lib.rs#L392-L403) as well as labeled blocks, in case you want to play around.

text/0000-loop-break-value.md
dhardy: Damn. The problem with being British is you expect people to spot the sarcasm.

1.  Having to mutable a variable just to return a value is inelegant
2.  Having to wrap with `Option` then unwrap just because there isn't an appropriate default value is clumsy
3.  Having to do this while considering the lifetimes of generated references is confusing
4.  It's a lot of extra code. Compare the above to:

```rust
let computation = loop {
        if let Some(r) = self.do_something() {
            break r;
        }
    }.do_computation();
self.use(computation);
```
(noting that due to reference lifetimes it may be possible to give the value fed to `do_computation()` a name in the outer scope).

Is that sufficient motivation?

The new lifetime rules under discussion may well simplify this problem, both with-and-without break-with-value, but I still think break-with-value is a neat pattern (or alternately, a generic break-from-block-with-value, but that opens another can of worms: the value if not broken out of).

birkenfeld: Well, @sfackler 's example works fine without a mutable variable or an `Option`. Am I missing something?

dhardy: Maybe, and maybe it's a good thing. Sounds like it's time I un-learned to *never leave variables uninitialised*.

Points 3 and 4 still hold (the example above can be four words shorter), but this weakens the argument.

text/0000-loop-break-value.md
dhardy: Can you delete this comment please? It adds nothing to the argument (not even clarity).

text/0000-loop-break-value.md
birkenfeld: Point 3 - please give an example where these confusing-to-consider lifetimes are involved.
Point 4 - "a lot of extra code" when talking about essentially one extra assignment?

JelteF: The current pattern is not necessarily hard, but the main problem I think is that the new pattern expresses much better what is going on.
It is immediately clear that the loop is meant to set the value, while with the code of @sfackler this is not te case. Especially when considering loops with more content or with multiple exit points.

notriddle:
Considering the fact that `println!("{:?}", panic!())` type-checks, do we really want to require break's expression to converge? AFAIK, diverging expressions are literally allowed everywhere else.

upsuper:
`for` and `while` may need something like Python's for-else structure. That could be another RFC I suppose.

dhardy:
@notriddle good point; I have no idea.

@upsuper: already discussed a *lot*; explicitly not part of the RFC

golddranks:
I think that the requirement for `EXPR` in `break EXPR;` to converge, as stated in the RFC text, is unnecessary. Like @notriddle said, the expressions in Rust work even without such requirements, so changing that would be a complication.

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,240 @@
...
+in which case it yields the value resulting from the evaulation of the
+statement's expression (`EXPR` above), or `()` if there is no `EXPR`
+expression.
+
+Examples:
+
+```rust
+assert_eq!(loop { break; }, ());
+assert_eq!(loop { break 5; }, 5);
+let x = 'a loop {
```
solson: syntactic nit: this should be `'a: loop {` and `'b: loop {` on the line below.

chris-morgan:
`break value` should be behaving *exactly* like `return value`, except producing a value for a loop rather than a function. `return break break return break up (please, continue)` should therefore compile (presuming a suitable function `up` and variable `please`), just as `return return return return return` works fine now.

I look forward to this compiling:

```rust
fn main() {
}

struct X;
type B = Result<(X), (X)>;  // That signature looks scary.
const please: () = ();
macro_rules! escape { ($x:expr, $y:expr) => (()); }
fn out(_: (), _: ()) { }
const now: B = Ok(X);

fn attempt_escape() -> B {
    'prison: loop {
        if break out (please, continue) {
            escape! (oh, yeah?)
        } else {
            return now :(
            // You got caught. Maybe next time you should try wearing cool sunnies like this:
            B)
        }
    }
}
```

wycats:
I like this improvement a lot.

Ruby is also an expression-oriented language and allows `break EXPR`. Even though I don't use the feature all that often in Ruby, it's pretty easy to remember (it has a good analogue with `return`) and makes code terser without a loss in clarity when it's useful.

I especially like the fact that this is a nice improvement to an existing expression form with minimal and intuitive new syntax.

Like the original issue thread, I have some concerns with other extensions, and am glad this RFC split out just `break EXPR` so it could be evaluated on its own.

nikomatsakis:
I'm basically 👍 on this. Limiting to `loop` makes sense to me, at least initially.

dhardy:
Nice. I'm still watching this but from what I gather no changes are needed to the RFC?

dhardy:
Wrong button!

sanxiyn:
Note that `break (0, 0);` currently parses as a call. (Result of `break` expression is called as a function with two arguments.) This RFC will change the parse. Not a big problem, since it can be handled exactly like `return (0, 0);`.

notriddle:
The result of break expression isn't callable, is it? So even if it parses differently, it won't compile.

solson:
@notriddle Correct. It'll be a "type must be known in this context" problem.

nrc:

Summary: this RFC proposes letting loops take values which are produced by `break`. That allows loops to be treated as expressions rather than statements. This is a fairly minor change with a small but significant benefit. Response has been positive, but not overwhelming. There are no major sticking points worth highlighting. There is a minor point of uncertainty around divergence, which I expect is affected by the recent acceptance of the `!`-is-a-type RFC.

I propose moving to FCP with an inclination to accept.

- [ ] @nikomatsakis
- [x] @nrc 
- [ ] @aturon 
- [ ] @eddyb 
- [ ] @pnkfelix (on vacation)

pnkfelix:
@nrc the point of uncertainty around divergence was regarding a prior draft of the RFC, wasn't it? That is, I thought that before d6e1594 people were concerned about a requirement that all expressions given to `break` converge, but now that is no longer a requirement -- so, is there still uncertainty about `!` ?

(In any case, I'm still going to check my name off, since I'm cool with the RFC a currently written.)

aturon:
🔔 **This RFC is entering its final comment period**. 🔔

diwic:
Any thoughts on inference? Is there a way to explicitly specify the type that `loop` returns in case the type is difficult to infer; or will we have to do something like

    { let x: MyType = loop { /* ... */ }; x }

notriddle:
That's type ascription. It's not supported yet.

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,290 @@
...
+    let result;
+    loop {
+        if let Some(r) = self.do_something() {
+            result = r;
+            break;
+        }
+    }
+    result.do_computation()
+};
+self.use(computation);
```
Thiez: Why not the following?

```rust
let computation;
loop {
    if let Some(r) = self.do_something() {
        computation = r.do_computation();
    }
}
self.use(computation);
```

Or

```rust
let computation;
loop {
    if let Some(r) = self.do_something() {
        computation = r;
        break;
    }
}
let computation = computation.do_computation();
self.use(computation);
```

JelteF:
@Thiez I agree that these examples seem to needlessly complex. However, when you imagine loops with multiple break points these types of constructs would be necessary.

dhardy:
@Thiez you're right, those examples can be simplified. I wasn't very familiar with the `let x; loop { ... { x = y; break; } }` pattern when I wrote the RFC. Contrary to what @JelteF says, the pattern even appears to work with multiple break points (though each "break" requires two lines of code).

```
let computation;
loop {
    if let Some(r) = self.do_something() {
        computation = r;
        break;
    }
}
let computation = computation.do_computation();
self.use(computation);
```
This example however won't always work with the current lifetime restrictions, since `r` may contain a reference to `self` (hence doing `r.do_computation()` early; this was one of the examples motivating this RFC).

dhardy:
@Thiez actually, on further thought, there may be examples where it's not possible to move function calls like `foo(bar)` or `r.do_computation()` into the loop due to further lifetime restrictions, though I'm not sure how realistic such cases are and in any case non-lexical lifetimes should solve these issues.

So I'm not sure how much motivation for this RFC remains. There is still some syntactic improvement, especially in cases with multiple returns, but this RFC is mostly sugar:
```rust
    let x;
    loop {
        ...
        if ... {
            x = v1;
            break;
        } else if ... {
            x = v2;
            break;
        }
    }
    now_do_something(x);
```
becomes
```rust
    let x = loop {
        ...
        if ... {
            break v1;
        } else if ... {
            break v2;
        }
    };
    now_do_something(x);
```
(where `now_do_something(x)` remains at the end for sequential clarity).

notriddle:
It is syntactic sugar (I always understood it as such, and never really thought about it as a workaround for lexical lifetimes), but it jives so well with the rest of the language that it seems like it should have been there from the beginning.

p.s. @dhardy's second example isn't just shorter, it's also clearer, since there's no way to know that the first example actually fills `x` in inside the loop without reading the loop body.

nikomatsakis:
> It is syntactic sugar (I always understood it as such, and never really thought about it as a workaround for lexical lifetimes), but it jives so well with the rest of the language that it seems like it should have been there from the beginning.

This is basically how I feel. I've always thought of it as syntactic sugar. Moreover, Rust's currently a very "expression-friendly" language, but a lot of the more imperative constructs are perhaps not as well-integrated as they could be. This RFC seems to help mend that gap a bit in a very natural way.

Thiez:
> p.s. @dhardy's second example isn't just shorter, it's also clearer, since there's no way to know that the first example actually fills `x` in inside the loop without reading the loop body.

There is actually a very easy way to know this: because `x` is not assigned before the loop, and it is used after the loop, then it must have been assigned inside the loop, or the compiler would reject the code for using a possibly uninitialized variable.

notriddle:
Half the time, I'm reading code to figure out why it's not compiling. If I added another break into a loop, which would you rather see:

    Variable may not have been initialized:
        now_do_something(x)
        ~~~~~~~~~~~~~~~~~^

Or this:

    Expected break of type `T`, found break of type `()`:
            break;
        ~~~~^

dhardy:
@notriddle good point about error messages, but I suspect that the compiler may not always get the type right. Error messages may be more like:
```
Expected break types to agree; found break of type `str`:
    break "text";
          ~~~~~^
found break of type `()`:
    break;
         ^
```
It depends... sometimes the output type may be clear (if typed or passed directly into/out of a function), and sometimes it may not be (e.g. if cast with `.into()`).

notriddle:
Your hypothetical error message is still better than what unassigned variables give us. It points us at the break statement, which is where the problem is, and it points at another break statement, giving an example of how it should have been written. If the loop is being used in an expression context, a no-brainer heuristic would hint that the unit break is probably the wrong one.

The point is that `let x = for { break f(); }` is explicit, while `let x; for { x = f(); break; }` is implicit. The error messages are just this extra, explicit information being exploited by the compiler. Even for human readers, figuring out that a variable must have been assigned inside a loop because it's used right afterward requires the person reading the code to do deductive reasoning, whereas figuring out that a loop is being assigned to a variable directly merely requires someone to read it.

vitiral:
I don't think this should be accepted without having the `break` statement for `while` and `for` as well. Having it for only `loop` would diverge the construct of all 3 even more.

Mostly, I don't understand why this would be invalid:

```
while n < value {
   if n == 3 {
       break "3";
    }
} else {
       "default"
}
```

The RFC states that you need `else default` but it doesn't make sense as to why you would need the extra keyword.

JelteF:
Lots of people were saying else was not clear wording, read
https://github.com/rust-lang/rfcs/issues/961 for details. Loop would not
need an else/default, so that's why this is not opposed.

vitiral:
@JelteF thanks, left an idea to use `!break` instead of `else`

glaebhoerl:
@vitiral For practical purposes, given the (very) long-standing impasse on the syntax to use with `while` and `for`, your position amounts to "I don't think this should be accepted". If that's really what you think, then OK.

vitiral:
@glaebhoerl unless we could resolve those conflicts in that thread -- I think loop break values is a useful feature to have, I just don't want the different loop constructs to start diverging.

dhardy:
@vitiral @glaebhoerl [is this a suitable solution](https://github.com/rust-lang/rfcs/issues/961#issuecomment-250200311)?

Separate question: is it sufficiently related to integrate into this RFC, potentially delaying the `loop` case? (I propose: yes, unless there is still significant disagreement over syntax.)

Ericson2314:
No it's not worth delaying any more. It's been 2 years (?) since this was first proposed.

vitiral:
I think I am for this RFC provided that it has a flag to not be stabilized until the `for` and `while` implementations are also finished. Putting it in nightly to get data on the usefulnes is good I think, but I think it would be a mistake to release this without all the loop constructs having this functionality (there shouldn't be special cases).

glaebhoerl:
@dhardy See my [earlier comment](https://github.com/rust-lang/rfcs/issues/961#issuecomment-223347468). I think any of those are fine by me if they work technically and are acceptable to others.

notriddle:
> I just don't want the different loop constructs to start diverging.

They're already different. `loop {}` is the only construct that can be recognized as diverging (it evaluates to `!` when there's no break statement). It can also evaluate to `()` when there is a break statement in it. `while` and `for` always evaluate to `()` no matter what.

Thiez:
> Putting it in nightly to get data on the usefulnes is good I think, but I think it would be a mistake to release this without all the loop constructs having this functionality (there shouldn't be special cases).

`loop` is inherently a special case; it is the one loop structure where the compiler understands that the loop body will execute at least once, so you can do this:

```rust
let x;
loop {
    x = 5;
    break;
}
println!("{}", x);
```

but not this:

```rust
let x;
while true {
    x = 5;
    break;
}
println!("{}", x);
```

I think Java solves this by special-casing `while (true)` in the compiler.

nikomatsakis:
So I was just about to approve this RFC, but @aturon mentioned he has some concerns. Therefore, I want to try and move this onto the RFC-bot process and start the FCP over. =)

@rfcbot fcp merge

nikomatsakis:
@rfcbot I ❤️ you.

aturon:
@rfcbot concern Cost/benefit and tail calls

As I've chewed on this RFC over time, I've gotten increasingly uneasy. While in general I'm open to syntactic sugar and conveniences, they do come at some complexity cost for the language, and thus should be justified by fairly widespread benefit (and should fit into the language as smoothly as possible).

My specific concerns here are:

- There's a very reasonable encoding in terms of delayed `let` binding, as others have pointed out. While this encoding is *slightly* harder to follow, the `let x;` form gives a very clear heads up. I'm just not 100% sure this bit of sugar winds up helping that much.

- While I agree with @nikomatsakis that the RFC moves us further in the direction of expression-orientation, it feels like an "incomplete" move, since the change only applies to `loop` and not our other looping constructs. This is also a source of complexity with the sugar, making the feature somewhat more obscure and introducing edges for its use.

- This sort of loop is very naturally encoded via tail recursion; if we ever get tail call optimization, I would *probably* prefer to right such loops via tail recursion -- making this feature perhaps even more niche?

Overall, I'm sympathetic, but I worry that we have a limited total budget for syntactic sugar, and there are a lot of purchases [we might want to make](https://internals.rust-lang.org/t/roadmap-2017-productivity-learning-curve-and-expressiveness/4097/27).

burdges:
Yes, tail recursion is normally the nicer way to express this sort of thing.

notriddle:
> * There's a very reasonable encoding in terms of delayed let binding, as others have pointed out. While this encoding is slightly harder to follow, the let x; form gives a very clear heads up. I'm just not 100% sure this bit of sugar winds up helping that much.

This is also true for many other cases, including `if`, `match`, and unadorned block. The big advantages are in the composability (I can put a loop in the terminal position of an if branch and have the break value just propagate out) and the explicitness (it's not a guess; I know the value is coming from the loop).

> * While I agree with @nikomatsakis that the RFC moves us further in the direction of expression-orientation, it feels like an "incomplete" move, since the change only applies to loop and not our other looping constructs. This is also a source of complexity with the sugar, making the feature somewhat more obscure and introducing edges for its use.

If you think you can resolve the default value :bike: :house: then be my guest.

> * This sort of loop is very naturally encoded via tail recursion; if we ever get tail call optimization, I would probably prefer to right such loops via tail recursion -- making this feature perhaps even more niche?

I actually agree with that. If and when we get TCO then I'll consider this use case solved and just relegate loops to the niche situations.

ubsan:
@aturon The issue I have with that is "*if* we ever get tail call optimization". It seems unlikely to happen in the near (or medium term) future... And, also, imho, `let x = loop {...}` is much more readable.

notriddle:
Pick between this:

    fn factorial(n: u64) {
        become factorial_(n, 1)
    }
    fn factorial_(n: u64, acc: u64) {
        if n == 0 { acc }
        else { become(n - 1, acc * n) }
    }

And this:

    fn factorial(mut n: u64) {
        let mut acc = 1;
        loop {
            if n == 0 { break acc }
            acc *= n;
            n -= 1;
        }
    }

p.s. I know that the second example works with return instead of break.

nikomatsakis:
@aturon 

> This sort of loop is very naturally encoded via tail recursion; if we ever get tail call optimization, I would probably prefer to right such loops via tail recursion -- making this feature perhaps even more niche?

I don't really consider tail calls an alternative. For one thing, while of course it's technically true that any loop can be rewritten using a fn, it's not a trivial rewrite in many cases, and which construct "feels right" depends (I think) on the example. In my opinion the *actual* competitor to this feature is iterators: most of the time that I want to carry a value out of a loop, it's something that I would prefer to write using an iterator chain with an explicit call to `next` (to get the first item produced). And I think people should do that, it's more explicit. :)

For another, tail calls are actually **significantly** more complex to add to the language. You have to figure out the interactions with destructors and the C ABI, neither of which play particularly nice with tail calls (destructors because they frequently add "extra" computation after the tail-call, the C ABI because it has the caller allocate space for arguments, so the tail call must use the same amount of stack space). We probably need a new keyword (`be`).

Still, I don't see this feature as a *must have* in any respect. I like it because I feel like Rust has gone about 90% of the way in terms of unifying exprs/statements and this seems to push us to 92% (arbitrary statistics, of course, not measuring anything real). I like this unification so going a bit further doesn't bother me. To me it makes the language feel just a bit more rounded.

> Overall, I'm sympathetic, but I worry that we have a limited total budget for syntactic sugar, and there are a lot of purchases we might want to make.

Yes, so: I don't see this as a major ergonomics hurdle or something that will boost productivity in particular. Certainly if I have to pick, I would **not** take this RFC over the things in that thread.

Really the question then is: do we have to pick? I wasn't really viewing this as "syntactic sugar", at least not any more than `if cond { v } else { w }` is. Basically, when you learn Rust you learn that many things you are accustomed to seeing as statements can be used as expressions. Currently, that set does not include loops. If we take this RFC, that set will include `loop` (but not `while` or `for`).

Diggsey:
@aturon 
> it feels like an "incomplete" move

So if this RFC had a provision that it would not actually be implemented until a design had been accepted and implemented for the remaining looping constructs (with the benefit that those RFCs would not need to re-hash the motivation for doing this) you would be in favour? Or implemented, but behind a feature flag and not stabilised until the other loops were implemented too? It seems an unnecessarily high bar to require a complete design to be accepted in one go.

burdges:
Just noticed two reasons  `let = loop { .. }` might be more competitive with delayed `let` bindings than I first though : First, a delayed `let` binding need not be `mut`, which rocks, but actually seems kinda subtle.  Second, `loop` does not exist in most imperative languages anyways, limiting confusion.

Thiez:
I would be interested in an analysis of how many times people currently use the `let some_var; loop { ... some_var = some_value; break; }` pattern. How many instances of 'loop { ... }` would even benefit from this sugar?

Ericson2314:
Yeah right now Rust (and MIR), for better or worse, makes a big distinction between inter- and intra-procedural control flow. On one hand, from a functional background this seems ludicrously arbitrary, on the other hand it's great for compilation (not having to recover first-order control flow, explicitness about what is a machine level "function").

Guaranteed tails calls (this is, strictly speaking *not* an optimization because the programmer is requiring constant space) would be a big step towards unifying inter- and intra-procedural control flow, and for those reasons above may or may not be good. The alternative is "expression-orient goto", which is basically the same thing but *only* for intra-procedural control flow and thus obviates concerns about calling convention, destructors, unwinding, etc.

I plan (when, no idea), write an RFC for the latter, "Goto Considered Harmless". At that point, it should be clear that all our looping constructs are skeuomorphisms to lure stodgy C programmers, so there's no aesthetic loss hot-rodding those constructs with more sugar.

eddyb:
For implementing `catch` a useful pattern might be `'name: { ... break 'name value; ... default }`.

nrc:
@aturon:

>    There's a very reasonable encoding in terms of delayed let binding

This same argument could be applied to our other structural expressions too - why allow `let x = if ... ` or `match` when you can do the `let x;` thing? It seems this RFC is a move towards uniformity in that respect.

>    it feels like an "incomplete" move, since the change only applies to loop and not our other looping constructs.

I somewhat agree with this one. I was much happier about this RFC when it was applied to other kinds of loops too. I had hoped we could use the value of the last iteration as the value of the whole loop, but I can see that might be confusing (although, it matches the underlying semantics of break, I believe. Perhaps also back compat concerns?).

Anyway, the way I see this is that there is a set of expressions which are void and a set which take the type of a sub-expression. This is simply moving one expression from the first set to the second. Although one might assume all 'looping' constructs work the same, that is not true, for example of all 'branching' constructs (e.g., `if` vs `if ... else`).

>    This sort of loop is very naturally encoded via tail recursion; if we ever get tail call optimization, I would probably prefer to right such loops via tail recursion -- making this feature perhaps even more niche?

I'm not sure whether this will be true or not. Even with tail recursion, using `let x = loop ...` feels more Rusty than a recursive function (which feels more like an FP solution). I also share the concerns voiced by others that tail recursion is not even on the horizon at the moment, so seems like a poor reason to delay.

nrc:
> Overall, I'm sympathetic, but I worry that we have a limited total budget for syntactic sugar, and there are a lot of purchases we might want to make.

I don't think that this is true. Whilst I think it makes sense sometimes to think of 'complexity' (however that is defined) in terms of a budget, I think syntactic sugar is much less fungible. It has costs of course, but these are usually very local - we worry about obfuscating semantics etc. in one area,rather than in the language as a whole (obvs, this is a simplification and there is some global interaction, but it is a smaller factor, I think). I think that none of the other sugar/ergonomics proposals touch on control flow in a way that is very similar to this proposal, and so I don't think this proposal would cause us to pass on those others.

glaebhoerl:
> I had hoped we could use the value of the last iteration as the value of the whole loop, but I can see that might be confusing

The issue here is that if the loop runs 0 times, there is no last iteration. A proposed solution was to make the type of such loops be `Option`, but this breaks loops which are currently in tail position, where their types need to be `()`. And then you're wading into the murk of either fudging it by making the type of the loop depend on whether it contains a `break`-with-value or a similar "solution", or instead having an additional `else` clause to provide the default value, but of course not called `else` because that is highly confusing to approximately 50% of people, but one of the other imperfect possibilities, and so on and so on in https://github.com/rust-lang/rfcs/issues/961.

arielb1:
This is syntactic sugar of the least-harmful kind. While this feature has low value, I don't see any strong reason *not* to include it.

pnkfelix:
@nrc: 
> I had hoped we could use the value of the last iteration as the value of the whole loop, but I can see that might be confusing (although, it matches the underlying semantics of break, I believe. Perhaps also back compat concerns?).

I think the biggest problem with that supposed semantics is that it is ill-defined for the case when a loop has zero iterations.

(Maybe we could find some way to distinguish `for` loops that guarnatee > 0 iterations by adding some new `trait AtLeastOnce: Iterator`, so then you could apply that semantics to both `for` and `loop`, but you would still be left with `while` and `while let` ...)

Amanieu:
IMO the only reasonable way of supporting returning a value from a `for` or `while` loop is with the `for ... else ...` syntax where the `else` clause is only executed if the loop terminated normally (as opposed to a `break`). Then all you need to do is ensure that the type of the `break`s is the same as that of the `else` clause.

This therefore leaves us with only two options: either only support returning a value for `loop`, or add `else` to `for` and `while`.

Ericson2314:
It uses existing keywords, it makes sense, and it has precedent in other languages. It's also wildly controversial at the moment but that hopefully will change once this lands and becomes popular.

dhardy:
I'm not convinced tail-call recursion has much to offer the argument for or against this RFC: Rust is a hybrid language allowing users to program with both functional and imperative features, however they are comfortable. Additionally, sometimes it is not convenient to use a function due to many variable/constant bindings from the current scope being in use.

(@Ericson2314's comment regarding Goto I don't fully understand but sounds intriguing.)

@pnkfelix I really do not like the sound of an `AtLeastOnce` trait (extra complexity in syntax & in iterators, and you still have to handle the *usual* case where iterators are not guaranteed to be non-empty).

@Amanieu please see [issue 961](https://github.com/rust-lang/rfcs/issues/961), there is considerable opposition to simply using the `else` keyword.

I left `for` and `while` out of the RFC since there is no clear consensus, but if it helps with closure on this RFC, my current favourite is [Gabor's suggestion](https://github.com/rust-lang/rfcs/issues/961#issuecomment-223347468):
```
for elem in elems {
    if pred(elem) {
        break elem
    }
} else {
    break default_value()
}
```

vitiral:
@Ericson2314 
> It uses existing keywords, it makes sense, and it has precedent in other languages. It's also wildly controversial at the moment but that hopefully will change once this lands and becomes popular.

I would also like to point out that `for ... else` is significantly better supported in a language like rust, where it will be used (mostly) to return values and will be much more clear. In python, the use case is much more rare, but in rust -- where the expression can be used to return a value -- it will be much more in sync with how `if ... else` works

dhardy:
@vitiral you can't *tell people* how a feature must be used — for sure it would sometimes get used that way. But having a `break` keyword there does help. I think the above example should have semi-colons terminating the `break ...` lines:

```
for name in suggest_names() {
    if like(name) {
        break name;
    }
} else {
    let last_resort = random_name();
    break last_resort;
}
```
The `else` does look a little weird here, though the `break` at the end is a clue as to the purpose. I can't personally tell what proportion of users would find this syntax confusing.

vitiral:
@dhardy I agree that break should have a semi-colon at the end (just like `return`). However, for clarity I am saying that it would be *typically* used like this:

```
let x = for name in suggest_names() {
    if like(name) {
        break name;
    }
} else {
    let last_resort = random_name();
    last_resort // no `break` necessary for else, just like `for ... else`
}
```
My argument that this is the way it would typically be used is because this is how patterns like this are typically used in rust. The `for ... else` in python is hardly used and is somewhat confusing precisely because it doesn't return any values.

Some have issues with `for ... else` because they think that it should work like so:

```
while x < 10 {
    // do stuff normally, break not required
} else {
    // do stuff here if the while loop was never entered
    // (i.e. if x was already < 10)
}
```

In other words, it could be replaced with:
```
if x < 10 {
    while x < 10 {
        // do stuff normally
    }
} else {
     // do stuff here if the while loop was never entered
}
```

However, this alternate implentation **cannot make sense if break must return a value**. If break returns a value, then clearly the `else` clause in `while ... else` must have something to do with break and nothing to do with whether or not the while loop was never entered.

There is much more ambiguity around what `while ... else` *could* do in python. In rust, that ambiguity wouldn't exist.

dhardy:
@vitiral interesting point. But the RFC states that `break ();` is equivalent to `break;`, and as I understand it the return value of a statement ending `;` is `()`, so if there is no `break` in the `else` block, the Python-esque `for ... else ...` with no `break` or return value would be completely legal (I think).

Besides, sometimes side-effects are much more important than return values. TBH I don't understand the Python ambiguity problem very well since it always seemed clear to me (yes, in Python).

vitiral:
the return value of any statement ending in `;` is `()`, so you could write:
```
for name in suggest_names() {
    if like(name) {
        print!("{}", name);
        break;
    }
} else {
    let last_resort = random_name();
    print!("{}", last_resort);
}
```
So you don't need a `break` in the `else` clause (just like you don't need one in `if ... else` even though that can return a value)

`for ... else` with no break value would **not** be legal in rust, since the `for` block would not have a return value. This is good, as `for ... else` without `break` makes no sense at all -- `else` would always get executed (assuming you didn't return or panic).

Amanieu:
Would it be better if `finally` (obviously this is open to bikeshedding) was used instead of `break`? This makes it clearer that this block is only executed if no `break` occurred in the loop.

```rust
let x = for name in suggest_names() {
    if like(name) {
        break name;
    }
} finally {
    random_name()
};
```

vitiral:
@Amanieu do you mean use `finally` instead of `else`? No, I think that would be even worse. For one thing, in python `finally` means "execute me no matter what".

Diggsey:
I'd like to know from the people who object to just using `for {} else {}` etc on the basis that it's confusing, whether using another keyword instead of `else` would solve the problem for them, or whether they'd still find it confusing?

If not, there's no point bike-shedding further on alternative names. The choices involving `Option` or some variation thereof seem to be flawed, so it's a simple choice between `else` or nothing.

JelteF:
I'm still very pro my suggestion of using `!break`. I think it really
simply shows the meaning of the second block. Without even reading the loop
body (and seeing that values are being returned with break).

glaebhoerl:
Something like the last 10 comments (I didn't count) are reinventing or rehashing ideas, and then the objections to them, almost verbatim from [the other thread](https://github.com/rust-lang/rfcs/issues/961). Nicely illustrating, if anyone needed it, why that feature was deliberately left out of this RFC.

aturon:
@rfcbot resolved Cost/benefit and tail calls

I've been persuaded that, on the whole, this RFC moves the language toward greater uniformity, and that in particular proper tall calls are not necessarily a better choice. Let's do it!

aturon:
🔔 **This RFC is entering its final comment period**. 🔔

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,290 @@
...
+1.  `break;`
+2.  `break 'label;`
+3.  `break EXPR;`
+4.  `break 'label EXPR;`
+
+where `'label` is the name of a loop and `EXPR` is an expression.
+
+### Result type of loop
+
+Currently the result type of a 'loop' without 'break' is `!` (never returns),
```
nagisa: s/'/`/

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,290 @@
...
+which may be coerced to any type), and the result type of a 'loop' with 'break'
```
nagisa: s/'/`/

I will not nit on any more of the formatting, but the issues with it probably should be fixed across the document. Namely, plain apostrophe (') does not begin/end an inline code snippet.

text/0000-loop-break-value.md
nagisa: Now `!` is a legal type everywhere, but this special case is still useful as a note, even though it follows just fine from rules applicable to `!` (i.e. `!` coerces to any other type).

text/0000-loop-break-value.md
```diff
@@ -0,0 +1,290 @@
...
+    }
+}
+```
+
+This proposal changes the result type of 'loop' to `T`, where:
+
+*   if a loop is "broken" via `break;` or `break 'label;`, the loop's result type must be `()`
+*   if a loop is "broken" via `break EXPR;` or `break 'label EXPR;`, `EXPR` must evaluate to type `T`
+*   as a special case, if a loop is "broken" via `break EXPR;` or `break 'label EXPR;` where `EXPR` evaluates to type `!` (does not return), this does not place a constraint on the type of the loop
+*   if external constaint on the loop's result type exist (e.g. `let x: S = loop { ... };`), then `T` must be coercible to this type
```
nagisa: This list still has issues with handling of coercion. For example I imagine something like this ought to be valid:

```
loop {
     // ...
     break (expr: SomeT);
     // ...
     break (expr: OtherT);
}: RT
```

where `SomeT` and `OtherT` both coerce to `RT`. If *this* discrepancy was fixed, the list could avoid having to contain the last two notes/requirements. I also dislike those two requirements, because they have potential to diverge from type rules elsewhere in the language.

(Why this matters? Substitute `SomeT = &[u8; 42]`, `OtherT = &[u8; 32]`, `RT = &[u8]`.)

Ericson2314: I think the typing rule here can be exactly like that for match arms or if-else.

text/0000-loop-break-value.md
dhardy: The RFC is written in English, not code. *shrug*

text/0000-loop-break-value.md
dhardy: Please write a diff against the RFC; I was not 100% sure how to handle coercions (same with the `!` special case).

JelteF:
I just created a new issue to discuss other implementation details for returing from `for` and `while` than the the name of keywords. This issue can be found here for other people that are interested: https://github.com/rust-lang/rfcs/issues/1767

burdges:
Just noticed this slightly confusion `loop` that is not a loop : 
```
let x = 'a loop { 
    ...
    if foo { break `a bar; }
    ...
    break `a baz;
};
```
It's obvious doing this with a closure would be more readable, but that's frequently true for `loop`s too, if one had tail call optimization, so.. 

Is it desirable to support a less confusing form of this?
```
let x = 'a { 
    ...
    if foo { break `a bar; }
    ...
    baz
};
```
If so, does `continue` work in such a block?  I presume no.

glaebhoerl:
@burdges That's the orthogonal (but synergistic) "break from any block" feature, I'm surprised I can't find any existing issue or RFC for it. cc @Ericson2314?

Ericson2314:
@glaebhoerl I don't think anybody wrote an RFC but we've both brought it up many times (e.g. ? And catch). Maybe it will finally stick :).

vitiral:
@burdges this would not be backwards compatible. If I had code like:

```
loop {
    // do something
    {
        let foo = Bar();
        if x == foo {
            break;
        }
    }
}
```
currently this would break out of the `loop` block, but if we implemented your code it would break out of the inner block (and the loop would continue). I also don't like making `break` into a kind of `goto` (even though it often fulfills that role). If we want something like a generalized `goto` we should reserve a word for it and implement it separately rather than overload `break`

burdges:
It's obviously "break from named blocks" @vitiral not literally "break from any block", so no compatibility issues.  I donnot if it's a good idea or not, just thought it might be worth asking.

aturon:
RFC has been merged! 

Further discussion should happen on the [tracking issue](https://github.com/rust-lang/rust/issues/37339).

PR #1632, opened 2016-05-27
Replace synchronization primitives with those from parking_lot

Amanieu:
[Rendered](https://github.com/Amanieu/rfcs/blob/parking_lot/text/0000-parking-lot.md)

### DIFF
--- a/text/0000-parking-lot.md
+++ b/text/0000-parking-lot.md
@@ -49,6 +49,7 @@ in the Rust standard library:
 11. `Condvar::notify_all` will only wake up a single thread and requeue the rest
     to wait on the associated `Mutex`. This avoids a thundering herd problem
     where all threads try to acquire the lock at the same time.
+12. `RwLock` supports atomically downgrading a write lock into a read lock.
 
 Here are some benchmark results of `parking_lot` synchronization primitives
 compare to those of the standard library, showing the rate of lock acquisition:
--- a/text/0000-parking-lot.md
+++ b/text/0000-parking-lot.md
@@ -1,0 +1,102 @@
+- Feature Name: parking_lot
+- Start Date: 2016-05-27
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes replacing the `Mutex`, `Condvar`, `RwLock` and `Once` types in
+the standard library with those from the [`parking_lot`](https://github.com/Amanieu/parking_lot) crate. The synchronization
+primitives in the `parking_lot` crate are smaller, faster and more flexible than
+those in the Rust standard library.
+
+# Motivation
+[motivation]: #motivation
+
+The primitives provided by `parking_lot` have several advantages over those
+in the Rust standard library:
+
+1. `Mutex` and `Once` only require 1 byte of storage space, while `Condvar`
+   and `RwLock` only require 1 word of storage space. On the other hand the
+   standard library primitives require a dynamically allocated `Box` to hold
+   OS-specific synchronization primitives. The small size of `Mutex` in
+   particular encourages the use of fine-grained locks to increase
+   parallelism.
+2. Since they consist of just a single atomic variable, have constant
+   initializers and don't need destructors, these primitives can be used as
+    `static` global variables. The standard library primitives require
+   dynamic initialization and thus need to be lazily initialized with
+   `lazy_static!`.
+3. Uncontended lock acquisition and release is done through fast inline
+   paths which only require a single atomic operation.
+4. Microcontention (a contended lock with a short critical section) is
+   efficiently handled by spinning a few times while trying to acquire a
+   lock.
+5. The locks are adaptive and will suspend a thread after a few failed spin
+   attempts. This makes the locks suitable for both long and short critical
+   sections.
+6. `Condvar`, `RwLock` and `Once` work on Windows XP, unlike the standard
+   library versions of those types.
+7. `RwLock` takes advantage of hardware lock elision on processors that
+   support it, which can lead to huge performance wins with many readers.
+8. `MutexGuard` (and the `RwLock` equivalents) is `Send`, which means it can be
+   unlocked by a different thread than the one that locked it.
+9. `RwLock` will prefer writers, whereas the standard library version makes no
+   guarantees as to whether readers or writers are given priority.
+10. `Condvar` is guaranteed not to produce spurious wakeups. A thread will only
+    be woken up if it timed out or it was woken up by a notification.
+11. `Condvar::notify_all` will only wake up a single thread and requeue the rest
+    to wait on the associated `Mutex`. This avoids a thundering herd problem
+    where all threads try to acquire the lock at the same time.
+
+# Detailed design
+[design]: #detailed-design
+
+The API of `Mutex`, `Condvar`, `RwLock` and `Once` will mostly stay the same.
+The only user-visible API changes are the following:
+
+- `Once` is no longer required to be `'static`.
+- `MutexGuard`, `RwLockReadGuard` and `RwLockWriteGuard` will be `Send` if the
+  underlying type is also `Send`. This allows them to be unlocked from a
+  different thread than the one that created them.
+- `Condvar` is guaranteed not to produce any spurious wakeups. A thread will
+  only be woken up if its wait times out or if the `Condvar` is notified by
+  another thread.
+- `Condvar` is no longer restricted to being associated with a single `Mutex`
+  for its entire lifetime. The only restriction is that you cannot wait using
+  a `Mutex` if there are currently threads waiting on the `Condvar` with a
+  different `Mutex` (this is the same restriction that pthreads has). This
+  situation is detected and a panic will be generated.
+- `Mutex`, `Condvar` and `RwLock` will have `const fn` constructors and no not
+  require any drop glue. This makes them suitable for use in `static` variables.
+- Calling `RwLock::read` when already holding a read lock may result in a
+  deadlock if there is a writer thread waiting. Note that this was already the
+  case in the Windows `RwLock` but it is now explicitly documented.
+
+The internal parking lot APIs `park`, `unpark_one`, `unpark_all` and
+`unpark_requeue` are not publicly exposed in the standard library API. Users
+who wish to use these to create their own synchronization primitives should use
+the `parking_lot` crate directly.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+`Mutex`, `Condvar` and `RwLock` are no longer simple wrappers around OS primitives.
+
+The implementation of `parking_lot` is quite complicated because it needs to
+support many advanced features like thread requeuing, hardware lock elision and
+spin waiting.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The main alternative is to keep the existing synchronization primitives as they
+are, which is essentially wrappers around OS synchronization primitives. This is
+undesirable since there are many issues with these, such as the lack of support
+for Windows XP or glibc's support for lock elision causing memory safety issues.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
--- a/text/0000-parking-lot.md
+++ b/text/0000-parking-lot.md
@@ -54,6 +54,7 @@ Here are some benchmark results of `parking_lot` synchronization primitives
 compare to those of the standard library, showing the rate of lock acquisition:
 - [x86_64 Linux](https://gist.github.com/Amanieu/6a4b4151b89b78224992106f9bc4374f)
 - [x86_64 Windows](https://gist.github.com/Amanieu/6812507e66c5cbaa6ab5ab04d9c71eac)
+- [AArch64 Linux](https://gist.github.com/Amanieu/0f10ea5b2acb2819b75442390f2855f8)
 
 # Detailed design
 [design]: #detailed-design
--- a/text/0000-parking-lot.md
+++ b/text/0000-parking-lot.md
@@ -50,6 +50,11 @@ in the Rust standard library:
     to wait on the associated `Mutex`. This avoids a thundering herd problem
     where all threads try to acquire the lock at the same time.
 
+Here are some benchmark results of `parking_lot` synchronization primitives
+compare to those of the standard library, showing the rate of lock acquisition:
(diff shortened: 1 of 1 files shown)


AndrewBrinker:
Just as a first comment, I'm not sure how the parking_lot primitives can be adopted in favor of the current standard library primitives without violation of Rust's stability guarantees.

ticki:
If this can be done in a way that is compatible, I would absolutely _love_ to do this.

Amanieu:
> Just as a first comment, I'm not sure how the parking_lot primitives can be adopted in favor of the current standard library primitives without violation of Rust's stability guarantees.

As mentioned in the RFC, the parking_lot primitives have the same API as the standard ones, so there is no compatibility issue if we replace one with the other.

nagisa:
I expected to see benchmarks on all tier 1 and at least some tier 2 platforms included into the RFC text as the reason to/not to accept the RFC.

Amanieu:
@nagisa I added some benchmark results for Linux and Windows. Unfortunately I don't have any other platforms to test on.

rkruppe:
I seem to recall from the Webkit article that the parking-lot-style locks are not entirely fair. While I also recall good reasons why this isn't so bad, I would expect that it (as well as any other downsides that may exist &mdash; for example does the crate leak memory as Webkit does?) are honestly discussed in the RFC text.

nagisa:
I personally haven’t came across a single case where entirely-fair locks were necessary or even desirable. Is there one? I don’t recall us ever guaranteeing this for any locking mechanism in the standard library so far.

Either way, from the [same webkit blog post](https://webkit.org/blog/6161/locking-in-webkit/) it seems like making locks fair is also an option.

rkruppe:
I'm not particularly keen on fair locks either (then again, I'm not really qualified to judge anyway). I just want *all* the differences, including the ones that may be irrelevant or drawbacks, documented in the RFC to make it easier to form an opinion without scavenging N articles across the internet. In particular if there are so many minor design decisions where `parking_lot` may differ from Webkit's implementation.

Amanieu:
Neither pthread mutexes or Windows SRWLOCK make any fairness guarantees, so there is no difference here.

comex:
👍 needing an allocation for each mutex is awful.

oli-obk:
The rfc does not mention, that `Mutex` doesn't get poisoned anymore when the thread panics. This can probably be abused to exhibit UB and definitely changes UX

Amanieu:
I have poisoning support for parking_lot in a [branch](https://github.com/Amanieu/parking_lot/tree/poison), and this is the version that would be integrated into the standard library.

I don't plan on adding poisoning support to the parking_lot crate itself because poisoning has a significant performance cost and I feel that it makes the API uglier. By the way, lack of poisoning doesn't make anything unsafe, it just means that user code needs to be aware that locked data may be left in an inconsistent (but safe) state after a panic. This is why `AssertUnwindSafe` does not require `unsafe`.

Amanieu:
As an example, on AArch64, adding support for poisoning makes `Mutex` 70% slower. Most of this overhead comes from the need to call `std::thread::panicking` which in turn accesses TLS.

kennytm:
Is it possible to turn off poisoning when we are linking to the `abort` panic runtime?

nagisa:
Abort still runs various handlers before actually quitting process, therefore all data behind mutex that may be accessed there should also be poisoned.

ticki:
@nagisa It only runs the panic hook, nothing else.

ticki:
So, it should be possible (abort doesn't call destructors anyways).

nagisa:
You may have mutexes in statics and the panic hook may run arbitrary code.

Amanieu:
@nagisa With panic = abort a mutex can never *become* poisoned because that is done in the destructor of `MutexGuard`, which will poison the mutex if the current thread is unwinding due to a panic.

ticki:
@nagisa The point is that the inner value cannot become poisoned because the destructor won't be called.

eternaleye:
One thing I feel is worth noting - this lock is a traditional test-and-compare-and-swap lock, which a type which degrades *badly* as cores increase due to cache-coherency traffic[1]. This can be somewhat ameliorated by adding exponential backoff, but the worst case is actually unaffected by that. In the worst-case, where lock hold time is short and core count is above a (realistically reachable) threshold, TACAS locks suffer *catastrophic* throughput failure[2] compared to MCS, CLH, or other "scalable" locks.

[1] http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.103.8168&rep=rep1&type=pdf
[2] https://pdos.csail.mit.edu/papers/linux:lock.pdf

Amanieu:
@eternaleye `WordLock` uses the same basic algorithm as `Mutex`, which is to spin a few times trying to acquire the lock ([with exponential backoff](https://github.com/Amanieu/parking_lot/blob/master/src/spinwait.rs#L89)) but then fall back to adding the thread to a queue so that it is woken up later.

eternaleye:
@Amanieu: Ah, had missed the backoff. However, I do strongly suggest reading my second link - it shows catastrophic throughput loss for the ticket spinlocks in the Linux kernel of the time, and compares against their MCS lock implementations patched into Linux.

Amanieu:
@eternaleye 

One big assumption that this paper makes is that all "threads" are always running and never scheduled out. This is a perfectly fine assumption to make in the context of a kernel where a "thread" is a raw CPU. However this doesn't necessarily apply in the context a userspace process where the scheduler will arbitrarily suspend threads for a (relatively) long duration.

And this leads to the big issue with queue-based locks: you have no way of knowing whether the next thread in the queue has been scheduled out. If it has then you have one whole timeslice during which nobody owns the lock. If you instead have threads spinning on the lock word waiting for it to get unlocked then you are pretty much guaranteed to pass the lock to a thread that is running.

Now, with that said, the spin waiting strategy in parking_lot is a bit ad-hoc and could use some fine-tuning (read: I made up some numbers and it seemed to give good results). In particular I'm considering reducing the number of times I yield the thread in favor of just queuing up the thread and having it sleep until explicitly woken up.

arthurprs:
In order to really discuss this we need benchmarks with the poisoned branch on most tier A platforms.

Amanieu:
The benchmarks linked in the RFC were done on the poison branch.

Amanieu:
If someone else want to contribute benchmark number for their machine, just clone the git repository (on the `poison` branch) and run these commands:

```
cd benchmark
cargo run --features nightly --release --bin mutex 1:8 1 0 1
cargo run --features nightly --release --bin rwlock 0:4 0:4 1 0 1
```

brson:
My understanding is that parking lot does not support priority inheritance, while locks built on the OS primitives do (if the OS does). That's a major semantic difference.

I think it is far to early to consider this RFC. Let people use parking lot for a while and prove that it is good and stable.

alexcrichton:
I personally agree with @brson that it's far too early to consider this RFC. The `parking_lot` crate looks great, and it seems like it's definitely got some solid foundations as well, though! That being said these are also the **core synchronization primitives for all of Rust**, which in turn has a very high bar for an implementation. The `parking_lot` crate was first published only 20 days ago, and probably hasn't seen the level of usage that, e.g., pthread mutexes and Windows srwlocks have seen.

I would feel much more comfortable moving these into the standard library if the `parking_lot` crate was incredibly battle tested and basically the defacto standard among the Rust community, as that would give me quite a bit of confidence in both the implementation and benefits over the system primitives. I unfortunately do not have many concrete technical points I know of *against* this crate, but I would be incredibly wary of stating that we have soundly beaten all platform locking primitives in all respects in all situations. I probably don't even know about half the situations they were designed for!

This crate seems like a great candidate for the nursery, however, and perhaps even inclusion into rust-lang one day!

bjz:
parking_lot looks very exciting. But to echo what @brson and @alexcrichton  have said, it's probably not a good idea to rush into this. Let's see how it goes in practice first! Perhaps you could talk to the Servo community and see how it performs in comparison? I'm not how much they are using the std concurrency primitives at the moment though. Or perhaps try it on other libraries or projects in the ecosystem that you think would benefit from them?

briansmith:
I would like to see the RFC proposal extended to address the following issues:

1. Why not make each mutex a full page, and force them to be page-aligned, to minimize cache contention? Shouldn't this be provided, at least, as an option?

2. To what extent would this allow the synchronization primitives to be made available with `#[no_std]`/libcore-only configurations?

Amanieu:
> To what extent would this allow the synchronization primitives to be made available with #[no_std]/libcore-only configurations?

There are two reasons why this won't work on `#[no_std]`:

1. Condvar waits with timeout use the `std::time::{Instant,Duration}` types.

2. Putting a thread to sleep requires an implementation of [`ThreadParker`](https://github.com/Amanieu/parking_lot/tree/master/src/thread_parker), which needs to call OS-specific APIs.

gereeter:
A more conservative change would be to just move the OS-specific implementations of `ThreadParker` into the standard library. Currently, the stardard library always uses the generic implementation of `ThreadParker` to implement `thread::park` and `thread::Thread::unpark`, and `Once` and channels would be sped up significantly from having faster park implementations.

gereeter:
Additionally, I'd like to see benchmarks against the standard library implementation of `Once`. The standard library already uses a parking-based algorithm, and it'd be nice to see the difference between using a centralized parking lot and a queue for each `Once`.

Amanieu:
I expect the performance of `Once` to be roughly the same in both cases, and also mostly irrelevant since parking threads is only done in the rare slow path.

As for `ThreadParker`, there are no performance differences between my implementations and just using a `Mutex` + `Condvar`. Again, any performance benefit is vastly outweighed by the process of putting a thread to sleep. I only have custom implementations for two reasons: to support Windows XP and to avoid a dependency on the standard library `Mutex` and `Condvar` types.

briansmith:
> Why not make each mutex a full page, and force them to be page-aligned, to minimize cache contention? Shouldn't this be provided, at least, as an option?

Sorry, I meant to write "full cache line" and "cache-line-aligned." AFAICT, it should be really easy to provide both aligned-and-padded and single-byte variants.

Amanieu:
@briansmith What you really want to do is align the mutex *and the data it is protecting* to a cacheline. This can easily be done by marking `Mutex<T>` with `#[aligned = "64"]`. However this has not been implemented in rustc yet.

Ericson2314:
This would be a nice thing to optionally enable if we gave `std` some features. Would also making battle-testing much easier.

CodesInChaos:
What seems a bit risky with this change is that it not only replaces the implementation, but that it also makes additional promises. This restricts alternative implementations and prevents backing out of the change.

arthurprs:
@CodesInChaos the guarantees can always be reduced and this would still give good improvements in the performance side. 

Anyway I fell like this is a great proposal, we just need to wait a few months (let parking lot prove itself) before really making a decision.

Amanieu:
The `RwLock` type in parking_lot now supports atomically downgrading a write lock into a read lock. This fixes rust-lang/rust#32527.

bstrie:
Regarding real-world usage of parking_lot, today has brought us http://plhk.ru/trash/netmap-20160710.pdf .

alexcrichton:
:bell: This RFC is now entering its week-long **final comment period** 🔔 

The libs team felt that this RFC's discussion has reached a standstill now and as a result it's ready to move into FCP. Our feeling matches [what I wrote earlier](https://github.com/rust-lang/rfcs/pull/1632#issuecomment-222853398) about how `parking_lot` may not be quite ready just yet to replace all the primitives in the standard library as we feel it needs some mroe time to prove itself out across platforms and implementations.

tmiasko:
Having an implementation of synchronization primitives directly in standard
library has also an additional benefit in the form of more consistent behaviour
across supported platforms.  This is not true when building them on top of
OS-specific primitives, for example:

- On some platforms pthead_mutex_t contains embedded mutex and
  pthread_mutex_init never fails (unless misused). On the others it is merely a
  pointer, and initialization requires additional memory allocation (on top of
  the one that is required to box the handle itself). This is the case for
  example in FreeBSD, OpenBSD, and NetBSD.

- Statically allocated pthread mutexes are lazy initialized in
  pthread_mutex_lock, this means that this function can also fail, for the same
  reasons that initialization may fail.

- Non error-checking mutexes are allowed to perform deadlock detection, 
  and pthread_mutex_lock could fail with EDEADLK.

- pthread_cond_timedwait uses system clock by default, and doesn't match the
  documented semantics of Condvar::wait_timeout in case system time is changed
  (workaround contained in standard library doesn't really help when time is
  moved backwards).

... (and many others, not to mention differences between Windows API).


It seems that at some point it would be a really nice thing to have.

aturon:
@tmiasko 

> It seems that at some point it would be a really nice thing to have.

I agree, there are lots of potential benefits to offering these constructs. But it's a pretty big move to make, and for the time being gaining more experience with them externally seems prudent. (In general, it's not 100% clear to me what direction the concurrency story within `std` should go long term.)

alexcrichton:
The libs team got a chance to talk about this RFC this week at libs triage, and our conclusion remains the [same as before](https://github.com/rust-lang/rfcs/pull/1632#issuecomment-233538985) in that while `parking_lot` seems like a great crate with some really nice benefits, we'd prefer to wait for it to get more battle tested before moving it wholesale into the standard library. Thanks though for the RFC @Amanieu!

techtonik:
Can this still be included as a fallback mechanism to allow Rust to run not only on shiny new Windows?

bstrie:
Another blog post mentioning parking_lot: https://blog.mozilla.org/nfroyd/2017/03/29/on-mutex-performance-part-1/

It's been about a year since the crate was first published, so at some point it would be worth thinking about a timeline or otherwise mulling over specific criteria for potentially moving forward on this. Rather than trying to go into the stdlib directly, maybe a more reasonable first step would be to get this into rust-lang-nursery and see where it goes from there.

Ericson2314:
I can easy imagine adding some Cargo features to make this opt-in (c.f. allocator choice).

sanmai-NL:
@aturon: it seems @bstrie’s last post warrants a comment/update from your team’s side, now that about 1,5 years have passed since its decision to hold this off.

TerjeBr:
It has been a long time now. Should this pull request be reopened?

Ericson2314:
The work of the up-and-comming portability time this year will hopefully allow this for free.

faern:
Work on this has been going on for quite a while over at https://github.com/rust-lang/rust/pull/56410.

PR #1633, opened 2016-05-27
Make mem::uninitialized() and mem::zeroed() const fn

Amanieu:
[Rendered](https://github.com/Amanieu/rfcs/blob/const_uninitialized/text/0000-const-uninitialized.md)

### DIFF
--- a/text/0000-const-uninitialized.md
+++ b/text/0000-const-uninitialized.md
@@ -1,0 +1,97 @@
+- Feature Name: const_uninitialized
+- Start Date: 2016-05-27
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Turn `std::mem::uninitialized()` and `std::mem::zeroed()` into `const fn` so
+that they can be used to initialize `static` variables.
+
+# Motivation
+[motivation]: #motivation
+
+The main motivation for this is to allow static variables to be initialized to
+and undefined value which is guaranteed to be overwritten later on. It is mainly
+intended to be used in macros like `lazy_static!` and `thread_local!` rather
+than invoked by users directly. This would allow these macros to avoid the need
+to wrap values in an `Option` which can be statically initialized as `None`.
+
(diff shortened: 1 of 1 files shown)


nagisa:
I’m inclined to not do anything like this before we get const fn implemented proper, not hacked onto the ye olde const_eval.

Amanieu:
Actually I think there is a better way of doing this by using unions:
```rust
union MyUnion<T> {
    data: T,
    no_data: (),
}

static X: MyUnion<T> = MyUnion { no_data: () };
```

So maybe this RFC isn't needed after all...

Ericson2314:
@nagisa I'll agree based on how close miri is :).
@Amanieu I rather only use unions for FFI. They add zero safety here over the original RFC. And practically, I think unions are a longer way off.

Amanieu:
The advantage of using unions is that it will cleanly handle `Some(uninitialized)` since the `no_data` in the union inhibits enum layout optimizations.

Ericson2314:
Hmm? Aren't we getting rid of the option for the proposed usecase? What layout optimizations are there without a tag? I am really confused by that comment.

rkjnsn:
@Ericson2314 My understanding is that with this RFC as proposed, it would be possible to do something like this:

```rust
static X: Option<&T> = unsafe { Some(std::mem::zeroed()) };
```

Because of the enum layout optimization, the final value of `X` at runtime would be `None`. The compiler would have to deal with this during compile-time constant evaluation.

Without this RFC, but with unions, one would instead write

```rust
static X: Option<MyUnion<&T>> = Some(MyUnion { no_data: () });
```

The union would inhibit the layout optimization, so the `Option` would have a tag, avoiding the corner case.

(Note that I don't think anyone would write either of these in actual code. The point is to demonstrate that the former introduces a compile-time corner case that the latter does not.)

alexcrichton:
Thanks for the RFC @Amanieu! The libs team discussed this today and the conclusion was that this likely doesn't need an RFC, just a compiler implementation. That is, from the libs perspective, we're fine doing this any time. In light of that I'm going to close, but PRs are of course always welcome!

PR #1638, opened 2016-06-04
add: duration_checked_add RFC

benaryorg:
This would, as described in the RFC add the method `Duration::checked_add()`.

### DIFF
--- a/text/0000-duration-checked-sub.md
+++ b/text/0000-duration-checked-sub.md
@@ -1,0 +1,92 @@
+- Feature Name: duration_checked_sub
+- Start Date: 2016-06-04
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+This RFC adds `checked_sub()` already known from various primitive types to the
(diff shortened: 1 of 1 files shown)


durka:
add or sub?

benaryorg:
Wow. During writing *add* was just stuck in my head. I always wrote add, but I actually meant sub.

Yes it is sub.

I read the Readme and it states that no rebases, squashes or similar should be used after sending a PR, so, is it okay to just change the commit message?

It's in the title though, so it might be better to just close this PR and open a new one, what do you think?

kennytm:
Why don't we include `(checked|saturating|overflowing|wrapping)_(add|sub|mul|div)` as well?

benaryorg:
Okay, forget it, I'm closing this now.

I am sorry.

benaryorg:
I would actually leave this to others but or at least wait for it to come up in this discussion.

benaryorg:
I'll close this now and send another PR, this time hopefully correct.

I am sorry.

PR #1642, opened 2016-06-06
add CStr::with_ptr and deprecate CStr::as_ptr

durka:
Spawned from rust-lang/rust#34111 (cc @oli-obk).

[Rendered](https://github.com/durka/rfcs/blob/cstr-with-ptr/text/0000-cstr-with-ptr.md)

### DIFF
--- a/text/0000-cstr-with-ptr.md
+++ b/text/0000-cstr-with-ptr.md
@@ -107,6 +107,7 @@ of the closure:
 [alternatives]: #alternatives
 
 - Do nothing. It remains very easy to write unsound code using `CString` and `CStr::as_ptr`.
+- Just make the warnings louder and the font larger in the docs for `CStr::as_ptr`. Continue to assume that people read this documentation despite the fact that the documentation examples show correct usage, but incorrect usage abounds in the ecosystem.
 - Move the `temporary_cstring_as_ptr` lint, which warns on the most common way to write said unsound
 code (calling `as_ptr` on a temporary `CString`) from Clippy to rustc.
 - Deprecate `as_ptr` and introduce a new function that does the same thing (we would have to bikeshed
--- a/text/0000-cstr-with-ptr.md
+++ b/text/0000-cstr-with-ptr.md
@@ -1,0 +1,126 @@
+- Feature Name: `cstr_with_ptr`
+- Start Date: 2016-06-06
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Deprecate `CStr::as_ptr`, which returns a raw pointer and thereby invites unsound code, in favor of
+a new method, `CStr::with_ptr`, which passes the pointer to a user-provided closure.
+
+# Motivation
+[motivation]: #motivation
+
+`CString` wraps a C string and provides easy, safe access to the raw pointer to the string, for
+FFI interop. The safety is correct, as there's nothing unsound about having a raw pointer around
+until you dereference it. However, raw pointers are not tracked by the borrow checker, and once you
+have one it's as easy in Rust as it is in C to keep it around too long (namely, after the backing
+`CString` is dropped) and have a dangling pointer, which leads to unsound code.
+
+There are [hundreds of projects](https://users.rust-lang.org/t/you-should-stop-telling-people-that-safe-rust-is-always-safe/6094/7)
+including a call like `CString::new(...).unwrap().as_ptr()`, which is evidence that this UB
+mistake is widespread.
+
+By changing the API from a method _returning_ a raw pointer to one that runs a user-provided
+closure, we make it easier to write sound code than to write unsound code. Sure, you can still
+have the pointer escape the closure (see [Drawbacks](#drawbacks)), but it's a clear choice. The
+API change tips the scales towards soundness.
+
+# Detailed design
+[design]: #detailed-design
+
+1. Deprecate `CStr::as_ptr`, with the following message: "too easy to misuse because it returns a
+raw pointer; use `with_ptr` instead". The method is stable, so it will not be removed in any 1.x
+release.
+
+2. Add the following method to `CStr`:
+
+    ```rust
+        /// Calls the provided closure, passing the inner pointer to this C string.
+        ///
+        /// The pointer will be valid for as long as `self` is and points to a contiguous region of
+        /// memory terminated with a 0 byte to represent the end of the string.
+        #[allow(deprecated)] // for the as_ptr call
+        fn with_ptr<F: FnOnce(*const c_char)>(&self, f: F) {
+            f(self.as_ptr());
+        }
+    ```
+
+    For example usage see [this playpen](https://play.rust-lang.org/?gist=b6b1495ebee03fea679e95acb6b51ed6).
+
+3. Modify the `CStr` and `CString` examples that use `as_ptr` to use `with_ptr` instead.
+
+    `CString::new` has the following example code:
+
+    ```rust
+    use std::ffi::CString;
+    use std::os::raw::c_char;
+
+    extern { fn puts(s: *const c_char); }
+
+    fn main() {
+        let to_print = CString::new("Hello!").unwrap();
+        unsafe {
+            puts(to_print.as_ptr());
+        }
+    }
+    ```
+
+    Under this proposal, it would change to:
+
+    ```rust
+    use std::ffi::CString;
+    use std::os::raw::c_char;
+
+    extern { fn puts(s: *const c_char); }
+
+    fn main() {
+        let to_print = CString::new("Hello!").unwrap();
+        to_print.with_ptr(|p| unsafe {
+            puts(p);
+        });
+    }
+    ```
+
+    There are nearly identical examples on `CString` and `CStr` themselves which would change in the
+    same way.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- It deprecates another stable method, which contributes to perception of API churn.
+- It adds surface area to the API of `CStr`.
+- It's still rather easy to circumvent the help and write unsound code by "leaking" the pointer out
+of the closure:
+
+    ```rust
+    let mut ptr = ptr::null();
+    {
+        let s = CString::new("foo").unwrap();
+        s.with_ptr(|p| { ptr = p; });
+    }
+    /* ptr is now dangling */
+    ```
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Do nothing. It remains very easy to write unsound code using `CString` and `CStr::as_ptr`.
+- Move the `temporary_cstring_as_ptr` lint, which warns on the most common way to write said unsound
+code (calling `as_ptr` on a temporary `CString`) from Clippy to rustc.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should `with_ptr` allow the closure to return a value? It could be
+
+    ```rust
+        fn with_ptr<F: FnOnce(*const c_char) -> O, O>(&self, f: F) -> O {
+            f(self.as_ptr())
+        }
+    ```
+
+    which might be convenient but would make it easier to "leak" the pointer (as easy as
+    `let ptr = s.with_ptr(|p| p);`).
+
--- a/text/0000-cstr-with-ptr.md
+++ b/text/0000-cstr-with-ptr.md
@@ -127,3 +127,5 @@ the name), but mark it `unsafe` due to the potential for unsoundness in combinat
     which might be convenient but would make it easier to "leak" the pointer (as easy as
     `let ptr = s.with_ptr(|p| p);`).
 
+- Does `f(CString::new(...).unwrap().as_ptr())` actually invoke undefined behavior, if `f` doesn't store the pointer? The author's reading of the Rust reference implies that the `CString` temporary is kept alive for the entire expression, so it's fine. However, some commenters in the RFC thread have opined that the behavior of this code is unspecified at best.
+
--- a/text/0000-cstr-with-ptr.md
+++ b/text/0000-cstr-with-ptr.md
@@ -109,6 +109,8 @@ of the closure:
 - Do nothing. It remains very easy to write unsound code using `CString` and `CStr::as_ptr`.
 - Move the `temporary_cstring_as_ptr` lint, which warns on the most common way to write said unsound
 code (calling `as_ptr` on a temporary `CString`) from Clippy to rustc.
+- Deprecate `as_ptr` and introduce a new function that does the same thing (we would have to bikeshed
+the name), but mark it `unsafe` due to the potential for unsoundness in combination with unsafe code.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions


BurntSushi:
Is every use of `CString::new(...).unwrap().as_ptr()` always incorrect?

Are there any uses of `as_ptr` that are impossible or difficult to express using `with_ptr`? What if I want to return a `*const c_char` from a function?

oli-obk:
> Is every use of CString::new(...).unwrap().as_ptr() always incorrect?

No. If the entire expression is part of a function argument, then the current rustc makes sure that the drop call happens after the function returns. This is necessary for taking the reference to an owned value. It gets problematic with pointers, because they are just values. I'll do some opimization experiments whren I'm back on a PC

durka:
> Is every use of `CString::new(...).unwrap().as_ptr()` always incorrect?

No. If the call occurs in an expression, the `CString` is kept alive until the end of an expression. For example, if it is immediately passed to a function which does not store the pointer, then it is safe. [Here's an example](https://is.gd/lUGu3x) showing the drop timing.

> Are there any uses of as_ptr that are impossible or difficult to express using with_ptr? What if I want to return a *const c_char from a function?

The Drawbacks section shows how to emulate `as_ptr` using `with_ptr`. In particular:

```rust
fn get_ptr(s: &CString) -> *const c_char {
    let mut ptr = ptr::null();
    s.with_ptr(|p| { ptr = p; });
    ptr
}
```

nagisa:
> Is every use of CString::new(...).unwrap().as_ptr() always incorrect?

From my watchtower, using the pointer returned by `as_ptr` here is *always* UB, regardless of how compiler behaves currently.

durka:
@nagisa I would say changing drop order in expressions is a breaking change.

oli-obk:
Not if llvm already does that after optimizations... Which we should definitely figure out

durka:
My example prints the same in release configuration.

matklad:
>Are there any uses of as_ptr that are impossible or difficult to express using with_ptr?

If you have two `CStrings`, then you'll need two level of indents, which may be ugly. And if you have `Vec<CString>`, then you'll need some kind of a recursive function, which wight be significantly trickier than a simple loop. 

Another alternative is "to mark as_ptr with unsafe", of course with introducing a new name and deprecating and old one.

durka:
I guess we could add a function that works on sequences as well. Is working with slices or Vecs of CStrings a common use case?

> Another alternative is "to mark as_ptr with unsafe", of course with introducing a new name and deprecating the old one.

The community has been very reluctant to use `unsafe` to mark areas of potential-but-not-guaranteed misuse. Since you can't cause UB using _just_ `as_ptr` (without unsafe code somewhere else), I think this is a non-starter. But I'll add it to the alternatives section.

durka:
Another alternative possibility (maybe, haven't thought this all the way through) is to deprecate `CStr::as_ptr` and add `CStr::into_ptr` which consumes `self`. Then you can't call it through a deref coercion from `CString`. This seems like it adds some safety because as long as you've got an explicit `CStr` then there is non-temporary backing storage somewhere.

oli-obk:
Well... the true way to do this is to add a (future miri-based) const fn that does the `str` -> `CStr` conversion at compile-time and yields a `&'static CStr`. So we only need to think about the case where the contents of the `CString` depend on some runtime calculations.

Maybe we could create a method that freezes the `CString` object, and thus forcing it to be 

> My example prints the same in release configuration.

That's a test, not a guarantee that UB doesn't occur. Especially since a use after free might not even get noticed in such a small example (the memory might not be used again due to other allocations having requirements placing them elsewhere). 

Have a look at this example: https://is.gd/AEiXxM

Activating release mode and creating llvm IR shows the following fun IR (first block of the main function):

```
invoke void @f({}* null)
```

I forgot what combo caused it, but I've had 

```
invoke void @f({}* undef)
```

at some point or another.

matklad:
Has anybody thought of improving the [docs] for `as_ptr`? It might be a good idea to add something like this. 

<hr>
**WARNING**: make sure that the `CString` is alive while you are using the returned pointer. In particular, for 

```Rust
let p = let c_to_print = CString::new("Hello, world!").unwrap().as_ptr();
```
dereferencing `p` inside an unsafe block or passing it to a C function will almost surely trigger undefined behavior.

Use this instead:

```Rust
let hello_world = CString::new("Hello, world!").unwrap()
let p = let c_to_print = hello_world.as_ptr();
```

<hr>

I would have done this myself, but I am not sure about my English :( 


[docs]: https://doc.rust-lang.org/nightly/std/ffi/struct.CString.html#method.as_ptr

durka:
@oli-obk 

I'd be very interested to see how you get an undef there. null seems to be correct as the `CString::new` in that playpen calls `ptr::null`. In any case I consider it a bug, and only somewhat relevant to the RFC, if LLVM is changing drop orders between debug and release mode. Certainly if `undef` ever appears that seems like it can't be correct.

@matklad 

Docs are a part of the solution, but since all the examples show correct use of `as_ptr` (keep the `CString` alive, don't call `as_ptr` on a temporary) but people do it anyway, I don't think it's the full answer.

oli-obk:
> I'd be very interested to see how you get an undef there

Ah, apparently this occurs, if the function doesn't use the argument: https://is.gd/dKLVoH

durka:
One also wonders whether `str::as_ptr` and `<[T]>::as_ptr` should get the same treatment. That might be more churn than we want, and I bet those methods are less commonly used.

oli-obk:
both `str::as_ptr` and `<[T]>::as_ptr` require you to know the length of the object, but there's no way to obtain both the length and the pointer in the same expression without storing the object in a binding.

durka:
That's a very good point. So those functions are less prone to misuse.

bluss:
From experience in our user help channels, this feels like by far the most common use after free bug that users stumble into. Sometimes they find bugs and come ask because of it, sometimes the issue is found latently. It would be swell if a lint could catch this well, and very much worth it to introduce lints that help users to write sound Rust code in practice.

starkat99:
This gets really nasty if you have to work with multiple `CString`s at once, either nested or in a vector. I think a strong lint and extra documentation warnings would be good enough to address the issue.

durka:
@starkat99 I've added a simple macro to my [example](https://is.gd/wc6oyW) for that use case.

kylewlacy:
IIUC, there was a time in Rust's history (long before I started using it) where [`with`-style methods were common](https://doc.rust-lang.org/0.9/std/cell/struct.RefCell.html#method.with) :)

What about instead using a newtype over `*const c_char`? Something like:

```rust
struct CCharRef<'a> {
    ptr: *const c_char,
    _phantom: PhantomData<&'a [c_char]>
}

impl<'a> Deref for CCharRef<'a> {
    type Target = *const c_char;
    
    fn deref(&self) -> &*const c_char {
        &self.ptr
    }
}

impl CStr {
    // ...

    #[allow(deprecated)]
    fn as_ref(&self) -> CCharRef {
        CCharRef {
            ptr: self.as_ptr(),
            _phantom: PhantomData
        }
    }
}
```

Then, the following would be an error:

```rust
let s = CString::new(...).unwrap().as_ref();
//      ^~~~~~~~~~~~~~~~~~~~~~~~~~
//      error: borrowed value does not live long enough
```

(Note that this would still allow `let s = *CString::new(...).unwrap().as_ref()`, and the `CString` will still be dropped immediately)

wycats:
> From my watchtower, using the pointer returned by as_ptr here is always UB, regardless of how compiler behaves currently.

This is probably the #1 gotcha I've experienced when writing Rust bindings to Ruby code. I've probably lost a cumulative 5+ hours over the years from this error.

wycats:
In addition to this proposal to make it less error prone to dynamically construct C strings, I would like to see c"hello world", which would create a static null terminated string. In addition to the gotchas described here, I find it annoying that this API forces me to allocate just because I need a null terminated string.

I've tried to write macros like `cstr!("hello world")` to create `b"hello world\0"), but there's no way in stable macros to convert a string into a C string. I've tried `cstr!(b"hello world")`, but even that is not possible, as `concat!` doesn't work with binary strings.

matklad:
>In addition to this proposal to make it less error prone to dynamically construct C strings, I would like to see c"hello world", which would create a static null terminated string.

I would say that `b"hello world\0"` is much more obvious then `c"hello world"`, while it is only two characters longer. 

But what type would `c"hello world" have"`? The problem with `b"hello\0"` is that it is `[u8]`, while `c_char` is `i8`. This can be handled by `CStr::from_bytes_with_nul`, but this is a mouthful.

ubsan:
@wycats There is, I believe, this:

```rust
macro_rules! cstr {
    ($s:expr) => (concat!($s, "\0") as *const str as *const ::std::os::raw::c_char)
}
```

durka:
There are docs now (thanks @matklad!), but I still think this is a good idea. Any other opinions? It's been quiet for a while. FCP time?

alexcrichton:
I think the addition of `with_ptr` is fine to have, but I don't think that `as_ptr` should be deprecated. Long ago we used to *only* have `with_ptr` and then we added `as_ptr` in a general trend towards "let's cut down on that rightward drift".

The `as_ptr` itself is not incorrect 100% of the time and it is much more ergonomic much of the time as well, so I'm not personally convinced that deprecation is warranted for an API like that. There's a downside for leaving it stable, it can go wrong, but there's a downside for deprecating it, the replacement is less ergonomic.

iliekturtles:
If accepted I would like to see `with_ptr` return a value. My main use case is FFI where checking the return value of the foreign function is necessary.

rkjnsn:
-1 to removing `as_ptr`, as I find it more ergonomic in many cases. I personally don't think I'd choose to use `with_ptr`, but have no objection to adding it if other's would find it useful.

I would very much like to see a definitive answer to this unresolved question:

> Does `f(CString::new(...).unwrap().as_ptr())` actually invoke undefined behavior, if `f` doesn't store the pointer? The author's reading of the Rust reference implies that the `CString` temporary is kept alive for the entire expression, so it's fine. However, some commenters in the RFC thread have opined that the behavior of this code is unspecified at best.

Coming from C++, I would expect this to be valid, as temporaries there aren't destroyed until the end of the full expression (after `f` returns). If this is not the case in Rust, it should be very clearly documented as such.

alexcrichton:
The libs team discussed this briefly at the triage meeting yesterday. We felt that there definitely is a real problem here in terms of misusage of `as_ptr`, but we didn't quite feel that this RFC is the best solution to the problem.

This may be a case where the motivation may want to be fleshed out a little more before the detailed design is tackled. For example what exact cases are we targeted at solving and are we willing to compromise cases unrelated to this? (things like that)

durka:
[Here](http://stackoverflow.com/a/39171100/1114328) is another example of people writing unsound code, even though there is a giant warning in the documentation.

matklad:
Here is a more tricky example of this issue (for which `with_ptr` would be uneconomic): https://users.rust-lang.org/t/solved-how-to-export-vec-string-to-c-with-ffi/7121/3?u=matklad

An insane idea: What if `drop`in `CString` would overwrite the buffer with `b"DEAD MEMORY X_X"`?

durka:
Yeah, slices of strings have been brought up already as a clear case where
this API would be less ergonomic than the current one. Maybe we could pair
it with another API that works on slices or iterators or something.

Your insane idea doesn't work unless CString also always leaks its
allocation.

matklad:
>Your insane idea doesn't work unless CString also always leaks its allocation.

The point is not to make this memory safe, but to make this fail at runtime with an easy to google reason. We obviously can't add runtime range checks, but overwriting the buffer with some garbadge will almost certainly lead to the the loud failure, which is much better than silent UB. 

And perhaps we can "leak" CStrings in debug mode? We can allocate them from some kind of free list, such that deallocated CStrings are not immediately overwritting by random garbage, and instead are overwritten by our deterministic garbage :)

matklad:
Ah, looks like I can't simply

```Rust
#[cfg(debug_assertions)]
impl Drop for CString {
    fn drop(&mut self) {
        let pattern = b"X_X DEAD MEMORY ";
        let bytes = &mut self.inner[..self.inner.len() - 1];
        for (d, s) in bytes.iter_mut().zip(pattern.iter().cycle()) {
            *d = *s;
        }
    }
}
```

because the standard library is always build with `--release` :( 

Adding a `Drop` impl to the release build is a non-starter, isn't it?

text/0000-cstr-with-ptr.md
```diff
@@ -0,0 +1,131 @@
...
+    ```rust
+        fn with_ptr<F: FnOnce(*const c_char) -> O, O>(&self, f: F) -> O {
+            f(self.as_ptr())
+        }
+    ```
+
+    which might be convenient but would make it easier to "leak" the pointer (as easy as
+    `let ptr = s.with_ptr(|p| p);`).
+
+- Does `f(CString::new(...).unwrap().as_ptr())` actually invoke undefined behavior, if `f` doesn't store the pointer? The author's reading of the Rust reference implies that the `CString` temporary is kept alive for the entire expression, so it's fine. However, some commenters in the RFC thread have opined that the behavior of this code is unspecified at best.
```
arielb1: Sure. Temporaries stay alive at least up to the nearest arena (see https://github.com/nikomatsakis/rust-memory-model/issues/17), and function calls do not create these.

durka:
I'm going to close this as there seems to be significant resistance to adding friction to the `CString::as_ptr` interface, and @matklad's PR adds some sort of runtime defense against the bug that `as_ptr` invites.

goertzenator:
Wouldn't implementing `AsRef<c_char> for CString` and using it instead of `as_ptr()` solve this problem? `&c_char` coerces to `*const c_char` but still gets borrow checked.

``` Rust
use std::ffi::{CString, NulError};
use std::os::raw::c_char;

fn ffi_function(_hello: *const c_char) {}


// minimal wrapper for CString so we can add traits
struct MyCString(CString);
impl MyCString {
    fn new<T: Into<Vec<u8>>>(t: T) -> Result<MyCString, NulError> {
        CString::new(t).map(|c| MyCString(c))
    }
}

impl AsRef<c_char> for MyCString {
    fn as_ref(&self) -> &c_char {
        unsafe{ &*self.0.as_ptr() }
    }
}

fn main() {
    { // as_ptr() undefined behavior case
        let p =   CString::new("some_string").unwrap().as_ptr();
        ffi_function(p);
    }
    { // as_ptr() proper use case
        let s =   CString::new("some_string").unwrap();
        ffi_function(s.as_ptr());
    }
    { // as_ref() undefined behavior case, but borrow checker catches it
        let r = MyCString::new("some_string").unwrap().as_ref();
        ffi_function(r);
    }
    { // as_ptr() proper use case, reference coerces to pointer
        let s = MyCString::new("some_string").unwrap();
        ffi_function(s.as_ref());
    }
}
```

eddyb:
Ideally, `CString` and `&CStr` would be ABI compatible with `*const c_char`, so we can use them directly.

BatmanAoD:
I don't quite understand why this issue is closed. Doesn't this invalidate Rust's guarantee that safe-rust *cannot* have undefined behavior, unless there is an error in an `unsafe` block somewhere?

BatmanAoD:
...upon reflection, it seems that the unsafety is in the passing of a pointer by-value to a function. I realize this would be a breaking change, but perhaps the only way to truly uphold that guarantee would be to require any function that takes a raw pointer as a parameter to be marked `unsafe`?

BatmanAoD:
@sfackler Right, but there can be a *safe* public function `f` that takes a raw pointer and (unsafely) dereferences it. Since such a function cannot guarantee that the dereference will be safe, the function itself ought to be `unsafe`, I would think.

That said, when I wrote that comment I was under the impression (based [this old comment](https://users.rust-lang.org/t/you-should-stop-telling-people-that-safe-rust-is-always-safe/6094/8?u=batmanaod)) that there existed an `open` function in the standard library that took a raw pointer, but it seems that's not the case.

sfackler:
> Since such a function cannot guarantee that the dereference will be safe, the function itself ought to be `unsafe`, I would think.

Yes, that's correct. You can do things with raw pointers that don't involve dereferencing them, though. This function consumes a raw pointer and is totally safe to call with any value:

```rust
fn is_pointer_even(x: *mut u8) -> bool {
    x as usize % 2 == 0
}

fn main() {
    is_pointer_even(1 as *mut u8);
}
```

BatmanAoD:
Ah. You're right, of course, but that doesn't seem....particularly useful.

In any case, there _aren't_ any non-`unsafe` functions in the standard library that consume and dereference a raw pointer, correct?

sfackler:
That's correct.

PR #1647, opened 2016-06-13
Allow `Self` to appear in the where clause of trait impls

sgrif:
[Rendered](https://github.com/sgrif/rfcs/blob/sg-self-in-where-clauses/text/0000-allow-self-in-where-clauses.md)

### DIFF
--- a/text/0000-allow-self-in-where-clauses.md
+++ b/text/0000-allow-self-in-where-clauses.md
@@ -6,8 +6,8 @@
 # Summary
 [summary]: #summary
 
-This RFC proposes allowing the `Self` type to be used in where clauses for trait
-implementations, as well as referencing associated types for the trait being
+This RFC proposes allowing the `Self` type to be used in every position in trait
+implementations, including where clauses and other parameters to the trait being
 implemented.
 
 # Motivation
@@ -46,28 +46,40 @@ on the associated type. It would be nice to reduce some of that duplication.
 # Detailed design
 [design]: #detailed-design
 
-The first half of this RFC is simple. Inside of a where clause for trait
-implementations, `Self` will refer to the type the trait is being implemented
-for. It will have the same value as `Self` being used in the body of the trait
-implementation.
+Instead of blocking `Self` from being used in the "header" of a trait impl,
+it will be understood to be a reference to the implementation type. For example,
+all of these would be valid:
 
-Accessing associated types will have the same result as copying the body of the
-associated type into the place where it's being used. That is to say that it
-will assume that all constraints hold, and evaluate to what the type would have
-been in that case. Ideally one should never have to write `<Self as
-CurrentTrait>::SomeType`, but in practice it will likely be required to remove
-issues with recursive evaluation.
+```rust
+impl SomeTrait for SomeType where Self: SomeOtherTrait { }
+
+impl SomeTrait<Self> for SomeType { }
+
+impl SomeTrait for SomeType where SomeOtherType<Self>: SomeTrait { }
+
+impl SomeTrait for SomeType where Self::AssocType: SomeOtherTrait {
+    AssocType = SomeOtherType;
+}
+```
+
+If the `Self` type is parameterized by `Self`, an error that the type definition
+is recursive is thrown, rather than not recognizing self.
+
+```rust
+// The error here is because this would be Vec<Vec<Self>>, Vec<Vec<Vec<Self>>>, ...
+impl SomeTrait for Vec<Self> { }
+```
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-`Self` is always less explicit than the alternative
+`Self` is always less explicit than the alternative.
 
 # Alternatives
 [alternatives]: #alternatives
 
-Not implementing this, or only allowing bare `Self` but not associated types in
-where clauses
+Not implementing this is an alternative, as is accepting Self only in where clauses
+and not other positions in the impl header.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-allow-self-in-where-clauses.md
+++ b/text/0000-allow-self-in-where-clauses.md
@@ -1,0 +1,75 @@
+- Feature Name: `allow_self_in_where_clauses`
+- Start Date: 2016-06-13
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes allowing the `Self` type to be used in where clauses for trait
+implementations, as well as referencing associated types for the trait being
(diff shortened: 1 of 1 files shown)


durka:
Slightly related: https://github.com/rust-lang/rfcs/issues/1588

This seems like a good idea. The only backcompat thing I could come up with is this, and it already doesn't compile, so no issue:

```rust
trait Foo {
    fn foo(&self);
}

trait Bar {}

struct Baz;
struct Quux;

impl Foo for Baz {
    fn foo(&self) {
        impl Bar for Quux where Self: Sized {}
            // before this RFC: erroneous attempt to use outer `Self` (i.e. `Baz`)
            // after this RFC: `Self` here refers to `Bar`
    }
}

fn main() {}
```

Ericson2314:
IIUC, the associated type thing is weird in that `<Self as Trait>::Assoc: Bound` will sometimes fail, but then `<Self as Trait>::Assoc` isn't bound to begin with.

sgrif:
Not really any different than having `T: Trait, <T as Trait>::Assoc: Bound` where `T: Trait` doesn't hold.

Ericson2314:
Say a different impl with a different associated type succeeds?

Ericson2314:
If `<SelfWrittenOut as Trait>::Assoc: Bound` is already allowed, then nevermind.

sgrif:
`<ExplicitSelfType as Trait>::Assoc: Bound` is already allowed. I could probably sum this up more easily in that the goal is to have `Self` in `impl Trait for T where Self: IsUsedHere` to mean the same thing as `fn method_in_trait_body() where Self: IsUsedHere` in effectively every way

text/0000-allow-self-in-where-clauses.md
```diff
@@ -0,0 +1,75 @@
...
+from the traits being implemented. When dealing with generic code, patterns like
+this often emerge:
+
+```rust
+trait MyTrait {
+    type MyType: SomeBound;
+}
+
+impl<T, U, V> MyTrait for SomeStruct<T, U, V> where
+    SomeOtherStruct<T, U, V>: SomeBound,
```
Ericson2314: If you could change this from `SomeOtherStruct<T, U, V>` to `<Self as MyTrait>::MyType`, that would make clear the current verbosity is not quite as bad :).

Ericson2314:
@sgrif ah, sorry for the noise then. I think the goal is clear enough, but I commented on one line you might consider changing then to reflect the status-quo as you just clarified it to be.

nikomatsakis:
In general we already allow `Self` inside impls, iirc, so I am a bit surprised that it doesn't work in where clauses. Seems like an oversight or bug to me, to be honest.

When it comes to associated types, though, I am a bit dubious about this paragraph:

> Accessing associated types will have the same result as copying the body of the associated type into the place where it's being used. That is to say that it will assume that all constraints hold, and evaluate to what the type would have been in that case. Ideally one should never have to write <Self as CurrentTrait>::SomeType, but in practice it will likely be required to remove issues with recursive evaluation.

This seems to suggest some kind of distinct pathway for handling associated types in this location. I am not too keen on that, I would prefer for this to fall out from normalization (and indeed I think it very well might, though there may be some issues with resolving recursive references -- we could probably do better in this respect than we do in general).

withoutboats:
@rfcbot fcp merge

(I hope I'm doing this right!)

I agree with @nikomatsakis that this seems like a bug. The 'type header' of an impl block (its where clauses, etc) should be considered a part of that impl block, and the `Self` type should be understood there as the receiving type.

withoutboats:
@rfcbot concern "Include type parameters"

I think this should also be accepted:

```rust
trait Foo<T> { }

struct Bar;

impl Foo<Self> for Bar { }
// equivalent to:
impl Foo<Bar> for Bar { }
```

petrochenkov:
EDIT: @withoutboats was faster at typing, but I'll still write this.

Technically, `Self` can be used in the "whole impl" except for the impl type itself, this includes the trait and bounds on impl parameters:
```
trait Tr<T> {}

impl<T: Tr<Self>> Tr<Self> for u8 {}
```
The only question is "How confusing this may be?", especially if the impl lives in another impl with its own `Self`.
The impl syntax is arbitrary, if it looked like `for u8 impl Trait<Self>`, probably it would look more intuitive, because `Self` is used after percieved point of "type introduction".

petrochenkov:
@withoutboats 
```
impl Foo for Bar<Self> { }
// equivalent to
impl Foo for Bar<Bar<Self>> { }
// equivalent to
impl Foo for Bar<Bar<Bar<Self>>> { }
// equivalent to
impl Foo for Bar<Bar<Bar<Bar<Self>>>> { }
// equivalent to
impl Foo for Bar<Bar<Bar<Bar<Bar<Self>>>>> { }
...
```

withoutboats:
@petrochenkov of course :sweat_smile: it should fail with some recursive type definition, and not a "use of `Self` outside of an impl or trait" error

petrochenkov:
>fail with some recursive type definition error, and not a "use of Self outside of an impl or trait" error

That... sounds very reasonable.

nikomatsakis:
Wait, I am confused. The RFC concerns using `Self` in *where-clauses*, right? Do we intend to permit `impl Trait<Self> for u8`?

withoutboats:
@nikomatsakis I would like to see the RFC amended to support that.

petrochenkov:
For reference, trait items, unlike impl items, already permit `Self` everywhere and not only inside of the block.
This, for example, compiles:
```
trait Tr<T: Tr<Self> = Self> where Self: Copy {}
```

nikomatsakis:
@withoutboats huh, really? Interesting. I guess it seems intuitive enough what it means, even if I find it a bit surprising. Presumably `Self` would also be "in scope" for bounds, right?

e.g.,

```rust
impl<T: Foo<Self>> Bar for u8 
```

withoutboats:
@nikomatsakis Yea. I agree that it seems weird to actually do, but we have a history of `Self` not always working for no particular reason, which IMO teaches users that Rust has complicated and arbitrary rules. :-\ I'd like to see the system be consistent between impls and traits - in any position in any impl or trait, `Self` is the receiving type.

nikomatsakis:
@withoutboats yeah, seems ok. I can get behind it.

withoutboats:
@rfcbot resolved "Include type parameters"

briansmith:
@sgrif Could you please write a patch to the Rust Reference for this change? Perhaps it would be an empty patch since the Rust Reference already says "All traits define an implicit type parameter Self that refers to "the type that is implementing this interface," which supports the conclusion above that this is fixing a bug.

ubsan:
Yay!

PR #1651, opened 2016-06-15
Extend Cell to non-Copy types

Amanieu:
This RFC extends the `Cell` type to work with non-`Copy` types.

[Rendered](https://github.com/Amanieu/rfcs/blob/movecell/text/0000-movecell.md)

### DIFF
--- a/text/0000-movecell.md
+++ b/text/0000-movecell.md
@@ -24,7 +24,7 @@ impl<T> Cell<T> {
 }
 
 impl<T: Copy> Cell<T> {
-    fn get(&self);
+    fn get(&self) -> T;
 }
 
 impl<T: Default> Cell<T> {
@@ -42,6 +42,8 @@ Finally, a `take` method is added which is equivalent to `self.replace(Default::
 [drawbacks]: #drawbacks
 
 It makes the `Cell` type more complicated.
+
+`Cell` will only be able to derive traits like `Eq` and `Ord` for types that are `Copy`, since there is no way to non-destructively read the contents of a non-`Copy` `Cell`.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/text/0000-movecell.md
+++ b/text/0000-movecell.md
@@ -18,8 +18,8 @@ It allows safe inner-mutability of non-`Copy` types without the overhead of `Ref
 
 ```rust
 impl<T> Cell<T> {
-    fn set(&mut self, val: T);
-    fn replace(&mut self, val: T) -> T;
+    fn set(&self, val: T);
+    fn replace(&self, val: T) -> T;
     fn into_inner(self) -> T;
 }
 
@@ -28,7 +28,7 @@ impl<T: Copy> Cell<T> {
 }
 
 impl<T: Default> Cell<T> {
-    fn take(&mut self) -> T;
+    fn take(&self) -> T;
 }
 ```
 
--- a/text/0000-movecell.md
+++ b/text/0000-movecell.md
@@ -1,0 +1,54 @@
+- Feature Name: move_cell
+- Start Date: 2016-06-15
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Extend `Cell` to work with non-`Copy` types.
+
+# Motivation
+[motivation]: #motivation
+
+It allows safe inner-mutability of non-`Copy` types without the overhead of `RefCell`'s reference counting.
+
+# Detailed design
+[design]: #detailed-design
+
+```rust
+impl<T> Cell<T> {
+    fn set(&mut self, val: T);
+    fn replace(&mut self, val: T) -> T;
+    fn into_inner(self) -> T;
+}
+
+impl<T: Copy> Cell<T> {
+    fn get(&self);
+}
+
+impl<T: Default> Cell<T> {
+    fn take(&mut self) -> T;
+}
+```
+
+The `get` method is kept but is only available for `T: Copy`. The `set` method is available for all `T`.
+
+The `into_inner` and `replace` methods are added, which allow the value in a cell to be read even if `T` is not `Copy`. The `get` method can't be used since the cell must always contain a valid value.
+
+Finally, a `take` method is added which is equivalent to `self.replace(Default::default())`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It makes the `Cell` type more complicated.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternative is to use the `MoveCell` type from crates.io which provides the same functionality.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
--- a/text/0000-movecell.md
+++ b/text/0000-movecell.md
@@ -32,7 +32,9 @@ impl<T: Default> Cell<T> {
 }
 ```
 
-The `get` method is kept but is only available for `T: Copy`. The `set` method is available for all `T`.
+The `get` method is kept but is only available for `T: Copy`.
+
+The `set` method is available for all `T`. It will need to be implemented by calling `replace` and dropping the returned value. Dropping the old value in-place is unsound since the `Drop` impl will hold a mutable reference to the cell contents.
 
 The `into_inner` and `replace` methods are added, which allow the value in a cell to be read even if `T` is not `Copy`. The `get` method can't be used since the cell must always contain a valid value.
 
--- a/text/0000-movecell.md
+++ b/text/0000-movecell.md
@@ -13,6 +13,10 @@ Extend `Cell` to work with non-`Copy` types.
 
 It allows safe inner-mutability of non-`Copy` types without the overhead of `RefCell`'s reference counting.
 
+The key idea of `Cell` is to provide a primitive building block to safely support inner mutability. This must be done while maintaining Rust's aliasing requirements for mutable references. Unlike `RefCell` which enforces this at runtime through reference counting, `Cell` does this statically by disallowing any reference (mutable or immutable) to the data contained in the cell.
+
+While the current implementation only supports `Copy` types, this restriction isn't actually necessary to maintain Rust's aliasing invariants. The only affected API is the `get` function which, by design, is only usable with `Copy` types.
+
 # Detailed design
 [design]: #detailed-design
 


alexcrichton:
Oddly enough @SimonSapin, @nox, and I were *just* talking about this, and we ended up concluding that we probably wouldn't want to extend the `Cell` type itself primarily for consistency with the other trait impls on `Cell`. For example the `PartialEq` and other assorted traits require `T: Copy` whereas this means that not *all* `Cell` types would be equatable/comparable.

I'd also personally err on the side of being conservative and drop `take` in favor of `replace`, but other than that this all seems great to me!

text/0000-movecell.md
```diff
@@ -0,0 +1,54 @@
...
+
+# Summary
+[summary]: #summary
+
+Extend `Cell` to work with non-`Copy` types.
+
+# Motivation
+[motivation]: #motivation
+
+It allows safe inner-mutability of non-`Copy` types without the overhead of `RefCell`'s reference counting.
```
sfackler: Which types? What is a concrete use case for this?

durka:
@alexcrichton 

> For example the PartialEq and other assorted traits require `T: Copy`

Can you explain this remark? As far as I can tell it's not true for PartialEq, so I want to know what assorted traits you refer to :)

alexcrichton:
@durka ah yeah sure! So right now the standard library has:

```rust
impl<T: PartialEq + Copy> PartialEq for Cell<T> { ... }
```

If, however, `Cell` can accept an arbitrary `T` it'll be expected for that to look like:

```rust
impl<T: PartialEq> PartialEq for Cell<T> { ... }
```

Unfortunately, though, we can't implement that. This means that we'd probably have to have `PartialEq` only defined for `Copy` types, which seems inconsistent enough that I'd prefer to have a separate cell type.

durka:
Oh I see. I thought you were saying that it was `trait PartialEq<Rhs>:
Copy`. I agree it'd be kinda weird if not all Cells were comparable.

text/0000-movecell.md
eefriedman: https://github.com/SimonSapin/kuchiki/blob/23e92d45b7406865e035defdc5f5ca9f1809c406/src/tree.rs#L116 is the best example I could find with a quick search.

nox:
I actually formalised `MoveCell` in Mitochondria. You can find another Cell idea in it (`OnceCell`) for which I should probably write a RFC myself.

I have various use cases for both in Servo.

https://github.com/nox/mitochondria

text/0000-movecell.md
```diff
@@ -0,0 +1,54 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+```rust
+impl<T> Cell<T> {
+    fn set(&mut self, val: T);
+    fn replace(&mut self, val: T) -> T;
```
glaebhoerl: These should all be `&self` right?

glaebhoerl:
Along with

    fn replace(&self, other: T) -> T

we could also have

    fn swap(&self, other: &mut T)
    fn swap_cells(&self, other: &Cell<T>)

(the first two are symmetric with `std::mem`)

cc #1106

ubsan:
@alexcrichton You should be able to implement `PartialEq`, at least, unless I'm not thinking right. It takes an `&T` parameter.

```rust
fn eq (&self, other: &Self) {
  let self_ptr = self.as_unsafe_cell().get();
  let other_ptr = other.as_unsafe_cell().get();
  unsafe { &*self_ptr == &*other_ptr }
}
```

nox:
That is unsound because you call PartialEq from inside the UnsafeCell.

-- 
Anthony Ramine

text/0000-movecell.md
Amanieu: Oops

text/0000-movecell.md
```diff
@@ -0,0 +1,54 @@
...
+
+```rust
+impl<T> Cell<T> {
+    fn set(&self, val: T);
+    fn replace(&self, val: T) -> T;
+    fn into_inner(self) -> T;
+}
+
+impl<T: Copy> Cell<T> {
+    fn get(&self);
```
kennytm: `-> T`

SimonSapin:
I think this needs to be in the detailed design: 

`Cell::set` is currently:

```rust
    pub fn set(&self, value: T) {
        unsafe {
            *self.value.get() = value;
        }
    }
```

But for `T: Drop` this can be unsound:

```rust
    struct Evil(Box<u32>, Rc<MoveCell<Option<Evil>>>);
    impl Drop for Evil {
        fn drop(&mut self) {
            mem::drop(self.1.take());  // Mess with the "other" node, which might be `self`.
            self.0.clone();  // use after free!
        }
    }
    let a = Rc::new(MoveCell::new(None));
    a.replace(Some(Evil(Box::new(5), a.clone())));  // Make a reference cycle.
    a.set(None);  // Trigger Evil::drop while in the cell
```

To avoid this, the method needs to be rewritten so that `Drop` is not called "from inside" the cell:

```rust
    pub fn set(&self, value: T) {
        self.value.replace(value);  // semi-colon to drop the return value of replace
    }
```

nikomatsakis:
@ubsan indeed, that is not sound, as the whole point of `Cell` is that you never get a reference to the interior. (Hence the name `Cell` vs `RefCell`.)

nikomatsakis:
I'm in favor of having something like `Cell` where I can swap the contents -- but I guess having a separate type (to resolve the `PartialEq` discrepancy) might also be OK (e.g., `MoveCell`) due to the `PartialEq` discrepancy. Regardless it galls me to use `RefCell` in this case and pay the price for flags and unnecessary runtime checks.

(I'd also like `RefCell` to have a `take` option... which in many cases might obviate the need for me to use `Cell`, since I suspect I would often write `Cell<Option<Vec<...>>>` so I can swap with `None`, and in that case I'd probably rather use `RefCell<Vec<...>>` if it had a `take()` method -- since it already has the flags.)

alexcrichton:
@rfcbot fcp close

My personal preference would be to not enhance `Cell` with these methods, and if we did anything to go in the direction of https://github.com/rust-lang/rfcs/pull/1659. 

The second drawback mentioned is a showstopper for me. In general if you're working with a `Cell<T>` type you might expect common traits like `Debug`, `PartialEq`, etc to work by default. When you run into *that* one `Cell<T>` that works differently, however, I feel like it's too big of a surprise factor to be worth it for a niche method like this.

I'm going to propose that the libs team closes this RFC as a result of that drawback, but I'm curious to hear others' thoughts!

nikomatsakis:
Hmm. I am sad, because I think that having a plethora of `Cell` types is kind of confusing; but the `PartialEq` impl does seem to sort of box us in here. 👍

glaebhoerl:
I don't see why the `PartialEq` thing is such a big deal at all. The `PartialEq` impl stays the same as it always was, and independently of it, some new capabilities get added.

There's no principle saying that every method/impl of a type should require the same trait bounds. I don't even understand why you would expect that.

alexcrichton:
For me at least it's all about the surprise factor of using `Cell`. It's documented and intended to be primarily used as `Copy` types, so then it can be very surprising to see it with a non-copy type or where functionality no longer works.

nikomatsakis:
@glaebhoerl 

> I don't see why the PartialEq thing is such a big deal at all. The PartialEq impl stays the same as it always was, and independently of it, some new capabilities get added.....There's no principle saying that every method/impl of a type should require the same trait bounds. I don't even understand why you would expect that.

I agree with you on the general principle; but of course one has to evaluate each case on its merits. In this case particular case I feel conflicted. You're making me rethink my opinion a bit. Let me just throw out some various chaotic thoughts.

### Combining sharing and mutability is still a weak spot

I feel like in general we need to smooth out our story on sharing + mutability. This begins with education: explaining better *when* it is legal to mutate (the naming of `&mut T` strongly suggests to people that it is the only way to mutate, reasonably enough, and I think our messaging on sharing + mutability needs to get more crisp). 

But also just looking at the APIs: The cell vs ref-cell split is logical, but not ideal. I think it's not well-explained, for one thing, but also it is often stricter than normal. Many uses of `RefCell` that I find myself doing in practice would be perfectly safe as a `Cell` that called `clone`. For example, `RefCell<Rc<T>>`, which is known not to invoke any methods of `T`; similarly I often want ref-cells with more complex types where I do a lot of swapping (as well as things like a `RefCell<&mut Vec<T>>` where I just push to accumulate, and then later recover mutable access).

**Moreover, and most importantly I think, `RefCell` is really hard to use right.** It's all too easy to hold the lock longer than you mean to, and tracking down a "double borrow" is more annoying than I would like (that's something else we could try to address, at least in debug builds).

I'd prefer if we had a widely usable alternative that avoided these pitfalls, so that we could promote it as the preferred way to have sharing and mutability (something like "avoid it if you can, but if not, use `Cell`").

Note that there is nothing stopping me from prototyping these ideas in a crates.io crate except for laziness. =)

### What I would prefer: an expanded Cell 

I would prefer if we could expand on `Cell` so that it can do a wide range of operations on a wide range of types. Ideally we would maintain the invariant that none of these operations will lead to a borrow error or other **cell-related** panic at runtime, but it might be better to loosen that to being "extremely unlikely" to lead to a borrow error.

Since we want to eliminate cell-related panics, we can't permit open-ended borrows like `foo.borrow()`. That requires a ref-cell still. But we can permit a bunch of operations:

    - `get` -- clones the value (requires `T: Clone`, but see below)
    - `set` -- overwrites the value (works for all `T`)
    - `swap` -- swaps the value (works for all `T`)
    - `take` for `Cell<Option<T>>`, which swaps with `None`
    - `push` for `Cell<Vec<T>>`, which just pushes an element
    - `pop` for `Cell<Vec<T>>`, which just pops an element

The last few might seem surprising. The idea here is that we know that `push()` on `Vec<T>` doesn't invoke any user-defined code, and that it doesn't mutate any cells, so this should be ok. (I think?) We can basically grow the set of operations here over time.

Note that I said `get` could work any `T: Clone`. Obviously in the general case this would require a safety flag and could panic -- but it's very unlikely to happen in practice unless you have some wacky clone impls. I think I'd prefer to support it for all `T: Clone` and use specialization to optimize away the flag, but I could see it also being nice to never have a flag and only support `get()` for `T: Copy` or types (notably `Rc<T>` and `Arc<T>`) that can be cloned without risking mutating the cell (this would be an unsafe trait).

### Why partial eq bothers me

OK, without that context, why do I care about `PartialEq`? I'm not sure if I should, but I guess it just feels like a surprising case of the API changing (more so than other things). It's because the reason it can't work for arbitrary T is tied to this sort of subtle fact that `PartialEq` might do random things and bad code could cause problems (but in practice such problems are very unlikely). 

### Random middle ground

I wonder if it'd be a good first step to just extend `Cell` to copy types *or* those that implement an unsafe `CellSafeClone` trait (notably `Rc` and `Arc`). This would basically be things that promise that their `clone` method cannot affect any cell that contains `self`.

### An irrelevant, whiny aside: why I wish cell did not have `PartialEq`

In retrospect, my preference would be that `Cell<T>` does **not** implement `PartialEq` under any circumstances. `Cell` is the marker that converts types from being "a value" into something with their own identity, and that is precisely where the semantics of `PartialEq` get murky. I'd rather we had *no* default, so that you can write your own wrappers. If we **were** to have behavior, I would have chosen pointer identity. 

My analogy is to things like Python:

```python
class Counter:
    def __init__(value):
        self.value = 0

x = Counter(0)
y = Counter(0)
x == x // True
x == y // False
x.value += 1
x == x // still True
x == y // still False
```

I consider this Rust code to be a direct translation, but of course it behaves differently:

```rust
let x = Rc::new(Cell::new(0));
let y = Rc::new(Cell::new(0));
x == x // true
x == y // true!
x.set(x.get() + 1);
x == x // still true, of course
x == y // now false, hmm.
```

This is because `==` in Python *defaults* to identity. This is a safer default when you have sharing+mutability, since it means that if `x == y` is true, it remains true no matter what you do. 

OTOH, there are times when you want a value-like semantics. Usually if you are "freezing" a value. But in rust those are much rarer, I think, because you would usually use ownership for such scenarios and not require `Cell`. So I'd rather that people "opt-in" to that by wrapping the `Cell` with their own type.

But ... yeah, water under the bridge.

glaebhoerl:
> What I would prefer: an expanded Cell

I feel like there's a conceptual difference here between [`replace`, `swap`, and `swap_with`](https://github.com/rust-lang/rfcs/pull/1651#issuecomment-226338744), which work with *any* type and rely on nothing besides Rust's own semantics for their safety, and the others, which are more ad hoc extensions to support specific operations on specific `std` library types whose implementations we happen to control. In other words, `replace` and friends feel to me like basic primitives which have been left out accidentally. Another qualitative difference is that `replace` & co. are "memcpy-only" operations - which they have in common with `Cell`'s existing operations - while the others are not.

(Note that I'm not arguing against potentially adding the others as well at some point. Just against lumping them in with `replace`.)

-----

> It's because the reason it can't work for arbitrary T is tied to this sort of subtle fact that `PartialEq` might do random things and bad code could cause problems (but in practice such problems are very unlikely).

I feel like the rationale here is much more straightforward than that:

"Hey how come there isn't a `PartialEq` on `Cell` for non-`Copy` types?"

"Try implementing it and find out."

"Oh, I see. It seems like you can't express it using the API."

----

> In retrospect, my preference would be that Cell<T> does not implement PartialEq under any circumstances. Cell is the marker that converts types from being "a value" into something with their own identity, and that is precisely where the semantics of PartialEq get murky. I'd rather we had no default, so that you can write your own wrappers. If we were to have behavior, I would have chosen pointer identity.

I agree with this completely and the same thought has occurred to me as well. (Another reference point is Haskell where `instance Eq (IORef a)` compares by address - *obviously*, because the contents of the reference cannot even be accessed from pure code. And I haven't checked but I would be almost shocked if ML's `ref`s didn't have the same behavior as well.)

nikomatsakis:
> The current behavior is very convenient for that use.

Indeed. I'm not saying the current behavior isn't convenient sometimes. And anyway it's not like we can change it.

aturon:
@rfcbot concern PartialEq

I'm not convinced by the argument about `PartialEq` -- it's not *so* different from cases like `Vec<T>` being `Clone` only when `T` is. It's also a bit confusing to mention in the context of an RFC that's all about API surface that varies depending on the specifics of a type parameter. I'm skeptical that this will actually be a problem in practice, especially for people knowledgeable enough to be using `Cell` in the first place.

@alexcrichton 

> For me at least it's all about the surprise factor of using Cell. It's documented and intended to be primarily used as Copy types, so then it can be very surprising to see it with a non-copy type or where functionality no longer works. 

I think *this* is really the core of the negative reaction many have had: we have, for a long time, though of and messaged `Cell` as "interior mutability for `Copy` types", and this RFC moves away from that.

I think it's important to step back a bit, and try to ignore what the current mental model of `Cell` is, and consider whether we can give a coherent *new* mental model that covers both `Cell` and `RefCell`. A promising such model is "`RefCell` is for when you need to get references into the object; otherwise, use `Cell`". Conveniently, this matches the names :-)

(I don't have time to go into more detail than that, but wanted to at least voice these concerns.)

SimonSapin:
@aturon The relationship between `Cell<T>: PartialEq` and `T: Copy` is not as obvious as the one between `Vec<T>: Clone` and `T: Clone`, but yes, we *can* do it.

I agree that this is about messaging (which includes docs). At first I didn’t want to be the one to push for that change in messaging, but if you and other people think it’s worth doing then I’d like to see it happen.

nikomatsakis:
Ah, I'm glad @aturon commented. I've been meaning to respond to @glaebhoerl's comment:

> I feel like the rationale here is much more straightforward than that:

Basically I tried to find a persuasive counterargument and kind of... failed. I think I agree now. =) That is, it's ... unfortunate, but not that big a deal ultimately. 

Regarding @aturon's philosophy of when to use what, "use `RefCell` when you want a ref" is what I always thought the philosophy was. I guess that frequently aligns with "use `Cell` for copy types" but I agree that's just sort of an accident of the current API. 

I was thinking it'd be nice to extend `Cell` (using an unsafe trait for "side-effect-free clone", perhaps, instead of `Copy`) so that it works for some clone impls, notably `Cell<Rc<T>>` and `Cell<Arc<T>>`. This would fit with the philosophy.

alexcrichton:
@aturon 

I agree that it's a good exercise to step back and consider what would happen if we drop our prior assumption that `Cell` is for `Copy` types. In doing this, though, there are still a few parts of the API of `Cell` which would be surprising to me, all of which boil down to the trait implementations.

Right now the important trait impls on `Cell` are `PartialEq` through `Ord` and `Clone`. All of these crucially require a reference, yet they are implemented on a type that would conceptually be a type for "you can't get a reference to this". This only works because the trait impls require `Copy`, obviously, and with @nikomatsakis's "side-effect-free clone" trait (or something like it) we could extend for "types that are safe to refer to sometimes, but not always". Even in this world, though, we're left with a type that you use when you don't need a `ref`, but only sometimes. I feel like those impls really are engineered to an API which requires `Copy` (as we've assumed in the past).

Now there's also an argument that having a "principled" type is just a hindrance. This line of reasoning is help up with "well it's nice to have `Cell` work purely over `Copy` types", but sometimes I just want *my* cell to work with this obviously-implementable functionality on `Cell`. Put another way, we could bolster `std::cell::Cell` with as many bells and whistles we can think of to serve everyone's needs. There may be confusion, but perhaps no more confusion than we already have today with reaching for `Cell` in the first place.

I think I personally still feel like this enhancement to `Cell` isn't worth it, but I'm not as convinced as I was earlier.

glaebhoerl:
For what it's worth, my own mental model is that "`Cell` is for `Copy` types" was the wrong model all along, and that the 'right' model is "`Cell` is for any safe operations that can be expressed as just `memcpy`s" (or, from a higher-level view, as "moves" -- with the definition of `Copy` being that the source is still available even after a "move"). So by this logic it's always felt "obvious" to me that alongside the existing `Copy` methods, these methods also belong. (And why I'm less sure about other extensions like supporting `Rc` or `Arc`; while clearly useful, there's no clear principle to determine where such extensions should end, as far as I can see.) I can see how others might have different mental models though.

And then I think the right intuition is that you need `Copy` for any operation which *non-destructively observes* the contents of the `Cell`, which explains why e.g. `PartialEq` requires it.

eddyb:
My intuition, since we seem to be sharing, is that `Cell` is an opaque inline container with no space overhead and a "duplicates, not references" API membrane.

Therefore, a Java `ArrayList<T>` or a JS array could be seen (single threaded and ignoring GC) as `Cell<Vec<Rc<T>>>` - there is no reason Rust shouldn't be able to express such patterns, with most common operations being safe.

There's even a way to prove that arbitrary functions won't result in two references into the `Cell`, conservatively, via the callgraph.
At least in cases like the `Vec` above, there are no indirect calls, not even a dependence on generics, i.e. the callgraph is statically inspectable, with neither `Vec` methods nor `Rc` cloning touching the same `Cell`.

Here's the gotcha though: the *same* `Cell`, not *a* `Cell`, since `Rc` uses `Cell` for its refcounts. It could be type-based (i.e. a form of TBAA), which would require strict rules around manipulating references to `Cell` types, otherwise they could bypass the safety checks in question.

Even if it can work, peering at bodies of functions, possibly from other crates, may be frowned upon, so for now it's just an idea I've mentioned a few times.
If we *do* want to do this, it can also apply to `#[thread_local] static mut`, making it as ergonomic as its C counterpart, and possibly defensive reentrant programming (signal/interrupt handlers).

nikomatsakis:
I'm enjoying the share-fest!

@glaebhoerl 

> And why I'm less sure about other extensions like supporting Rc or Arc; while clearly useful, there's no clear principle to determine where such extensions should end, as far as I can see.)

So certainly @eddyb and I share an "intution" that `Cell` is basically all about a "duplicates, not references" membrane. My feeling is that this notion is a good one to build upon -- unlike `RefCell`, it's semantics are fairly clear, and it doesn't inject panics. This is not to say that code which uses `Cell` won't be buggy, but it is sort of easier to debug and imagine how it's happening. It's sort of similar to how sequentially consistent code can still have race conditions, but they are easier to understand. :)

You're right though that if you limit yourself to `Copy` types, you have the nice property that `Cell<T>::get()` always executes a known set of instructions that is independent from `T` (except for the size). If we use an unsafe trait, then the set of instructions becomes arbitrary -- I think why `Rc<U>` and `Arc<U>` feel like natural extensions to me, though, is that the set of instructions is independent from `U`. (In contrast to something like `Vec<V>`.) But once you allow the instructions to be dependent on `T`, it's not clear why we should stop at any particular number of "degrees".

I am wary of @eddyb's proposal that we can use a sort of ad-hoc conservative static analysis to decide what types `T` is compatible with, but of course what I'm proposing (an `unsafe` trait) is sort of no different, except that we've outsourced the analysis to unsafe code authors. Still, these ultimately feel like orthogonal concerns (that is, we could have an `unsafe` trait, but eventually have some mechanism to make safe impls, based on said analysis).

In any case, I suppose that this is mildly separate from the current proposal. That is, things like `swap` preserve the original invariant in that the code they execute for `Cell<T>` is independent of `T`.

SimonSapin:
> if you limit yourself to Copy types, you have the nice property that Cell<T>::get() always executes a known set of instructions that is independent from T (except for the size). If we use an unsafe trait, then the set of instructions becomes arbitrary -- I think why Rc<U> and Arc<U> feel like natural extensions to me, though, is that the set of instructions is independent from U.

If we add some way to clone e.g. `Arc<T>` out of `Cell<Arc<T>>`, I think it should definitely be an API separate from `Cell::get` which should still require `T: Copy` and be guaranteed to only do a (comparatively cheap) `memcpy`.

brson:
Although there's a lot of good discussion here, we are heading steadily toward an FCP to close. Does that still seem right? From what I see it seems like there's a whole lot of debate left on this matter, so I think close still makes sense.

@aturon waiting on your opinion to move the FCP forward.

aturon:
@brson What's under debate right now is a mixture of motivation and mental model, rather than the details of the RFC. I don't think we need to close/restart it -- we just need to reach a consensus about the tradeoffs here.

@alexcrichton, I wonder what you think of the mental model that `Cell` is for data that can only be accessed via moves in or out, which in the special case of `Copy` can leave the existing data behind? In that case, you get some special APIs like `PartialEq` and `get` precisely when moving doesn't destroy the underlying data.

I'll note that one of the biggest problems with this RFC is the very thin stated motivation -- it's much better to spell out practical examples where this really matters. I think one of the simplest but most common ones is the `Option<T>` case, where I want to optionally store a value that I will always work with by `take`ing the value out -- a typical "option dance". Here, `Cell` would give a more ergonomic and less redundant interface than `RefCell` does today. On the other hand, though, there are some cases where you want `Option<T>` and interior mutability, but you *do* want to peak into the data without `take`ing it -- and then you'd be forced into `RefCell`. Ideally, it'd be easy to decide up front which case you rely on, by asking yourself "Do I ever need to access this data without moving it out?"

On the whole, I'm in favor of this RFC; I found @glaebhoerl's mental model particularly persuasive.

Finally, a small note regarding stability: landing this change would require relaxing the bounds on `new` and `set`, which effectively mean the change is "insta-stable".

SimonSapin:
> On the whole, I'm in favor of this RFC; I found @glaebhoerl's mental model particularly persuasive.

At first I preferred a separate  `MoveCell` type, but I was convinced by this proposed shift in mental model and I’m now in favor of this RFC as well.

----

> I'll note that one of the biggest problems with this RFC is the very thin stated motivation

I can provide an example of how I’d use this: [Kuchiki](https://crates.io/crates/kuchiki) currently uses a custom cell based on `UnsafeCell`. With this RFC implemented, it could use `Cell` like so:

```rust
struct Node {
    first_child: Cell<Option<Rc<Node>>>,
    next_sibling: Cell<Option<Rc<Node>>>,

    parent: Cell<Option<Weak<Node>>>,
    last_child: Cell<Option<Weak<Node>>>,
    previous_sibling: Cell<Option<Weak<Node>>>,

    data: NodeData,
}

enum NodeData {
    // …
}
```

(My initial design for this had each node entirely in a single refcell: `Rc<RefCell<Node>>`, but juggling mutable and shared refcell borrows to do *anything* turned out to be a pain. I’ve found it more convenient to have more, smaller (ref)cells. But the space overhead of `RefCell` adds up if there’s five of them in each node just for maintaining the tree structure.)

This uses `Option` as @aturon mentioned, so some of the code does the option dance with `if let Some(rc) = some_cell.take()` or `if let Some(rc) = some_cell.replace(None)`. In some cases I’d use:

```rust
fn clone_cell_rc<T>(cell: &Cell<Rc<T>>) -> Rc<T> { unsafe { (*cell.as_ptr()).clone() } }
fn clone_cell_weak<T>(cell: &Cell<Weak<T>>) -> Weak<T> { unsafe { (*cell.as_ptr()).clone() } }
```

… which as discussed before is safe because `Rc<T>::clone` and `Weak<T>::clone` are known not to touch `T` (only increment refcounts).

It might be nice to have this cloning functionality in `std`, but since `Cell::as_ptr` is public and now stable it doesn’t need to be there. But including it would raise the question of what else needs to be included, or how to generalize it. Which is tricky on its own, so I’d rather *not* include it in this RFC (and leave it for a possible future RFC to reconsider).

aturon:
Thanks @SimonSapin, that's a helpful example, and I feel exactly the same way about the `Rc` interaction. I think it's great to keep pushing on the ergonomics of `(Ref)Cell` + `Rc`/`Arc`, since you can use that to recover a Java-like programming model, but we should take it one step at a time.

Manishearth:
While I'm generally very much in favor of this RfC, I would like to note that this has a diagnostics impact. Currently using `Cell<T>` with a non-Copy type gives an error (which could be made to recommend `RefCell`). Now, the error will be deferred till you try to use `.get()`, which is a more annoying user experience. The whole learning curve around figuring out cell/refcell isn't great -- part of why I wrote the wrapper types blog post / doc page -- and this makes it worse. Not worse enough for me to say I don't want this RfC to happen, but it's something we should be aware of and try to remedy (through clippy? through the compiler? something.)

alexcrichton:
> @alexcrichton, I wonder what you think of the mental model that Cell is for data that can only be accessed via moves in or out, which in the special case of Copy can leave the existing data behind? 

Yeah I certainly understand this, but I feel like it's much easier to explain, however, that `Cell` is for copy types rather than "well `Cell` is for all types, but it's even more special for `Copy` types". That sort of special-casing I don't believe exists elsewhere in the standard library, and can provide interesting speed bumps for picking these up.

I think @Manishearth [put it pretty well](https://github.com/rust-lang/rfcs/pull/1651#issuecomment-259759965) basically.

brson:
> @brson What's under debate right now is a mixture of motivation and mental model, rather than the details of the RFC. I don't think we need to close/restart it -- we just need to reach a consensus about the tradeoffs here.

There's is an FCP proposal to _close_ this RFC right now that nearly everyone else has signed off on. If you see this RFC as going forward maybe something needs to happen, because right now, process-wise, this RFC is on the verge of being closed.

SimonSapin:
@brson As I see it, new arguments have come up since this was proposed. In my opinion, enough that people who signed off should be asked to reconsider now.

Kimundi:
I think I've been persuade by the neweset rounds of discussion that a extension of `Cell` for non-Copy types would be worthwhile after all. `Cell` can be a thing that is about memcopies instead of references, and not just with `Copy`. ;)

About documentation ergonomics: Retargetting from "for Copy types" to "for cases where inner references are not needed" is indeed a bit tricky, but maybe not that important - after all the old wisdom is still true, it just becomes incomplete.

---

In not entirely unrelated news, I've opened #1789 for an orthogonal feature around `Cell` that might synergize well with this proposal.

brson:
It seems like @aturon is still in favor of pushing on this. I think the RFC needs a lot better motivation and usage examples showing what patterns this is going to affect. Right now it's hard to square the tiny RFC, the overwhelming sentiment to close, the ongoing discussion, and the renewed enthusiasm to do this. The RFC's weak motivation hasn't changed. Why is this good?

alexcrichton:
@rfcbot fcp cancel

Ok, enough discussion has happened that I don't think the arguments are as strong for closing now as they were when I initially proposed to close. I believe @aturon will follow-up soon with a request to in fact merge

nikomatsakis:
@brson the motivation doesn't seem weak to me. `Cell` feels like a primitive building block and I think it makes sense to make it as powerful as it can be without stretching the mental model. The mental model of "Cell supports moving bits around, but never gives a reference to the inside" seems pretty coherent.

Adding the ability to swap data around is a pretty powerful extension to `Cell`. The alternatives seem strictly worse to me:

1. Requiring people to use `RefCell` for this adds both space and runtime overhead (albeit minimal) and also introduces footguns (`RefCell` is sort of panic-prone).
2. `MoveCell` means we now have to pick between three types; working with cells is complex enough in my opinion.
    - It's a bit of a stretch, but I also think it's handy to have one type because sometimes you may have generic code that requires only the ability to swap, but extended functions where you can use `get()` because you know that `T: Copy` in that area. In cases like that, having just one type is advantageous.

Kimundi:
@glaebhoerl Idle observation: If #1789 should get accepted then you'd only need `swap(&self, &Cell<T>)` to swap with both `&Cell<T>` and `&mut T`. Though it would be somewhat unergonomic unless we start working with conversion traits:

```rust
let mut tmp = ...;
cell.swap(Cell::from_mut(&mut tmp));
```

glaebhoerl:
Yep, good point.

antrik:
Regarding mental model / documentation ergonomics, I believe this will actually *improve* the situation: I was recently thinking about how to best explain `Cell<>` and related types; and I was really struggling with explaining why `Cell<>` is only for `Copy` types... Because really I could not think of a fundamental reason.

I'm surprised that some people are so vehemently contesting the usefulness of this. (On the now-closed `MoveCell<>` RFC even more than here...) To me, the use cases seem pretty much the same as for the "regular" `Cell<>` type; and questioning the usefulness of `Cell<>` on move types doesn't feel much different than questioning the usefulness of `Cell<>` altogether... I believe it's a basic language building block, that is likely to come up in pretty much any sufficiently complex program -- relegating this to some obscure external library, or to another type that is both less efficient and less ergonomic (run-time vs. compile-time checking) in cases where it's not really needed, affects the efficiency and ergonomics of the language as a whole. It kinda breaks Rust's promise that you never have to pay for abstractions you are not using...

The specific use case that brought me here is for Unix file descriptors, which I want to wrap in a non-copy type. (While *technically* they are just a `c_int`, *semantically* they are actually references to kernel data structures -- and need to be treated just like other references regarding aliasing/cloning/dropping, to avoid undue overhead, potential leaking, or use-after-free...) However, sometimes we need to put file descriptors into a `Cell<>` for interior mutability. (In the specific case of `ipc-channel`, because of the way the `Serde` API works.)

This is an interesting example I think, because it demonstrates that the difference between a low-cost abstraction and a zero-cost abstraction can be fundamental: if we could continue using `Cell<>`, the non-copy wrapper would be a no-brainer, as it enables better employing the type system for more robust and ergonomic code at no cost. If it requires a switch to `RefCell<>` on the other hand, it suddenly becomes an unfortunate tradeoff between avoiding overhead or improving robustness...

aturon:
Apologies for the long delay moving forward on this!

Given the thrust of the overall discussion here, it seems both that there's a [pretty decent motivation](https://github.com/rust-lang/rfcs/pull/1651#issuecomment-259620591), and a [reasonable mental model](https://github.com/rust-lang/rfcs/pull/1651#issuecomment-257736645).

@Amanieu, can you update the RFC to try to incorporate some of the discussion, in particular around the motivation and mental model issues? (No changes needed to the actual design.)

Meanwhile, I'll kick off the process for heading toward FCP.

@rfcbot fcp merge

pnkfelix:
@rfcbot reviewed

nrc:
ping @BurntSushi, @brson, @sfackler, @withoutboats please review!

solson:
ping @withoutboats, only review left

aturon:
The FCP period has elapsed without any further discussion, so the RFC has now been merged! Thanks @Amanieu!

[Tracking issue](https://github.com/rust-lang/rust/issues/39264)

RalfJung:
Not sure if this is the right place to ask, but I have a question concerning this RFC: Why is `get_mut` still restricted to `Copy` types? As far as I can see, it would be perfectly fine to allow `get_mut` for any `T`. (Actually, I have [proven that this is the case](https://gitlab.mpi-sws.org/FP/LambdaRust-coq/blob/9aeb8f2b89a708a6517bbc399876cff136190a41/theories/typing/unsafe/cell.v#L107). ;)

Manishearth:
Yeah, it seems sound to do that, it's probably still restricted because nobody thought of it. Might be worth opening up a quick rfc for this, or perhaps just bringing it up on the tracking issue.

alexcrichton:
@RalfJung sounds good to me! (there's probably a few more methods that can be generalized as well)

alexcrichton:
Also I don't think such an RFC would require an RFC any more

aturon:
Agreed, a PR is fine.

RalfJung:
All right, will prepare one.

@alexcrichton well, the other two are taking a shared borrow of the Cell, so they are way more complicated to talk about. On the other hand, they are returning `&UnsafeCell<T>` and `*mut T` respectively, so they don't make any promises anyway -- they are trivially safe to call on non-Copy types, but what you can do with the resulting pointer is a whole different story.

SimonSapin:
I think that only `get` really requires `Copy`. Everything else can be generalized. (Though I don’t care so much for `as_unsafe_cell` since it’s deprecated.)

F001:
I did a comparison of the methods of `Cell` and `RefCell`, and the current situation confuses me. The differences are:

1. `Cell::as_ptr` requires `Sized`, `RefCell::as_ptr` does not.
2. `Cell` implements `Clone` when `T: Copy`, `RefCell` implements `Clone` when `T: Clone`.
3. `RefCell` has these methods `borrow` `try_borrow` `borrow_mut` `try_borrow_mut`.
4. `Cell` has these methods: `Cell::get` when `T: Copy`, `Cell::take` when `T: Default`.

The main questions on top of my mind are: 
1. **is it still necessary to split `std::cell` to two flavors** ? 
2. **why not merge these two types into one** ?

It is totally fine to me if `Cell` also has `borrow_*` methods, and `RefCell` has a `take` method if `T: Default`, and of course `RefCell` has a `get` method if `T: Copy`.

As niko stated above, the mental model of "Cell supports moving bits around, but never gives a reference to the inside" seems pretty coherent, before this RFC is merged. But after this RFC is merged, the distinction between these two types are very vague, and the motivation of this separation seems not strong enough.

To make cell types as easy as it can be, I'm proposing to merge the methods together and make these two types exactly identical.

Manishearth:
@F001 yes, it is necessary to have two, the inner implementation differs and the guarantees provided are different.

SimonSapin:
@F001 To expand a bit on Manish’s explanation: an important invariant of safe Rust is that mutation cannot happen while there multiple ways (aliasing) to access something. This leads to the separation of `&T` and `&mut T`. Cells (of various kinds) allow mutation through `&T`, so they have to be careful in their safe public API to maintain the invariant. `Cell` and `RefCell` are two ways to achieve that, with different tradeoff.

* `Cell` is less flexible: it never provides to arbitrary code (which might include a `Clone for T` impl) references to the inside of the cell. You can only copy or swap entire `T` values. See https://github.com/kuchiki-rs/kuchiki/blob/v0.6.0/src/move_cell.rs#L99-L127 for an example of why `Clone for Cell<T> where T: Clone` would be bad.
* `RefCell` can provide inner references, but need to count them to make sure that a `RefMut` is the only reference as long as it exists. So it has space overhead (an extra `usize`) to store that counter, and time overhead (for these counter increments and decrements).

This tradeoff is why they both exist. And the different constraints explain the different API. Except for `as_ptr`, I think. Maybe this is just an oversight and `Cell::as_ptr` can be extended to `T: ?Sized`. Wanna file a PR for that one?

F001:
@SimonSapin I see. `set` `take` should not coexist with `borrow_*`. If the refcell is borrowed, and `take` is called in this period, it can only panic. The `Clone for Cell<T> where T: Clone` is much trickier.

Thank you all for clarification.

SimonSapin:
`set` and `take` could be added to `RefCell`, but they’d not be super useful since they’re equivalent to `*refcell.borrow_mut() = new_value` and `refcell.borrow_mut().take()` respectively. `borrow_mut` already panics when the refcell is already borrowed.

PR #1653, opened 2016-06-17
propose assert_ne

ashleygwilliams:
this is an rfc to add a `assert_ne` macro to compliment `assert_eq`. it exists as a crate: https://crates.io/crates/assert_ne.

### DIFF
--- a/text/0000-assert_ne.md
+++ b/text/0000-assert_ne.md
@@ -29,7 +29,7 @@ macro_rules! assert_ne {
         match (&$left, &$right) {
             (left_val, right_val) => {
                 if *left_val == *right_val {
-                    panic!("assertion failed: `(left !== right)` \
+                    panic!("assertion failed: `(left != right)` \
                            (left: `{:?}`, right: `{:?}`)", left_val, right_val)
                 }
             }
--- a/text/0000-assert_ne.md
+++ b/text/0000-assert_ne.md
@@ -1,0 +1,57 @@
+- Feature Name: Assert Not Equals Macro (`assert_ne`)
+- Start Date: (2016-06-17)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+`assert_ne` is a macro that takes 2 arguments and panics if they are equal. It
+works and is implemented identically to `assert_eq` and serves as its compliment.
+
+# Motivation
+[motivation]: #motivation
+
+This feature, among other reasons, makes testing more readable and consistent as
+it compliments `asset_eq`. It gives the same style panic message as `assert_eq`,
+which eliminates the need to write it yourself.
+
+# Detailed design
+[design]: #detailed-design
+
+This feature has exactly the same design and implementation as `assert_eq`.
+
+Here is the definition:
+
+```rust
+macro_rules! assert_ne {
+    ($left:expr , $right:expr) => ({
+        match (&$left, &$right) {
+            (left_val, right_val) => {
+                if *left_val == *right_val {
+                    panic!("assertion failed: `(left !== right)` \
+                           (left: `{:?}`, right: `{:?}`)", left_val, right_val)
+                }
+            }
+        }
+    })
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Any addition to the standard library will need to be maintained forever, so it is
+worth weighing the maintenance cost of this over the value add. Given that it is so
+similar to `assert_eq`, I believe the weight of this drawback is low.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Alternatively, users implement this feature themselves, or use the crate `assert_ne`
+that I published.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None at this moment.
--- a/text/0000-assert_ne.md
+++ b/text/0000-assert_ne.md
@@ -7,13 +7,13 @@
 [summary]: #summary
 
 `assert_ne` is a macro that takes 2 arguments and panics if they are equal. It
-works and is implemented identically to `assert_eq` and serves as its compliment.
+works and is implemented identically to `assert_eq` and serves as its complement.
 
 # Motivation
 [motivation]: #motivation
 
 This feature, among other reasons, makes testing more readable and consistent as
-it compliments `asset_eq`. It gives the same style panic message as `assert_eq`,
+it complements `asset_eq`. It gives the same style panic message as `assert_eq`,
 which eliminates the need to write it yourself.
 
 # Detailed design
--- a/text/0000-assert_ne.md
+++ b/text/0000-assert_ne.md
@@ -8,6 +8,7 @@
 
 `assert_ne` is a macro that takes 2 arguments and panics if they are equal. It
 works and is implemented identically to `assert_eq` and serves as its complement.
+This proposal also includes a `debug_asset_ne`, matching `debug_assert_eq`.
 
 # Motivation
 [motivation]: #motivation
@@ -35,6 +36,14 @@ macro_rules! assert_ne {
             }
         }
     })
+}
+```
+
+This is complemented by a `debug_assert_ne` (similar to `debug_assert_eq`):
+
+```rust
+macro_rules! debug_assert_ne {
+    ($($arg:tt)*) => (if cfg!(debug_assertions) { assert_ne!($($arg)*); })
 }
 ```
 


KalitaAlexey:
@ashleygwilliams How often do you need assert_ne? I didn't want it once

ashleygwilliams:
i found that when writing tests i wanted it pretty often and on tweeting that i made the crate i got feedback that people were excited and wanted me to try to get it into the language. hence this rfc.

text/0000-assert_ne.md
```diff
@@ -0,0 +1,57 @@
+- Feature Name: Assert Not Equals Macro (`assert_ne`)
+- Start Date: (2016-06-17)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+`assert_ne` is a macro that takes 2 arguments and panics if they are equal. It
+works and is implemented identically to `assert_eq` and serves as its compliment.
```
apasel422: s/compliment/complement/

(And below.)

text/0000-assert_ne.md
```diff
@@ -0,0 +1,57 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This feature, among other reasons, makes testing more readable and consistent as
+it compliments `asset_eq`. It gives the same style panic message as `assert_eq`,
```
apasel422: s/asset_eq/assert_eq/

text/0000-assert_ne.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here is the definition:
+
+```rust
+macro_rules! assert_ne {
+    ($left:expr , $right:expr) => ({
+        match (&$left, &$right) {
+            (left_val, right_val) => {
+                if *left_val == *right_val {
+                    panic!("assertion failed: `(left !== right)` \
```
apasel422: Should this be `!=` instead of `!==`?

ashleygwilliams: yes it should. my javascript is showing 😳

bjz:
I personally have found it annoying sometimes fall back on `assert!(x != y)`, since it yeilds a much less helpful error message on failure. :+1:

jdm:
I have often come across cases where I would prefer to use assert_ne.

ereichert:
Yes, to this.  assert_ne! is easier to read, less error prone, and more ergonomic.  I usually wind up writing some form of this in a lot of my projects.

pnkfelix:
Bikeshed alert:

I have a slight preference for a more informative suffix than `_ne`. I would be happy with e.g. `assert_neq!`

  * (I would also be happy with `assert_not_eq!`, but that is a bit more difficult to type especially with the second underscore.)

I think that `assert_neq!` is more likely to be immediately comprehensible than `assert_ne!`, at least for newcomers.

----

I understand where `ne` comes from, and `ne` makes a lot of sense in the context of e.g. bash test conditions, or assembly programming (e.g. `repne` in x86, `bne` in mips and alpha, etc).

But in either of those cases, there tends to be an emphasis on both:
   1. brevity and
   2. regularity (thus encouraging all predicates have same length in characters).

In Rust we have certainly de-emphasized 1 in favor of familiarity (e.g. the shift from `next` to `continue`). I'm not sure how much emphasis we place on 2, but I don't think it trumps familiarity.

ashleygwilliams:
@pnkfelix totally hear this and mostly agree! i chose `assert_ne` based on the overwhelming preponderance of `ne` in the source (https://github.com/rust-lang/rust/search?utf8=%E2%9C%93&q=ne&type=Code).

would be totally open to changing it if others agreed.

bjz:
I agree with your sentiments, but also think about other things in the API, like `PartialEq::ne` - I would favor consistency in this case.

BurntSushi:
@bjz One rarely types or reads `ne` for `PartialEq` though, instead using `!=`. (Of course, this doesn't necessarily rebuke your consistency argument, but maybe dulls it slightly.)

oli-obk:
There are many other operations that would make sense like all the `Ord` ops etc. If one new macro is added, it would make sense to add the others, too. But there's also an alternative:

If `assert!` were rewritten as a syntax extension instead of as a macro as it is now, then it would be possible to automatically detect `assert!(x < y)`, `assert!(x == y)`, `assert!(x != y)`, ...

This has the major advantage of not needing textual identifiers, but using the actual operation and still producing a nice error message. It would probably be possible to improve the errors produced by these macros, since the syntax extension can produce better spans than macros.

dgrunwald:
I think making `assert!` a syntax extension that inspects the condition is the way to go.

Pytest does the same thing ([demo](http://pytest.org/latest/example/reportingdemo.html#tbreportdemo)) and I've found that extremely helpful.

mcarton:
`assert_eq!(x, y)` requires `x: Debug` and `y: Debug` while `assert!(x == y)` does not.
An `assert!` syntax extension could not require `x: Debug` and `y: Debug` for compatibility so it wouldn't do much.

dgrunwald:
A syntax extension could convert `assert!(x == y)` into a function call `assert_eq(&x, &y)`, and that function can then use specialization to act differently depending on the types of `x` and `y`.

bjz:
Given that we already have an `assert_eq!` macro, would it not make more sense to plug that gap in the interim before embarking on a more ambitious change to the semantics of the `assert!` macro?

wycats:
@ashleygwilliams Thanks so much for this. I have wanted this occasionally, and falling back to bare `assert` is annoying as you said. It's encouraging that your crate is already getting use, and that people on this thread have expressed demand for the feature.

I agree that a more ambitious refactor of `assert!` is a good thing to expore, but as @bjz said, given that we already have an `assert_eq!`, adding this functionality seems like a good medium-term approach to adding the necessary functionality, without adding too much avoidable cruft.

I'm for this :smile:

kennytm:
Why not include `assert_{gt|lt|ge|le}!` as well?

luser:
I rolled my own `assert_neq!` by copy-pasting `assert_eq` out of the standard library, so I would have liked to have this:
https://github.com/luser/sccache2/blob/fd5da79fe896fbdc39a711e257dd370644b0d526/src/test/utils.rs#L37
(I can't remember why I chose neq instead of ne.)

tbu-:
I don't think we need this in the standard library, this should be fine to live in a crate. You could use this: [`more-asserts`](https://github.com/thomcc/rust-more-asserts).

joshtriplett:
I specifically wished for this a few days ago when writing test code.

I'd love to see a smarter `assert!`, too, but I'd like to have `assert_ne!` available for consistency in the interim.

Evrey:
Like @kennytm I'd prever having all comparison operators represented by `assert` macros, i.e.:

`assert_eq!`, `assert_ne!`, `assert_ge!`, `assert_gt!`, assert_le!`, and `assert_lt!`, plus their debug siblings. So, in a way, I propose making the crate [`more_asserts`](http://www.thomcc.io/rust-more-asserts/more_asserts/index.html) standard stuff.

Having only `assert_eq!` and nothing else does not just feel lazy, but I, too, also use various comparing assertions besides equality testing in my tests. And having not just `assert!` but also `assert_eq!` is more than enough justification for me to import all the others.

ashleygwilliams:
i'm happy to write the rest of these as per @Evrey , i actually have a crate called `asserts` where i'm collecting this functionality. (looks like this was published the exact same day i wrote that crate and opened this rfc? by a mozilla employee no less- were we in the same room?)

in other news, i've updated my `assert_ne` crate to reflect the recent addition of custom messages to `assert_eq` https://github.com/rust-lang/rust/pull/33976/files

ashleygwilliams:
i would propose that using a crate is not ideal, if only because when updates like the one i just mentioned happen the behavior is out of sync. it'd be great to just be able to edit and improve them all at once.

lambda:
I have also found a need for `assert_ne!` in the past. The other comparison operators I don't think would come up as often, but would also be helpful.

Another that I ran into a need for recently is `assert_match!(err, SomeErrorVariant(_))`. I originally tried `assert_eq!`, but then I would need to implement `PartialEq` for my error type, and I couldn't do that conveniently with `#[derive(PartialEq)]` since one variant contained `std::io::Error`, so I'd have to do it by hand. What I wound up with was `assert!(match err { SomeErrorVariant(_) => true, _ => false })`, but that's kind of cumbersome and doesn't give good error reporting if the assertion fails. After implementing it myself, I discovered there already was an [assert_matches crate](https://crates.io/crates/assert_matches) that provides that macro.

I haven't encountered it in Rust yet, but in Python I've found it useful to have `assertIn`/`assertNotIn`, and `assertRaises` (which would be `assert_panic!()` in Rust), and `assertItemsEqual` (tests that two collections have the same elements even if in different order).

Of course, there are two reasons for wanting all of these different types of asserts; one is just code reuse, of commonly used predicates, but you could also just use write the predicates as operators, method calls, or function calls inside an `assert!` macro. The main reason for needing macros is to be able to also get at and print the value of the subexpressions, to make it easier to determine why a particular assertion failed.

As mentioned earlier in the thread, instead of introducing a number of special purpose assert macros, the existing `assert!` macro could be extended to print out subexpressions. In fact, there is already the [power-assert](https://crates.io/crates/power-assert) crate that does just this.

So, it would probably be a good idea to figure out the approach that we want to take; the `power-assert` approach, of just making `assert!` give better diagnostics, or the approach of having one `assert_<pred>!` per predicate, and with the latter approach, whether to integrate it into the standard library or leave it to external crates. If we do leave it as an external crate, it would be nice to at least propose it for `rust-lang-nursery`/`rust-lang`, so that there's a single central crate of nice assert macros that can be updated together and has consistent contribution guidelines. Right now, I'd have to pull in two tiny crates for `assert_ne!`, `assert_matches!`, and write my own `assert_panic!`, which just makes finding all of these things harder.

tbu-:
@lambda I think there's no particular need for the community to standardize on one assert crate, these don't affect downstream users of your library.

Making `assert!` print out subexpressions would unfortunately also lead to code bloat when asserts are just used to signal runtime invariants.

[Asserting a panic can be done using `#[should_fail="unique part of message"]` or `#[should_fail]`.]

lambda:
@tbu- Consolidating them in one crate would address one of the concerns that @ashleygwilliams brought up, of making it easy to apply enhancements consistently, such as formatting improvements or consistently expanding them to add extra arguments.

It would also make them more discoverable; when I [searched for "assert" on crates.io](https://crates.io/search?q=assert) to see if there was already a comprehensive assertion crate that provided `assert_ne!` or `assert_match!`, I see 289 results, the first one named `assert` just provides an approximate floating point assertion, and I have to trawl through things like `more-asserts`, `power-assert`, `assert_matches`, `static_assert`, `static_assert_macro`, `assert_ng`, `assert_ne`, `assert_approx_eq`, `asserts`, `bassert`, `passert`, etc to see which, if any, I actually want to use.

Also, I recently got burned when I went on crates.io, found a crate that looked like it did exactly what I needed, tried it out and it seemed to work at first, I was only using it for tests so I figured I didn't need to audit it too deeply, and then later on some of my tests started failing and I spent a long time debugging my code until I discovered that the crate I was depending on for testing was completely buggy. So I'm a lot more wary now of quick little hack crates from people I don't know, even for tests.

So rather than dig through all of the above mentioned crates, I just went ahead and implemented what I needed myself and moved on. Which kind of defeats the purpose of libraries. That's the big issue with an ecosystem of lots of little tiny libraries that each implement a very small piece of functionality; the work to find, validate, and trust each one individually can outweigh the benefit of using a library in the first place.

> Making `assert!` print out subexpressions would unfortunately also lead to code bloat when asserts are just used to signal runtime invariants.

This could be dealt with by making that functionality conditional on `#[cfg(test)]` or `#[cfg(debug)]`. I agree, however, that this is a good concern for a fancy `assert!` macro, rather than just having a library of `assert_<pred>!`; I don't really have a preference either way, but would find it helpful to have one or the other readily available.

> [Asserting a panic can be done using `#[should_fail="unique part of message"]` or `#[should_fail]`.]

Looks like you mean `#[should_panic]`/`#[should_panic(expected="message")]`. Which is also not easily discoverable, and more limited than an `assert_panic!`, in case you're testing something where you want some piece of code to panic, and continue to assert that some invariant holds after the panic, such as a lock held by code that panicked being poisoned.

withoutboats:
I'm not sure `assert_eq!` would be accepted into the standard library if it was proposed today, but since it is there I think it should be paired with its twin. I don't think assert ord relations are comparably analogous - eq / ne are defined by the same trait, whereas gt/lt etc are defined by a different trait which is implemented for far fewer types.

Basically, I think the reasons against this being in std all apply to assert_eq as well, and assert_eq is there.

tbu-:
I believe it's far more common (by at least one order of magnitude) to compare for equality in tests: it just provides more guarantees. In that way, I don't think that `assert_ne!` is really warranted by the existence of `assert_eq!`.

joshtriplett:
I would suggest that this seems like an eminently bikesheddable proposal.  Rather than continuing to bikeshed it, I would suggest narrowing the question: is `assert_ne!` useful enough to go into the library alongside `assert_eq!`?

If people want to see variants for lt/le/gt/ge, they could propose them.  And I'd love to see someone propose a version of `assert!` that gives details about the components of the asserted expression.  But this particular pull request proposes `assert_ne!`, and I think that's worth evaluating for inclusion, rather than blocking it because a half-dozen other things might *also* be useful above and beyond it.

I would also argue that a family of assert macros makes sense to have in the core, because they go along with `#[test]`. Having to pull in a separate crate for those seems like overkill. And these do not take up any code size at all unless used.

wycats:
> If people want to see variants for lt/le/gt/ge, they could propose them. And I'd love to see someone propose a version of assert! that gives details about the components of the asserted expression. But this particular pull request proposes assert_ne!, and I think that's worth evaluating for inclusion, rather than blocking it because a half-dozen other things might also be useful above and beyond it.

I agree. I'd find `assert_ne` useful regardless of whether we end up adding other members of the family of assertions.

I would **also** love to see a follow-up proposal to this that added the most obvious remaining members of the family, and agree with you that given that `#[test]` is built-in, a handful of assertions that are connected to Rust's operators feels like a good fit for the standard library.

ashleygwilliams:
hey! as this is my first rfc, i'm interested in how we decide to resolve this and move forward- should i make a PR?

tbu-:
The RFC will get into the final comment period (FCP) for one week at some point, afterwards it'll either get accepted or not. If it gets accepted, then it can be implemented.

ashleygwilliams:
thanks @tbu- !

@wycats: rfc submitted, https://github.com/rust-lang/rfcs/pull/1662

alexcrichton:

The libs team discussed this RFC today (along with #1662) and the overall feeling was that these are a very low maintenance burden, if any, and the road is already paved with `assert_eq` so it seems reasonable to add more.

tbu-:
Could we instead try to introduce a more integlligent macro, such as @nagisa's proposal in https://github.com/rust-lang/rfcs/pull/1662#issuecomment-229372157? I don't think introducing a new macro (which is then dropped into everyone's namespace because macro namespaces are global) to the standard library (which should have a high barrier to entry) is proper for something that is not used very much.

From the Rust compiler and standard library (i.e. the rust-lang/rust repo):
```
$ grep -r 'assert!.* < ' src/lib* src/test/  | wc -l
263
$ grep -r 'assert!.* > ' src/lib* src/test/  | wc -l
160
$ grep -r 'assert!.* >= ' src/lib* src/test/  | wc -l
174
$ grep -r 'assert!.* <= ' src/lib* src/test/  | wc -l
184
$ grep -r 'assert!.* != ' src/lib* src/test/  | wc -l
217
```

Compare that to equality:
```
$ grep -rE '(assert!.* == |assert_eq!)' src/lib* src/test/  | wc -l
10673
```

Equality is used two orders of magnitude more often than the others.

tbu-:
For all crates:
```
$ grep -rIE '(assert!.* == |assert_eq!)' * | wc -l
96910
```
vs
```
$ grep -rI 'assert!.* < ' *  | wc -l
4389
$ grep -rI 'assert!.* > ' *  | wc -l
3124
$ grep -rI 'assert!.* != ' *  | wc -l
1291
$ grep -rI 'assert!.* <= ' *  | wc -l
2276
$ grep -rI 'assert!.* >= ' *  | wc -l
1304
```

pnkfelix:
@tbu I'm not sure your grepping for various occurrences of assert actually proves anything about the actual utility of a specialized macro here.

In particular, if I want the error message on a failure to actually include descriptions of the values that were not equal, I cannot use `assert` in *any* form. I have to resort instead to something like
```rust
if (x != y) { panic!("Oops x != y for x: {:?} y: {:?}", x, y); }
```

Your data does not tell me how many times the above pattern occurs. (I certainly understand the difficulty in actually *gathering* such data; I am not trying to suggest that you attempt to gather it. I just want to warn against false inferences.)

ereichert:
@tbu In addition to @pnkfelix's comment (with which I agree) I would also suggest that a lot of programmers use what's available especially given that most tutorials and docs use nothing but the asserts.

I like that you tried to use real data but in this case I don't think past usage is a good indicator of what will be used.  The data is skewed.

tbu-:
@pnkfelix Is this just a generic counter argument or do you actually believe that this pattern exists in the wild?

I wouldn't have guessed that it does, can you point me to one instance of it? I'll try to gather data on that too when I understand how the pattern can look like.

@ereichert The data isn't the only argument. I made the argument in this thread that inequality is rarely useful to test for; you want to check for exact results in tests. The data I provided just underlines that fact.

tbu-:
@pnkfelix Also, I would probably have written that as `assert!(x != y, "{} != {}", x, y)`.

pnkfelix:
@tbu- ah I forgot that the alternative format of `assert!` allows for arbitrary format inputs after the string literal.

In a quick search over the `rustc` code base, I didn't see too many instances of that particular pattern; there were variations of it, like this: https://github.com/rust-lang/rust/blob/master/src/librustc_back/dynamic_lib.rs#L133 but even that includes a little bit more context (the `argument` to cos), and as you say, even it could be written with `assert!` instead.

lambda:
One good way to look at this is to look at examples from a language that has a richer set of assertions built in, and has for a long time so people have had a chance to use them.

Python has a [fairly rich set available for unit tests](https://docs.python.org/2.7/library/unittest.html#assert-methods). If I grep the [Django codebase](https://github.com/django/django) for anything that looks like a call to one of the unit test assertions, I see the following top 20. Note that several of these are custom assertions, not the built in ones:

```
$ git grep -E 'self.assert[a-zA-Z0-9_]+\(' | grep -Eo 'self.assert[a-zA-Z0-9]+' | sort | uniq -c | sort -nr | head -20
12279 self.assertEqual
1375 self.assertQuerysetEqual
1192 self.assertRaises
1050 self.assertTrue
 842 self.assertIn
 727 self.assertRaisesMessage
 672 self.assertFalse
 664 self.assertContains
 445 self.assertHTMLEqual
 443 self.assertIsNone
 431 self.assertIs
 380 self.assertNumQueries
 356 self.assertIsInstance
 298 self.assertNotIn
 180 self.assertRedirects
 152 self.assertOutput
 144 self.assertNoOutput
 137 self.assertNotContains
 135 self.assertNotEqual
 133 self.assertTableNotExists
```

`assertEqual` is by far the most common, with an order of magnitude more uses than the next. The next appears to be a Django specific equality assertion. After that comes `assertRaises` for testing that some thunk raises an exception, followed by the generic `assertTrue`, and then `assertIn` for testing for membership in a collection. There are a variety of uses of both built in and custom methods after that. 

`assertNotEqual` is down at number 19. In total, if I replace the `head -20` with `wc -l`, there are 122 different assertions used, most of them custom. Python itself comes with 30 different assertions; `assertNotEqual` is 11 out of the assertions that Python provides (or 12 if Django used the built-in but newer `assertRaisesRegexp` rather than its custom `assertRaisesMessage`).

I'll also note that I first tried this on [Flask](https://github.com/pallets/flask), but only found one use of the above assertions. It turns out that Flask uses [Pytest](http://docs.pytest.org/en/latest/), which instead [provides support for printing values of subexpressions directly from the `assert` statement](http://docs.pytest.org/en/latest/assert.html#assert-with-the-assert-statement).

I've also run these same search on my dayjob codebases, and see similar results; more than an order of magnitude more use of `assertEqual` than anything else, followed by `assertRaises` (and `assertFailure`, the async Twisted version of `assertRaises`), then a mix of build in ones like `assertIn`, `assertTrue`, `assertFalse`, `assertNotIn` and `assertNotEqual` along with a few custom ones, then a long tail of handfuls of uses of more obscure built in ones and custom ones.

Given the large proliferation of both built-in assertions, and custom assertions, in Python codebases using the standard unittest framework, I'd probably be more in favor of moving towards a pytest/[power-assert](https://crates.io/crates/power-assert) style `assert!` macro that prints subexpressions (probably conditionally based on `#[debug]` or `#[test]` to avoid bloat in production code) and possibly adds some kind of custom diff display, as defining these endless custom assertions mostly for the sake of just displaying the values of the subexpressions seems like a lot of wasted time and effort.

lambda:
I may have sounded a little too negative in the previous comment, so want to clarify.

I support either adding richer assertions, or making `assert!` smarter; I find it frustrating when I have to reinvent the wheel, just use `assert!` and deal with adding extra code to print out the subexpressions every time, or add more dependencies on external crates for trivial things like this. I don't think that adding `assert_ne!` should be blocked by bikeshedding or slippery-slope arguments on proliferation of assertions; it is a demonstrably useful assertion to have, I have needed it recently, so I would like to see it included, unless there is a more immediate plan to go with a more powerful `assert!`.

If we're not going with a more powerful `assert!`, then I'd probably want to propose more custom `assert_<predicate>!` macros for the same reason, but that can go in a separate RFC. I would have a slight preference for improving `assert!` over adding more individual `assert_<predicate>!`, but I find them both acceptable ways of solving the problem.

withoutboats:
I don't think it matters that assert_ne is needed less frequently than asset_eq. Its my expectation that when I have an "eq" form of an action, its paired with an "ne" form as well. For the occasion when "ne" is what I need, it is frustrating and disruptive to my work that the macro isn't at hand.

Amanieu:
I think the key here is that the majority of not-equal assertions have the form `assert!(x != CONSTANT)`. This means that if the assertion fails, you immediately know that `x == CONSTANT`, so there is no need to print the values being compared.

joshtriplett:
@Amanieu ...unless you're testing a PartialEq implementation that defines "semantic equivalence" for values that aren't structurally equal.  In which case you'd like to know what actual values you had on both sides.

lambda:
@Amanieu Where are you getting your "majority" from? When I look at one of my (Python) codebases, I find out of 20 `assertNotEqual` uses, only 6 of them have a constant on one side. Knowing what the values are, when the discrepancy in what you're expecting could come from either side of the comparison, can be useful.

When I look at Django, I see a higher percentage of comparing against a primitive constant, but I also see a good number in which neither side is a constant, and a number of tests which are testing a custom equality predicate to ensure that particular different objects don't compare as equal, while other ones do compare as equal, like @joshtriplett mentions.

Amanieu:
@lambda In one of my C codebases I have 27 asserts which contain `!=`. Of these, 24 involve comparing against a constant.

alexcrichton:
The libs team got a chance to discuss this RFC during triage yesterday, and the decision was to **merge**. This macro helps reduce the "surprise factor" when looking at the standard library's macros and also isn't much of a maintenance burden, so it seems like an all-around win. We have also decided to not merge https://github.com/rust-lang/rfcs/pull/1662 for now (the other family of assertion macros), but I'll write more over there in a second.

@ashleygwilliams could you also update this to include the `debug_assert_ne` macro akin to the `debug_assert_eq` macro as well? After that I'll merge.

steveklabnik:
@alexcrichton 

@ashleygwilliams asked me to watch out for this decision, because she's in a few meetings today. As it turns out, she's in one literally right now. She says she'll update it later today, once she's out of said meeting.

tbu-:
@alexcrichton Could the response by the libs team acknowledge that inequality assertions are used less frequently? I'm fine with the decision, but I think it's not nice to only mention the arguments in favor of the decision of the libs team, rather it should state relevant counter-arguments and say that the benefits outweigh them.

Sorry to ask for this in a thread that I have participated in as well, but I think the proclaimed "all-around win" is maybe a bit disingenuous.

tbu-:
(Unless the libs team doesn't think that frequency of use is a relevant argument.)

alexcrichton:
@tbu- ah yes I forgot to mention that we discussed that although this seems to be used not as often as the equality macro, for example, it's still a surprise that it doesn't exist. That, when coupled with the fact that the maintenance overhead is basically zero, convinced us that it'd be fine to merge.

ashleygwilliams:
@alexcrichton updated. lemme know if you'd like me to squash/rebase etc.

ashleygwilliams:
also: is the next step that i make a PR on the language repo?

tbu-:
@ashleygwilliams Yup.

alexcrichton:
Awesome, thanks again @ashleygwilliams! We actually prefer to avoid squashes on RFCs to preserve history, so as-is was fine. Also yeah the next step is to make a PR against rust-lang/rust, referencing the...

[Tracking issue](https://github.com/rust-lang/rust/issues/35073)

ashleygwilliams:
:tada: i'm on it!

PR #1665, opened 2016-07-03
Windows subsystem support

Diggsey:
[Rendered](https://github.com/Diggsey/rfcs/blob/windows-subsystem/text/0000-windows-subsystem.md)

### DIFF
--- a/text/0000-windows-subsystem.md
+++ b/text/0000-windows-subsystem.md
@@ -57,7 +57,7 @@ whichever linker is actually being used.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- Additional platform-specific API surface.
+- Additional platform-specific code.
 - The difficulty of manually calling the rust initialization code is potentially
   a more general problem, and this only solves a specific (if common) case.
 
--- a/text/0000-windows-subsystem.md
+++ b/text/0000-windows-subsystem.md
@@ -1,0 +1,96 @@
+- Feature Name: Windows Subsystem
+- Start Date: 2016-07-03
+- RFC PR: ____
+- Rust Issue: ____
+
+# Summary
+[summary]: #summary
+
+Rust programs compiled for windows will always flash up a console window on
+startup. This behavior is controlled via the `SUBSYSTEM` parameter passed to the
+linker, and so *can* be overridden with specific compiler flags. However, doing
+so will bypass the rust-specific initialization code in `libstd`.
+
+This RFC proposes supporting this case explicitly, allowing `libstd` to
+continue to be initialized correctly.
+
+# Motivation
+[motivation]: #motivation
+
+The `WINDOWS` subsystem is commonly used on windows: desktop applications
+typically do not want to flash up a console window on startup.
+
+Currently, using the `WINDOWS` subsystem from rust is undocumented, and the
+process is non-trivial:
+
+A new symbol `pub extern "system" WinMain(...)` with specific argument
+and return types must be declared, which will become the new entry point for
+the program.
+
+This is unsafe, and will skip the initialization code in `libstd`.
+
+# Detailed design
+[design]: #detailed-design
+
+When an executable is linked while compiling for a windows target, it will be
+linked for a specific *Subsystem*. The subsystem determines how the operating
+system will run the executable, and will affect the execution environment of
+the program.
+
+In practice, only two subsystems are very commonly used: `CONSOLE` and
+`WINDOWS`, and from a user's perspective, they determine whether a console will
+be automatically created when the program is started.
+
+The solution this RFC proposes is to always export both `main` and `WinMain`
+symbols from rust executables compiled for windows. The `WinMain` function
+will simply delegate to the `main` function.
+
+The end result is that rust programs will "just work" when the subsystem is
+overridden via custom linker arguments, and does not require `rustc` to
+parse those linker arguments.
+
+A possible future extension would be to add additional command-line options to
+`rustc` (and in turn, Cargo.toml) to specify the subsystem directly. `rustc`
+would automatically translate this into the correct linker arguments for
+whichever linker is actually being used.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Additional platform-specific API surface.
+- The difficulty of manually calling the rust initialization code is potentially
+  a more general problem, and this only solves a specific (if common) case.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Choosing to emit only `WinMain` or `main` using `cfg` attributes or similar.
+
+  This is problematic because it requires the compiler to know which subsystem
+  is being compiled for, and all dependencies would have to be compiled for that
+  specific subsystem.
+
+  Pushing the decision further down the line, making it purely a
+  link-time/run-time decision reduces the potential for binary incompatibility.
+
+- Add a `subsystem` function to determine which subsystem was used.
+
+  The `WinMain` function would first set an internal flag, and only then
+  delegate to the `main` function.
+
+  A function would be added to `std::os::windows`:
+
+  `fn subsystem() -> &'static str`
+
+  This would check the value of the internal flag, and return either `WINDOWS` or
+  `CONSOLE` depending on which entry point was actually used.
+
+  The `subsystem` function could be used to eg. redirect logging to a file if
+  the program is being run on the `WINDOWS` subsystem. However, it would return
+  an incorrect value if the initialization was skipped, such as if used as a
+  library from an executable written in another language.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
--- a/text/0000-windows-subsystem.md
+++ b/text/0000-windows-subsystem.md
@@ -9,7 +9,8 @@
 Rust programs compiled for windows will always flash up a console window on
 startup. This behavior is controlled via the `SUBSYSTEM` parameter passed to the
 linker, and so *can* be overridden with specific compiler flags. However, doing
-so will bypass the rust-specific initialization code in `libstd`.
+so will bypass the rust-specific initialization code in `libstd`, as the entry
+point must be named `WinMain`.
 
 This RFC proposes supporting this case explicitly, allowing `libstd` to
 continue to be initialized correctly.
@@ -33,7 +34,7 @@ This is unsafe, and will skip the initialization code in `libstd`.
 [design]: #detailed-design
 
 When an executable is linked while compiling for a windows target, it will be
-linked for a specific *Subsystem*. The subsystem determines how the operating
+linked for a specific *subsystem*. The subsystem determines how the operating
 system will run the executable, and will affect the execution environment of
 the program.
 
@@ -91,36 +92,46 @@ whichever linker is actually being used.
 # Alternatives
 [alternatives]: #alternatives
 
-- Emit either `WinMain` or `main` from `libstd` based on `cfg` options.
+- Only emit one of either `WinMain` or `main` from `rustc` based on a new
+  command line option.
 
-  This has the advantage of not requiring changes to `rustc`, but is something
-  of a non-starter since it requires a version of `libstd` for each subsystem.
+  This command line option would only be applicable when compiling an
+  executable, and only for windows platforms. No other supported platforms
+  require a different entry point or additional linker arguments for programs
+  designed to run with a graphical user interface.
 
-- Emit either `WinMain` or `main` from `rustc` based on `cfg` options.
+  `rustc` will react to this command line option by changing the exported
+  name of the entry point to `WinMain`, and passing additional arguments to
+  the linker to configure the correct subsystem. A mismatch here would result
+  in linker errors.
 
-  This would not require different versions of `libstd`, but it would require
-  recompiling all other crates depending on the value of the `cfg` option.
+  A similar option would need to be added to `Cargo.toml` to make usage as
+  simple as possible.
 
-- Emit either `WinMain` or `main` from `rustc` based on a new command line
-  option.
+  There's some bike-shedding which can be done one the exact command line
+  interface, but one possible option is shown below.
 
-  Assuming the command line option need only be specified when compiling the
-  executable itself, the dependencies would not need to be recompiled were the
-  subsystem to change.
+  Rustc usage:
+  `rustc foo.rs --crate-subsystem windows`
 
-  Choosing to emit one or the other means that the compiler and linker must
-  agree on the subsystem, or else you'll get linker errors. If `rustc` only
-  specified a `subsystem` to the linker if the option is passed, this would be
-  a fully backwards compatible change.
+  Cargo.toml
+  ```toml
+  [package]
+  # ...
 
-  A compiler option is probably desirable in addition to this RFC, but it will
-  require bike-shedding on the new command line interface, and changes to rustc
-  to be able to pass on the correct linker flags.
+  [[bin]]
+  name = "foo"
+  path = "src/foo.rs"
+  subsystem = "windows"
+  ```
 
-  A similar option would need to be added to `Cargo.toml` to make usage as simple
-  as possible.
+  The `crate-subsystem` command line option would exist on all platforms,
+  but would be ignored when compiling for a non-windows target, so as to
+  support cross-compiling. If not compiling a binary crate, specifying the
+  option is an error regardless of the target.
 
-- Add a `subsystem` function to determine which subsystem was used at runtime.
+- Export both entry points as described in this RFC, but also add a `subsystem`
+  function to `libstd` determine which subsystem was used at runtime.
 
   The `WinMain` function would first set an internal flag, and only then
   delegate to the `main` function.
@@ -137,8 +148,8 @@ whichever linker is actually being used.
   an incorrect value if the initialization was skipped, such as if used as a
   library from an executable written in another language.
 
-- Use the undocumented MSVC equivalent to weak symbols to avoid breaking
-  existing code.
+- Export both entry points as described in this RFC, but use the undocumented
+  MSVC equivalent to weak symbols to avoid breaking existing code.
 
   The parameter `/alternatename:_WinMain@16=_RustWinMain@16` can be used to
   export `WinMain` only if it is not also exported elsewhere. This is completely
--- a/text/0000-windows-subsystem.md
+++ b/text/0000-windows-subsystem.md
@@ -9,8 +9,8 @@
 Rust programs compiled for windows will always flash up a console window on
 startup. This behavior is controlled via the `SUBSYSTEM` parameter passed to the
 linker, and so *can* be overridden with specific compiler flags. However, doing
-so will bypass the rust-specific initialization code in `libstd`, as the entry
-point must be named `WinMain`.
+so will bypass the rust-specific initialization code in `libstd`, as when using
+the MSVC toolchain, the entry point must be named `WinMain`.
 
 This RFC proposes supporting this case explicitly, allowing `libstd` to
 continue to be initialized correctly.
@@ -22,13 +22,15 @@ The `WINDOWS` subsystem is commonly used on windows: desktop applications
 typically do not want to flash up a console window on startup.
 
 Currently, using the `WINDOWS` subsystem from rust is undocumented, and the
-process is non-trivial:
+process is non-trivial when targeting the MSVC toolchain:
 
 A new symbol `pub extern "system" WinMain(...)` with specific argument
 and return types must be declared, which will become the new entry point for
 the program.
 
 This is unsafe, and will skip the initialization code in `libstd`.
+
+The GNU toolchain will accept either entry point.
 
 # Detailed design
 [design]: #detailed-design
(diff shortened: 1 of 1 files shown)


sfackler:
cc @retep998

camlorn:
I haven't tried to do this myself, but the RFC is implying that it requires special compiler flags.  If this is a use case we want to support, then perhaps Cargo should also gain something for it.

There might be a similar concern with mac.  I'm not sure what the difference between a console app and a nonconsole app is over there, or even what the proper names are.  But it might be worth covering them both at the same time and with the same Cargo option.  If this is also a concern, designing it to hit both with as few changes as possible could be helpful.  Even if we need a separate RFC to finish it.

One of the things I want to do with Rust is Windows desktop apps. I had assumed this was trivially possible and that I could just kill the console at any time, but I guess I should have checked.

retep998:
Right now you can change the subsystem by passing` -Clink-args="/SUBSYSTEM:WINDOWS"` either via `cargo rustc` or `RUSTFLAGS`. It's not ideal, but to have a proper way to configure that would need to also support choosing the target NT version as well.

Diggsey:
@camlorn AFAIK, there's no equivalent for OSX or linux, all executables are the same there, the only difference between a GUI app and a console app are code differences.

camlorn:
@Diggsey 
Just verified. The only difference is whether or not it's in an app bundle.  [Source](http://stackoverflow.com/questions/23861914/how-to-hide-terminal-window-in-mac-osx).

@retep998 
Perhaps subsystems should be handled by letting us specify an oldest Windows version and having Rustc or Cargo automatically determine what it should be.  I've looked into the subsystem stuff before and came away vowing to never touch it because of poor documentation.  It's possible being able to configure the subsystem has uses I don't know about, but it feels like an option that would exist just to let me get it wrong.

I would personally vote for defaulting such an option to Windows 7, with Vista and XP requiring you to explicitly set it.

Diggsey:
@camlorn Right, but the executable does not need to be linked specially - it is sufficient to simply place it in the correct location in the application bundle. I see this as being a problem for one of the various packaging plugins for cargo to solve.

alexcrichton:
Thanks for the PR @Diggsey! I've got a few questions, but mostly along the lines of clarification.

Could you provide a detailed explanation of what `WinMain` would look like and what the compiler generates (the compiler generates `main`, so I assume it'd also generate `WinMain`). The [signature](https://msdn.microsoft.com/en-us/library/windows/desktop/ms633559(v=vs.85).aspx) looks like it has some information like `hInstance` and `nCmdShow` which we'd probably want to make sure are accessible through at least some means, right?

Also, this would in theory be a breaking change for anyone who explicitly has a `WinMain` function in their crate today I believe, so this may want to explicitly mention that and perhaps try to find a few crates that either do this or confirm that it's quite rare.

Being relatively unfamiliar with this space as well, I wouldn't mind expanding the alternatives a bit more in detail. For example I'm not sure what a `#[cfg]` attribute would look like here. Also is it true that *object files* need to be compiled with a particular subsystem in mind? If that's true then we need to consider how we ship the standard library, but I thought that this only had to do with linker flags at the very end. Along those lines, it's not clear to me at least why dependencies would need to be recompiled.

Also, how would a `subsystem` function help? Doesn't compile-time generation need to be different as opposed to something at runtime? It seems like libraries could discover whether they're part of an application or not, but it's also not clear to me how it'd solve the problem at hand.

retep998:
>Could you provide a detailed explanation of what WinMain would look like and what the compiler generates

```Rust
extern "system" WinMain(_: HINSTANCE, _: HINSTANCE, _: LPSTR, _: c_int) -> c_int {
    // call Rust's std::rt entry point
    return exit_code;
}
```
None of the arguments matter and can all be obtained via other means.

> perhaps try to find a few crates that either do this or confirm that it's quite rare.

https://github.com/search?q=WinMain+language%3Arust&type=Code&utf8=%E2%9C%93

>Also is it true that object files need to be compiled with a particular subsystem in mind

Nope, the only thing that matters is choosing the subsystem at link time. Since any non-console app can still be run from within a console, it is up to the application to determine whether stdout/stderr are valid at runtime and redirect logging appropriately.

Diggsey:
@alexcrichton Hopefully these changes address your comments.

> Doesn't compile-time generation need to be different as opposed to something at runtime?

No, the aim is for the entire compilation process to be completely identical up until link time. If libraries want to have different code-gen, they should do so via the existing feature or cfg systems.

It's not all that different from other linker options such as what icon to use for the produced executable: it greatly simplifies things if it's independent of the compilation process.

text/0000-windows-subsystem.md
```diff
@@ -0,0 +1,150 @@
...
+
+  A possible work-around for this is described below.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Emit either `WinMain` or `main` from `libstd` based on `cfg` options.
+
+  This has the advantage of not requiring changes to `rustc`, but is something
+  of a non-starter since it requires a version of `libstd` for each subsystem.
```
alexcrichton: To clarify here, the compiler generates a `main` symbol as the very last thing when an executable is generated. This main symbol is then wired up to libstd's `#[start]` function. In that sense, I don't think that this would require multiple versions of libstd?

Also, are you thinking that this alternative is basically `rustc --cfg subsystem=windows foo.rs` or something like that?

text/0000-windows-subsystem.md
```diff
@@ -0,0 +1,150 @@
...
+
+- Emit either `WinMain` or `main` from `rustc` based on `cfg` options.
+
+  This would not require different versions of `libstd`, but it would require
+  recompiling all other crates depending on the value of the `cfg` option.
```
alexcrichton: I'm not sure I understand the implications here unfortunately, how come multiple versions of libstd are needed or why would other crates need to be recompiled?

text/0000-windows-subsystem.md
```diff
@@ -0,0 +1,150 @@
...
+
+  `fn subsystem() -> &'static str`
+
+  This would check the value of the internal flag, and return either `WINDOWS` or
+  `CONSOLE` depending on which entry point was actually used.
+
+  The `subsystem` function could be used to eg. redirect logging to a file if
+  the program is being run on the `WINDOWS` subsystem. However, it would return
+  an incorrect value if the initialization was skipped, such as if used as a
+  library from an executable written in another language.
```
alexcrichton: I'm not sure I understand this alternative in terms of how it might replace the detailed design section of this RFC, wouldn't this be a nice-to-have regardless of how this RFC turns out?

text/0000-windows-subsystem.md
Diggsey: This is considering the hypothetical that the `main` symbol generation is moved out of the compiler and into `libstd` - libstd would then conditionally export a `main` or `WinMain` function depending on `cfg` options.

I've seen similar approaches in C++ libraries which attempt to abstract over platform differences.

text/0000-windows-subsystem.md
Diggsey: Here, `main` or `WinMain` is directly emitted by the compiler, but it's reusing the existing `cfg` system for the configuration (which avoids adding extra surface area to the CLI and Cargo.toml). The problem is that upstream dependencies may condition on that `cfg` option, requiring them to be recompiled for different subsystems.

text/0000-windows-subsystem.md
Diggsey: I originally included this in the RFC, but moved it to the alternatives (the alternative would be this RFC + the additional function) because of the issues with its implementation.

It would also arguably be bad practice to use such a function. Take the logging example: it's probably better practice to just directly check whether `stdout` exists before writing to it, rather than inferring it based on the subsystem.

Personally I prefer *not* having the function, so that the choice of subsystem has no effect whatsoever on rust code: everything's *much* simpler if the choice of subsystem changes one thing, and one thing only (whether a console is automatically created). However, I wanted to include it as an alternative in case someone could justify its inclusion.

DemiMarie:
One major difference is that console programs need to use `WriteConsole` and UTF-16 to do console I/O correctly.

retep998:
@DemiMarie It's trivial to figure out whether the handle you get via `GetStdHandle` is a console handle or a pipe (libstd already does this). It's also easy to determine whether writing to that handle results in an error, so you can fallback to logging to a file. There's simply no need to know which subsystem you were linked for.

text/0000-windows-subsystem.md
alexcrichton: Right now that's not actually possible with the compiler, something like a `--cfg` flag can't conditionally export different symbols when we're shipping a precompiled artifact, so maybe that's why I'm finding this confusing? Are you assuming here that we're compiling std from source on demand or something like that?

text/0000-windows-subsystem.md
alexcrichton: Hm ok, perhaps in that case this "alternative" could be left out for now?

text/0000-windows-subsystem.md
alexcrichton: Is this what I was thinking above where you pass something like `--cfg subsystem=windows` to the compiler? Sorry if I'm not following these alternatives, I'm just trying to get a better picture for what the alternatives might be in terms of technical details of how they're implemented.

CryZe:
I guess without the WinMain you can't get access to the hInstance which is probably why the Window with glutin is always so disconnected from the Main Process in the Task Manager?!

tomaka:
@CryZe That problem is probably due to the fact that you have the console subsystem.
You have to pass the `hInstance` when creating a window, however it's mostly for corner cases when DLLs want to register window classes. Glutin passes `null` for this parameter, which means "the main app".

Diggsey:
@tomaka It doesn't, either something is exporting `WinMain` (maybe one of your dependencies, it would be useful to know if that's the case!) or the linker arguments are not getting through to the linker.

retep998:
>I guess without the WinMain you can't get access to the hInstance which is probably why the Window with glutin is always so disconnected from the Main Process in the Task Manager?!

`GetModuleHandle(NULL)`. There's your hinstance.

Diggsey:
I've removed some of the "hypothetical" alternatives at @alexcrichton's suggestion, and I've fleshed out a possible design for the "command-line configurable" alternative.

I initially went with the "dual symbol" solution as the main proposal because I thought it would be a quick way to get this support into rustc, with room for later improvement. That may still be the case, but if there's no objection to the specific design I went with for the alternative, it may be better to jump straight to that.

The main problem I see is that it adds extra command-line surface for everyone, but it's only going to be relevant when targeting windows, and I can imagine there may be some objection to that.

Diggsey:
@tomaka Ah, you may have been using the GNU toolchain. This is more of a problem for MSVC. I've updated the RFC accordingly.

retep998:
@Diggsey The `/SUBSYSTEM` flag can also be used to specify the minimum NT version the binary should run on. How would you specify that with this RFC, or how is there room left for this in a future RFC?

Diggsey:
The duplicate symbol solution which this RFC proposes does not affect linker flags at all, and the version numbers specified for the subsystem do not affect which entry point is used, so I don't think there's any overlap at all there.

The alternative which adds a subsystem config to rustc and cargo does have some overlap: as proposed, rustc would not understand a version number, and so would not pass it through to the linker, but it would still be possible to manually pass in the option as we do now. I don't believe it restricts us in future: we could backwards compatibly extend the set of possible values for `subsystem` to include a version number, or we could add a separate option controlling the version.

alexcrichton:
Thanks for the update @Diggsey! This to me like it's ready for FCP, I'll bring it up at the tools triage this week.

metajack:
I think you should update the RFC text to include the linker options required for both GCC and MSVC.

Also, this is going to block Servo moving to the native Windows toolchain, which we arein the process of doing. Is there a workaround?

metajack:
I think some tooling is also needed so that this is easily controllable from rustc and Cargo. Having to use RUSTFLAGS to accomplish this is not very ergonomic. Overriding the linker in Cargo is also not a good option as unless the linker program is a native executable it will be run by cmd.exe which has 25% of the command line length that CreateProcess does. Servo hits this limit if you use fake-ld.cmd to try and add this link args that way.

So I suggest rustc be able to emit correct link args even if it can't parse them. I guess that puts me in the option 1 camp with the addition of supporting --crate-subsystem.

metajack:
cc @vvuk @larsbergstrom

alexcrichton:
The tools team got a chance to discuss this yesterday, but some points were brought up that made us think that this isn't ready for FCP just yet. @vadimcn brought up that we may be able to sidestep the `WinMain` symbol entirely by forcing the entry point as `mainCRTStartup` instead of `WinMainCRTStartup` for applications. It's thought that the two CRT entry points do the exact same initialization, and then just delegate to different entry points (would be good to confirm!). In that case, we can simply always pass the entry point for executables as `mainCRTStartup`, and then all we need to do is to pass the subsystem somehow.

I personally have come to feel that we need to address passing the linker argument as well. Figuring out how to pass linker arguments can be relatively arcane and difficult sometimes, and it's best if this kind of information is encoded into the crate anyway. @brson brought up the idea of perhaps having:

```rust
#![windows_subsystem = "windows"]
```

That way crates could encode this information (as it's probably local to a crate anyway), it's clearly windows-specific (just ignored elsewhere), is extensible to other values if we need, and can also be controlled with `#[cfg_attr]`.

Thoughts on that strategy?

alexcrichton:
Ah and a downside of that approach is now we're accumulating "cargo culted knowledge" that whenever you write a windows application you just have to blindly apply that attribute to your crate (probably without understanding it). It'd be "best" if you could, for example, link to the "windows application" crate which does all the right things for you, but seems like it's more difficult to get working.

Diggsey:
The `mainCRTStartup` option is alternative 2 in the RFC, and alternative 1 proposes a way to specify the linker arguments through rustc and cargo.

A crate attribute could work, although it would be somewhat magical. It *could* potentially be done via an external crate, but libstd would have to add and then stabilise a few new features for that to happen. I also feel like this is a primitive enough feature that it should be possible without external crates.

I can rewrite the RFC to make one of these alternatives the primary proposal if one in particular is preferred.

retep998:
I just tested and passing `-Clink-args="/subsystem:windows /entry:mainCRTStartup"` to rustc appears to work correctly. The process is spawned without a console.

metajack:
@retep998 That certainly makes things easier. It looks like this will at least unblock Servo's msvc port.

vvuk:
Unfortunately, going this route for RUSTFLAGS for servo is going to cause problems -- it's going to mean that all build_script_build.exe's built by cargo are going to have the windows subsystem set, and I think aren't going to get consoles for stdout/stderr.  Have yet to confirm that.

Also, I can't figure out how to pass anything with a space to -Clink-args via RUSTFLAGS; setting RUSTFLAGS to either "-C link-args='/subsystem:windows /entry:mainCRTstartup'" or with inner "s causes a "multiple input filenames" error.  It would be helpful if -C link-args were to append, instead of replace (e.g. behave like -Wl,... for gcc).

metajack:
I didn't consider the build.rs case. You're right that this will cause them to get built as /subsystem:windows too. I don't know whether that is a blocker or just a pain for debugging the build system.

retep998:
@vvuk When Cargo runs a build script it specifies stdout/stderr/stdin for the build script. The Windows subsystem does not override stdout/stderr/stdin set from the parent process, it only changes the default of whether a console is spawned if there isn't already a console. Therefore even if the build script was compiled for the windows subsystem, it would still work just fine.

retep998:
@vvuk For the space issue, pass it as `-Clink-args="/subsystem:windows" -Clink-args="/entry:mainCR
TStartup"` so they can be interpreted as two separate arguments just fine.

alexcrichton:
@Diggsey what do you think about updating the RFC according to what was [mentioned earlier](https://github.com/rust-lang/rfcs/pull/1665#issuecomment-235654612)? I think that should solve the motivation, plus some of the problems servo [has been running into](https://github.com/rust-lang/cargo/issues/2986) as well?

vvuk:
> @vvuk When Cargo runs a build script it specifies stdout/stderr/stdin for the build script. The Windows subsystem does not override stdout/stderr/stdin set from the parent process, it only changes the default of whether a console is spawned if there isn't already a console.

Windows subsystem binaries don't get a console at all -- a process would need to AttachConsole(ATTACH_PARENT_PROCESS) to get access to the parent's stdout/err/in if it's flagged as subsystem:windows.  Console processes attach to parent, or make a new one if needed.  So things like build.rs compiled with subsystem:windows won't print any output, so cargo won't see any cargo:... directives, nor see errors etc.

retep998:
@vvuk Having a console has little to do with stdout/stderr/stdin. If the parent process explicitly sets the stdout/stderr/stdin of the child process, then the child process can read and write to them without issue regardless of whether there is a console attached.

vvuk:
I'll retest, but it did not work for me -- cargo did not capture any of the output, at least in my local builds. Now that I think about it, it does work on appveyor, so...

retep998:
`[build] rustflags = ["-Clink-args=/subsystem:windows /entry:mainCRTStartup"]`

Tested building some stuff which relies on build scripts with that and the build scripts ran fine and the output was a valid program. Granted the cmake invocations popped up new console windows which can be a bit annoying, but it works fine otherwise. So long as you don't need to pass `-Clink-args` for anything else.

metajack:
@retep998 Popping up console windows during build makes my computer basically unusable when compiling Servo. I'd say that's a non-starter.

Diggsey:
I've rewritten the RFC based on the feedback.

text/0000-windows-subsystem.md
```diff
@@ -0,0 +1,161 @@
+- Feature Name: Windows Subsystem
+- Start Date: 2016-07-03
+- RFC PR: ____
+- Rust Issue: ____
+
+# Summary
+[summary]: #summary
+
+Rust programs compiled for windows will always flash up a console window on
```
retep998: Personally I would use the terminology `allocate a console window`, due to the behavior matching `AllocConsole`.

text/0000-windows-subsystem.md
```diff
@@ -0,0 +1,161 @@
...
+the MSVC toolchain, the entry point must be named `WinMain`.
+
+This RFC proposes supporting this case explicitly, allowing `libstd` to
+continue to be initialized correctly.
+
+# Motivation
+[motivation]: #motivation
+
+The `WINDOWS` subsystem is commonly used on windows: desktop applications
+typically do not want to flash up a console window on startup.
```
retep998: `flash up` -> `allocate` here as well.

Diggsey: This is more about what the user sees rather than a technical detail: ie. console applications can hide the console, but you can't stop it flashing up momentarily. The motivation being to improve the user experience.

camlorn:
I don't see why you need to build build scripts with a specific subsystem or with the console disabled; why can't they just build as they are now?

I can't immediately think of something that a build script might want to check for that would change if the subsystem changed.  As far as I can figure, the important thing from the perspective of build scripts is that it's build with MSVC, not that it happens to not have a console or that the binary of the build script itself runs on XP (you never distribute that piece).

Diggsey:
@camlorn The problem is that when using `RUSTFLAGS`, it will affect both the program *and* any build scripts. This is exactly one of the issues this RFC is intended to solve: you often want your program to be compiled for the windows subsystem, but you don't want that to affect the build scripts.

camlorn:
@Diggsey 
Ah, I see. I kind of had it backwards.  I read the comments as saying that this RFC would introduce that problem, not fix it.

alexcrichton:
This looks great to me, thanks @Diggsey!

3442853561:
并没有理解如何隐藏后台窗体。
Help!There is something wrong in my program.

3442853561:
Help！How can I hide the DOS form?
请问，我应该如何隐藏后台窗体？

ticki:
@3442853561 This repository is for RFCs, proposals for the Rust programmer language. This is not the right place to ask questions.

Secondly, your question is vague and cannot be answer as is. Please be specific, and ask it at either https://reddit.com/r/rust or https://stackoverflow.com.

joshtriplett:
This wouldn't involve libstd support, but for the benefit of applications using libcore, could this support the "NATIVE" subsystem as well?

retep998:
@joshtriplett the `NATIVE` subsystem is not for simple `no_std` applications. It is for kernel mode drivers which is an area that nobody has even attempted with Rust yet, so there is basically no information on what the right way forward is yet. There's nothing stopping the proposed attribute from eventually supporting that, but it'll have to be part of a larger effort to figure out how to write Windows kernel drivers in Rust.

joshtriplett:
@retep998 I wasn't suggesting that the RFC needed to define how that would work.  I just wondered if a Rust crate producing a Windows driver could use this mechanism to declare the correct target subsystem, assuming it doesn't use std and otherwise meets the assumptions of a Windows driver, or if such a crate would still have to use linker options directly.

retep998:
@joshtriplett At the moment this RFC doesn't specify any subsystems other than windows or console. To make a driver you have to pass other linker args anyway such as `/DRIVER`, so it shouldn't be too hard to pass along `/SUBSYSTEM:NATIVE` manually as well.

alexcrichton:
@rfcbot fcp merge

Discussion I think has reached a close at this point on the RFC, and having discussed it in a tools team meeting I believe our concerns are addressed as well. I propose we merge!

michaelwoerister:
@rfcbot reviewed

vadimcn:
@rfcbot reviewed

alexcrichton:
🔔 This RFC is now entering its week-long **final comment period** for merging :bell:

alexcrichton:
Ok, looks like nothing new came up during FCP, so merging! Thanks again for the RFC @Diggsey!

Tracking issue: https://github.com/rust-lang/rust/issues/37499

pravic:
I find it strange that `panic!` behavior isn't mentioned in this RFC. Currently it just exits with code 3 and prints nothing. I understand that in general you can't expect any console output in GUI executable, but the fact that the `panic` behavior isn't specified in this case is odd. 

What would be the next steps to address such an issue? Issue in the https://github.com/rust-lang/rust repository?

rpjohnst:
@pravic Panic behavior is probably better handled with [custom panic hooks](https://doc.rust-lang.org/std/panic/fn.set_hook.html).

PR #1668, opened 2016-07-06
Propose RFC for IpAddr common methods

mathphreak:
[rendered](https://github.com/mathphreak/rfcs/blob/master/text/0000-ipaddr-common-methods.md)

### DIFF
--- a/text/0000-ipaddr-common-methods.md
+++ b/text/0000-ipaddr-common-methods.md
@@ -1,0 +1,60 @@
+- Feature Name: ipaddr-common-methods
+- Start Date: 2016-07-06
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+As of Rust v1.7.0, net::Ipv4Addr and net::Ipv6Addr both expose `is_loopback()` and `is_multicast()`.
+net::IpAddr should expose those directly rather than requiring a redundant `match`.
+This should carry forward as [more methods are stabilized](https://github.com/rust-lang/rust/issues/27709).
+
+# Motivation
+[motivation]: #motivation
+
+If I have an IpAddr and I want to test whether or not it is a loopback address, I do not need or want to know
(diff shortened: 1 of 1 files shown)


eefriedman:
Please include the exact signature of the proposed methods in the proposal... something like this:

```
impl IpAddr {
    fn is_loopback(&self) -> bool { ... }
    fn is_multicast(&self) -> bool { ... }
}
```

mathphreak:
@eefriedman Just added that.

Ryman:
Similar suggestion (without required details) https://github.com/rust-lang/rfcs/issues/1554 cc @canndrew

sfackler:
It looks like there are some other shared methods between the two types (`is_documentation`, etc). Seems like we should probably add them all at once?

mathphreak:
@sfackler I only mentioned the stable methods. If it would be better to list all of them, I can make that change.

sfackler:
Yeah I figure we should add all of them - these new methods would land unstable anyway.

mathphreak:
@sfackler Done.

Also, I noticed that Ipv4Addr has an `is_link_local()` method, while Ipv6Addr has an `is_unicast_link_local()` method. It seems like those are nearly the same; should I add an `is_link_local()` as well?

sfackler:
I don't think I know enough about IPv6 to be able to answer that :)

mathphreak:
Neither do I. Unless an IPv6 expert shows up and says that those are the same, I think it's best to leave `is_link_local()` out.

alexcrichton:
I'm actually surprised to see this RFC because I thought these methods already exists! I could have sworn we either had PR or an issue about this...

@mathphreak this is actually the kind of addition where the libs team generally doesn't require an RFC. If you'd like to close this out and send a PR directly to rust-lang/rust proposing this change, I suspect we'd merge pretty quickly!

mathphreak:
@alexcrichton I wasn't quite sure if that was the case. I'll get right on that. I'll leave this open until I submit that PR.

mathphreak:
That took longer than expected, but https://github.com/rust-lang/rust/pull/34694 is now live.

PR #1671, opened 2016-07-09
Semantic "private in public" enforcement.

eddyb:
[Rendered](https://github.com/eddyb/rfcs/blob/semantic-private-in-public/text/0000-semantic-private-in-public.md)

### DIFF
--- a/text/0000-semantic-private-in-public.md
+++ b/text/0000-semantic-private-in-public.md
@@ -1,0 +1,121 @@
+- Feature Name: semantic_private_in_public
+- Start Date: 2016-07-09
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Enforce that public APIs do not expose private definitions at the semantic level, while allowing the use of private aliases and blanket implementations for convenience and automation.
+
+# Motivation
+[motivation]: #motivation
+
+The "private-in-public" rules ensure the transitivity of abstraction. That is, one must be able to name types and bounds used in public APIs, from anywhere else.
+This property can be relied upon to create perfect proxies for generic types, functions and trait implementations.
+However, the current set of rules is too strict and ignores any semantic equivalence:
+```rust
+type MyResult<T> = Result<T, String>;
+
+#[derive(Clone)]
+struct Wrap<T>(T);
+
+pub fn foo<T>(_: T) -> MyResult<()>
+    where Wrap<T>: Clone { Ok(()) }
+```
+
+The example above does not compile right now, because of `MyResult` and `Wrap` being private, even though a perfect proxy can be written as such:
+
+```rust
+fn bar<T: Clone>(x: T) -> Result<(), String> {
+    api::foo(x)
+}
+```
+
+This limitation most notably prevents derive and similar macros from generating bounds on field types, as they may contain private types, although most of the time the bound can be written in terms of type parameters and public types (`T` below), or is not needed at all (`U` below):
+```rust
+#[derive(Clone)]
+pub struct Foo<T, U>(Wrap<T>, Wrap<Rc<U>>);
+```
+
+Deriving cannot but add both a `T: Clone` and a `U: Clone` bound, in the current implementation, which is more restrictive than necessary, and ironically prevents automatic generation of perfect wrapper types.
+
+# Detailed design
+[design]: #detailed-design
+
+Function signatures, public field types, types of statics, constants and associated constants, types assigned to type aliases and associated types, and where clauses (after elaboration) must not be *less public* than the item they are found in.
+
+The previous definition of *less public* relied solely on paths as they appear in the source, but after this RFC, it is more fine-grained:
+
+An item `X` is *less public* than another item `Y` if there exists a module from where `Y` can be referred to (by any name) whereas `X` can't, taking into account `pub(restricted)` and any other privacy semantics.
+
+```rust
+pub mod m {
+    struct A;
+    // A is less public than B
+    pub(crate) trait B {}
+    // B is less public than c
+    pub fn c() {}
+}
+```
+
+A type or bound is *less public* than an item `X` if it refers to any type or trait definition that is *less public* than `X`, after resolving aliases and associated types.
+
+Where clauses in an item `X` are elabored as follows:
+ * type aliases and associated types are resolved as with all types
+ * lifetime bounds *less public* than `X` are replaced with lifetime bounds on type and lifetime parameters
+ * for each trait bound *less public* than `X`:
+  * a list of applicable implementations is computed
+  * because the bound refers to items that cannot be exported, coherence will prevent applicable implementations from existing in downstream crates
+  * type parameters of `X` are assumed to match any type, regardless of what other bounds `X` has
+  * if there is exactly one applicable `impl`, the bound is replaced with the where clauses of that `impl`, after elaborating them as well
+
+The set of bounds left after the recursive elaboration of `X`'s where clauses must not be *less public* than `X`, even if the original where clauses are allowed to.
+
+Example for use in deriving, without restricting the user or exposing private details:
+```rust
+#[derive(Debug)]
+struct Wrap<T>(T);
+
+#[derive(Copy, Clone, Debug)]
+pub struct Ref<'a, T: 'a>(&'a Wrap<T>);
+
+// deriving will produce:
+impl<'a, T> Copy for Ref<'a, T>
+    where &'a Wrap<T>: Copy {}
+impl<'a, T> Clone for Ref<'a, T>
+    where &'a Wrap<T>: Clone {...}
+
+impl<T> Debug for Wrap<T>
+    where T: Debug {...}
+impl<'a, T> Debug for Ref<'a, T>
+    where &'a Wrap<T>: Debug {...}
+
+// after elaborating where clauses:
+impl<'a, T> Copy for Ref<'a, T> {}
+impl<'a, T> Clone for Ref<'a, T> {...}
+
+impl<T> Debug for Wrap<T>
+    where T: Debug {...}
+impl<'a, T> Debug for Ref<'a, T>
+    where T: Debug {...}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Browsing sources and generating documentation becomes more complex, as private details need to be replaced with equivalent public versions before using them from other modules/crates.
+
+The "one applicable `impl`" rule works well with deriving, but adding a private `impl` can break adjacent public items, whereas the existing strategy of placing bounds on type parameters would continue to work, however restrictive it may be in general.
+
+# Alternatives
+[alternatives]: #alternatives
+
+We could leave the current situation as-is, or just resolve type aliases and associated types, leaving deriving in the same suboptimal state.
+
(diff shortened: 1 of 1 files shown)


text/0000-semantic-private-in-public.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Summary
+[summary]: #summary
+
+Enforce that public APIs do not expose private definitions at the semantic level, while allowing the use of private aliases and blanket implementations for convenience and automation.
+
+# Motivation
+[motivation]: #motivation
+
+The "private-in-public" rules ensure the transitivity of abstraction. That is, one must be able to name types and bounds used in public APIs, from anywhere else.
```
petrochenkov: "able to name" is not entirely correct, "able to name after addition of arbitrary number of reexports" is closer to the truth. E.g.
```
mod m {
    mod detail {
        pub struct S;
    }
    pub fn f() -> detail::S { detail::S }
}
```
is allowed because `S` is *potentially* nameable outside of `m` (if reexported), but not *actually* nameable.
I'd like to avoid promising namebility in docs, including RFCs, because it's already one of the most common misconceptions about our privacy system.

text/0000-semantic-private-in-public.md
```diff
@@ -0,0 +1,121 @@
...
+This property can be relied upon to create perfect proxies for generic types, functions and trait implementations.
+However, the current set of rules is too strict and ignores any semantic equivalence:
+```rust
+type MyResult<T> = Result<T, String>;
+
+#[derive(Clone)]
+struct Wrap<T>(T);
```
petrochenkov: The basic idea behind the private-in-public checker is that *values* of private types (in addition to types themselves) can't leave their module.
Can a function be tricked into returning a private type (with traits + associated types + type parameters + whatever) if it's used in one of its predicates?
Also, where can I read something about predicate elaboration?

eddyb: The return type is in the signature and checked as part of that, regardless of what happens with bounds.
With this RFC you can still *always* write a perfect proxy for functions, methods, impls, etc.

"Elaboration" as used in this RFC is described in the detailed design section. I should probably be more specific about it because it only does the work necessary to rewrite private bounds in terms of more public ones.

text/0000-semantic-private-in-public.md
eddyb: Ah, I remember reading about this before, but I wasn't clear on what decisions were taken.
It does weaken the "transitivity of abstraction" argument in that case, sadly.

text/0000-semantic-private-in-public.md
petrochenkov: Here's an example of leaking a private type through bounds, the returned type `T::Alias` seems to be completely public if checked separately. I wonder if something like this can be constructed with `PrivateWrap<T>: Bound`-like bounds. (I fully support the goals of the RFC, just trying to figure out possible holes.)
```
pub trait Tr<T> {
    type Alias = T;
}
impl<T, U> Tr<T> for U {}

struct Priv;
pub fn f<T>() -> T::Alias
    where T: Tr<Priv>
{
    panic!()
}
```

eddyb: After associated type resolution, the return type of `f` is `Priv` and that doesn't pass the check.
Even if it didn't resolve to any concrete type, the return type would be `<T as Tr<Priv>>::Alias` which still contains `Priv` so it doesn't pass the check either.

text/0000-semantic-private-in-public.md
Ericson2314: How odd, link to rational for this?

text/0000-semantic-private-in-public.md
petrochenkov: By the way, if you do the `T::Alias` -> `<T as Tr<Priv>>::Alias` transformation before checking, then predicates can become completely unchecked, because nothing private can be leaked through them<sup>1</sup>.
This is more significant relaxation of the rules, but it is simpler and removes the need in the elaboration procedure.
We discussed this with @pnkfelix [last year](http://blog.pnkfx.org/blog/2015/12/19/signatures-and-surfaces-thoughts-on-privacy-versus-dependency/) and both were generally in favor.

<sup>1</sup>"nothing private can be leaked" is the core guarantee from the current privacy system, as opposed to "nameability" from the old system.

text/0000-semantic-private-in-public.md
petrochenkov: @Ericson2314
[historical overview (with links!)](https://github.com/jseyfried/rfcs/commit/a23a5c9fc51cabc478e6ae62da084446a1410533#commitcomment-16157895)
:)

text/0000-semantic-private-in-public.md
eddyb: Well, that does reduce this RFC to "check types after normalization", I wish that was better advertised as an option (it seems perhaps even simpler than the current implementation).

text/0000-semantic-private-in-public.md
Ericson2314: Thanks so much!

eddyb:
I want to have this outside of diff comments too: @petrochenkov [pointed out](https://github.com/rust-lang/rfcs/pull/1671#discussion_r70170548) that name, alias, UFCS and associated type resolution combined (i.e. everything the compiler does to go from the syntactical to the semantic type-system) are enough to reveal all possible private *type* leaks in public APIs.
We then wouldn't even need to look at bounds, if all we care about is preventing private *type* leaks, and not the ability to write perfect proxies or anything else.

I agree this would greatly simplify the RFC and the implementation work and allow deriving to work even in dubious situations where the set of implementations available to a bound is completely hidden.

Does anyone else (e.g. from @rust-lang/lang) have anything against that?
If not, I'll update (or rewrite, even) the RFC to reflect the simpler solution.

Ericson2314:
@eddyb sounds great to me! I'd ideally like the docs to evaluate/resolve types as little as possible---just enough not leak. Maybe we can someday leverage incremental compilation for this. (Type language is pure and thus can be safely lazily evaluated.)

petrochenkov:
@eddyb 
>We then wouldn't even need to look at bounds

The main argument against this change probably lies outside of the language - documentation.
A function like `fn f<T: PrivateBounds>(arg: T) { .... }` is a black box, unless it's well documented in prose. It's not clear what arguments it can accept. If wrong arguments are provided then users see error messages like "PrivateTrait is not satisfied", what can they do with them? ([more detailed comment about this](https://internals.rust-lang.org/t/pre-rfc-sealed-traits/3108/6)).
On the other hand, derived impls - the main beneficiary from private bounds - don't care about documenting bounds.

petrochenkov:
>"nothing private can be leaked" is the core guarantee from the current privacy system (https://github.com/rust-lang/rfcs/pull/1671#discussion_r70170548)

I should write this out in more detail. The guarantee is that entities with visibility `pub(m)` "can't escape" from the module `m`.
What "can't escape" means for different entities?

1. For imports, extern crates, modules, type aliases, functions, statics, constants, foreign functions, foreign statics, struct fields, associated functions/types/constants this simply means that they can't be *named* outside of module `m`. This is ensured by the usual privacy checker ("can't access private X") and limitations on reexports ("X is private, and cannot be reexported").
2. For types (structs and enums) and traits-as-types this additionally means that *values* of those types cannot be obtained. This is ensured by the private-in-public checker. The private-in-public has to check bounds and where clauses due to situations like [this](https://github.com/rust-lang/rfcs/pull/1671#discussion_r70170231), unless types are fully transformed into their "semantic" form (`ty::Ty`) before checking.
3. Traits. When a trait should be considered escaping from its module (In addition to conditions from 1) and 2))? This is the key question to answer before relaxing the private-in-public rules.

    Regardless of the discussion about predicates above, I think a trait should be considered leaked if its items are available outside of its module. This can happen if the private trait has public sub-traits (`pub trait PubTr: PrivTr { .... }`), then its methods are available through its sub-traits. Returning a public subtrait like a trait object `fn f() -> Box<PubTr> { .... }` also kind of makes `PrivTr` leaked as a type too in some sense.
So I think private-in-public checker still needs to check bounds on trait definitions (i.e. supertraits).

crumblingstatue:
Does this aim to address https://github.com/rust-lang/rust/issues/30905?
More specifically, I am very interested in resolving this type of API weirdness:

```Rust
// This type is inacessible outside of the crate, but "public" to the entire
// crate, since it's at the crate root.
struct CratePrivType;

mod private {
    // This function is inacessible outside of the crate, so it should be okay
    // to return a crate-private type
    pub fn gimme_crate_priv() -> super::CratePrivType { unimplemented!() }
}

fn main() {}

```

```
error: private type in public interface [--explain E0446]
 --> <anon>:8:34
  |>
8 |>     pub fn gimme_crate_priv() -> super::CratePrivType { unimplemented!() }
  |>                                  ^^^^^^^^^^^^^^^^^^^^
```

To work around this without exposing the crate-private type, one can put it in a private module.
```rust
mod crate_priv {
    pub struct CratePrivType;
}

mod private {
    pub fn gimme_crate_priv() -> ::crate_priv::CratePrivType { unimplemented!() }
}

fn main() {}

```

But this once again shows the stupidity of the current analysis.
You can now use this type in your public API, even though it's supposed to be private.

```rust
mod crate_priv {
    pub struct CratePrivType;
}

mod private {
    pub fn gimme_crate_priv() -> ::crate_priv::CratePrivType { unimplemented!() }
}

pub fn pub_api_func() -> ::crate_priv::CratePrivType { unimplemented!() }

fn main() {}

```

Even rustdoc gets confused by this, and the crate private type will not get documented, even though
it will appear in the signature of pub_api_func.

eddyb:
@crumblingstatue We could restrict the check to the same module, which would allow that code to compile.
It's orthogonal to how the check is done though, i.e. it could be implemented with the current system too.

petrochenkov:
@crumblingstatue 
This was resolved by [RFC 1422](https://github.com/rust-lang/rfcs/pull/1422), your example is supposed to be written like this:
```
#![feature(pub_restricted)]

// This type is inacessible outside of the crate, but "public" to the entire
// crate, since it's at the crate root.
struct CratePrivType;

mod private {
    // This function is inacessible outside of the crate, so it should be okay
    // to return a crate-private type
    pub(crate) fn gimme_crate_priv() -> super::CratePrivType { unimplemented!() }
}

fn main() {}
```

crumblingstatue:
@petrochenkov You're right, pub(restricted) does solve the issue I was having. Thanks, and apologies for invading this thread.

nrc:
Alternative proposal on discuss (also from @eddyb): https://internals.rust-lang.org/t/rfc-reduce-false-positives-for-private-type-in-public-interface/3678

nrc:
cc @pnkfelix do you remember the ins and outs of this from the last RFC discussion?

ubsan:
ping @eddyb 

status?

eddyb:
@ubsan See the latest comments on https://github.com/rust-lang/rust/issues/34537. We're more likely to switch to a different scheme, checking types as they appear after inference, as opposed to only checking direct uses.
Not sure what the next step is, though.

ubsan:
@eddyb sounds like a close to me.

eddyb:
@ubsan Oh right, it's my PR so I can just do this.

PR #369, opened 2014-10-08
RFC: numerics reform

aturon:
This RFC is preparation for API stabilization for the `std::num` module.  The goal is to finish the simplification efforts started in [@bjz's reversal of the numerics hierarcy](https://github.com/rust-lang/rust/issues/10387).

Broadly, the proposal is to collapse the remaining numeric hierarchy in `std::num`, and to provide only limited support for generic programming (roughly, only over primitive numeric types that vary based on size).  Traits giving detailed numeric hierarchy can and should be provided separately through the Cargo ecosystem.

Thus, this RFC proposes to flatten or remove most of the traits currently provided by `std::num`, and generally to simplify the module as much as possible in preparation for API stabilization.

[Rendered](https://github.com/aturon/rfcs/blob/num-reform/active/0000-num-reform.md)

PR #1676, opened 2016-07-14
Unified machine word trait

Karl-D-Asmussen:
A proposal to unify functionality pertaining to the primitive types in a single trait, easing generic programming with primitive types.

### DIFF
--- a/0000-unified-machine-word-trait.md
+++ b/0000-unified-machine-word-trait.md
@@ -1,0 +1,36 @@
+- Feature Name: (fill me in with a unique ident, my_awesome_feature)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+One para explanation of the feature.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Detailed design
+[design]: #detailed-design
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody familiar
+with the language to understand, and for somebody familiar with the compiler to implement.
+This should get into specifics and corner-cases, and include examples of how the feature is used.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still TBD?
--- a/0000-unified-machine-word-trait.md
+++ b/0000-unified-machine-word-trait.md
@@ -1,36 +1,192 @@
-- Feature Name: (fill me in with a unique ident, my_awesome_feature)
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
-- RFC PR: (leave this empty)
-- Rust Issue: (leave this empty)
+- Feature Name: unified_machine_word_arithmetic
+- Start Date: 2016-07-13
+- RFC PR:
+- Rust Issue:
 
 # Summary
 [summary]: #summary
 
-One para explanation of the feature.
+Unify functionality peculiar to `i8`…`i64` and `u8`…`u64` in a trait containing the
+family of `overflowing`/`checked`/`wrapping`/`saturating` variants of arithmetic operations,
+as well as a few new ones.
 
 # Motivation
 [motivation]: #motivation
 
-Why are we doing this? What use cases does it support? What is the expected outcome?
+The current design has the following functions replicated over eight different types:
+
+- `checked_add`
+- `checked_div`
+- `checked_mul`
+- `checked_neg`
+- `checked_rem`
+- `checked_shl`
+- `checked_shr`
+- `checked_sub`
+- `count_ones`
+- `count_zeros`
+- `leading_zeros`
+- `overflowing_add`
+- `overflowing_div`
+- `overflowing_mul`
+- `overflowing_neg`
+- `overflowing_rem`
+- `overflowing_shl`
+- `overflowing_shr`
+- `overflowing_sub`
+- `rotate_left`
+- `rotate_right`
+- `saturating_add`
+- `saturating_mul`
+- `saturating_sub`
+- `trailing_zeros`
+- `wrapping_add`
+- `wrapping_div`
+- `wrapping_mul`
+- `wrapping_neg`
+- `wrapping_rem`
+- `wrapping_shl`
+- `wrapping_shr`
+- `wrapping_sub`
+
+Further functions that are replicated include:
+
+- `from_be`
+- `from_le`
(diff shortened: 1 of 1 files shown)


kennytm:
This is the same as the `Int` trait, which was removed from rust-lang/rust#23549 when rust-lang/rust#23104 was introduced. 

Meanwhile you could use [num_traits::PrimInt](http://rust-num.github.io/num/num_traits/int/trait.PrimInt.html) as a substitute, if you don't need to use every method (e.g. saturating_add etc).

(One of the goals of #369 was to provide "*Limited* generic programming: being able to work generically over the natural classes of *primitive* numeric types that vary only by size." But then the remaining traits proposed in #369 were removed as well `¯\_(ツ)_/¯`)

0000-unified-machine-word-trait.md
```diff
@@ -0,0 +1,196 @@
...
+
+    trait MachineArith : Copy + Eq {
+      fn abs(self) -> Self;
+      fn bitsize() -> u32; // new
+      fn bit_test(self, u32) -> bool; // new
+      fn bit(u32) -> Self; // new
+      fn bitwise_and(self, Self) -> Self; // new
+      fn bitwise_not(self) -> Self; // new
+      fn bitwise_or(self, Self) -> Self; // new
+      fn bitwise_xor(self, Self) -> Self; // new
```
kennytm: Why not simply make the trait implement `BitAnd`, `Not`, `BitOr` and `BitXor`?

withoutboats:
I think this trait should be proposed as part of a comprehensive numerics proposal to adds all traits for numerics. I don't think we can evaluate this trait without thinking about how it overlaps and interacts with potential traits like `Integer`.

MagnetoHydroDynamics:
@withoutboats According to philosophies of numerical computing proposed by John L. Gustafson in his book *The End of Error: Unum Computing*, there is merit in separating algebraic properties of the idealized numeric datatypes, from the "fast, dirty and specific" machine instructions.

In particular, we want algebraic operations to obey algebraic laws, while we want machine instructions to do one particular job in the right situation. This is particularly egregious in floating point optimizations that might insert a fused multiply-add whenever you write addition and multiplication in the same expression. Problem with FMA is that it does not obey algebraic laws (yields different results than multiply followed by add, due to rounding) but provides a more accurate result overall.

The point of having a separate machine word trait is to separate concerns: the numeric hierarchy can focus on algebra, while this trait aims to be as close to metal as possible. If there was to be a `MachineFloat` trait, it could have FMA and Exact Doc Product and all the other nice hardware-level stuff for floats. This is an attempt at the same for machine integers.

withoutboats:
@MagnetoHydroDynamics I don't disagree, I just think the whole system should be proposed concurrently, so we can discuss it holistically.

0000-unified-machine-word-trait.md
```diff
@@ -0,0 +1,196 @@
...
+
+Generalizing the arithmetic functionality into a trait, will have the following benefits:
+
+- Enable generic code working with arithmetic overflow — a clunky process at present.
+- Pave the way for general-case software implementations of large integer sizes on platforms
+    without support.
+- Make it easy to implement `i128`/`u128` (and larger types) when compiler/hardware support permits, letting
+    generic code predating the addition of the larger types work with it.
+- Generic implementation of `std::ops::*` traits with the usual panic-on-overflow semantics, using the
+    `checked_*` functions.
```
brson: This could be more explicit about the benefits. What kind of code needs to work with generic overflow and what will be better about it? How does this trait make it easier to implement large integers in software? Why are generic ops traits desirable?

brson:
Neat proposal. Thanks for writing it up @MagnetoHydroDynamics!

Why does `MachinArith` depend on `Copy + Eq`? Can you make sure that's explicit in the RFC?

Do you have an implementation already? It might be good to prove this serves its purpose well in real generic code.

This is proposing a lot of new functions. It may be more appropriate and require less debate to consider the unification of existing features independently of adding new features.

The RFC should say where the trait will end up. `core::num` and `std::num` seem good.

0000-unified-machine-word-trait.md
```diff
@@ -0,0 +1,196 @@
+- Feature Name: unified_machine_word_arithmetic
+- Start Date: 2016-07-13
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+Unify functionality peculiar to `i8`…`i64` and `u8`…`u64` in a trait containing the
+family of `overflowing`/`checked`/`wrapping`/`saturating` variants of arithmetic operations,
```
bjz: What about inheriting these from the existing traits instead of duplicating them?

0000-unified-machine-word-trait.md
bjz: Yep - same for the others.

0000-unified-machine-word-trait.md
```diff
@@ -0,0 +1,196 @@
...
+for saner arithmetic hierarchies: machine-level arithmetic is seldom easily expressible in terms of
+abstract algebra; indeed this RFC may be said to enforce a separation of concerns in library design.
+
+# Detailed design
+[design]: #detailed-design
+
+I propose the following trait to be added to the standard library where convenient:
+
+    trait MachineArith : Copy + Eq {
+      fn abs(self) -> Self;
```
bjz: I'm concerned about including `abs` here, seeing as we might want to use this in an algebraic trait hierarchy. What is your reasoning for including it here?

bjz: Ditto for `signum`, `is_positive`, and `is_negative`.

0000-unified-machine-word-trait.md
kennytm: I don't think there are existing traits for the overflowing/checked/wrapping/saturating methods.

bjz: Oh, woops! I must still be thinking in pre-1.0 land.

ubsan:
ping @aturon @Karl-D-Asmussen 

status?

burdges:
Appears `min_value()` and `max_value()` are missing from this, although probably they should be replaced by the associated constants `MIN` and `MAX` anyways.

aturon:
Thanks much for the RFC. As [some others have mentioned](https://github.com/rust-lang/rfcs/pull/1676#issuecomment-232627265), numeric traits like this have had a long, sad history in the standard library. The strong preference of the libs team has been for both generic numeric programming, and a numeric hierarchy, to live outside of the standard library, at least at first. If such an external crate saw widespread use, we could imagine moving it into the standard library. But we've gotten it wrong too many times to start immediately with `std`. We even had something relatively close to this design, and ended up abandoning it.

As such, I'm going to propose closing this RFC for the time being, in favor of development in crates.io.

@rfcbot fcp close

aturon:
We discussed this RFC in libs triage, and in general there was consensus around the [previous summary](https://github.com/rust-lang/rfcs/pull/1676#issuecomment-276536926); we don't feel ready to revisit this question in the standard library, and would much prefer this to be developed out of tree. I'm going to close, in the meantime.

PR #1678, opened 2016-07-14
Add non-panicking abs() functions to all signed integer types.

jethrogb:
[Rendered](https://github.com/jethrogb/rfcs/blob/wrapping_abs/text/0000-no-panic-abs.md)

### DIFF
--- a/text/0000-no-panic-abs.md
+++ b/text/0000-no-panic-abs.md
@@ -1,4 +1,4 @@
-- Feature Name: wrapping_abs
+- Feature Name: no_panic_abs
 - Start Date: 2016-07-14
 - RFC PR: (leave this empty)
 - Rust Issue: (leave this empty)
@@ -6,15 +6,16 @@
 # Summary
 [summary]: #summary
 
-Add a `wrapping_abs` function to all signed integer types.
+Add the non-panicking `checked_abs`, `overflowing_abs` and `wrapping_abs` 
+functions to all signed integer types.
 
 # Motivation
 [motivation]: #motivation
 
 Currently, calling `abs()` on one of the signed integer types might panic (in 
 debug mode at least) because the absolute value of the largest negative value 
-can not be represented in that signed type. However, this number could be 
-represented correctly when converting to an unsigned type of the same size.
+can not be represented in that signed type. Unlike all other integer 
+operations, there is currently not a non-panicking version on this function.
 
 # Detailed design
 [design]: #detailed-design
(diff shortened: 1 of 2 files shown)


ExpHP:
Can you name a use case?  It's hard for me to picture when it would be useful to receive a value of a different type when you want the absolute value of a number.

An alternative could be `checked_abs` (which would use `checked_neg` for negatives).

jethrogb:
Obtaining the absolute value is going to always result in a positive number, so why would you not want to store that in an unsigned type? In any case, one example where this is useful for converting a number to a sign-and-magnitude representation:

```
(signed.is_negative(),signed.wrapping_abs() as u64)
```

The alternative you mention is not really an alternative but it is something that could be added to this RFC.

jethrogb:
Added `checked_abs` and `overflowing_abs` to the RFC. I didn't add `saturating_abs` since there is no `saturating_neg`.

glaebhoerl:
Rendered link seems to be broken

glaebhoerl:
earlier proposal for reference: [Change abs() to return unsigned integers](https://github.com/rust-lang/rfcs/pull/1017)

jethrogb:
> Rendered link seems to be broken

Fixed

text/0000-no-panic-abs.md
```diff
@@ -0,0 +1,86 @@
...
+	} else {
+		self
+	}
+}
+```
+
+This RFC proposes to add the following:
+
+```rust
+pub fn checking_abs(self) -> Option<Self> {
```
brson: `checked_abs`, not `checking`, would seem to be more consistent with existing conventions.

brson:
Do all these have concrete use cases or are they just here to provide symmetry to other numeric methods? ISTM like `checked_abs` is the only one that really matters.

text/0000-no-panic-abs.md
jethrogb: Yes, of course.

jethrogb:
See above for a usecase for wrapping. I don't know of one for overflowing.

alexcrichton:

The libs team is leaning towards merging this as is, and we'd also be willing to just take a PR for this as well if you'd prefer to take that route @jethrogb :)

jethrogb:
https://github.com/rust-lang/rust/pull/35058

alexcrichton:
Thanks @jethrogb! In that case I'll just close this in favor of that.

PR #1685, opened 2016-07-24
Deprecate anonymous parameters

matklad:
[Rendered](https://github.com/matklad/rfcs/blob/anon-params/text/0000-deprecate-anonymous-parameters.md)

Issue: #1351

### DIFF
--- a/text/0000-deprecate-anonymous-parameters.md
+++ b/text/0000-deprecate-anonymous-parameters.md
@@ -1,0 +1,152 @@
+- Feature Name: deprecate_anonymous_parameters
+- Start Date: 2016-07-19
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Currently Rust allows anonymous parameters in trait methods:
+
+```Rust
+trait T {
+    fn foo(i32);
+
+    fn bar_with_default_impl(String, String) {
+
+    }
+}
+```
+
+This RFC proposes to deprecate this syntax. This RFC intentionally does not
+propose to remove this syntax.
+
+# Motivation
+[motivation]: #motivation
+
+Anonymous parameters are a [historic accident]. They do not pose any
+significant problems, but lead to a number of annoyances.
+
+1. Surprising pattern syntax in traits
+
+    ```Rust
+    trait T {
+        fn foo(x: i32);        // Ok
+        fn bar(&x: &i32);      // Ok
+        fn baz(&&x: &&i32);    // Ok
+        fn quux(&&&x: &&&i32); // Syntax error
+    }
+    ```
+
+    That is, patterns more complex than `_, foo, &foo, &&foo, mut foo` are
+    forbidden.
+
+2. Inconsistency between default implementations in traits and implementations
+   in impl blocks
+
+    ```Rust
+    trait T {
+        fn foo((x, y): (usize, usize)) { // Syntax error
+        }
+    }
+
+    impl T for S {
+        fn foo((x, y): (usize, usize)) { // Ok
+        }
+    }
+    ```
+3. Inconsistency between method declarations in traits and in extern blocks
+
+    ```Rust
+    trait T {
+        fn foo(i32);  // Ok
+    }
+
+    extern "C" {
+        fn foo(i32); // Syntax error
+    }
+    ```
+
+4. Slightly more complicated syntax analysis for LL style parsers. The parser
+   must guess if it currently parses a pattern or a type.
+
+5. Small complications for source code analyzers (e.g. [IntelliJ Rust]) and
+   potential alternative implementations.
+
+6. Potential future parsing ambiguities with named and default parameters
+   syntax.
+
+
+None of these issues is significant, but they exist.
+
+[historic accident]: https://github.com/rust-lang/rust/pull/29406#issuecomment-151859611
+[IntelliJ Rust]: https://github.com/intellij-rust/intellij-rust/commit/1bb65c47341a04aecef5fa6817e8b2b56bfc9abb#diff-66f3ba596f0ecf74a2942b3223789ab5R41
+
+
+# Detailed design
+[design]: #detailed-design
+
+Removing anonymous parameters from the language is formally a breaking change.
+The breakage can be trivially and automatically fixed by adding `_:` (suggested by @nagisa):
+
+```Rust
+trait T {
+    fn foo(_: i32);
+
+    fn bar_with_default_impl(_: String, _: String) {
+
+    }
+}
+```
+
+However this is also a major breaking change from the practical point of view.
+Parameter names are rarely omitted, but it happens. For example,
+`std::fmt::Display` is currently defined as follows:
+
+```Rust
+trait Display {
+    fn fmt(&self, &mut Formatter) -> Result;
+}
+```
+
+So the proposal is just to deprecate this syntax in the hope that the removal
+would become feasible and practical in the future. The hypothetical future may
+include:
+
+* Rust 2.0 release.
+* A tool to automatically fix deprecation warnings.
+* Storing crates on crates.io in "elaborated" syntax independent format.
+
+Enabling deprecation early makes potential future removal easier in practice.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Deprecation will require code changes without bringing any immediate benefits.
+  Until the deprecation is turned into a hard error the underlying issues will
+  stay.
+
+* It is not clear if it will ever be possible to remove this syntax entirely.
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+* Status quo.
+
+* Decide on the precise removal plan prior to deprecation.
+
+* Try to solve the underlying annoyances in some other way. For example,
+  unbounded look ahead can be used in the parser to allow both anonymous
+  parameters and the full pattern syntax.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* How often are anonymous parameters used in practice? There is a rough
+  estimate: Servo and its dependencies omit parameter names 34 times.
+
+* Is there a consensus that anonymous parameters are not a useful language
+  feature?
--- a/text/0000-deprecate-anonymous-parameters.md
+++ b/text/0000-deprecate-anonymous-parameters.md
@@ -24,8 +24,8 @@ propose to remove this syntax.
 # Motivation
 [motivation]: #motivation
 
-Anonymous parameters are a [historic accident]. They do not pose any
-significant problems, but lead to a number of annoyances.
+Anonymous parameters are a [historic accident]. They cause a number of technical
+annoyances.
 
 1. Surprising pattern syntax in traits
 
@@ -55,6 +55,7 @@ significant problems, but lead to a number of annoyances.
         }
     }
     ```
+
 3. Inconsistency between method declarations in traits and in extern blocks
 
     ```Rust
@@ -79,12 +80,33 @@ significant problems, but lead to a number of annoyances.
 
 None of these issues is significant, but they exist.
 
+
+Even if we exclude these technical drawbacks, it can be argued that allowing to
+omit parameter names unnecessary complicates the language. It is unnecessary
+because it does not make Rust more expressive and does not provide noticeable
+ergonomic improvements. It is trivial to add parameter name, and only a small
+fraction of method declarations actually omits it.
+
+Another drawback of this syntax is its impact on the learning curve. One needs
+to have a C background to understand that `fn foo(T);` means a function with
+single parameter of type `T`. If one comes from dynamically typed language like
+Python or JavaScript, this `T` looks more like a parameter name.
+
+Anonymous parameters also cause inconsistencies between trait definitions and
+implementations. One way to write an implementation is to copy the method
+prototypes from the trait into the impl block. With anonymous parameters this
+leads to syntax errors.
+
+
 [historic accident]: https://github.com/rust-lang/rust/pull/29406#issuecomment-151859611
 [IntelliJ Rust]: https://github.com/intellij-rust/intellij-rust/commit/1bb65c47341a04aecef5fa6817e8b2b56bfc9abb#diff-66f3ba596f0ecf74a2942b3223789ab5R41
 
 
 # Detailed design
 [design]: #detailed-design
+
+
+## Backward compatibility
 
 Removing anonymous parameters from the language is formally a breaking change.
 The breakage can be trivially and automatically fixed by adding `_:` (suggested by @nagisa):
@@ -109,23 +131,53 @@ trait Display {
 }
 ```
 
-So the proposal is just to deprecate this syntax in the hope that the removal
-would become feasible and practical in the future. The hypothetical future may
-include:
+Of the 5560 packages from crates.io, 416 include at least one usage of
+an anonymous parameter ([full report]).
+
+[full report]: https://github.com/rust-lang/rfcs/pull/1685#issuecomment-238954434
+
+
+## Benefits of deprecation
+
+So the proposal is just to deprecate this syntax. Phasing the syntax out of
+usage will mostly solve the learning curve problems. The technical problems
+would not be solved until the actual removal becomes feasible and
+practical. This hypothetical future may include:
 
 * Rust 2.0 release.
-* A tool to automatically fix deprecation warnings.
+* A widely deployed tool to automatically fix deprecation warnings.
 * Storing crates on crates.io in "elaborated" syntax independent format.
 
 Enabling deprecation early makes potential future removal easier in practice.
 
 
+## Deprecation strategy
+
+There are two possible ways to deprecate this syntax:
+
+### Hard deprecation
+
+One option is to produce a warning for anonymous parameters. This is backwards
+compatible, but in practice will force crate authors to actively change their
+code to avoid the warnings, causing code churn.
+
+### Soft deprecation
+
+Another option is to clearly document this syntax as deprecated and add an
+allow-by-default lint, a clippy lint, and an IntelliJ Rust inspection, but do
+not produce compiler warnings by default. This will make the update process more
+gradual, but will delay the benefits of deprecation.
+
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
-* Deprecation will require code changes without bringing any immediate benefits.
-  Until the deprecation is turned into a hard error the underlying issues will
-  stay.
+* Hard deprecation will cause code churn.
+
+* Soft deprecation might not be as efficient at removing the syntax from usage.
+
+* The technical issues can not be solved nicely until the deprecation is turned
+  into a hard error.
 
 * It is not clear if it will ever be possible to remove this syntax entirely.
 
@@ -145,8 +197,4 @@ Enabling deprecation early makes potential future removal easier in practice.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-* How often are anonymous parameters used in practice? There is a rough
-  estimate: Servo and its dependencies omit parameter names 34 times.
-
-* Is there a consensus that anonymous parameters are not a useful language
-  feature?
+* What deprecation strategy should be chosen?
--- a/text/0000-deprecate-anonymous-parameters.md
+++ b/text/0000-deprecate-anonymous-parameters.md
@@ -168,6 +168,11 @@ allow-by-default lint, a clippy lint, and an IntelliJ Rust inspection, but do
 not produce compiler warnings by default. This will make the update process more
 gradual, but will delay the benefits of deprecation.
 
+### Automatic transition
+
+Rustfmt and IntelliJ Rust can automatically change anonymous parameters to
+`_`. However it is better to manually add real names to make it obvious what
+name is expected on the `impl` side.
 
 # Drawbacks
 [drawbacks]: #drawbacks


WaDelma:
Well I have used these quite few times when parameter name wasn't nessary to understand the purpose. Didn't even realise that they cause this much problems.

nikomatsakis:
So I'm torn here. On the one hand, I wish we had removed them long ago, or found another way to resolve the parser ambiguities. I am not happy with the arbitrary limitations!

On the other hand, those could be overcome with more parser lookahead -- maybe a cover grammar too? -- and I worry a lot about this affecting a lot of people. I'm not sure it's really feasible. (I mean, it's only a deprecation, but still.)

ubsan:
@nikomatsakis parser lookahead is bad :(

Deprecations are easy. We can leave them in for a long time, only taking them out when we're sure that people have stopped using them, or almost all people have stopped using them.

ticki:
> @nikomatsakis parser lookahead is bad :(

But unavoidable.

matklad:
Is there perhaps any script to download the source code of all crates.io packages locally? It shouldn't be that much data, right? I may try to quantify "a lot of people".

nikomatsakis:
@matklad indeed there is <https://github.com/brson/stabworld>

matklad:
Here are the results:

```
416 of 5560 crates are affected.
1747 occurrences of anonymous args in total.

Top occurrences by crate:
149: adapton-0.2.4
101: gfx_core-0.4.0
47: mayda-0.2.1
36: unjson-0.0.4
35: zdd-0.1.0
34: ncollide_geometry-0.1.1
33: glium-0.15.0
31: ncollide_entities-0.5.0
24: delix-0.2.4
22: nalgebra-0.8.2
21: cs-0.23.1
20: oaty-0.1.0
17: stree-0.0.4
13: gfx_phase-0.6.0
12: vulkano-0.2.0
11: aster-0.0.4
11: rui-0.0.2
11: core-nightly-2015.1.7
11: podio-0.1.5
11: gfx_app-0.2.0
11: ntrusive-0.0.4
10: unctor-0.1.2
10: opal-0.1.1
10: simdop-0.1.0
09: cpython-0.0.5
09: nphysics3d-0.4.0
09: psilonz_algebra-0.0.1
09: gfx_device_dx11-0.3.0
09: specs-0.7.0
09: repl-0.7.1
```

Here is the full report:

https://gist.github.com/matklad/3ab67778a15778717e8b28bb01f7bacf

(warning, 700kb of XML)

brson:
I think it makes sense to deprecate, put it on the breakage wishlist, then revisit in a few years.

nikomatsakis:
I feel mildly uncomfortable with the impact of this deprecation -- particularly since it seems useful to be able to elide parameter names. The main downside of today's setup is patterns, right? So another option would be to basically require more lookahead (or a cover grammar for patterns/types, etc), so that we could support anonymous parameters without some of their downsides, right? I haven't investigated this deeply, but it seems worth considering.

matklad:
>it seems useful to be able to elide parameter names.

I personally think that anonymous parameters are a language feature of negative value, even if we forget about all the problems the RFC mentions. 

1. It adds very little convenience in writing and reading code. 
2. It introduces an alternative syntax for a relatively rare use case (there much more `impl`s than `trait`s). 

Almost zero benefit + two syntaxes for the same thing + usual bias of not adding features < 0

matklad:
>So another option would be to basically require more lookahead (or a cover grammar for patterns/types, etc), so that we could support anonymous parameters without some of their downsides, right? 

And yet another option is to do *both* parser tricks and a deprecation, to allow using full range of patterns today, and to be able to remove old syntax and tricks somewhere in the distant future.

matklad:
Looks like discussion has died down. FCP maybe?

mglagla:
:+1: on this.

Whenever i'm implementing a trait, i usually copy-paste the function signature into my code. Doing this with fmt::Display often makes me stumble a bit because of the anonymous parameter, as this is surprising.
Newcomers to the language also might get confused, since this is syntax that is not explained the book (I think?).

nrc:
> Looks like discussion has died down. FCP maybe?

I also want to do this, but fear the size of the breakage. Although I agree it is not nice to have alternate syntaxes around (especially when `_: Type` is so easy to use), I fear that hitting users with a bunch of warnings about deprecation needs more motivation than it being irritating. My feeling is thus that we should not accept this RFC. Put another way, I agree that "anonymous parameters are a language feature of negative value", but that the magnitude of negative value is outweighed by the user inconvenience of deprecating.

nrc:
@rfcbot close

There has not been much discussion on this thread, although it has quite a few upvotes. There seems to be a viable alternative for some of the motivation (parser lookahead), and the pain of deprecation (this would affect 7.5% of all crates on crates.io (https://github.com/rust-lang/rfcs/pull/1685#issuecomment-238954434)) does not seem to be out-weighed by the benefits.

OTOH, I have to say that I also don't feel strongly about this and wouldn't argue strongly against accepting, if the lang team think it is worth the user pain.

vi:
So is it for Rust 2.0?

petrochenkov:
I feel like if somebody just sent a PR implementing a compatibility lint instead on an RFC, this would have higher chances. 😄 
Even if this is not going to be removed in the close future, issuing a warning still makes sense, this won't break anyone unless they proactively `deny` warnings.

vi:
Maybe first add the lint to Clippy?

petrochenkov:
@vi 
>Maybe add the lint to the Clippy first?

Whatever helps to get this out of the language eventually.

matklad:
>Maybe add the lint to the Clippy first?

There's already an inspection/quick fix in IntelliJ Rust, though it's not enabled yet :)

steveklabnik:
> I feel like if somebody just sent a PR implementing a compatibility lint instead on an RFC, this would have higher chances

New lints require RFCs these days.

aturon:
@nrc: just noticed that your comment for rfcbot was in the incorrect format (missing "fcp"), and didn't trigger the close. Are you still inclined to close?

nrc:
@rfcbot fcp close

There has not been much discussion on this thread, although it has quite a few upvotes. There seems to be a viable alternative for some of the motivation (parser lookahead), and the pain of deprecation (this would affect 7.5% of all crates on crates.io (#1685 (comment))) does not seem to be out-weighed by the benefits.

OTOH, I have to say that I also don't feel strongly about this and wouldn't argue strongly against accepting, if the lang team think it is worth the user pain.

In reply to comments since, I think that adding this as a Clippy lint sounds like a great idea.

@aturon - ah dammit, and yes still inclined to close.

nikomatsakis:
I think the backcompat issue is just too much.

vi:
Shall it be tagged "rust-2.0"?

aturon:
@petrochenkov 

> Even if this is not going to be removed in the close future, issuing a warning still makes sense, this won't break anyone's code unless they proactively deny warnings, and won't break dependencies.

I think there's a basic disconnect here, in terms of the impact of a deprecation. While it's not technically breakage, in that the code continues to compile, the compiler is still asking you to change your code, introducing an irritation and/or churn. When that affects a large number of crates, it needs to be well-motivated and can't be done lightly (IMO).

petrochenkov:
@vi
>Shall it be tagged "rust-2.0"?

2.0 can't just happen and break stuff, it needs preparations.
The preparations include lints (built-in or third-party) and discouraging in docs.
@aturon, if warn-by-default lint is too much, then allow-by-default lint + clippy lint is the way to go.

matklad:
>Shall it be tagged "rust-2.0"?

As @petrochenkov said, this is the main idea of the RFC: if we ever want to drop this syntax, we must figure a way to phase it out without breaking the code. The deprecation is the only effective way I know.

> allow-by-default lint + clippy lint is the way to go.

What would be a motivation for this lint? I see two possible options

1. We want to drop this syntax, so disabled by default lint will be some day enabled (i.e., this RFC)

2. Anonymous parameters are considered a bad style. There is objective reason for this is https://github.com/rust-lang/rfcs/pull/1685#issuecomment-263211263: the syntax is indeed surprising if you don't know C. There are several not so important reasons (name is documentation, uniform style).

Should anonymous parameters be discouraged in idiomatic Rust, regardless of this RFC?

jimmycuadra:
I disagree with the proposal to close this RFC. While I appreciate the team's sensitivity to irritating developers by introducing warnings that might not be necessary, I think @matklad is right that we want to get rid of this "negative feature" and that deprecating it is the best way to stop the bleeding before there are significantly more crates that do this. If the feature will eventually be removed, in Rust 2.0 or otherwise, people will have to change then. Better to risk slightly annoying ~400 crate authors now than to have hard breakage for who knows how many developers a few years down the road.

matklad:
Am I correct that the biggest irritation from the deprecation is that some people have `deny(warnings)`, and so every deprecation **will** break their code?

aturon:
@jimmycuadra I agree that *if* we want to eventually remove this feature, deprecating now makes sense. But I wasn't personally convinced that the benefits outweigh the costs here. Can you elaborate on what convinced you?

petrochenkov:
@aturon 
>I wasn't personally convinced that the benefits outweigh the costs here. Can you elaborate on what convinced you?

The RFC text describes everything good enough.
It's a really nasty hack in the parser/grammar (worse than anything recently discussed in https://github.com/rust-lang/rust/issues/32409), it prevents using full pattern syntax, it prevents copy-pasting traits into impls and filling in method bodies (with misleading parsing errors).

As I said, I'd be okay with allow-by-default lint, clippy lint and discouraging in docs - this won't break anyone's code unless explicitly asked.

aturon:
> As I said, I'd be okay with allow-by-default lint, clippy lint and discouraging in docs - this won't break anyone's code unless explicitly asked.

That all seems totally fine to me, for sure.

> It's a really nasty hack in the parser/grammar . . .

I don't disagree with any of this. But we won't be able to clean up the tooling (well, rustc at least) until the feature is actually *removed*, and it's not clear if an actually-breaking Rust 2.0 is ever on the horizon. So, again, it just didn't seem worth the churn to me. 

@matklad, would you consider cleaning up your tooling so that it simply *doesn't handle* anonymous parameters, after some deprecation period? Or would you wait until Rust 2.0?

Now, it is true that if we lint, we increase the probability that things like copy-pasting traits into impls will work.

My sense is that no one on the lang team feels very strongly about this; I think we're all sympathetic to the RFC. But if we want to reconsider, I'd like to make the benefits and costs as concrete as possible.

est31:
@matklad AFAIK Rust's stability promise does not extend to codebases/people with `deny(warnings)`. The compiler is allowed to add and improve warnings.

jimmycuadra:
> @jimmycuadra I agree that if we want to eventually remove this feature, deprecating now makes sense. But I wasn't personally convinced that the benefits outweigh the costs here. Can you elaborate on what convinced you?

The benefits are the ones that have already been mentioned in the RFC itself, and in other comments. The only downside I see is annoying people that are using it in their crates now. Obviously it's more friction than none, but the fix (just naming any anonymous parameters) is also super easy.

comex:
> What convinced me not to do this now is that, as @aturon said, even if you don't have warnings on, when you get a lot of warnings in your build you're going to feel like you have to fix them. This is still churn, even if your build isn't broken.
>
> If we are going to 2.0 some day, I think this should be included. But I don't think we should do this if we're not committing to someday breaking it. We could make it an ignore-by-default lint, or put it in clippy.

Forgive me for gratuitous philosophy, but... fwiw, this reasoning basically seems backwards to me.  You're implying that churn must be justified by breakage, which is justified when it simplifies things from the implementer's perspective.  I think that usually it should work like this: breakage is justified by being necessary to implement a language change (i.e. churn), which is justified when it improves things from the user's perspective.

In this case, the less code in the wild that uses this syntax, the less likely users are to:

- run into its issues when modifying the code...

  - including to add use of the features that might not play well with it (according to the RFC text);

- or try to use it in their own code;

- or get confused if they never learned how it works ('cause it's deprecated).

None of that requires a hard break.

matklad:
>@matklad, would you consider cleaning up your tooling so that it simply doesn't handle anonymous parameters, after some deprecation period?

Yeah, I think we can do this: an IDE, unlike the compiler, does not have to be 100% correct. So, if this syntax is not actually used in practice, I think it would be OK to remove it. 

However I don't think we need to focus on the implementation side. In my opinion, slight tooling complications are one of the least compelling reason for this. 

Making the language itself more predictable and easier to learn is a more important aspect. 

So, as explained by @comex , annoying ~400 crate authors should be weighed against not a bunch of compiler and tools writers, but against a potentially unbounded set of the **users** of those crates, who would have to learn this rather obscure piece of syntax. Is it documented anywhere at all? Note that this argument, unlike the tooling one, does not depend on Rust 2.0 ever happening.

matklad:
>As I said, I'd be okay with allow-by-default lint, clippy lint and discouraging in docs - this won't break anyone's code unless explicitly asked.

> That all seems totally fine to me, for sure.

I think that this actually means deprecation, it's just not enforced. I'll update the RFC text to explicitly include this option.

matklad:
Updated the rfc. Thanks @mglagla and @comex for pointing out the "learing curve" issues with the anonymous parameters and @vi and @petrochenkov for suggesting a softer deprecation strategy!

aturon:
@nrc, can you please cancel the FCP given the additional discussion and updates to the RFC?

matklad:
Here is another spin on the copy-paste method prototype problem: 

In IntelliJ Rust, (and hopefully in RLS in the future) we have a feature to fill the missing impl methods ([gif](https://zippy.gfycat.com/GlisteningCostlyAntarcticfurseal.gif)).  For `::std::fmt::Dispaly`, we generate this

```
impl fmt::Display for S {
    fn fmt(&self, _: &mut Formatter) -> Result {
        unimplemented!()
    }
}
```

because there is no easy way to guess the name (`fmt`) that should be used in place of `_`.

scottmcm:
Would the `foo(i32)` -> `foo(_: i32)` rewrite be within the style remit of rustfmt?  Feels like switching from the soft plan to the hard one would be much easier if nobody running rustfmt would be impacted...

matklad:
>Would the foo(i32) -> foo(_: i32) rewrite be within the style remit of rustfmt? Feels like switching from the soft plan to the hard one would be much easier if nobody running rustfmt would be impacted...

This falls under the "A widely deployed tool to automatically fix deprecation warnings" bucket. Rust fmt could certainly do this, though arguably it should be handled by the RLS. IntelliJ Rust can do this already. 

In theory, IntelliJ Rust can be used to mass-fix all affected crates on github, but in practice sending 400 PRs does not seems like a good idea, and it would be better if crate authors picked meaningful names themselves.

WaDelma:
Yeah mass sending PRs is bad idea. You are just going to be reported for spam as happened when there was mass PRs for changing licenses from MIT or Apache to Apache/MIT.

tspiteri:
Would giving names to anonymous parameters in the standard library traits be a step in the right direction? It would have two benefits I can think of: firstly it would provide names to tools that can use the names from the std traits, and secondly std can be used as a template by people writing their first traits; I know I have used anonymous parameters in traits because that's what I saw in std.

aturon:
@tspiteri Sounds good to me!

nikomatsakis:
Although the FCP has completed, it seems that we had a lot of additional discussion in the interim, so I think that the original decision is moot.

nikomatsakis:
I think my opinion had mildly shifted in light of the revised text. I am in favor of the "soft deprecation" strategy, and in particular I would want `rustfmt` to automatically rewrite code. I think we should have an allow-by-default lint for now. I would consider changing that to warn-by-default in the future, but let's start soft and get louder.

(@matklad perhaps you can add a line about rustfmt too?)

nikomatsakis:
@rfcbot fcp merge

Per [my earlier comments](https://github.com/rust-lang/rfcs/pull/1685#issuecomment-279127782), I propose that we **merge** this RFC and adopt the "soft deprecation" strategy that it describes (at least initially). There is universal agreement that the feature is painful for implementors and unnecessary, and the softer deprecation strategy seems to at least put us on the path towards removing it without immediately generating a lot of warnings.

One question we may wish to discuss: what would be the criteria for adopting the harder deprecation strategy? It seems like if we want to prevent the problem from getting worse, we need at least one of the following:

- widespread use of rustfmt or IDEs, either of which rewrite code automatically;
- warnings when people use it.

But perhaps if we just don't teach it (cc @rust-lang/docs) that will suffice. In any case, I think we can merge the RFC and make those decisions down the line.

jimmycuadra:
Happy to see this resolution! Once rustfmt and/or clippy are shipped as a standard component that is installable with rustup, it'll likely become standard practice to use them, at which point the migration pain is pretty much solved with automation. That seems like the appropriate time to move to a harder deprecation of anonymous parameters.

est31:
@jimmycuadra note that clippy adoption is not required for this, as the lint is going to be included into rustc itsel, and it shouldn't be required. clippy is very opinionated and some people (including me) don't use it out of this reason.

est31:
> the lint is going to be included into rustc itself

Oh, seems I was wrong. Seems that "soft deprecation" will mean that rustc will never warn by default. I think this approach is wrong. While I wouldn't agree with removing that syntax, or not making it possible to silence the warning, I do think that at one point in time (should be after rustfmt is mature enough to produce good output so that it can be used for automatically removing that syntax), the lint should be turned to warn by default, so that the remaining users switch, and its clear that the syntax is deprecated.

vi:
Maybe "soft deprecation" may be just a beginning. A minimal action nobody against. After that the discussion about whether to make the deprecation harder or not will happen while already reaping benefits from the soft deprecation. For example, crate authors with `#![deny(anonymous_parameters)]` may enjoy some additional external tooling.

aturon:
@nikomatsakis JFYI, your rfcbot command won't work until @nrc [cancels the FCP](https://github.com/rust-lang/rfcs/pull/1685#issuecomment-277291262).

matklad:
>(@matklad perhaps you can add a line about rustfmt too?)

Fixed in https://github.com/rust-lang/rfcs/pull/1685/commits/cef568ce59b8c76ae1e00d9e2bde36967074f61d

nikomatsakis:
@aturon 

> JFYI, your rfcbot command won't work until @nrc cancels the FCP.

Oh, interesting. =) Um...

gnzlbg:
@aturon rightfully argued that: 

>  think there's a basic disconnect here, in terms of the impact of a deprecation. While it's not technically breakage, in that the code continues to compile, the compiler is still asking you to change your code, introducing an irritation and/or churn. When that affects a large number of crates, it needs to be well-motivated and can't be done lightly (IMO).

But the "fix" for a hard deprecation warning is just `#![allow(anonymous_parameters)]`.

As @aturon mentions, that is indeed some irritation and/or churn, but... it looks like pretty minor churn to me. Whether the crate authors want to just silence the warning everywhere, or just fix it everywhere (e.g. with rustfmt), or maybe fix it slowly over time by silencing them in a module by module basis, it's kind of up to them.

But I think the biggest advantage of a hard deprecation warning is that everybody gets informed (independently of whether they read github, reddit, or the release notes or not) that something is deprecated. 

We could remove the churn by showing only the first warning in rustc (or a fixed number of them), and improving the error message to inform the crate authors about their options (e.g. silence the warning globally by putting something in the root of the crate, by passing a flag to cargo / rustc, or in each module, or fixing it automatically with rustfmt, or explaining how they can fix it manually if they don't want to use rustfmt). We should also clearly informed that, although the feature is deprecated, we are probably never going to remove it from the compiler, but want to discourage its use (e.g. bug fixes performance optimizations for the feature are not prioritzied, and maybe in the future they can add some `#[!no(anonymous_parameters)]` information to the root of the crate to completely disabling the feature, which might speed up their compiletimes).

est31:
> We can start with a soft warning, that is switched to hard in the next release after ironing possible bugs with the feature, to provide a smooth experience though.

Just noting that some people have started getting rustc from their distro, and while I think its not the best way to obtain rust as rust is evolving at a rapid pace (nor do I want to support anything older than the latest stable in my crates), they only update it in large intervals, e.g. every 6 months, or for LTS distros every 4-5 years. For them it won't be smooth.

gnzlbg:
> For them it won't be smooth.

By smooth I meant a "bugless" hard warning experience, not time-continuous kind of smooth.

porky11:
I like the ability to define type signatures without names, when the names aren't used anyway.

aturon:
Let's see if this works:

@rfcbot fcp cancel

aturon:
@rfcbot fcp merge

(As per [@nikomatsakis's comment](https://github.com/rust-lang/rfcs/pull/1685#issuecomment-279127902))

text/0000-deprecate-anonymous-parameters.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+
+This RFC proposes to deprecate this syntax. This RFC intentionally does not
+propose to remove this syntax.
+
+# Motivation
+[motivation]: #motivation
+
+Anonymous parameters are a [historic accident]. They cause a number of technical
```
pnkfelix: This claim is misrepresenting the text it references.

What Niko wrote there was:
> I don't believe it was intended that parameter names can be omitted in trait methods *with* a body

(emphasis added)

I do not think it was *ever* an accident that the design allowed anonymous parameters for trait methods *without* a body.

pnkfelix:
@rfcbot reviewed

text/0000-deprecate-anonymous-parameters.md
matklad: I was referring to another part of the comment, but maybe I've misinterpreted it

> I thought we planned to require parameter names in trait definitions for this reason -- but I guess that never happened?

@nikomatsakis what would be the right way to phrase this in the RFC? Or we can just drop this sentence: it does not add anything to the current situation, it just an interesting historical detail :)

matklad: Or perhaps I was intended to provide this link instead? https://internals.rust-lang.org/t/pre-rfc-deprecating-anonymous-parameters/3710/9?u=matklad

nikomatsakis: Either way. I do remember discussing this at some point but -- as you say -- it never happened.

aturon:
This RFC has been merged -- thanks @matklad! Tracking issue is [here](https://github.com/rust-lang/rust/issues/41686).

gbutler69:
@ubsan - I would disagree with your comment:

> Deprecations are easy. We can leave them in for a long time, only taking them out when we're sure that people have stopped using them, or almost all people have stopped using them.

The problem with this is, they NEVER get removed. See the situation with Java. Almost nothing that is deprecated is EVER removed (years later). Over time, it has become a problem for the language and JVM designers because they need to continue to support things that have been deprecated for years even though it keeps them from migrating the language the way they would like to easily.

Deprecations should come with a promise to remove/disable the feature at a pre-defined time in the future. Ideally, deprecation should say which version the feature will be removed/disabled at the time it is deprecated. At that version it should be removed. Otherwise, the notion of deprecation is meaningless.

WaDelma:
As far as I understand the plan is to only remove deprecated items in rust 2.0 as removing it 1.* would be a breaking change, but I could be wrong.

PR #1686, opened 2016-07-25
Braceless Else on Braced Expressions

phoenixenero:
Make braces around `else` body optional on single expressions that require braces to increase brevity.

Addresses issue #1616.

### DIFF
--- a/text/0000-braceless-else-on-braced-expressions.md
+++ b/text/0000-braceless-else-on-braced-expressions.md
@@ -18,7 +18,7 @@ the bodies of `if` and `else`.
 This is good because it avoids the ambiguity and the potential for mistakes around nested `if`s and
 `else`s that exists in other C-family languages.
 
-But this can get sverbose when writing single nested expressions after `else` clauses.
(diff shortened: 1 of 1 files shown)


text/0000-braceless-else-on-braced-expressions.md
```diff
@@ -0,0 +1,210 @@
...
+
+But this can be too verbose when writing stuff like `match` statements inside `else` clauses.
+
+```rust
+if foo() {
+    do_this()
+} else if bar() {
+    do_this_other_thing()
+} else {
+    match baz() {
```
sfackler: How often does something like this actually occur?

nagisa:
RFC is missing a section on changes to the grammar.

nrc:
I'm not really in favour of any change here, but if we really want something, I'd prefer `else match` in the spirit of `else if`/`else if let` rather than a generic skip the braces rule.

phoenixenero:
I'll make a different, more specific proposal then.

PR #1689, opened 2016-07-26
RFC to allow self parameter be any type

KalitaAlexey:
[Rendered](https://github.com/KalitaAlexey/rfcs/blob/1577f265b3a69c6e38d71e23266500b44277229b/text/0000-any_type_self.md)

### DIFF
--- a/text/0000-any_type_self.md
+++ b/text/0000-any_type_self.md
@@ -1,0 +1,79 @@
+- Feature Name: any_type_self
+- Start Date: 2016-07-26
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow self parameter be any type.
+
+# Motivation
+[motivation]: #motivation
+
+It allows to avoid wrapper structs.
+
+# Detailed design
+[design]: #detailed-design
+
+Currently self parameter type is allowed to be only `Self`, `&Self`, `&mut Self`, `Box<Self>`.
+
+Solution: Allow `self` be any type that depends on `Self`.
+
+```rust
+struct Sound {
+}
+
+impl Sound {
+    fn play(&self) {
+        // Implementation details
+    }
+
+    fn play_with_continuation<F>(&self, continuation: F)
+        where F: FnOnce()
+    {
+        // Implementation details
+        continuation();
+    }
+}
+
+struct Game {
+    first_sound: Sound,
+    second_sound: Sound,
+}
+
+impl Game {
+    fn play_sounds(self: &Rc<UnsafeCell<Self>>) {
+        let sound = {
+            let self = unsafe {
+                &*self.get()
+            };
+            &self.first_sound
+        };
+        sound.play(move || {
+            let sound = {
+                let self = unsafe {
+                    &*self.get()
+                };
+                &self.second_sound
+            };
+            sound.play()
+        });  
+    }
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+No drawbacks.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Without it wrapper structs are required.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+No unresolved questions.


text/0000-any_type_self.md
```diff
@@ -0,0 +1,79 @@
...
+            sound.play()
+        });  
+    }
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+No drawbacks.
```
steveklabnik: There is never a proposal with no drawbacks.

steveklabnik:
I am not completely opposed to this idea, but this RFC is _extremely_ light on details.

text/0000-any_type_self.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Summary
+[summary]: #summary
+
+Allow self parameter be any type.
+
+# Motivation
+[motivation]: #motivation
+
+It allows to avoid wrapper structs.
```
jonas-schievink: An example showing where a wrapper struct is currently required would be nice

jonas-schievink:
What happens when I do this:

```rust
struct MyStruct;

impl MyStruct {
    pub fn len(self: &Vec<MyStruct>) -> usize { ... }
}

fn main() {
    let v: Vec<MyStruct> = Vec::new();
    v.len();   // which method is called?
}
```

Will this shadow `Vec::len()`? Will it not compile?

KalitaAlexey:
@steveklabnik I know but I can't figure out any more details.

eddyb:
The guarantee needs to be stronger, in order to be able to implement this in a meaningful way.
Last time I looked into this, `Deref<Target=Self>` seemed to be enough for inherent methods.
That means that the `Self` type is found when doing recursive dereferences of `

Trait methods, however, if they can be used in trait objects, require that `typeof self` is some `X<Self>` where `X<Self>: CoerceUnsized<X<Trait>>` where `X<Self>` and `X<Trait>` are structurally identical down to a matching pointer/reference to `Self`, and `Trait`, respectively.
Otherwise, associated types could be used to convert a pointer into a reference, in safe code.

cc @rust-lang/lang

KalitaAlexey:
@jonas-schievink It should be ambiguous code error.

eddyb:
@KalitaAlexey @jonas-schievink No, that method can't be in an `impl MyStruct`, at most `impl [MyStruct]`. It *bears no direct relation* to `MyStruct`, only to `[MyStruct]`.

KalitaAlexey:
@eddyb As you wish it is an option also.

text/0000-any_type_self.md
KalitaAlexey: @Jonas-schievink See detailed design section.

text/0000-any_type_self.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Currently self parameter type is allowed to be only `Self`, `&Self`, `&mut Self`, `Box<Self>`.
+
+Solution: Allow `self` be any type that depends on `Self`.
```
KalitaAlexey: @jonas-schievink Type depends on `Self` = There is `Self` in Type. For example:
`Vec<Self>`, `Box<Self>, `MyStruct<Self>`.

nrc:
I believe this has previously been discussed in the context of UFCS, it would be good to dig up and reference that discussion.

This RFC is more at the 'feature request' stage rather than ready for an RFC PR. I would recommend the author close it for now and open either an issue on this repo or a thread on internals.rust-lang.org in order to gather the details required for an RFC.

KalitaAlexey:
As @nrc suggested I closed this PR and opened a [thread on internals.rust-lang.org](https://internals.rust-lang.org/t/pre-rfc-allow-self-parameter-be-any-type/3739).

text/0000-any_type_self.md
ticki: But, how would you construct a value of `Vec<Self>` from a value of `Self`? Clearly, that doesn't make sense.

What you would need is a trait, like `From<Self>` which defines that `self` is constructed from the value in question, but at that point you introduce hidden conversions.

I believe that some form of generalized `AsRef` is what's needed.

KalitaAlexey: @ticki I can't understand you.
*Type* depends on `Self` = *Type* has `Self` as generic parameter or associated type.

ticki: Yes, but you need to define how to construct the value of that type when using the dot operator.

KalitaAlexey: @ticki I don't quite understood you. What is *dot operator*?

PR #1694, opened 2016-08-01
RFC: derive(Deref) & derive(DerefMut)

ticki:
[Rendered.](https://github.com/ticki/rfcs/blob/e1ca97c23870aa1bb2f3d04b4a46818f20d70eb3/text/0000-derive-deref.md)

### DIFF
--- a/text/0000-derive-deref.md
+++ b/text/0000-derive-deref.md
@@ -1,0 +1,47 @@
+- Feature Name: derive-deref
+- Start Date: 2016-08-01
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow deriving `Deref` and `DerefMut`.
+
+# Motivation
+[motivation]: #motivation
+
+Newtypes are a common pattern in Rust, and dereferencing is a convenient way to
+work with these. Implementing `Deref` and `DerefMut` is quite mildly annoying,
+compared to the triviality.
+
+# Detailed design
+[design]: #detailed-design
+
+Add a `derive_Deref` and `derive_DerefMut` attribute (that is, `derive(Deref)`
+and `derive(DerefMut)`), which derives the respective implementation for
+single-field `struct`s to deref to their only field.
+
+```rust
+#[derive(Deref, DerefMut)]
+struct MyType<T> {
+    inner: T,
+}
+
+// Now MyType<T> implements Deref<Target = T> as well as DerefMut.
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None.
+
+# Alternatives
+[alternatives]: #alternatives
+
+None.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None.
--- a/text/0000-derive-deref.md
+++ b/text/0000-derive-deref.md
@@ -34,12 +34,14 @@ struct MyType<T> {
 # Drawbacks
 [drawbacks]: #drawbacks
 
-None.
+This encourages using `Deref` for non-pointer types, which could be considered
+an anti-pattern.
 
 # Alternatives
 [alternatives]: #alternatives
 
-None.
+Newtype derive, where every implementation of the inner type is reimplemented
+for the newtype.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions


oli-obk:
I think this is handled pretty well by [newtype derive](https://danielkeep.github.io/rust-custom-derive/doc/newtype_derive/index.html). This should probably wait for procedural macros and then get a clean nice solution.

text/0000-derive-deref.md
```diff
@@ -0,0 +1,47 @@
...
+and `derive(DerefMut)`), which derives the respective implementation for
+single-field `struct`s to deref to their only field.
+
+```rust
+#[derive(Deref, DerefMut)]
+struct MyType<T> {
+    inner: T,
+}
+
+// Now MyType<T> implements Deref<Target = T> as well as DerefMut.
```
eddyb: You should at least write out the `impl`s in clear.

ticki:
I disagree. An upstream solution would be a lot more convinient. Also, it won't break consistency.

oli-obk:
> An upstream solution would be a lot more convinient.

Definitely. But it should still be listed as an alternative.

> Also, it won't break consistency.

but it will set the rules on how any future newtype derives should work (that should be in the disadvantage list).

Instead we could add a general `newtype_derive`, that will allow forwarding any (even user defined) traits to the inner object. Not sure if that is feasible, but it's definitely another alternative.

Stebalien:
> Instead we could add a general newtype_derive, that will allow forwarding any (even user defined) traits to the inner object. Not sure if that is feasible, but it's definitely another alternative.

That would be nice but isn't the same thing. Given `struct NewType(OldType)`, I would expect `#[newtype_derive(Trait)]` to implement `Trait` for `NewType` by forwarding all calls to `<NewType as Trait>::some_function(self, ...)` to `<OldType as X>::some_function(self.0, ...)`. `Deref::deref(&self)` would be equivalent to `Deref::deref(&self.0)`, not `&self.0` as in this proposal.

text/0000-derive-deref.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None.
+
+# Alternatives
+[alternatives]: #alternatives
+
+None.
```
nrc: Submitting an RFC with no drawbacks and no alternatives makes it very likely to be closed without the attention it might otherwise deserve.

In this case newtype deriving seems to be the main alternative. Perhaps also support for custom derives would allow this to be in a crate, rather than part of the language.

A major disadvantage I see is that this encourages using `Deref` with non-pointer types. It might be that newtypes (like slices/collections) are a reasonable use of `Deref` (c.f., using `Deref` for polymorphism which I regard as an anti-pattern, and which this is dangerously close to). It would be good to see that argument made in this RFC.

ticki: I can understand why you can see this as a disadvantage, and tbh I haden't thought of it, but I don't really think that it is actually an antipattern.

I will add it, though.

Havvy:
For completeness, you should explain what happens if you try to use derive(Deref) on a struct with more than one field or on an enum. Or using derive(DerefMut) on e.g. `struct<'a>(&'a SomeType)`.

tikue:
Are there any derives that work differently depending on the number of fields in the stuct? I feel like that inconsistency should be listed as a downside.

nikomatsakis:
We discussed this in the @rust-lang/lang meeting and decided that we'd like to move to **close** the RFC. The reasoning is basically that this would be something enabled by custom-derive, as @oli-obk [mentioned here](https://github.com/rust-lang/rfcs/pull/1694#issuecomment-236565927), and with [Macros 1.1](https://github.com/rust-lang/rfcs/pull/1681) it seems like stable custom derive will be coming up. So basically we'd rather wait and re-evaluate in that context.

nikomatsakis:
**Hear ye, hear ye!** This RFC is now entering **final comment period**. As discussed in the previous comment, the inclination is to close.

nrc:
Agreed

nikomatsakis:
Thanks to everyone for the discssion. The @rust-lang/lang team has decided to **close** this RFC for the [reasons previously given](https://github.com/rust-lang/rfcs/pull/1694#issuecomment-239554109).

PR #1096, opened 2015-04-28
remove static_assert

steveklabnik:
[rendered](https://github.com/steveklabnik/rfcs/blob/remove_static_assert/0000-remove-static-assert.md)

PR #1695, opened 2016-08-01
Add a `compile_error!` macro to libstd

sgrif:
[Rendered](https://github.com/sgrif/rfcs/blob/sg-add-error-macro/text/0000-add-error-macro.md)

### DIFF
--- a/text/0000-add-error-macro.md
+++ b/text/0000-add-error-macro.md
@@ -1,0 +1,73 @@
+- Feature Name: compile\_error\_macro
+- Start Date: 2016-08-01
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes adding a new macro to `libstd`, `compile_error!` which will
+unconditionally cause compilation to fail with the given error message when
+encountered.
+
+# Motivation
+[motivation]: #motivation
+
+Crates which work with macros or annotations such as `cfg` have no tools to
+communicate error cases in a meaningful way on stable. For example, given the
+following macro:
+
+```rust
+macro_rules! give_me_foo_or_bar {
+    (foo) => {};
+    (bar) => {};
+}
+```
+
+when invoked with `baz`, the error message will be `error: no rules expected the
+token baz`. In a real world scenario, this error may actually occur deep in a
+stack of macro calls, with an even more confusing error message. With this RFC,
+the macro author could provide the following:
+
+```rust
+macro_rules! give_me_foo_or_bar {
+    (foo) => {};
+    (bar) => {};
+    ($x:ident) => {
+        compile_error!("This macro only accepts `foo` or `bar`");
+    }
+}
+```
+
+When combined with attributes, this also provides a way for authors to validate
+combinations of features.
+
+```rust
+#[cfg(not(any(feature = "postgresql", feature = "sqlite")))]
+compile_error!("At least one backend must be used with this crate. \
+    Please specify `features = ["postgresql"]` or `features = ["sqlite"]`")
+```
+
+# Detailed design
+[design]: #detailed-design
+
+The span given for the failure should be the invocation of the `compile_error!`
+macro. The macro must take exactly one argument, which is a string literal. The
+macro will then call `span_err` with the provided message on the expansion
+context, and will not expand to any further code.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
+
+# Alternatives
+[alternatives]: #alternatives
+
+Wait for the stabilization of procedural macros, at which point a crate could
+provide this functionality.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
--- a/text/0000-add-error-macro.md
+++ b/text/0000-add-error-macro.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-This RFC proposes adding a new macro to `libstd`, `compile_error!` which will
+This RFC proposes adding a new macro to `libcore`, `compile_error!` which will
 unconditionally cause compilation to fail with the given error message when
 encountered.
 


sfackler:
I'd love something like this. I'm currently forcing type errors to do these kinds of checks, but that's not a great user experience: https://github.com/sfackler/rust-postgres/blob/306a6a9d187167b864ed5b28147b5da176549f58/src/feature_check.rs

alexcrichton:
Tagging as T-libs, but also relevant to the @rust-lang/lang team as it would essentially be adding a new macro the language (e.g. libcore wouldn't define it)

Stebalien:
This would make macro libraries like horrorshow suck significantly less. However, it would actually be nice to pass a custom span to `compile_error!`. This way DSLs (like horrorshow) could tell the user where their mistake is.

golddranks:
What about error message localizations (as there has been some discussion recently about supporting localized compiler error messages)? Even if `compile_error!` wouldn't support localization in the end, I'd like them to be mentioned so it won't go unconsidered.

sgrif:
@golddranks I'd imagine that localization will include some sort of way to detect them using `cfg` attributes, so it could be accomplished via:

```rust
#[cfg(locale = "en_US")]
compile_error!("LOOKOUT!");

#[cfg(locale = "de")]
compile_error!("ACHTUNG!");
```

text/0000-add-error-macro.md
```diff
@@ -0,0 +1,73 @@
...
+    Please specify `features = ["postgresql"]` or `features = ["sqlite"]`")
+```
+
+# Detailed design
+[design]: #detailed-design
+
+The span given for the failure should be the invocation of the `compile_error!`
+macro. The macro must take exactly one argument, which is a string literal. The
+macro will then call `span_err` with the provided message on the expansion
+context, and will not expand to any further code.
```
nrc: Would it be worth accepting a format string rather than a plain string literal? Errors seem like the kind of thing that would be commonly customised?

Is it also worth accepting alternative spans some how? That seems like a useful tool, but I don't have an idea for how it would work.

text/0000-add-error-macro.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
```
nrc: It adds another language-aware macro.

When we do have proc macros, there will be overlap between this feature and more sophisticated error handling for proc macros which might be exposed to macro_rules somehow.

text/0000-add-error-macro.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+Wait for the stabilization of procedural macros, at which point a crate could
+provide this functionality.
```
nrc: Seems there is some overlap with `static_assert!`, which iirc, was removed. It would be worth contrasting the two and motivating why this is worth adding, but `static_assert!` was removed.

text/0000-add-error-macro.md
sgrif: Given that this entirely at compile time, I'm not sure that there's anything that we could format that wouldn't be equally handled by `concat!`.

sgrif: The only way I can think of to accept alternative spans would be to take a file/line number/char number triple, which I think is fine as an option if folks are in favor of it.

text/0000-add-error-macro.md
kennytm: IIRC Rust did not had `static_assert!`, but the awkward `#[static_assert]` (#1096).

kennytm:
cc #1146. 

On nightly there are the [static_assert](https://crates.io/crates/static_assert) and [compile_msg](https://crates.io/crates/compile_msg) crates.

nikomatsakis:
I am in favor of this general idea (along with some kind of static-assert, but I guess that (could be) a separate thing). The question of how to "specify" a span is interesting -- I am presuming this error would be reported during macro expansion?

sciyoshi:
:+1:, I think this is a good first step towards improving the readability of macro errors in general. How would custom spans be specified? @sgrif any thoughts on also including a `compile_warning!` macro?

sgrif:
> I am presuming this error would be reported during macro expansion?

Correct. The implementation would essentially just be a procedural macro which passes its single argument to `cx.span_err`.

> The question of how to "specify" a span is interesting

I'm really not sure that there's a compelling use case for that, as you really don't generally have span information without procedural macros in the first place.

> any thoughts on also including a compile_warning! macro?

Seems like a reasonable addition if others are in favor of it.

Stebalien:
> I'm really not sure that there's a compelling use case for that, as you really don't generally have span information without procedural macros in the first place.

Is the span information not attached to the tokens passed to a macro? I'd like to be able point out which token in the macro input caused the macro expansion to fail (and why).

briansmith:
Semantically, this would be just a special case of `static_assert!(false, msg)`, right? Does it make sense to have both this and `static_assert!`? IMO, no, because they overlap too much. Consider, in particular, `static_assert!(cfg!(locale = "de"), "ACHTUNG!")`, in constrast to the example mentioned above.

eddyb:
@briansmith `static_assert!`, to actually evaluate its first argument, would have to run after type-checking instead of erroring at expansion time, which could result in other derived errors showing up first.
It might not be a bad idea to only have a "static assert", although consider:
```rust
// A potential desugaring, assuming CTFE that turns panics into compile errors.
const _: () = assert!(cfg!(locale = "de"), "ACHTUNG!");
```

I'm not sure if being able to implement `static_assert!` like that makes the macro unnecessary, but it certainly would affect the design.

shepmaster:
I always imagined that something like this would be implemented as a `panic!` inside of a `const fn` (which is what it appears @eddyb mentions).

eddyb:
@shepmaster I would like @tsion to confirm it, but I'm pretty sure [miri](https://github.com/tsion/miri) already supports `panic!` as a compiler error (but not unwinding or running destructors).

solson:
@shepmaster @eddyb Miri can't reach the internal panic machinery inside `panic!()` quite yet, but turning uncaught unwinds into compiler errors was always the plan.

DanielKeep:
Just popping by to note that the "force a type error on a constant" trick doesn't work in macros any more: the compiler just points at the macro invocation, and never shows the erroneous construct.

Right now, I have no way of catching and reporting input mistakes to the user.  Can we get something, *anything* for this?

aturon:
Ping @sfackler

sfackler:
Ah yes hello! I'd like to FCP to merge - while this will be implementable in the future via third party syntax extensions, it's a common enough workflow that I feel it's good to natively support it.

@rfcbot fcp merge

kennytm:
@rfcbot concern compile_warning

Do we include `compile_warning!` as well? https://github.com/rust-lang/rfcs/pull/1695#issuecomment-237656364 https://github.com/rust-lang/rfcs/pull/1695#issuecomment-238057165

sfackler:
Ah good point. I'd probably say we add it since it's a pretty straightforward addition.

iliekturtles:
What about `no_std` libraries? Should/can the macros be in `libcore`?

solson:
@iliekturtles Good call, it seems to me like they can and should be in `libcore`.

sgrif:
I think `compile_warning!` should be left to a separate RFC if it is deemed useful. It has more questions around it than `compile_error!` does (how does it interact with `#[deny(warnings)]`, is it `#[allow]`able? How is the lint specified), and has a less obvious use case.

KalitaAlexey:
@sgrif,
As you want to add the macro to the libcore, you should rename the PR to mention libcore instead of libstd.

alexcrichton:
ping @BurntSushi (checkbox)

CasualX:
As a neat workaround you can use the `env!` macro :)

I asked if this feature existed some time ago in [/r/rust](https://www.reddit.com/r/rust/comments/5trkco/hey_rustaceans_got_an_easy_question_ask_here_72017/ddsykpx/?context=4)

sfackler:
The comment period has elapsed, and this RFC is accepted! Tracking issue: https://github.com/rust-lang/rust/issues/40872

PR #1706, opened 2016-08-08
Add current RFCs to README list

pthariensflame:


durka:
Would be nice if this were automatic somehow. Maybe we could teach the highfive bot to do it?

pthariensflame:
I could certainly try!  Where is the bot's code?

aturon:
Thank you!

PR #1708, opened 2016-08-09
Use title case in template section headers

Ericson2314:


aturon:
I'm tagging with the docs team, as this question seems closest to their purview (and they're likeliest to have opinions based on the rest of the ecosystem). cc @rust-lang/docs

GuillaumeGomez:
I'm not sure to see the usefulness of such a change...

golddranks:
I think that typographic considerations like this are mainly tied to what variant of English we consider using as the "default". And remember that "titles" (like titles of CD's or something) are a different thing from "headlines". I think it's customary to use title casing with headlines in AmE, but sentence casing in BrE.

golddranks:
For what it's worth, I think that using sentence casing is easier for read, but that might be just the effect of my native language, where we don't use title casing at all.

burdges:
Yes.  I know one never does this in French, not even in titles of films, books, etc., although newspapers can capitalize their own name, ala Le Monde.  I believe Spanish and Italian are similarly opposed too title case, albeit less stringently perhaps. 

In English, I think [house style determines casing](https://en.wikipedia.org/wiki/Letter_case#Headings_and_publication_titles), but normally section headings like these are sentence case.  And [wikipedia opposes](https://en.wikipedia.org/wiki/Wikipedia:Naming_conventions_%28capitalization%29) title case even in English.  

I'd always argue against title case, even in titles, because less title case in the world saves humans effort, as sentence case contains more information by identifying proper nouns, meaning you can automatically convert from sentence case to title case rule sets, but not conversely.

steveklabnik:
So, talked about this at the docs meeting. Since this was just opened, "FCP" is a little... wrong, but just to record things:

  - [x] @GuillaumeGomez : 👎 
  - [x] @steveklabnik: 👎 
  - [x] @frewsxcv: 👎 
  - [ ] @peschkaj: (out for a few weeks, not likely to vote)
  - [ ] @jonathandturner : (out for a few weeks, not likely to vote)
  - [ ] @kmcallister: (not officially on the team yet but want to include)

Unless everyone else _really_ wants this change, and it seems it's controversial at best, I doubt we'll be accepting this.

Ericson2314:
@burdges's argument is especially compelling to me.

PR #1711, opened 2016-08-09
Target Bundles

nagisa:
cc @japaric @alexcrichton

### DIFF
--- a/text/0000-target-bundles.md
+++ b/text/0000-target-bundles.md
@@ -28,7 +28,7 @@ use-cases.
 We’ve observed a considerable amount of desire by users of the language to customize targets they
 use in the ways currently not supported by our current infrastructure (sans making changes to the
 compiler itself, of course), and noted that the current scheme is not very feasible in the long
-run. This RFC should go a great amount towards fixing the issues.
+run. This RFC should go a long way towards fixing the issues.
 
 Then, there also is a strong need to be able to inspect arbitrary parts of target specification,
 regardless of their origin. For example, in a cross-compilation setting, when the crate uses a
@@ -94,13 +94,14 @@ while LLVM appears to use the highest supported version of Dwarf by default.
 * similarly for many other variables which tweak the way compilation is done.
 
 ¹: Technically, `target_pointer_width` is used in trans, but it does not provide any extra
-information over `data_layout`, which is what it should be using.
+information over `data_layout`, which is already used elsewhere in the compiler to calculate
+layout, sizing information and alignment of all types, including the pointers.
 
 This would allow selectively reusing various conditional implementations that are present in the
 compiler for a custom target, when none of the `is_like_*` variables would fully suit the target.
 Moreover, being able to specify arbitrary cfg variables would allow easily adapting for various
 miniscule details related to the targets. For example, the targets for ARM CPUs with NEON support
-could export `target_has_neon` without any extra language support.
+could export `target_has_neon` without any extra language or compiler support.
 
 That being said, it might make sense to have a whitelist of stable options and cfg variables and
 keep everything else unstable for some duration.
@@ -110,11 +111,11 @@ keep everything else unstable for some duration.
 ```js
 {
     // REQUIRED
-    "llvm-target": "x86_64-unknown-linux-gnu",              // LLVM target triple (does not need to match with rustc triple)
-    "data-layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128", // DataLayout for the target
+    "llvm_target": "x86_64-unknown-linux-gnu",              // LLVM target triple (does not need to match with rustc triple)
+    "data_layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128", // DataLayout for the target
     "arch": "x86_64",                                       // Architecture of the target
 
-    // OPTIONAL
+    // OPTIONAL (should have sane defaults)
     // Configuration variables injected into the compilation units.
     "cfg": {
         "target_os": "linux",
@@ -131,64 +132,59 @@ keep everything else unstable for some duration.
     },
 
     // Type of, and the linker used
-    "linker-kind": "gnu",                                   // previously linker_is_gnu: bool
+    "linker_kind": "gnu",                                   // previously linker_is_gnu: bool
+                                                            // possible values: "gnu", "msvc", "osx"
     "linker": "cc",
     "ar": "ar",
-    "archive-format": "gnu",
+    "archive_format": "gnu",
 
-    // Various stuff adjusting how compilation is done
-    "function-sections": true,
-    "dynamic-linking": true,
-    "disable-redzone": false,
-    "obj-is-bitcode": false,
-    "allow-asm": true,
-    "allows-weak-linkage": true,
-    "no-default-libraries": true,
-    "custom-unwind-resume": false,                          // might make sense to merge into below
-    "eh-method": "dwarf",                                   // NEW: what EH method to use
-    "dll-storage-attrs": false,                             // NEW: should use dll storage attrs?
-    "debug-info": ["Dwarf", 4],                             // NEW: what debug info format
-    "system-abi": "C",                                      // NEW: what "system" ABI means
-    "c-abi-kind": "cabi_x86_64",                            // NEW: what C-ABI implementation to use
+    "function_sections": true,
+    "dynamic_linking": true,
+    "disable_redzone": false,
+    "obj_is_bitcode": false,
+    "allow_asm": true,
+    "allows_weak_linkage": true,
+    "no_default_libraries": true,
+    "custom_unwind_resume": false,                          // might make sense to merge into below
+    "eh_method": "dwarf",                                   // NEW: what EH method to use
+    "dll_storage_attrs": false,                             // NEW: should use dll storage attrs?
+    "debug_info": ["Dwarf", 4],                             // NEW: what debug info format
+    "system_abi": "C",                                      // NEW: what "system" ABI means
+    "c_abi_kind": "cabi_x86_64",                            // NEW: what C_ABI implementation to use
 
-    // Various stuff adjusting how linkage is done
-    "pre-link-args": ["-Wl,--as-needed", "-Wl,-z,noexecstack"],
-    "post-link-args": [],
-    "pre-link-objects-dll": [],
-    "pre-link-objects-exe": [],
-    "post-link-objects": [],
-    "late-link-args": [],
-    "gc-sections-args": [],                                 // NEW: how to strip sections
-    "rpath-prefix": "$ORIGIN",                              // CHANGED: to allow specifying rpath prefix, null to disable rpath altogether
-    "no-compiler-rt": false,
-    "metadata-section": ".note.rustc",                      // NEW: name of section for metadata
-    "has-frameworks": false,                                // NEW: is concept of frameworks supported?
-    "position-independent-executables": true,               // should become a plain linker argument?
+    "pre_link_args": ["_Wl,__as_needed", "_Wl,_z,noexecstack"],
+    "post_link_args": [],
+    "pre_link_objects_dll": [],
+    "pre_link_objects_exe": [],
+    "post_link_objects": [],
+    "late_link_args": [],
+    "gc_sections_args": [],                                 // NEW: how to strip sections
+    "rpath_prefix": "$ORIGIN",                              // CHANGED: to allow specifying rpath prefix, null to disable rpath altogether
+    "no_compiler_rt": false,
+    "metadata_section": ".note.rustc",                      // NEW: name of section for metadata
+    "has_frameworks": false,                                // NEW: is concept of frameworks supported?
+    "position_independent_executables": true,               // should become a plain linker argument?
     "lib_allocation_crate": "alloc_system",
     "exe_allocation_crate": "alloc_jemalloc",
-    // should become templates? `lib{}.so` is much nicer
+    // should become a template? `lib{}.so` is much nicer
     "dll_prefix": "lib",
     "dll_suffix": ".so",
     "exe_suffix": "",
     "staticlib_prefix": "lib",
     "staticlib_suffix": ".a",
 
-    // LLVM options
-    "cpu": "x86-64",                                        // CPU of the target
+    "cpu": "x86_64",                                        // CPU of the target
     "features": "",                                         // LLVM features
     "relocation_model": "pic",
     "code_model": "default",
-    "eliminate-frame-pointer": true,
+    "eliminate_frame_pointer": true,
 
-    // Apparently OS X is so much of a special case that even moving all the special cases into
-    // configuration is painful.
-    is_like_osx: false,
-
-    // is_like_solaris is handled by the extra gc-sections key
-    // is_like_msvc is handled by the extra metadata-section, linker-kind, eh-method,
-    // dll-storage-attrs, debug-info keys
-    // is_like_windows is handled by the extra system-abi and c-abi-kind keys
-    // is_like_android is handled by the extra debug-info key
+    // is_like_solaris is handled by the extra gc_sections_args key
+    // is_like_msvc is handled by the extra metadata_section, linker_kind, eh_method,
+    // dll_storage_attrs, debug_info keys
+    // is_like_windows is handled by the extra system_abi and c_abi_kind keys
+    // is_like_android is handled by the extra debug_info key
+    // is_like_osx is handled by linker_kind: "osx", rpath_prefix, has_frameworks, c_abi_kind
 }
 ```
 
@@ -240,6 +236,11 @@ values, though.
 
 This RFC is still very viable without the proposed changes to JSON keys (the “target quirks and
 cfg” as well as the “proposed JSON key-values” sections).
+
+Keep distributing all the built-in targets with the rustc package.
+
+Since we’re already tweaking the structure of the target json spec, instead of allowing comments in
+JSON, just switch to TOML.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-target-bundles.md
+++ b/text/0000-target-bundles.md
@@ -30,6 +30,12 @@ use in the ways currently not supported by our current infrastructure (sans maki
 compiler itself, of course), and noted that the current scheme is not very feasible in the long
 run. This RFC should go a great amount towards fixing the issues.
 
+Then, there also is a strong need to be able to inspect arbitrary parts of target specification,
+regardless of their origin. For example, in a cross-compilation setting, when the crate uses a
+build.rs script, `#[cfg]` variables are for the host, rather than the target. This way the author
+is forced to parse the target triple and figure out particularities of the target on their own, as
+rustc does not provide any way to inspect any of the built-in targets.
+
 # Detailed design
 [design]: #detailed-design
 
--- a/text/0000-target-bundles.md
+++ b/text/0000-target-bundles.md
@@ -1,0 +1,239 @@
+- Feature Name: target_bundles
+- Start Date: 2016-08
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Combine distribution of standard libraries and targets into bundles for targeting a particular
+platform. Such bundle is all you need from the Rust side to cross-compile for your target.
+
+The major ideas of this RFC is to:
+
+1. Make JSON targets as full featured as they need to be in order to support specifying targets as
+   custom as they need to be;
+2. Convert current built-in targets to the JSON targets;
+3. Change distribution of libstd to include a corresponding JSON target.
+
+# Motivation
+[motivation]: #motivation
+
+Currently there’s two different ways rustc targets are distributed: built-in targets and custom
+JSON targets. Built-in targets are very inflexible – they cannot be changed without changing and
+recompiling the compiler itself. On the other side of the spectrum, custom JSON targets are easy to
+adjust and adapt, but feature wise are very limited and rarely are suitable for the more uncommon
+use-cases.
+
+We’ve observed a considerable amount of desire by users of the language to customize targets they
+use in the ways currently not supported by our current infrastructure (sans making changes to the
+compiler itself, of course), and noted that the current scheme is not very feasible in the long
+run. This RFC should go a great amount towards fixing the issues.
+
+# Detailed design
+[design]: #detailed-design
+
+## What constitutes a fully working target?
+
+In order to have any meaningful discussion about targets we need to decide on what constitutes a
+full, complete target.
+
+Currently to compile for a specific target a number of pieces are necessary: compiler knows
+information about the target in some way, there is a set of rust standard libraries compiled for
+the target, system has native library dependencies for the target and there exists a linker which
+is capable of linking code for the target.
+
+In this RFC we will *not* propose how to call custom linkers or “extend” capabilities of the LLVM
+used.
+
+## Changes to target format
+
+### Comments and reuse
+
+There has been attempt already to migrate built-in targets to JSON targets, but it didn’t go all
+the way because of [loss of comments][comments]. One proposal was to migrate towards TOML, however
+another, [more elegant solution][jsmin] which allows us to keep using JSON was proposed by Douglas
+Crockford himself:
+
+> Suppose you are using JSON to keep configuration files, which you would like to annotate. Go
+> ahead and insert all the comments you like. Then pipe it through JSMin before handing it to your
+> JSON parser.
+
+This RFC proposes to use similar scheme: adjust the build system to remove comments from all the
+JSON target specifications before packaging them for use by rustc. This way we get to keep using
+JSON and can keep on having comments in the checked-in target specifications.
+
+Similar preprocessing step could be used to implement some form of target inheritance so the
+duplication between built-in targets could be reduced greatly.
+
+[comments]: https://github.com/rust-lang/rust/pull/34980#issuecomment-234683183
+[jsmin]: https://plus.google.com/+DouglasCrockfordEsq/posts/RK8qyGVaGSr
+
+### Target quirks and cfg
+
+Currently we have a few keys dealing with the configuration variables only: `target_os`,
+`target_env`, `target_vendor`, `target_family`, `target_endian`, `target_pointer_width`¹, etc;
+Then there’s options from which some configuration variables are derived, but they are also used to
+tweak compilation: `is_like_windows`, `is_like_osx`, `is_like_solaris`, etc.
+
+This RFC proposes replacing these keys with a different set of keys which explicitly control
+configuration variables and keys which control compilation details:
+
+* `cfg: {"target_env": "msvc", windows: null }` would result in two cofiguration variables
+`#[cfg(windows)]` and `#[cfg(target_env = "msvc")]` evaluating to true, but no variables provided
+by the `cfg` key would get used to influence the behaviour of the compiler; and
+* `debuginfo: ["CodeView", 1]` some targets require different debuginfo format than what LLVM
+generates by default. MSVC targets want CodeView version 1, OS X and Android want Dwarf version 2,
+while LLVM appears to use the highest supported version of Dwarf by default.
+* similarly for many other variables which tweak the way compilation is done.
+
+¹: Technically, `target_pointer_width` is used in trans, but it does not provide any extra
+information over `data_layout`, which is what it should be using.
+
+This would allow selectively reusing various conditional implementations that are present in the
+compiler for a custom target, when none of the `is_like_*` variables would fully suit the target.
+Moreover, being able to specify arbitrary cfg variables would allow easily adapting for various
+miniscule details related to the targets. For example, the targets for ARM CPUs with NEON support
+could export `target_has_neon` without any extra language support.
+
+That being said, it might make sense to have a whitelist of stable options and cfg variables and
+keep everything else unstable for some duration.
+
+### Proposed JSON key-values
+
+```json
+{
+    // REQUIRED
+    "llvm-target": "x86_64-unknown-linux-gnu",              // LLVM target triple (does not need to match with rustc triple)
+    "data-layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128", // DataLayout for the target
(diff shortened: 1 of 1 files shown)


text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+    "cfg": {
+        "target_os": "linux",
+        "target_family": "unix",
+        "target_arch": "x86_64",
+        "target_endian": "little",
+        "target_pointer_width": "64",
+        "target_env": "gnu",
+        "target_vendor": "unknown",
+        "target_has_atomic": ["8", "16", "32", "64"],       // any of #[cfg(target_has_atomic={"8", "16", "32", "64")] work.
+        "target_has_atomic_ptr": null,
```
Amanieu: I would prefer the previous system where you only need to specify a single value for atomic support: just the maximum atomic bit width supported by the target. It's much simpler and gives the same information.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+We’ve observed a considerable amount of desire by users of the language to customize targets they
+use in the ways currently not supported by our current infrastructure (sans making changes to the
+compiler itself, of course), and noted that the current scheme is not very feasible in the long
+run. This RFC should go a great amount towards fixing the issues.
+
+Then, there also is a strong need to be able to inspect arbitrary parts of target specification,
+regardless of their origin. For example, in a cross-compilation setting, when the crate uses a
+build.rs script, `#[cfg]` variables are for the host, rather than the target. This way the author
+is forced to parse the target triple and figure out particularities of the target on their own, as
+rustc does not provide any way to inspect any of the built-in targets.
```
jmesmon: Well, there is `rustc --print cfg`, which does allow quite a bit of inspection for built-in targets (presuming one can execute rustc) [which implies that parsing target triples should never be needed].

That said, I'm in favor of unifying our target specifications.

jmesmon:
If we're going to change the target format anyway (by moving around json keys as proposed here), I'm not sure what the downside of switching to toml would be (and at that point we'll get comments natively, and have the same config format as cargo)

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+
+`#![no_core]` users will have to download the rust-std even if they have no use for libraries in
+there.
+
+Proposed change to add a key for each option, instead of having an umbrella `is_like_*` keys, will
+result in big increase of such options. All of these are optional and should have sane default
+values, though.
+
+# Alternatives
+[alternatives]: #alternatives
```
jmesmon: Seems like an alternative (in some sense) would be to have a `rustc --print target-spec` option that spit out more info.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+Moreover, being able to specify arbitrary cfg variables would allow easily adapting for various
+miniscule details related to the targets. For example, the targets for ARM CPUs with NEON support
+could export `target_has_neon` without any extra language support.
+
+That being said, it might make sense to have a whitelist of stable options and cfg variables and
+keep everything else unstable for some duration.
+
+### Proposed JSON key-values
+
+```json
```
japaric: Changing `json` to `js` would generate a better render (comments wouldn't be highlighted in red)

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+Currently every built-in target is distributed along with the rustc compiler. This is suboptimal,
+because in majority of cases users are interested in targets which rustc can compile for, rather
+than the built-in targets rustc knows about, therefore distributing targets built-in into rustc is
+providing no benefits.
+
+Native libraries and linkers aside, it is obvious there’s little sense in distributing targets
+separately from the standard libraries for them. This RFC proposes to change the distribution so
+the rust-std and rust (not rustc or rust-docs) packages begin including the JSON target description
+for the target which the standard library targets. The target JSONs would get installed in
+`$sysroot/share/rustc/targets/` or a similar directory, and the `RUST_TARGET_PATH` environment
```
jmesmon: I'd propose placing the target specifications into the `$sysroot/lib/rustlib` directory instead, so they live right next to the std libraries they specify. This allows us to avoid claiming another directory and provides a "1-stop-shop" of sorts for examining the targets available to the compiler.

In meta-rust we've been shipping a [patch](https://github.com/meta-rust/meta-rust/blob/master/recipes-devtools/rust/files/rust/0002-Target-add-default-target.json-path-libdir-rust-targ.patch) that does that for a while (to make our life a bit easier).

text/0000-target-bundles.md
nagisa: Its not an alternative at all, because that option does not help *creating* and distributing custom targets.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+{
+    // REQUIRED
+    "llvm-target": "x86_64-unknown-linux-gnu",              // LLVM target triple (does not need to match with rustc triple)
+    "data-layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128", // DataLayout for the target
+    "arch": "x86_64",                                       // Architecture of the target
+
+    // OPTIONAL
+    // Configuration variables injected into the compilation units.
+    "cfg": {
+        "target_os": "linux",
```
japaric: Some of these are currently [mandatory](https://github.com/rust-lang/rust/blob/f0139140f6a2d1207cb21336b0faca69b5a337b2/src/librustc_back/target/mod.rs#L441-L442).

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently there’s two different ways rustc targets are distributed: built-in targets and custom
+JSON targets. Built-in targets are very inflexible – they cannot be changed without changing and
+recompiling the compiler itself. On the other side of the spectrum, custom JSON targets are easy to
+adjust and adapt, but feature wise are very limited and rarely are suitable for the more uncommon
+use-cases.
+
+We’ve observed a considerable amount of desire by users of the language to customize targets they
```
japaric: With this change, the compiler will have to learn to *not* use `std` binaries pre-compiled for `$TARGET` (e.g. `x86_64-unknown-linux-gnu`) if its corresponding target file (`$TARGET.json`) has been modified because codegen options may have been changed in an incompatible way. This is not a problem today because one can't override a built-in target specification with a target specification file.

text/0000-target-bundles.md
nagisa: I address this in lines 96-97. Namely, the `target_pointer_width` is used in trans, but this same information is already available from data-layout, which trans really should use in the first place. Similarly, `target_endian` is also available from data-layout, but isn’t used anywhere in the compiler AFAICT.

nagisa:
@Amanieu commented on the line 128 (about `target_has_atomic`):

> I would prefer the previous system where you only need to specify a single value for atomic support: just the maximum atomic bit width supported by the target. It's much simpler and gives the same information.

GitHub really needs a way to mark comments as not-outdated.

text/0000-target-bundles.md
jmesmon: It's a solution to the issue of being able to determine information about a target externally (which I saw as the motivation for moving all targets to json), but you're right that it doesn't have anything to do with creating & distributing custom targets.

text/0000-target-bundles.md
japaric: I also meant `target_os`, which can't be derived from `data-layout`; it should have a comment about it being a mandatory field. I'm not sure about deriving `target_pointer_witdth` et al. from `data-layout` because we don't have control over its format/contents; it might stop containing information about the pointer width in the future.

text/0000-target-bundles.md
japaric: I prefer `$sysroot/share/` because `$sysroot` already resembles my `/usr` directory and `share` is where a distro package would place these files.

jmesmon: > share is where a distro package would place these files

Not necessarily. Packages regularly put non-shared libraries (xml & conf) in `/lib` and `/usr/lib` (systemd, valgrind, etc).

Also, `rustlib` is already where all the target specific files go. Having a second directory makes a bit of trouble for us in https://github.com/meta-rust/meta-rust as we currently need to put libraries & binaries for `-cross` packages (cross compilers) into different bindirs & libdirs than libraries & binaries for `-native` packages (those built for the host system).

There is no (current) allowance for `-cross` share-dirs due to the presumption that they don't contain anything specific to the target of cross compilation.

Ericson2314:
Speaking of cross compiling, I think rustc should support separate `build` and `host` platforms (to use the autoconf terminology---which we may want to abandon as there's no need to ever bake in a `target` so 2 platforms suffices). This would certainly make `build.rs` easier to work with.

Also, while I'd like to do this ASAP, I don't think it should be stabilized ASAP as there is some interaction with [scenarios](https://internals.rust-lang.org/t/pre-rfc-a-vision-for-platform-architecture-configuration-specific-apis/).

Ericson2314:
Also, Now more than ever, it should be clear that that the target triple does not determine the target. So that we still by default name build directories by the target triple is prone for errors. I think there should at least be an option (if not the default) to instead normalize the config JSON, hash it, and use the hash instead.

jmesmon:
> Now more than ever, it should be clear that that the target triple does not determine the target. 

+1 to that.

>  I think there should at least be an option (if not the default) to instead normalize the config JSON, hash it, and use the hash instead.

As long as this can be overridden in cargo via a flag or env var (so higher level build systems can know where files end up without hashing things themselves) that would work for me.

Ericson2314:
> As long as this can be overridden in cargo via a flag or env var (so higher level build systems can know where files end up without hashing things themselves) that would work for me.

Quickly glancing over `--help` and the man page, I didn't see anything, but I swear this already existed.

jmesmon:
> I swear this already existed.

There is a env var to change the `target` directory (`CARGO_TARGET_DIR`, it's addition broke some of my makefiles that used a variable of the same name to track the deeper directory), but not (afaik) to change the `target/<target-name>` directory.

text/0000-target-bundles.md
eternaleye: The distinction is that `lib`/`${LIBDIR}` is for stuff specific to the architecture of the _installed package_ (i.e. the architecture rustc was compiled for), while `share`/`${DATADIR}` is for things independent of that. As the whole point of target vs. host in the GNU cross parlance is that they're independent, target specs really do belong in `share`.

text/0000-target-bundles.md
eternaleye: Does it sound sensible to include a hash of the target spec in emitted binaries? The issue goes beyond `std`, and including a spec hash would resolve it.

eternaleye:
I'd also like to chime in supporting the idea of using TOML rather than JSON, if we make incompatible changes. IMO, it'd make it considerably easier to work with targets - in particular, the built-in targets, well-commented, could themselves serve as half-decent documentation.

text/0000-target-bundles.md
jmesmon: > ${LIBDIR} is for stuff specific to the architecture of the installed package (i.e. the architecture rustc was compiled for)

Well, then should all of `rustlib/<target-name>` go under `${DATADIR}` instead of `${LIBDIR}`? They're not related to the architecture of the installed package.

I'd also consider that `gcc` is using `${LIBDIR}/gcc/<target-name>/<gcc-version>` to store it's target libraries & include files (at least, on arch linux they are), which could be seen as similar to our usage of `${LIBDIR}/rustlib`.

I guess my point is that target specifications and their associated compiled libraries are clearly related, and as a result I'd prefer they are stored "close" to each other. If `${DATADIR}` is the right place for target config, it's likely the right place for the target libs too.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+
+That being said, it might make sense to have a whitelist of stable options and cfg variables and
+keep everything else unstable for some duration.
+
+### Proposed JSON key-values
+
+```js
+{
+    // REQUIRED
+    "llvm-target": "x86_64-unknown-linux-gnu",              // LLVM target triple (does not need to match with rustc triple)
```
cardoe: Don't use dashes and instead use underscores so that you can just use `#[derive(RustcEncodable, RustcDecodable)]`. The dashes are the problem as I pointed out in https://github.com/rust-lang/rust/pull/32988

text/0000-target-bundles.md
cardoe: I've got the code for `rustc --print target-spec` at https://github.com/cardoe/rust/tree/target-spec I'll gladly resubmit it.

Ericson2314:
Actually, there might be a use for all 3 autoconfig-style platforms after all. "host" for the project becomes "target" for build scripts and procedural macros.

eternaleye:
@Ericson2314: Moreover, build scripts and procedural macros probably should be aware of `target` (consider build scripts that compile C, or procedural macros that need to care about `mem::size_of(usize)`)

text/0000-target-bundles.md
japaric: A hash of the target spec makes sense to me. I think it would have to be appended to the target triple of the rustlib directories: `$sysroot/lib/rustlib/$TARGET-$HASH`. That way several variants of the same target triple can coexist in the sysroot. It may also be worth to encode the hash in the rlibs' metadata for "extra protection"; someone might rename a rustlib directory from `$TARGET-$HASH1` to `$TARGET-$HASH2` -- the hash in metadata would prevent using the wrong variant in that case.

japaric:
> I'd also like to chime in supporting the idea of using TOML rather than JSON, if we make incompatible changes.

It might even be possible to provide a migration path for this: Support both `$TARGET.json` and `$TARGET.toml` where the TOML spec has precedence over the JSON spec. Warn when a JSON spec is used and point to the new TOML spec. After a few cycles, ignore the JSON spec. Though I'm not sure if it's worth the trouble.

text/0000-target-bundles.md
Ericson2314: Excellent, triple-hash solves the discoverability/ergonomics problem.

jmesmon:
> Though I'm not sure if it's worth the trouble.

Target specs have gotten members added (some mandatory) and removed as a regular occurrence, without warnings. I don't think there is really stability with the specs yet. I don't see it as a big deal to just remove the json spec support (though I say this as someone using target specs with release versions, not with nightlies, so adjusting to changes here isn't so bad for me).

text/0000-target-bundles.md
eternaleye: The thing with rustlib (and GCC's _compiled_ objects for cross architectures) is that they fall into a grey area: They are architecture specific (i.e. can't be _used_ on an architecture other than what they're compiled for) and thus don't belong in `${DATADIR}`, but aren't for the architecture of the _package_, and thus don't strictly belong in `${LIBDIR}`. This is not true of target _descriptions_, be they JSON or TOML.

This arises because the FHS doesn't account for architecture at all - on Exherbo, by contrast, _all_ compilation is cross compiling, and so there's a single global `${DATADIR}` at `/usr/share`, while each architecture has its own `${BINDIR}`, `${LIBDIR}`, and `${LIBEXECDIR}` under `/usr/${TRIPLE}` - the package manager, then, natively handles cross-compilation.

jmesmon: > This is not true of target descriptions, be they JSON or TOML.

I guess this is the point I disagree on. The main difference between the compiled object files and the target descriptions is just that one happens to be compiled, so people are more comfortable with them sitting in lib. The target descriptions are specific to a target, they are just parse-able outside of that target, but aren't useful outside of the context of using them for a specific target.

I'd like to re-iterate that configuration files for valgrind and systemd (among other things) are also placed in lib. These aren't compiled objects. They are as specific to the architecture as our target specification is (and perhaps somewhat less specific to it in some cases).

And gcc's headers in lib aren't compiled objects.

jmesmon: On thing to consider is that if we had `${rustlib}/<target-triple>/target.toml` (instead of putting the target name in the file name, we put the file within the folder with the target name, along with the compiled libraries).

text/0000-target-bundles.md
jmesmon: I don't think it makes sense to suffix the rustlib-target-dir with a hash. Consider that if we have a directory blessed to contain target specifications, and the name of those specifications is `<target-name>.ext`, we already know that all the _installed_ specfications have unique names.

If we're worried about handling targets that don't have specifications installed, why would they have `std` installed in `rustlib`? Instead, for handling non-installed target specs (those not in the blessed dir), I'd propose the following:

 - allow specification of the directory containing `std` crates either in the target-spec (brittle for dynamic uses, but potentially automatic) or as a seperate argument to rustc (easy for dynamic, but no automation)

Alternately (and I prefer this):

 - allow `--target` to accept a directory (so `--target ./my-target`) that contains a target spec (`./my-target/target.json` or `./my-target/target.toml`) and the `std` crates in the same way as `rustlib` today.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+## Distribution of targets
+
+Currently every built-in target is distributed along with the rustc compiler. This is suboptimal,
+because in majority of cases users are interested in targets which rustc can compile for, rather
+than the built-in targets rustc knows about, therefore distributing targets built-in into rustc is
+providing no benefits.
+
+Native libraries and linkers aside, it is obvious there’s little sense in distributing targets
```
japaric: I don't agree with this. Once "on the fly compilation of std" becomes a thing, it actually makes sense to provide all the target specifications "upfront" with `rustc`. Then people, specially the ones working with/on embedded systems, can choose which std features to enable/disable (e.g. remove RUST_BACKTRACE support) (\*) and/or compile std with CPU specific optimizations for maximum performance (\*\*). In this scenario, rust-std becomes a convenience to avoid the compile time required to build std the first time.

(\*) This actually makes more sense for final deployments/releases.
(\*\*) This makes sense for both development (slightly faster execution of test suites but the savings add up quickly) and for deployment to embedded systems. For releases that must work everywhere, one should instead not enable any CPU-specific optimization -- this would be the default, zero-configuration option, which is also how rust-std is compiled.

Instead of what's proposed here, I propose adding a new `targets` component that will be installed by `rustup` *once* regardless of which or how many toolchains (`rustc` components) one installs. After all, target specifications are indepedent of the host system. This also means one less step (`rustup target add foo`) in the initial cross compilation setup (once "on the fly compilation of std" is here).

jmesmon: Getting rid of precompiled `std` altogether isn't something I had considered in my other comments here. If we're sure that is going to happen (and as a result `${libdir}/rustlib/<target>/*` is going to go away as most places that isn't writable) then it definitely makes sense to distribute specs for all supported targets at once rather than requiring them to each be downloaded separately because the size of a `target` will just be the size of it's spec.

Ericson2314: > After all, target specifications are indepedent of the host system. 

They aren't independent of the compiler however, as the fields will probably be subject to change for a while.

japaric: You are right. Then we could use the same "channel" policy that the toolchain uses. If you do `rustup update nightly`, you get an updated `targets-nightly`; If you do `rustup default stable`, you get `targets-stable`; and so for.

text/0000-target-bundles.md
eternaleye: > If we're worried about handling targets that don't have specifications installed, why would they have std installed in rustlib?

That's really, really nasty if we put the hash of the target in the path to rustlib - you need the target to hash it to find the directory in which you can find the target. You wind up having to scan every directory, looking for the target you want.

text/0000-target-bundles.md
eternaleye: > On thing to consider is that if we had ${libdir}/rustlib/<target-triple>/target.toml (instead of putting the target name in the file name, we put the file within the folder with the target name, along with the compiled libraries).

I don't think that works very well; see https://github.com/rust-lang/rfcs/pull/1711#discussion_r74292019 for my reply to your other comment suggesting this.

jmesmon: > units can very well contain architecture-specific paths, thus making them architecture-dependent in exactly the way ${LIBDIR} captures.

Let me clarify my point in bringing up valgrind & systemd here:

 - both of these are host-architecture specific files (they don't have the same problem as gcc & ourselves)
 - both of the files provided are examinable (and in some sense usable in that information about the host-architecture can be extracted out of them).
 - but the information contained isn't usable except to "work with" the host-architecture.

I'm then considering our target spec bits:

 - target specs are examinable (and in some sense usable in that information about the target-architecture can be extracted out of them)
 - but the information contained isn't usable except to "work with" the target-architecture.

The key difference between the target specs & target libs is that one is "examinable" (in a traditional sense that I don't need to parse elf files :).

text/0000-target-bundles.md
jmesmon: >  it solves uniqueness at one point in time, but if the target spec is edited, you can now have a target dir that contains a mix of objects built under different target spec contents. That's unsafe.

Very true. But using a hash in a directory name as a way of avoiding that unsafety isn't perfect. For avoiding unsafety, we'd really want to embed the target hash in the object files.

> That's really, really nasty if we put the hash of the target in the path to rustlib 

Very true. Which is why the beginning of my comment discards the idea of using a hash.

Ericson2314: > embed the target hash in the object files.

We can do both. but remember that even if rustc can check rlib metadata to verify the hash, ld is not rust-specific and won't. For interopt with other tools it's still nice to get the safety of one directory / target.

text/0000-target-bundles.md
eternaleye: "Examinable" is a red herring, and I suspect _why_ is rooted in this:

> but the information contained isn't usable except to "work with" the target-architecture.

The crucial misstep is that "work with" isn't the constraint - it's "work on". If a systemd unit file reference an arch-dependent path (say, in `ExecStart`), that unit cannot be _used for its purpose_ unless you are _running_ that arch - i.e., can execute the program at that path, which was built for that arch, referred to in `ExecStart`.

Meanwhile, if I had a native rustc and a canadian-cross rustc, of the same versions, they could both use the exact same target specifications unchanged. This is exactly the property denoting what should go in `${DATADIR}`.

And before you then go back to saying this is true of rustlib, it's not - not in the face of dynamic linking, which is supported.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+    "lib_allocation_crate": "alloc_system",
+    "exe_allocation_crate": "alloc_jemalloc",
+    // should become templates? `lib{}.so` is much nicer
+    "dll_prefix": "lib",
+    "dll_suffix": ".so",
+    "exe_suffix": "",
+    "staticlib_prefix": "lib",
+    "staticlib_suffix": ".a",
+
+    // LLVM options
```
arielb1: Both this and "Various stuff adjusting how compilation is done"?

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+    "cpu": "x86-64",                                        // CPU of the target
+    "features": "",                                         // LLVM features
+    "relocation_model": "pic",
+    "code_model": "default",
+    "eliminate-frame-pointer": true,
+
+    // Apparently OS X is so much of a special case that even moving all the special cases into
+    // configuration is painful.
+    is_like_osx: false,
```
arielb1: Maybe this should be "linker_kind": "osx"?

nagisa: A nice alternative.

text/0000-target-bundles.md
```diff
@@ -0,0 +1,245 @@
...
+Currently there’s two different ways rustc targets are distributed: built-in targets and custom
+JSON targets. Built-in targets are very inflexible – they cannot be changed without changing and
+recompiling the compiler itself. On the other side of the spectrum, custom JSON targets are easy to
+adjust and adapt, but feature wise are very limited and rarely are suitable for the more uncommon
+use-cases.
+
+We’ve observed a considerable amount of desire by users of the language to customize targets they
+use in the ways currently not supported by our current infrastructure (sans making changes to the
+compiler itself, of course), and noted that the current scheme is not very feasible in the long
+run. This RFC should go a great amount towards fixing the issues.
```
eternaleye: Nit: "a great amount" -> "a long way"

text/0000-target-bundles.md
nagisa: I do not disagree with any points raised here, but I feel like solving this concern should become a part of eventual “custom std on demand” RFC, which AFAICT not even feasible at the moment due to stability concerns.

text/0000-target-bundles.md
nagisa: > I'm not sure about deriving target_pointer_witdth et al. from data-layout because we don't have control over its format/contents; it might stop containing information about the pointer width in the future.

We already use data-layout to calculate stuff like …ehem… layout in rustc, including sizing and alignment information of pointers. Even if data-layout format changed somehow to not contain (or have defaults for) this information anymore, that particular part of the compiler would need some way to have both pointer size *and* alignment specified. I’d argue that `cfg` stuff is not suitable for that, and we’d have to figure out a different way to specify such information.

I’m also not proposing to derive `cfg`s themselves from other data. If the key is not specified in this list, then it is not available for the crate to see. Rather, I’m saying that nowhere in rustc the `cfg`s should be used to figure out how to compile stuff.

> I also meant target_os

AFAICT there’s exactly two uses of this cfg in the compiler, both when deciding particulars of ABI, and these IMO should be handled by distinct keys. One of the uses is already handled by `c-abi-kind` key, the other use might need an extra key (since it is [a hack](https://github.com/rust-lang/rust/blob/099b9fdb1a170b57ffd7174b3c3042cc86b7fe91/src/librustc_trans/abi.rs#L319-L323)).

ubsan:
ping @nagisa 

status?

text/0000-target-bundles.md
```diff
@@ -0,0 +1,246 @@
...
+    // Configuration variables injected into the compilation units.
+    "cfg": {
+        "target_os": "linux",
+        "target_family": "unix",
+        "target_arch": "x86_64",
+        "target_endian": "little",
+        "target_pointer_width": "64",
+        "target_env": "gnu",
+        "target_vendor": "unknown",
+        "target_has_atomic": ["8", "16", "32", "64"],       // any of #[cfg(target_has_atomic={"8","16","32","64"}] work.
```
nagisa: @Amanieu previously complained about this being an explicit listing. See https://github.com/rust-lang/rust/pull/38579

nagisa:
@ubsan I’m considerably comfortable with current PR as it is written. I still think it makes a lot of sense to bundle libraries with their target specs. 

There has been some concerns raised about future std-on-demand RFC, but it seems to me like concern is better handled in the future RFC. I do not see anything in this RFC that would prohibit on-demand compilation of std or even core.

All in all, this RFC is mostly waiting for somebody to shepherd this and move the RFC into FPC for either acception or rejection.

Ericson2314:
As the author of std-aware cargo, I see no issue with this.

alexcrichton:
Overall this RFC seems great to me, it looks like a fantastic cleanup of the current json target specs and should also simplify some compiler internals by making them a little more agnostic to platforms. 

It's also worth pointing out that since this RFC was opened the internal targets are all "JSON-ified" in the sense that the compiler only consumes them through the JSON format. They're not literally stored in JSON yet (to preserve comments), but we should be able to easily convert them over.

One thing I've noticed is that over time we continually add `is_like_foo` keys to the target JSON, typically because one platform just sticks out from the rest. @nagisa do you feel that we'd avoid this and instead add a "proper" configuration targeted at the exact use case of where you'd otherwise write `if is_like_foo`? That means that we'll probably want some story for the default for each key as it may be common to hit specs without keys specified as the compiler evolves.

The final point of this RFC, distributing target specs with libraries, I may still need convincing for. The RFC claims:

> Currently every built-in target is distributed along with the rustc compiler. This is suboptimal, because in majority of cases users are interested in targets which rustc can compile for, rather than the built-in targets rustc knows about, therefore distributing targets built-in into rustc is providing no benefits.

I'd personally argue that there's no *downside* to the current scheme, and the drawback of a different scheme is that it will take time and effort to implement. I also agree with @japaric's [sentiments](https://github.com/rust-lang/rfcs/pull/1711#r74173303) that this may be useful in the future.

Is this part of the RFC, distributing target specs with libraries, crucial to the RFC? It seems to me like we've got a lot to gain from cleaning up the format (I'm a fan of TOML too) and making them less `is_like_*` and more `real_option: ...`.

nagisa:
> do you feel that we'd avoid this and instead add a "proper" configuration targeted at the exact use case of where you'd otherwise write if is_like_foo?

Yes, that’s one of the goals by the new schema proposed by this RFC. (note the lack of the `is_like` keys in it).

> That means that we'll probably want some story for the default for each key as it may be common to hit specs without keys specified as the compiler evolves. 

Since we already have `is_like_unix: true` and `is_like_everything_else: false` by default, one option is to keep those keys’ defaults as applicable to unix.

> Is this part of the RFC, distributing target specs with libraries, crucial to the RFC? It seems to me like we've got a lot to gain from cleaning up the format (I'm a fan of TOML too) and making them less is_like_* and more real_option: ....

There’s one counter-point I’d like to make here. I feel that just cleaning up and converting the target specs all by themselves doesn’t necessarily deserve a full RFC – it is mostly an implementation detail of the rustc compiler. It is also easy to support the old custom target specs for a while alongside the new ones to provide a migration path. In that sense, my original intent with this RFC was more to propose the idea of distributing target specs with the libcore/libstd and less the target spec refactor.

I was thinking about working the target spec refactor on the weekends for a while now, and I feel like it would be able to land it just fine with a PR FCP (those didn’t exist back when I wrote the RFC either) for tools and compiler teams.

alexcrichton:
Ok that sounds reasonable to me, and yeah for defaults I think we'd just need to be proactive about choosing a good default to break as few targets as possible, but in theory the targets that would be broken are esoteric regardless and as @jmesmon [mentions](https://github.com/rust-lang/rfcs/pull/1711#issuecomment-238682600) custom targets aren't really *that* stable today anyway.

I'd also be ok yeah with just a PR to modify these pieces of the compiler (and rework the target specs a bit). If you're ok with that and separating out the distribution of the target specs for now perhaps we can close this RFC?

PR #1712, opened 2016-08-11
Add 'else match' blocks to if expressions.

phoenixenero:
A more specific version of my previous proposal.

### DIFF
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -1,0 +1,228 @@
+- Feature Name: Else Match
+- Start Date: 2016-07-26
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+Extend the `if` expression to accept an `else match` block.
+
+# Motivation
+[motivation]: #motivation
+
+This proposal is meant to reduce the verbosity of writing `if ... else { match ... } ` style
+statements.
+
+It also makes code similar to:
+
+```rs
+let flag = false;
+
+if foo() {
+    do_this();
+} else {
+    match bar() {
+        baz() => do_that(),
+        _ => flag = true
+    }
+}
+
+if flag {
+    do_something();
+}
+```
+
+simpler and more concise:
+
+```rs
+if foo() {
+    do_this();
+} else match bar() {
+    baz() => do_that()
+} else {
+    do_something();
+}
+```
+
+## Use-cases
+
+Though rare, this pattern does exist in several Rust projects.
+
+### [Servo](https://github.com/servo/servo)
+
+<https://github.com/servo/servo/blob/master/components/layout/model.rs#L557>
+
+Before:
+
+```rs
+/// Clamp the given size by the given `min` and `max` constraints.
+pub fn clamp(&self, other: Au) -> Au {
+    if other < self.min {
+        self.min
+    } else {
+        match self.max {
+            Some(max) if max < other => max,
+            _ => other
+        }
+    }
+}
+```
+
+After:
+
+```rs
+/// Clamp the given size by the given `min` and `max` constraints.
+pub fn clamp(&self, other: Au) -> Au {
+    if other < self.min {
+        self.min
+    } else match self.max {
+        Some(max) if max < other => max,
+        _ => other
+    }
+}
+```
+
+### [xsv](https://github.com/BurntSushi/xsv)
+
+<https://github.com/BurntSushi/xsv/blob/master/src/cmd/stats.rs#L311>:
+
+Before:
+
+```rs
+if !self.typ.is_number() {
+    pieces.push(empty()); pieces.push(empty());
+} else {
+    match self.online {
+        Some(ref v) => {
+            pieces.push(v.mean().to_string());
+            pieces.push(v.stddev().to_string());
+        }
+        None => { pieces.push(empty()); pieces.push(empty()); }
+    }
+}
+```
+
+After:
+
+```rs
+if !self.typ.is_number() {
+    pieces.push(empty()); pieces.push(empty());
+} else match self.online {
+    Some(ref v) => {
+        pieces.push(v.mean().to_string());
+        pieces.push(v.stddev().to_string());
+    }
+    None => { pieces.push(empty()); pieces.push(empty()); }
+}
+```
+
+### [trust-dns](https://github.com/bluejekyll/trust-dns)
+
+<https://github.com/bluejekyll/trust-dns/blob/master/src/authority/authority.rs#L558>:
+
+Before:
+
+```rs
+if class == self.class {
+  match rr.get_rr_type() {
+    RecordType::ANY | RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
+    _ => (),
+  }
+} else {
+  match class {
+    DNSClass::ANY => {
+      if rr.get_ttl() != 0 { return Err(ResponseCode::FormErr) }
+      if let &RData::NULL(..) = rr.get_rdata() { () }
+      else { return Err(ResponseCode::FormErr) }
+      match rr.get_rr_type() {
+        RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
+        _ => (),
+      }
+    },
+    DNSClass::NONE => {
+      if rr.get_ttl() != 0 { return Err(ResponseCode::FormErr) }
+      match rr.get_rr_type() {
+        RecordType::ANY | RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
+        _ => (),
+      }
+    },
+    _ => return Err(ResponseCode::FormErr),
+  }
+}
+```
+
+After:
+
+```rs
+if class == self.class {
+  match rr.get_rr_type() {
+    RecordType::ANY | RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
+    _ => (),
+  }
+} else match class {
+  DNSClass::ANY => {
+    if rr.get_ttl() != 0 { return Err(ResponseCode::FormErr) }
+    if let &RData::NULL(..) = rr.get_rdata() { () }
+    else { return Err(ResponseCode::FormErr) }
+    match rr.get_rr_type() {
+      RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
+      _ => (),
+    }
+  },
+  DNSClass::NONE => {
+    if rr.get_ttl() != 0 { return Err(ResponseCode::FormErr) }
+    match rr.get_rr_type() {
+      RecordType::ANY | RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
+      _ => (),
+    }
+  },
+} else {
+  return Err(ResponseCode::FormErr)
+}
+```
+
+# Detailed design
+[design]: #detailed-design
+
+## Grammar
+
+See the following document for an (incomplete) guide to the grammar used in Rust:
+[Rust Documentation → Grammar](https://doc.rust-lang.org/grammar.html).
+
+This proposal modifies the
+[if expression grammar](https://doc.rust-lang.org/grammar.html#if-expressions).
+
+```
+else_tail : "else" [ if_expr | if_let_expr
++                  | match_expr
+                   | '{' block '}' ] ;
+```
+
+## Execution
+
+An `else match` block should be treated similar to an `else` block with a single `match`
+expression, with one key addition:
+
+When an arbitrary expression `exp` in `else match <exp>` fails to match any of the cases in the
+`else match` block, the next block (if it exists) is run.
+
+### Dead code
+
+An `else match` block with a `_ => ...` match means the next clause (if exists) will never run.
+There might be more complicated cases to optimize for, and is outside the scope of this document.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Slight maintainability problems whenever you need to add additional logic to an `else` block.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Don't do this.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None.
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -210,7 +210,7 @@ When an arbitrary expression `exp` in `else match <exp>` fails to match any of t
 ### Dead code
 
 An `else match` block with a `_ => ...` match means the next clause (if exists) will never run.
-There might be more complicated cases to optimize for, and is outside the scope of this document.
+There might be more complicated cases to optimize for and is outside the scope of this document.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -11,14 +11,9 @@ Extend the `if` expression to accept an `else match` block.
 # Motivation
 [motivation]: #motivation
 
-This proposal is meant to reduce the verbosity of writing `if ... else { match ... } ` style
-statements.
-
-It also makes code similar to:
+This proposal is meant to reduce the verbosity of writing `if ... else { match ... } ` expressions.
 
 ```rust
-let flag = false;
-
 if foo() {
     do_this();
 } else {
@@ -26,10 +21,6 @@ if foo() {
         baz() => do_that(),
         _ => flag = true
     }
-}
-
-if flag {
-    do_something();
 }
 ```
 
@@ -40,48 +31,12 @@ if foo() {
     do_this();
 } else match bar() {
     baz() => do_that()
-} else {
-    do_something();
 }
 ```
 
 ## Use-cases
 
 Though rare, this pattern does exist in several Rust projects.
-
-### [Servo](https://github.com/servo/servo)
-
-<https://github.com/servo/servo/blob/master/components/layout/model.rs#L557>
-
-Before:
-
-```rust
-/// Clamp the given size by the given `min` and `max` constraints.
-pub fn clamp(&self, other: Au) -> Au {
-    if other < self.min {
-        self.min
-    } else {
-        match self.max {
-            Some(max) if max < other => max,
-            _ => other
-        }
-    }
-}
-```
-
-After:
-
-```rust
-/// Clamp the given size by the given `min` and `max` constraints.
-pub fn clamp(&self, other: Au) -> Au {
-    if other < self.min {
-        self.min
-    } else match self.max {
-        Some(max) if max < other => max,
-        _ => other
-    }
-}
-```
 
 ### [xsv](https://github.com/BurntSushi/xsv)
 
@@ -177,8 +132,7 @@ if class == self.class {
       _ => (),
     }
   },
-} else {
-  return Err(ResponseCode::FormErr);
+  _ => return Err(ResponseCode::FormErr),
 }
 ```
 
@@ -194,28 +148,27 @@ This proposal modifies the
 [if expression grammar](https://doc.rust-lang.org/grammar.html#if-expressions).
 
 ```
-else_tail : "else" [ if_expr | if_let_expr
-+                  | match_expr
-                   | '{' block '}' ] ;
+ else_tail : "else" [ if_expr | if_let_expr
++                   | match_expr
+                    | '{' block '}' ] ;
 ```
 
 ## Execution
 
-An `else match` block should be treated similar to an `else` block with a single `match`
-expression, with one key addition:
-
-When an arbitrary expression `exp` in `else match <exp>` fails to match any of the cases in the
-`else match` block, the next block (if it exists) is run.
+An `else match` block should be treated exactly the same as an `else` block with a single `match`
+expression.
 
 ### Dead code
 
-An `else match` block with a `_ => ...` match means the next clause (if exists) will never run.
-There might be more complicated cases to optimize for and is outside the scope of this document.
+The next `else` block after an `else match` is never run. This is because `match` itself does not
+take on an `else` block. Whether `match` should allow an `else` or not should be addressed in a
+separate proposal.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 - Slight maintainability problems whenever you need to add additional logic to an `else` block.
+- Can be more of a stylistic issue than an ergonomics issue.
 
 # Alternatives
 [alternatives]: #alternatives
@@ -224,17 +177,18 @@ Not an alternative but an addition to the proposal: `if match` expressions. This
 additional grammar rule and modify an existing one:
 
 ```
++ if_match_expr : "if" match_expr else_tail ? ;
+```
+
+```
  expr : literal | path | tuple_expr | unit_expr | struct_expr
       | block_expr | method_call_expr | field_expr | array_expr
       | idx_expr | range_expr | unop_expr | binop_expr
       | paren_expr | call_expr | lambda_expr | while_expr
       | loop_expr | break_expr | continue_expr | for_expr
       | if_expr | match_expr | if_let_expr | while_let_expr
-~     | if_match_expr | return_expr ;
-
-...
-
-+ if_match_expr : "if" match_expr else_tail ? ;
+-     | return_expr ;
++     | if_match_expr | return_expr ;
 ```
 
 Should work nearly the same as `else match`.
@@ -242,4 +196,5 @@ Should work nearly the same as `else match`.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-None.
+Whether to allow `match` to take on an `else` block. This should be addressed
+in a separate proposal.
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -33,6 +33,7 @@ if foo() {
     do_this();
 } else match bar() {
     baz() => do_that()
+    _ => flag = true
 }
 ```
 
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -220,15 +220,21 @@ There might be more complicated cases to optimize for and is outside the scope o
 # Alternatives
 [alternatives]: #alternatives
 
-Not an alternative but an addition to the proposal: `if match` expressions. This would modify the
-grammar as so:
+Not an alternative but an addition to the proposal: `if match` expressions. This would add an
+additional grammar rule and modify an existing one:
 
 ```
-+ if_match_expr : "if" match_expr else_tail ? ;
+ expr : literal | path | tuple_expr | unit_expr | struct_expr
+      | block_expr | method_call_expr | field_expr | array_expr
+      | idx_expr | range_expr | unop_expr | binop_expr
+      | paren_expr | call_expr | lambda_expr | while_expr
+      | loop_expr | break_expr | continue_expr | for_expr
+      | if_expr | match_expr | if_let_expr | while_let_expr
+~     | if_match_expr | return_expr ;
 
-else_tail : "else" [ if_expr | if_let_expr
-+                  | if_match_expr | match_expr
-                   | '{' block '}' ] ;
+...
+
++ if_match_expr : "if" match_expr else_tail ? ;
 ```
 
 Should work nearly the same as `else match`.
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -13,6 +13,8 @@ Extend the `if` expression to accept an `else match` block.
 
 This proposal is meant to reduce the verbosity of writing `if ... else { match ... } ` expressions.
 
+Code such as:
+
 ```rust
 if foo() {
     do_this();
@@ -24,7 +26,7 @@ if foo() {
 }
 ```
 
-simpler and more concise:
+can be simpler and more concise:
 
 ```rust
 if foo() {
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -16,7 +16,7 @@ statements.
 
 It also makes code similar to:
 
-```rs
+```rust
 let flag = false;
 
 if foo() {
@@ -35,7 +35,7 @@ if flag {
 
 simpler and more concise:
 
-```rs
+```rust
 if foo() {
     do_this();
 } else match bar() {
@@ -55,7 +55,7 @@ Though rare, this pattern does exist in several Rust projects.
 
 Before:
 
-```rs
+```rust
 /// Clamp the given size by the given `min` and `max` constraints.
 pub fn clamp(&self, other: Au) -> Au {
     if other < self.min {
@@ -71,7 +71,7 @@ pub fn clamp(&self, other: Au) -> Au {
 
 After:
 
-```rs
+```rust
 /// Clamp the given size by the given `min` and `max` constraints.
 pub fn clamp(&self, other: Au) -> Au {
     if other < self.min {
@@ -89,7 +89,7 @@ pub fn clamp(&self, other: Au) -> Au {
 
 Before:
 
-```rs
+```rust
 if !self.typ.is_number() {
     pieces.push(empty()); pieces.push(empty());
 } else {
@@ -105,7 +105,7 @@ if !self.typ.is_number() {
 
 After:
 
-```rs
+```rust
 if !self.typ.is_number() {
     pieces.push(empty()); pieces.push(empty());
 } else match self.online {
@@ -123,7 +123,7 @@ if !self.typ.is_number() {
 
 Before:
 
-```rs
+```rust
 if class == self.class {
   match rr.get_rr_type() {
     RecordType::ANY | RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
@@ -154,7 +154,7 @@ if class == self.class {
 
 After:
 
-```rs
+```rust
 if class == self.class {
   match rr.get_rr_type() {
     RecordType::ANY | RecordType::AXFR | RecordType::IXFR => return Err(ResponseCode::FormErr),
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -178,7 +178,7 @@ if class == self.class {
     }
   },
 } else {
-  return Err(ResponseCode::FormErr)
+  return Err(ResponseCode::FormErr);
 }
 ```
 
@@ -220,7 +220,18 @@ There might be more complicated cases to optimize for and is outside the scope o
 # Alternatives
 [alternatives]: #alternatives
 
-Don't do this.
+Not an alternative but an addition to the proposal: `if match` expressions. This would modify the
+grammar as so:
+
+```
++ if_match_expr : "if" match_expr else_tail ? ;
+
+else_tail : "else" [ if_expr | if_let_expr
++                  | if_match_expr | match_expr
+                   | '{' block '}' ] ;
+```
+
+Should work nearly the same as `else match`.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions


text/0000-else-match.md
```diff
@@ -0,0 +1,245 @@
...
+simpler and more concise:
+
+```rust
+if foo() {
+    do_this();
+} else match bar() {
+    baz() => do_that()
+} else {
+    do_something();
+}
```
oli-obk: with single variant matches `if let` is definitely more readable:

```rust
enum Bazz {
    Baz,
    Bak,
}

fn foo() -> bool { unimplemented!() }
fn baz() -> Bazz { unimplemented!() }
fn do_this() -> () { unimplemented!() }
fn do_that() -> () { unimplemented!() }
fn do_something() -> () { unimplemented!() }

fn main() {
    if foo() {
        do_this();
    } else if let Bazz::Baz = baz() {
        do_that();
    } else {
        do_something();
    }
}
```

louy2:
I am not convinced this is necessary unless more useful optimization can be done against it. It seems pure aesthetic now.

text/0000-else-match.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+## Use-cases
```
ticki: To be honest, in no of these cases I see a clear improvement. In fact, the opposite is true. I find it "uglier".

text/0000-else-match.md
```diff
@@ -0,0 +1,245 @@
...
+```rust
+/// Clamp the given size by the given `min` and `max` constraints.
+pub fn clamp(&self, other: Au) -> Au {
+    if other < self.min {
+        self.min
+    } else match self.max {
+        Some(max) if max < other => max,
+        _ => other
+    }
+}
```
killercup: No, I'm sorry, to me, that doesn't look good at all. If you really want to get rid of that additional level of indentation, just use an early return in the if branch (works for the next example as well).

camlorn:
If I was going to do anything in this regard, it would be to let `else if` and `else if let` be valid without braces.  I don't much see the utility of this either.  It seems like `else if` would be more useful, assuming that's not allowed already (I'm assuming it's not because I've never seen it, but haven't tried).

`else if let` would let you match multiple arms as well, though without the exhaustion checks:

```
if flag {
    do_something();
}
else if let pattern {
    do_something_else();
}
else if let other_pattern {
    do_a_third_thing();
}
```

Etc.  While this has the downside of not getting the compiler to check for you, it also looks good in the case where you need to match patterns against multiple, different values; the one the RFC proposes would require indentation for this idiom.

It might be worth expanding the RFC with this, or closing the RFC and making another along these lines, but something like the above seems much more generally useful.

text/0000-else-match.md
```diff
@@ -0,0 +1,245 @@
...
+                   | '{' block '}' ] ;
+```
+
+## Execution
+
+An `else match` block should be treated similar to an `else` block with a single `match`
+expression, with one key addition:
+
+When an arbitrary expression `exp` in `else match <exp>` fails to match any of the cases in the
+`else match` block, the next block (if it exists) is run.
```
withoutboats: I am in favor of `else match`, but I am really against this! `match` should work consistently.

There is an orthogonal proposal to allow match to take `else`s at the end, so you don't have to include the catchall branch inside the match intend. But that proposal should apply to all `match` statements, not only the `else match` ones.

withoutboats:
Integrating `match` and `if let` (as @camlorn notes) into the if/else construct seems obvious and normalizing. As I commented on a line note, it also make sense for `match` to take `else` in general. In my opinion these changes would make the control flow constructs of the language more consistent in how they work.

nrc:
@camlorn both `else if` and `else if let` are already accepted (although you need `... pattern = expr` in your example to parse).

This RFC somewhat makes sense to me since we allow `else if let`, but I'm not completely convinced of the value here. `if ... else if ... else` seems a very common pattern. However, I can't imagine that `if ... else match` would be so common. It is also not clear that it makes sense since `else match` can't be chained, e.g., `if ... else match ... else` doesn't make sense unless we relax the requirement for `match`s to be exhaustive.

glaebhoerl:
Instead of special-casing `match` I would rather extend this to any control-flow construct which has a block attached. This rule would naturally encompass the three specific constructs which are already allowed after an `else`: plain blocks `{ ... }`, `if ... { ... }`, and `if let ... { ... }`. So instead of adding more special cases, it would result in fewer.

withoutboats:
@glaebhoerl you mean that this RFC should be written to support `else loop`, `else while`, `else for`?

glaebhoerl:
I think there should be a syntactic category of "things that have blocks", and anything in that category should be allowed. So yes those, and also `catch` once it's implemented, and `unsafe`, and anything else that might be slipping my mind. But the point would be to uniformly distinguish on the basis of "does it have a block or not" instead of a laundry-list of special cases.

(Not necessarily *this* RFC. Just stating my preference in general.)

phoenixenero:
Personally, for `match` statements I can agree on because match is basically a glorified if-else-if/switch structure. Loop, for, while, unsafe, etc. belong in a different "class" of structure that does not deal with Boolean logic (and pattern matching in general) and would be confusing. Allowing only for `match` seems less prone to mistaken intuition.

nikomatsakis:
I've wanted to propose this for a while. This pattern seems to come up often for me.

withoutboats:
@glaebhoerl I think "having blocks" is just a syntactic category and there is some semantics to this. Basically I think it should make sense as a 'branching' construct. This could make sense for loops if loops ever evaluate to values, and I have to think about `catch` more, but for `unsafe` I don't think it makes sense because `unsafe` doesn't have anything to do with control flow. I also think its good to make an exception for `unsafe` for the reason that we want that keyword to be extremely visible, and this would make it easier to miss that this else block is actually also an unsafe block.

glaebhoerl:
> I think "having blocks" is just a syntactic category and there is some semantics to this. Basically I think it should make sense as a 'branching' construct.

To be honest, I don't think so. The way I look at it is, what's the core requirement? The core requirement is to avoid the "`goto fail`" class of bugs, where something can be visually mistaken for being part of the `else` clause, when it is not. What do we need to satisfy this requirement? We need that where there is the possibility to add an additional statement after a statement belong to an `else` clause, that that statement must be enclosed in braces. And my disposition would be to just satisfy the requirement in the simplest way possible without getting other perceived distinctions mixed up in it. I mean, C and most C-family languages allow *anything* to come after an `else`, "branching construct" or not, block-having or not. Relative to C the only difference we really have is that we want to avoid `goto fail`.

> I also think its good to make an exception for `unsafe` for the reason that we want that keyword to be extremely visible, and this would make it easier to miss that this else block is actually also an unsafe block.

This *might* make some sense for `unsafe` in particular. Though requiring `unsafe` and only `unsafe` to always use double-nesting in this case also feels a bit like the "make `unsafe` code unergonomic to write so people are discouraged from writing it" philosophy of design which we've tried to avoid following.

mdinger:
This type of idea was previously discussed [here](https://internals.rust-lang.org/t/match-fn-sugar/431). As an alternative, if all you're trying to do is reduce rightward drift for matches, https://github.com/rust-lang/rfcs/issues/1745 might be interesting.

nikomatsakis:
@rfcbot fcp merge

This is a small bit of syntactic sugar, but it seems harmless enough, and I know that I have found myself wanting it many times. It frequently happens because I have some guards I need to execute all the time, and then I want to do a bunch of matching in the general case. So I propose we merge it. I would start small: just `else match`, basically on the premise that `match` and `if` are generalizations of one another.

withoutboats:
@nikomatsakis how would you feel about allowing match to carry an else, which would be equivalent to a `_ =>` arm?

```rust
match an_option {
    Some(foo) => { ... }
} else {
    ...
}
```

It could take an if, which would be the same as a guard:


```rust
match an_option {
    Some(foo) => { ... }
} else if a_bool {
    ...
} else {
    ...
}
```

I don't feel particularly strongly that this is good style, but if we're allowing `if { } else match { }` it seems more consistent to allow `match { } else if { }`

nagisa:
I would like to note that this RFC generally has more of the negative feedback than it has of positive. I personally find that a sign that it ought to be moving towards `fcp close` regardless of the harmlessness, smallness or personal experiences of the lead of most of the teams. 

I’m personally fine with this getting merged *provided* we also investigate making things like `else loop`, `else while`, `else for` and so forth (hence the :/ reaction).

---

I very much dislike the idea of `else` after `match`.

nikomatsakis:
@nagisa 

> I’m personally fine with this getting merged provided we also investigate (now, as part of accepting this RFC) making things like else loop, else while, else for and so forth (hence the :/ reaction), even if we are not planning of including any of these as part of this RFC.

What do you mean by "investigate" here? (Is there something to figure out, other than just ... would we maybe want that? Like, some sort of parsing ambiguity?)

nagisa:
@nikomatsakis 

> What do you mean by "investigate" here? (Is there something to figure out, other than just ... would we maybe want that? Like, some sort of parsing ambiguity?)

Yes, the investigation answering “would we like that sort of syntax as the extension to this proposal”. To me it is important that the language is consistent. I can deal with only `if { ... } else if { ... }` working at the moment by thinking of it as if it were `if { ... } elif { ... }` instead (i.e. `else if` is a distinct construct from both `else` and `if`).

If I think about being able to do `else match`, my head infers that I ought to be able to also be able to `else loop` (especially with the loop-break-value thing accepted recently), `else while`, `else for`, and perhaps even `else EXPR` in general.

Sadly, to me some of those constructs (`else while`, `else loop`) seem kinda useless because they restrict the type of the whole branch to `()`. Refactoring an `if { ... } else while cond { ... }` to have some other type than `()` would involve an extra menial change of adding `{}` around the `while`, which I see as a significant loss in ergonomics.

nikomatsakis:
@nagisa 

> To me it is important that the language is consistent.

Agreed! To me, I guess, supporting `match` is consistent -- if and match "feel" closely related to me -- but give than you're not alone in raising this question it seems like we ought indeed to answer the question.

> Sadly, to me, some of those constructs (else while, else loop) seem kinda useless because they restrict the type of the whole conditional to ().

Hmm, I think this is not quite correct in two ways. First, it still seems useful to be able to do things like:

```rust
if let Some(data) = have_data {
      process(data);
} else for data in channel {
     process(data);
}
```

but also, given that one can now `break` with a value, there is still the option of 

```rust
if foo {
    value
} else loop {
    break other_value
}
```

Those examples are pretty interesting! They feel...unfamiliar to me, but maybe cool.

withoutboats:
I've thought about it and @nagisa 's extension makes sense to me. Like @nagisa, my concern with extensions like this is primarily that the language be consistent and discoverable.

Though I sort of expect `else TERM { }` to imply `TERM { } else { }` is valid, its very difficult to extend this practically to loops (and extending it to matches requires changing their behavior in a big way). I think just allowing `else` to take any control flow block type makes sense.

nrc:
So, I've approved this, but not strongly. The various other extensions proposed here (`else for`, `match else`, etc.) all seem a bit alien and weird. All this proposal saves us is a set of braces and personally I haven't encountered the pattern often. I agree that it is probably worth having a look at our control flow language as a whole and thinking about how we would like it to look, rather than considering very minor changes like this in isolation.

glandium:
FWIW, as a user, `else TERM {}`, to me, just is a convenient shortcut for `else { TERM { } }` where the extra brackets are not adding anything useful. Making `else TERM` work, to me, raises more questions about why brackets are needed for one-liners than why `TERM {} else` is not supported. (And I'm not arguing that their requirement be lifted, merely pointing that `TERM {} else` sounds like a stretch)

Sherif89:
> I agree that it is probably worth having a look at our control flow language as a whole and thinking about how we would like it to look, rather than considering very minor changes like this in isolation.

I agree. If you were going to add this expression, you may add all other possible ones as well.

aturon:
Apologies for the slow response.

I've read through the thread, and feel particularly persuaded by @glaebhoerl:

> Instead of special-casing `match` I would rather extend this to any control-flow construct which has a block attached. This rule would naturally encompass the three specific constructs which are already allowed after an `else`: plain blocks `{ ... }`, `if ... { ... }`, and `if let ... { ... }`. So instead of adding more special cases, it would result in fewer.

a point echoed by @glandium:

> FWIW, as a user, `else TERM {}`, to me, just is a convenient shortcut for `else { TERM { } }` where the extra brackets are not adding anything useful.

As [@nikomatsakis points out](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-260743501), new combinations like `else loop` and `else for` actually seem natural and potentially useful. If there's a consistent story to be told syntactically, let's tell it at that level.

====

Overall, I think that there is indeed a paper cut being addressed here, and one worth tackling in some fashion. I'd prefer to extend `else` with the full syntactic class @glaebhoerl mentions (so: `else loop`, `else unsafe`, etc), since I buy the argument that this leads to an overall simplification of the language. I'm not particularly worried about such additions being confusing, since the block is still always delimited.

But what about further extensions? 

- [@withoutboats mentiones `match .. else`](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-260075431). The rationale, as I understand it, is to make conditional control flow more consistent by always allowing an `else` branch. But I think the argument for consistency there is more flimsy than for extending the existing `else if`-like cases. For one, the fact that `match` is currently intended to be exhaustive means this is a deeper change. I'm not necessarily opposed, but to me this seems like a step that can and should be considered separately.

- You could imagine similar extensions to other constructs that currently require a block. For example, `loop match { .. }` or `while EXPR unsafe { .. }`. Assuming this doesn't run into parsing trouble, should we consider such additional extensions? On the one hand, they seem to have much the same flavor as the `else if` case. OTOH, unlike with `else if`, we don't currently allow *any* such combinations. FWIW, some of these combinations definitely do come up; `loop { match { .. }}` being one I've written a handful of times.

Interestingly, extensions of `else` and `loop` feel better to me than cases like `for` -- I think because `else` and `loop` otherwise stand on their own as very short keywords, so it'd be hard to miss additional control flow operations that follow them.

====

In summary:

- I'm open to trying out extensions to `else`.
  - I'd strongly prefer to include *all* such syntactic extensions from the start. 
- Adding `else` to `match` seems trickier and I'd like to postpone that. 
- Extensions to `loop`, `while` etc, I'm unsure about -- but I imagine they could/should be informed by experience with `else` first.

So I'd sign off on this RFC if it was updated to include all brace-taking forms.

glaebhoerl:
I agree that `loop match` also seems worthwhile, and also that it seems like material for a separate RFC. 

Just for the record, does *that* feature have any obvious dimension along which it could/should be generalized? Are there any other values of `CONSTRUCT` for which you wind up writing `loop { CONSTRUCT { ... } }` besides `match`?

birkenfeld:
I'd be careful about allowing `loop if ... { ... }` - for one, it might not announce clearly what it does, and also it would mean having
```
loop if {
} else {
}
```
which makes it easy to overlook that it `loop`s.

As for `while expr CONSTRUCT`, it looks pretty odd because of the paren-less-ness of `expr`.

nikomatsakis:
Huh, I find `loop match` and `loop if` very confusing. Can someone give an example where this would be a useful thing? I guess it'd be useful in a circumstance similar to `while let Some(x) = option.next() { ... }`?

In any case, I'm ok with @glaebhoerl more extended version of `else <control-flow-keyword>`, though I don't consider it a *more* obvious contour than just supporting `else match`. That is, both seem like ok places to draw the line, and for different reasons. (Specifically, I consider `if` to be syntactic sugar for `match` so it seems to be natural to support them interacting in various ways, whereas loops and other things feel distinct to me.)

aturon:
@nikomatsakis 

One example for `loop match` is state machines. It's not so common that I'd consider it independently motivated, but I could get behind it as part of a general move toward consistency.

text/0000-else-match.md
```diff
@@ -0,0 +1,203 @@
...
+- Can be more of a stylistic issue than an ergonomics issue.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Not an alternative but an addition to the proposal: `if match` expressions. This would add an
+additional grammar rule and modify an existing one:
+
+```
++ if_match_expr : "if" match_expr else_tail ? ;
```
pnkfelix: I'm trying to understand the semantics of the construct being proposed in this alternative.

Is the idea of this alternative extended proposal that the new `if match` form would provide a way to describe a (non-exhaustive) match that has an (non-optional) `else` attached? (And if the given match is exhaustive, then the attached else becomes optional and in fact pointless?)

E.g. are you suggesting something like this:

```rust
enum Cases { Case1, Case2, Case3 }
if match Case2 { Case1 => 1, Case2 => 2 } else { 3 }; // => 2
if match Case3 { Case1 => 1, Case2 => 2 } else { 3 }; // => 3
```

I know this is only an alternative so it may not seem as important to tie down the described semantics; but I would like to understand the proposal so that I can evaluate the main text in the appropriate context.

burdges:
All these `{ }` fighting RFCs make me wonder if people are really fighting `rustfmt` and the ideology behind it. 

An alternative, or complementary, move might be tweaking `rustfmt` so that
```
loop { match ... {
    ...
} }
```
was kept as is, or even slightly preferred over the right drifting
```
loop {
    match ... {
    ...
    } 
}
```
After all, if you just worry about `{ }` then house styles can resurrect the rightward drift like
```
loop
    match ... {
    ...
    } 
```
In this vein, one could imagine `rustfmt` giving these blocks lacking `{ }` two extra spaces, so maybe
```
loop match {
      ...
  }
```

As an aside, I think the general form of this blocks approach winds up with some amusing nested loops
```
'a loop 'b loop 'c loop { 
    ... logic with various breaks to various levels ...
}
``` 
and 
```
loop if .. { 
    ... break;
} else loop if .. { 
    ... break;
} else loop if .. {
    ... break;
} else { 
    ...
}
```
I'd wager whatever the person writing such code might be trying to do, they would benefit more from `?` being able to handle looping monads, not that I have any proposal for doing that.  :)

leodasvacas:
@burdges A good place to make your argument would be rust-lang-nursery/fmt-rfcs#34.

text/0000-else-match.md
pnkfelix: @rfcbot concern what-the-heck-is-if-match

pnkfelix:
@rfcbot concern what-the-heck-is-if-match

This is transcribed from a comment I left on the diff (apparently rfcbot is not monitoring the comments there)

I'm trying to understand the semantics of the construct being proposed in this `if match` alternative.

Is the idea of this alternative extended proposal that the new `if match` form would provide a way to describe a (non-exhaustive) match that has an (non-optional) `else` attached? (And if the given match is exhaustive, then the attached else becomes optional and in fact pointless?)

E.g. are you suggesting something like this:

```rust
enum Cases { Case1, Case2, Case3 }
if match Case2 { Case1 => 1, Case2 => 2 } else { 3 }; // => 2
if match Case3 { Case1 => 1, Case2 => 2 } else { 3 }; // => 3
```

I know this is only an alternative so it may not seem as important to tie down the described semantics; but I would like to understand the proposal so that I can evaluate the main text in the appropriate context.

nikomatsakis:
@pnkfelix I think that [@withoutboats's comment here](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-260075431) suggests that adding an `else` to a match was intended to be sugar for an `_`. However, his comment did not include an `if`. That *may* be an error in the RFC.

burdges:
Isn't `else` as sugar for `_ =>` harmful though?  Yes, it reduces right ware drift, but `else` introduces confusion that `_ =>` avoids.  

If right ward drift were a big enough concern to justify this strangeness, then it could be avoided by reducing the standard indentation in match blocks to 2 or 3 spaces, ala
```
match foo {
   _ => {
      bar()
   }
}
```
Also `if match` already means `match` returns a `bool`, not that such cases cannot be handled with a `let`, but it's a strange construction.  Now `else match` is a bit different in that people expect `else if`.

joshtriplett:
I have a concern about extending constructs that currently accept braced blocks to also accept other constructs that themselves have braced blocks.  This can lead to language ambiguities, and constrain future language design.

For example, imagine that we extended `if condition` and `else` to accept `match`, `loop`, and `for`.  Now, imagine that later, we want to extend `for` with an `else` block, similar to Python's `for`...`else` construct.  What would that mean for constructs like these?

```rust
if condition for x in y {
} else {
}
```

This could mean `if condition { for { ... } } else { ... }`, or it could mean `if condition { for { ... } else { ... } }`.

```
if condition {
} else for x in y {
} else {
}
```

This doesn't have a parsing ambiguity, but it seems extremely confusing.

If you want to respond to this by saying "well, let's not allow `if condition for x in y {`, only `else for x in y {`, then that introduces an inconsistency into the language.  Why can `else` accept things that `if condition` (and `else if condition`) cannot?

glaebhoerl:
(I feel like the discussion has been meandering? As far as I can tell, none of the comments since [@aturon's comment](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-266592764) have been directly about the proposal at hand, the tendency being more toward people expressing concerns about *other* possible syntactic extensions which... weren't being seriously proposed in the first place.)

> If you want to respond to this by saying "well, let's not allow `if condition for x in y {`, only `else for x in y {`, then that introduces an inconsistency into the language. Why can `else` accept things that `if condition` (and `else if condition`) cannot?

This is inconsistent *currently*. Things which can currently come after `else`: `{ ... }`, `if condition { ... }`, `if let assignment { ... }`. Things which can currently come after `if condition`: `{ ... }`. So, at least, it would not be any kind of regression w.r.t. consistency.

(And yes, that's how I want to respond. ;) Let's generalize the thing that's straightforward to generalize and think about other possibilities later.)

joshtriplett:
@glaebhoerl As I said, I don't see an issue with the very specific case of `else match`.  However, since several people (including you) have proposed generalizing this to accepting any construct that itself takes a block, I wanted to go ahead and raise the concern about that breaking/limiting future semantic/syntactic extensions.

glaebhoerl:
@joshtriplett If I'm understanding your previous comment correctly, the actual problems are with extending `if`, which we're not proposing, and not with `else`?

joshtriplett:
@glaebhoerl Not necessarily, no.  It wouldn't necessarily create a parsing *ambiguity* in that case, but it could still result in confusing constructs.  The problem comes in if you want to introduce `else` blocks for other constructs, such as `for`, `while`, or `catch`.

nrc:
Summary

This PR is currently in 'pre-RFC' with an inclination to merge. @nmatsakis proposed FCP to merge [here](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-260062667) with the justification that this is a small change which improves ergonomics (albeit in a minor way).

@pnkfelix registered a concern [here](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-269222247) that the semantics of `if ... match` are unclear.

There has been some discussion about whether this proposal should be extended, either to allow all control flow constructs after `else` (i.e., not just `match`), and/or to allow `match ... else`, and/or to allow a more general removal of braces, e.g., `loop match { ... }` instead of `loop { match { ... }}` (there are been some discussion on this last item, but I don't think anyone strongly thinks it should be part of *this* RFC).

@glandium has raised the point that this change is somewhat analagous to allowing `if foo` instead of `if { foo }` (https://github.com/rust-lang/rfcs/pull/1712#issuecomment-260775231).

@aturon supports the RFC (I think, although hasn't checked his box, presumably waiting on changes?), as long as it supports all constructs after `else`, but is against `match ... else` (https://github.com/rust-lang/rfcs/pull/1712#issuecomment-266592764)

ping @aturon, @pnkfelix, @withoutboats for review

aturon:
Ping @phoenixenero (author), any thoughts on the latest discussion?

Mark-Simulacrum:
I am against this change, because I find it harder to reason about control flow with else match. While it is equivalent to the desugaring, I do not like it.

The change seems minor (which is why I believe some of the team is in favor or at least not opposed), but I think the implications are relatively large. This change seems to affect a very small number of use cases: it is rare that one would want to match in an else, I think. Most often I see that match code wants to operate on one value, and compare it against various other patterns or matches.

Because the change proposed seems to make an edge case that I think decreases readability and encourages a pattern that is non ideal.

However, it is hard for me to come up with specific reasons against it. It simply doesn't feel Rusty to me.

I think that what we might really want is a match like block or even match itself able to take arms with arbitrary ”if" blocks: if let and if. That is, "_ if let ... = ...` be a valid LHS for an arm. This is cleaner I think than slowly extending what can appear after else.

My main point is that the usability increase is marginal in comparison to the potential difficulties created.

aturon:
@Mark-Simulacrum Thanks for the feedback! I had one question:

> I am against this change, because I find it harder to reason about control flow with else match. While it is equivalent to the desugaring, I do not like it.

Can you explain why you find it harder to reason about control flow? I found this a bit confusing, since the desugaring is so simple and (to me) straightforward to understand.

I do agree that the gain here is marginal; it's papercut reduction.

Mark-Simulacrum:
Thinking about this more, I think I misspoke. It's not that I find else match to be confusing, or other similar constructs, I think. However, if we add else match, I would somewhat expect to have else match else since match feels like if to me in that case.

More so, I think the advantages that else match give are limited. Even the RFC acknowledges this. It is used rarely in the wild, and gives no advantage over marginally reducing the right side drift. It also somewhat adds to our strangeness budget, since the current options exist in all languages, at least in broad strokes. I almost feel like I might want this if it did more: allowing else to take any "blocked" expression. This has been suggested in this thread before. Currently, I am worried that users will expect these to work after seeing else match, and I certainly want a good error message, something I suspect doesn't happen currently (can't check right now).

Does that help, @aturon?

joshtriplett:
@Mark-Simulacrum It does seem to add to the level of explanation required.  Before: "`if` and `else` bodies always require a braced block." After: "...or a (subset of) constructs that themselves take a braced block."

In particular, I can imagine someone seeing `else match ... { ... }` and inferring, incorrectly, that `else` doesn't itself require braces.

I like how Rust opted for simplicity (including simplicity of human parsing) and lack of ambiguity instead of brevity, by always requiring braces.  This change to that seems harmless enough in isolation, but it doesn't seem clear to me that the benefits outweigh the cost.

withoutboats:
I'd like to only extend the composability of our control flow constructs like this in a way that is the assemblage of a set of straightforward, orthogonal rules. I don't want to add more and more special cases based on use until we have a hodge podge of 1 off parsing rules.

I think the most coherent extension that would enable `else match` (and some other useful constructs) would be this:

First, we define a set of 'control flow constructs' which all should have certain things in common (in particular, they should `{ }` to create new scopes). I think this is a complete list of the control flow constructs we have today:

* a plain block
* if
* if let
* match
* loop
* while
* while let
* for in
* unsafe (this creates a scope so it counts IMO)

Second, we say that else and loop, instead of taking blocks, take any of the control flow constructs.

This allows `else match`, `loop match`, `else if let`, `else loop`, and so on. It also allows certain very strange & definitely bad constructs like:

    'x: loop loop loop { break 'x; }

Basically, I think that's the trade off here - we can make these rules more flexible, allowing for some useful constructs, but only if we also accept some definitely pathological constructs like `loop loop`.

glandium:
> I think this is a complete list of the control flow constructs we have today:

Shouldn't `else` be in that list too?

> Second, we say that else and loop, instead of taking blocks, take any of the control flow constructs.

Add "except themselves" and you got yourself out of the `loop loop` weirdness.

withoutboats:
> Shouldn't else be in that list too?

No, you can't have a freestanding `else`.

withoutboats:
We could apply this rule only to `else` and not `loop`, and that would avoid the confusing constructs.

glandium:
Note, `loop while` would be weird too.

glandium:
Or `loop for`, for that matter.

joshtriplett:
@glandium To the extent any of these make sense, `loop for` does have use cases.  For instance, consider `loop for thing in get_next_iterator()`.  I think writing it that way obfuscates the meaning, but the more I look at any of these the more I feel like all of them sacrifice clarity for brevity.

Looking at the motivating example in Servo, linked from the RFC:

```rust
    pub fn clamp(&self, other: Au) -> Au {
        if other < self.min_size {
            self.min_size
        } else {
            match self.max_size {
                Some(max_size) if max_size < other => max_size,
                _ => other
            }
        }
    }
```

This could just as easily use:

```rust
    pub fn clamp(&self, other: Au) -> Au {
        if other < self.min_size {
            return self.min_size;
        }
        match self.max_size {
            Some(max_size) if max_size < other => max_size,
            _ => other
        }
    }
```

Does anyone have a use case where `else match` would make code significantly clearer, and no other alternative possible in Rust today would provide comparable clarity?

Ixrec:
A futile attempt to be somewhat comprehensive about all the CFC combinations (assuming withoutboats' list is complete).

- Any non-block CFC followed by a block seems to make sense, and a block containing any of the other CFCs makes sense. We already have all of these combinations.
- `unsafe` followed by any non-block CFC makes logical sense, but part of me really doesn't like the idea of making `unsafe` syntax more concise.
- A CFC and its "let-variant" like `if if let` or `while while let` is clearly pointless.
- Because the loops while, while let, for in and (at least for the moment) loop are not expressions returning a value, it seems pointless to put them immediately after any conditional CFC or other loop. @joshtriplett's example is so weird it seems like the exception proving this rule.

Which leaves us with only two "interesting" categories for expansion:
1) loop-conditional: `loop if [let]`, `loop match`, `while if [let]`, `while match`, and `for if ... in ...` all seem like fairly reasonable things to allow.
2) conditional-conditional: We already allow `if [let] match`, `match if [let]`, `else if [let]`. I don't think `if [let] else` or `match else` makes any sense, and then the only combination left is `else match`.

-----

Although I still don't believe any of the suggestions here are particularly well-motivated, I do like the idea of reframing this in terms of "orthogonal rules", I think there is a smaller version of @withoutboats' suggestion that drastically reduces the amount of pathological silliness:

First, we define a set of 'control flow _expression_ constructs' which all can use {} to create new scopes, and _can evaluate to a value other than ()_. I believe that list is:

- a plain block
- if
- if let
- match
- unsafe (though if I were to write the RFC I'd probably exclude this as a special case)

Second, we say that `else` and `loop`, instead of taking blocks, take any of the control flow expression constructs.

withoutboats:
@Ixrec `loop` now can evaluate to any type: https://github.com/rust-lang/rust/pull/37487

It seems like your extension is more correctly defined as "non-looping control flow constructs." This seems like a relatively simple & decoupled set of rules that avoid weird constructs; I don't think anyone is craving `else loop` anyway.

Ixrec:
@withoutboats Oops, I thought the loop break value thing hadn't happened yet. Given that, I agree "non-looping control flow constructs" makes more sense.

glandium:
Mmmmm from a user perspective, if both `match` and `loop` blocks can evaluate to any type, and `else match` is allowed, it seems oddly restrictive that `else loop` wouldn't be allowed.

Mark-Simulacrum:
`else 'a: loop` looks rather ugly, however. I'm not sure we want that, though it does feel like an odd restriction to not permit `else loop`.

nikomatsakis:
@joshtriplett 

> Does anyone have a use case where else match would make code significantly clearer, and no other alternative possible in Rust today would provide comparable clarity?

I think it will never make code "significantly clearer". From time to time, though, it lets you write without early return, in a more compact style. Here is an example of some unification code; the idea is that we first want to apply normalizations -- only if no normalization is possible, do we start examining cases. 

Before (the code I have today):

```rust
        if let Some(n_a) = self.table.normalize_shallow(a) {
            return self.unify_ty_ty(&n_a, b);
        } else if let Some(n_b) = self.table.normalize_shallow(b) {
            return self.unify_ty_ty(a, &n_b);
        }

        match (a, b) {
            (&Ty::Var(depth1), &Ty::Var(depth2)) => ...,
            (&Ty::Var(depth), &Ty::Apply(ref apply)) => ...,
            (&Ty::Apply(ref apply), &Ty::Var(depth)) => ...,
        }
```

After (what I might rather have written):

```rust
        if let Some(n_a) = self.table.normalize_shallow(a) {
            return self.unify_ty_ty(&n_a, b);
        } else if let Some(n_b) = self.table.normalize_shallow(b) {
            return self.unify_ty_ty(a, &n_b);
        } else match (a, b) {
            (&Ty::Var(depth1), &Ty::Var(depth2)) => ...,
            (&Ty::Var(depth), &Ty::Apply(ref apply)) => ...,
            (&Ty::Apply(ref apply), &Ty::Var(depth)) => ...,
        }
```

Now, is this a deal-breaker? Obviously not. It's just a little thing that comes up for me from time to time. As I said, to me, it really is something special to `if/else` and `match`, which are our two ways of "querying" state and I find I want to link them together sometimes.

Interestingly, I cheated a little though. =) In my real code today I have two debug printouts before the match:

```rust
        if let Some(n_a) = self.table.normalize_shallow(a) {
            return self.unify_ty_ty(&n_a, b);
        } else if let Some(n_b) = self.table.normalize_shallow(b) {
            return self.unify_ty_ty(a, &n_b);
        }

        debug!("unify_ty_ty(normalized a={:?})", a);
        debug!("unify_ty_ty(normalized b={:?})", b);

        match (a, b) {...}
```

This seems like the right place to do a debug. I remember when writing the code that I thought "I wish I had `else match`", but maybe it was better that I didn't!

nikomatsakis:
Also, I have to say that I really find stuff like `loop for x in y { }` impossible to parse. My mind just can't comprehend what it means. I keep expecting the `{ }` to be the body of the `loop` keyword or something. Similarly, `else loop { ... }` feels a bit confusing. I think I'm starting to sour on this whole endeavor and leaning towards just keeping things as they are.

withoutboats:
> I'm starting to sour on this whole endeavor and leaning towards just keeping things as they are. =/

Yea I kind of am too. Even statements like `loop if false { }` are quite confusing; quite different from `while false { }` !

We could add `else match` as a one-off special case, but I pretty much never use it and I'm inclined to say that people can abide the rightward drift rather than have a special case. What I use more is actually `loop match`, but I feel like *I* will just abide the rightward drift for that case.

nikomatsakis:
@rfcbot fcp cancel

I'm ready to back off. I would at this point support closing. I'm still not opposed to `else match`, but it opens lots of annoying questions, and I have to admit that I also find it... a little harder for my eyes to parse than I expected.

aturon:
@nikomatsakis I'm still a fan of this idea, generalized in some way as we've been discussing on thread. I'm not too bothered by cases like `loop for x in y {}` any more than I am by things like `return break` or `return return`. These are completely analogous situations to my mind: in both cases, we strive for uniformity in the grammar, which may allow for nonsensical combinations that people simply don't use.

I remember when `if let` was first proposed, I thought that it was a very minor thing and probably not worth the extra complexity. Boy was I wrong! *Still to this day* I feel happy every time I'm able to use it to avoid a more verbose `match`. These little syntactic affordances add up to making the language feel more pleasant and less pedantic.

While I don't want to accept the RFC as written (as I [said before](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-266592764)), I'd still love to see a fresh/revised RFC that attempted to lay out a design based on orthogonal syntactic rules, as [@withoutboats so beautifully articulated](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-271119282).

nikomatsakis:
@aturon I think I'd be ok w/ using those rules after an `else`. I am not yet comfortable with them after a `loop`, but perhaps I could be persuaded.

withoutboats:
The one which concerns me most is `loop if`, much more than `loop loop`. If we allowed `loop if` I can easily imagine someone using it when they meant `while`, and then - because they have the `while` semantics in their mind - they have absolutely no idea why their program is looping indefinitely. Seems like an extremely bad footgun.

I don't see any of these issues with `else`, but I feel like the savings are a lot less than `if let`. `if let` let you drop a whole match arm that felt quite boilerplate, this just lets you drop a single brace set.

aturon:
@withoutboats 

> The one which concerns me most is `loop if`, much more than `loop loop`. If we allowed `loop if` I can easily imagine someone using it when they meant `while`, and then - because they have the `while` semantics in their mind - they have absolutely no idea why their program is looping indefinitely. Seems like an extremely bad footgun.

That is a very convincing argument. I wonder if a similar confusion could arise with `loop match`, which is mildly useful (for state machines).

This raises further questions about the importance of uniformity. Would we consider adding only clear/sensible combinations? **Are there more than philosophical downsides here?** Special case complexity can make the language harder to learn, but it's not obvious to me to what degree that would happen in practice here, and whether that outweighs the (admittedly small) ergonomic win.

I will note that, given that `else if` is already accepted, further expansion for `else` doesn't make things any *less* consistent (and arguably makes the special case less narrow).

withoutboats:
Ruby has an `unless` construct, which is sugar for "if not". In general, it can be used like `if`:

```ruby
unless foo.is_bar?
  do_some_stuff
end

// Ruby also has post-conditionals on expressions
do_some_stuff unless foo.is_bar?
```

However, it cannot be combined arbitrarily with `else`:

```ruby
// Invalid
unless foo.is_bar?
    do_some_stuff
else
    do_some_other_stuff
end

// Invalid
if foo.is_baz?
    do_this_thing
else unless foo.is_bar?
    do_some_stuff
end
```

This kind of irregularity presents a real speedbump for learning, because users don't look up what all the valid control flow constructs are - they see them in code and then combine them through mimicry. The invalid constructs above seem perfectly reasonable, from my knowledge of the valid constructs I inferred that they should be accepted. When they instead failed to parse, this was an immediate frustration and stumbling block, but it also decreased my confidence in the language as a whole, because it seemed arbitrary and overly complicated.

This is why I think they need to be grounded in orthogonal rules, not from a philosophical purity but from the way that enables users to have a generative experience of learning - they can pick up these underlying orthogonal rules from the code they use and then guess what other instances those rules permit in the language.

`if let` is actually a great example of this, because rather than special casing `Option` or something, `if let` takes any pattern, so as you learn about patterns that are applicable in matches, you can guess correctly that they are also valid in `if let`s.

Irrefutable patterns are an example that tends toward special cases, but is very well justified & so it feels permissible.

I wouldn't want to allow `else {`, `else match {` and `else if {`, but then allow `loop {` and `loop match {`, because I think users would infer from that `loop if {` works.

withoutboats:
Also highlighting `else if` as a special case feels a bit dubious - `else if` is supported because its one of the most common constructs in mainstream programming languages & so its justified by being a universal vernacular users already expect.

Mark-Simulacrum:
Perhaps a good reasoning method is to only allow things which are "extendable" after the `else`. That is, `else if` is valid because `if` can be extended with `else`. Match, therefore, would not currently be extendable (AFAIK) because `match ... { ... } // nothing can go here`.

This rule makes cases such as else match only have one extra level of indentation, which I think is an acceptable loss, given that I believe this rule is rather clear, and easy to apply to any case: "Can it take an "else" (or other similar control flow construct) branch? Yes: Can go directly after else. No: Cannot."

aturon:
@withoutboats Thanks for that very lucid and evocative description of the learning process. I'm persuaded that having different rules around `loop` and `else` would cause more harm than good.

I'm still a little uncertain about expanding `else` alone. While I take your point that `else if` is universal (and hence wouldn't lead to an expectation that `else match` would work), is there room to make an improvement here without sacrificing the kind of "generativity" you mention? It's hard to know what people's mental model is likely to be here.

withoutboats:
It seems like we could allow any control flow after an else, but not expand loop, and that would be a simple rule. But my original intuition, which seems to be the same as @Mark-Simulacrum's, was that `if { } else X { }` implies `X { } else if { }`, which would only work for `match` if we allowed it to be non-exhaustive and hit the `else` case if it isn't matched.

& FWIW, if I am confronted with a situation where I want this control flow, I am more likely to do:

```rust
match (boolean, matchable) {
    (true, _) => { ... }
    (_, Some(_)) => { ... }
    (_, None) => { ... }
}
```

But maybe that would change after this proposal.

joshtriplett:
@withoutboats I tend to do that using guards:

```rust
match matchable {
    _ if boolean => { ... }
    Some(_) => { ... }
    None => { ... }
}
```

solson:
I'm strongly in favour of @aturon's conclusions in https://github.com/rust-lang/rfcs/pull/1712#issuecomment-266592764. The proposals unrelated to `else <thing> {` seem like they could be decided separately (and I don't have strong opinions on them), but the idea of allowing any block-like control structure after `else` seems eminently teachable and clear to me.

When you have this situation:

```rust
} else {
    <thing> {
        ...
    }
}
```

You may optionally shorten it to:

```rust
} else <thing> {
    ...
}
```

That's it!

withoutboats:
Yea "an `else` can take any control flow" is a pretty simple rule. I'm just wondering how often users will think the rule is more expansive than it is, and think e.g. a loop can anything, or anything can take an else.

nikomatsakis:
@withoutboats 

Note though that this alternative (nor @cuviper's variant with guards) is not at all equivalent, because both of them still evaluate `matchable`. The scenario where I want this form is when I do not want to evaluate the expression I am matching upon unless I need to.

```rust
match (boolean, matchable) {
    (true, _) => { ... }
    (_, Some(_)) => { ... }
    (_, None) => { ... }
}
```

nikomatsakis:
Sounds like there might be some consensus emerging that we could accept `else <keyword>` but not `else match`?

solson:
@nikomatsakis Did you mean "but not `match else`"? As in `match <expr> { ... } else { ... }`.

If so, I agree, `else <keyword>` seems less controversial than `match else`.

cuviper:
(the guard variant is @joshtriplett's, not mine...  #notalljoshes)

aturon:
> Sounds like there might be some consensus emerging that we could accept `else <keyword>` but not `else match`?

Does anyone strongly object? If not, I suggest that @phoenixenero update the RFC accordingly.

aturon:
Paging the shepherds: @joshtriplett @solson @whitequark -- is one of you interested in updating this RFC [as proposed](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-273318459)?

Mark-Simulacrum:
I believe the general consensus is `else <keyword>` but not `match else` (and other like blocked expressions for the time being, at least). Is that the correct interpretation?

aturon:
@Mark-Simulacrum Yep, that's right.

louy2:
Since @burdges 's opinion seems to be just brushed away and not brought up by anyone, I really wonder: are we going to change the parser of the language just to prevent right shift?

I remember when writing C++ with vim, it does a simple thing to prevent right shift on switch-case: align the case lines with the switch line. It thus unifies switch and if in that all cases are at the same level of indentation as the construct itself, and resulting executions are one level deeper. Or as @burdges has proposed, if you prefer the elegance of else and match on the same line. Either way, it is much more simple to just let rustfmt do it for you, or at least let it ignore a manually done one. 

In the end, we change either parsing rules or formatting rules. I know this is the place to discuss parsing rules, but I think parsing rules are more important than formatting rules and should not change with only cosmetic benefits, because cosmetics is what formatting rules are for in the first place.

scottmcm:
Looking at the two examples in the RFC, I'm not convinced even the "just `else match`" version of this RFC is worth doing.

The XSV snippet looks like it could be written as something like
```
match self.online {
    Some(ref v) if self.typ.is_number() => {
        pieces.push(v.mean().to_string());
        pieces.push(v.stddev().to_string());
    }
    _ => { pieces.push(empty()); pieces.push(empty()); }
}
```
I won't say that's necessarily better than the current version; the reading order is certainly different.  It makes me wonder, though, how often the if-else-match pattern is sortof a hybrid match where for some reason making the if an arm is awkward, and whether where are features that could just help make that transformation easier.

Similarly, the trust-dns example is `if class == self.class { ... } else { match class { ...`.  So the first `if` body could presumably become an `x if x == self.class` arm.  (Again, I won't take a position on whether that would be _better_, just that it's potentially a way to have less typing and rightwards-drift using current features.)

So I think that if the `if` is conceptually different enough to the code author from the `match` that it's not incorporated as an arm, then I'm not convinced that the extra brace and indent is bad enough to make the change to the language.

Thinking more, to me the thing that makes `else if` compelling, but not `else blah`, is that it can continue to chain consistently at the same "level".  I'm not convinced that `else match foo { ... } else match bar { ... }` would be desirable.  And that would be particularly bad if extended to `if ... { ... } else while ... { ... } else { ... }`, where the second else is something quite different from the first.  (Though that needs two things that, while they've been proposed, seem to be heading to a "no" decision.)

aturon:
@scottmcm 

> Thinking more, to me the thing that makes `else if` compelling, but not `else blah`, is that it can continue to chain consistently at the same "level".  I'm not convinced that `else match foo { ... } else match bar { ... }` would be desirable.

This is a good point, which has been somewhat implicit in the thread so far: if we don't want `match exp { .. } else { .. }` to be a thing, then `match` must terminate an `if`/`else` sequence, making it even less analogous to `else if`.

I don't personally see this as a death knell for the idea, but it does put to rest any argument that `else match` (etc) makes the language more consistent. It's purely a small ergonomics improvement with several restrictions. (I'll note that `if let` also comes with its own restrictions, e.g. the fact that the `let` must be the only clause of the `if`; it's still a net win.)

glaebhoerl:
> then `match` must terminate an `if`/`else` sequence, making it even less analogous to `else if`

It's analogous to `else { }` though :)

solson:
> it does put to rest any argument that `else match` (etc) makes the language more consistent

It seems like "more consistent" is kind of fuzzy in this case. This proposal *does* generalize the `else { if` => `else if` shortening to `else { <block>` => `else <block>`. Whether those other blocks can be followed by `else` is a separate idea. However, I agree that it's really just an ergonomic improvement for end users.

withoutboats:
I looked through my own code a few weeks ago and I couldn't find any places I would have used this (the match form or the expanded form); that is a place where I currently have `else { control flow expression { } }` (I didn't look for things like `match`s that could be transformed with it). I've grown more wary of making our `if/else` work in a more unique way from other languages'.

scottmcm:
Seeing the replies, I realized a better way to say what I said before:

Regardless what the grammar says, I _think_ of `if {} else if {} else {}` as one unit of things, not as `if {} else`-followed-by-`if {} else`.  I tend to write `if{}else{if{}else{}}` _with_ the unnecessary braces when the second condition is mentally separate from the first, especially if it's `if{ if{}else{} }else{ if{}else{} }`.

Phrased as a desugar, I think of `if A { B } else if C { D } else { E }` as `match () { _ if A => B, _ if C => D, _ => E }`, _not_ a nested match.

So from that perspective, `else match` is a weird exemption to the "blocks have braces" rule, not a consistency improvement.  That gives analogies like expecting `fn foo(a:A) match a { ... }` to be valid function definition syntax because `else match` taught you (incorrectly, obviously) that "match doesn't need to be wrapped in braces".

(It may partly be Perl's fault, with its `if-elsif-else`, but that mental model has stuck with me through many other languages, even weird things like `?:` chaining in C.)

iopq:
I'm not opposed to the idea itself, but I'm opposed to special-casing syntax like this unless it's a clear win. The problem is that most repos don't even have a use case for this. When someone actually sees this syntax they will be confused because it's so rare and strange.

aturon:
@iopq I understand this perspective in general, but I'm a bit skeptical that `else match` would cause much confusion; it's meaning seems very hard to get wrong. (That's in contrast to some of the other cases we've discussed, like `loop match`, which you could imagine working in a few different ways.)

nikomatsakis:
@scottmcm 

> Regardless what the grammar says, I think of if {} else if {} else {} as one unit of things, not as if {} else-followed-by-if {} else. I tend to write if{}else{if{}else{}} with the unnecessary braces when the second condition is mentally separate from the first, especially if it's if{ if{}else{} }else{ if{}else{} }...So from that perspective, else match is a weird exemption to the "blocks have braces" rule, not a consistency improvement. 

Interestingly, I think your argument is *precisely* why I might want `else match` (but don't care as much about supporting other keywords). I sometimes want to indicate that there is one "logically connected" condition being tested, which sometimes requires guards, but if those guards don't apply, requires matching. I don't like having the first line be `_ if ...` because, among other things, it requires evaluating the `match` discriminant, which may be expensive to do, for no particular reason.

Regardless, I think this is not going to be something we use everyday, and less of a win than (e.g.) `if let`. It's just something that comes up occasionally for me.

nikomatsakis:
> Sounds like there might be some consensus emerging that we could accept `else <keyword>` but not `else match`?

To be clear on what I meant by this (since I see there was some confusion a while back): I meant that we would not want an RFC that *only* supports `else match { ... }`, but we would perhaps want one that supports `else loop { ... }` and other such keywords.

(At least that was the sense I had from the thread at the time; I think I could go other way).

I am however vociferously opposed to `loop <keyword>`, as I think this is actively harmful to clarity and mostly opposed to `match { } else { }`, which just seems silly to me -- though I guess I see the idea that "anywhere you could have type `if cond`, you can type `match discr`".

glaebhoerl:
Howzabout we do some straw polls...

glaebhoerl:
:+1: for "yes please", :-1: for "no please", :confused: for "meh"

`else match foo { ... }` (instead of `else { match foo { ... } }`), no other `else keyword` extensions

glaebhoerl:
:+1: for "yes please", :-1: for "no please", :confused: for "meh"

`else keyword { ... }` (instead of `else { keyword { ... } }`), where `keyword` is any construct that has a block attached, including `match` (but also `loop`, the already-permitted `if` and `if let`, etc.)

glaebhoerl:
:+1: for "yes please", :-1: for "no please", :confused: for "meh"

`loop match x { ... }` (instead of `loop { match x { ... } }`), but not other `loop keyword`s

glaebhoerl:
:+1: for "yes please", :-1: for "no please", :confused: for "meh"

`loop keyword { ... }` (instead of `loop { keyword { ... } }`), where `keyword` is any construct that has a block attached, including `match`

glaebhoerl:
:+1: for "yes please", :-1: for "no please", :confused: for "meh"

`match foo { ... } else ...`, a partial `match` that falls back to the `else` clause if none of the arms match

nagisa:
This is not a bikeshed colour level question to be doing polls.

I find doing polls for language design ill advised as it encourages taking decisions on these poll results and also encourage people who aren’t familiar with design space to cast the votes. These votes then have the same weight, in GH interface at least, as the votes by people intimate with the design issues and whatnot.

glaebhoerl:
I agree with your concerns about this method **in general**... but. I feel like this thread has been drifting around in circles for a long while now, with a great lack of clarity around which features are actually "being proposed", or should be pursued, or have any support, or how many of them there even are, with different people drawing different associations between them, and concerns and sentiments around one getting conflated with another, and so on. This was an attempt to try to get some more clarity on at least some of those fronts -- and maybe to winnow the contenders.

> I also find it amusing that all of these polls are generally negative, even if its only open for 2 hours yet.

Especially since one of them got +10 positive "votes" when I first floated it near the beginning of the thread ;)

aturon:
I feel like we've been slowly building toward consensus around `else keyword`, but I guess the emoticons will tell...

scottmcm:
Another thought: there are a bunch of languages that already allow `else keyword` by not requiring braces after `else`.  Does that possibility get used in them?

By my reading, it's disallowed by the Mozilla [Coding Style](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Control_Structures) and the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Conditionals), for example.

Personally, I've never seen things like `else while` in C++ nor `else foreach` in C#.  I feel like if I wrote it, I'd get code review feedback telling me to just put the braces in.

Hmm, should RFCs for alternative syntax propose which version (if either) should be canonical?  For those in favour of `else keyword`, would you expect rustfmt to reformat existing `else { keyword` to use it?

glandium:
Note that Mozilla and Google's C++ coding styles are about
```
if (foo) {
  do_one_thing();
} else {
  do_another();
}
```

They are *not* advocating for:
```
if (foo) {
  do_one_thing();
} else {
  if (bar) {
    do_another();
  }
}
```
instead of
```
if (foo) {
  do_one_thing();
} else if (bar) {
  do_another();
}
```

scottmcm:
@glandium I agree that they don't require `else { if { } }`; both have explicit counterexamples.

But I still don't think they allow
```
if (foo) {
  do_one_thing();
} else switch (bar) {
  case 1:
    do_another();
    break;
  default:
    do_that();
    break;
}
```

For the Mozilla one, for example, I can see no exception or counterexample to the "Always brace controlled statements" rule that would allow `else switch` or `else while`.

Hmm, DXR looks nice.  Let's ask it about `else switch` in C++: https://dxr.mozilla.org/mozilla-central/search?q="else+switch"+ext%3Acpp  In 9,130 files, it finds three uses.  Two are in machine-generated code (from YACC), where two more braces wouldn't be a hardship.  The third is from ICU, in code that clearly doesn't follow Mozilla's coding style.

`"else while" exp:cpp`? No matches.  `"else for" exp:cpp`?  4 matches, three of which are comments.  The other has a loop in the `if` block, so IMHO would actually look better with the braces-and-indent.  And that particular use throws off my mental pattern matching, as my brain sees
```
  else
    ...
    return NS_OK;
  }
```
but the else doesn't always return OK because that return is inside a for loop over a linked list.

So overall, I can't help thinking that generalized `else keyword` isn't that common a case, isn't that helpful when it does come up, and might end up being considered bad style anyway.

glaebhoerl:
@scottmcm Thanks for bringing that up, it's definitely something that's worth investigating. Do we know what their reasons are for disallowing it? That is, is there a substantive reason why it *should* be considered "bad style"? If there *are* reasons behind it, we should very much take them into consideration. If we don't know of any, though, then it's still an interesting data point, but I think we should make the decision based on substance ("are there any problems this could conceivably cause?") rather than heuristics ("how do other people do it?").

For what it's worth, I suspect the reason for it is that, in guidelines and conventions, because you're solely reliant on people to keep them all in their head, simplicity and clarity are at a premium, and it's preferable to have few blunt, conservative, but easy to remember rules, than nuanced but more flexible ones. Having the compiler check it for you alters the balance between tradeoffs significantly: if the compiler lets you do it, then you're good to go, and you don't have to second-guess yourself nearly as much. Complexity still has a price, of course, but it's considerably less acute. (And at least in terms of how the rules are formulated, I've [argued](https://github.com/rust-lang/rfcs/pull/1712#issuecomment-239396686) that `else keyword` *reduces* complexity -- in terms of how brains conceive of the rules, it's of course harder to tell.)

For what it's also worth, I've personally been using `else for` and `else switch` in my C++ code for a long time now, precisely because I don't see any substantive reason there'd be a problem with it.

burdges:
If `else keyword` is allowed, then I suggest that `rustfmt` should *not* reformat 
```
if .. {
    foo(0);
} else for i in .. {
        foo(i);
    }
```
or
```
if .. {
    foo(0);
} else for i in .. { foo(i); }
```
or
```
if .. {
    foo(0);
} else 
    for i in .. {
        foo(i); 
    }
```
or even 
```
if .. {
    foo(0);
} else 
for i in .. {
    foo(i);
}
```
into an obnoxious indentation that hides the loop like
```
if .. {
    foo(0);
} else for i in .. {
    foo(i);
}
```
If anything, `rustfmt` should convert this fifth indentation into the first or third.

I've no feelings against `else match` itself because `match` has the same time complexity as `if`, and `match` often requires double indentation anyways.

repax:
@burdges: In your first and third example:

```rust
if .. {
    foo(0);
} else for i in .. {
        foo(i);
        ...
        ...
        // Imagine some more code here...
        ...
        ...
    }

//<-- The lack of an ending bracket on this indentation level
// makes me think that the brackets are unbalanced.
// I feel I'd had to go back and count them.
more_code();
```

I don't think that any particular reformatting is needed. If any, then:

```rust
if .. {
    foo(0);
}
else for i in .. {
    foo(i);
}
```

nikomatsakis:
@rfcbot fcp close

Honestly, I feel like this discussion has been kind of going around in circles. It seems clear that people's intuitions differ here. Therefore, I am moving to close this RFC: I think we ought to just leave things as they are for now. The status quo is reasonable and we can consider more extensions in the future. 

To summarize some of the recent discussion:

- Just support `else match { ... }` strikes some (ahem, at least me) as reasonable, but strikes others as a kind of surprising limitation. They would expect to be able to substitute any keyword, at minimum, in that spot.
- But for others, supporting `else match { ... }` seems to imply that `match { ... } else { ... }` ought to work. Others find this form sort of surprising.
- Still others find that if we support `else match { ... }` we might consider keywords in other places, such as `loop if { ... }`, but the meaning of such constructs can be quite confusing.

solson:
While I was in favor of `else <keyword>` I now agree with @nikomatsakis that none of the proposals seem to have clear support. I also don't think the win from `else <keyword>` is very big, so it's not worth introducing if it would confuse others.

withoutboats:
I've also been feeling increasingly inclined to close this. None of this seems obvious enough for the learning cost to not outweigh the convenience cost.

glaebhoerl:
This is a bit frustrating. I've felt like throughout this thread, there has only been one proposal that has had clear support, but people have kept conflating it with other extensions that very few people think are a good idea. This had the effect of muddying the waters, and making it seem like it was a complicated, confusing and ambiguous situation when it really wasn't. So I thought of doing the straw polls to try to restore some clarity to the situation, and it seems to me that this was successful (moreso than I had expected): it's now *completely clear* that there is, in fact, only one proposal that has any real support. I had hoped that this would have allowed us to refocus discussion around that one proposal, and to stop getting the others mixed up in it. And now the motion is to close, because there are too many different proposals, and the water is too muddy. I... don't see how this makes any sense. The verbiage in @nikomatsakis's comment and my earlier one is remarkably similar: "I feel like this thread has been drifting around in circles"; "I feel like this discussion has been kind of going around in circles". Yes. Maybe now that it's clear that there's only one direction to go in, we could stop going around in circles?

withoutboats:
I'm in favor of closing this for a slightly different set of reasons from Niko.

1. The only proposal raised in this thread has ambivalent support, and I have trouble finding strong motivation for it.
2. I do not believe this will seem like a natural and comprehensible extension to the language. I think it will seem capricious & arbitrary to people when they encounter it (even in the generalized and not match-specific form).
3. The fact that there are many other proposals, even if they have weak or no support, indicates that this is not an 'obvious' extension to language.

aturon:
@glaebhoerl I agree with you that there's been only one real contender for a while; that was clear to me even before the straw poll. And I personally would still be happy to land such a feature. But I think @nikomatsakis was despairing of reaching *full* consensus around the proposal, since many are still opposed, and those on the lang team are a bit ambivalent; meanwhile, this RFC has been a bit of a timesink with pretty minimal benefits if we do land it.

So, for me: I'd love to land this, I'm OK with closing it, but I'm ready to reach a decision.

nikomatsakis:
@aturon @glaebhoerl can one of you clarify what you see as the "one true contender"? Is it extending `else` to permit any control-flow keyword following (i.e., in addition to `if`, also accept `loop`, `for`, `match`, `while`, `break`, `continue`, `return`, are there others?)

I could get behind this proposal but I also feel a lot of ambivalence. I am definitely ready to be done talking about it either way. The win seems relatively slim (less, e.g., than `if let`). The potential for surprise (in the sense of people not really understanding what will and won't work) seems, well, surprisingly high.

aturon:
@nikomatsakis yes, that's the "consensus" proposal. I'm sympathetic to @withoutboats's concerns for sure, but still would be OK trying it.

iopq:
Yes, that's the proposal that makes the most sense to me. It applies rules in a consistent manner, where we can just say `else` can be followed by a block or a construct that has a block attached to it.

It "back-generalizes" why `else if` works, since it's just another construct that has a block attached to it.

withoutboats:
I was in favor of the `else keyword` proposal until about 2 or 3 weeks ago, as I think the discussion record indicates. But at some point I stepped back & I feel like I was driven by a bias for action. The original proposal seemed too specific, so I generalized it to something plausible. But just because its plausible doesn't mean its preferable to doing nothing & I realized that I would really rather do nothing.

I understand the argument that `else if` is a special case that this eliminates, but I don't agree. `else if` is an expected/understood construct from other language. Like much of our control flow*, `if`/`else` is empowered by familiarity as much as convenience, and making it act strangely from other languages only because its more general in some abstract sense doesn't make it more intuitive or usable.

*The most glaring example of this "familiarity sugar" is `while`: I doubt we would have considered that use case deserving of sugar if it weren't for the fact that its a universallyy common construct in imperative languages.

glaebhoerl:
I think that's the strongest counterargument, fwiw. Finally the discussion we should be having. :)

I'm not so convinced that allowing only `else if` as a special case is more consistent with expectations from other languages, however. As far as I'm aware, more or less all of the mainstream statically typed C-heritage languages - C, C++, Java, C#, Objective-C, D, ... - allow **anything** to come after an `else`, `if` or not, keyword or not, braces-having or not. (Correct me if I'm wrong.) It's mainly the various "scripting languages" which deviate from this by having special `elif` or `elsif` constructs. So we wouldn't be out of line by generalizing, ours would just be slightly narrower. I think of this idea as providing the best of both worlds: protection against `goto fail` and similar mistakes, and the convenience of `else match` and friends where there's no potential danger.

withoutboats:
@glaebhoerl But the consistency in the languages you cite is between `if` and `else` not requiring braces for a single statement, whereas `if` certainly still requires it in Rust. I don't find this appreciably more consistent with those languages' behavior. Rather, we have said that Rust requires braces around the result and does not require parens around the conditional, and that `else if` is "an exception" doesn't even need to be said, users just assume it.

glaebhoerl:
Sure, not more consistent, but not less consistent either. *shrug*

nrc:
I'm also a bit sad about the circles here and I'm voting to close because I think the effort is far outweighing the potential benefit at this point.

FTR I am in favour of `else match` only. `else keyword` feels like consistency for the sake of consistency. I fear it would lead to confusing, hard to follow code with many rare combinations to trip up the reader.

And just to go off topic and risk wasting more time and energy, a note on language design philosophy: IMO, consistency is a means to an end, not a goal. If we try to make consistency an overarching goal (without considering practicality) we will end up with a language which is loved by uber-geeks, popular in academia, and abandoned in practice. As with most things there is a delicate trade off to consider.

iopq:
Consistency is not for the sake of itself, it's for the principle of least surprise. `else if` is not surprising, but `else match` is surprising unless `else for ...` also works

glaebhoerl:
Substance level: I agree this is a very minor thing, there are reasonable arguments both for and against, and in the end it's not going to make much difference either way -- it's obviously not worth wrangling over.

Process level: It kind of bothers me from an incentive-compatibility perspective that a proposal can get sidetracked to death and the official response is, "this proposal has been sidetracked to death so we're going to close it". I might've hoped there would be someone with the responsibility to help discussions *stay* on track.

joshtriplett:
@glaebhoerl I don't think I'd characterize this as "this has been sidetracked to death so we're going to close it".  I think I'd characterize this as "this has too much pull in too many different directions to conclude that we have a reasonable consensus in favor of the change".

It's not clear whether the "consensus" proposal was "else match" or "else keyword" (multiple people described each as the consensus at various times0, and both had substantial pushback.

aturon:
I'm going to go ahead and close out this RFC. While there may still be some merit in ideas along these lines, it has proved too difficult to reach a lasting consensus here, and we have bigger fish to fry in the meantime. Thanks @phoenixenero for the RFC!

phaux:
How about using `=>` or `:` when nesting control statements?

```rust
for x in 0..10 => for y in 0..10 {
    println!("{:?}", (x, y));
}
```

Or simply bring back the people's favorite `if <expr> then <stmt>` and `for <pat> in <expr> do <stmt>` keywords.

;)

gustavolsson:
I don't know if one is allowed to post more comments now that this is closed, but I just wanted to chime in on the discussion: I followed the same path of reasoning for my toy language and eventually settled on two types of statements: action statements and flow control statements. This is very similar to what was discussed here but I think _the concept of splitting statements into 2 categories and separating them from blocks is worth emphasizing_.

By allowing any flow control statement after `else`, a block being one type of (but not limited to) flow control statement, one allows and generalizes the common C-like `else if` without letting any pointless single statement through. New statements would require a block making scoping clear and forcing pointless single statements to visually stand out.

In Rust's case, `unsafe` would not be considered a flow statement even though it has a block, statements and blocks need not to be connected.

Here is part of my grammar:
```
-- Statement block
StmBlock.   StmBlock ::= "{" [Stm] "}";

-- Statement
SAction.    Stm ::= Action;
SFlow.      Stm ::= Flow;
terminator Stm "";

-- Action statement
ADecl.      Action ::= Decl;
AAssign.    Action ::= Ident "=" Exp;
...
AFunApp.    Action ::= FunApp;

-- Flow statement
FIf.        Flow ::= "if" Exp StmBlock;
FIfElse.    Flow ::= "if" Exp StmBlock "else" Flow;
FWhile.     Flow ::= "while" Exp StmBlock;
FFor.       Flow ::= "for" Action "," Exp "," Action StmBlock;
FBreak.     Flow ::= "break";
FBlock.     Flow ::= StmBlock;
```
Note that it is only `else` that allows any flow statement, loops are restricted to blocks.

I'm very much a beginner when it comes to Rust, so please excuse me if this is already done internally or if I've missed something important! :)

PR #1716, opened 2016-08-15
Fix typo in fn variance

aidanhs:
Per https://github.com/rust-lang/rfcs/pull/738#discussion_r23849477

I'm a little unsure though since this hasn't been fixed for so long?

### DIFF
--- a/text/0738-variance.md
+++ b/text/0738-variance.md
@@ -264,7 +264,7 @@ as desired:
 PhantomData<T>         // covariance
 PhantomData<*mut T>    // invariance, but see "unresolved question"
 PhantomData<Cell<T>>   // invariance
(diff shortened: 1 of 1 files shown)


eddyb:
cc @rust-lang/lang

aidanhs:
> Finally, there can be cases where it is ok to make a lifetime longer, but not shorter. This comes up when the lifetime is used in a function return type (and only a fn return type).

By the same line of thought, I think this is wrong as well and should refer to function arguments?

I.e. it is safe to treat a `fn(&'a T)` as a `fn(&'static T)`

nikomatsakis:
Looks like a correct typo.

nikomatsakis:
Er, a correct fix =)

PR #1724, opened 2016-08-22
Update unresolved questions and drawbacks for the FusedIterator RFC

Stebalien:
1. We can't remove the `done` bool without making `Fuse` invariant.
2. One drawback I failed to mention is that implementing `FusedIterator` locks implementors into obeying the `FusedIterator` spec.


alexcrichton:
Thanks!

PR #1732, opened 2016-08-31
Trait alias.

phaazon:
Trait alias would enable aliasing traits.

Useful especially when we want to lift the name from generic to specific (a bit like `Result` on a lot of modules, but for traits):

```
mod gen {
  trait Foo<T> {
    // ...
  }
}

mod spec {
  struct Bck0 {}

  trait Foo as gen::Foo<Bck0>;
}
```


phaazon:
Oops fucked up.

PR #1737, opened 2016-09-03
Longer question marks RFC

est31:
[Rendered](https://github.com/rust-lang/rfcs/blob/fd5fe18b09b35209488e82f3ac52de1d701e62c9/text/0000-longer_question_mark.md).

### DIFF
--- a/text/0000-longer_question_mark.md
+++ b/text/0000-longer_question_mark.md
@@ -1,0 +1,108 @@
+- Feature Name: long_question_mark
+- Start Date: 2016-09-03
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Improve code readability and language consistency by
+renaming an operator, increasing its visibility.
+
+The changes are:
+
+ * Rename the current `?` operator to `?!`.
+
+This RFC amends [RFC 243][243].
+
+[243]: https://github.com/rust-lang/rfcs/pull/243
+
+# Motivation
+[motivation]: #motivation
+
+The `?` operator is very short, and has been criticised
+in the past to "hide" a return. Most times code is read,
+not written. Renaming it to `?!` will make it two times
+as long and therefore harder to overlook.
+
+A single character feels too short for an operator that may return.
+
+Function returns that are not at the bottom
+(aka "implicit return") right now are only doable with the
+`return` keyword and macros. Giving the `?` operator an
+exclamation mark will retain the simple "return keyword or
+exclamation mark" rule that code readers can use if they
+want to find the places a function can return.
+
+The current behaviour of `?` is different from other language `?`
+notations, which make it work on a expression level instead of a
+function level. Those languages often exceed Rust in popularity
+and coders coming from those contexts will be confused by this
+tiny difference in functionality.
+
+This RFC removes the inconsistency both for coders from
+different languages and for readers, and leaves space for a
+future possible `?` operator that matches the other languages better.
+
+## What use cases does it support?
+
+The same use cases are supported for `?!` as for `?`,
+but with increased readability for readers who want to know all
+edge cases of code's behaviour (this is an use case as well!).
+
+# Detailed design
+[design]: #detailed-design
+
+The `?` operator stays a language feature, but gets renamed to `?!`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The process of going through this PR and delaying stabilisation of the `question_mark` feature will
+further prevent use of the `?` feature by stable Rust users.
+RFC 243 has initially been proposed in 2014, and has been discussed since, for a very long time.
+
+Expressitivity wise, writing `?` is shorter than `?!`, and it will make reading code that uses
+`?!` instead of `?` harder if you want to focus on the main functionality of the code and want to
+blend out edge cases like error handling.
+
+If `?` can be overlooked easily, `?!` can be overlooked easily as well.
+
+The new name `?!` may be mistaken for a macro.
+
+The rename might be bikeshedding.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Alternatively, the `?` feature could be dropped completely. The general need
+for non-prefix based error handling feature exists though.
+
+One could drop `?` to wait for postfix macros which may implement postfix error
+propagation on a libs level, and not inside the language, which would be overall
+cleaner. But its not sure whether they will come, and when this will happen. Also,
+the bad name of macros in C/C++ based languages has made macros in rust
+unpopular as well, despite the design advantages. Plus extra syntax would be needed
+to work with `catch` from RFC 243, that the postfix macros would then expand to.
+
+The `?` could be made match other languages instead, making it `.map(...)` sugar.
+The use case of the `?` operator that allows chained execution would still be supported
+if the whole expression is wrapped inside a `try!` invocation, but futures which need a
+short notation for early returns will suffer.
+
+## What is the impact of not doing this?
+
+If things are kept moving the way they are, with letting `?` stabilise,
+it will become easier to overlook where a function returns than with `?!`.
+Also, all the other arguments from the motivation section apply.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Should there be a `?` operator that works as `.map(...)` sugar?
+Which additional use cases will that cover? Will it help to prevent use of `?!`
+in cases where it might disturb a reader?
+
+Maybe the `?!` name is still too short? Should a valid identifier be chosen
+instead of "?" so that if Rust gets postfix macros, the operator can be moved
+into libcore?
--- a/text/0243-trait-based-exception-handling.md
+++ b/text/0243-trait-based-exception-handling.md
@@ -6,6 +6,8 @@
 
 # Summary
 
+** Amended by RFC [fill in RFC number] **. See the RFC for details.
+
 Add syntactic sugar for working with the `Result` type which models common
 exception handling constructs.
 
@@ -314,11 +316,11 @@ However, it was removed for the following reasons:
 - There was some concern about potential user confusion about two aspects:
   - `catch { }` yields a `Result<T,E>` but `catch { } match { }` yields just `T`;
   - `catch { } match { }` handles all kinds of errors, unlike `try/catch` in other languages which let you pick and choose.
- 
+
 It may be worth adding such a sugar in the future, or perhaps a
 variant that binds irrefutably and does not immediately lead into a
 `match` block.
- 
+
 ## Choice of keywords
 
 The RFC to this point uses the keyword `catch`, but there are a number


kennytm:
I don't know why we must be restricted to `[?!]+` 😄. In my Swift project I used the symbol `^` to perform something similar to Rust's `?`.

```swift
postfix operator ^ {}
postfix func ^<T>(opt: T?) throws -> T { ... }
...
let address = try IPAddress(string: addresses[ip]^)^.withPort(12345)
```

durka:
Seems pointless. If we were to add "method macros" then `expr?` would change to `expr.try!()` and you could define other macros like `expr.foo!(bar, baz)`. This RFC seems to churn the syntax (and unecessarily re-open a highly pressurized can of worms) without actually adding extensibility.

est31:
@durka how should method macros work together with the `catch` keyword that the RFC has suggested as well? To "throw" towards the next `catch`, extra syntax is needed. The `?` has served as that extra syntax before.

camlorn:
Going to join in the bikeshedding.  Why not get rid of the operator and make it `maybe <expr>` or `try <expr>`?

The biggest problem with `try!` in my book is that it introduces extra parens, but I'm fine with a longer identifier for this sort of thing personally.  I don't exactly see the problem with `?`, not outside stupid screen reader bugs and limitations land, but if we want to change it I'm more than happy to suggest something that I would find more friendly to read for the hell of it.

I feel like alternative syntaxes were probably already bikeshedded on the original RFC, though. Has anyone checked that thread?

Method macros are interesting and, while off topic, please, yes, let's do that.  I think the big problem with macros generally (at least, why I don't just use them) is that you can't namespace them, and it's going to be a long, long time before that's fixed.  The current system makes it too expensive to use the convenient names; you can effectively only use it once per project or, if the crate is popular, once for the entire ecosystem.

Being able to make custom operators in Rust would also be a nice feature in limited circumstances, but probably you need to combine that with method macros to extend error handling.

durka:
@est31 actually something like `catch` _can_ be implemented as a macro, but it's [horrifying](https://github.com/durka/named-block/blob/dd28a8d16610f0b91b12ac59eaeb0d3636e6c8f3/src/lib.rs#L70).

@camlorn one of the big arguments in favor of `?` is that things like `try!(try!(try!(foo).bar()).baz())` turn into `foo?.bar().baz()?`, and `try (try (try foo).bar()).baz()` (or any other prefix keyword) is no better.

comex:
Longer question marks, eh?  How about ？ FULLWIDTH QUESTION MARK?  There's also the emoji version,❓.  Nice and red!

est31:
@camlorn 

> I think the big problem with macros generally (at least, why I don't just use them) is that you can't namespace them

That's wrong. Macros are always namespaced to the block they are defined. If you define a macro in mod.rs and want to use it in lib.rs that includes mod.rs, you will get an error.

You can even import macros into 
> The current system makes it too expensive to use the convenient names; you can effectively only use it once per project or, if the crate is popular, once for the entire ecosystem.

camlorn:
@durka 
I did say limited circumstances.

That said, it may not matter. I think, if you're willing to use words, you can have any custom operator you want inside something like a `context!` macro.  I think you can even have precedence, though you can't get precedence above the built-in ones that I can figure.

And, with that scary thought, I'll avoid taking this further off topic.

est31:
@camlorn 

> I think the big problem with macros generally (at least, why I don't just use them) is that you can't namespace them

That's wrong. Macros are always namespaced to the block they are defined. If you define a macro in mod.rs and want to use it in lib.rs that includes mod.rs, you will get an error. If you define a macro inside a crate, users of that crate won't get it unless you add `#[macro_use]` when using the crate. You can define macros inside methods, and they don't leak to outside scope.

What you think about is C/C++ preprocessor macros, which do act on a global scope and leak through headers per default until they are undefined, but which destroys any macros with the same name that were used before. Yes, you can't rename Rust makros (unlike with use ... as notation that works for all other language items), and if you use a crate's macros, you need to use all of them, but generally the situation is much much better than on say C/C++.

But your dislike against macros is symptomatic, and the unpopularity is listed as one of the drawbacks of a macro based solution to the problem in the RFC, although I personally would even prefer to use macros over custom operators in my code.

est31:
@durka if you make an RFC for adding method macros and removing the `?` operator, I would really welcome it and would close this RFC in favour of yours considering that mine has so many negative votes already (although nobody from the core team has commented yet, so I don't know their thoughts yet). Let's discuss the idea in its own thread.

@camlorn

> I feel like alternative syntaxes were probably already bikeshedded on the original RFC, though. Has anyone checked that thread?

I've skimmed over it before doing the RFC, but most of the discussion was about entirely different approaches, like one can observe in this thread already. Almost none was about which notation exactly should be used.

To get discussion on topic again: What criticism is there to this proposal other than "I think RFC 243 was wrong, and this is basically RFC 243, lets use [...] instead, I'm against this RFC"?

Why do you think the `?` operator suits better than the `?!` operator? Do you want it to stay short?

durka:
> Why do you think the ? operator suits better than the ?! operator? Do you want it to stay short?

Everything starts at [-100 points](https://blogs.msdn.microsoft.com/ericgu/2004/01/12/minus-100-points/). The `?`/`catch` RFC already climbed out of that hole (it took several years). So it seems like a misdirection to present `?!` and say "OK, what are the downsides?". Rather, what are the upsides of switching from `?` to `?!`? I see two potential reasons to move away from `?` (did I miss any?):

1. It's too short (easy to miss when reading code).
2. It doesn't generalize (we might want a postfix modifier that does something else).

(Also a note that some people consider (1) a positive, because often when reading code you want to focus on the main flow and worry about error handling later.)

This RFC would solve (1) (sort of -- two characters really isn't much more than one) but doesn't touch (2) at all. That's why I don't see it as a major gain.

> if you make an RFC for adding method macros and removing the ? operator, I would really welcome it

I'm not even sure they are a good idea, so I probably won't be the one to write that RFC. We'll see what happens when Macros 2.0 shakes out, maybe.

camlorn:
@est31 
My dislike of macros is not symptomatic of C++, and I'm not wrong.  From the perspective of a public interface, they are unnamespaced.  This means that the list of good names that don't start with `mycrate_` is valuable and finite.

I could say more, but this is off topic.  Suffice it to say that the lack of namespacing support for macros frustrates me greatly, especially since we don't have default and keyword parameters and you have to use macros to emulate them.

comex:
@camlorn It is off topic, but allowing macros to be namespaced and imported like other items is [planned for macros 2.0](https://github.com/rust-lang/rfcs/pull/1584).

whitequark:
How about renaming it to ‽ ? We could also have a right-associative version, ⸘.

dathinab:
@durka, is there a "methode macro" RFC?? 

I think such a think wouldn't be to bad e.g.  `expr?` would first become `expr.try!()` but that woudln't work with `catch {...}` but then you could "upgrade" try like in the current `?` RFC to work with catch and to make sure it is clear separated to "normal" method macros you could again rename it to `?` which means `catch { foo?.bar()?.baz()? }` would become `catch { foo.?!().bar().?!().baz.?!() }`. 
Hm, which I just noticed is not very readable. And if you ad more special cases for a "methode macro" `try/?` then you would and with  `catch { foo.?.bar().?.baz.? }` or  `catch { foo.?().bar().?().baz.?() }` which is where we started. 
So why not keep `?` and maybe if there is a "methode macro' RFC and it gets accepted define `?` as a (very) special case of it... (so is there such an (pre)RFC?).

keeperofdakeys:
I haven't seen any full proposal on implementing method macros, just a vague idea being thrown around. There has been some other discussion about it in https://github.com/rust-lang/rust/issues/31436.

kennytm:
@dathinab There is an issue #676, no pre-RFCs yet.

Making `?` to be an identifier would require another RFC besides method-macro.

Mark-Simulacrum:
From the linked issue discussion on method macros:

[From nrc](https://github.com/rust-lang/rust/issues/31436#issuecomment-244644143):

> Just to note that I hate the idea of method macros and I can't think of a language feature I would oppose more strongly. They fuzz the phasing of the compiler, and unless we make really quite fundamental changes to the language there is no way they can exist and have unsurprising behaviour. They are also hard to parse sensibly and almost certainly not backwards compatible.

And a similar note [from withoutboats](https://github.com/rust-lang/rust/issues/31436#issuecomment-244581723).

nikomatsakis:
Overall, I'm pretty happy with `?` and not inclined to change it to anything.

That said: `?!` in particular has a "WTF" connotation to it that, while humorous, doesn't really seem suitable. (I can't tell if this is intentional?)

est31:
No, the "WTF" connotation is not intentional.

My main concern is that `?` is too short and can be overlooked too easily. Of course, its always a tradeoff between readability of the "normal" control flow and readability of the "return early" control flow, but I like the length of try!, and `?` alone makes the situation worse imo.

Any other ideas how a longer version of `?` could look like? Yes, method macros would use `try!`, but the name shouldn't be used imo when its a hardcoded operator. It would lead to confusion that the prefix `try!` is a macro and the postfix `try!` is an operator.

Mark-Simulacrum:
I think that a good reason for why "?" is significant enough is that I expect editors to change syntax highlighting to highlight it specially, and from what I can tell that should be good enough. @durka [seems to share this belief](https://github.com/rust-lang/rfcs/pull/1737#issuecomment-244563762).

As it relates to the RFC, `?!` does not feel like much of an improvement over `?` since it isn't significantly longer. I also believe that `try!` should be used for those cases where error handling is intended to be explicit; perhaps wrapping a chain of `?`-linked method calls.

sciyoshi:
I really like the current `?` operator (which mimics CoffeeScript's), but I also like the possibility of keeping it reserved for nullable types (`u32?` vs `Option<u32>`).

est31:
Note that it wouldn't be the first time to rename something during or close to stabilisation: https://github.com/rust-lang/rust/issues/27719#issuecomment-206998250

aturon:
I think this would be better served by discussion on the [tracking issue](https://github.com/rust-lang/rust/issues/31436) for `?`. We often make bikeshed-level decisions there.

FWIW, I also think it's very unlikely that this change will be accepted, since it cuts against much of the spirit of the `?` proposal. If we collectively believe that readability of `?` is a problem, we're likelier to stay with `try!` in some form, I suspect.

I'm going to close this RFC for the time being, in favor of tracking issue discussion.

est31:
> since it cuts against much of the spirit of the ? proposal

Wait, wasn't the main argument for `?` that it allows chaining? I mean the brevity was mostly a side effect no?

Still doesn't change my opinion  that `?` is too long and whose usage can in fact be dangerous, but seems this RFC won't be merged. `?` is still miles better than what other languages have, which is no sign of "stuff can early return here" at all.

ticki:
@est31 Eh, `?` isn't for chaining, it's for propagation. `?` is equivalent to `try!` in an extended form, and not `and_then`.

est31:
@ticki yes, I know that it is not `and_then`. The hidden early return is in fact part of the reason why i think `?` is too short.

With chaining I mean the use case outlined in this comment https://github.com/rust-lang/rfcs/pull/243#issuecomment-169423425 , aka builder pattern.

PR #1741, opened 2016-09-06
Delete stray quote char.

solson:


nagisa:
This quotation character probably shouldn’t have been deleted because it has a matching start quote at the start of the paragraph. 

---

None of this would’ve happened if we all used “proper” quotes too :)

solson:
Heh, I did not notice the entire code block was nested inside a quotation. It would be better to block quote the entire section.

PR #1752, opened 2016-09-18
Security advisories in crates.io

untitaker:
pre-RFC here: https://internals.rust-lang.org/t/pre-rfc-security-advisories-as-part-of-crates-io-metadata/4045

### DIFF
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -179,11 +179,11 @@ and then release the versions that contain the security fix.
   ``publish`` should refuse to upload my crate even if ``bar=1.2.3`` is not
   vulnerable, as another version satisfying that constraint may be.
 
-The author of a crate that directly depends on a vulnerable crate may disable
-these warnings with a switch in their `Cargo.toml`. If `iron==0.4.x` has
-an advisory with the ID `deadbeef`, the dependent author may use the
-`allow_vulnerable` parameter to disable all the above-described warnings and
-errors for this vulnerability:
+The author of a crate that directly depends on a vulnerable crate may still use
+vulnerable packages using switch in their `Cargo.toml`. If `iron==0.4.x` has an
+advisory with the ID `deadbeef`, the dependent author may use the
+`allow_vulnerable` parameter to disable the warnings for `build` and `install`
+and the errors for `publish` due to this vulnerability:
 
 ```
 [dependencies]
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -234,7 +234,7 @@ While this alternative meets the popular aesthetic preference of having generic
 commands with a large variety of usecases (over single-purpose commands), using
 yanking this way has a few drawbacks:
 
-- Cargo dosen't allow yanked packages to be used in new packages. In the
+- Cargo doesn't allow yanked packages to be used in new packages. In the
   author's opinion, people who know what they're doing should be allowed to
   depend on vulnerable packages, as they might use the crate in a way that
   poses no security threat.
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -141,9 +141,13 @@ advisory`, and then release the versions that contain the security fix.
 
 - `cargo test` make those warnings hard errors.
 
-- `cargo publish` will refuse to upload a crate if the latest version of a
-  direct dependency satisfying the constraints in `Cargo.toml` is vulnerable.
+- `cargo publish` will refuse to upload a crate if any version of a direct
+  dependency satisfying the constraints in `Cargo.toml` is vulnerable. 
   Indirect dependencies should not trigger this behavior.
+
+  For example, if I have a dependency such as ``bar = "^1.2.3"``, this means
+  ``publish`` should refuse to upload my crate even if ``bar=1.2.3`` is not
+  vulnerable, as another version satisfying that constraint may be.
 
 The author of a crate that directly depends on a vulnerable crate may disable
 these warnings/errors with a switch in their `Cargo.toml`. If `iron==0.4.x` is
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -15,23 +15,37 @@ versions during compilation.
 
 When compared to other ecosystems such as Python's, Rust's broader community
 prefers many single-purpose crates over larger monoliths. This situation,
-together with the strongly encouraged practice of pinning MINOR versions of
-dependencies, slows down the propagation of critical security fixes.
+together with the strongly encouraged practice of pinning version ranges or
+even exact versions of dependencies, slows down the propagation of critical
+security fixes.
 
-Assume a crate `W`, which depends on `X`, which depends on `Y`, which depends
-on `Z`.  If `Z` releases a new MINOR version including a security fix, it
-requires the attention of `Y`'s  and `X`'s maintainers to propagate that
-security fix to `W`. What makes this situation worse is that the author of `W`
-is never notified that they were running a vulnerable version of `Z` all the
-time.
+Here's an example. Assume that:
 
-Tooling that builds on top of the API provided by Crates.io (such as Cargo)
-could alert crate users of their vulnerabilities, which in turn spurs them to
-update their dependencies accordingly. Even if that does not happen, the
-additional metadata at least makes it clear which crates are potentially
-dangerous to use and which ones not. This not only helps Rust programmers, but
-potentially also distributors (such as packagers of Linux distros) and
-end-users.
+- There are two crates, `A` and `B`.
+- The latest version of `B` is `1.0.0`, which is a complete rewrite and comes
+  with an entirely new API. There are no bugfixes, but the API of `0.x` became
+  gradually more quirky, to a point that it had to be completely overhauled.
+- `A` depends on `B = ^0.9.0`, and did so before `B = 1.0.0` was released. The
+  author of `A` doesn't care about updating their code because why change a
+  running system?
+- The author of `B` doesn't officially support `0.9` anymore. No bugfixes or
+  security updates.
+
+If there is now a security issue in all existing versions of `B` (including the
+`0.9` and `1.0` series), `B` will get a security patch. However, that patch
+will only be available for `1.0` as a backport is costly. The author of `A` is
+never notified that they were running a vulnerable version of `B` all the time,
+unless they take extra measures to keep themselves informed. Especially in
+open-source development (not sponsored by a multinational company) that is
+unlikely to be the case.
+
+In any case it could be easier. Cargo and other tooling that builds on top of
+the proposed new API for Crates.io could alert crate users of their
+vulnerabilities, which in turn spurs them to update their dependencies
+accordingly. Even if that does not happen, the additional metadata at least
+makes it clear which crates are potentially dangerous to use and which ones
+not. This not only helps Rust programmers, but potentially also distributors
+(such as packagers of Linux distros) and end-users.
 
 # Detailed design
 [design]: #detailed-design
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -82,21 +82,23 @@ A correct invocation makes Cargo do the following:
 1. Cargo will open `$EDITOR` with a file ending with `.toml` that looks like
    this:
 
-      [vulnerability]
-      # Edit this as you wish
-      package = "mypackage"
-      versions = ["1.2.0", "1.2.3", "1.2.4", "1.2.5"]
+   ```
+   [vulnerability]
+   # Edit this as you wish
+   package = "mypackage"
+   versions = ["1.2.0", "1.2.3", "1.2.4", "1.2.5"]
 
-      # It is recommended to request a CVE at https://iwantacve.org/ and
-      # reference the assigned number here.
-      # cve = "CVE-YYYY-XXXX"
-      cve = false
+   # It is recommended to request a CVE at https://iwantacve.org/ and
+   # reference the assigned number here.
+   # cve = "CVE-YYYY-XXXX"
+   cve = false
 
-      # Enter a short-form description of the vulnerability here. Preferrably a
-      # single paragraph.
-      description = """
-      
-      """
+   # Enter a short-form description of the vulnerability here. Preferrably a
+   # single paragraph.
+   description = """
+   
+   """
+   ```
 
 2. After `$EDITOR` exits, Cargo validates the content. This includes verifying
    that:
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -57,32 +57,38 @@ page:
 
 ```
 $ cargo advisory --help
+Generate and upload security advisories for the given or the current crate.
 
 Usage:
     cargo advisory [options] -- [<crate>]
-    --vers VERSION      Versions to mark as vulnerable. Can be specified multiple times.
+    --filename PATH      The filename to use. Defaults to `./Advisory.toml`.
+                         If `-` is given, generated advisories are printed to
+                         stdout and advisories to upload are read from stdin.
+
+    --vers VERSION       Versions to mark as vulnerable. Can be specified
+                         multiple times. Only valid in conjunction with
+                         --generate.
+
+    --upload/--generate  Whether to upload or generate a advisory. The default
+                         is to generate. These options are mutually exclusive.
     [...]
 ```
 
-`advisory` has a similar CLI compared to `yank`.
+Like `yank` it takes a `--vers` option, with two differences:
 
-- It takes exactly the same positional arguments, defaulting to the crate in
-  the current working directory.
+- if a version is not specified, `advisory` will default to marking all
+  existing versions on Crates.io as vulnerable.
 
-- Like `yank` it takes a `--vers` option, with two differences:
+- Version ranges such as `<1.2.6, >1.0.0` can be specified. This is comparable
+  to the syntax used for specifying dependencies in the `Cargo.toml`, with the
+  exception that `x.y.z` is not equivalent to `^x.y.z`, but means the exact
+  version.
 
-  - if a version is not specified, `advisory` will default to marking all existing
-    versions on Crates.io as vulnerable.
+Here's the workflow:
 
-  - Version ranges such as `<1.2.6, >1.0.0` can be specified. This is
-    comparable to the syntax used for specifying dependencies in the
-    `Cargo.toml`, with the exception that `x.y.z` is not equivalent to
-    `^x.y.z`, but means the exact version.
-
-A correct invocation makes Cargo do the following:
-
-1. Cargo will open `$EDITOR` with a file ending with `.toml` that looks like
-   this:
+1. The user invokes `cargo advisory` without the `--upload` option. Cargo will
+   generate a file under `filename`. Cargo should abort if the file already
+   exists. The content looks like this:
 
    ```
    [vulnerability]
@@ -108,18 +114,17 @@ A correct invocation makes Cargo do the following:
    """
    ```
 
-2. After `$EDITOR` exits, Cargo validates the content. Some rules:
+2. The user invokes `cargo advisory --upload`. Cargo verifies the passed file
+   against the following rules:
 
-   - the file is valid TOML
+   - the file exists and is valid TOML
    - Optional keys may be either `false` or absent.
    - the `description` contains not only whitespace. More text than a paragraph
-     should be allowed, but not wished.
+     should be allowed, but not necessarily recommended.
    - `package` exists on Crates.io and the versions specified in `versions` exist
    - `dwf` is not an empty array. It should be ``false`` if there are none.
    
-   If not, Cargo should print an error message and wait for the user to either
-   hit enter or `^C`. In the former case, open the editor with the same file
-   again.
+   If not, Cargo should print one or more error messages and exit.
 
 3. When the vulnerability report is found to be valid, Cargo should print a
    summary, ask the user for confirmation and upload it to the package index.
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -25,12 +25,13 @@ security fix to `W`. What makes this situation worse is that the author of `W`
 is never notified that they were running a vulnerable version of `Z` all the
 time.
 
-An added API to Crates.io as described above would allow for the creation of
-third-party tooling that notifies the author of `Z` about security releases.
-
-The warning emitted by Cargo would further help downstream distributors (Linux
-packagers for example) and end users of Rust applications to identify potential
-risks in their usage.
+Tooling that builds on top of the API provided by Crates.io (such as Cargo)
+could alert crate users of their vulnerabilities, which in turn spurs them to
+update their dependencies accordingly. Even if that does not happen, the
+additional metadata at least makes it clear which crates are potentially
+dangerous to use and which ones not. This not only helps Rust programmers, but
+potentially also distributors (such as packagers of Linux distros) and
+end-users.
 
 # Detailed design
 [design]: #detailed-design
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -86,7 +86,6 @@ A correct invocation makes Cargo do the following:
 
    ```
    [vulnerability]
-   # Edit this as you wish
    package = "mypackage"
    versions = ["1.2.0", "1.2.3", "1.2.4", "1.2.5"]
 
@@ -98,9 +97,12 @@ A correct invocation makes Cargo do the following:
    # dwf = "CVE-YYYY-XXXX"
    # dwf = ["CVE-YYYY-XXXX", "CVE-ZZZZ-WWWW"]
 
+   # URL to a long-form description of this issue, e.g. a blogpost announcing
+   # the release or a changelog entry (optional)
+   url = false
 
    # Enter a short-form description of the vulnerability here. Preferrably a
-   # single paragraph.
+   # single paragraph (required)
    description = """
    
    """
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -13,14 +13,18 @@ versions during compilation.
 # Motivation
 [motivation]: #motivation
 
-When compared to other ecosystems such as Python's, Rust's broader community
-prefers many single-purpose crates over larger monoliths. This situation,
-together with the strongly encouraged practice of pinning version ranges or
-even exact versions of dependencies (as opposed to using `"*"` as a version
-identifier, which Crates.io rejects), slows down the propagation of critical
-security fixes.
+Keeping on top of security vulnerabilities for all dependencies of a typical
+Rust application is currently extremely hard, if not impossible. Particularly
+for Rust this task is relatively hard, since Rust's broader community prefers
+many single-purpose crates over larger "collections" of tools (like Boost in
+C++).
 
-Here's an example. Assume that:
+One might think that a regular `cargo update` may help in such situations.
+However, the application developer still does not know which updates were
+security updates, and which were not, and therefore doesn't know if a new
+release of their application is needed or not. `cargo update` doesn't even
+suffice to automatically recieve all security updates. Here's an example.
+Assume that:
 
 - There are two crates, `A` and `B`.
 - The latest version of `B` is `1.0.0`, which is a complete rewrite and comes
@@ -40,13 +44,13 @@ unless they take extra measures to keep themselves informed. Especially in
 open-source development (not sponsored by a multinational company) that is
 unlikely to be the case.
 
-In any case it could be easier. Cargo and other tooling that builds on top of
-the proposed new API for Crates.io could alert crate users of their
-vulnerabilities, which in turn spurs them to update their dependencies
-accordingly. Even if that does not happen, the additional metadata at least
-makes it clear which crates are potentially dangerous to use and which ones
-not. This not only helps Rust programmers, but potentially also distributors
-(such as packagers of Linux distros) and end-users.
+Cargo and other tooling that builds on top of the proposed new API for
+Crates.io could alert crate users of their vulnerabilities, which in turn spurs
+them to update their dependencies accordingly. Even if that does not happen,
+the additional metadata at least makes it clear which crates are potentially
+dangerous to use and which ones not. This not only helps Rust programmers, but
+potentially also distributors (such as packagers of Linux distros) and
+end-users.
 
 # Detailed design
 [design]: #detailed-design
@@ -219,7 +223,8 @@ semver-incompatible with the latest one are usually aware that they should
 eventually update (and in fact there's already tooling to keep on top of that,
 such as [cargo-outdated](https://github.com/kbknapp/cargo-outdated)), but, like
 in the example in [Motivation](#motivation), don't yet have a good reason to do
-so. A security issue would be a good reason.
+so. A security issue would be a good reason, but marking a package as
+deprecated does not imply that.
 
 ## Extending yanking for security advisories
 
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -201,7 +201,9 @@ still print warnings:
   ["deadbeef"]` set.
 
 
-Cargo must reject nonexistent vulnerability IDs with a fatal error.
+Note that `deadbeef` represents the vulnerability ID assigned by Crates.io, not
+a CVE or DWF ID. Cargo must reject nonexistent vulnerability IDs with a fatal
+error.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-security-advisories.md
+++ b/text/0000-security-advisories.md
@@ -108,11 +108,10 @@ A correct invocation makes Cargo do the following:
    """
    ```
 
-2. After `$EDITOR` exits, Cargo validates the content. This includes verifying
-   that:
+2. After `$EDITOR` exits, Cargo validates the content. Some rules:
 
    - the file is valid TOML
-   - none of the table keys have ben removed
+   - Optional keys may be either `false` or absent.
    - the `description` contains not only whitespace. More text than a paragraph
      should be allowed, but not wished.
    - `package` exists on Crates.io and the versions specified in `versions` exist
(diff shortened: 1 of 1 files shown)


text/0000-security-advisories.md
```diff
@@ -0,0 +1,259 @@
...
+  versions.
+
+- Version ranges such as `<1.2.6, >1.0.0` can be specified. This is comparable
+  to the syntax used for specifying dependencies in the `Cargo.toml`, with the
+  exception that `x.y.z` is not equivalent to `^x.y.z`, but means the exact
+  version.
+
+Here's the workflow:
+
+1. The user invokes `cargo advisory` without the `--upload` option. Cargo will
```
EPashkin: may be --generate here?

text/0000-security-advisories.md
```diff
@@ -0,0 +1,259 @@
...
+should offer a new command for the same purpose and warn about vulnerable crate
+versions during compilation.
+
+# Motivation
+[motivation]: #motivation
+
+When compared to other ecosystems such as Python's, Rust's broader community
+prefers many single-purpose crates over larger monoliths. This situation,
+together with the strongly encouraged practice of pinning MINOR versions of
+dependencies, slows down the propagation of critical security fixes.
```
gkoz: I'm not aware of pinning minor versions in `Cargo.toml` being encouraged (except the `0.x.y` case but then `x` is the major version for cargo-semver purposes). Can you elaborate on this?

text/0000-security-advisories.md
```diff
@@ -0,0 +1,259 @@
...
+
+Assume a crate `W`, which depends on `X`, which depends on `Y`, which depends
+on `Z`.  If `Z` releases a new MINOR version including a security fix, it
+requires the attention of `Y`'s  and `X`'s maintainers to propagate that
+security fix to `W`. What makes this situation worse is that the author of `W`
```
gkoz: If `X` and `Y`use the (default) caret requirement, performing `cargo update` on `W` will update `Z` (once again, for `0.a.b` MINOR version is `b` as far as caret requirement is concerned).

text/0000-security-advisories.md
untitaker: The idea behind that awkward phrasing is that both `cargo advisory` and `cargo advisory --generate` are valid for this step. "without --upload" is a phrasing that captures both.

(damn "start a review" button)

text/0000-security-advisories.md
untitaker: You're right about that, I only thought about the `0.x.y`-case, but changing MINOR to MAJOR in that sentence is also wrong. What about this:

>...with the strongly encouraged practice of pinning version ranges (sometimes specific versions)

Specific versions are pinned if a semver-compatible version of a dependency introduces a bug that breaks the dependent crate.

text/0000-security-advisories.md
untitaker: What about "semver-incompatible version"? Also I probably should clarify what the version pinning for all crates looks like. I also made the assumption that each crate exposes its dependency's API and therefore needs to do a new semver-incompatible release when their dependencies do.

text/0000-security-advisories.md
gkoz: I feel this bit of motivation doesn't work because
 * pinning to specific versions is not the norm,
 * cargo makes the ultimate consumer track precise versions via lockfile, giving them more power and responsibility.

So applying the fix should normally not be a problem. Only if the intermediate crates actively prevent this by strict version pinning do you need to act differently (e.g. prevent publishing a crate, which makes it impossible to avoid vulnerable dependency versions).

text/0000-security-advisories.md
gkoz: I guess if the fix requires a major version bump, you get the same scenario as with pinned versions.

text/0000-security-advisories.md
untitaker: In the [pre-pre-RFC thread I gave a different motivation](https://internals.rust-lang.org/t/idea-security-advisories-as-part-of-crates-io-metadata/3899). Money quote:

>Unfortunately I'm depending on open source libraries where nobody gets paid for their work. This means that the maintainer of Z is less likely to backport the patch to other MINOR versions, only the latest version gets a PATCH release. The maintainers of X and Y are less likely to update their dependencies either, because they do all of this in their freetime as well. This is a social problem though, and people are just getting started12 writing about this issue.

I agree with the motivation being a bit weak. Do you think the scenario described in that thread above is more realistic?

untitaker: I rewrote the entire motivation. Better?

gkoz: I think it's worth separating two concerns: notifications and dealing with upgrades.

Regardless of any semver issues it's important for end-users (crate `W` author from the `W->X->Y->Z` example) to be notified of any dependencies being vulnerable. This need stands on its own and might be the biggest motivator.

Then once you're aware of the vulnerabilities there's the case of `cargo update` not working for some reason (major version bump, version pining, etc). Whatever the reason, one might want cargo or crates.io to impose some limitations in this situation, but perhaps not bother the majority of crate authors who don't pin minor versions.

untitaker: I don't understand what you mean. Dealing with upgrades is not concern of this RFC at all.

Also the new, completely rewritten example doesn't literally mention minor versions and avoids general statements about pinning.

gkoz: I guess we're back to why you think there's

> strongly encouraged practice of pinning version ranges or even exact versions of dependencies

:)

untitaker: Well, as opposed to `"*"` as version specifier, which Crates.io rejects.

untitaker: I updated the text to concretely mention why pinning is encouraged. Note that I wrote "version ranges" (e.g. `^1.0.0`) as well.

gkoz: > as opposed to `"*"` as version specifier, which Crates.io rejects

I see, it's the "pinning" part because I wouldn't consider "^1.0.0" pinning, it implies something like "=1.0.0" to me. But is it right treat them the same way? I don't think "^1.0.0" poses any propagation problems. Dealing with more exotic cases is important but probably won't be the primary focus of this feature.

> I don't understand what you mean. Dealing with upgrades is not concern of this RFC at all.

It is because you're opening with slow propagation of security fixes. It's probably stronger to first discuss only the communication part: all users need to be made aware of the advisory. (BTW I wouldn't mind `cargo` outright failing the build by default if any dependency is found to be vulnerable.) The propagation discussion can build on that but the feature is valuable even without it.

gkoz: Perhaps I'm not taking seriously enough the scenario of a "1.x" branch being unmaintained and vulnerable (i.e. `cargo update` definitely won't work).

untitaker:
The new example I wrote in https://github.com/rust-lang/rfcs/pull/1752/commits/06259c1d4b191fff37011155aed98fd5a63054d0 doesn't imply that...

text/0000-security-advisories.md
untitaker: The new example I wrote in https://github.com/rust-lang/rfcs/pull/1752/commits/06259c1d4b191fff37011155aed98fd5a63054d0 doesn't imply that.

gkoz:
Well it seems like you're solving a pretty specific problem (a discontinued `B 0.9` is found out to be vulnerable and `A` depends on `B 0.9` and won't upgrade), which is an extension of more general goals (it's by itself valuable to be notified if you got a `B 0.9` in your lockfile).

untitaker:
No, that's not the problem I'm trying to solve, that's a pretty common *example* that exhibits that problem, where the entire issue is not just solved by running `cargo update` regularly. The problem is that people don't know if their dependencies contain security vulnerabilities, and that it's currently hard to keep on top of that.

arielb1:
Enterprise users might only want to do `cargo update` at Designated Release Points except for security issues. Automated security notifications would remind them to update.

untitaker:
Is this something I should write into the Motivation? I feel like it's a bit
dishonest as I wouldn't know of any current enterprise users of Rust, so that
claim is hard to verify.

untitaker:
@gkoz @arielb1 I rewrote the motivation again in https://github.com/rust-lang/rfcs/pull/1752/commits/06d5174ffa25d6b3ae77b2c1718f1fbe2982490d. I think this clarifies the problem this RFC is trying to solve without making any unnecessarily strong claims. What do you think?

text/0000-security-advisories.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Extending yanking for security advisories
+
+It has been proposed to [extend the semantics of yanking such that it could be
+used for security advisories](https://github.com/rust-lang/cargo/issues/2608).
+While this alternative meets the popular aesthetic preference of having generic
+commands with a large variety of usecases (over single-purpose commands), using
+yanking this way has a few drawbacks:
+
+- Cargo dosen't allow yanked packages to be used in new packages. In the
```
sdboyer: typo: "dosen't"

gkoz:
@untitaker thanks, it looks great now!

tarcieri:
This looks good to me on my first pass through

untitaker:
Next week it'll be a month this PR has been open. I understand that this is hard to decide on but I suspect that it is also slipping through the cracks.

arielb1:
@untitaker 

I think that an "incompatibility" feature is useful for other reasons than security - yanking indicates that a *specific* version contains some brokenness than its predecessor does not and must not be updated to, while "incompating" indicates that a whole range of versions contains some *newly-exposed* brokenness (because of a security issue, API server shutting down, rustc bugfix, other non-Cargo dependency, etc.) and must be upgraded *from*.

untitaker:
@arielb1 I chose to only mark specific versions as vulnerable because marking the entire "0.x" version range as vulnerable prevents me from backporting my security fix in a new "0.x.y". We can totally expand the report format to allow version ranges to have a more expressive format, but I do think that those version ranges should be (internally) expanded to an array of specific versions when the vulnerability report is submitted.

eternaleye:
@unitaker: IMO, that fails a very important criterion for a security measure, which is that it "fails closed" - that is, if the user takes _less_ effort, the _more_ secure result occurs.

With the specific-version approach, the "correct" action (mark all vulnerable versions, omit safe versions) is a large amount of effort - one must invoke the command for each version. Failing to do this results in vulnerable versions remaining unmarked.

With the blanket-range approach, the "correct" action is a relatively small amount of effort (`<0.7` or whatever). Failing to specify results in safe versions being erroneously marked, but that is safe in the security sense (if inconvenient).

Moreover:

> marking the entire "0.x" version range as vulnerable prevents me from backporting my security fix in a new "0.x.y" at some later point

Only with a naive implementation. The _semantic meaning_ isn't "mark the entire 0.x range as vulnerable" - it's "mark all releases that _have been made_ in 0.x as vulnerable". If the command internally does the necessary iteration to produce a tightly-scoped list of exact vulnerable versions, the safe behavior I described above is preserved, without causing the problem you describe.

untitaker:
By default all published versions are marked vulnerable and single ones can be
edited out. This is in the RFC already.

eternaleye:
@untitaker: Yeah, realized just after posting and edited my comment.

tarcieri:
> I think that an "incompatibility" feature is useful for other reasons than security

@arielb1 I hope the above exchange between @untitaker and @eternaleye helps illustrate why colluding security features with non-security features is a bad idea.

I know this sort of feature might feel somewhat general, and for reasons of parsimony you might want to couple it to non-security features. However, as soon as non-security-related requirements start getting mixed up into what is supposed to be a security feature, I feel they become actively harmful to security use cases.

I remain strongly opposed to trying to leverage existing non-security related features for this purpose, or tack-on non-security use cases as an afterthought. I think non-security use cases should be addressed separately, and if not, it will be detrimental to the security use cases.

vi:
Questions:

* Shall it be possible to undo uploading an advisory?
* Shall there be `--dry-run`?
* Is granularity finer than crate-level a good idea? If can be just one function used by minority of dependent crates that is vulnerable.
* Shall uploading an advisory without version range ("default to all existing versions") issue a warning?
* Shall some gradual disclosure scheme where important dependent crates and maintainers receive notification about the advisory first and all the rest receive it later (after fix already deployed)?
* Shall moderators of crates.io be able to file or grant the right to file advisories for non-owners of the crate?

> `cargo publish` will refuse to upload... 

Shall it although allow uploading crates with `allow_vulnerable = ...`?

Imagine big old no-so-popular single-crate abandoned (with no successor) framework full of arcane formuli no "mere mortal" users can comprehend. Although there has been no update to the framework for 5 years, new apps using it still appear once in a while.

But suddently a workaroundable security bug that affects 20% of apps using the framework... Somebody files an advisory. Now all those apps are broken (crates.io don't accept new apps and versions) and no clear way to fix (because of there will be no fixed version of framework - it is ever-vulnerable).

> Optional keys may be either false or absent.

The phrase looks not clear. Does it mean "if optional key is not absent, it must be `false`" or "ommitted optional keys may be absent or have value `false`"? Probably meant "If there is no url or dwf to specify, just omit the key or use value `false`" or "Even if there is no url or dwf to specify, the key should exists and have value `false`".

untitaker:
On Wed, Oct 19, 2016 at 06:00:29PM -0700, Vitaly Shukela wrote:
> Questions:
> 
> * Shall it be possible to undo uploading an advisory?

Crates.io could possibly offer this in the web-UI but I don't consider it
important to have in the initial implementation.

> * Shall there be `--dry-run`?

What would it do though? Do you mean a command that validates the edited
report?

> * Is granularity finer than crate-level a good idea? If can be just one function used by minority of dependent crates that is vulnerable.

Another feature that would be nice to have but which I don't consider essential
for now.

> * Shall uploading an advisory without version range ("default to all existing versions") issue a warning?

The file generated by cargo already contains all versions. I guess you mean the
case where the user empties the array. I've pushed a commit for that.

> * Shall some gradual disclosure scheme where important dependent crates and maintainers receive notification about the advisory first and all the rest receive it later (after fix already deployed)?

Also nice to have, but again, let's keep this stuff out for simplicity's sake.

> * Shall moderators of crates.io be able to file or grant the right to file advisories for non-owners of the crate?

Yes, though moderator intervention is something that already exists and doesn't
require explicit mention in the RFC IMO. Takeover of old broken crates
everybody depends on is something that has to be decided on a case-to-case
basis.

> 
> > `cargo publish` will refuse to upload... 
> 
> Shall it although allow uploading crates with `allow_vulnerable = ...`?

Yes that was the intention. Particularly this quote from the RFC:

>the dependent author may use the `allow_vulnerable` parameter to disable all
>the above-described warnings and errors for this vulnerability


> 
> Imagine big old no-so-popular single-crate abandoned (with no successor) framework full of arcane formuli no "mere mortal" users can comprehend. Although there has been no update to the framework for 5 years, new apps using it still appear once in a while.
> 
> But suddently a workaroundable security bug that affects 20% of apps using the framework... Somebody files an advisory. Now all those apps are broken (crates.io don't accept new apps and versions) and no clear way to fix (because of there will be no fixed version of framework - it is ever-vulnerable).
> 
> > Optional keys may be either false or absent.
> 
> The phrase looks not clear. Does it mean "if optional key is not absent, it must be `false`" or "ommitted optional keys may be absent or have value `false`"? Probably meant "If there is no url or dwf to specify, just omit the key or use value `false`" or "Even if there is no url or dwf to specify, the key should exists and have value `false`".

Pushed a commit for this.

vi:
Re-read and understood: refusal to publish happens on client (cargo) side, not on server (crates.io) side.

untitaker:
I'm not sure there's a semantic difference, allow_vulnerable can be read both
by the server and the client.

Of course it would probably be better to enforce this on the server such that
the package author can't use a modified cargo to publish their vulnerable
package. But I don't think that the user playing themselves like that is in our
threat model.

vi:
The main thing that this rule (not just warnings printed to console) is overridable by explicit listing of advisory IDs.

untitaker:
I don't understand what you mean, the behavior can be the same on server and
client?

vi:
There are two effects of using vulnerable crate as a dependency:

1. Warnings printed to console;
2. Refusal to upload.

I'd explicitly written that `allow_vulnerable` should grant exception not only from "1.", but also from "2.". (Currently it can be deduced from "... disable all the above-described warnings **and errors** for this vulnerability ...", with the only error described is refusal to upload)

If "2." will be checked on not only on client (cargo), but also on server (crates.io), what shall it do about old Cargo clients without this feature? Or shall server-side advisory checking be in scope of further separate RFCs?

text/0000-security-advisories.md
```diff
@@ -0,0 +1,297 @@
...
+overloading in Java).  And the name "yank" wouldn't make sense for crate
+versions that may still be available (depending on the "reason").
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## DWF vs CVE
+
+- It may be counterintuitive that one can specify CVEs in the DWF parameter.
+  Should it be called ``cve`` instead even though it can also be used for DWFs?
```
troplin: IMO the field should be called `id` or `ids`.

Also, there should be some kind of (enforced) namespacing for the IDs, just to be future proof. By this I mean a whitelist of prefixes that are accepted by crates.io.

text/0000-security-advisories.md
```diff
@@ -0,0 +1,297 @@
...
+Comparison:
+
+- CVEs are more popular
+- Applying for a CVE number is a manual process and requires review by a human.
+  DWFs can be automatically managed assigned by Crates.io
+
+## What to do if dwf = false
+
+- Crates.io could apply for blocks of DWF IDs and automatically assign them if
+  the user didn't specify one in the advisory (``dwf = false``).
```
troplin: Why not simply require at least one ID?
I think the uploaded TOML should already be complete and not be changed by the server.
There should be an easy way to get an ID beforehand though. For example `--generate` could already assign an ID which can then be overridden. If the auto-assigned ID is not used within certain timespan, it can be reused.

vi:
Shall `Advisory.toml` be under Git or in `.gitignore`?

text/0000-security-advisories.md
untitaker: CVE and DWF requests already have to contain descriptions and affected versions, neither of which is known when generating the template for a vuln report.

text/0000-security-advisories.md
```diff
@@ -0,0 +1,297 @@
...
+   
+   """
+   ```
+
+2. The user invokes `cargo advisory --upload`. Cargo verifies the passed file
+   against the following rules:
+
+   - the file exists and is valid TOML
+   - When an optional key is `false`, this is semantically equivalent to it
+     being omitted.
```
untitaker: In general I'd say keys are optional if they have a default value. And that default value is false for the optional keys `dwf` and `url`.

text/0000-security-advisories.md
untitaker: Instead of namespacing you might as well add more keys for different types of IDs to avoid overloading one key.

untitaker:
@vi I fixed the wording around allow_vulnerable.

`Advisory.toml` should probably be in the `.gitignore` (or even created outside of the repo) if we have a fixed filename and if we only use it as a temporary file.

vi:
Shall `cargo new` and `cargo init` automatically add `Advisory.toml` to `.gitignore`? Shall `cargo advisory` do it?

text/0000-security-advisories.md
eternaleye: Then why have a default value with a different _type_ than the "real" value? What is gained by using `false` and explicitly ruling out an empty array? It seems to me that the only effect of overloading the type in this manner is to make the schema more painful to work with (especially for statically-typed languages).

Ericson2314: I'm insufficiently familiar with how CVE and DVF numbers are assigned, but this `false` business is very unidiomatic. simply skipping optional fields is idiomatic.

> dwf is not an empty array. It should be false if there are none.

Much easier to make `dwf` a required field and use `[]`.

text/0000-security-advisories.md
troplin: Adding more keys might pose a forward compatibility issue. Namespaces are just a convention how to format a string ID. Strings can be compared even if the concrete namespace is unknown. Unknown keys should are usually rejected or ignored. You cannot just treat every unknown key as an additional ID key, that would severely hurt the extensibility of the format.

text/0000-security-advisories.md
troplin: This makes no sense to me. If an advisory has no URL, then it just doesn't have an URL, so just leave it away.
But it certainly does **not** have a default URL with the value `false`.

This feels like some ugly JavaScript where you can use strings and booleans interchangeably, with arbitrary conversion rules.

text/0000-security-advisories.md
untitaker: In that case I'd rather have a TOML-subtable called `ids` than an array of strings where the strings contain custom syntax.

text/0000-security-advisories.md
untitaker: I agree with you @Ericson2314 and @eternaleye but you're criticizing an entirely different part of this document. This is about *behavior of the server* when dwf is empty/false, you're arguing about syntax.

untitaker: Changed typedef of dwf field.

untitaker:
>Shall cargo new and cargo init automatically add Advisory.toml to .gitignore? Shall cargo advisory do it?

Definetly not `cargo advisory`, for the simple reason that `cargo build` also doesn't append `target` to gitignore.

>Is information from Advisory.toml expected to be also duplicated inside crate.io's internal state? If yes then way is the source of truth: crates.io or Advisory.toml on Github?

I would understand users who want to host security advisories in their own infra as well. In practice there will already be duplication between CVE descriptions, Crates.io descriptions and maybe security announcements of the project's blog.

In either case we definetly shouldn't include `Advisory.toml` in the uploaded tarball.

untitaker:
@vi I added your questions to the end of the document for lack of a better issue tracking mechanism. I left the IMO very optional features and IMO answered questions out, but feel free to point out ommissions.

text/0000-security-advisories.md
troplin: That would be overly complex IMO. IDs should be unique without considering the subkey.

If one one advisory has:
```toml
ids = { foo = "DEADBEEF" }
```
and another one has:
```toml
ids = { bar = "DEADBEEF" }
```
then, what would the following mean:
```toml
allow_vulnerable = ["DEADBEEF"]
```

Or do you also want the subkeys there like:
```toml
allow_vulnerable = [{foo = "DEADBEEF"}]
```
What's the advantage over a simple convention how to format the string?

untitaker: Crates.io assigns each vulnerability report a unique ID, which is the only way you can reference vulnerability reports in `allow_vulnerable`. You wouldn't use DWFs or CVEs. The reason is that you may reference one CVE in multiple vulnerability reports, or multiple CVEs in one report. If you would then use CVEs in `allow_vulnerable` you'd have a lot of ambiguity. 

Sorry for the confusion! I'll look into fixing this.

vi:
How should Cargo get information about new advisories (especially when there are no other reasons to reach Internet).

Options:

1\. Only when doing `Updating registry ...`
2. On `cargo build` (or `run` or `test`), maybe with additional things
3. Only explicitly.

Additional things:
1. Not sooner than N hours after the last [successful] update
2. In background (postponing the actual advisory checking until the next `cargo build`)

What do to with a failure to connect to Internet - a warning, an error or nothing?

Should there be a separate option of `cargo advisory` to explicitly download the list of advisories for current project?

I propose mixed mode as a default:

1. Adviosry list is fetched when updating registry or when doing explicit command for this;
2. `cargo build` only uses advisory list, not downloads it;
3. `cargo build` prints a warning mentioning the explicit command if advisory list is not updated for N days.

tarcieri:
@vi what you're proposing sounds good to me. Point #3 is a fun one which also comes up in The Update Framework (#75), which tries to enforce guarantees around the timeliness of updates.

vi:
How global should be cached the vulnerability list? Per-project or per-cargo-deployment?

Shall it keep global list of crates (from multiple projects) to keep informed about advisories, so they got checked in one move?

Shall it just download all new vulnerabilities for all crates (not scalable)?

With unmodified point "3." (especially if the list is per-project) users may get too many warnings: doing occasional little fixes on many mini-projects => getting the warning almost every time.

Alternative option:

3b. `cargo build/run/test`: If advisory list is not updated for N days, start updating it in background (while continuing compilation and testing and running, etc.). If the list gets successfully downloaded, check against it after `cargo test/run/build` main run. If can't download, print a short warning. The warning about an advisory should be bigger, redder and scarier than the warning about inability to download the advisory list.

Something like this:

```
$ cargo test
    Downloading advisory list for dependent crates in background...
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs
     Running target/debug/rt-10dd314f496c8c5a

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured
warning: unable to download advisory list
```

vi:
Shall there be advisories for Rust system components (like libstd or libcore)?

tarcieri:
@vi the [ruby-advisory-db](https://github.com/rubysec/ruby-advisory-db) tracks similar vulnerabilities in the Ruby standard library. I think it would be nice to have functionality-wise, but seems a bit orthogonal to advisories for specific crates, and might be something to punt on initially in the hopes of getting anything out the door.

eternaleye:
@tarcieri, @vi: In addition, the push towards [stdlib-aware Cargo](https://github.com/rust-lang/rfcs/pull/1133) (and the possible future of much of the stdlib being on crates.io natively) may make it a moot point.

untitaker:
1.  I don't think adding advisories for the stdlib is within scope, as the stdlib doesn't have a independent versioning scheme and it's unclear how `allow_vulnerable` is supposed to work with it. Also with the stdlib as-is (and without stdlib-aware cargo), this would push the entire logic into rustc rather than having it in cargo.
2.  I didn't make this clear, but I think forcing a fetch of the advisory list on `cargo publish` is enough. This list is also updated with `cargo update`, but not `build`. A separate command would be nice, but having warnings if the list is old seems very intrusive to me.

  What does Cargo do with yanked packages?

vi:
> I didn't make this clear, but I think forcing a fetch of the advisory list on `cargo publish` is enough. 

What about private Rust projects that are never published to crates.io? Shouldn't they also benefit from advisories?

That's why advisories also make sence for crates without reverse dependencies.

Also `cargo publish` is a point where developer may think that the job is almost done, and an advisory may knock it back into designing and architecting. With `cargo build`-triggered warnings one may just switch away from problematic dependency earlier.

tarcieri:
> What about private Rust projects that are never published to crates.io? Shouldn't they also benefit from advisories?

My inclination is no, or rather, if they'd like to use this feature, they should run an internal Cargo repository with something that speaks the same API as crates.io (similar to [Gemstash](https://github.com/bundler/gemstash)) and can thereby manage advisories for internal crates.

untitaker:
>With cargo build-triggered warnings one may just switch away from problematic dependency earlier.

Network access during compilation is a dealbreaker for many bigger Linux distributions, which is why I'm trying to avoid it. For the dev workflow it would be better of course.

vi:
There may be `--offline` option or distributions may use local crates.io-like interface to its packages.

alexcrichton:
Reading over this, I unfortunately don't have a lot of time to follow the conversation and help push this over the finish line, but some thoughts I had were:

* I feel like the alternative with yanking isn't suitably explored. As you've mentioned many of the problems you listed are simply fixed by adding a yank reason, and I think the remaining problems would require an arguably similar delta of work/understanding the current RFC proposes from the current behavior of `cargo` to implement. That is, I [agree with](https://internals.rust-lang.org/t/pre-rfc-security-advisories-as-part-of-crates-io-metadata/4045/18?u=alexcrichton) @wycats that it may be best to explore that route before adding more infrastructure to Cargo with a whole new subcommand.
* I'm very worried about the warning behavior here, so much so that I could consider it a hard blocker for this feature. The proposal is to *unconditionally warn* on all builds about vulnerable packages with an optional whitelist in `Cargo.toml` to turn off warnings. The RFC then later claims that this feature could be used for vulnerabilities that may affect optional functionality. This to me sounds like I'm going to start getting warning ridden builds in Rust for functionality I never use. This seem to me like very much the wrong default and would simply condition everyone to ignore these warnings (defeating their purpose in the first place)
* The `allow_vulnerable` key seems very underspecified. In one location it's a string array, in another it's a boolean, and it's also not the conventional separated-with-dashes that Cargo configuration typically has.

> Next week it'll be a month this PR has been open. I understand that this is hard to decide on but I suspect that it is also slipping through the cracks.

I'm sorry that this is taking awhile, but please remember that we're all very busy with lots of projects in flight, so we don't always have a lot of time to allocate to all RFCs coming down the pike.

untitaker:
>I feel like the alternative with yanking isn't suitably explored.

The assertion I and @tarcieri have made in this and other threads is that overloading yank with a multitude of "yank reasons" compromises the security aspect of the proposed feature. [Tony said it better than I could](https://github.com/rust-lang/rfcs/pull/1752#issuecomment-254951280). My personal viewpoint is that cargo should provide a panic button for security incidents where the user is guided through proper disclosure, not a swiss-army knife where the answer to "What do I do next?" is "it depends, there are endless possibilities". Whether this RFC achieves that is left to discuss.

>The proposal is to unconditionally warn on all builds about vulnerable packages with an optional whitelist in Cargo.toml to turn off warnings. The RFC then later claims that this feature could be used for vulnerabilities that may affect optional functionality. This to me sounds like I'm going to start getting warning ridden builds in Rust for functionality I never use.

I'm very confused about this reasoning. If a dependency has a vulnerability, and I can disable the corresponding warning with a whitelist, how does this generate warning ridden builds? The purpose of the whitelist is to turn off security warnings that affect featuresets I don't use.

>The allow_vulnerable key seems very underspecified. In one location it's a string array, in another it's a boolean, and it's also not the conventional separated-with-dashes that Cargo configuration typically has.

That's totally fair criticism. The bool value is a leftover from older drafts and the preference for underscores is my own bias given that there are no keys with dashes except for package names in the TOML. In either case it's not a conceptual criticism.

seanmonstar:
There exists `dev-dependencies`.

untitaker:
Fair enough, fixed.

tarcieri:
@alexcrichton I'm not necessarily opposed to `yank` being a means to an end here, and indeed yanking crates with vulnerabilities makes sense, so having a separate command may be redundant.

However, the goal of a feature like this is to build a security advisory database which security-related tooling can consume. One like this (except as an integrated part of https://crates.io as opposed to a git repo of YAML files on GitHub):

https://github.com/rubysec/ruby-advisory-db

More than just eliminating security vulnerabilities during development, such a system can aid security teams in monitoring deployed versions of apps containing vulnerable crates with the assistance of endpoint agents.

If this can be accomplished with adding a set of security-related flags to `yank`, great, but I'm not sure that actually makes sense.

Security advisories by nature span multiple versions (while still being associated with a single logical CVE/DWF-like ID), but yank (as far as I can tell) only operates on one version at a time. That feels like an impedance mismatch to me: I think I'd rather the filing of an advisory carry an optional flag to yank the affected crates. Perhaps `yank` can be modified to consult the advisory for the affected versions so you can omit the `--vers` flag, but at that point, is it really the same command?

vi:
I don't like fusing of yanking and advisories.

* Yanking seem to be better usef for retracting crates quickly after they are published. Yanking things that was published long ago is more destabilizing.
* Security vulnerabilities may be found and filed, but not fixed. Yanking all versions of a crate without providing non-yanked version seems to be too breaking.
* With yanking for security, old versions of big important frameworks may become increasingly yanked. If one expect that 2-year-old versions are 70%-probable to be yanked because of security, overriding the yanking though `Cargo.lock` hacks may become a norm. One may need to try various old versions to bisect bugs or for doing gradual, step-by-step upgrades when reiving something long abandoned. I want projects to be generally buildable even with `Cargo.lock` removed.
* Applications may not be affected by a particular vulnerability. Having vulnerability ID in special "we have checked that it's OK" is more explicit reasoning about security.
* Advisories may have more didactic effect, making programmers think more about security. Cargo may make the first move, reifying abstract far away security problems into an actual potential vulnerability in *your* project.
* Advisory tracking system can be used by more things, not just Rust/Cargo. For example, somebody may want an unmaintained long-running program to automatically shutdown if some even potential vulnerability is found.

brson:
Although this seems like a reasonable thing to want, I'm wary to devote a lot of infrastructure to this use case without having more evidence that it's needed now and the design is right.

There are a number of specifics here and it seems like there's a fair opportunity to overdesign something that ends up being the wrong design in practice.

Text in the RFC indicating precedent in other package managers, their success, and how this design relates to them, would bolster the argument.

A more general, and more incremental approach that doesn't preclude adding a full security advisory feature as specified here, might be "yank with message".

Another incremental approach would be to develop something out-of-tree and prove it out, before trying to add the feature to cargo. As an example, I'm not sure that there are even enough crates with known security vulnerabilities to warrant the infrastructure necessary to add a feature to cargo to let one publish advisories from the command line. We could start by just setting up a repo on github containing a blacklist, which people can submit PRs to when security problems become known. Then either cargo itself or a `cargo advisory` plugin could source that list and print notifications. Implementing this would be a simple step that lets the concept prove itself without requiring cargo commit to a specific user interface.

tarcieri:
@brson one of the Bundler/RubyGems maintainers is working on a similar feature (cc @indirect), so hopefully you'll have a data point there

tarcieri:
> Another incremental approach would be to develop something out-of-tree and prove it out, before trying to add the feature to cargo.

I would agree it might be best to start out by developing this out-of-tree. Maybe just an advisory database on GitHub you contribute to with pull requests storing the advisory data format, and a CLI tool you can `cargo install` to clone that repo and check it against your `Cargo.lock`.

untitaker:
Since so many core members of Rust are against this only form that I consider viable I'm going to close this pull request. I agree that developing this feature in a separate repo outside of crates.io is good as PoC, but I don't think that a separate project would attract the amount of users necessary to make it actually useful.

If anybody wants to work with me on this feel free to email me and I'll be happy to help wherever I can, but my motivation to e.g. implement this feature all by myself (which I initially had) has faded away.

tarcieri:
If anyone is interested in continuing this work, I've created a project to continue pursuing it:

https://github.com/rustsec/cargo-advisory

At my job we're looking at using Rust in a "high assurance" capacity, so vulnerability tracking is definitely something I'd love to see tackled.

PR #1784, opened 2016-11-03
Differentiate between the different uses of "windows"

shepmaster:
Throw in some proper noun and minor typo fixes as well.


alexcrichton:
Thanks!

PR #1787, opened 2016-11-06
Automatic features.

withoutboats:
[Rendered](https://github.com/withoutboats/rfcs/blob/conditional_dependencies/0000-automatic_features.md)


withoutboats:
This proposal originates from the conversation [in this thread](https://internals.rust-lang.org/t/vague-proposal-extending-coherence-with-workspaces/3216), about how to make the orphan rules less of a bummer.

This specifically solves the problem of "I want to implement some traits for my type, but I don't want my users to *have* to add this dependency / *have* to fiddle with features" problem. It *does not* solve the "this type in a library I'm using should implement this trait, but the author didn't provide that implementation" problem, which is much trickier.

The idea behind this RFC was mine, and I think it is the best solution to a problem that is happening right now, but I don't feel super motivated to drive this problem to completion, so if someone else wanted to "co-sponsor" this RFC and be really active about it, I would be stoked about it.

cc @nikomatsakis @aturon @wycats @sgrif all of whom have talked about issues in this area.

DanielKeep:
A concern I have with this are incidental dependencies.

For example, a crate I'm working on at the moment can optionally use `regex` for some parsing.  However, using `regex` *also* means using `lazy_static`.  Thus, I couldn't use this: Cargo doesn't allow one dependency to trigger the inclusion of another dependency.

A more general design might be to allow features to have an optional `activate-if-available` that lists dependencies that, if they're all already being used somewhere else in the dependency graph, cause the feature to be activated by default.

... of course, this same crate *also* wouldn't be able to use that because it'd be too easy to end up with multiple parsing implementations enabled.  At least, not without having to explicitly control activation order everywhere in the code... I really wish Cargo would let us define mutually exclusive features...

I suppose my position is: "I'd like this feature, but I don't think this specific design would work in practice."

withoutboats:
@DanielKeep This doesn't seem different from having multiple dependencies on the cfg: `#[cfg(depedency=regex, lazy_static)]`.

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+as the conditional dependency. This module will be tagged with the cfg, and the
+extern crate declaration and all code that uses symbols from that crate will
+go inside of that module.
+
+For example:
+
+```rust
+/// lib.rs
+
+#[cfg(dependency=foobar]
```
chris-morgan: `#[cfg(dependency = "foobar")]`

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+If you want to add a new conditional dependency, you should add a top level
+submodule (that is, mounted directly under lib, main, etc) with the same name
+as the conditional dependency. This module will be tagged with the cfg, and the
+extern crate declaration and all code that uses symbols from that crate will
+go inside of that module.
+
+For example:
+
+```rust
+/// lib.rs
```
chris-morgan: `///` is definitely not what you wanted.

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+
+#[cfg(dependency=foobar]
+mod foobar;
+```
+
+```rust
+/// foobar.rs
+
+extern crate foobar;
```
chris-morgan: Note that this is a departure from the *very* strong convention of only using `extern crate` in the crate root. It also has substantial potential to be confusing, because now for to import something from that crate you’d need `use self::foobar::Bar;` or `use foobar::foobar::Bar;` rather than `use foobar::Bar;`.

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+When packaging and uploading a crate to crates.io, some strategy will be used
+to ensure that its conditional dependencies could not introduce a cycle into
+the dependency graph of your crate.
+
+## `cargo doc` compiles with all dependencies
+
+Unlike other forms of compilation, the `cargo doc` command will treat
+conditional dependencies as present by default, in order to document the APIs
+which exist only when conditional dependencies are present. The conditional
+dependency may be provided somehow.
```
chris-morgan: > The conditional dependency may be provided somehow.

I don’t know what you mean. The sentence doesn’t make sense. Do you mean *documented*, i.e. some kind of banner, like with the stability attributes?

Nemo157:
I like it, especially the part about `cargo doc` documenting all the optional dependencies. I've just written a crate with a few features and really wishing that `cargo doc` would just auto-generate something letting users know what different features activate.  In fact I'm just about to add a `serde` feature to that crate for conditional deserialization support so if you get this implemented in the next couple of days I'll use it instead 😜

---

> if a matching source for a conditional dependency (e.g. a matching version number) is already present in the graph

I think this could warrant some extra warning information added to cargo, e.g. assuming you have a dependency `foo` with a conditional implementation for `serde 0.8.0` which you are currently using everything will work fine, if at some point in the future you update to `serde 0.9.0` in the root crate it will no longer match and you'll start getting build errors like `foo::Bar does not implement serde::Deserialize`.

If cargo detects this it would be nice to have it suggest that the reason is the non-matching dependency versions. In this case it would probably be easy to detect why it happened, but if it's related to interaction between dependencies of dependencies it might be harder. I think this can only happen with breaking updates (assuming all crates are following semver around publicly exposed dependencies properly), but there might be some case I haven't thought of that would allow a non-breaking dependency update to break this.

---

Also, what about conditional implementations for multiple versions of a dependency? It might be nice to have something like

```rust
#[cfg(dependency = "foobar:1.0.0")]
mod foobar {
    extern crate foobar;
    impl foobar::Bar for MyType {
        ...
    }
}
#[cfg(dependency = "foobar:2.0.0")]
mod foobar {
    extern crate foobar;
    impl foobar::SuperBar for MyType {
        ...
    }
}
```

although I have no idea how you would specify that in the `Cargo.toml` 😕

chris-morgan:
@withoutboats `#[cfg(dependency=regex, lazy_static)]` would be `#[cfg(all(dependency = "regex", dependency = "lazy_static"))]`

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+matching version number) is already present in the graph, cargo will
+automatically add that dependency.
+
+## `#[cfg(dependency=)]`
+
+A new kind of cfg attribute will be added to rustc. The `dependency` attribute
+will only be compiled when a certain dependency has been explicitly passed to
+rustc using the --extern flag.
+
+Because cargo automatically passes dependencies explicitly with this command,
```
Manishearth: I _think_ `cargo build -p foo` may need to be tweaked to support this case. Minor issue.

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+[drawbacks]: #drawbacks
+
+This adds more functionality to both cargo and rustc. The distinction between
+"conditional" dependencies and "optional" dependencies is sort of subtle and
+will have to be explained to users. More complexity always has a downside.
+
+# Alternatives
+[alternatives]: #alternatives
+
+We considered options like allowing the orphan rules to be broken by certain
```
Manishearth: FWIW the application of sibling crates is much more than just optional dependencies.

One example is Servo. Servo is nobody's dependency, and has a bunch of crates, split up for compile times. Between these crates it would be lovely if we could bypass the orphan rules -- the main reason for the orphan rules is so that downstream users shouldn't get conflict issues that they're powerless to fix (without editing their dependencies directly), but in this case we have no downstream users -- the downstream crates are all Servo crates, and if a change in `layout` forces me to make a change in `style` to avoid an ambiguity, so be it. All the Servo crates should really be considered as a single whole wrt orphan rules.

I recently was forced to merge three crates to get around orphan rule issues (and other things).

0000-conditional_dependencies.md
withoutboats: it is supposed to be documented, yes, will update

0000-conditional_dependencies.md
withoutboats: Yes, this isn't going to solve this problem. The solution to this problem is incremental compilation and then don't break up your projects into crates just to get coarse grained incremental compilation.

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+A single dependency objecy cannot contain both a `conditional` key and an
+`optional` key.
+
+When generating the dependency graph, cargo will not include dependencies
+tagged conditional. It will then traverse the graph looking for conditional
+dependencies; if a matching source for a conditional dependency (e.g. a
+matching version number) is already present in the graph, cargo will
+automatically add that dependency.
+
+## `#[cfg(dependency=)]`
```
withoutboats: Thought a bit about this and there are two syntaxes which are equivalent that we can decide between, will write up more in a comment but it probably won't be today.

0000-conditional_dependencies.md
Manishearth: I think that's a bit extreme though. We still want there to be crates for code organization and code reuse (for example now Firefox uses our style crate), but it would be nice if orphan rules could be broken amongst these crates.

(Yes, in this case that means we do have downstream users, but that's C++ code)

Ericson2314:
So I'd *really* like to enforce that all uses of these conditional dependencies are propoerly guarded by a `cfg` attribute. As I argued in the internals thread for scenarios, features + a similar cfg-check probably cover scenarios, so now we have two uses for such a thing.

0000-conditional_dependencies.md
```diff
@@ -0,0 +1,142 @@
...
+
+Or:
+
+```toml
+[dependencies.foobar]
+version = "1.0.0"
+conditional = true
+```
+
+A single dependency objecy cannot contain both a `conditional` key and an
```
sgrif: s/objecy/object

sgrif:
This is a great proposal, and I like the overall idea. This can potentially let us sidestep a lot of issues with Cargo's version resolution if done properly. Having rustdoc include conditional dependencies by default is a nice touch as well. I strongly disagree with the notion that this solves the problems created by the orphan rule, as this still requires a crate to care about every other crate in the ecosystem, or for every other crate in the ecosystem to care about that crate. However, this does greatly improve ergonomics for the end user without having to address the orphan rule directly.

withoutboats:
> So I'd really like to enforce that all uses of these conditional dependencies are propoerly guarded by a cfg attribute.

What we need are two compilation modes - one with the crate under compilation's conditional dependencies turned on, one with them turned off.

withoutboats:
> I strongly disagree with the notion that this solves the problems created by the orphan rule

Definitely didn't mean to imply that this solves all of the negative consequences of the orphan rules.

withoutboats:
So I think the actual issue that isn't solved by this scheme is not a situation where want to include some code if both (e.g.) `regex` and `lazy_static` are in the graph, but where you want to include some code if _regex_ is in the graph, but that requires adding lazy_static.

How often is this desirable? If its something you usually need, it seems like we can do it, but it introduces other questions - if my dependency adds `lazy_static` while resolving a conditional dependency for `regex`, and another dependency has a conditional dependency on `lazy_static`, do we do another round of conditional dependency checking? If that second round adds some more new dependencies, do we continue doing this indefinitely? So on.

The syntax for doing that which is consistent with the syntax in this proposal seems to be accepting a string for the `conditional` member of a dependency, which indicates that it is conditional on _another_ crate being in scope. As in:

```toml
[dependencies.regex]
version = "1.0.0"
conditional = true

[dependencies.lazy_static]
version = "1.0.0"
condiitonal = "regex"
```

0000-conditional_dependencies.md
Manishearth: > I'm open to any way to resolve the papercuts of the orphan rules that doesn't break that rule. 

Yeah, I think we can design this to work without cargo. But this is pretty off topic, just wanted to point out (like sgrif) that the orphan rules have a few more problems that this can't help with.

0000-conditional_dependencies.md
Ericson2314: The `magic` route is what Cabal uses for its flags, basically.

Somewhat relatedly, I think `target.cfg(....).foo` syntax is really good---way better than optional syntax. I would love:
```toml
[targets.'cfg(and(or(feature = "foo", feature = "bar"), create = "baz"))'.asdf]
#...
```

0000-conditional_dependencies.md
withoutboats: > Yeah, I think we can design this to work without cargo.

Cool. :-) I'm always very excited to hear new ideas to make the orphan rules less of a burden. I'd love to continue this conversation off of this RFC thread.

Ericson2314:
@withoutboats 

> What we need are two compilation modes - one with the crate under compilation's conditional dependencies turned on, one with them turned off.

Actually that wouldn't cut it. For example, it erroneously allows:
```rust
#[cfg(dependency=foo)]
extern bar;
#[cfg(dependency=bar)]
extern foo;
```

My originally rambling post is https://internals.rust-lang.org/t/fleshing-out-libstd-scenarios/4206/9?u=ericson2314, the tl;dr is we need to do SAT (and when we take into account upstream restrictions, perhaps easier to thank of as as SMT).

Manishearth:
> The syntax for doing that which is consistent with the syntax in this proposal seems to be accepting a string for the conditional member of a dependency, which indicates that it is conditional on another crate being in scope.

Could we make this still tie-in with features (so that there is no distinction between `cfg(feature)` and `cfg(conditional)`)? In your specific syntax I think just requiring people to define a `regex` feature (the way `optional=true` works right now) would be enough. However, I'd prefer marking the _feature_ as magic/conditional rather than the dependencies.

Your new proposal still misses one use case, where you have two partially overlapping sets of conditional dependencies. Sounds pretty rare, but that and more complicated overlappings are cleanly handled by the magic feature proposal.

Regarding having to re-resolve indefinitely, it's not indefinite -- there is a finite number of dependencies and the conditional resolution algorithm is monotonic.

withoutboats:
@Manishearth I imagine making this a 'property of features' would look something like this:

```toml
[features.serde]
dependencies = [serde]
automatic = true
```

And then the feature would be opted into automatically if the dependencies listed are already present. You could also imagine that `automatic` (or whatever) can take a list of dependencies separate from the dependencies listed as members of this feature, and the feature is opted into if the dependencies in `automatic` are listed.

These syntaxes seem equally capable to me. I'll list this as an alternative when I revise this RFC. I personally prefer the syntax in the RFC right now because it creates a distance between this and features. I don't see this as a system for making features easier to opt into, but as a system for ergonomically solving a problem that is being solved unergonomically with features right now. Maybe my view of it is limited though. I will admit I have a bias against feature systems in general, including ours. Because of their generality, they are always sort of frustrating to use in my experience.

> Your new proposal still misses one use case, where you have two partially overlapping sets of conditional dependencies.

I do not know what you mean by this.

Manishearth:
> These syntaxes seem equally capable to me. 

More or less yeah. The main difference is that this creates a second config system independent of features, whereas I'd prefer to work within the same framework. I really want to avoid a system where you have both `cfg(feature=)` and `cfg(conditional=)`. If using features, you get interoperability with features, easy upgrades without breaking changes (folks can make their features automatic, and users won't be forced to change their cargo files to upgrade, nor do maintainers have to keep both a feature and a conditional alive in parallel till the next breaking change), and the same ability to pick and choose features in deps that we already have (instead of including conditional deps that match the ). This especially matters for conditional deps that need more than one crate to work (serde+serde_derive, for example). As a library I may not be using serde_derive, but I would like to trigger the serde conditional in a dependency of mine. I can't without including serde_derive (which means I have to go back to the dependency's cargo.toml and pick up all the conditional deps). However, if it has a "serialize" feature I can easily specify `features = "serialize"` for the dependency and it will work. Furthermore, the dependency can add more deps to the feature and I don't have to change anything.

> I do not know what you mean by this.

I was wrong. Thought about it more.

withoutboats:
> If using features, you get ... easy upgrades without breaking changes

This is pretty compelling!

Ixrec:
Another reason I like the "automatic features" design is that I can imagine having features which some clients want to explicitly enable without caring exactly what set of (conditional) dependencies you use to implement those features.

If I use serde for everything, I'd like all my dependencies to automagically compile their serde impls. But if there's only one or two libraries I want to serialize types from, and I don't care whether Foo.serialize() uses serde or rustc-serialize or whatever, I'd prefer to turn on "the serialize feature" rather than pretend I depend directly on serde.

nikomatsakis:
FWIW, I prefer going through features for this.

withoutboats:
Based on feedback for the past three weeks I've intended to rewrite this from "Conditional dependencies" to "automatic features," but I have trouble finding the time. If anyone wants to rewrite the proposal to work as @Manishearth suggested, I would happily accept the PR into this branch.

Manishearth:
I will ... try, but I'm not sure if I have time. (So if someone else can pick it up that would be great!)

Manishearth:
Rewrote proposal at https://github.com/withoutboats/rfcs/pull/1

withoutboats:
RFC has been updated with @Manishearth's edits. There's still more things to disagree about on it. :-)

Manishearth:
I completely agree with the new version. Obviously.

0000-automatic_features.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+## Automatic features in `Cargo.toml`
+
+When declaring a feature in a Cargo.toml file, if that feature is an
+object, it can contain the `automatic` member. For example:
+
+```toml
```
sfackler: All dependencies define their own features, so I'm assuming `automatic` can be applied dependency declarations as well?

Manishearth: We could let that be the case. I'd prefer if it went by the route where you must create a feature of the same name and use that, but no strong reason (mostly just to avoid it getting confusing) for this.

sfackler: I would prefer that approach as well, but that seems like a whole other RFC!

sfackler:
```toml
[package]
name = "foo"
version = "0.1.0"
authors = ["Steven Fackler <sfackler@palantir.com>"]

[features]
libc = []

[dependencies]
libc = "0.2"
```

```
error: failed to parse manifest at `/Users/sfackler/foo/Cargo.toml`

Caused by:
  Features and dependencies cannot have the same name: `libc`
```

Manishearth: Oh, huh. I could swear that was the case. We could add wording to allow this here, or just go the way you proposed. @withoutboats thoughts?

withoutboats: The fact that (right now) you need to have a different name for the feature from the dependency is a big part of why I was in favor of the original proposal in the first place. Saying "automatic" could be applied to features is just returning to that proposal, but changing the key from "conditional" to "automatic" (which doesn't make sense to me).

Manishearth: @withoutboats Well, the difference is that the new proposal is still applying to _features_ (and thus not requiring any changes to rustc), and still integrates with the rest of the feature system (and still works for multi-dependency features). I would prefer to push for allowing dependency features to be shadowed, which fixes this problem for us.

@sfackler yeah, I was aware, I just thought that you could shadow that.

withoutboats: Thanks, I knew that dependencies were features but the implications of that hadn't been obvious to me before.

So it seems like the normal way to do this is going to be something like this:

```toml
[dependencies.libc]
version = "0.2"
optional = true
automatic = true
```

I don't _love_ having to attach two booleans to the dependency object, but I'm not sure some sort of third key that is the same as `optional && automatic` is a good idea.

Manishearth: FWIW I was talking with @wycats and it seems like the lack of feature shadowing is mostly an oversight. We should just include that in this rfc.

sfackler:
It would be fantastic if this interacted with #1133 to allow `#![no_std]` crates to automatically enable std functionality when its being pulled in.

alexcrichton:
Reading over this the RFC looks pretty good to me, along with the underlying idea. 

I have a number of questions here, but lots of them boil down to clarifying this statement.

> It will then look for features (in all crates in the graph) that only need dependencies which are already part of the graph.

More formally, any feature in a manifest can be expanded to a list of dependencies in the `[dependencies]` section plus a list of other features. Both of these lists could be empty. If the feature were enabled, then these two lists would also be enabled. My assumption is that this RFC is basically saying that if all of the dependencies are enabled, then the feature itself is enabled (as everything it wants is already there). 

If that's true, which I'm not 100% sure if it is, then I think it would be good to clarify at least that dependency matching is still performed on dependencies. For example if I have an automatic, optional dependency on serde 0.7 then if serde 0.8 is in the dependency graph then my serde dep shouldn't be activated.

Additionally, if the set of feature-dependencies of a feature is not empty, I don't quite understand the semantics of how that's automatically enabled. For example, if we have:

```toml
[dependencies]
serde = { version = "0.8", optional = true }

[features]
serialize = { dependencies = ["serde", "other"], automatic = true }
other = []
```

Is it possible for the `serialize` feature to ever get automatically enabled? It'd require at least serde 0.8 to be enabled in the dependency graph but would it *also* require the `other` feature to be activated by some other dependency? The wording in this RFC makes it sound like `serialize` is activated if serde 0.8 is included (regardless of `other`), but this seems slightly odd to me.

---

Along those lines, I think a more rigorous specification would then also assist in clarifying edge cases like:

```toml
[features]
foo = { automatic = true }
```

Is that an error? Always activated? Never activated?

Similarly it'd help clarify:

```toml
[features]
foo = { automatic = true, dependencies = ["bar"] }
bar = []
```

Does enabling `bar` enable `foo` through automatic features? Is this an error? (etc)

Manishearth:
> My assumption is that this RFC is basically saying that if all of the dependencies are enabled, then the feature itself is enabled (as everything it wants is already there).

correct

> Is it possible for the serialize feature to ever get automatically enabled? 

Not unless the `other` crate is pulled in by someone else.

> The wording in this RFC makes it sound like serialize is activated if serde 0.8 is included (regardless of other), but this seems slightly odd to me.

Which part of the wording makes this feel that way?

> Is that an error?

Error I'd say.

> Does enabling bar enable foo through automatic features? 

Yes.

@withoutboats do you have time to make these clarifications?

nikomatsakis:
@alexcrichton I don't know much (anything, really) about the dependency resolution algorithm that cargo uses. How hard do you think it would be to integrate this into that code?

nikomatsakis:
Oh, and my assumption was that if you had an automatic feature tied to `serde-0.7`, then indeed serde-0.8 would not trigger it (but presumably one could have other semver operators too).

alexcrichton:
@nikomatsakis I think it sort of depends on the precise meaning of the RFC, but my current belief is that given my own personal interpretation of this RFC it should be very easy to add to Cargo. It would look something like:

* For all automatic features, it depends on a list of dependencies, a list of local features, and lists of features in each dependency.
* Dependency resolution is performed as-is today, afterwards considering automatic features
* Automatic features are then activated if they meet the following criteria:
  * All dependencies of an automatic feature are satisfied with an existing node in the dependency graph
  * All feature dependencies of an automatic feature are already activated for the crate in question
  * All features enabled in dependencies are also already activated

If that's all true then the automatic feature is activated, new dependency edges are drawn, and the set of activated features for a crate is updated to include the automatic feature. I think this would need to be an iterative process run until the graph is stable to handle automatic features depending on automatic features, but it shouldn't take more than O(n) time where n is the number of nodes.

Manishearth:
Yeah, your interpretation is correct. With serde-0.7/0.8 an automatic feature won't be activated unless the crates that it depends on up to the precise version ranges are already in there.

Ericson2314:
Yes, the key rule is new edges, not new nodes. Since nothing currently can depend on a feature being *dis*abled, we have a single unique solution that all possible traces of the iterative process will arrive at. Note that if we did have mutually-exclusive features or similar, then there may not be a unique solution.

Ericson2314:
Hmm, I think I found another issue.

An important part of package management is adjusting one's dependencies should never in itself be a breaking change of one's own interface. But if a crate relies on some transitive dep triggering an automatic feature of an immediate dep, and that transitive dep disappears, then our crate will break because its intermediate dep has to rely on less.

There is two compatible solutions here, i.e. any crate in the ecosystem can adopt either, Cargo doesn't need to pick up front:

1. Crates don't, in fact, rely on the optional features. Instead, they condition any code that does on the dependency's feature (or equivalently declare their own optional feature wrapping the dependency's feature and cfg on that).

2. Optional dependencies are only *exposed* considering "sibling" dependencies: Our crate only gets access to its dependency's optional features if its own other dependencies [hence they both have the same parent, our crate, and are siblings]. If "half-siblings" from another crate would trigger the optional feature, then our crate should not be privy to this (in the bright scenario's future where non-requested things in general aren't leaked). Furthemore, if only non-siblings would cause the optional feature to be enabled under the old rules, then no one gets to use it, and there's not point compiling the dependency with the optional dependency enabled at all.
  
   If we wish to require more opt-in from the user (though this doesn't affect soundness), instead of looking at all sibling deps, crates could specify which sibblings should be aware of each others existance, e.g. `{ spin = ..., lazy_static = { knows about = ["spin"], ... } }`.

Besides soundness, I believe that this also has the benefit of cheapening optional dependency resolution. I think it can now be performed in parallel for every interior node, and the results aggregated at the end.

Ericson2314:
**Meta comment**: So far, I haven't seen very much effort put into formalizing Cargo and considering whether it upholds certain properties we care about, or even what those properties are. I get that, in the absence of enforcing users to package their code properly (or be able to automatically blacklist crates that aren't packaged properly), the soundness is Cargo won't be practically put to the test. But I'd hope that this RFC and https://github.com/rust-lang/rfcs/pull/1841 indicate that getting Cargo sound is very subtle---potentially as subtle as getting Rust itself sound. At the very least, I'd hope we'd be able to exhaustively document everything users need to worry about to type-check their crates' `Cargo.toml`s and `!#[cfg(..)]`s by hand.

sfackler:
The point about transitive dependencies is a good one. Scenarios would help here since we could prevent you from using those features unless you had the direct dependency even if another crate already was using it.

Ericson2314:
@hvr, a Cabal developer, just told me that Cabal's "automatic flags" (basically this) are not intended to affect a package's interface. This means it *is* sound to use the global algorithm, which can discover optional edges the local one won't.

I think both are useful---for example I suppose if optional features don't affect the public API, they probably improve performance, in which case the more the merrier! [At least in release builds, more edges do reduce build parallelism.] The global algorithm's disadvantage---enabling more non-trivial edges---becomes and advantage in this case.

Ericson2314:
@sfackler Well, the other thing I've been thinking is "scenarios" need not be a whole new feature of Cargo, but rather a reformation of features/`cfg`. [Apologies for using the word "feature" in two different senses.]

Manishearth:
> Crates don't, in fact, rely on the optional features. 

Except for toplevel crates.

But this doesn't solve the original problem then. Let's say I have a crate which needs serialization to work. It wants to enable serde in all of its transitive dependencies. I want to be able to do this easily, so it just specifies a serde dep in the top level and the automatic features pull it in in its dependencies. If there is a version mismatch that's not a different situation than what we'd be in previously, and you have that kind of problem with non-optional dependencies _anyway_.

Not relying on optional features brings us back to the status quo.

> Optional dependencies are only _exposed_


How would you gate that? Remember that an optional dependency is more than just an extra rlib; there will be cfgs in the parent crate that compile differently. You cannot control access to things affected by the cfgs.


------------------

Most of this is nicely solved by the fact that if you're relying on an optional dependency like serde you'll need to pull the crate in anyway to use it. It's not perfect, but basically we can declare that if a crate wish to forcibly trigger an optional dependency of its dependenc(ies) to be pulled in, you must specify its constituent crates/features in the crate that needs the dependency to be triggered. This can be a rule that is not enforced; I'm not sure if it _can_ be enforced.

Ericson2314:
> Except for toplevel crates.

I'm not sure what's special about them here? Moreover making the top ones of a DAG special is IMO a smell. Inductively, the leaf nodes are the base case and the roots are just more interior nodes.

> Let's say I have a crate which needs serialization to work. It wants to enable serde in all of its transitive dependencies....

Good point. Heh, oh the joys of trying to find something both sound and useful! One possible solution (possibly sound, in that I just thought of it :)) is feeding in crates to all their siblings' descendants. In your example, since serde is a dependency of the root crate, everything gets built with serde-deps enabled, but only the root crate gets to see those optional features enabled by serde.

Maybe that's good enough, or maybe the intermediate crates needs to know their deps also use serde so they can write the relevant instances. In that case, they would need their own optional dependencies anyways for their serde usage. I forgot about routing *optional* sibling crates in my previous post, but that would allow the serde optional dep to be propagated without the change from the previous paragraph. [This removes the embarrassingly parallel aspect, but keeps the algorithm "weakly inductive" (I think that's the term) where you only consider a node and its immediate children rather than all descendants.]

Say there is a "gap" in the dag, with some crates using serde not reachable via the root crate following edges between serde-caring crates (which is using serde). Does it matter? I'm not sure. If it does, then we need *both* ideas---propagating optional features and propagating siblings to siblings' children.

> If there is a version mismatch that's not a different situation than what we'd be in previously, and you have that kind of problem with non-optional dependencies anyway.

Ah! Let me add a detail I forgot in my previous paragraph (and which I am not adding there retroactively for the sake of information density). Each serde-aware crate should have have an optional automatic feature on serde, and that optional dependency should *require* (not merely allow to be enabled) the optional dependency of its immediate dependencies. Then, inductively, all the crates will be required to use serde together, and we, erm, somewhat side step your concern about whether versions should be resolved entirely before automatic features.

> How would you gate that? Remember that an optional dependency is more than just an extra rlib; there will be cfgs in the parent crate that compile differently. You cannot control access to things affected by the cfgs.

This is the hard crux of what I think is already proposed as part of scenarios. We need to mask the interface of crates to match these hypothethetical other resolutions. The "Additivity" rule from https://github.com/rust-lang/rfcs/pull/1841 is essential to make sure this is always upcasting and thus possible.

Manishearth:
> Each serde-aware crate should have have an optional automatic feature on serde, and that optional dependency should require (not merely allow to be enabled) the optional dependency of its immediate dependencies.


No. That's my entire point. _This is the status quo_, this is _exactly_ what you do when you want to use features. This has always been problematic, because you need to thread features through a million different Cargo.toml files. This is _exactly_ what this RFC is trying to avoid.


> One possible solution (possibly sound, in that I just thought of it :)) is feeding in crates to all their siblings' descendants.

Breaks down in the presence of generic types being pulled in from a different part of the dag.

Ericson2314:
> No. That's my entire point...

Well then at least we need to require that no two automatic features are triggered by different version of the same crate. Still thinking about the rest.

Manishearth:
That restriction feels strange, but I don't see any actual issues with it.

Ericson2314:
Yeah its weird to me too. Very global, very not inductive

> If there is a version mismatch that's not a different situation than what we'd be in previously, and you have that kind of problem with non-optional dependencies anyway.

Oh, do you mean the sketchy fact that we already allow multiple versions of the same crate in build plans?

Manishearth:
> Oh, do you mean the sketchy fact that we already allow multiple versions of the same crate in build plans?

Yeah.

Ericson2314:
> Breaks down in the presence of generic types being pulled in from a different part of the dag.

You mean generics + traits? I'd think generics alone would be fine.

aturon:
Based on discussion at the most recent Cargo team meeting, I'm going to close this as **postponed**.

While this still seems like a very promising idea, there are some more pressing revisions for the feature system (and in general bigger fish to fry with Cargo right now). Let's plan to revisit in 2018!

aturon:
Sorry, that was a bit terse -- I'm trying to push on a bunch of RFCs and went too quickly. In more detail, Cargo folks are definitely fans of this idea. But:

- There are some significant issues with the feature system right now, which we intend to revise as part of the build system integration work. We're generally holding off on *extensions* to the feature system until we've successfully rationalized what's there.

- Our hands are going to be extremely full with some of the tasks on deck (the build system work alone could easily take the rest of the year).

Put together, it seemed wisest to delay finalizing the design here until the dust settled in other parts of Cargo.

vi:
Has the dust already settled?

"Include tokio/async-specific features" seem to be a good example of an automatic feature. Otherwise either tokio/futures would suddenly appear unless `default-features = false`; or all async users need to write the curly brackets.

kentfredric:
Having vendored perl packages which have been doing this since forever, I can only say that in practice its a terrible idea.

The last thing you want when trying to build a reliable system is having to think about the possibility you will incur a *functional change* as an incidental side effect of something simply having been present, where it previously was not.

Clarity in dependencies and explicitly requiring things when you need them is worth the extra tiny amount of developmental pain.

I'd hate to ever update one of my dependencies, which in turn, pulls in a new dependency, and have it turn on a feature in a *different* dependency which changes how my code behaves.

That's right in "spooky action at a distance" territory, and can be a nightmare to work out where the breaking change was introduced.

(Isn't that part of why we like `use` as opposed to a great big global share of everything like C has? ... 'automatically visible functions'. )

PR #1789, opened 2016-11-13
Conversions from `&mut T` to `&Cell<T>`

Kimundi:
A proposal about legitimizing the cast of a `&mut T` to a `&Cell<T>`.

[Rendered](https://github.com/Kimundi/rfcs/blob/as-cell/text/0000-as-cell.md)

### DIFF
--- a/text/0000-as-cell.md
+++ b/text/0000-as-cell.md
@@ -118,7 +118,7 @@ for (i, j) in v_slice[n..].iter().zip(v_slice) {
 
 ## Language
 
-The core of this proposal is the ability to cast a `&T` to a `&Cell<T>`,
+The core of this proposal is the ability to convert a `&T` to a `&Cell<T>`,
 so in order for it to be safe, __it needs to be guaranteed that
 `T` and `Cell<T>` have the same memory layout__, and that there are no codegen
 issues based on viewing a reference to a type that does not contain a
@@ -175,7 +175,7 @@ effect on any existing code.
 
 ### Cell Slicing
 
-We enable a cast from `&Cell<[T]>` to `&[Cell<T>]`. This seems like it violates
+We enable a conversion from `&Cell<[T]>` to `&[Cell<T>]`. This seems like it violates
 the "no interior references" API of `Cell` at first glance, but is actually
 safe:
 
@@ -189,10 +189,10 @@ safe:
   through a `&Cell<[T]>`, or just a memcopy to a single element through a
   `&Cell<T>`.
 - Yet another way to look at it is that if we created a `&mut T` to each element
-  of a `&mut [T]`, and casted each of them to a `&Cell<T>`, their addresses
+  of a `&mut [T]`, and converted each of them to a `&Cell<T>`, their addresses
   would allow "stitching" them back together to a single `&[Cell<T>]`
 
-For convenience, we expose this cast by implementing `Index` for `Cell<[T]>`:
+For convenience, we expose this conversion by implementing `Index` for `Cell<[T]>`:
 
 ```rust
 impl<T> Index<RangeFull> for Cell<[T]> {
@@ -286,8 +286,8 @@ you have mutable access to it.
 > Would the acceptance of this proposal change how Rust is taught to new users at any level?
 How should this feature be introduced and taught to existing Rust users?
 
-As it is, the API just provides a few neat casting functions. Nevertheless,
-with the legalization of the `&mut T -> &Cell<T>` cast there is the
+As it is, the API just provides a few neat conversion functions. Nevertheless,
+with the legalization of the `&mut T -> &Cell<T>` conversion there is the
 potential for a major change in how accessors to data structures are provided:
 
 In todays Rust, there are generally three different ways:
@@ -317,8 +317,8 @@ of ownership and change it to a "rule of four".
 
 > What additions or changes to the Rust Reference, _The Rust Programming Language_, and/or _Rust by Example_ does it entail?
 
-- The reference should explain that the `&mut T -> &Cell<T>` cast,
-  or specifically the `&mut T -> &UnsafeCell<T>` cast is fine.
+- The reference should explain that the `&mut T -> &Cell<T>` conversion,
+  or specifically the `&mut T -> &UnsafeCell<T>` conversion is fine.
 - The book could use the API introduced here if it talks about internal mutability,
   and use it as a "temporary opt-in" example.
 - Rust by Example could have a few basic examples of situations where this API
@@ -378,13 +378,13 @@ conversions steps, while still offering essentially the same API.
 
 ## Just the language guarantee
 
-The cast could be guaranteed as correct, but not be provided by std
+The conversion could be guaranteed as correct, but not be provided by std
 itself. This would serve as legitimization of external implementations like
 [alias](https://crates.io/crates/alias).
 
 ## No guarantees
 
-If the casting guarantees can not be granted,
+If the safety guarantees of the conversion can not be granted,
 code would have to use direct indexing as today, with either possible
 bound checking costs or the use of unsafe code to avoid them.
 
@@ -416,10 +416,10 @@ Cons:
 
 ## Cast to `&mut Cell<T>` instead of `&Cell<T>`
 
-Nothing that makes the `&mut T -> &Cell<T>` cast safe would prevent
+Nothing that makes the `&mut T -> &Cell<T>` conversion safe would prevent
 `&mut T -> &mut Cell<T>` from being safe either, and the latter can be
 trivially turned into a `&Cell<T>` while also allowing mutable access - eg to
-call `Cell::as_mut()` to cast back again.
+call `Cell::as_mut()` to conversion back again.
 
 Similar to that, there could also be a way to turn a `&mut [Cell<T>]` back
 into a `&mut [T]`.
--- a/text/0000-as-cell.md
+++ b/text/0000-as-cell.md
@@ -131,7 +131,9 @@ However, if changes as proposed in https://github.com/rust-lang/rfcs/pull/1651 s
 get implemented, the `Copy` bound might get relaxed or removed entirely,
 which would affect the ergonomics here.
 
-The proposed implementation only covers the base case `&mut T -> &Cell<T>`
+It might also be possible to add `AsRef` implementations for this conversion.
+
+The proposal only covers the base case `&mut T -> &Cell<T>`
 and the trivially implementable extension to `[T]`,
 but in theory this conversion could be enabled for
 many "higher level mutable reference" types, like for example
--- a/text/0000-as-cell.md
+++ b/text/0000-as-cell.md
@@ -335,18 +335,6 @@ of ownership and change it to a "rule of four".
 # Alternatives
 [alternatives]: #alternatives
 
-## Just the language guarantee
-
-The cast could be guaranteed as correct, but not be provided by std
-itself. This would serve as legitimization of external implementations like
-[alias](https://crates.io/crates/alias).
-
-## No guarantees
-
-If the casting guarantees can not be granted,
-code would have to use direct indexing as today, with either possible
-bound checking costs or the use of unsafe code to avoid them.
-
 ## Removing cell slicing
 
 Instead of allowing unsized types in `Cell` and adding the `Index` impls,
@@ -387,6 +375,31 @@ for (i, j) in v_slice[n..].iter().zip(v_slice) {
 
 This would be less modular than the `&mut [T] -> &Cell<[T]> -> &[Cell<T>]`
 conversions steps, while still offering essentially the same API.
+
+## Cast to `&mut Cell<T>` instead of `&Cell<T>`
+
+Nothing that makes the `&mut T -> &Cell<T>` cast safe would prevent
+`&mut T -> &mut Cell<T>` from being safe either, and the latter can be
+trivially turned into a `&Cell<T>` while also allowing mutable access - eg to
+call `Cell::as_mut()` to cast back again.
+
+Similar to that, there could also be a way to turn a `&mut [Cell<T>]` back
+into a `&mut [T]`.
+
+However, this does not seem to be actually useful since the only reason to use
+this API is to make use of shared internal mutability.
+
+## Just the language guarantee
+
+The cast could be guaranteed as correct, but not be provided by std
+itself. This would serve as legitimization of external implementations like
+[alias](https://crates.io/crates/alias).
+
+## No guarantees
+
+If the casting guarantees can not be granted,
+code would have to use direct indexing as today, with either possible
+bound checking costs or the use of unsafe code to avoid them.
 
 ## Exposing the functions differently
 
--- a/text/0000-as-cell.md
+++ b/text/0000-as-cell.md
@@ -6,11 +6,11 @@
 # Summary
 [summary]: #summary
 
-For all `T`, define `Cell<T>` as having the same memory layout as `T` and
-add conversion functions with the following signatures to std:
-
-- `&mut T -> &Cell<T>`
-- `&mut [T] -> &[Cell<T>]`
+- Change `Cell<T>` to allow `T: ?Sized`.
+- Guarantee that `T` and `Cell<T>` have the same memory layout.
+- Enable the following conversions through the std lib:
+  - `&mut T -> &Cell<T> where T: ?Sized`
+  - `&Cell<[T]> -> &[Cell<T>]`
 
 > Note: https://github.com/rust-lang/rfcs/pull/1651 has been accepted recently,
 > so no `T: Copy` bound is needed anymore.
@@ -21,7 +21,7 @@ add conversion functions with the following signatures to std:
 Rust's iterators offer a safe, fast way to iterate over collections while avoiding
 additional bound checks.
 
-However, due to the borrow checker, they run into issues if you try to have
+However, due to the borrow checker, they run into issues if we try to have
 more than one iterator into the same data structure while mutating elements in it.
 
 Wanting to do this is not that unusual for many low level algorithms
@@ -53,7 +53,7 @@ for i in n..v.len() {
 
 ```
 
-...but this reintroduces the bound-checking costs.
+...but this reintroduces potential bound-checking costs.
 
 The alternative, short of changing the actual algorithms involved, is to use
 internal mutability to enable safe mutations even with overlapping shared views into the data:
@@ -79,13 +79,12 @@ This has the advantages of allowing both bound-check free iteration and
 aliasing references, but comes with restrictions that makes it not generally
 applicable, namely:
 
-- The need to modify the data structure containing the data (Which might not
-  always be possible because it might come from external code).
+- The need to change the definition of the data structure containing the data
+  (Which is not always possible because it might come from external code).
 - Loss of the ability to directly hand out `&T` and `&mut T` references to the data.
-- Refcounting overhead in case of `RefCell<T>`.
 
 This RFC proposes a way to address these in cases where `Cell<T>`
-would be used by introducing simple conversions functions
+could be used by introducing simple conversions functions
 to the standard library that allow the creation of shared borrowed
 `Cell<T>`s from mutably borrowed `T`s.
 
@@ -119,28 +118,27 @@ for (i, j) in v_slice[n..].iter().zip(v_slice) {
 
 ## Language
 
-In order for this proposal to be safe, __it needs to be guaranteed that
+The core of this proposal is the ability to cast a `&T` to a `&Cell<T>`,
+so in order for it to be safe, __it needs to be guaranteed that
 `T` and `Cell<T>` have the same memory layout__, and that there are no codegen
-issues based on viewing a reference to a `UnsafeCell`-less types as a
-reference to a `UnsafeCell`-containing type.
+issues based on viewing a reference to a type that does not contain a
+`UnsafeCell` as a reference to a type that does contain a `UnsafeCell`.
 
 As far as the author is aware, both should already implicitly
-fall out of the semantic of `Cell` and llvms notion of aliasing:
+fall out of the semantic of `Cell` and Rusts/llvms notion of aliasing:
 
 - `Cell` is safe interior mutability based on memcopying the `T`,
-  and thus does not need additional fields.
+  and thus does not need additional fields or padding.
 - `&mut T -> &U` is a sub borrow, which prevents access to the original `&mut T`
   for its duration, thus no aliasing.
 
-However, we might have to add an explicit language guarantee about
-the validity of this cast.
-
 ## Std library
 
-Add conversions functions for `&mut T -> &Cell<T>` and `&mut [T] -> &[Cell<T>]`
-to std, implemented with the equivalent of a simple `transmute()`.
+### `from_mut`
 
-As an initial design, `Cell` would provide them through additional constructors:
+We add a constructor to the cell API that enables the `&mut T -> &Cell<T>`
+conversion, implemented with the equivalent of a `transmute()` of the two
+pointers:
 
 ```rust
 impl<T> Cell<T> {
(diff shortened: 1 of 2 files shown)


nathan7:
Having this trait makes no real sense — I'd rather see `Cell::from_mut` and `Cell::from_mut_slice` than a trait that covers two semantically incompatible conversions.

Kimundi:
True, the trait only exists to enable method call syntax - maybe thats too extravagant...

nathan7:
Making this usable with method call syntax only serves to further obscure what's going on, and this whole proposal is already designed to do non-obvious (but useful) things by its very nature.

camlorn:
How does this interact with the borrow checker?  It seems to me that the point here is to get rid of the `&mut` so that you can do things with the Vec again, but if you're already iterating over the Vec via `iter_mut`, our current borrow checker isn't smart enough to let you.  At least, in so far as I know.

Can you add an example to the RFC that shows usage of the trait? You've only got one that shows how it's currently a problem and one that shows how I can make it work in current Rust.  I would like to see this so I can conceptualize it, because I really feel like the current borrow checker brings this crashing down.

nathan7:
@camlorn Casting `&mut T` to `&Cell<T>` works today — this RFC merely proposes adding it to the stdlib.

glaebhoerl:
Just for reference, there [is a crate](https://crates.io/crates/alias) (by @huonw) providing this on `crates.io`. (See also https://github.com/rust-lang/rfcs/issues/1106)

For this kind of thing, where the interesting part is not the writing down of the `transmute`s, but the clauses implicit in the language definition which say that they're OK, I do think it makes more sense to provide these directly in `std` than in an external crate. In other words my feeling is that, like `Cell` itself and `Rc` and others, these are essentially language primitives (part of the "trusted code base") that we're choosing to make available as `std` types and functions instead of as built-in syntax, for aesthetic reasons or whatever.

Kimundi:
Alright, I added an example of how it would be used, and switched over the using `Cell` constructors instead.

camlorn:
I see it now. I focused in on the first conversion and missed the second.

What is the use case of the first conversion?  This RFC is all about iteration, but I'm not seeing how `&mut T->Cell<T>` helps.  I think it will in the future, but not until the borrow checker is no longer based on blocks.  Is the future the point?

eddyb:
@camlorn Can you give an example of code you think won't work? If I transmute `&mut T` to `&Cell<T>` I can make as many copies as I want, and the compiler will treat the mutable borrow as expiring when the last `&Cell<T>` goes away (each lexically scoped, of course).

text/0000-as-cell.md
```diff
@@ -0,0 +1,244 @@
...
+the previous Example can be written as this:
+
+```rust
+let mut v: Vec<i32> = ...;
+
+// convert the mutable borrow
+let v_slice: &[Cell<T>] = Cell::from_mut_slice(&mut v);
+
+// example 1
+for i in v_slice.iter() {
```
oli-obk: nit: could just be `for i in v_slice {`

glaebhoerl:
I think the `&mut [T]` -> `&[Cell<T>]` conversion (`from_mut_slice`) could be further decomposed into:

1. `&mut [T]` -> `&Cell<[T]>` (`Cell<[T]>::from_mut`)

2. `&Cell<[T]>` -> `&[Cell<T>]` -- perhaps using an `Index` impl?

So the API could look like this:

```rust
impl<T: ?Sized> Cell<T> {
    fn from_mut(val: &mut T) -> &Cell<T>;
}

impl<T> Index<usize> for Cell<[T]> {
    type Output = Cell<T>;
    fn index(&self, idx: usize) -> &Cell<T>;
}

impl<T> Index<Range<usize>> for Cell<[T]> {
    type Output = [Cell<T>];
    fn index(&self, idx: Range<usize>) -> &[Cell<T>];
}

impl<T> Index<RangeFrom<usize>> for Cell<[T]> {
    type Output = [Cell<T>];
    fn index(&self, idx: RangeFrom<usize>) -> &[Cell<T>];
}

impl<T> Index<RangeTo<usize>> for Cell<[T]> {
    type Output = [Cell<T>];
    fn index(&self, idx: RangeTo<usize>) -> &[Cell<T>];
}

impl<T> Index<RangeFull> for Cell<[T]> {
    type Output = [Cell<T>];
    fn index(&self, idx: RangeFull) -> &[Cell<T>];
}
```

So given a cell of an array (`&Cell<[T]>`), you could index it however you like and get cells or slices of cells (`&[Cell<T>]`).

As an aside, it'd also analogously make sense to support `&Cell<(A, B)>` -> `&(Cell<A>, Cell<B>)`, and likewise for tuples of other sizes and `struct`s. It might be possible to express the former in the general case with https://github.com/rust-lang/rfcs/pull/1582, the latter would almost certainly need compiler support. No need to have these as part of this RFC, just mentioning it for the record.

eddyb:
I don't think `&Cell<[T]>` -> `&[Cell<T>]` is sound, it violates the no interior references principle, although I'm not sure I can prove any problematic interactions. You'd need code that works on `Cell<[T]>` and doesn't take the existence of the transformation into account.

That said, I do *quite* like `xs[i].set(x)`, there's a nice aesthetic to it.

glaebhoerl:
The idea is that internal references (`&`) together with shared mutability (`Cell`) is safe and sound as long as the structure is "flat": the references don't go under indirections (e.g. `Box`) or branches (`enum`s). That means there is no way for the outer references to invalidate the memory observed by the inner references, because mutation cannot change the structure (because there is only one possible structure), only the data.

Furthermore, given two references to the same memory, one typed as `&Cell<(X, Y, Z)>`, and the other as `&(Cell<X>, Cell<Y>, Cell<Z>)`, doing `cell.set((x, y, z))` with the first is completely equivalent to doing `cell.0.set(x); cell.1.set(y); cell.2.set(z)` with the second, which is some kind of algebraic property I'm sure. (The same principle applies to arrays (slices), it's just nicer to illustrate with tuples.)

camlorn:
@eddyb 
it's hard for me to come up with code here because I don't understand the proposed use case of the first conversion.  The problem as I see it is that the lifetime of the `&Cell<T>` cannot be larger than the lifetime of the `&mut T`, consequently you still have the `&mut T` in the way of all the interesting manipulations you might want to do.

Let me put it this way instead: can someone show me why it's useful, and then I'll come back with an objection if I still have one?

eddyb:
@camlorn I've used it in a constraint solver. A highly heterogenous UI tree contained `f32`s indicating coordinates on screen, scattered inside it. They were mutably traversed and `&mut f32` was turned into `&Cell<f32>`. Now once I had the `&Cell<f32>` I could copy it many types and use it as a "variable reference" in constraints, and every single one of them could update the original if they wanted to.

I could've done all of that through a `Vec<&mut f32>` and integer indices though, so maybe not the best example? It just felt cleaner, at least *that* part (you *really don't* want to read the actual "solver").

camlorn:
@eddyb 
Ah, I kind of see.  It sounds like the kind of thing where it's not useful save in larger programs.  That seems workable, though, yes.

Kimundi:
@camlorn: `&mut T -> &Cell<T>` is also the general approach you can use if you just have a iterator that yields you `&mut T` references, as oppose of a slice directly.

```rust

let set: HashSet<i32> = ...;

let tmp: Vec<&Cell<i32>> = set.iter_mut().map(|x| Cell::from_mut(x)).collect();

for i in &tmp {
    for j in &tmp {
        t.set(j.get() + i.get());
    }
}


```

camlorn:
@Kimundi 
Oh. That's clever.  Though I will say that I don't like the temporary Vec.

Kimundi:
Sure, but the need for the vector is kind of independent to this feature, and only caused by the need to have random access to all elements.

Also, the `&Cell<i32>`'s might just as well have been picked up by a consumer of the iterator directly and stashed away in whatever data structure works with them.

camlorn:
@Kimundi 
I know.  I still don't like the need for temporary intermediate data structures.

One of the problems of Rust (orthogonal to this RFC) is that some things like this suddenly require you to collect into a vec, which is fine until the container is large.  Other such scenarios include deleting multiple keys from a map based on a predicate.  The solution to this problem is for me to write up the extensions I think we should have and put them in an RFC, as I think most of them would be accepted without much argument.

Nevertheless, every time I see something that ends in `.collect::<Vec<_>>()` just because it's Rust and not C++, I get a little sad inside.

bluss:
Collecting into a vec is a workaround for when a more specific solution do not exist yet, for example an (random access?) iterator where each iterator element is a `&Cell<T>`. The canonicalization of this conversion does pave the way for such iterators to be written.

Havvy:
No discussion has been going on in this RFC for almost a month. I'd nominate for FCP, but I don't know think I can.

Kimundi:
Agreed - I updated the RFC text a bit for a post-`T: Copy` world, but so far I haven't seen any actual objection, so I'm putting this into fcp. :)

@rfcbot fcp merge

glaebhoerl:
@Kimundi What do you think about https://github.com/rust-lang/rfcs/pull/1789#issuecomment-260207639?

Kimundi:
@glaebhoerl:

Ah sorry, I meant to have replied to that as part of the last comment. My thoughts are that your proposal sounds like a useful extension I'd like to see in a follow-up RFC, but in this one I'd try to be conservative first by only having the RFC be about the basic `&mut T -> &Cell<T>` transformation (well, and the slice version, which is equivalent to it because you can split up a `&mut [T]` into a number of `&mut T` directly).

Specifically, these are the additional details your proposal would introduce:
- `Cell<T>` right now requires `T: Sized`, so enabling the proposed cast would also have to include allowing `?Sized` cells in general.
- `&Cell<[T]> -> &[Cell<T>]` would "technically" create a interior reference. While I agree with your reasoning that it should be safe, the compiler or language spec might reason about it differently, and I can't fully validate that myself yet. :)
- Related, since this would be a `&T -> &U` cast you would be able to access both representations concurrently, with the same uncertainty as in the argument above.
- Lastly, it seems that if you think it through fully you'd want both bidirectional casting (`&Cell<[T]> <-> &[Cell<T>]` ) as well as a proper extension to tuple types as mentioned by you, which would be a somewhat large API to throw in with the base proposal.

ubsan:
My only request for this is that we don't implement this with `transmute`. The standard library is already a mess of bad practice, and I don't want to add to that.

text/0000-as-cell.md
```diff
@@ -0,0 +1,243 @@
...
+- `&mut T -> &Cell<T>`
+- `&mut [T] -> &[Cell<T>]`
+
+> Note: https://github.com/rust-lang/rfcs/pull/1651 has been accepted recently,
+> so no `T: Copy` bound is needed anymore.
+
+# Motivation
+[motivation]: #motivation
+
+Rusts iterators offer a safe, fast way to iterate over collections while avoiding
```
ubsan: nit: "Rust's"

text/0000-as-cell.md
```diff
@@ -0,0 +1,243 @@
...
+As far as the author is aware, both should already implicitly
+fall out of the semantic of `Cell` and llvms notion of aliasing:
+
+- `Cell` is safe interior mutability based on memcopying the `T`,
+  and thus does not need additional fields.
+- `&mut T -> &U` is a sub borrow, which prevents access to the original `&mut T`
+  for its duration, thus no aliasing.
+
+However, we might have to add an explicit language guarantee about
+the validity of this cast.
```
ubsan: If explicit language is required, I have absolutely failed in the unsafe code guidelines.

Kimundi:
@ubsan Sure, literally using transmute is not meant as a requirement of this RFC.

text/0000-as-cell.md
```diff
@@ -0,0 +1,243 @@
...
+Add conversions functions for `&mut T -> &Cell<T>` and `&mut [T] -> &[Cell<T>]`
+to std, implemented with the equivalent of a simple `transmute()`.
+
+As an initial design, `Cell` would provide them through additional constructors:
+
+```rust
+impl<T: Copy> Cell<T> {
+    fn from_mut<'a>(t: &'a mut T) -> &'a Cell<T> {
+        unsafe { mem::transmute(t) }
+    }
```
ubsan: This is, pretty much literally, the least idiomatic implementation of this; we should not be adding to the standard library's problems as a bastion of unidiomatic and unsafe code. It has two major problems: first, transmute of a pointer (see the docs), and second, transmute without type parameters. Implement it in the following way:

```rust
fn from_mut(t: &mut T) -> &Cell<T> {
  unsafe {
    &*(t as *mut T as *const Cell<T>)
  }
}
```

As another idea, adding a `*mut T -> &Cell<T>` cast function would be good. In that case, the implementation would be:

```rust
fn from_mut(t: &mut T) -> &Cell<T> {
  unsafe { Cell::from_raw(t) }
}
```

text/0000-as-cell.md
```diff
@@ -0,0 +1,243 @@
...
+    fn from_mut_slice<'a>(t: &'a mut [T]) -> &'a [Cell<T>] {
+        unsafe { mem::transmute(t) }
+    }
```
ubsan: Same as the other function.

```rust
fn from_mut_slice(t: &mut [T]) -> &[Cell<T>] {
  unsafe {
    &*(t as *mut [T] as *const [Cell<T>])
  }
}
```

text/0000-as-cell.md
```diff
@@ -0,0 +1,243 @@
...
+in, say, `std::cell`:
+
+```rust
+fn ref_as_cell<T: Copy>(t: &mut T) -> &Cell<T> {
+    unsafe { mem::transmute(t) }
+}
+
+fn slice_as_cell<T: Copy>(t: &mut [T]) -> &[Cell<T>] {
+    unsafe { mem::transmute(t) }
+}
```
ubsan: Change these two functions as well.

text/0000-as-cell.md
```diff
@@ -0,0 +1,243 @@
...
+        unsafe { mem::transmute(t) }
+    }
+}
+
+impl<'a, T: Copy> AsCell for &'a mut [T] {
+    type Cell = &'a [Cell<T>];
+    fn as_cell(self) -> Self::Cell {
+        unsafe { mem::transmute(t) }
+    }
+}
```
ubsan: And these.

text/0000-as-cell.md
Kimundi: I might simply have failed in my research - where would I find the corresponding parts of the guidelines?

text/0000-as-cell.md
Kimundi: Just to be clear, I used `transmute` there because I thought it most concisely shows what the actual cast would do, not because I thought that its the best way to actual implement it.

text/0000-as-cell.md
ubsan: @Kimundi not yet written. But still, having these in the standard library, and being safe, is a better guarantee than anyone can write in documentation.

Kimundi: Sorry, then I don't understand what your initial comment was referring to.

My point with that line in the RFC is that I don't know if the current language spec (aka "what should happen") already validates this specific cast on its own, or if there needs to be a amendment to allow it "officially", rather than just by the implementation of the compiler.

ubsan: fine, it doesn't matter all that much.

glaebhoerl:
@Kimundi of your points:

 * 1 is true, but seems like an inconsequential detail?

 * 2 & 3: well, we can ask the lang team to see if they agree it would be safe, they're reviewing the RFC after all :)

 * 4: I don't see why tuples would have to be included any more than they'd have to be included in the current proposal: after all it's basically just a refactoring of the currently proposed `&mut [T] -> &[Cell<T>]` conversion into a chained `&mut [T] -> &Cell<[T]> -> &[Cell<T>]` conversion, not really adding anything. What would you want bidirectional casting for?

Kimundi:
Hi, sorry for only getting back to this now.

@glaebhoerl: I think I now agree with your proposal. I'll overhaul the PR text to include it.

Kimundi:
I updated the RFC text to include @glaebhoerl's proposal.

However, after having written it and having compared the actual usage the of the slicing API in an example implementation, I'm thinking that we might not to want to do it like this after all, simply because it gets more verbose while offering no real advantage:

- All index impls on `&Cell<[T]>` yield the same as the index impls on `&[Cell<T>]`.
- Additionally, indexing a `&Cell<[T]>` can be expressed as indexing the original `&mut [T]` if its just for a temporary mutation.
- Creating a `&[Cell<T>]`, eg to iterate over it, can become much more verbose:
  ```rust
  let mut v: Vec<i32> = ...;
  let v_slice: &[Cell<T>] = &Cell::from_mut(&mut v[..])[..];
  ```

I'm leaving the change in for now to gather feedback, but I would not be surprised if it ends up being scaled back again before a merge. :)

Kimundi:
I noticed that nothing would prevent keeping the mutability during the cast, so I added a alternative section about `&mut T -> &mut Cell<T>`. Seems to not be actually useful though.

edef1c:
Is there a reason for implementing `Index` over `Deref<Target=[Cell<T>]>`?
It would yield the same benefits, afaict.

Kimundi:
@edef1c: 

True, but making a `Cell<[T]>` derefeable  to `[Cell<T>]` seems like even more of an hack to me, since it is usually only done with smartpointer-like types and not for `as_foo` style conversions. :)

Maybe I'll add it to the alternatives later though.

ubsan:
It seems weird that this proposal doesn't include `UnsafeCell`.

aturon:
We very briefly discussed this RFC in the libs meeting, and are ready for a team-wide review with initial disposition to merge:

@rfcbot fcp merge

solson:
@ubsan Hmm, I went to look up if you could chain this RFC with [`Cell::as_unsafe_cell`](https://doc.rust-lang.org/stable/std/cell/struct.Cell.html#method.as_unsafe_cell) to reach `&UnsafeCell<T>`, but that method is deprecated, and its replacement `Cell::as_ptr` returns `*mut T`.

Is there a usecase for `&UnsafeCell<T>` not covered by `*mut T`?

solson:
Relevant past discussion in https://github.com/rust-lang/rust/issues/27708.

ubsan:
@solson `&UnsafeCell` is a safe-pointer, not a raw-pointer, which means it's always valid, never null, and has a lifetime. `*mut T` should only be used when you're implementing an owned pointer type, or when your pointer is not guaranteed to be valid.

solson:
If there is a concrete use case, the deprecation that was decided in the thread I linked should probably be reversed.

ubsan:
@solson I mean, I don't think there's a usecase there; `&UnsafeCell` is more useful for building unsafe abstractions, like `RefCell`, than actually being used; you'd want to use `as_ptr` on `&UnsafeCell` as well, anyways.

nrc:
Just to be clear, this RFC proposes adding conversion functions only, correct? It does not propose allowing actual casts using the `as` keyword or any implicit coercions? (It would be clearly to use the verb 'convert' rather than 'cast' in the RFC text).

aturon:
@nrc Yes, that's correct, and agreed.

eddyb:
Pinging @pnkfelix for checkmark.

pnkfelix:
The `&Cell<[T]> -> &[Cell<T>]` conversion is unnerving me enough that I don't want to rubber stamp this before going to bed. Will try to re-read that section more carefully tomorrow

aturon:
@pnkfelix BTW, @RalfJung has proved the basic conversion here sound in his model of Rust.

RalfJung:
In case anyone is curious how unreadable that proof is, you can check it out here:
<https://gitlab.mpi-sws.org/FP/LambdaRust-coq/blob/master/theories/typing/lib/cell.v#L192>
It's fairly trivial, because essentially `&mut T` and `&mut Cell<T>` mean the same thing (when you own them -- not when you wrap an `&` around them).

burdges:
Interesting, this is why borrows cannot be downgraded from mutable to immutable, so even fancy non-lexical lifetime stuff cannot make [this](https://is.gd/eOUiq6) work.

pnkfelix:
My issue was not with converting `&mut T` to `&Cell<T>`, but rather with converting `&Cell<[T]>` to `&[Cell<T>]`, which seems distinct to me in terms of the theorizing involved.

But at this point I'm willing to buy into the idea that unsized `[T]` inside of `&Cell<..>` can bubble out, yielding `&[Cell<T>]`. In the end it seems that it boils down to the core notion that `T` and `Cell<T>` must have same representation, regardless of the context they appear in, and any operation overwriting the contents of `&Cell<[T]>` will still preserve the structure of the slice (i.e. its type and, crucially, the number of elements), so the aliasing `&[Cell<T>]` will still remain valid...

RalfJung:
Btw, I completely forgot to stress that the actual proof was not actually done by me, but by Jacques-Henri Jourdan. Credit where credit is due!

> My issue was not with converting &mut T to &Cell<T>, but rather with converting &Cell<[T]> to &[Cell<T>], which seems distinct to me in terms of the theorizing involved.

Indeed that's very different.
And I am actually wondering... why is this okay?  I cannot find a counterexample given the restricted API, but that seems to me more a failure of the API than a feature of this conversion function. In particular, why is `swap` not available for unsized types? It should be perfectly implementable, and excluding it feels unnatural.

However, with `swap`, I think I can break the slice conversion function...
```rust
fn truth_about_slices<T, F: FnOnce()>(x: &[T], f: F) {
  let len = x.len();
  f();
  // This must never fail -- given a shared borrow to a slice, it's structure cannot ever change.
  // That's something compiler optimizations and unsafe code should be able to rely on.
  assert_eq!(x.len(), len);
}

fn bad_cell() {
  let init : Box<[i32]> = Box::new([1]);
  let x : &Cell<[i32]> = Cell::from_mut(&mut *init);
  let y : &[Cell<i32>] = convert_cell_slice(x);
  truth_about_slices(y, || {
    let emp : Box<[i32]> = Box::new([]);
    x.swap(&mut Cell::from_mut(&mut *emp));
  }); // *oops* assertion fails because length of the slice changed from 1 to 0
}
```

Basically, it seems the only reason `&Cell<[T]> -> &[Cell<T>]` is sound is that `Cell<T>` for unsized `T` is not actually permitting replacing the entire content of the `Cell` with something else. How does that make any sense? Sounds like a very unnatural restriction to me, given the nature of `Cell`. `&mut [T] -> &[Cell<T>]` should be fine though.

I know I'm kind of late to the party here, but IMHO the proposed API seems to work for the wrong reasons, and that's not a good ground for accepting it.

withoutboats:
>  In particular, why is swap not available for unsized types? It should be perfectly implementable, and excluding it feels unnatural.

How would you implement `swap` for an unsized type? swap memcpys the bytes from one address to aonther, but different values of unsized types can be different sizes, so they will require differently sized allocations.

eddyb:
@RalfJung `&Cell<[T]>` and `&[Cell<T>]` are isomorphic and they both keep the length in *the reference*.
What you wrote there is fundamentally impossible in Rust: you can at most swap the references, or mutate the bytes inside, but not change a reference *given a copy of it*.

RalfJung:
Ah, I see... even with a `&mut [T]`, the size cannot actually change because there is no way to change the size of the allocation. So even the following assertion would always hold:
```rust
fn truth_about_slices<T, F: FnOnce(&mut [T])>(x: &mut [T], f: F) {
  let len = x.len();
  f(x);
  assert_eq!(x.len(), len);
}
```
I was not aware of this, sorry for the noise.

Well, actually the reason is even deeper than not changing the size... if I think about trait objects, it's not legal even to change the type to sth. else of the same size, right? The existential quantifier is, so to speak, outside of the reference, because it is literally stored on the "outside". Hm.

Kimundi:
Glad to see this RFC received so much formal scrutiny. :)

@nrc / @aturon: Sorry, I totally forgot to address your comments. I've pushed updated wordings.

@pnkfelix: Yeah, the slice conversion is independent to the core proposal, and weird in its own way. If there where major objections I'd be happy to remove it from this RFC, but apparently no one sees an actual issue with it. (notwithstanding @RalfJung's confusion 😄)

aturon:
Huzzah! This RFC has been **merged**. [Tracking issue](https://github.com/rust-lang/rust/issues/43038).

whitequark:
Question, why is this merged but [slice::ref_slice is deprecated](https://github.com/rust-lang/rust/issues/27774)? Aren't these fundamentally fairly similar?

eddyb:
@whitequark There is no slice <-> non-slice conversion here. I don't know why `ref_slice` is deprecated, it's really useful in quite a few cases.

whitequark:
@eddyb I mean this [comment](https://github.com/rust-lang/rust/issues/27774#issuecomment-150058618) by @alexcrichton:
>The libs team discussed this during triage today and the decision was to deprecate these functions. The standard library currently doesn't house many nuggets of functionality like this (e.g. see the recent discussions around Box::leak), and we've also deprecated similar functionality (Option::as_slice).
>
>While certainly useful from time to time, it's not clear that these functions belong in the standard library right now. There can always be a helper crate implementing all sorts of conversions like this, however!

This seems to me to fall right into the "little nugget of functionality" area. I would rather we get ref_slice back, it's pretty useful in embedded contexts, if you have no Box!

mzji:
@whitequark Then should it go to `libcore`? These useful functions could be a part of core, just won't be exported by std.

whitequark:
I don't think we have precedent for that and it seems like a really odd decision to me.

edef1c:
Note that `ref_slice` is available as a [Cargo crate](https://crates.io/crates/ref_slice), should you need it.
It doesn't require any additional unsafe guarantees from the compiler, so it is fine to implement outside the stdlib.

whitequark:
> It doesn't require any additional unsafe guarantees from the compiler, so it is fine to implement outside the stdlib.

This proposal could have been [done the same way](https://github.com/rust-lang/rfcs/pull/1789/files#diff-885ec33ec14f3f25b044597d029dc473R379) and without even any new RFCs, just mark `Cell` as `#[repr(transparent)]` I think.

aturon:
I think the libs team would be open to reconsidering `ref_slice` (which was deprecated quite a long time ago), if there are more documented use-cases.

nikomatsakis:
For what it's worth, I still find uses for `ref_slice` relatively frequently. Luckily, in the compiler, we have a copy available. Here is a quick analysis of the various uses in the compiler. As one would expect, there are two main patterns.

**You have a function that takes an `&[T]`, and you have a `&T`.** Here are the cases that seem to be like that:

```
src/librustc/middle/resolve_lifetime.rs:431:            self.resolve_elided_lifetimes(slice::ref_slice(lifetime_ref));
src/librustc_const_eval/check_match.rs:100:        self.check_patterns(false, slice::ref_slice(&loc.pat));
src/librustc_const_eval/check_match.rs:108:            self.check_patterns(false, slice::ref_slice(&arg.pat));
src/librustc_lint/unused.rs:93:                self.check_unused_mut_pat(cx, slice::ref_slice(&l.pat));
src/librustc_lint/unused.rs:106:            self.check_unused_mut_pat(cx, slice::ref_slice(&a.pat));
src/librustc_typeck/astconv.rs:850:        self.prohibit_type_params(slice::ref_slice(item_segment));
src/librustc_typeck/astconv.rs:925:        self.prohibit_type_params(slice::ref_slice(item_segment));
src/librustc_typeck/check/mod.rs:3987:        (def, Some(ty), slice::ref_slice(&**item_segment))
src/librustc_typeck/check/mod.rs:4119:                Some(e) => ref_slice(e),
```

**You have an enum that sometimes has `T` and sometimes `Vec<T>`, and you want to return a `&[T]`.** This occurs primarily in the MIR, except that we are returning `Cow<[T]>`:

```
src/librustc/mir/mod.rs:579:            Goto { target: ref b } => slice::ref_slice(b).into_cow(),
src/librustc/mir/mod.rs:586:                slice::ref_slice(t).into_cow(),
src/librustc/mir/mod.rs:587:            Call { destination: None, cleanup: Some(ref c), .. } => slice::ref_slice(c).into_cow(),
src/librustc/mir/mod.rs:595:                slice::ref_slice(target).into_cow()
src/librustc/mir/mod.rs:598:            Assert { ref target, .. } => slice::ref_slice(target).into_cow(),
```

I continue to think this function is a natural candidate for the standard library. It's a bit of unsafe code that is useful as a bit of glue in various scenarios and absolutely no dependencies or runtime implications. It is not entirely **obvious** that it is safe -- in particular, it requires understanding Rust's memory layout rules (which of course are not entirely defined). So putting it in the standard library is a way for us to advertise that indeed it is safe. =) It's also something that people might not realize is possible, and hence they wouldn't know to search for the crate. 

It's also ok for it to exist in a crate, of course, but I don't think there's any good reason to put it there.

whitequark:
I'd like to get it back in std. What's the process here? Is it an RFC or just a PR?

aturon:
@whitequark I think it's a small enough addition that a PR suffices. Feel free to CC me.

comex:
This is late, but I have a concern about this RFC's design.  To go from `&mut [T]` to `&[Cell<T>]` it requires going through `&Cell<[T]>`.  That's a cute decomposition and works fine - for `Cell`.

However, a natural extension to the proposed use case would be a thread safe variant: in particular, a way to go from `&mut u8` to `&AtomicU8` and from `&mut [u8]` to `&[AtomicU8]`, and similarly for other atomic types.  That would be its own RFC of course, but notably, the API design couldn't directly mirror this one; there would have to be a direct conversion function/method from `&mut [u8]` to `&[AtomicU8]`, since there's no type like `&Atomic<[u8]>` to serve as an intermediary.  For the sake of consistency, therefore, it might make sense to revise this RFC to add a direct conversion from `&mut [T]` to `&[Cell<T>]`.

edef1c:
@comex You also can't swap out the contents of a `Cell<[T]>` with another `[T]` freely, so that seems analogous?

glaebhoerl:
I was going to reply with something similar but I think their point was that there is no `Atomic<T>` type in the first place that you could instantiate to `Atomic<[T]>`.

comex:
@edef1c Ah... I kind of forgot about that.  It might be possible to support the full `Cell` API on `Cell<[T]>` in the future with stack-based DSTs.  But if that doesn't happen / isn't desired, it only strengthens my belief that the design should be changed.  If `Cell<[T]>` does not provide any of the normal Cell functionality, what's the point of having it?  It's not 'modular' to go through an intermediary type if the intermediary type is completely useless - other than for the purpose of being an intermediary for that exact conversion.

But in any case, as @glaebhoerl said, there is no `Atomic<T>`.  Perhaps there should be one in the future, but it would be quite strange to have an `Atomic<T>` that can't actually be read/written atomically...

PR #1792, opened 2016-11-14
Introduce Option::<&T>::borrowed

nox:


BurntSushi:
[Rendered.](https://github.com/nox/rust-rfcs/blob/option-borrowed/text/0000-option_borrowed.md)

sfackler:
I'm not sure about `Borrow` vs `Deref` and naming, but I've definitely run into this, and it's kind of annoying/ugly to have to do the current `foo.map(|s| &**s)` dance. We'd probably want this on `Iterator` as well?

BurntSushi:
Yeah, I know I've at least burned `x.as_ref().map(|x| &**x)` into my brain as well. At this point, I barely even notice it, but taking a step back, I can see how unintuitive it might be.

One thing I worry about is the overlap between `as_ref` and `borrowed`. As a thought experiment, how would you explain the difference between them? (Keeping in mind that [`as_ref`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref) in this context is a concrete method on `Option` and not part of the `AsRef` trait.)

sfackler:
I believe that the only difference is that `borrowed` is required to be consistent wrt Hash/Eq/Ord, but yeah, the number of traits that do basically the same thing here is not great.

bluss:
I would like to focus on AsRef, and not spread Borrow too far away from its use case (which was originally HashMap::get, BTreeMap::get).

sfackler:
That seems reasonable, though naming will be a bit less obvious (please no `as_refed`!).

eddyb:
There's probably some connection with #1403, as people have mentioned `AsRef` coercions recently.

nox:
> I would like to focus on `AsRef`, and not spread Borrow too far away from its use case (which was originally `HashMap::get`, `BTreeMap::get`).

Isn't that already a lost battle, with `ToOwned` and `Cow`? And neither `HashMap` nor `BTreeMap` need `BorrowMut`, right? I would rather make `AsRef` less used.

nox:
The book says:

> Choose `Borrow` when you want to abstract over different kinds of borrowing, or when you’re building a data structure that treats owned and borrowed values in equivalent ways, such as hashing and comparison.

> Choose `AsRef` when you want to convert something to a reference directly, and you’re writing generic code.

So I would rather prefer we use `Borrow` for this.

bluss:
Borrow is constrained by its hash/ordering guarantees, and those have no special connection to this use case. AsRef is more general and versatile, supporting `str` → `Path`, `String` → `[u8]` and so on.

steveklabnik:
I also believe that this would be `AsRef`; @aturon and I had a lot of conversations about asref/borrow in the past...

alexcrichton:
I [agree with](https://github.com/rust-lang/rfcs/pull/1792#issuecomment-260348202) @bluss that we likely want to leverage `AsRef` here instead of `Borrow` to allow for other conversions like `Option<Cow<str>>` to `Option<&str>` and such. I'm not entirely sure what the method would be however, as I also [agree with](https://github.com/rust-lang/rfcs/pull/1792#issuecomment-260348827) @sfackler that `as_refed` is pretty bad.

Also the RFC motivation may want to be updated slightly, the stated problem is:

> How to convert from `Option<String>` to `Option<&str>` is sometimes asked

Yet the RFC does not solve this problem as it defines a method on `Option<&'a T>`. Instead, I would have expected the signature:

```rust
impl<T> Option<T> {
    fn name<U: Bound<T>>(&self) -> Option<&U> {
        self.as_ref().map(Bound::method)
    }
}
```

nox:
I just meant that with this method, we have the solution `opt.as_ref().borrowed()`, it wasn't a mistake, will reformulate.

nox:
> I agree with @bluss that we likely want to leverage `AsRef` here instead of `Borrow` to allow for other conversions like `Option<Cow<str>>` to `Option<&str>` and such.

AFAIK, `Cow` implements `Borrow`, and `AsRef` doesn't have that many more implementations.

nox:
It should also be noted that there is no implementation of `AsRef<T>` for `T`, which may or may not be a problem.

lfairy:
I'm not bothered by `.as_refed()` that much. After all, the convention of naming single-method traits as verbs (e.g. `Read` and `Write`) could be considered "ugly" as well, yet we got used to that over time.

If `.as_refed()` isn't feasible then these names would make good alternatives:

* `.inner_as_ref()` -- by analogy with `.into_inner()`

* `.map_as_ref()` -- contraction of `.map(AsRef::as_ref)`

nox:
That would mean calling .as_ref().as_refed() which is even uglier.

I don't understand why AsRef is promoted as the way to go even for generic functions when that trait doesn't even have an impl for T.

-- 
Anthony Ramine

lfairy:
On the other hand, the lack of a general identity impl lets `AsRef` cut through multiple layers of references in one go.

For example, you can convert from `&&String` to `&str` using `AsRef` but similar code using `Borrow` will fail. ([playground](https://play.rust-lang.org/?gist=d3e50ea9c30558da1b9111bcee348943&version=stable&backtrace=0))

Coherence stops us from adding a corresponding feature to `Borrow`. So for the use cases I'm familiar with -- various combinations of `&`, `&mut`, `String`, and `str` -- the `AsRef` solution is superior.

nox:
How often do you hold onto ˋOption<&&String>ˋ though?

-- 
Anthony Ramine

eddyb:
I really hope this ends up going through the coercion mechanism and not more ad-hoc hierarchies.

oli-obk:
> I really hope this ends up going through the coercion mechanism and not more ad-hoc hierarchies.

This please! Allow any `enum Foo<T> { X(T), .. }` value `Foo<T>` to coerce to `Foo<U>` if `T` can coerce to `U` (as long as `Foo` has no uses of T in subtypes where coercion isn't possible (like `GenericStruct<T>`)). All enum variants are public, so we could write a match by hand that does the coercion for every variant, so there's no reason to prevent it being done automatically.

eddyb:
I would prefer it opt-in (like `CoerceUnsized` just generalized), but yes.

nox:
Using `Borrow` means we could also have `borrowed_mut` (on `Option<&mut T>`) using `BorrowMut`, whereas `as_muted` using `AsMut` would also be quite weird.

As for implicit coercions, that's a rabbit hole I'm very wary of.

nox:
It's not like we must limit ourselves to only one choice... Would people find `Option::<&T>::dereferenced` controversial?

nox:
Any news on this?

nox:
Happy new year. Was a decision reached for that?

lfairy:
Another advantage of `AsRef`: we can handle both `&Option<String>` and `Option<&String>` using identical code. Instead of thinking about whether to use `.as_ref().borrowed()` or `.borrowed()` based on the input type, we can just write `.map_as_ref()` every time. See this [playground][1] for how this could work.

[1]: https://play.rust-lang.org/?gist=8b3c2eba4f9a03a00d4d30bf71b12d47&version=stable&backtrace=0

nox:
I don't understand how that is an advantage, and how `map_as_ref` is a name that makes sense.

lfairy:
> I don't understand how that is an advantage

Since this method is sugar, it is an advantage if the user doesn't have to think about how to use it.

We have deref conversions for the same reason: they let us just write `&x` instead of thinking about whether to write `&*x` or `&**x` or `&x[..]`.

That said, the lack of general identity impl for `AsRef` means that this solution won't work everywhere either. Bother. That coercion [idea](https://github.com/rust-lang/rfcs/pull/1792#issuecomment-261224476) is looking better and better.

> and how `map_as_ref` is a name that makes sense.

I was using that name as an placeholder, and am not that attached to it.

nox:
> Since this method is sugar, it is an advantage if the user doesn't have to think about how to use it.

I disagree, the user already has to care about `&Option<T>` vs `Option<&T>` in plenty of case, so sometimes calling `.as_ref()` and sometimes not should already be a known thing to them.

> That coercion idea is looking better and better.

That coercion idea is implicitness over explicitness, and I think that would be a bad precedent to set in Rust.

aturon:
The libs team discussed this RFC during triage yesterday, and had some thoughts:

As @lfairy argues, the ideal situation is to have a single method, `as_ref`, that can support both the current use case and the one proposed by this RFC. We actually tried out such a method before 1.0, but ran into type inference problems, since the target type becomes unknown.

It's possible that we could introduce some opt-in type inference fallback scheme, where a method could indicate that a given generic parameter should default to a certain type if inference is unable to otherwise determine it. But fallback in general is quite brittle, and we've been wary of adding more cases.

The coercion idea that @eddyb mentions is a much more promising direction for making the above possible.

In any case, that's the long term view. In the near term, the question is whether we're open to landing a separate method that may eventually be deprecated. I worry that the relationship between `borrowed` and `as_ref` will only add to the confusion, and that `borrowed` won't be a terribly discoverable method. OTOH, it's certainly much nicer than mapping with `&**`. The libs team in general was on the fence here.

FWIW, given the current design of the `AsRef` and `Borrow` traits, going with `Borrow` here is effectively the only option (as @lfairy has discussed). We do hope to eventually add a blanket impl for `AsRef`, once intersection impls are possible.

nox:
@aturon Do I need to make any amends on the RFC? Any action required of me?

aturon:
After another round of discussion on the libs team, I propose we go forward with this RFC as is, with the intention to *also* explore additions to the type inference system that would obviate the need for it. If there's impetus to stabilize this method before we improve type inference, so be it; we can always deprecate it once it's no longer needed.

@rfcbot fcp merge

BurntSushi:
In addition to the `AsRef` concerns mentioned by others, I would personally like to see the section on "how do we teach this" fleshed out before I'd be comfortable with this, because it seems like a pretty confusing addition to the existing API. I'll repeat my thought experiment:

> One thing I worry about is the overlap between `as_ref` and `borrowed`. As a thought experiment, how would you explain the difference between them?

brson:
Since this draws a comparison to `Option::cloned` to justify `Option::borrowed`, and there exists a similar `Iterator::cloned`, we should probably figure out the appropriate parallel in `Iterator::borrowed`.

nox:
@brson I would very much fancy this!

I'll try to write the teaching section tomorrow if that's ok with you @BurntSushi.

alexcrichton:
This was discussed at libs triage yesterday and the conclusions were that if the following changes were made we'd all be comfortable going to FCP:

- Considering `borrowed` on iterators as well as `Option` (as proposed here) for symmetry with `cloned`
- Changing the method to `&self` instead of `self`
- Updating the teaching section according to @BurntSushi 
- Continue using `Borrow`, instead of `AsRef` as it would be too "weird" to have `Option::as_ref` not use `AsRef` and then have `Option::borrowed` use `AsRef` with `Borrow` not used anywhere.

nox:
> Changing the method to &self instead of self

What? Why?

nox:
https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned takes `self`, so this makes no sense to me. I'll do the other changes though.

alexcrichton:
I [commented above](https://github.com/rust-lang/rfcs/pull/1792#issuecomment-298723739) that I [commented farther above](https://github.com/rust-lang/rfcs/pull/1792#issuecomment-260460316) with rationale for `&self`, not `self`

alexcrichton:
er didn't mean to close

Xion:
This RFC doesn't really make the common use case of "borrowing" `Option<String>` as `Option<&str>` (i.e. the stated goal in the motivation section) any more convenient.

To the best of my knowledge, the shortest way to do it currently is:
```rust
opt_string.as_ref().map(|s| s.as_str())
```

which this RFC improves only marginally to:
```rust
opt_string.as_ref().borrowed()
```

What I'd like to do is get rid of `as_ref` entirely here, and allow `borrowed` to coerce to whatever exact borrowed type is necessary in the current context. I'm not sure this is expressible in current Rust; my attempt would be somewhere along these lines:
```rust
impl<T> Option<T> {
    fn borrowed<U>(&self) -> Option<&U>
        where T: Borrow<U>
    {
        self.as_ref().map(|x| x.borrow())
    }
}
```

which may or may not be rephrasable to an equally powerful `AsRef` variant. (Since I don't think I understand the `AsRef`/`Borrow` distinction well enough, I'm preferring one over the other here).

nox:
> Changing the method to `&self` instead of `self`

If we do that, what happens for the type `Option<&String>`? We wouldn't be able to call `borrowed` anymore. I still don't think that is a good idea.

daboross:
@nox If I understand correctly, this is a reason to do both `&self` and `AsRef` at once: `AsRef` works for `Option<&String>` to `Option<&str>` as well, and having the method accept `&self` would let it work for `Option<String>` too.

`x.as_ref().map(|x| x.to_str())` really isn't that bad: is it worth adding an additional method if `x.as_ref().borrowed()` is still required?

nox:
@daboross `AsRef<T>` does not allow getting a `Option<&T>` for all `self: Option<&T>`, as mentioned earlier. `x.as_ref().map(|x| x.to_str())` really is bad, and my use cases aren't about strings anyway.

daboross:
@nox Sorry, I know I should probably be getting this, but I don't seem to be.

Here's an reference implementation of `borrowed` with `AsRef`, which as far as I can tell works for all of the case you mentioned?


`Option<&String>`, `Option<String>` and `Option<&str>` can all be turned into `Option<&str>` with:

```
    fn borrowed<U: ?Sized>(&self) -> Option<&U>
        where T: AsRef<U>
```

test case: http://play.integer32.com/?gist=6686926f4114f2104b45588b9973711c&version=undefined

If there was a case I missed, could you elaborate on it specifically?

I really do want to find a good solution for this, I just can't find a place where `AsRef` doesn't work and `Borrow` does. As far as I can tell, `AsRef` just allows for more use cases.

nox:
It doesn't work for arbitrary types.

-- 
Anthony Ramine

daboross:
OK - I think I understand now.

If it isn't about the String use case though, could the RFC be updated to remove that from the motivation?

I can get behind a Borrow-based method, but I don't think that it is the best solution if the only motivation in the RFC is for `Option<&String>` to `Option<&str>`.

Thank you for clarifying that, and sorry I didn't completely get that before.

aturon:
I'm sorry for the lack of follow-up here; it's been difficult to prioritize this particular API, given some of the issues we've covered on this thread. At this point, I'm inclined to postpone consideration; there's enough else going on that I'd like to let the dust settle a bit and see whether language-based solutions are shaping up.

@rfcbot fcp postpone

nox:
> I commented above that I commented farther above with rationale for `&self`, not `self`

@alexcrichton Sorry but I don't see a rationale in any of these two comments.

aturon:
whoa there rfcbot...

aturon:
I'm going to close this RFC as **postponed**. This is still a significant paper cut that would be great to fix, but work during the impl period on the trait system may provide an alternative, more general way of doing so. Thanks, @nox!

PR #1806, opened 2016-12-02
Default Struct Field Values

KodrAus:
[Rendered](https://github.com/KodrAus/rfcs/blob/master/text/0000-default-fields.md).

### DIFF
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -152,7 +152,7 @@ let foo = Foo {
 };
 ```
 
-The mechanism isn't exactly a shorthand because the structure can still be initialised using field defaults even if `b` is private. The caller still can't interact with private fields directly so privacy isn't violated.
+The mechanism isn't exactly a shorthand because the `struct` can still be initialised using field defaults even if `b` is private. The caller still can't interact with private fields directly so privacy isn't violated.
 
 When a caller doesn't supply a field value during initialisation and there is no default available then the `E0063` missing field error applies.
 
@@ -274,20 +274,17 @@ let foo = data::Foo {
 
 ## Field Privacy
 
-Default values for fields are opted into by the `struct` definition, rather than the caller initialising the structure. Field privacy doesn't need to be violated to initialise a structure.
+Default values for fields are opted into by the `struct` definition, rather than the caller initialising the `struct`. Field privacy doesn't need to be violated to initialise a `struct`.
 
 # How We Teach This
 [how-we-teach-this]: #how-we-teach-this
 
 Field defaults look similar to functional record updates, but solve different problems. New users could be confused by the similarity and be unsure when to use either feature. We can easily distinguish the two for new users:
 
-- If you're in control of the structure definition, you should opt for field defaults.
-- If you're only initialising structures, you should opt for functional record updates.
+- If you're in control of the `struct` definition, you can use field defaults.
+- If you're only initialising `struct`s that don't have defaults or private fields, you can use functional record updates.
 
-As for constructors, structure literals and builders, we can recommend the following rule of thumb:
-
-- If your initialiser needs non-constant expressions or various default values can be logically grouped, you should opt for constructor functions.
-- If your initialiser needs non-constant expressions, has many common permutations or is otherwise too complex for structure literals or constructors, then you should opt for builders.
+Field defaults are a tool for producers and functional record updates are a tool for consumers.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -340,9 +337,9 @@ This has the effect of simplifying the definition, but also requiring readers to
 
 ## Implicit syntax for opting into field defaults
 
-Invoke field defaults implicitely instead of requiring a `..` in the struct initialiser. This would be more in line with how other languages handle default values, but is less explicit. It would also be different from pattern matching for structures, that require a `..` to ignore unnamed fields.
+Invoke field defaults implicitely instead of requiring a `..` in the initialiser. This would be more in line with how other languages handle default values, but is less explicit. It would also be different from pattern matching for `struct`s, that require a `..` to ignore unnamed fields.
 
-A future RFC could propose making the `..` optional for all places it's required when dealing with structures.
+A future RFC could propose making the `..` optional for all places it's required when dealing with `struct`s.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -18,6 +18,7 @@ struct Foo {
 let foo = Foo {
     a: "Hello",
     c: 42,
+    ..
 };
 ```
 
@@ -64,17 +65,19 @@ struct Foo {
 
 // Overriding a single field default
 let foo = Foo {
-    b: false
+    b: false,
+    ..
 };
 
 // Override multiple field defaults
 let foo = Foo {
     a: "Overriden",
     c: 1,
+    ..
 };
 
 // Override no field defaults
-let foo = Foo {};
+let foo = Foo { .. };
 ```
 
 # Detailed design
@@ -125,6 +128,7 @@ Field defaults are like a shorthand for the 'real' initialiser, where values for
 let foo = Foo {
     a: "Hello",
     c: 42,
+    ..
 };
 ```
 
@@ -141,6 +145,19 @@ let foo = Foo {
 The mechanism isn't exactly a shorthand because the structure can still be initialised using field defaults even if `b` is private. The caller still can't interact with private fields directly so privacy isn't violated.
 
 When a caller doesn't supply a field value during initialisation and there is no default available then the `E0063` missing field error applies.
+
+Field defaults are only considered for missing fields when the caller supplies a `..` at the end of the initialiser. Otherwise the standard `E0063` error applies with additional help when the field has a default value available:
+
+```rust
+let foo = Foo {
+    a: "Hello",
+    c: 42,
+};
+
+// error: missing field `b` in initializer of `Foo`.
+// help: `b` has a default value. Try adding `..` so its default value will be used:
+// `let foo = Foo { a: "Hello", c: 42, .. }`
+```
 
 ## Order of precedence
 
@@ -197,16 +214,18 @@ mod data {
     pub struct Foo {
         pub a: &'static str,
         pub c: i32,
+        _marker: () = ()
     }
 }
 
 let foo = data::Foo {
     a: "Hello",
     c: 42,
+    ..
 }
 ```
 
-We can add a new field `b` to this `struct` with a default value, and the calling code doesn't change:
+Using a private marker field with a default value forces callers to opt-in to field defaults. We can now add a new field `b` to this `struct` with a default value, and the calling code doesn't change:
 
 ```rust
 mod data {
@@ -214,12 +233,14 @@ mod data {
         pub a: &'static str,
         pub b: bool = true,
         pub c: i32,
+        _marker: () = ()
     }
 }
 
 let foo = data::Foo {
     a: "Hello",
     c: 42,
+    ..
 }
 ```
 
@@ -237,6 +258,7 @@ mod data {
 let foo = data::Foo {
     a: "Hello",
     c: 42,
+    ..
 }
 ```
 
@@ -293,13 +315,11 @@ struct Foo {
 
 This has the effect of simplifying the definition, but also requiring readers to manually work out what the type of the right-hand-side is. This could be less of an issue with IDE support.
 
-## Explicit syntax for opting into field defaults
+## Implicit syntax for opting into field defaults
 
-Field defaults could require callers to use an opt-in syntax like `..`. This would make it clearer to callers that additional code could be run on struct initialisation, weakening arguments against more powerful default expressions. However it would prevent field default from being used to maintain backwards compatibility, and reduce overall ergonomics. If it's unclear whether or not a particular caller will use a default field value then its addition can't be treated as a non-breaking change. 
+Invoke field defaults implicitely instead of requiring a `..` in the struct initialiser. This would be more in line with how other languages handle default values, but is less explicit. It would also be different from pattern matching for structures, that require a `..` to ignore unnamed fields.
 
-This could be worked around by adding a private field to a `struct` that does nothing, forcing a caller to opt-in to potential field defaults.
-
-The goal of the design proposed in this RFC is to let users build a struct as if its default fields weren't there.
+A future RFC could propose making the `..` optional for all places it's required when dealing with structures.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -286,6 +286,8 @@ Field defaults look similar to functional record updates, but solve different pr
 
 Field defaults are a tool for producers and functional record updates are a tool for consumers.
 
+We should document field defaults in the Rust Reference sections 6.1.5 "Structs" and 7.2.4 "Struct expressions", with a one-sentence mention in 6.1.6 "Enumerations" that struct-like enum variants can have default fields just like structs.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -88,8 +88,18 @@ let foo = Foo { .. };
 In the definition of a `struct`, a default value expression can be optionally supplied for a field:
 
 ```
-struct_field ::= vis? ident ':' type_path |
-                 vis? ident ':' type_path '=' expr
+struct_field : vis? ident ':' type_path |
+               vis? ident ':' type_path '=' expr
+```
+
+Initialisers can then opt-in to use field defaults for missing fields by adding `..` to the end of the initialiser:
+
+```
+struct_init_fields : struct_field_init ? [ ',' struct_field_init ] *
+
+struct_init : '{' 
+    struct_init_fields [ ".." | ".." expr ] ?
+'}'
 ```
 
 The syntax is modeled after constant expressions. Field defaults for tuple structs are not supported.
@@ -265,6 +275,19 @@ let foo = data::Foo {
 ## Field Privacy
 
 Default values for fields are opted into by the `struct` definition, rather than the caller initialising the structure. Field privacy doesn't need to be violated to initialise a structure.
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+Field defaults look similar to functional record updates, but solve different problems. New users could be confused by the similarity and be unsure when to use either feature. We can easily distinguish the two for new users:
+
+- If you're in control of the structure definition, you should opt for field defaults.
+- If you're only initialising structures, you should opt for functional record updates.
+
+As for constructors, structure literals and builders, we can recommend the following rule of thumb:
+
+- If your initialiser needs non-constant expressions or various default values can be logically grouped, you should opt for constructor functions.
+- If your initialiser needs non-constant expressions, has many common permutations or is otherwise too complex for structure literals or constructors, then you should opt for builders.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -1,0 +1,305 @@
+- Feature Name: Default Fields
+- Start Date: 2016-12-03
+- RFC PR: 
+- Rust Issue: 
+
+# Summary
+[summary]: #summary
+
+Allow `struct` definitions to supply default values for individual fields, and then allow those fields to be omitted from `struct` initialisation:
+
+```rust
+struct Foo {
+    a: &'static str,
+    b: bool = true,
+    c: i32,
+}
+
+let foo = Foo {
+    a: "Hello",
+    c: 42,
+};
+```
+
+# Motivation
+[motivation]: #motivation
+
+Rust allows you to create an instance of a `struct` using a literal syntax. This requires that all fields in the `struct` are assigned a value, so it can be inconvenient for large `struct`s whose fields usually receive the same values. Literals also can't be used to initialise `struct`s with innaccessible private fields. Functional record updates can reduce noise when a `struct` derives `Default`, but are also invalid when the `struct` has private fields.
+
+To work around these shortcomings, you can create constructor functions:
+
+```rust
+struct Foo {
+    a: &'static str,
+    b: bool,
+    c: i32,
+}
+
+impl Foo {
+    // Constructor function.
+    fn new(a: &'static str, c: i32) -> Self {
+        Foo {
+            a: a,
+            b: true,
+            c: c
+        }
+    }
+}
+
+let foo = Foo::new("Hello", 42);
+```
+
+The problem with a constructor is that you need one for each combination of fields a caller can supply. To work around this, you can use builders, like [`process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html) in the standard library. Builders enable more advanced initialisation, but need additional boilerplate.
+
+This RFC proposes a solution to improve `struct` literal ergonomics, so they can be used for `struct`s with private fields, and to reduce initialisation boilerplate for simple scenarios. This is achieved by letting callers omit fields from initialisation when a default is specified for that field. This syntax also allows allows fields to be added to `struct`s in a backwards compatible way, by providing defaults for new fields.
+
+Field defaults allow a caller to initialise a `struct` with default values without needing builders or a constructor function:
+
+```rust
+struct Foo {
+    a: &'static str = "Hello",
+    b: bool = true,
+    c: i32 = 42,
+}
+
+// Overriding a single field default
+let foo = Foo {
+    b: false
+};
+
+// Override multiple field defaults
+let foo = Foo {
+    a: "Overriden",
+    c: 1,
+};
+
+// Override no field defaults
+let foo = Foo {};
+```
+
+# Detailed design
+[design]: #detailed-design
+
+## Grammar
+
+In the definition of a `struct`, a default value expression can be optionally supplied for a field:
+
+```
+struct_field ::= vis? ident ':' type_path |
+                 vis? ident ':' type_path '=' expr
+```
+
+The syntax is modeled after constant expressions. Field defaults for tuple structs are not supported.
+
+## Interpretation
+
+The value of a field default must be a compile-time expression. So any expression that's valid as a `const` can be used as a field default. This ensures values that aren't specified by the caller are deterministic and cheap to produce. A type doesn't need to derive `Default` to be valid as a field default.
+
+Valid:
+
+```rust
+struct Foo {
+    a: &'static str,
+    b: bool = true,
+    c: i32,
+}
+```
+
+Invalid:
+
+```rust
+struct Foo {
+    a: &'static str,
+    b: Vec<bool> = Vec::new(),
+                   ^^^^^^^^^^
+                   // error: calls in field defaults are limited to struct and enum constructors
+    c: i32,
+}
+```
+
+The above error is based on `E0015` for trying to initialise a constant with a non-constant expression. As the scope of constant expressions changes this message will change too.
+
+Field defaults are like a shorthand for the 'real' initialiser, where values for missing fields are added with the supplied default expression:
+
+```rust
+let foo = Foo {
+    a: "Hello",
+    c: 42,
+};
+```
+
+is equivalent to:
+
+```rust
+let foo = Foo {
+    a: "Hello",
+    b: true,
+    c: 42,
+};
+```
+
+The mechanism isn't exactly a shorthand because the structure can still be initialised using field defaults even if `b` is private. The caller still can't interact with private fields directly so privacy isn't violated.
+
+When a caller doesn't supply a field value during initialisation and there is no default available then the `E0063` missing field error applies.
+
+## Order of precedence
+
+Supplied field values take precedence over field defaults:
+
+```rust
+// `b` is `false`, even though the field default is `true`
+let foo = Foo {
+    a: "Hello",
+    b: false,
+    c: 42,
+};
+```
+
+Supplied field values in functional updates take precedence over field defaults:
+
+```rust
+// `b` is `false`, even though the field default is `true`
+let foo = Foo {
+    a: "Hello",
+    c: 0,
+    ..Foo {
+        a: "Hello",
+        b: false,
+        c: 0
+    }
+};
+```
+
+## Deriving `Default`
+
+When deriving `Default`, supplied field defaults are used instead of the type default. This is a feature of `#[derive(Default)]`.
+
+```rust
+#[derive(Default)]
+struct Foo {
+    a: &'static str,
+    b: bool = true,
+    c: i32,
+}
+
+// `b` is `true`, even though `bool::default()` is `false`
+let foo = Foo::default();
+```
+
+Field defaults allow `#[derive(Default)]` to be used more widely because the types of fields with default values don't need to implement `Default`.
+
+## Enabling backwards compatibility
+
+With no special syntax, additional fields can be added to a struct in a non-breaking fashion. Say we have the following API and consumer:
+
+```rust
+mod data {
+    pub struct Foo {
+        pub a: &'static str,
+        pub c: i32,
+    }
+}
+
+let foo = data::Foo {
+    a: "Hello",
+    c: 42,
+}
+```
+
+We can add a new field `b` to this `struct` with a default value, and the calling code doesn't change:
+
+```rust
+mod data {
+    pub struct Foo {
+        pub a: &'static str,
+        pub b: bool = true,
+        pub c: i32,
+    }
+}
+
+let foo = data::Foo {
+    a: "Hello",
+    c: 42,
+}
+```
+
+By using field defaults, callers can use `struct` literals without having to know about any private fields:
+
+```rust
+mod data {
+    pub struct Foo {
+        pub a: &'static str,
+        pub c: i32,
+        private_field: bool = true
+    }
+}
+
+let foo = data::Foo {
+    a: "Hello",
+    c: 42,
+}
+```
+
+## Field Privacy
+
+Default values for fields are opted into by the `struct` definition, rather than the caller initialising the structure. Field privacy doesn't need to be violated to initialise a structure.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Field defaults are limited to constant expressions. This means there are values that can't be used as defaults, so any value that requires allocation, including common collections like `Vec::new()`. It's expected that users will use a constructor function or builder for initialisers that require allocations.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Allow arbitrary expressions instead of just constant expressions
+
+Allowing arbitrary expressions as field defaults would make this feature more powerful. However, limiting field defaults to constant expressions maintains the expectation that struct literals are cheap and deterministic. The same isn't true when arbitrary expressions that could reasonably panic or block on io are allowed.
+
+For complex initialisation logic, builders are the preferred option because they don't carry this same expectation.
+
+## Allow `Default::default()` instead of just constant expressions
+
+An alternative to allowing any expression as a default value is allowing `Default::default()`, which is expected to be cheap and deterministic:
+
+```rust
+struct Foo {
+    a: &'static str,
+    b: Vec<i32> = Vec::default(),
+    c: i32,
+}
+
+let foo = Foo {
+    a: "Hello",
+    c: 42,
+}
+```
+
+It could be argued that supporting `default()` is an artificial constraint that doesn't prevent arbitrary expressions. The difference is that `Default` has an expectation of being cheap, so using it to inject logic into field initialisation is an obvious code smell.
+
+Allowing functionality to be injected into data initialisation through `default()` means struct literals may have runtime costs that aren't ever surfaced to the caller. This goes against the expectation that literal expressions have small, predictable runtime cost and are totally deterministic (and trivially predictable).
+
+## Type inference for fields with defaults
+
+Reduce the amount of code needed to define a `struct` by allowing type inference for field defaults:
+
+```rust
+struct Foo {
+    a: &'static str,
+    b = Bar,
+    c: i32,
+}
+```
+
+This has the effect of simplifying the definition, but also requiring readers to manually work out what the type of the right-hand-side is. This could be less of an issue with IDE support.
+
+## Explicit syntax for opting into field defaults
+
+Field defaults could require callers to use an opt-in syntax like `..`. This would make it clearer to callers that additional code could be run on struct initialisation, weakening arguments against more powerful default expressions. However it would prevent field default from being used to maintain backwards compatibility, and reduce overall ergonomics. If it's unclear whether or not a particular caller will use a default field value then its addition can't be treated as a non-breaking change. 
+
+This could be worked around by adding a private field to a `struct` that does nothing, forcing a caller to opt-in to potential field defaults.
+
+The goal of the design proposed in this RFC is to let users build a struct as if its default fields weren't there.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -215,63 +215,6 @@ let foo = Foo::default();
 
 Field defaults allow `#[derive(Default)]` to be used more widely because the types of fields with default values don't need to implement `Default`.
 
-## Enabling backwards compatibility
-
-With no special syntax, additional fields can be added to a struct in a non-breaking fashion. Say we have the following API and consumer:
-
-```rust
-mod data {
-    pub struct Foo {
-        pub a: &'static str,
-        pub c: i32,
-        _marker: () = ()
-    }
-}
-
-let foo = data::Foo {
-    a: "Hello",
-    c: 42,
-    ..
-}
-```
-
-Using a private marker field with a default value forces callers to opt-in to field defaults. We can now add a new field `b` to this `struct` with a default value, and the calling code doesn't change:
-
-```rust
-mod data {
-    pub struct Foo {
-        pub a: &'static str,
-        pub b: bool = true,
-        pub c: i32,
-        _marker: () = ()
-    }
-}
-
-let foo = data::Foo {
-    a: "Hello",
-    c: 42,
-    ..
-}
-```
-
-By using field defaults, callers can use `struct` literals without having to know about any private fields:
-
-```rust
-mod data {
-    pub struct Foo {
-        pub a: &'static str,
-        pub c: i32,
-        private_field: bool = true
-    }
-}
-
-let foo = data::Foo {
-    a: "Hello",
-    c: 42,
-    ..
-}
-```
-
 ## Symmetry with Destructuring
 
 Field defaults allow `struct`s to be non-exhaustively constructed and deconstructed using the same syntax:
@@ -358,6 +301,10 @@ This has the effect of simplifying the definition, but also requiring readers to
 Invoke field defaults implicitely instead of requiring a `..` in the initialiser. This would be more in line with how other languages handle default values, but is less explicit. It would also be different from pattern matching for `struct`s, that require a `..` to ignore unnamed fields.
 
 A future RFC could propose making the `..` optional for all places it's required when dealing with `struct`s.
+
+## Enhancements to functional record updates
+
+Instead of adding a new language feature, provide some alternative desugaring for functional record updates that avoids the problems outlined in [736](https://github.com/rust-lang/rfcs/blob/master/text/0736-privacy-respecting-fru.md) and still works with private fields.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-default-fields.md
+++ b/text/0000-default-fields.md
@@ -272,9 +272,25 @@ let foo = data::Foo {
 }
 ```
 
+## Symmetry with Destructuring
+
+Field defaults allow `struct`s to be non-exhaustively constructed and deconstructed using the same syntax:
+
+```rust
+// Construct
+let foo = Foo {
+    a: "Hello",
+    c: 42,
+    ..
+}
+
+// Deconstruct
+let Foo { a, c, .. } = foo;
+```
+
 ## Field Privacy
 
-Default values for fields are opted into by the `struct` definition, rather than the caller initialising the `struct`. Field privacy doesn't need to be violated to initialise a `struct`.
+Default values for fields are opted into by the `struct` definition, rather than the caller initialising the `struct`. Field privacy doesn't need to be violated to initialise a `struct` with private default values.
 
 # How We Teach This
 [how-we-teach-this]: #how-we-teach-this


killercup:
Interesting, I like it. Not something that really bothered my so far, but I can see the appeal.

You mention "constant expressions" a few times. Does mean you want to allow `const fn` as well? How much complexity do you want to allow in the default value? I think a pretty common use case might be to initialize some heap-allocated data structure like `Vec` or `HashMap`, which AFAIK is not possible in a constant expression.

For alternatives, I see there is [derive-new](https://github.com/nrc/derive-new) crate (using macros 1.1 which is currently unstable), that has [an open issue](https://github.com/nrc/derive-new/issues/2) for adding default values with attributes (I assume something like adding `#[default(42)]` to an `i32` field).

leodasvacas:
@killercup `const fn` is included. To allow an empty `Vec` we would need the alternative of allowing `Default::default`. It is not theoretically impossible for `Vec::new()` to be `const fn` since it dosen't actually allocate anything, but that is probably far from being possible in practice.

eddyb:
@leodasvacas We could make `Vec::new` `const fn` *right now* AFAIK, if we wanted to. But a `const fn` `<Vec as Default>::default` is trickier to design (you don't want a separate `const fn` `default`).

leodasvacas:
@eddyb Ah, I missed #1440. Awesome!

KodrAus:
The goal is to maintain that expectation of cheapness for fields that can be initialised without you knowing. We were talking about allowing `Default::default` too, because you expect that to be cheap, but what's proposed is the conservative option.

jFransham:
I have a problem with this not playing nice with the trait machinery, if you have a function that takes a generic `A: Default` it won't operate on a struct that has default values for all the fields, despite that meaning the same thing. It means that there are two ways to do the same thing that don't interact. I can understand that

```rust
#[derive(Default)]
struct Foo {
    a: i32,
    b: i32,
}

Foo {
    a: 10,
    ..Default::default(),
}
```

is slightly unergonomic, but it's explicit and allows a user wondering where values come from to follow the `Default::default`. If you want to only have defaults for _some_ values, you can use the builder pattern. Anonymous structs/named arguments would fix that too, since you could have a function that looks like

```rust
initialize_b_only({
    a: 10
})
```

(syntax is imaginary).

dtolnay:
Another drawback: [this parsing ambiguity](https://github.com/rust-lang/rfcs/pull/92) would show up more often.

```rust
#[derive(Eq, PartialEq)]
struct S {
    a: u8 = 1,
    b: u8 = 1,
}

let s = /* ... */;
if s == S{} {
    /* ... */
}
```

Currently empty braced structs like `struct S {}` are pretty rare, but with default struct field values if it is common for every field to have a default, this will be something everybody hits sooner or later.

KodrAus:
@jFransham I think this plays just as nicely with traits as `Clone` does. You'd have the same problem with `A: Clone` unless you derive `Clone`, even though all fields may be `Clone`.

Maybe this is a different case because you're actively _giving_ each field a default value and the result is that it can be initialised with no outside data. So it basically is `Default`. The question is whether the compiler should automatically take the step of deriving it for you if you don't.

I think an important distinction between what we've got now and what this RFC proposes is that it lets structs define their own default for types on a per-field basis. So instead of having one value for `Default` that may not be suitable for your needs (what if I don't want my `i32` to default to `0`?) you can make the default value of a field the responsibility of that field instead of the external type.

Having less boilerplate around builders would definitely make them more useful here, but are still more effort.

KodrAus:
@dtolnay Following that parsing ambiguity around lools like you're all over it? Are you concerned there are cases being missed? I'd think it would be better to nail that behaviour down deterministically and avoid the problem altogether rather than making it less common and hoping people don't do it. That's said with no appreciation of how much work might be involved though :)

text/0000-default-fields.md
```diff
@@ -0,0 +1,305 @@
...
+    }
+}
+
+let foo = data::Foo {
+    a: "Hello",
+    c: 42,
+}
+```
+
+We can add a new field `b` to this `struct` with a default value, and the calling code doesn't change:
```
petrochenkov: Unless `..` is made optional in patterns new field `b` can't be added here because it would break patterns `Foo { a, b }`.

petrochenkov: However, if structure has a special private field to force `..` in patterns - `Foo { a, b, .. /*__hidden: ()*/ }`, then it can be extended.

text/0000-default-fields.md
```diff
@@ -0,0 +1,305 @@
...
+    b = Bar,
+    c: i32,
+}
+```
+
+This has the effect of simplifying the definition, but also requiring readers to manually work out what the type of the right-hand-side is. This could be less of an issue with IDE support.
+
+## Explicit syntax for opting into field defaults
+
+Field defaults could require callers to use an opt-in syntax like `..`. This would make it clearer to callers that additional code could be run on struct initialisation, weakening arguments against more powerful default expressions. However it would prevent field default from being used to maintain backwards compatibility, and reduce overall ergonomics. If it's unclear whether or not a particular caller will use a default field value then its addition can't be treated as a non-breaking change.
```
petrochenkov: >However it would prevent field default from being used to maintain backwards compatibility

There's no backward compatibility already (https://github.com/rust-lang/rfcs/pull/1806/files#r90957633), so mirroring pattern syntax or omitting `..` is mostly a stylistic choice (unless, again, `..` is made optional in patterns).

text/0000-default-fields.md
KodrAus: My immediate thought is that `..` in patterns would also have to be optional then for consistency. I'm cautious about that though because it might increase the  scope of this thing too much.

The other option is make `..` required for structs with defaults, or stop pretending it's always backwards compatible.

dtolnay:
> @dtolnay Following that parsing ambiguity around lools like you're all over it? Are you concerned there are cases being missed? I'd think it would be better to nail that behaviour down deterministically and avoid the problem altogether rather than making it less common and hoping people don't do it. That's said with no appreciation of how much work might be involved though :)

Oh it is nailed down deterministically on the compiler side. If someone tries to write the code I wrote, they see this pretty reasonable error:

```
error[E0423]: `S` is the name of a struct or struct variant, but this expression uses it like a function name
 --> <anon>:6:9
  |
6 | if s == S{} {
  |         ^ struct called like a function
  |
  = help: did you mean to write: `S { /* fields */ }`?
```

But the fact is, almost nobody ever sees that error because almost nobody uses empty braced structs. I am pointing out that with this feature I would expect more people to stumble over this wart in Rust syntax which is a (minor) drawback.

withoutboats:
Not sure how I feel about this RFC, but syntactically I would strongly prefer some indication of the missing fields in the constructor, possibly just `..` with no expression.

KodrAus:
@withoutboats I'm not against requiring `..` to invoke field defaults. It'd work pretty much the same as FRU does now so you could add a private field with a default value to force callers to use the `..`. But I don't think it's worth peppering things with syntax if that's not solving a real problem. What kinds of problems do you see this explicitness solving?

Mark-Simulacrum:
I support adding `..` to initialization expressions that want to use default field values, since it makes it clear that there are more fields in the struct being instantiated. When reading code (especially `rustc` code, where many structs have 10+ fields) I want to know that either the initialization expression I'm seeing is complete or that I can look for X to find the other fields.

Right now, X is usually an implementation of `Default`, or another struct; both of which are easy to find. Omitting the `..` makes it more likely that someone reading the code wouldn't know if they've found the correct struct or not (especially with the re-exports prevalent in larger codebases which make tracing imports difficult).

KodrAus:
Well that would also save having to change the destructuring pattern so ok I'm sold.

withoutboats:
Yea, the goal would be to make it clear to people what is happening in the code they're reading, since they aren't necessarily familiar with the definition of the structs involved. I think this is a case where "explicit is better than implicit" is a helpful maxim.

Mark-Simulacrum:
One thing that I haven't yet seen brought up is the interaction with (potential, admittedly) function argument defaults, along with defaults in other places. Do we want to settle on `=` as the syntax for those? I think it's good; but if there's any problems with that, we should discuss them now.

withoutboats:
@Mark-Simulacrum `EXPR = EXPR` is unfortunately a valid expression which evaluates to `()`.

KodrAus:
Now something to figure out is how to explain field defaults with `..` syntax vs FRU to new users. From a caller point of view I think it could be pretty unintuitive knowing which feature they need to use and why.

keeperofdakeys:
Requiring `..` is definitely a good thing, it fits in with the more explicit nature of rust. It also improves semantics with future changes. For example, if a new field is added with a default, only code that opts-in with `..` will receive it - the rest will get compile errors.

crumblingstatue:
> The problem with a constructor is that you need one for each combination of fields a caller can supply.

I would rather like to see work on named and optional arguments, which would allow constructors to solve this problem, and make other APIs more ergonomic as well.

KodrAus:
@crumblingstatue I don't see why Rust couldn't have both default struct fields and default function arguments.

leodasvacas:
As @KodrAus already mentioned, making this explicit with `..` draws a subtlely false parallel with FRU, because this can be used to default private fields and FRU can't. The ugly part is: _explicit syntax would leak the presence of private fields to the public API_. Adding a private field to a struct that has only public fields becomes a breaking change. This is already a problem because of patterns, but we can relax pattern syntax so that users may opt-out of breakage, also struct patterns are less common than struct literals. Explicit syntax would make this a pervasive and permanent problem, by design, so it seems a lot worse than the downsides of the implicit syntax.

@crumblingstatue This is a much smaller adition to the language than named or optional parameters and yet solves some use cases that would need both of those features, so it's a win for those who want such features even if it dosen't solve all use cases.

KodrAus:
@leodasvacas I'm thinking a pragmatic way forward would be to propose this with an explicit syntax so it's in line with the current state of the world. The biggest benefit of implicit syntax kind of falls over with patterns right now, so they seem like a coupled issue. If callers want expliciteness it's going to come at the cost of incompatibility because everything is surfaced to them. A future RFC could propose allowing implicit `..` for field defaults and patterns together. Unless that's not the best aproach. I'm open to suggestions.

The similarity but difference to FRU is the problem I'm thinking about at the moment.

crumblingstatue:
@KodrAus 
> I don't see why Rust couldn't have both default struct fields and default function arguments.

@leodasvacas 
> This is a much smaller adition to the language than named or optional parameters and yet solves some use cases that would need both of those features, so it's a win for those who want such features even if it dosen't solve all use cases.

My question is this: If Rust had optional _and_ named parameters, would this feature have any use cases that aren't covered by them? If not, then this feature would just become additional bloat to the language.

And I do believe that optional and named parameters are generally more useful and ergonomic to use than default struct fields, so if both cover the same use cases, and we don't want to add features that will become obsolete, I would strongly recommend going with optional and named parameters.

crumblingstatue:
I also have concerns over optional and named arguments becoming even less of a priority if this is implemented, perhaps even rejected outright, because "technically, you can use structs with default fields in place of optional and named arguments", even if the API is much less pleasant to use that way (having additional `FooArguments` structs in place of actual function arguments).

joshtriplett:
I share the same concerns as @withoutboats here.  With this marked in the struct literal, it seems like a good idea; in the absence of such marking, though, it seems like a little too much implicit magic to me.

I do very much appreciate that this can only use compile-time constants, rather than arbitrary expressions.

bombless:
> But the fact is, almost nobody ever sees that error because almost nobody uses empty braced structs. I am pointing out that with this feature I would expect more people to stumble over this wart in Rust syntax which is a (minor) drawback.

```rust
#[derive(Eq, PartialEq)]struct A{ a: u32 }
fn main() {
    if A{ a: 1 } == A{ b: 2 } {}
}
```

It is already the case that we need parenthesis in expressions inside `if` if the expression have braces.
So it just cannot be worse.

dtolnay:
> It is already the case that we need parenthesis in expressions inside `if` if the expression have braces.
So it just cannot be worse.

Good call! Somehow I was thinking it was only empty braces that were a problem.

scottmcm:
I see two separable concerns here.

1) Struct update syntax doesn't work if the struct has private fields.

This restriction feels really weird to me, since you can do the same thing explicitly.  For example:
```
mod foo {
    #[derive(Default)]
    pub struct Bar {
        pub i : i32,
        j : i32,
    }
}

pub fn does_not_compile() -> foo::Bar {
    foo::Bar { i: 34, ..Default::default() }
}

pub fn does_compile() -> foo::Bar {
    let mut x = foo::Bar::default();
    x.i = 34;
    x
}
```

I can't see a reason that this shouldn't work, given that you can do the exact same thing less elegantly with more typing.  So I'm definitely in favour of allowing update syntax for structs with private fields.

(Perhaps the update syntax should just be defined by a desugar? Roughly `Blah { a: b, c: d, ..e }` => `{ let mut foo: Blah = e; foo.a = b; foo.c = d; foo }`.)

2) Creating a struct with a bunch of defaults is a pain

I'll throw in another option for this: What if `..` (without an expression) in the struct literal just meant `..Default::default()`?

- That meets the "there should be something in the expression denoting that more is happening" vibe I've been hearing in the thread.
- That allows punting on syntax decisions around specifying them in the struct declaration, since they wouldn't be.  (And thus avoids the "what would happen with default `fn` arguments?" discussion.)
- That avoids offering both `S{}` and `S::default()` to do the same thing in the case where all fields have default values.
- That means the default method can do non-const things, if the author believes it to be a good choice.  (Like for member containers that are known to not allocate if empty.)
- That avoids copying all the values into every call site, with the corresponding potential weirdness (see C defines, C# const vs readonly, etc) around library updates.
- That way it's just more-ergonomic syntax for something already being done, for example with [rust TargetOptions](https://github.com/rust-lang/rust/blob/ec7679b460d49a89b5d23bb04426229f4add235b/src/librustc_back/target/haiku_base.rs#L15) and [gilium DrawParameters ](https://internals.rust-lang.org/t/struct-field-defaults/3412/25)

leodasvacas:
@scottmcm FRU used to work for private fields, this was changed in RFC #736.

nrc:
> I also have concerns over optional and named arguments becoming even less of a priority if this is implemented, perhaps even rejected outright

For me, this is a step towards optional arguments and makes it more likely for us to support them. It seems likely that named arguments would be desugared to structs, and then this work gives us optional arguments 'for free'.

My concern about requiring `..` is that it significantly weakens the backwards compatibility motivation (I prefer allowing to elide `..` in patterns, perhaps only for default fields). Although the explicitness is appealing, I don't find it persuasive - one often needs to look at a stuct's definition to understand it because you need to see the types of fields. With this change, you need to look to also check on default fields. Tools (e.g., IDEs) could make this much easier by showing you the definition of the struct (e.g., on hover).

Put another way, I believe that using a structs default fields, and/or wanting backwards compatibility will be the (by far) common case, and so we shouldn't make it the less ergonomic choice.

In many cases structs are already constructed using functions (`new`, etc.) and in all these cases fields may be implicitly set, I never find myself tripped up by this. So, although programmers might need to change their expectations when seeing a struct literal, I don't think this will actually make code harder to comprehend.

Mark-Simulacrum:
I think requiring `..` isn't about not having to look at the struct definition, it's about making it obvious that there is more to the struct being declared here. For example, I want it explicit that the two struct expressions below potentially produce different results, something that is not obvious to a reader. This hurts auditability of code, since it's difficult to determine if there is another field on the struct.

```
Foo { a: 10, b: 20 }
Bar { a: 10, b: 20 }
```

The argument that IDEs can display the definition of the struct is a good one, but the reality is that not all people use an IDE, so making "this is okay, because IDEs" an argument for accepting this feature isn't a good idea.

KodrAus:
I think this question of implicit vs explicit is a wide ranging one and deserves attention across the board. I'm on board with proposing this feature with an explicit syntax so it's in line with what currently exists, with the intention of revisiting ergonomics of all the places structures are unpacked.

But I'm not convinced there's anything unsafe about fields being initialised that aren't visible in the initialiser. It's a common pattern in many programming languages so it shouldn't be surprising.

@Mark-Simulacrum In your above example I think it's already obvious that the two struct expressions produce different results because they're different types. I don't mean that to sound condescending, the apparent data equivalence doesn't matter because those types are different.

Mark-Simulacrum:
Yeah, I understand that it's obvious that data equivalence doesn't matter for different types, but to me it's not really about data equivalence so much as visible equivalence of the inside. When doing code review, I don't usually have the tooling handy to query struct definitions from, so it's helpful to be able to glance at structs without worrying about their names (which Github for example can, and often does, hide, especially for larger structs) and be able to say "they're probably the same type".

It's also confusing to me to see a struct declared and a field used on it that wasn't in anyway mentioned in the definition. That looks like an error to me; though I'm sure I'd get used to it over time.

For extensibility, the way I understand @nrc's argument is that we want it to be easy for API authors to add "default" fields even after the struct definition has been declared without any, and preserve backwards compatibility. I don't really see how this is different from the enum `__Extensible`/`__Nonexhaustive` variant pattern. I think the language should possibly add something to support that rather common pattern, but that's clearly out of scope for this RFC.

> But I'm not convinced there's anything unsafe about fields being initialised that aren't visible in the initialiser. It's a common pattern in many programming languages so it shouldn't be surprising.

Can you provide some examples? I've personally never dealt with a programming language that hides fields when using something akin to struct literal syntax (except for, perhaps, Javascript's prototypal inheritance of properties, but I don't think that's overly similar since JS isn't a compiled language where I would expect such checks. I would frown upon code doing hidden instantiations of fields there, as well).

It's also worth noting that accepting this feature with the explicit syntax doesn't prevent us from not requiring it later down the line if it's deemed reasonable (e.g. during the stabilization period). I think this is a question that really needs the experience that would give us.

withoutboats:
> In many cases structs are already constructed using functions (new, etc.) and in all these cases fields may be implicitly set, I never find myself tripped up by this. So, although **programmers might need to change their expectations when seeing a struct literal**, I don't think this will actually make code harder to comprehend.

This is the issue precisely - I want to know if any other fields are being set when I see a struct literal (I _expect_ that a constructor function sets additional fields). I consider that clarity well worth having to type `, ..` when invoking this feature.

KodrAus:
> Can you provide some examples? I've personally never dealt with a programming language that hides fields when using something akin to struct literal syntax ...

@Mark-Simulacrum C#'s classes will do that, but that is a language that's always been designed with strong IDE support in mind. 

@withoutboats can you elaborate on _why_ it's important to you to know if additional fields are being set?

joshtriplett:
It's not that it's *unsafe* to have a structure initialization mean more than what it looks like.  It's that it's confusing, and it invokes non-obvious behavior specified elsewhere.

Rust avoids that kind of "action at a distance", where a separate part of the program completely changes the interpretation of a line of code, and not knowing that causes you to interpret it incorrectly.  Rust does have features like overloaded operators, but those can't introduce an ambiguity between an obvious-but-wrong interpretation and a less obvious correct interpretation; instead, Rust has cases where you don't know, but you *know* you don't know, so you have to find out.

This change would introduce a case where you don't know, but you think you *do* know.  *That's dangerous.*

KodrAus:
@joshtriplett That sounds reasonable. So let's talk about cause and effect. You're describing a cause; interpretting a struct literal as wholy representing the data it captures, when it actually doesn't. What's an example of an effect that makes it dangerous? I can't really think of a case, and don't find the idea of a literal initialiser implicitely setting values confusing or unexpected. But again maybe that's just my C# experience and it's not actually a thing. So I'd like to understand why it's dangerous in this case.

jFransham:
@KodrAus Effect: someone fully initialises their "options" struct to make sure that everything is explicit, which is not entirely uncommon to mitigate security concerns. Library owner adds a `fire_the_missiles` option with the default `true`, when the library consumer updates the library (which, since that's not a breaking change, might just be when they delete their `Cargo.lock`) their program silently starts firing the missiles. Obviously this is silly, but as a library consumer the only way to protect against this is manually reviewing the changes of each version of the library, and pinning the library version until that review is done.

KodrAus:
@jFransham That sounds like a problem with backwards-compatible feature changes in general, and one that constructor functions could suffer from too. A change like that would still constitute at least a minor version bump, so you could avoid that by pinning a minor version of that dependency.

Having said that you do have a valid point that if you could opt-out of backwards compatibility then you wouldn't have a problem in that case. Especially since versions don't always solve the problem of versioning.

withoutboats:
I don't think "dangerous" is the right word to describe the negative consequences I think come of allowing silently initialized fields to exist. I think its more likely to be irksome than dangerous - I am imagining myself in a large, unfamiliar codebase, and seeing something like this:

```rust
let foo = Foo { bar: ..., baz: .. };
...
...
foo.quux.frobulate();
```

I expect I would likely not remember if I saw quux in the initialization, step back to the declaration of foo, realize its a default field, and then have to reach for the file where Foo is defined. It's nice to avoid this kind of frustration when we can. If I saw the `..`, I think I would be more likely to jump directly to Foo when I wanted to know what quux was.

Turning this around, the advantage of avoiding the `..` seems to me to be, in combination with the removal of the required `..` from the destructuring pattern, that fields can be added to fully public structs in a backward compatible way. Its not clear to me that that is even desirable.

It sounds like @nrc (and maybe @KodrAus as well) has a larger vision of how this will relate to optional arguments and so on. I think if I had a better understanding of how these ideas all fit in together I might have a different opinion. Not sure if I haven't read the RFC closely enough (I admit I only skimmed it several weeks ago) or if there's more context that could be provided here.

petrochenkov:
An example of code with `S { a, b, c }` pattern that *wants* to be broken when new fields are added - a visitor: https://github.com/rust-lang/rust/pull/38449/files#diff-120c6d0afb7349758f45d19ce33f2b4cR824

pnkfelix:
@leodasvacas wrote:

> making this explicit with `..` draws a subtlely false parallel with FRU, because this can be used to default private fields and FRU can't.

One way to address the above concern would be to only allow default-based initialization of fields that are visible. Another way of describing this: we could continue to restrict struct literals to expression contexts where *all* fields are visible.

For example, with this new restriction:

```rust
mod a {
    #[derive(Default)] pub struct Foo { pub m: i32, pub n: i32 = 1 }
    #[derive(Default)] pub struct Bar { pub x: i32, y: i32 = 2 }
}

fn f() {
  let foo_01: a::Foo = Default::default(); // both fine
  let bar_02: a::Bar = Default::default();

  let foo_11 = a::Foo { m: 1, .. }; // fine again

  let bar_22 = a::Bar { x: 2, .. }; // compile error: `y` is not visible.
}
```


(To be clear: This proposed alternative would *remove* one of the goals of the RFC as written. Namely it would remove the ability to use struct literal construction syntax if there are non-visible fields.)

On the plus side, the analogy of explicit `..` with FRU would be much stronger if we did adopted this alternative, since `..` *would* be sugar, expanding into all the remaining fields with their default expressions.

----

I don't personally know how much value to place on keeping the model consistent with FRU. There was a strong motivation for the limitations we put on FRU with RFC #736, but I don't think those examples apply here.

Nor do I know how much value to place on the feature of allowing struct literal construction with non-visible fields.

The current RFC text does not consider this alternative of restricting default-based initialization to visible fields. It should probably be added to the Alternatives section, because I *do* think that this RFC would still be quite useful even if it were reduced in scope in the manner I describe.

petrochenkov:
I don't think this RFC needs to be consistent with FRU, because FRU itself has suboptimal semantics.
There's a [pre-RFC](https://internals.rust-lang.org/t/proto-rfc-expanded-functional-record-update/3077) describing better semantics (`{a, ..x}` => `let tmp = x; tmp.a = a;` as opposed to current `{a, ..b}` => `{a, s.b}`), but it can't be implemented fully backward compatibly unfortunately.

KodrAus:
@withoutboats My vision is really just for all the places we assign groups of data to labels to be consistent. Which it currently is. So if we're going to use an explicit `..` for struct literals I'd find it surprising if that wasn't the same for function arguments and patterns. If it's implicit then I'd find it surprising if it was only required in just one or two places. I don't see any of these cases as truly distinct.

I've taken the feedback so far into account and updated the RFC to require an explicit `..` at the end of an initialiser for field defaults to be considered.

I've also added a basic _How We Teach This_ section, but am not sure I've got that right. It'd be great if someone could share their thoughts on that.

withoutboats:
@KodrAus there's a certain distinction between constructive patterns and destructive patterns, but don't we require the `..` in struct destructuring right now?

KodrAus:
@withoutboats yes you're right, I mentioned that in my last comment, but I think I worded it rather poorly. With the explicit `..` it's all consistent for constructive and destructive patterns. I quite like that distinction, but think the semantics of the two can be aligned in terms of how that data is packed or unpacked.

joshtriplett:
@KodrAus I like the idea of maintaining consistency in all those cases.  And thank you for the change to include the `..` to indicate defaulted initializers.

For the *How Do We Teach This* section, I'd suggest specifying exactly which part of the Rust documentation should document this feature, and sketching out what that documentation should look like.  For example, we should document this in the Rust Reference, section 6.1.5 "Structs" and section 7.2.4 "Struct expressions", with a one-sentence mention in 6.1.6 "Enumerations" that *struct-like enum variants* can have default fields just like structs.  And we should also document this in https://doc.rust-lang.org/book/structs.html .

jFransham:
I like the explicit `..`, but I can't help but wonder if there's a solution to partial initialisation that doesn't involve such a large change to the language (i.e. putting values in types, which has never been done anywhere in Rust, not to mention taking up two syntactic forms that could have been used for another feature down the line) and isn't so special-cased to a single, pretty-rare use-case. The latter is my main objection. Right now, we can just have a substruct of `Foo` called `OptionalFoo` or whatever, which implements `Default`, and we might be able to get some partial initialisation semantics out of virtual structs/fields in traits (if they ever arrive). I can't see how this significantly improves on either of those.

KodrAus:
@jFransham I think that's an interesting question :+1:

With field defaults we're allowing a caller to supply any combination of fields that offer a default value. I guess you could consider it analogous to language support for builders. What you're proposing is a different thing; a fixed collection of subsets to choose from in the form of 'substructs'. That sounds analogous to constructor functions. Which also sounds interesting, but in my mind has a weaker value proposition since it's not offering anything above what constructors already allow. Maybe if FRU allowed structural type equality over the missing fields it could interesting to use that existing feature for supported initialisation: `Foo { a: 1 ..FooDefaults { b: true ..Default::default() } }`. Or something like that.

I think this proposal is setting a precedent for filling in the gaps between what a caller supplies and what's required, and that's where it's real value is.

nrc:
I believe that the main thrust against implicit defaults is "if I assume the old semantics, then the new semantics will be surprising". Whilst true, that doesn't seem a strong criticism to me. The real argument is what the semantics of a struct literal *ought* to be.

On the other hand, I am persuaded that the back-compat motivation is not so strong, can be handled (in the future) even with explicit opt-in, and was crucially missing an opt-out. Therefore, I think that we should go the explicit route, which means we don't need to change the interpretation of completeness of struct literals, so win-win :-)

The subtle differences around private fields c.f., FRU are a little worrying, but I think the pros outweigh the cons here.

jFransham:
Can I say that changing the semantics of `Foo { a: 1, b: 2, ..somestruct }` to mean:

```rust
{
    let out = somestruct;
    out.a = 1;
    out.b = 2;
    out
}
```

Is a more usable feature than having defaults baked into the language. The fact that `Zero::zero` and `One::one` exist shows that there is more than one valid way to synthesise a value, depending on use-case. `Default::default` is only one, and granting one synthesis method more powers (i.e. the ability to modify private fields) while not granting the same to others means that you're implying that there is one valid way to synthesise values, and all the others are second-class. What if I have an options struct like:

```rust
struct Toppings {
    pub meats: Vec<Meat>,
    pub tomato_sauce: Bool,
    // We don't want this to be set by the user, we might run out of cheese!
    cheese: u16, 
}

fn the_works() -> Toppings { unimplemented!() }
fn margherita() -> Toppings { unimplemented!() }

fn main() {
    let awful_pizza = Toppings {
        tomato_sauce: False,
        ..margherita()
    };
    let nice = Toppings {
        // The works probably comes with tomato sauce but I
        // want to be really sure in case the ingredients change
        // in the future
        tomato_sauce: True,
        ..the_works()
    };
}
```

There's no "right" default that I could have for both. Having all the options set and having none of the options set are both reasonable defaults, but if we implement this I have to choose one default that's allowed to set private fields.

leodasvacas:
@jFransham What you want used to work for private fields, see why was changed in RFC #736.

petrochenkov:
@leodasvacas 
You are misunderstanding the point of RFC #736 i think. It introduces field privacy checks and explains why they are necessary for the *current* FRU desugaring, @jFransham suggests an *alternative* FRU desugaring for which field privacy checks are not needed anymore.

leodasvacas:
@petrochenkov I think the point is that copying over private fields can break invariants and should not be allowed regardless of the desugaring.

KodrAus:
I think it's reasonable to allow features utilised by the API producer to work with private fields when it's inside the existing privacy boundary. I also think it's reasonable to deny features utilised by the API consumer to work with private fields, because as @leodasvacas says they could be breaking invariants or otherwise messing with things the consumer expects to be out of reach. FRU as it stands is a tool for consumers rather than producers.

So I think @JFramsham's suggestion makes sense if the API producer was the one defining the valid substructs. Having a single substruct that contains all field defaults is functionally equivalent to this RFC, but I like the ergonomics of field defaults.

petrochenkov:
@leodasvacas 
>I think the point is that copying over private fields can break invariants and should not be allowed regardless of the desugaring.

@jFransham's desugaring doesn't "copy over private fields" (like the current desugaring does) and can't break invariants, it moves the whole value instead, this is the fundamental difference. You can't reproduce unsoundness examples from https://github.com/rust-lang/rfcs/pull/736 with it.

pnkfelix:
Indeed, I suspect @jFransham desugaring falls under ["Alternative 1" from RFC #736](https://github.com/rust-lang/rfcs/blob/master/text/0736-privacy-respecting-fru.md#alternatives), with the added bonus that its semantics are actually quite clear.

(Switching to this semantics would certainly be a breaking change, but we might get lucky; maybe not very many people are relying on the non-consuming nature of the current FRU form.)

scottmcm:
The name "Functional Record **Update**" does sound to me like it should be consuming.  And it's unusual for Rust that something (non-`Copy`) named without a borrowing sigil isn't consumed.  If nothing else, it feels odd that a language feature can't work with non-`Clone` stuff when there's a reasonable interpretation that doesn't require it.  FWIW, the two examples in the (current) book would work whether FRU was consuming or non-consuming.

Does Rust have an equivalent of C++'s Copy Elision?  It'd be nice if the compiler could replace `foo.clone()` with just `foo` when there are no further uses of `foo`.  (Or, I suppose, even omit the clone entirely if the result is unused.)  Would let a desugar (transitionally) `.clone()` without introducing cost, and would also help code that `clone`s unnecessarily, especially with `Vec`s, `Arc`s, and similar...

Ixrec:
@scottmcm Since move semantics are the default in Rust and you have to explicitly opt-in to any non-trivial copy with `.clone()`, I'm not sure what you'd need "copy elision" for. I'd actually find it more confusing if the compiler sometimes chose to ignore my explicit request for a potentially expensive copy operation (unlike in C++ where it's usually very implicit). It might be nice to have a lint for `.clone()` calls that are _provably_ unnecessary, but that probably won't catch many of the interesting mistakes like cloning Vecs when you could use more iterator adapters.

I do agree that it's weird and probably unfortunate if FRU does not consume the original value. Unfortunately it's not at all clear to me when if ever the current FRU is supposed to move or copy each field. Is there another RFC I'm unable to find that describes current FRU semantics in more detail than RFC 736?

petrochenkov:
@lxrec
>Unfortunately it's not at all clear to me when if ever the current FRU is supposed to move or copy each field.

Current FRU moves necessary fields of its operand individually, i.e. the operand is left in partially initialized state if one of the fields were non-`Copy`.
```
struct S {
    a: Vec<u8>,
    b: Vec<u8>,
}

fn main() {
    let s1 = S { a: vec![0], b: vec![1] };
    let s2 = S { a: vec![2], ..s1 }; // Moves s1.b, leaves s1.a
    s1.a; // OK
    s1.b; // ERROR use of moved value: `s1.b`
    s1; // ERROR use of partially moved value: `s1`
}
```

KodrAus:
Does anyone have any more thoughts on this? As it stands I think it's a pretty straightforward and accessible feature with some helpful use cases, but I'm also a fan of @jFransham's Super-FRUs, and would consider this a subset of that feature if it were proposed.

crumblingstatue:
I still stand by my belief that we should prioritize named and optional arguments over this.

They would cleanly satisfy all the motivation of this RFC, and would be more generally useful and flexible.

With default field values, you can only have one set of defaults. With optional and defaulted arguments, you could create multiple constructors with different defaults.

burdges:
Just curious, does anyone know why Rust writes FRUs like `Foo { a: 1, b: 2, ..somestruct }` as opposed to `somestruct { a: 1, b: 2 }` like in Haskell?  Is it simply influenced by OCaml's `{ somestruct with a=1, b=2 }` here as opposed to Haskell?  Or is there some clarity argument somewhere?

bombless:
@burdges Basically we discuss syntax from scratch, so it is not about pro or con Haskell. It was only after 1.0 that people widely consider other programming languages when we discuss syntax.

KodrAus:
The topic of [default and named arguments](https://www.reddit.com/r/rust/comments/645e8x/named_and_default_arguments/) came up on Reddit recently so I thought I'd try revive some constructive conversation around this RFC, since there isn't much movement elsewhere.

So the way I see things, this RFC isn't about providing the one and only tool for optional segments in an API, it's about providing a tool for solving a particular kind of problem; instantiating structs with a bunch of fields, where some of them have clear defaults; and, importantly,  describing a pattern for defaultness at the field-level that would translate to function arguments.

I'm in favour of named/optional function arguments, but don't see it as a competing feature, rather a complimentary one, where users have tools in appropriate places for defaultness/optionalness that are obvious and syntactically similar. Once there's an RFC for named/optional arguments we can see how it affects default struct fields.

kestred:
@KodrAus, @crumblingstatue:

With respect to named/optional function arguments, there are a few cases where they would not solve (or solve less nicely) the same problem that I'd like default arguments for:

- Nested structs with their own optional/default values
- The definition for functions which have a very large number of optional arguments may be me much easier to read if data is separated as a struct.  (Note that I completely agree there is a use for named/optional arguments that I do not have to separate out as a struct, but in different parts of the code than this use-case).
- The general distinction between ordered and named arguments (although I'm sure this will be figured out in the Default and Named Arguments RFC).

Here is an example demonstrating a use that I don't think would map well on to Named arguments.
```
use some_module::Context;

#[derive(Default)
struct SomeData {
    a: usize,
    b: usize,
    c: OtherData,
    d: OtherData,
    e: isize,
    f: f64,
}

#[derive(Default)]
struct OtherData {
    one: String,
    two: Vec<u8>,
    three: bool,
}

fn do_something(ctx: &Context, data: SomeData);

fn main() {
    let ctx = Ctx::new();
    do_something(&ctx, SomeData{a: 3, d: OtherData{three: true, ..}, ..});
}
```

-----------------------

For reference, I don't normally design interfaces like this within my own code-base, but rather I have wanted to use this pattern when writing API bindings for external interfaces which are defined in this way (such as HTTP or Python APIs).

KodrAus:
I think the main open question here now (assuming we're on board with doing _something_) is whether we want to move forward with field defaults using the `=` syntax as proposed here or take a closer look at @jFransham's enhancements to FRU, or maybe even something else.

The backwards compatibility story proposed here is obsolete because you can use [a `non_exhaustive` attribute](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md#structs-1) to achieve the same thing, but better.

I like that the `field: Type = expr` form could also be applied to function arguments too and be consistent across the board, but I also like the idea of consuming FRUs. It's a shame that it's a breaking change, and at this stage I can't see how we could manage rolling it out.

So given that I'm still on board with what's currently proposed, but the RFC needs to be updated for the current state of the world. You should all just stop writing new RFCs so my neglected prose doesn't get out of date.

petrochenkov:
@KodrAus 
There is a backward-compatible way to consume in FRUs, btw. See https://github.com/rust-lang/rfcs/pull/2008#discussion_r122558311.
I'll cite because stupid GitHub doesn't understand links to "outdated" messages.
>There's a way to add the alternative desugaring without removing the old one, inspired by evaluation order.
Imagine we changed the desugaring to `let mut s = base; s.field = field;`, but kept the syntax `S { field, ..base }`, then we are breaking the left-to-right (LTR) evaluation order! `base` is evaluated first, unlike with the old desugaring.
Then let's put it into the first position in syntax as well - `S { ..base, field }`. Voila, we added the alternative desugaring in backward compatible way and kept LTR evaluation order.

eternaleye:
@petrochenkov @KodrAus Retaining the current evaluation order with a simple desugaring doesn't require any syntax change at all; just let-bind the fields before the base, and then assign after.

```
S { field: val, ..base }
```
becomes
```
let field = val;
let mut s = base;
s.field = val;
```

Ixrec:
@KodrAus Now that I've finally caught up on this thread, my opinion is that this field defaults feature is separate from the FRU desugaring change and both are worth pursuing independently, much like how default/optional arguments are a separate and worthwhile feature. Neither field defaults or default arguments solve all the use cases on their own, because you can have a struct with no methods or a function that's not associated with any struct. Field defaults allows you to give a field a default value without having to migrate all struct initializers in your code over to static factory methods. Analogously, a "proper" default arguments feature would prevent the need to go invent a FooOptions struct as soon as you wanted fn foo() to take a single defaultable argument. They're both erognomic wins, similar to the proposed lifetime elision changes in that they prevent a conceptually small change from having a disproportionately large annotation/churn cliff, but they are targeting different cliffs.

Overall, I'm in favor of this RFC as currently written. I don't think it's as big of a win as most of the other ergonomics RFCs, so I wouldn't mind if this missed the impl period (we do have a _lot_ of other RFCs in flight right now...), but the even the subtler aspects of this RFC seem completely intuitive to me now that the "explicit" `..` is mandatory and this doesn't seem to conflict with any other desirable future features (whereas it's harder to tell how default function arguments might interact with optional/variadic arguments).

aturon:
Given that the impl period is coming quickly, I'm going to go ahead and put this up for lang team review (to determine whether we want to consider this for the impl period or punt until next year).

@rfcbot fcp merge

KodrAus:
Ok, I've removed the section on backwards compatibility and added a quick note about alternative FRU desugaring to alternatives so the text should be up to date with the state of conversation here.

aturon:
@rfcbot fcp cancel

aturon:
We discussed this RFC in the lang team meeting today, and while there was broad agreement that there's work to be done in this area, there remains significant disagreement about the route forward. For example:

- Arguably this RFC introduces a "parallel" system of defaults, alongside `Default`
- As proposed earlier in the thread, a similar goal could be accomplished by using FRU and `Default`, though at a loss of some expressiveness (since that would force you to provide a default value for *all* fields, which may not even be possible).

It seems unclear how to weigh these tradeoffs, or whether there might be a path to a fully expressive, but fully unified approach to default values. Thus, we don't feel confident enough about the direction to approve this for the impl period, and would like to consider this topic again later on.

Thanks, @KodrAus for the RFC, and sorry if I raised false hopes earlier!

@rfcbot fcp postpone

withoutboats:
I'm excited to come back to this after the impl period!

burdges:
If we had structural (sub)records, then your `=` could define a structural (sub)record, which you access with an FRU, but you can make structural (sub)records for other purposes and access those with FRUs too.  It might create a trait like : 
```
pub trait FieldDefaults {
    type FieldDefaults;  // Some structural subrecord type of Self 
    const FIELD_DEFAULTS : FieldDefaults;
}
```
And `#[derive(Default)]` could be adapted to complete the definition of `FieldDefaults` and define 
```
impl Default for S {
    fn default() -> S { FieldDefaults::FIELD_DEFAULTS }
}
```

nikomatsakis:
Just a note: A point I found convincing (probably raised earlier in the thread, sorry) that argued *against* making this sugar for `..Default::default()` was that this RFC allows one to have just *some* fields with defaults, while others remain required.

kevincox:
Yes, I think it makes sense to have this be the "primary" default technique and if you add a default for all fields you could automatically get Default (or have it derivable). That way templates and traits can use the Default trait while "average code" just doesn't need to specify these fields.

gbutler69:
Would a useful addition to this RFC (or perhaps a different RFC) be to allow fields to be declared as "initializable private" in addition to "public" and "private". "initializable private" fields could be assigned a value with an initializer expression during initialization, but, would be "private" thereafter? For example:

````rust
struct Foo {
    a: i8 = 3,
    pub b: bool = true,
    init c: i32,
    pub d: i64
}
````

The above would permit initialization values to be given for 'b', 'c' and 'd' but would only require values to be given for 'c' and 'd'. 'a' could not be assigned a value via the initializer, but, would have the default value of 3. 'c' must be given a value during initialization, but, would be fully private thereafter. Such a concept might obviate the need/desire for non-constant default parameters among other useful things.

**Comment by aturon** [comment]
Closing as postponed -- thanks @KodrAus! As mentioned above, this is definitely an area where we anticipate doing work in the future, but there are a number of thorny, interlocking design considerations.

**Comment by derekdreery** [comment]
.. a year later: I'm a big fan of this RFC. Builder pattern can be verbose, and you are relying on the compiler to make it efficient.

**Comment by Centril** [comment]
I like this RFC; I like everything about it.

- That the RHS of `=` is a const context ensures that no potentially uncontrolled global side-effects happens at a distance. This is good for the readability of code. I think it also helps maintainability because global state / side-effects makes code harder to test, refactor, and reason about.

- That `..` is used clearly informs the reader that there are more fields elsewhere that could be necessary for understanding. This reduces the reasoning footprint, improves local reasoning and thus readability and maintainability is enhanced.

- That this makes implementing `Default` manually for structs a very rare thing you have to do; that is a big win to me. In particular, seeing the default values and `#[derive(Default)]` makes for quite readable code that is easy to reason about. All you need to understand is co-located neatly.

- It clearly improves ergonomics for some of the most mundane boilerplate you have to write; line noise is thus reduced and you can focus on the important things instead of it being drowned out by unimportant things.

- Dealing with `PhantomData<T>` is likely to become more palatable as a result of this RFC. In particular, you don't have to mention `_marker: PhantomData,` anymore in struct literals.

- The proposed syntactic mechanism and interaction with `#[derive(Default)]` is natural and *not a hack*. This can easily be reproduced by other macros and the default field values can be taken advantage of.

- As @nikomatsakis noted, "this RFC allows one to have just _some_ fields with defaults, while others remain required."

All in all, there's absolutely nothing I would want to change in this proposal (except to update the wording to reflect 2018); it is perfect in my view.

Given that this RFC was postponed, I'd like to revisit it now. ...@KodrAus are you interested in pursuing this proposal further? If so, I'll reopen the PR. (some of the text needs to be updated to reflect changes in `const fn` since then...).

**Comment by H2CO3** [comment]
I oppose this RFC, for the following reasons:

1. > When deriving Default, supplied field defaults are used instead of the type default. This is a feature of `#[derive(Default)]`.

    I'm not a fan of this. It makes sense, but how much more special will the `Default` trait be as a consequence of it? Parallel to recent discussions about `From`/`Into`, the beauty of these traits was that they didn't require language or compiler support so far. (I know that `#[derive(Default)]` is in the compiler, but theoretically it could be in a library instead – it's not very highly coupled with the language, and it doesn't need to be.)

2. > ## Symmetry with Destructuring
    >
    > Field defaults allow structs to be non-exhaustively constructed and deconstructed using the same syntax:

    Why is that a symmetry? Conversely, they are doing two completely different things with the same syntax, not two aspects or "directions" of the same thing. Destructuring a `struct` and throwing away fields doesn't need extra information, there's nothing magical going on there, you just don't use certain fields.

    Construction with default values, however, hides the fact that there is extra information required to instantiate a `struct` without specifying all of its fields. `..` suggests that there's something "unimportant" or "ignorable" going on, whereas that is pretty much not the case – the user is supplying values without knowing about it. In contrast, `.. Default::default()` makes you know that _you_ are supplying the values, while still keeping it simple (and syntactically lightweight).

3. The RFC also lists the ability to construct structs with private fields using the struct literal syntax as an advantage. I disagree – to me, that's a straight-up disadvantage. The reason is this. Most if not all `struct`s with private fields have some sort of internal invariant, state, or other delicate property which must be maintained by their public API.

    Changing the internals of how these types are supposed to work might sometimes be impossible without enforcing certain initial conditions on construction. For example, one might want to perform some checks in the constructor and make a decision base don them, or use values not local to the call site (that can't be, as such, specified once in the struct definition, for all future instantiations, much like there exist expressions which just couldn't be supplied as default function arguments).

    This in turn means that a `struct` with private fields can transition from being able to be initialized using a literal, to the state of requiring a constructor function (e.g. `T::new()` and/or a manual`Default::default` impl). This means that previously non-breaking private changes (ie. changes in the implementation of an already-existing constructor function) now leak to the public interface and cause a breaking change. This seems highly undesirable.

Overall, this RFC, while seems *somewhat* useful, doesn't seem to provide enough and strong enough advantages to warrant a language change, while it is also problematic (in terms of being future-proof) in the face of structs with private fields.

I don't find the motivation very convincing either. I understand that the proposed syntax is somewhat more convenient to use than writing a builder pattern or more than one constructor, but these mainly (with one exception) concern the person implementing an API, not those consuming it. The exception is of course the "config struct in constructor parameter" pattern, `Window::new(WindowArgs{ width: 200, height: 200 })`, as mentioned in the comments. While this is indeed a few characters longer than just saying `Window { width: 200, height: 200, .. }`, I specifically find it beautiful how such a simple pattern can solve a large spectrum of problems (including even eg. optional/default function arguments) without altering the language in any way. And I think this is more important than saving a few keystrokes.

**Comment by Centril** [comment]
@H2CO3 

> I'm not a fan of this. It makes sense, but how much more special will the `Default` trait be as a consequence of it? Parallel to recent discussions about `From`/`Into`, the beauty of these traits was that they didn't require language or compiler support so far.

This makes the `Default` trait and the `#[derive(Default)]` "macro" *less*, not more, special (or at least the impact is neutral). Simply put, the trait would by no means have to be a `#[lang_item = "default"]` and the "macro" would simply use the tokens provided by the `TokenStream` to implement `Default` for the trait.

> (I know that `#[derive(Default)]` is in the compiler, but theoretically it could be in a library instead – it's not very highly coupled with the language, and it doesn't need to be.)

It couldn't because making it a library would be a breaking change; at most it can be a sort of quasi library proc macro that the compiler provides by default. In any case, if we implement this RFC, nothing changes. It can still be a library and the crate [`derivative`](https://crates.io/crates/derivative) and other procedural macros can make use of the provided defaults.

> Why is that a symmetry? Conversely, they are doing two completely different things with the same syntax, not two aspects or "directions" of the same thing. Destructuring a `struct` and throwing away fields doesn't need extra information, there's nothing magical going on there, you just don't use certain fields.

"Magic" is in the eye of the beholder. When pattern matching a tuple with `..`, i.e. `(a, b, ..)` you are essentially having row polymorphism in the type system to deal with the pattern typing. That seems magical to me. But magic isn't always a bad thing.

I also feel that there's a *sort of* duality between pattern matching and ignoring the extra fields and constructing and providing extra fields. I don't think the usual "pattern matching follows construction" is violated here.

There are even nice laws here:

```rust
let x = Foo { alpha, beta, .. };
let Foo { alpha, beta, .. } = x;
let y = Foo { alpha, beta, .. };
assert_eq!(x, y);
```

> Construction with default values, however, hides the fact that there is extra information required to instantiate a `struct` without specifying all of its fields. `..` suggests that there's something "unimportant" or "ignorable" going on, whereas that is pretty much not the case – the user is supplying values without knowing about it. In contrast, `.. Default::default()` makes you know that _you_ are supplying the values, while still keeping it simple (and syntactically lightweight).

I think *there is* something mostly "unimportant" and "ignorable" going on, otherwise I wouldn't use default values on fields. I also don't think `.. Default::default()` is syntactically light weight (relatively speaking) and it may also have side-effects... With `..` I both know that *semantic* defaults (instead of merely *structural* defaults, which is what you ordinarily get with `#[derive(Default)]` otherwise...) are being used and that no side effects are happening behind my back.

I think that if defaults were being used when writing `Foo { bar, baz }` you would have a good case here and I would tend to agree, but in this case you have to write `Foo { bar, baz, .. }` so nothing is happening behind your back. You opted into it.

> 3. The RFC also lists the ability to construct structs with private fields using the struct literal syntax as an advantage. I disagree – to me, that's a straight-up disadvantage. The reason is this. Most if not all `struct`s with private fields have some sort of internal invariant, state, or other delicate property which must be maintained by their public API.

Sure, sometimes they do; however, even with `..` being able to use the `def_site`'s visibility instead of the `call_site`'s visibility, you will likely still have other fields that don't have default and so you still cannot use `Foo { .. }` to construct a value of the type that may break invariants. For example:

```rust
pub struct Unique<T: ?Sized> {
    pointer: NonZero<*const T>,
    _marker: PhantomData<T> = PhantomData,
}

Unique { .. }  // Error! `pointer` is missing.
Unique { pointer: <value>, .. }  // Still, error! `pointer` isn't visible...
```

> Changing the internals of how these types are supposed to work might sometimes be impossible without enforcing certain initial conditions on construction. For example, one might want to perform some checks in the constructor and make a decision base don them, or use values not local to the call site (that can't be, as such, specified once in the struct definition, for all future instantiations, much like there exist expressions which just couldn't be supplied as default function arguments).
> This in turn means that a `struct` with private fields can transition from being able to be initialized using a literal, to the state of requiring a constructor function (e.g. `T::new()` and/or a manual`Default::default` impl). This means that previously non-breaking private changes (ie. changes in the implementation of an already-existing constructor function) now leak to the public interface and cause a breaking change. This seems highly undesirable.

It is *always* possible to add a private field typed at some unit type; now the type cannot be constructed with `Foo { .. }`.

> I don't find the motivation very convincing either. I understand that the proposed syntax is somewhat more convenient to use than writing a builder pattern or more than one constructor, but these mainly (with one exception) concern the person implementing an API, not those consuming it.

I think the language (+ team) should be concerned with making libraries (and libraries within applications) convenient to write.

H2CO3:
> It couldn't because making it a library would be a breaking change

Of course not now – but it could have been, that's what I meant.

> There are even nice laws here:

I see that, but that's mostly ignoring the semantics. (Also just because it's mathematically appealing, it can have issues in practice, which I think this construct definitely has, as I mentioned before.)

> I also don't think `.. Default::default()` is syntactically light weight

Sorry, then I think we are just disagreeing here.

> and it may also have side-effects

So could this RFC with two of the proposed semantics. Only one of them prevents side effects; the others (allowing arbitrary expressions or only `Default::default()`) could also do everything nasty one could do within `default`.

> It is always possible to add a private field typed at some unit type; now the type cannot be constructed with Foo { .. }.

Yes (but I doubt many users would know this or do it by default out of caution…), and then this part of the RFC completely loses its use anyway. By the way, this would now be yet another thing that programmers implementing an API would need to remember – "always add a mostly meaningless ZST private field by default, if you want to **prevent** construction via a literal". I think this is the wrong way around, as it makes the potentially dangerous situation the default, requiring one to opt out of it (and not even in a way that might be obvious to people).

> I think the language (+ team) should be concerned with making libraries (and libraries within applications) convenient to write.

Definitely. I still don't think this particular RFC is justified. It's not like structs are *hard* or *inconvenient* to use today.

Centril:
@H2CO3 

>  > It couldn't because making it a library would be a breaking change
>
> Of course not now – but it could have been, that's what I meant.

But what value does the argument have if it's not actionable?
Furthermore, as I pointed out, it could still have been a library macro if we had made this change pre 1.0.

> > There are even nice laws here:
> 
> I see that, but that's mostly ignoring the semantics. (Also just because it's mathematically appealing, it can have issues in practice, which I think this construct definitely has, as I mentioned before.)

Yeah sure, just because there are mathematical laws doesn't mean it cannot have other problems, but then those need to be demonstrated (and I don't think sufficient problems have been...).
But mathematical beauty is a plus in my view. It also hints at a construct that is easy to understand.

> > and it may also have side-effects
> 
> So could this RFC with two of the proposed semantics. Only one of them prevents side effects; the others (allowing arbitrary expressions or only `Default::default()`) could also do everything nasty one could do within `default`.

I think we should discuss what the RFC actually proposes, and it proposes that the default value of a `struct` is a [`const` context](https://github.com/rust-lang-nursery/reference/blob/master/src/const_eval.md) and thus the expression must be a constant expression. Those may not (at least not currently and that is very unlikely to change) have side effects.

> > It is always possible to add a private field typed at some unit type; now the type cannot be constructed with Foo { .. }.
> 
> Yes (but I doubt many users would know this or do it by default out of caution…), and then this part of the RFC completely loses its use anyway. By the way, this would now be yet another thing that programmers implementing an API would need to remember – "always add a mostly meaningless ZST private field by default, if you want to **prevent** construction via a literal". I think this is the wrong way around, as it makes the potentially dangerous situation the default, requiring one to opt out of it (and not even in a way that might be obvious to people).

I disagree that what you need to remember is "always add a mostly meaningless ZST"... I think almost always the other fields that don't have defaults are private (and then you wouldn't `#[derive(Default)]` anyways) so in that case it isn't possible possible to construct it via a literal anyways.

The language is set up in such a way that the default thing is to respect the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) since fields are private by default. If your API is dangerous in some way such as if it involves `PhantomData<T>` + `unsafe`, then you are unlikely to have public fields. An example of where it could be dangerous is:

```rust
struct Id<A, B> {
    prf: PhantomData<(fn(A) -> A, fn(B) -> B)> = PhantomData,
}

impl<S: ?Sized, T: ?Sized> Id<S, T> {
    pub fn cast(self, value: S) -> T where S: Sized, T: Sized {
        unsafe {
            let cast_value = mem::transmute_copy(&value);
            mem::forget(value);
            cast_value
        }
    }
}

let refl: Id<u8, Box<u8>> = Id { .. }; // And we have bricked the type system.
```

but in this case you used `unsafe` so you should know what you are doing in any case.
I think these scenarios are unlikely to arise and I would make this trade-off personally.

H2CO3:
> But what value does the argument have if it's not actionable?

The point was that `#[derive(Default)]` doesn't need to be coupled with the compiler in principle.

> I pointed out, it could still have been a library macro if we had made this change pre 1.0.

This is exactly what I'm trying to say (assuming by "it" you are also referring to `#[derive(Default)]`).

Centril:
@H2CO3 

> > But what value does the argument have if it's not actionable?
> 
> The point was that `#[derive(Default)]` doesn't need to be coupled with the compiler in principle.

Right, I think that's a worthwhile goal. But do you agree that this RFC doesn't change anything in this respect -- so *the principle is retained*?

With respect to how the principle is even *enhanced*, consider for example [`#[derive(Arbitrary)]`](https://github.com/AltSysrq/proptest/issues/102) for `proptest`, which we are working on shipping. It could make good use of the default values provided by the user. Ostensibly `serde` could do the same (ofc that is up to the authors of `serde`). For example:

```rust
#[derive(Debug, Arbitrary, Deserialize)]
struct Foo {
    // The proptest_derive macro will see this
    // and fix the value of `bar` to always be 42.

   // Serde will see this and assume that if no value is
   // provided then `42` will be used on deserialization.
    bar: u8 = 42,
    // other fields...
}
```

H2CO3:
> But do you agree that this RFC doesn't change anything in this respect -- so the principle is retained

I was asking if it does - and if it doesn't, I do agree it is retained.

burdges:
I'll tentatively disagree with the constructor `..` here being "dual" to the pattern `..`, but that's easy to fix by using say `...` here.  In fact I'll use `...` below because even discussing the feature as `..` gets confusing.

If I understand, the `...` proposed here ensures that values are instantiated by constants explicitly specified in the struct definition, while `..Default::default()` executes arbitrary code distributed through arbitrarily complex type hierarchies.  In particular, we cannot expand `...` to instantiate a `HashMap` because doing so requires sampling random numbers.  

I'd agree with @H2CO3 opposition to this feature if `...` were merely a synonym for `..Default::default()`, but..  I believe the restriction to constants explicitly specified in the struct definition makes this feature favor code readers, auditors, etc. over code writers. 

I also think `...` permitting only a partial default helps enormously, although conceivably structural records with magical cross-struct FRU and some magical traits might enable this: 
```
trait PartialDefault {
    type Part : Substruct<Self>;
    const fn partial_default() -> Part;
}
```
And this RFC's ergonomics can be recaptured with polymorphic constants:
```
const DEFAULT<T: PartialDefault> : <T as PartialDefault>::Part = <T as PartialDefault>::partial_default();
```
We could then simple write `{ foo , ..DEFAULT }` instead of the `{ foo , ... }` proposed here, but we've used similar magic to this RFC albeit general purpose.

crumblingstatue:
My main problem with this feature is that it's a less generic solution than what could be achieved with constructor functions if Rust had named and optional args.

Quoting from the motivation:
> The problem with a constructor is that you need one for each combination of fields a caller can supply. To work around this, you can use builders, like process::Command in the standard library. Builders enable more advanced initialisation, but need additional boilerplate.

This problem would be nullified if Rust had optional and named arguments.

## Reasons why it's less generic than constructor functions

1. Doesn't allow multiple sets of defaults
You can have as many constructor functions as you want, and each can have a different set of defaults.
With default struct field values, you can only define one set of defaults that are the defaults in all usage contexts.

2. Doesn't allow non-constant expressions
    As stated in the drawbacks:
    >  Field defaults are limited to constant expressions. This means there are values that can't be used as defaults, so any value that requires allocation, including common collections like Vec::new(). It's expected that users will use a constructor function or builder for initialisers that require allocations.

    If constructor functions had optional/named arguments, they could be used in these cases with the same amount of convenience as default struct field values.

3. Not applicable if you want to check for invariants at construction time
For example, a `Window` could have a `width` field that is `u32`, but the actual width can't be larger than `1000000`. Or you could have a struct with `min` and `max` fields, and `min` can't be more than `max`. A constructor function can easily check for these invariants.

I don't fundamentally oppose this feature, but I believe optional and default arguments would cover all the motivation for this feature, without the constant-only drawback, while also fulfilling other motivations.

As a closing statement, I want to acknowledge that optional/default arguments are indeed controversial, and they may take years to land, if ever. I just wanted to make sure this information is out there for people to consider.

burdges:
I believe structural records #2584 along with this feature provide precisely named and optional args since you could write:
```
fn foo(args: {a : bool = true, b: bool = false} ) { .... }
foo({ a: false, ...});
```

crumblingstatue:
> I believe structural records #2584 along with this feature provide precisely named and optional args since you could write:
> 
> ```
> fn foo(args: {a : bool = true, b: bool = false} ) { .... }
> foo({ a: false, ...});
> ```

Making sure these two features work together to allow this use case is definitely worth looking into.

jplatte:
This seems like a pretty useful feature, but I don't understand the rationale for allowing the construction of structs that contain invisible fields with it, especially given that FRU isn't supported for structs with invisible fields.
Allowing this even has an additional fallback over allowing it for FRU: It makes it easier to accidentally limit the evolution of the implementation of a struct by specifying defaults for all private fields. If one does that without considering this feature, adding or modifying fields in a way that they can't have defaults becomes a breaking change.

derekdreery:
@jplatte Maybe this feature could be combined with a non-exhaustive flag, so

```rust
#[non_exhaustive]
struct Options {
   first: bool,
}
```

would not allow instantiating the struct, similar to if you did

```rust
struct Options {
  first: bool,
  #[doc(hidden)]
  __private: ()
}
```

and 

```rust
#[non_exhaustive]
enum Example {
  Variant1,
  Variant2,
}
```

is equivalent to

```rust
enum Example {
  Variant1,
  Variant2,
  #[doc(hidden)]
  __NonExhaustive,
}
```

This has been discussed before I think - there was talk of syntax like

```rust
enum Example {
  Variant1,
  ..
}
```

but I can't remember the details of the conversation.

sbrl:
Newbie Rust programmer here. Just tried to use default values for struct fields, and discovered it doesn't work. Now I'm a bit lost. What's the recommended alternative to default struct values if my struct has lots of fields? A constructor would have lots of arguments and look messy.

Lokathor:
Uh, this isn't really a Q and A place, usually you want to try reddit or the discourse forums or the discord servers for rust.

This RFC was not merged, so it's not an accurate description of how the language works. this repo is for language change proposals.

scottmcm:
@dhardy Cool to have that in a crate!

I'd rather not do that plan in the language, though, since it'd make it harder to make those `= expr`s actually do something outside of macros in the future.  I'd rather just give them meaning at the same time as making them parse.

burdges:
It's maybe handy if `Default::default` could be declared `const fn` ala https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4dc60134a2c8a128974dc9f307e9ae32 even though such code could not be generic.

PR #1838, opened 2016-12-30
Remove last core-team reference from 1636

chriskrycho:
I thought I'd caught them all, but apparently NOPE.

This fix is for you, @brson. 😀


aturon:
Thanks!

PR #1860, opened 2017-01-20
Manually Drop

nagisa:
[Rendered](https://github.com/nagisa/rfcs/blob/manually-drop/text/0000-manually-drop.md)

### DIFF
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -112,8 +112,8 @@ impl Drop for FruitBox {
             // Explicit ordering in which field destructors are run specified in the intuitive
             // location – the destructor of the structure containing the fields.
             // Moreover, one can now reorder fields within the struct however much they want.
-            peach.manually_drop();
-            banana.manually_drop();
+            peach.drop();
+            banana.drop();
         }
         // After destructor for `FruitBox` runs (this function), the destructor for Melon gets
         // invoked in the usual manner, as it is not wrapped in `ManuallyDrop`.
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -105,12 +105,13 @@ impl<T> ManuallyDrop<T> {
 // Other common impls such as `Debug for T: Debug`.
 ```
 
-Let us apply this union to the example from the motivation:
+Let us apply this union to a somewhat expanded example from the motivation:
 
 ```rust
 struct FruitBox {
     // Immediately clear there’s something non-trivial going on with these fields.
     peach: ManuallyDrop<Peach>,
+    melon: Melon, // Field that’s independent of the other two.
     banana: ManuallyDrop<Banana>,
 }
 
@@ -123,6 +124,8 @@ impl Drop for FruitBox {
             peach.manually_drop();
             banana.manually_drop();
         }
+        // After destructor for `FruitBox` runs (this function), the destructor for Melon gets
+        // invoked in the usual manner, as it is not wrapped in `ManuallyDrop`.
     }
 }
 ```
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -17,7 +17,7 @@ exposed the fact that the current, but unspecified behaviour is relied onto for 
 that there’s at least a few instances of such code in the wild.
 
 While a move to stabilise and document the order of destructor evaluation would technically fix the
-problem describe above, there’s another important aspect to consider here – implicitness. Consider
+problem described above, there’s another important aspect to consider here – implicitness. Consider
 such code:
 
 ```rust
@@ -47,7 +47,7 @@ annotate the dependencies somehow.
 [design]: #detailed-design
 
 This RFC proposes adding following `union` to the `core::mem` (and by extension the `std::mem`)
-module. `mem` module is a most suitable place for such `struct`, as the module already a place for
+module. `mem` module is a most suitable place for such type, as the module already a place for
 functions very similar in purpose: `drop` and `forget`.
 
 ```rust
@@ -130,7 +130,7 @@ impl Drop for FruitBox {
 }
 ```
 
-It is proposed that such pattern would become idiomatic for structures where fields must be dropped
+It is proposed that this pattern would become idiomatic for structures where fields must be dropped
 in a particular order.
 
 # How We Teach This
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -71,24 +71,6 @@ impl<T> ManuallyDrop<T> {
         }
     }
 
-    /// Extracts the value from the ManuallyDrop container.
-    /// Could also be implemented as Deref.
-    #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
-    pub fn as_ref(&self) -> &T {
-        unsafe {
-            &self.value
-        }
-    }
-
-    /// Extracts the value from the ManuallyDrop container.
-    /// Could also be implemented as a DerefMut.
-    #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
-    pub fn as_mut(&mut self) -> &mut T {
-        unsafe {
-            &mut self.value
-        }
-    }
-
     /// Manually drops the contained value.
     ///
     /// # Unsafety
@@ -97,9 +79,18 @@ impl<T> ManuallyDrop<T> {
     /// with the value within invalid. The fact that this function does not consume the wrapper
     /// does not statically prevent further reuse.
     #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
-    pub unsafe fn manually_drop(&mut self) {
-        ptr::drop_in_place(&mut self.value)
+    pub unsafe fn drop(slot: &mut ManuallyDrop<T>) {
+        ptr::drop_in_place(&mut slot.value)
     }
+}
+
+impl<T> Deref for ManuallyDrop<T> {
+    type Target = T;
+    // ...
+}
+
+impl<T> DerefMut for ManuallyDrop<T> {
+    // ...
 }
 
 // Other common impls such as `Debug for T: Debug`.
@@ -163,4 +154,4 @@ glance;
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-* Best way to expose value inside the wrapper.
+None known.
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -46,14 +46,14 @@ annotate the dependencies somehow.
 # Detailed design
 [design]: #detailed-design
 
-This RFC proposes adding following `struct` to the `core::mem` (and by extension the `std::mem`)
+This RFC proposes adding following `union` to the `core::mem` (and by extension the `std::mem`)
 module. `mem` module is a most suitable place for such `struct`, as the module already a place for
 functions very similar in purpose: `drop` and `forget`.
 
 ```rust
 /// Inhibits compiler from automatically calling `T`’s destructor.
 #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
-#[allow(missing_debug_implementations, unions_with_drop_fields)]
+#[allow(unions_with_drop_fields)]
 pub union ManuallyDrop<T>{ value: T }
 
 impl<T> ManuallyDrop<T> {
@@ -101,6 +101,8 @@ impl<T> ManuallyDrop<T> {
         ptr::drop_in_place(&mut self.value)
     }
 }
+
+// Other common impls such as `Debug for T: Debug`.
 ```
 
 Let us apply this structure to the example from the motivation:
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -105,7 +105,7 @@ impl<T> ManuallyDrop<T> {
 // Other common impls such as `Debug for T: Debug`.
 ```
 
-Let us apply this structure to the example from the motivation:
+Let us apply this union to the example from the motivation:
 
 ```rust
 struct FruitBox {
@@ -127,8 +127,8 @@ impl Drop for FruitBox {
 }
 ```
 
-It is proposed that such code would become idiomatic for structures where fields must be dropped in
-a particular order.
+It is proposed that such pattern would become idiomatic for structures where fields must be dropped
+in a particular order.
 
 # How We Teach This
 [how-we-teach-this]: #how-we-teach-this
@@ -136,9 +136,9 @@ a particular order.
 It is expected that the functions and wrapper added as a result of this RFC would be seldom
 necessary.
 
-In addition to the usual API documentation, this structure should be mentioned in
-reference/nomicon/elsewhere where drop ordering guarantees are described as a way to explicitly
-control the order in which the structure fields gets dropped.
+In addition to the usual API documentation, `ManuallyDrop` should be mentioned in
+reference/nomicon/elsewhere as the solution to the desire of explicit control of the order in which
+the structure fields gets dropped.
 
 <!--
 # Drawbacks
--- a/text/0000-manually-drop.md
+++ b/text/0000-manually-drop.md
@@ -1,0 +1,161 @@
+- Feature Name: manually_drop
+- Start Date: 2017-01-20
+- RFC PR: (leave this empty)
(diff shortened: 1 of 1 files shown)


KalitaAlexey:
@nagisa,
I like the idea.
I see that you don't propose to make the compiler emit an error if a struct has a manually dropped field but doesn't impl Drop.
I think that the compiler must emit an error.

KalitaAlexey:
@nagisa,
You don't describe the situation where a struct has as a manually dropped field as a common one.

text/0000-manually-drop.md
```diff
@@ -0,0 +1,161 @@
...
+
+This RFC proposes adding following `struct` to the `core::mem` (and by extension the `std::mem`)
+module. `mem` module is a most suitable place for such `struct`, as the module already a place for
+functions very similar in purpose: `drop` and `forget`.
+
+```rust
+/// Inhibits compiler from automatically calling `T`’s destructor.
+#[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
+#[allow(missing_debug_implementations, unions_with_drop_fields)]
+pub union ManuallyDrop<T>{ value: T }
```
KalitaAlexey: You write **adding following `struct`**, but it is a union.

text/0000-manually-drop.md
```diff
@@ -0,0 +1,161 @@
...
+[design]: #detailed-design
+
+This RFC proposes adding following `struct` to the `core::mem` (and by extension the `std::mem`)
+module. `mem` module is a most suitable place for such `struct`, as the module already a place for
+functions very similar in purpose: `drop` and `forget`.
+
+```rust
+/// Inhibits compiler from automatically calling `T`’s destructor.
+#[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
+#[allow(missing_debug_implementations, unions_with_drop_fields)]
```
KalitaAlexey: I think it should have `impl Debug` if `T: Debug`.

text/0000-manually-drop.md
```diff
@@ -0,0 +1,163 @@
...
+    #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
+    pub unsafe fn manually_drop(&mut self) {
+        ptr::drop_in_place(&mut self.value)
+    }
+}
+
+// Other common impls such as `Debug for T: Debug`.
+```
+
+Let us apply this structure to the example from the motivation:
```
KalitaAlexey: Replace **this structure** with **this union**

nagisa:
>  I see that you don't propose to make the compiler emit an error if a struct has a manually dropped field but doesn't impl Drop. I think that the compiler must emit an error.

`ManuallyDrop` may also be used as a method to *suppress* the destructors, similar to the functionality provided by the [`nodrop`](https://crates.io/crates/nodrop) crate. Automatic `mem::forget`. In that sense I disagree with compiler having to emit an error.

> You don't describe the situation where a struct has as a manually dropped field as a common one.

What do you mean? I’d expect `ManuallyDrop` to be used quite rarely (for its intended purpose, anyway). I even said this in the “How we teach this” section:

> It is expected that the functions and wrapper added as a result of this RFC would be seldom necessary.

That is, I believe that large majority of code does not care about the drop order.

KalitaAlexey:
> ManuallyDrop may also be used as a method to suppress the destructors, similar to the functionality provided by the nodrop crate. Automatic mem::forget. In that sense I disagree with compiler having to emit an error.

The one who needs to suppress the destructor can add `#[allow]` to the field.
It is better than people will get annoyed because of leaks.

>What do you mean? I’d expect ManuallyDrop to be used quite rarely (for its intended purpose, anyway). I even said this in the “How we teach this” section:

```rust
struct FruitBox { 
    peach: ManuallyDrop<Peach>, 
    banana: ManuallyDrop<Banana>,
    other_fruit: OtherFruit,
} 
```

text/0000-manually-drop.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+For structures with dependencies between fields it is worthwhile to have ability to explicitly
+annotate the dependencies somehow.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposes adding following `struct` to the `core::mem` (and by extension the `std::mem`)
+module. `mem` module is a most suitable place for such `struct`, as the module already a place for
```
joshtriplett: It seems odd to call this a `struct`.  Perhaps "type" instead?

Ericson2314:
I feel like `&move` drop is much more elegant?

KalitaAlexey:
@Ericson2314,
What are you talking about?

Ericson2314:
@KalitaAlexey Sorry, it's in a bunch of old RFCs scattered around.

Ericson2314:
At random, there's newest https://github.com/rust-lang/rfcs/pull/1646 and oldest https://internals.rust-lang.org/t/pre-rfc-allow-by-value-drop/1845

text/0000-manually-drop.md
```diff
@@ -127,18 +127,18 @@ impl Drop for FruitBox {
 }
 ```
 
-It is proposed that such code would become idiomatic for structures where fields must be dropped in
-a particular order.
+It is proposed that such pattern would become idiomatic for structures where fields must be dropped
```
jmesmon: Perhaps: "this pattern" instead of "such pattern"? Alternates would be "such patterns" and "such a pattern", but "this" seems more natural here.

nagisa:
> I feel like &move drop is far more elegant? Could we just do this on an experimental/temporary basis?

`&move` is not possible to produce from a `&mut`, and we cannot change the signature of `Drop::drop`. If we had `&move` already, I’d be interested in discussing a `MoveDrop::move_drop(&move self)`, but it does not exist yet. Stabilising `union`s and having something like this proposal as a crate also sounds okay to me, although it is awkward to recommend people to use a crate for what I propose to make the idiomatic (while we haven’t `&move`, at least) solution to the problem.

joshtriplett:
Based on discussion in the lang team meeting: this is a libs RFC, not a lang RFC.  Labeling accordingly.

nagisa:
I feel that its at least to some extent related to `T-lang`, even if the only relation is about the intent to consider `ManuallyDrop` the idiomatic approach.

alexcrichton:
I wonder, was perhaps a `Deref` and `DerefMut` implementation considered for this type? This seems like a swell type to have in libstd, and with deref we'd trim the API down to:

```rust
impl<T> ManuallyDrop<T> {
    pub fn new(T) -> Self;
    pub fn into_inner(self) -> T;
    pub unsafe fn manually_drop(slot: &mut ManuallyDrop<T>); // not inherent to avoid conflicts
}

impl<T> Deref for ManuallyDrop<T> {
    type Target = T;
}

impl<T> DerefMut for ManuallyDrop<T> {}
```

SimonSapin:
```rust
    pub fn into_inner(self) -> T;
    pub unsafe fn manually_drop(slot: &mut ManuallyDrop<T>); // not inherent to avoid conflicts
```

`pub fn into_inner(ManuallyDrop<T>) -> T` for the same reason?

alexcrichton:
@SimonSapin yeah I was initially going to recommend that, but I hesitated because almost all `into_*` methods consume `self`, so if you're reaching them through `Deref` you're by definition unable to call them. In that sense the reason we avoid inherent methods on `Deref` types, to avoid method conflicts, may not actually apply to the `into_*` terminology.

Although with that in mind it may be too clever, we have `Box::into_raw`, so it may be not worth it. Either way I'm fine.

SimonSapin:
> if you're reaching them through Deref you're by definition unable to call them

Ah, good point.

nagisa:
@alexcrichton 

> I wonder, was perhaps a `Deref` and `DerefMut` implementation considered for this type?

Yes, of course. I’m torn between just providing these impls and not providing them. `Deref` and `DerefMut` work pretty auto-magically a lot of the time, and I’m worried about it not being obvious enough where a mistake in

```
let mut x = ManuallyDrop::new(something);
unsafe { x.manually_drop(); }
x.some_method()
```

is. On the other hand, `Deref` and `DerefMut` make the type much more ergonomic.

@SimonSapin 

> not inherent to avoid conflicts

That’s something to consider, but I’m not too worried about conflicts. It seems to me like `manually_drop` is so unique a name that it just doesn’t matter either way.

sfackler:
I personally see
```rust
let mut x = ManuallyDrop::new(something);
unsafe { x.manually_drop(); }
x.some_method()
```
as pretty equivalent to
```rust
let mut x = ManuallyDrop::new(something);
unsafe { x.manually_drop(); }
x.as_ref().some_method()
```
as far as obviousness that you're doing something wrong.

I would like to avoid non-static methods if we are going to go the `Deref` route, but `ManuallyDrop::manually_drop(&mut x)` seems pretty gross. We could just call it `drop` I guess?

alexcrichton:
I think I agree with @sfackler where I'd consider those two snippets equivalent in terms of "ease of spotting the bug", so I think I'd personally at least prefer Deref/DerefMut, and I like @sfackler's idea of calling the method `drop`

nagisa:
I’m convinced :) I hadn't a strong opinion either way, so I pushed the `Deref` version.

burdges:
If I understand `manually_drop` is meant to be called only from `Drop::drop`, right?  If so then is there any safety to be gained by keeping the mutable borrow about `manually_drop` finishes?
```
struct ManuallyDropped<'a,T>(PhantomData<&'a mut T>);

impl<T> ManuallyDrop<T> {
    ...
    pub unsafe fn manually_drop<'a>(slot: &'a mut ManuallyDrop<T>) -> ManuallyDropped<'a,T>;
```
I donno if the `unsafe` makes this useless, but it's certainly does nothing if you call `manually_drop` from a function besides `Drop::drop` that was passed the `&mut T`. 

In fact, if the `manually_drop` was somehow restricted to only be called from `Drop::drop` then perhaps it could act as a a borrow to a top level value in `Drop::drop`.  Would that be enough to make `manually_drop` safe?

sfackler:
That doesn't change behavior at all if you don't bind the result.

alexcrichton:
This looks great to me, has sound motivation, and even if `union` is stabilized seems like a great abstraction to have in the standard library. Thanks for the rfc @nagisa! In light of this I'd like to propose that we merge this RFC:

@rfcbot fcp merge

burdges:
Right of course @sfackler   I guess you'd need the compiler to auto-magic bound the result to a new top level value in `Drop::drop` for that to do anything.

joshtriplett:
@burdges @sfackler You could mark it as `#[must_use]`, which would force binding it.  However, that doesn't actually fix the safety issue, just make a few errors somewhat less likely.

joshtriplett:
As far as I can tell, if not for `Drop::drop`, we'd make this method take and consume `self` rather than taking `&mut self`; doing so would eliminate the safety considerations.  Does it seem worthwhile to provide that safe function, in addition to the unsafe version?

For that matter, could we add some kind of compiler-level support to help detect use-after-drop cases?

(As an aside, why *doesn't* `Drop::drop` take and consume `self` rather than taking `&mut self`?)

sfackler:
@joshtriplett if `drop` took `self`, it'd have to magically disable drops from running on `self` if it was passed by-value into another function to avoid recursing.

alexcrichton:
Isn't the whole point of this type that drop *doesn't* consume self? If that were true then this type wouldn't be needed at all.

SimonSapin:
> if the manually_drop was somehow restricted to only be called from Drop::drop

That sounds undesirable. Sometimes part of a `Drop::drop` impl is moved to a separate function. Either as a refactoring to manage complexity for human readers/maintainers, or to guide inlining for the compiler.

burdges:
Good point @SimonSapin   You'd need some attribute that forced returning the `PhantomData<&mut T>`, or whatever held the lifetime, all the way back to `Drop::drop`, which sounds extremely annoying.

scottmcm:
Is there any way the compiler could complain about this, or that it could be made to work?
```
    let x = ManuallyDrop::new(something);
    // do some stuff
    std::mem::drop(x);
```
That hopefully won't get written, but it _looks_ like it's doing the right thing, so I'd probably miss it in CR.


Random thought: Should `E0040` mention this?  Maybe a line at the end along the lines of "If you do need to control drops manually, such as to enforce struct drop order, consider `ManuallyDrop`?"


As a bonus, I think this RFC obviates `intrinsics::forget`, as `std::mem::forget` can now be written:
```
pub fn forget<T>(t: T) {
    let _ = ManuallyDrop::new(t);
}
```
That does make me wonder: should the type be `#[must_use]`?  (I can't imagine it'll be returned from methods very often, if ever, but if it were, not using it is probably wrong.)

Ericson2314:
> `&move` is not possible to produce from a `&mut`, and we cannot change the signature of Drop::drop.

These were never interested. It's easy to write a blanket impl from Old Drop to new Drop.

> If we had &move already, I’d be interested in discussing a `MoveDrop::move_drop(&move self)`, but it does not exist yet.

I think the reason we *don't* have `&move` is because people were insufficiently peeved with the problem this RFC duct-tapes. `&move`-Drop is, I'll wager, the only safe solution to the drop-from-drop instance problem. I was completely OK with the reasoning in https://github.com/rust-lang/rfcs/pull/1646#issuecomment-27912302, but if this RFC is going to be accepted perhaps we should revisit some of the thoughts there.

Taking a step back, I'm fine with this as a stop-gap, especially if it remains unstable until further thinking about `&move`. I hope it *increases* the pressure for the more elegant and safe solution, but I fear it will decrease it instead.

alexcrichton:
ping @BurntSushi (checkbox)

alexcrichton:
also ping @aturon

text/0000-manually-drop.md
```diff
@@ -0,0 +1,157 @@
...
+}
+
+impl Drop for FruitBox {
+    fn drop(&mut self) {
+        unsafe {
+            // Explicit ordering in which field destructors are run specified in the intuitive
+            // location – the destructor of the structure containing the fields.
+            // Moreover, one can now reorder fields within the struct however much they want.
+            peach.manually_drop();
+            banana.manually_drop();
```
gereeter: This should be updated to `ManuallyDrop::drop(...)`.

oli-obk:
Is it valid to ptr::read from the deref of a ManualDrop to obtain an owned value in a Drop impl?

nagisa:
I do not immediately see any reason you wouldn't be able to. Operation
leaves wrapper in a undefined state so you should take care to not drop or
act on the wrapper in any other way after the ptr::read

est31:
Cross linking https://github.com/rust-lang/rust/pull/40559

text/0000-manually-drop.md
```diff
@@ -0,0 +1,157 @@
...
+impl<T> ManuallyDrop<T> {
+    /// Wraps a value to be manually dropped.
+    #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
+    pub fn new(value: T) -> ManuallyDrop<T> {
+        ManuallyDrop { value: value }
+    }
+
+    /// Extracts the value from the ManuallyDrop container.
+    #[unstable(feature = "manually_drop", reason = "recently added", issue = "0")]
+    pub fn into_inner(self) -> T {
```
sfackler: This should be a static method like `drop`.

alexcrichton:
@sfackler there's some discussion [above](https://github.com/rust-lang/rfcs/pull/1860#issuecomment-279829731) about whether it's a free function or method, and the method was chosen because conventionally all `into_*` methods require self-by-value which is typically unusable through `Deref` and `DerefMut`

sfackler:
I was unfortunately not tuned in to the conversation at that point, oops :(

I feel somewhat strongly that it should be static - since we have an otherwise universal convention that smart pointer methods are static, if I see a `foo.bar()`, I can assume the `bar` is calling into the inner type. In addition, if/when we support `DerefMove`, an inner `into_inner` call *could* be made!

alexcrichton:
That's true yeah, I'd personally be fine either way

alexcrichton:
Looks like the main issue to arise during FCP is whether `into_inner` is an inherent method or not, but overall nothing much major arised so I'm going to merge! We can always tweak the implementation during stabilization and come back to tweak the RFC if necessary.

Thanks for the RFC @nagisa!

PR #1869, opened 2017-01-23
RFC: add `eprint(ln)!`

zackw:
Continuing discussion from https://internals.rust-lang.org/t/extremely-pre-rfc-eprintln/4635/7 and https://github.com/rust-lang/rust/pull/39229.

### DIFF
--- a/text/0000-eprintln.md
+++ b/text/0000-eprintln.md
@@ -1,0 +1,129 @@
+- Feature Name: eprintln
+- Start Date: 2017-01-23
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes the addition of two macros to the global prelude,
+`eprint!` and `eprintln!`.  These are exactly the same as `print!` and
+`println!`, respectively, except that they write to standard error
+instead of standard output.
+
+# Motivation
+[motivation]: #motivation
+
+This proposal will improve the ergonomics of the Rust language for
+development of command-line tools and "back end" / "computational
+kernel" programs.  Such programs need to maintain a distinction
+between their _primary output_, which will be fed to the next element
+in a computational "pipeline", and their _status reports_, which
+should go directly to the user.  Conventionally, standard output
+should receive the primary output and standard error should receive
+status reports.
+
+At present, writing text to standard output is very easy, using the
+`print(ln)!` macros, but writing text to standard error is
+significantly more work: compare
+
+    println!("out of cheese error: {}", 42);
+    writeln!(stderr(), "out of cheese error: {}", 42).unwrap();
+
+The latter may also require the addition of `use std::io::stderr`
+and/or `use std::io::Write;` to the top of the file.
+
+Because writing to stderr is more work, and requires introduction of
+more concepts, all of the tutorial documentation for the language uses
+`println!` for error messages, which teaches bad habits.
+
+# Detailed design
+[design]: #detailed-design
+
+Most of the design is already nailed down by the existing `println!`
+macro.  It is my intention to clone the existing definition verbatim,
+changing only the name and the I/O stream written to.  Anything other
+than strict parallelism with `println!` would be surprising and
+confusing.
+
+There remain two design decisions to make:
+
+ * Should there be an `eprint!` analogous to `print!`?  (That is, a
+   macro that writes formatted text to stderr _without_ appending a
+   newline.)  I suspect that it will be rarely used, but I also
+   suspect that its absence will be surprising.  Leaving out `eprint!`
+   might enable a better choice of name for `eprintln!` (see below).
+
+ * What should the name(s) of the macro(s) be?  There were four
+   candidates proposed in [the pre-RFC][pre-rfc]:
+
+   * `eprintln!` and `eprint!` -- easy to type, and two different
+     people said they were already using these names in their own
+     code; but perhaps too cryptic.
+
+   * `println_err!` and `print_err!` -- less cryptic, but significantly
+     more awkward to type; it is the author's personal opinion that
+     these names will lead people to continue using `println!` for
+     error messages.
+
+   * `errorln!` and possibly also `error!` -- ruled out by the
+     [`log` crate][log-crate] already using `error!` for something
+     else, IMHO.
+
+   * `errln!` and possibly also `err!` -- `err!` is too likely to be
+     taken by a crate already IMHO.  If we could omit the no-newline
+     macro, however, I rather like `errln!`.
+
+An [implementation][] (using `eprint(ln)!`) already exists and should
+need only trivial revisions after the above decisions are made.
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+Since these macros are exactly the same as the existing `print(ln)!`
+macros but for writing to stderr, it will not be necessary to teach
+how they work.
+
+It will, however, be necessary to add text to the reference manual and
+especially to the tutorials explaining the difference between "primary
+output" and "status reports", so that programemrs know _when_ to use
+them.  All of the existing examples and tutorials should be checked
+over for cases where `println!` is being used for a status report, and
+all such cases should be changed to use the new macro instead.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The usual drawbacks of adding macros to the prelude apply.  In this
+case, I think the most significant concern is to choose names that are
+unlikely to to conflict with existing library crates.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Conceivably, the ergonomics of `writeln!` could be improved to make
+this unnecessary.  There are three fundamental problems with that,
+though:
+
+1. `writeln!(stderr(), ...)` is always going to be more typing than
+   `eprintln!(...)`.  People live with `fprintf(stderr, ...)` in C, so
+   perhaps that's not that bad.
+
+1. `writeln!` returns a Result, which must be consumed; this is
+   appropriate for the intended core uses of `writeln!`, but means
+   tacking `.unwrap()` on the end of every use to print diagnostics
+   (if printing diagnostics fails, it is almost always the case that
+   there's nothing more sensible to do than crash).
+
+1. `writeln!(stderr(), ...)` is unaffected by `set_panic()` (just as
+   `writeln!(stdout(), ...)` is unaffected by `set_print()`).  This is
+   arguably a bug.  On the other hand, it is also arguably the Right Thing.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+See discussion above.
+
+[pre-rfc]: https://internals.rust-lang.org/t/extremely-pre-rfc-eprintln/4635/10
+[log-crate]: https://crates.io/crates/log
+[implementation]: https://github.com/rust-lang/rust/pull/39229/files
--- a/text/0000-eprintln.md
+++ b/text/0000-eprintln.md
@@ -11,6 +11,8 @@ This RFC proposes the addition of two macros to the global prelude,
 `println!`, respectively, except that they write to standard error
 instead of standard output.
 
+An [implementation][] already exists.
+
 # Motivation
 [motivation]: #motivation
 
@@ -40,74 +42,121 @@ more concepts, all of the tutorial documentation for the language uses
 # Detailed design
 [design]: #detailed-design
 
-Most of the design is already nailed down by the existing `println!`
-macro.  It is my intention to clone the existing definition verbatim,
-changing only the name and the I/O stream written to.  Anything other
-than strict parallelism with `println!` would be surprising and
-confusing.
+Two macros will be added to the global prelude.  `eprint!` is exactly
+the same as `print!`, and `eprintln!` is exactly the same as
+`println!`, except that both of them write to standard error instead
+of standard output.  "Standard error" is defined as "the same place
+where `panic!` writes messages."  In particular, using `set_panic` to
+change where panic messages go will also affect `eprint!` and
+`eprintln!`.
 
-There remain two design decisions to make:
-
- * Should there be an `eprint!` analogous to `print!`?  (That is, a
-   macro that writes formatted text to stderr _without_ appending a
-   newline.)  I suspect that it will be rarely used, but I also
-   suspect that its absence will be surprising.  Leaving out `eprint!`
-   might enable a better choice of name for `eprintln!` (see below).
-
- * What should the name(s) of the macro(s) be?  There were four
-   candidates proposed in [the pre-RFC][pre-rfc]:
-
-   * `eprintln!` and `eprint!` -- easy to type, and two different
-     people said they were already using these names in their own
-     code; but perhaps too cryptic.
-
-   * `println_err!` and `print_err!` -- less cryptic, but significantly
-     more awkward to type; it is the author's personal opinion that
-     these names will lead people to continue using `println!` for
-     error messages.
-
-   * `errorln!` and possibly also `error!` -- ruled out by the
-     [`log` crate][log-crate] already using `error!` for something
-     else, IMHO.
-
-   * `errln!` and possibly also `err!` -- `err!` is too likely to be
-     taken by a crate already IMHO.  If we could omit the no-newline
-     macro, however, I rather like `errln!`.
-
-An [implementation][] (using `eprint(ln)!`) already exists and should
-need only trivial revisions after the above decisions are made.
+Previous discussion has converged on agreement that both these macros
+will be useful, but has not arrived at a consensus about their names.
+An executive decision is necessary.  It is the author's opinion that
+`eprint!` and `eprintln!` have the strongest case in their favor,
+being (a) almost as short as `print!` and `println!`, (b) still
+visibly different from them, and (c) the names chosen by several
+third-party crate authors who implemented these macros themselves for
+internal use.
 
 # How We Teach This
 [how-we-teach-this]: #how-we-teach-this
 
-Since these macros are exactly the same as the existing `print(ln)!`
-macros but for writing to stderr, it will not be necessary to teach
-how they work.
-
-It will, however, be necessary to add text to the reference manual and
-especially to the tutorials explaining the difference between "primary
-output" and "status reports", so that programemrs know _when_ to use
-them.  All of the existing examples and tutorials should be checked
-over for cases where `println!` is being used for a status report, and
-all such cases should be changed to use the new macro instead.
+We will need to add text to the reference manual, and especially to
+the tutorials, explaining the difference between "primary output" and
+"status reports", so that programmers know when to use `println!` and
+when to use `eprintln!`.  All of the existing examples and tutorials
+should be checked over for cases where `println!` is being used for a
+status report, and all such cases should be changed to use `eprintln!`
+instead; similarly for `print!`.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 The usual drawbacks of adding macros to the prelude apply.  In this
 case, I think the most significant concern is to choose names that are
-unlikely to to conflict with existing library crates.
+unlikely to to conflict with existing library crates' _exported_
+macros.  (Conversely, _internal_ macros with the same names and
+semantics demonstrate that the names chosen are appropriate.)
+
+The names `eprintln!` and `eprint!` are terse, differing only in a
+single letter from `println!` and `print!`, and it's not obvious at a
+glance what the leading `e` means.  ("This is too cryptic" is the
+single most frequently heard complaint from people who don't like
+`eprintln!`.)  However, once you do know what it means it is
+reasonably memorable, and anyone who is already familiar with stdout
+versus stderr is very likely to guess correctly what it means.
+
+There is an increased teaching burden---but that's the wrong way to
+look at it.  The Book and the reference manual _should have_ been
+teaching the difference between "primary output" and "status reports"
+all along.  This is something programmers already need to know in
+order to write programs that fit well into the larger ecosystem.  Any
+documentation that might be a new programmer's first exposure to the
+concept of "standard output" has a duty to explain that there is also
+"standard error", and when you should use which.
 
 # Alternatives
 [alternatives]: #alternatives
 
-Conceivably, the ergonomics of `writeln!` could be improved to make
-this unnecessary.  There are three fundamental problems with that,
-though:
+It would be inappropriate to introduce printing-to-stderr macros whose
+behavior did not exactly parallel the existing printing-to-stdout
+macros; I will not discuss that possibility further.
+
+We could provide only `eprintln!`, omitting the no-newline variant.
+Most _error_ messages should be one or more complete lines, so it's
+not obvious that we need `eprint!`.  However, standard error is also
+the appropriate place to send _progress_ messages, and it is common to
+want to print partial lines in progress messages, as this is a natural
+way to express "a time-consuming computation is running".
+[For example][progress-ex]:
+
+```
+Particle        0 of      200: (0.512422, 0.523495, 0.481173)  ( 1184 ms)
+Particle        1 of      200: (0.521386, 0.543189, 0.473058)  ( 1202 ms)
+Particle        2 of      200: (0.498974, 0.538118, 0.488474)  ( 1146 ms)
+Particle        3 of      200: (0.546846, 0.565138, 0.500004)  ( 1171 ms)
+Particle        4 of      200: _
+```
+
+We could choose different names.  Quite a few other possibilities have
+been suggested in the [pre-RFC][] and [RFC][] discussions; they fall
+into three broad classes:
+
+ * `error(ln)!` and `err(ln)!` are ruled out as too likely to collide
+   with third-party crates.  `error!` in particular is already taken
+   by the [`log` crate][log-crate].
+
+ * `println_err!`, `printlnerr!`, `errprintln!`, and several other
+   variants on this theme are less terse, but also more typing.  It is
+   the author's personal opinion that minimizing additional typing
+   here is a Good Thing.  People do live with `fprintf(stderr, ...)`
+   in C, but on the other hand there is a lot of sloppy C out there
+   that sends its error messages to stdout.  I want to minimize the
+   friction in _using_ `eprintln!` once you already know what it means.
+
+   It is also highly desirable to put the distinguishing label at the
+   _beginning_ of the macro name, as this makes the difference stand
+   out more when skimming code.
+
+ * `aprintln!`, `dprintln!`, `uprintln!`, `println2!`, etc. are not
+   less cryptic than `eprintln!`, and the official name of standard
+   I/O stream 2 is "standard _error_", even though it's not just for
+   errors, so `e` is the best choice.
+
+Finally, we could think of some way to improve the ergonomics of
+`writeln!` so that we don't need the new macros at all.  There are
+four fundamental problems with that, though:
 
 1. `writeln!(stderr(), ...)` is always going to be more typing than
-   `eprintln!(...)`.  People live with `fprintf(stderr, ...)` in C, so
-   perhaps that's not that bad.
+   `eprintln!(...)`.  (Again, people do live with `fprintf(stderr,
+   ...)` in C, but again, minimizing usage friction is highly
+   desirable.)
+
+1. On a similar note, use of `writeln!` requires `use std::io::Write`,
+   in contrast to C where `#include <stdio.h>` gets you both `printf`
+   and `fprintf`.  I am not sure how often this would be the _only_
+   use of `writeln!` in complex programs, however.
 
 1. `writeln!` returns a Result, which must be consumed; this is
    appropriate for the intended core uses of `writeln!`, but means
@@ -124,6 +173,8 @@ though:
 
 See discussion above.
 
-[pre-rfc]: https://internals.rust-lang.org/t/extremely-pre-rfc-eprintln/4635/10
+[pre-RFC]: https://internals.rust-lang.org/t/extremely-pre-rfc-eprintln/4635/10
+[RFC]: https://github.com/rust-lang/rfcs/pull/1869
+[progress-ex]: https://github.com/rust-lang/rfcs/pull/1869#issuecomment-274609380
 [log-crate]: https://crates.io/crates/log
 [implementation]: https://github.com/rust-lang/rust/pull/39229/files


jonas-schievink:
[Rendered](https://github.com/zackw/rust-rfcs/blob/eprintln/text/0000-eprintln.md)

jethrogb:
I think there should be both `eprintln!` and `eprint!`, using those names.

I think the final name shouldn't focus too much on the “error” part. Yes, the output stream is called stderr, but it really just means “out-of-band” messages that are not part of the regular output.

comex:
However, `errln` has the advantage of having its purpose immediately apparent, whereas `eprintln` could be anything (extended? enhanced? extra?).

ExpHP:
Adding only `eprintln` without `eprint` would present an immediate challenge to anyone migrating stderr output away from `write{,ln}` and/or `print{,ln}`.

I migrated such a piece of code yesterday after seeing the pre-RFC:

```rust
for dla_step in 0..NPARTICLE {
    err!("Particle {:8} of {:8}: ", dla_step, NPARTICLE);

    // ... long, time consuming computation ...

    errln!("({:8.6}, {:8.6}, {:8.6})  ({:5?} ms, avg: {:5?})",
        (pos.0).0, (pos.1).0, (pos.2).0, timer.last_ms(), timer.average_ms()
    );
}
```

```
Particle        0 of      200: (0.512422, 0.523495, 0.481173)  (  184 ms, avg:     6)
Particle        1 of      200: (0.521386, 0.543189, 0.473058)  (  202 ms, avg:    13)
Particle        2 of      200: (0.498974, 0.538118, 0.488474)  (  146 ms, avg:    18)
Particle        3 of      200: (0.546846, 0.565138, 0.500004)  (  171 ms, avg:    24)
Particle        4 of      200: _
```

That said, this kind of usage of stderr also conflicts with its typical designation as "a safe place to dump stuff"; if I added debugging output to methods called during that computation, it would mess up my nice output.

ubsan:
I'm of the opinion that, whatever we call it, we should add it. I've reimplemented these macros in far too many Rust programs.

BurntSushi:
I would also like to have this. Like @ubsan, I've written this macro a gagillion times.

Some lightly held opinions:

1. I like `eprint` and `eprintln` the best.
2. I would be disappointed if we wound up with an underscore in the name.
3. I could live without `eprint`, but there's really not much downside to adding it for consistency with `print`.

liigo:
`errprint`/`errprintln` is another alternativea. `err` is more meaningful than `e`.

joshtriplett:
On the name bikeshedding: I feel like `err` or `errln` seem too likely to conflict with non-printing-related things, and they don't make their purpose as clear.  :+1: to `eprint!` and `eprintln!`.  (Also, I like that these start with a different letter than `println!` does, which makes them easily differentiable at a glance.)

Havvy:
I would rather see it be `err_println`. `eprintln` starts looking really esoteric (in the same way as C's function names).

There's also a consistency or learning issue. We don't use `e` to mean `error` anywhere else in the stdlib. This makes an ad hoc naming scheme that is only used for two macros, adding to the cognitive load needed to learn the language.

What does the stdlib use for errors right now? Whatever it is, we should align the name with that.

zackw:
My objection to `println_err!` (it's enough longer and harder to type than `println!` that people will be discouraged from using it even when they should) applies equally to any other name with an underscore in it.  The exclamation point is already one too many uses of the shift key; let's not add more.

text/0000-eprintln.md
```diff
@@ -0,0 +1,129 @@
...
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+Since these macros are exactly the same as the existing `print(ln)!`
+macros but for writing to stderr, it will not be necessary to teach
+how they work.
+
+It will, however, be necessary to add text to the reference manual and
+especially to the tutorials explaining the difference between "primary
+output" and "status reports", so that programemrs know _when_ to use
```
jplatte: typo: *programemrs*

jplatte:
I like @Havvy's suggestion of `err_println` from a readability perspective, and removing the underscore would negate that IMHO. But I can also see the problem with having an underscore in the name.

Overall I think I'm slightly in favor of the harder-to-type, more explicit `err_println` over `eprintln` because I personally only have very few such print-to-stderr statements in my programs, to handle errors propagated through `Result`s from inside the program logic.

petrochenkov:
:bike: I like `eprintln!`/`eprint!` specifically (as written in the RFC) because most of the time I use them for "printf debugging" and they need to be quickly typable and won't stay in the code for long :bike:

text/0000-eprintln.md
zackw: Noted, thank you.

ticki:
I dislike this strongly, for this reason:

You aren't supposed to use `print`/`eprint`. It is computationally expensive (locking a mutex = atomic operations) and unnecessary most of the time (i.e. you can do `let stdout = io::stdout(); let stdout = stdout.lock();` and only lock the mutex once). This makes it a pretty bad habit, and adding `eprint!()` will only make that more common.

I think `print` was a mistake, but it's not revertible, because it is used so much, but adding `eprint` is like digging the hole deeper, IMO.

BurntSushi:
@ticki Every command line program I've *ever* written has never needed to care about the performance of writing to `stderr`. Conversely, almost every command line program I've written writes to stderr at some point---whether it's a warning, or debug message or an error---and this macro makes this a little more convenient. Seems like a nice trade off to me.

I'm sympathetic to `println!` being a footgun because I have written programs where use of `println!` causes it to slow down. I think it is a potential stumbling block for newcomers. But that ship has sailed and I don't think the presence of `eprintln!` makes it worse.

eddyb:
FWIW it might be a better idea to go with the `log` crate, except that is too verbose by default.

sfackler:
Logging is pretty orthogonal to output designed for consumption by an end user.

ticki:
@BurntSushi 

I can't say my experiences are the same. If you log extensively, you can easily severely slow down the program, whereas using `write` means that you won't need to lock a mutex for updating the cache.

Moreover, I would say that just throwing out some logs into `stderr` is, well, hacky. There is no leveling, prefixes, etc., yet this macro gives a high-level "feel", when that's clearly not the case. So I am concerned if it will cause a reduction in use of third-party logging crates, which are more idiomatic.

That being said, this macro can be useful for debugging, but even then you rarely need to filter between stderr and stdout when you debug.

ubsan:
@ticki It's pretty common for me to use `eprintln` in order to have both normal output in the course of the program, and `println` debugging, because I need to be able to do `>/dev/null`.

joshtriplett:
My most common use case for `eprintln!` would be error messages, not logging.  For general-purpose logging, I'd use a structured logging mechanism.

BurntSushi:
There's nothing hacky about printing an error message to `stderr`.

theduke:
Please don't call it `eprintln`, that's a really esoteric/cryptic name.

These would all be better, with my preference towards `println_err`

* `println_err`
* `printerrln`
* `errln`

seanmonstar:
Out of all the suggestions, I still prefer one from the internals thread:

```rust
println!(stderr, "this happened: {}", err);
```

Adding a variant to the macro that can be either `stderr` or `stdout`, to specify.

iamcodemaker:
Bikeshed: I prefer `eprintln!()`. It's concise and similar enough to `println!()` to be relatively discoverable. You only have to learn it once, so I think it should be easy to type and remember vs optimizing for discoverabability.

dpc:
`eprintln!(...)` is a bit unclear at first, but only takes one web search to decipher it, and afterwards it's obvious and doesn't hurt the eyes in any way. Plus I think `stderr` has the same rank as `stdout` so deserves similar treatment: it's own macro. `println!(stderr, ...)` is not really cutting it (I would still macro it myself in my code).

Havvy:
You can argue that it's a thing you only have to learn once and that a web search would help, but every cryptic addition could use that rationale individually and then we'd have a lot of cryptic things.

There's already so many things that have to be learned than adding more trivial (as in trivia) things to learn is going to unfocus people from what really matters.

ubsan:
@Havvy as @dpc notes, if we go the `println!(stder,...)` route, many (like me) would still macro it to be called `eprintln!(...)` (or `errln!`, or whatever else they like) in their own code, which just means the same thing to learn, but now it doesn't have a common name.

dpc:
@Havvy : Agreed, but it's always a balance to be made. One needs to learn `println` already, and `eprintln` is trivial to understand at this point.

It seems to me that  `eprintln` is just missing, since [people keep creating it](https://github.com/search?utf8=%E2%9C%93&q=macro_rules+stderr&type=Code&ref=searchresults) . And at least once I [did to](https://github.com/dpc/rdedup/blob/95494448d09fcad56f2ae525eb7154dcd3fa90ea/src/bin.rs#L16). Having a differently named `println` in every other Rust program is more complexity for users than one standard one.

seanmonstar:
> many (like me) would still macro it to be called eprintln!(...) (or errln!, or whatever else they like) in their own code, which just means the same thing to learn,

I'm fine determining objective reasons why one or the other is better, but this reason is more like "If it is not named the way I prefer, I'll create a macro with the name I like, and thus, the other name is bad."

ubsan:
@seanmonstar More that `println!(stderr, ...)` is too long.

nnethercote:
I frequently use `fprintf(stderr, ...)` in C++ programs for temporary, ad hoc profiling (usually in combination with some light post-processing). These calls are always temporary. stderr is good for this because it's separable from stdout, and also because it's not buffered.

`eprintln!` would serve the same purpose nicely in Rust.

vi:
What if `error!` and `warning!` from `log` crate becomes eventually accessible in libstd with a logger that writes everything to stderr by default?

Would this `[e]print[err]ln[_err]!` become useless?

BurntSushi:
@vi I think this was already mentioned, but it's reasonable to consider logging and printing user errors as orthogonal.

vi:
Maybe the word "error" is overused here? `stderr` is not just for errors, also for `10%... 20%...` and other diagnostics. If one just want to pring a message from program, it probably should go to stderr.

liigo:
> My objection to println_err! (it's enough longer and harder to type than println! that people will be discouraged from using it even when they should) applies equally to any other name with an underscore in it. The exclamation point is already one too many uses of the shift key; let's not add more.

It's a shame to discourage from using underscores in API names, with the truth that many/most Rust API names has underscores.

letheed:
Another reason compelling reason to add `eprint!` is that `write!`'s behavior is actually different from `print!`. `write!(stdout(), ...)` writes directly on `stdout()` whereas `print!` uses the sink in `LOCAL_STDOUT`. Currently the only way to write to the `LOCAL_STDERR` sink is through the `panic!` macro.

Programs like `cargo test` that rely on `set_print()` and `set_panic()` to capture outputs from `print!` and `panic!` fail to capture `write!` outputs. That may be acceptable for `write!(stdout(), ...)` but currently we don't have an alternative to `write!(stderr(), ...)`.

See [this discussion](https://users.rust-lang.org/t/cargo-doesnt-capture-stderr-of-running-tests/9222).

cristicbz:
It looks like conversation here's settled and everybody likes it with a slight preference towards `eprint{ln,}` should go ahead with this?

BurntSushi:
@zackw I think the detailed design needs to be updated. Right now, the "detailed design" essentially discusses a bunch of choices rather just detailing the proposed design. I think a good chunk of it should be moved to the "alternatives" section.

BurntSushi:
@rfcbot fcp merge

zackw:
@BurntSushi I have revised the text as requested.

nikomatsakis:
So, I'm not going to disturb the consensus, but I will say that when I saw the names `eprintln!` referenced from somewhere else I had **no idea whatsoever** what it could possibly be (I imagined something to do with the environment?). I definitely want these macros, but am surprised we didn't opt to include "err" or "stderr" in the names, which would have made it crystal clear to me.

I do not understand the objection names that include `_` -- all of our functions and method names use `_` since that's the naming convention we adopted -- clearly we ain't changing it now!

joshtriplett:
@nikomatsakis The primary objection: it'd make a name we're going to use frequently longer and slower to type.  For instance, why do we have `println!` rather than `print_line!`?

`errprintln!` is the only alternative I saw that wouldn't have been excessively longer, but that particular variant didn't get much discussion/consideration.

I think the consensus of `eprintln!` will work out fine in the long run.  20 years from now, I don't think we'll look at it the way C folks look at `creat`.  (Hey, *that's* where the missing 'e' went!)

bjadamson:
I'm excited to see this merged

alexcrichton:
Reading over the discussion again it doesn't seem like there's necessarily consensus around the names `eprint` and `eprintln`. I sympathize with @nikomatsakis's reaction as well in that it does seem that unless you're "in the know" these could be confusing on first glance. Now that being said I do think that the name is evocative enough that there's only so many possibilities for what it could be doing. (e.g. it's different than println, but it's still printing, just somewhere...)

I'm not personally 100% sold by the "shorter names are more ergonomic" line of reasoning, and I'd be up for longer names like `println_err` or `println_stderr` (or something like that). We've done this in the past with `debug_assert`, for example ([loc1](https://github.com/rust-lang/rfcs/pull/50) [loc2](https://github.com/rust-lang/rust/pull/12108)). These aren't really called in every other line of code, so there's not really a great need to save a few bytes. Otherwise more evocative names tend to definitely be helpful for readers.

jethrogb:
I agree with @alexcrichton that I don't think there's consensus around the name, although I myself am in favor of eprintln.

> These aren't really called in every other line of code

`debug_assert` isn't, but eprintln arguably should be the default people reach for when printing.

joshtriplett:
I don't think there's complete consensus, but for a topic like this I don't think we'll *ever* have complete consensus (🚳).

nikomatsakis:
I agree people will get used to `eprintln!`. I just think that `println_err!(...)` is perfectly fine -- arguably better, since then I can start typing `prin` and have the IDE complete the rest for me, whereas I have to know whether I want `eprintln!` before I start typing.

dpc:
@nikomatsakis Argument about auto-completion is quite convincing for me.

joshtriplett:
(Note: we're treading into serious bikeshed territory here.)

@nikomatsakis That makes things worse for people who don't have autocompletion, though.  And in my opinion, it also decreases code readability-at-a-glance, by making the two less quickly distinguishable.

nikomatsakis:
@joshtriplett 

> (Note: we're treading into serious bikeshed territory here.)

I agree, and I think this'll be my last comment on the topic, since I think I've said my 2c, and I'm happy to abide by whatever the @rust-lang/libs team decides. That said, let me respond to your two comments.

> That makes things worse for people who don't have autocompletion, though.

Worse in what way? Because they have to type an `_`? It seems like this is par for the course when writing Rust code to me. 

>  And in my opinion, it also decreases code readability-at-a-glance, by making the two less quickly distinguishable.

Hmm, let's compare:

```
println!("foo");
eprintln!("foo");
```

vs

```
println!("foo");
println_err!("foo");
```

If anything, the latter seems much more clearly distinguishable to me -- it's much longer and has an `_` in it. But perhaps if they were embedded in a bunch of code that would be less true?

aturon:
FWIW, I didn't give much thought to the name when I reviewed, but I definitely agree with @nikomatsakis's suggestion of `println_err`. This is already a huge improvement over `writeln!(::std::io::stderr(), ...)`. The name is, to me, more immediately obvious in meaning than `println_err`, and easier to remember. And I don't think it's used *so* often as to necessitate minimizing character count. (See also [my recent blog post](https://blog.rust-lang.org/2017/03/02/lang-ergonomics.html) on some of these considerations).

As far as bikeshedding goes: I think it's fine to spend some time discussing the name before we land it, but we shouldn't block the RFC on achieving final consensus here; we'll have a separate FCP for the tracking issue anyway.

joshtriplett:
For the sake of bikeshed-avoidance, I'll stop commenting on this as well; almost any reasonable name would be better than none.

nnethercote:
C has printf, fprintf, sprintf, vprintf, etc. So that's a precedent for single letter prefixes of this type. Whether it's a good precedent is a separate question. I'm not aware of deep distaste for those names but I may have overlooked it.

Havvy:
@nnethercote When I first saw those names, and still even now, I still don't like those names. I know `f` means `format` as a suffix, but I'm not sure what any of the prefixes mean. Granted, I've also not actually gone into the documentation for C to see what they are. It just looks arcane to me.

eddyb:
`f` is "file", `s` is "string", `v` is "variadic" (only the last one is rare enough that you may never see it)

vi:
Doesn't `println_err!` imply that some error has happened?
I think `stderr` is currently a misnomer because of majority of things getting printed to it are not errors.

Maybe Rust should not continue this chain and start calling the "print directly to user, not to the next program in pipeline" action without the loaded word "error"?

jplatte:
> it seems to be named like a thing specifically for errors.

I'm pretty sure that was the original intention, yes. But two output buffers were standardized in POSIX and probably elsewhere, and people had to cope, so they reused stderr for non-error output in some situations.

IMHO the best solution is just to stick with the names, at least they will be obvious to some people. Moving the perceived line of what belongs to stdout / stderr might be helpful for some purposes, but will be unhelpful for others; and a `println_err` renamed so it doesn't include `err` will be obvious to noone. Also, we can't change the name of `std::io::stderr` until Rust 2.0 anyway, even if we wanted to. Introducing an inconsistency for just `println` seems like a bad idea to me.

BurntSushi:
I said this in the beginning:

> * I like eprint and eprintln the best.
> * I would be disappointed if we wound up with an underscore in the name.

I think I still largely have these opinions, and they are almost entirely aesthetic. I can't articulate my argument beyond that though. (Which isn't to say that I'm conceding, I think aesthetics are... valuable... to a degree.) With that said, I think someone just needs to make a decision. While it seems a bit overwrought, is this an example of where the core team would just make the call?

N.B. I find it absolutely delightful that [`eprintln!`](https://github.com/search?l=Rust&q=eprintln%21&type=Code&utf8=%E2%9C%93), [`println_err!`](https://github.com/search?l=Rust&q=println_err%21&ref=searchresults&type=Code&utf8=%E2%9C%93) and [`println_stderr!`](https://github.com/search?utf8=%E2%9C%93&q=println_stderr%21&type=Code&ref=searchresults) already have widespread use, with a preference for `println_stderr!`.

nathanaeljones:
As far as autocomplete goes, keep in mind the number of disambiguating home and top-row keystrokes required. Grouping terms by prefix is good for autocomplete discovery, but tends to work strongly against autocomplete ergonomics. 

Is there an easy way to alias a macro?

GuillaumeGomez:
> Is there an easy way to alias a macro?

Calling them into another macro?

n3vu0r:
Maybe another view on consistency emphasizing the actual intention and not the technical background of either `stdout` or `stderr`:

We have the most general macros `write(ln)!` which *write* messages to arbitrary output streams. Special streams are `stdout`/`stderr`. The first is for *printing* to the terminal in general. The second is also for printing to the terminal, but more precisely for *alerting*. This might be for alerting about errors in the original sense of `stderr` or just for alerting about status updates while `stdout` is used to transfer the actual data of interest, for instance, piping it to another process. For both, the errors and the status updates, the word *alert* suits well in my opinion. So I would suggest the macros `alert(ln)!` which have exactly 5 and 7 characters like `print(ln)!`. Using a completely different word like `alert`, instead of some characters prefixed to `print` like `eprint`, might better emphasize the intention why we distinguish how we write (print/alert) to a terminal, as long as this word is intuitive for both, errors and status updates. I think `print(ln)!`/`alert(ln)!` most precisely reflect the real world usage of `stdout`/`stderr` and would be most intuitive for new users reading the examples of the Rust book. So we would go from general `write(ln)!` over less general `print(ln)!` to most specific `alert(ln)!`.

aldanor:
Big thumbs up for having macros with *some* names for printing to stderr.

In re: naming, I personally and subjectively concur with @BurntSushi in both that (1) `eprint!` and `eprintln!` seem like the most ergonomic choices, and (2) that having `_` in this macro would be sad -- it's `println!` and not `print_line!`, after all, although one could argue that the latter is more transparent to an unwary reader who's never seen Java or Pascal.

To add to the already epic amount of bikeshedding in this thread, I actually did measure the average time to type `eprintln!` and `println_stderr!`, and the latter is 2.3x slower for me on average, with lower average accuracy, due to (1) increased length, (2) added underscore, and (3) `stderr` being uncomfortable to type in general since all letters 's', 't', 'd', 'e', 'r' are all clustered on the left side of the keyboard. YMMV though.

nikomatsakis:
@BurntSushi 

> While it seems a bit overwrought, is this an example of where the core team would just make the call?

I feel this is abdicating your rightful responsibility as a member of the libs team! You are the one designated to make annoying calls like this! =)

nagisa:
IME it doesn’t really matter how short, pleasant and whatnot this macro is to type down. After all, how often will you litter your code with printouts to stderr? If you have significant alerting and messaging to do, you will surely use some proper logging framework?

All things considered I this function would be written rarely and by extension *read* rarely as well, which calls for a name that’s obvious in its purpose from its name. In that sense `print{,ln}_error` or `print{,ln}_stderr` seem like the no-brainers.

iamcodemaker:
Having heard arguments from various sides, my bikeshed remains a deep shade of `eprint{,ln}!()` 🙃

zackw:
As the author of the RFC, I want to state for the record that I, too, care far more about having these macros under _some_ name than about what the name actually is.  (But my personal preference is for `eprint(ln)!`, and failing that, for having the difference from `print(ln)!` at the _beginning_ of the name rather than the end.)

I'd like to try to answer this objection, though:

> All things considered I think this function would be written rarely and by extension _read_ rarely as well, which calls for a name that’s obvious in its purpose from its name. In that sense `print{,ln}_error` or `print{,ln}_stderr` seem like the no-brainers.

(This is not to pick on @nagisa specifically, several other people have expressed more-or-less the same sentiment.)

I suspect that people who want a short, quick-to-type, yet still visually obvious name, and don't mind if it's cryptic (such as myself) habitually use a different debugging strategy than people who want a long and transparent name.  I expect it is true that in production code there will _normally_ not be very many instances of `eprintln!`.  But if I am debugging something, one of the first things I will try is peppering the functions that I think are at fault with `eprintln!`.   In extreme cases, _every other line_ will become an `eprintln!` call.  I don't check them in, because they're often much too detailed for even "debug-level" general-purpose logging, and at the same time, they are useless for any scenario but the one I care about right now.

And then when I discover that I'm wrong about the cause of the problem, I delete all of the `eprintln!` calls I added, form a new hypothesis, and invent a whole new set of printouts to probe the new hypothesis.

So my workflow involves typing `eprintln!` a whole lot, even though it might not look like it to look at my code; and my workflow also involves _deleting_ `eprintln!` a whole lot, so I want it to be easy to grep for and visually skim for.  (This is why it matters to me to have the discriminant at the beginning of the token.  In C, finding all the `fprintf(stderr, ...)` calls that I put in involves more brain and/or more complicated regexps, because `fprintf` is used for other things.

zackw:
Oh, also:

> So we would go from general `write(ln)!` over less general `print(ln)!` to most specific `alert(ln)!`.

I would be fine with `alert(ln)!` as long as it was known not to have been claimed already by some third-party crate.

n3vu0r:
@zackw I have just downloaded the latest versions of all the 8357 crates and egrep'ed every `*.rs` file against the pattern `alert!|alertln!` without a single match.

stjepang:
If I saw `alertln!("message")` in Rust code without seeing this discussion, I'd be sure that the message goes through a logging framework into a file or perhaps through some kind of custom notification system.

Honestly, I've never before seen the word "alert" in the context of stderr, so this is a very unintuitive choice for me. Java uses `System.err.println`, C++ uses `std::cerr`, and OCaml uses `eprintf`. The file descriptor is usually referred to as "stderr" or "standard error". Naturally, all those functions have "err" or at least an "e" in their names.

"Standard error" might be a slight misnomer (not too bad, IMHO), but everyone knows what it is. This name is not going away and noone has ever tried going against it.

My vote goes to `eprintln!`, but I'm also fine with `println_err!`.

golddranks:
I think `errprintln!` is underrepresented in the discussion, even though it has generally the same properties as `eprintln!` but being less cryptic.

ticki:
@golddranks I like that, except that I prefer an underscore: `err_println!`.

ExpHP:
Now we're getting into shades of maroon and periwinkle!

I'm not even going to state my preference.  I say we just blindfold somebody, walk them into the paint can aisle, and hand them a magnetic dart.

mmstick:
I'm still in favor of my original proposal in another thread -- just expand the existing `println` / `print` macros, and then there's no longer a need to invent a new keyword.

```rust
println!("Same println as usual");
println!(stderr, "comm: error occurred: {}", err);
println!(stdout, "not an error {}", value);
```

Note that the `stderr` and `stdout` keywords are not input values, but keywords.

mmstick:
@joshtriplett It's not an obect/type but a keyword, and it's trivial to implement:

```rust
macro_rules! println {
    // Current macros
    () => { ... };
    ($fmt:expr) => { ... };
    ($fmt:expr, $($arg:tt)*) => { ... };
    // Proposed stderr expansion
    (stderr, $fmt:expr) => { ... };
    (stderr, $fmt:expr, $($arg:tt)*) => { ... };
    // Or this one
    (stderr $fmt:expr) => { ... };
    (stderr $fmt:expr, $($arg:tt)*) => { ... };
}
```

Simple.
```

**Comment by dpc** [comment]
@mmstick  I'm OK with any shade of paint here, but this versions seem to me to have only disadvantages:  confusing, longer to type, harder to discover, harder to autocomplete. People will have trouble telling if `stderr` is from `io::stderr`, since you already had to clarify it here. And generally the least macro magic, the better. I with `println` wasn't a macro in the first place.

**Comment by mmstick** [comment]
@dpc In no way, shape, or form is it confusing. I'm not sure how you're coming to that conclusion.

- Confusing: This isn't true because we already know that the `println!()` macro prints to stdout. If you already know that you're wanting to print to stderr, then it's not confusing to specify that intent in the macro as so: `println!(stderr "message")`.
- Longer to type: This is irrelevant -- inventing a new, short form keyword will only cause confusion and will likely break existing code -- basically the opposite of what you're wanting. If you want a shorthand for `println!(stderr, "message")`, then you could create a macro yourself.
- Harder to Discover: Not at all, it would be front and center on the println page with existing println examples. I'd say it's much harder to discover a completely new keyword rather than using one that already exists.
- Harder to Autocomplete: 100% false -- your autocompleter would have no problems with this.

> People will have trouble telling if stderr is from io::stderr, since you already had to clarify it here.

stderr is stderr -- there should be zero confusion here in that `println!(stderr, "message")` will print to stderr instead of stdout. That there is a function in `io` named `stderr` doesn't mean anything here.

> And generally the least macro magic, the better.

There is no macro magic involved, and certainly less macro magic than if you were to create a new keyword. Creating a new keyword would give you two separate macros in two separate locations in the source code. Just adding a few more branches to the existing `println!` macro would mean keeping everything conveniently in the same place.

**Comment by louy2** [comment]
For reference, the current `println!` macro implementation is here: https://github.com/rust-lang/rust/blob/fc6f092c21a7a7249a9f8860f3cd10160aa36c02/src/libstd/macros.rs#L121

**Comment by birkenfeld** [comment]
@mmstick 

> stderr is stderr -- there should be zero confusion here in that println!(stderr "message") will print to stderr instead of stdout. It doesn't come from anywhere -- it's a keyword. That there is a function in io named stderr that returns a handle to the process's stderr doesn't mean anything here.

Maybe you want to scale back the assertiveness a little here. Knowing the current `println!()`, I would be quite confused upon seeing `println!(stderr, ...)` and looking everywhere for where `stderr` got imported, and wondering if someone tried to spell `writeln!` wrong.

I don't think an macro that usually takes lots of expressions should now take something that looks like and identifier but is actually a magic token, especially when there's a closely related macro that *actually* takes an expression in that place.

**Comment by mmstick** [comment]
@birkenfeld 

> I don't think an macro that usually takes lots of expressions should now take something that looks like and identifier but is actually a magic token, especially when there's a closely related macro that actually takes an expression in that place.

It doesn't look like an identifier, and it's not magic. You're not using the same syntax I outlined above:

```rust
println!(stderr "message to stderr")
```

In addition, you wouldn't be confused once it became the standard.

SamWhited:
> In addition, you wouldn't be confused once it became the standard.

People who know that it's a standard aren't the ones we need to worry about confusing though; people who are brand new and don't understand why one of those funny exclamation mark functions has a thing in it that's not imported are the ones we need to worry about confusing.

I tend to agree with @mmstick the `println!(stderr…` syntax looks simple and nice, but also with @birkenfeld: Please don't invent new syntax, the implementation is simpler, but it makes using, learning, and understanding things much more complex.

birkenfeld:
> It doesn't look like an identifier, and it's not magic. You're not using the same syntax I outlined above:

Well, when I started writing the message there were still both forms in your mock `macro_rules` :) For this form I agree that because it's much more jarring, the confusion with an unimported identifier is less likely.

mmstick:
@SamWhited People new to Rust will be confused by pretty much everything to begin with, such as why there needs to be an exclamation mark when using `println!`, and why the formatter takes curly braces as tokens. If you have to explain to a new user that writing stderr before the message will print to stderr instead of stdout, that won't change by using a new keyword and then having to explain to a user that there's a completely different macro that does the same thing, but prints to stderr. Then it's memorizing two macros, not one.

Like I said before, when I started learning Rust two years ago and saw that the `println!` macro was the preferred way to write to standard output, I wondered if the macro supported printing to stderr too by specifying stderr before the message, since this is the sort of thing that macros were logically made for. Sadly, it does not. It'd be a bit silly if we were to invent a new keyword that could have been solved by using the existing macro and just adding a stderr routine. That would make Rust's macro system look bad, in that Rust's own designers aren't confident enough to utilize it's strengths here.

BurntSushi:
I'm firmly against the `println!(stdout, ...)`/`println!(stderr, ...)` form. IMO, it's a new and surprising syntax.

zackw:
I don't like the `println!(magic[,] "format", ...)` either. I left it out of the RFC on purpose, because it seemed so clearly worse than a new `somethingprintln!(...)`.

There are three things wrong with it in my opinion:

 * It uses context-sensitive keywords, that _look like_ variables, but aren't.

    Context-sensitive keywords are troublesome to begin with — just look at SQL — and they're especially bad when they look like variable names.  People are going to wonder why they can't put an arbitrary writeable in that position, and then when we tell them they have to use `writeln!` for that, they are going to be confused about why we have both of these things.

* `println!`, as is, is easily understood as a shorthand for a particular use of `writeln!` (it's not _exactly_ the same but that's not relevant for a beginner).  The proposed `whateverprintln!` will also be understandable as a shorthand.  `println!(magic, ...)` complicates the shorthand again.  It's a wrinkle on top of a wrinkle.  It'll be harder to teach.

 * It doesn't put the discriminant at the _beginning_ of the statement, which improves code-skimmability, which is very important to me in this case.

aldanor:
So, counting the positive voices in this thread (and not counting negative ones), we have 4 most popular options so far, sorted by number of voices:

- `eprint(ln)!` > 10
- `err_print(ln)!` or `print(ln)_(std)err!` < 10
- `errprint(ln)!` < 5
- `print(ln)!(stderr, ...)` < 5

Coincidentally, GH has 4 reaction icons, can we vote? :)

:+1: = `eprint(ln)!`
:heart: = `err_print(ln)!` or `print(ln)_(std)err!`
:tada: = `errprint(ln)!`
:-1: = `print(ln)!(stderr, ...)`

(icons have no meaning attached other than being markers)
(:+1:ing this message is a vote for `eprint(ln)!`)

aldanor:
Amen.

alexcrichton:
I don't think we're *quite* ready to merge this as we're still bikeshedding the name. That being said I plan to at least bring this up during libs triage (this coming Tuesday) and we'll discus the best course of action to make progress on the "Error Macro Naming Crisis v2017"

liigo:
`println!(stderr, ...)` is a really nice api IMHO, maybe this is the future, ... though somebody don't like it *now*.
or `stderr_println!(...)`, which is more descriptive than eprintln.

n3vu0r:
> Honestly, I've never before seen the word "alert" in the context of stderr, so this is a very unintuitive choice for me. Java uses System.err.println, C++ uses std::cerr, and OCaml uses eprintf. The file descriptor is usually referred to as "stderr" or "standard error". Naturally, all those functions have "err" or at least an "e" in their names.

> "Standard error" might be a slight misnomer (not too bad, IMHO), but everyone knows what it is. This name is not going away and noone has ever tried going against it.

I'm not trying to start hiding `stdout`/`stderr`. That is already `print(ln)!` doing by not mentioning `stdout` as in `out.println()` or `cout`. And it's a good thing, there is no need to tell about `stdout`/`stderr`. `print(ln)!` just says *what* it does, compactly and expressively. The documentation then states that it's using `stdout`, no need to carry this artifact with every invocation of the macro. My point is that it would be consistent to give `stderr` its own macro *as well*, just saying *what* it does, alerting about errors and status updates, *and not* printing in general but just to a different stream. That is the real world usage of `stdout`/`stderr`, so the names *print* and *alert* suit well. But if someone wants to be technical and neutral regarding the common use cases, there is already `write(ln)!` accepting both `stdout`/`stderr` with a more general name *write*. For me, `alert(ln)!` elegantly complements `print(ln)!`, being consistent and on the same level as compact and expressive as  `print(ln)!`.

nicoburns:
-1 to `alert(ln)!` on the basis that I don't think all (or even most) uses of `stderr` are for alerts. Alert implies an urgent message, which error logs may not be. Furthermore, an "alert" (in the common usage of the word) may not be an "error" at all and as such one might want to send an "alert" to `stdout`.

+1 for `eprint(ln)!

stjepang:
@n3vu0r Well, we'll just have to disagree then.

Sorry, but big -1 to `alertln!` from me as well. It sounds like it's going to pop up a message box or append a message to the logs.

We're talking about a macro that writes to stderr - no more, no less. Literally all other suggestions we saw have understandable names. The only one whose meaning I'd never be able to guess is `alertln!`. Of course, YMMV. :)

carols10cents:
Given that teaching this is a large component, and the book is expected to be revised to use this appropriately, and the book is going to be printed in the ever-more-near future, I'd like to request that a name, any name, be chosen quickly so that this can stabilize before the book goes to print. 

Also, at that point, I would love if someone who feels strongly about this could read through the second edition of the book and file an issue or PR for all the cases that should change.

dpc:
While community can keep bikeshedding, I think we will all respect a call made by Rust Team (Core/Language/Library?).

cristicbz:
> So, I'm not going to disturb the consensus
-- @nikomatsakis

matthieu-m:
I fully second @dpc here. No matter the name we'll need to get used to it anyway, so I don't see much point in bikeshedding it to death.

Let's just make it obvious that it's about *printing* to *stderr*.

k3d3:
One vote for `eprintln!()`

zackw:
@carols10cents 
> I would love if someone who feels strongly about this could read through the second edition of the book and file an issue or PR for all the cases that should change.

I will make time to do that if no one beats me to it. Where can I find the publication schedule for the book, so I can know when it needs to happen by?

kennytm:
| Proposal | Pros | Cons |
|---|---|---|
| `eprint[ln]!` | Short | Cryptic |
| `error[ln]!` | Clear | All good names are taken by [log](https://doc.rust-lang.org/log/log/index.html) already 😡 |
| `print[ln]_err!` | Descriptive | Underscore |
| `print[ln]err!` | No underscore | Long |
| `errprint[ln]!` | Doesn't start with "p" | Long, and flipped word order |
| `stderr_print[ln]!` | Very descriptive | Very long |
| `print[ln]!(stderr)` | Similar to `fprintf` | Magical |
| `alert[ln]!` | Specific | No this name is for the GUI message box |
| `write[ln]!(stderr())?` | Status quo | Status quo |

Here let me propose two more bike-shed colors 🚴‍♀️🚴.

1. Since the macro in this RFC is proposed for status **report**s, why not `report[ln]!`?

    ```rust
    reportln!("out of cheese error: {}", 42);
    ```

2. The main problems of the status quo `write[ln]!(stderr(), "help").unwrap()` are:

    * `stderr()`, it is a function call, function call gives an impression of being expensive. All other languages provide stderr as a constant or property.
    * `.unwrap()` written by user, 'nuff said.
    * `use std::io::stderr;`, you need to import a module, importing a module gives an impression of bloating the program.
    * [`LOCAL_STDERR` ≠ `stderr()`](https://github.com/rust-lang/rfcs/pull/1869#issuecomment-276566460) (though I think it's not a real motivation)

    So what if we:

    * Put `STDERR` to std::io that `deref()` to `LOCAL_STDERR` (similar for STDOUT and STDIN), and `use` it in prelude, and
    * Provide a macro that simply `.unwrap()`/`.expect()` the result of `write[ln]!`?

    ```rust
    fprintln!(STDERR, "out of cheese error: {}", 42);
    ```

----

I am actually skeptical that we cannot use `error[ln]!` or `warn[ln]!` due to conflict with logging crates, because #1561 has been merged.

```rust
#![feature(use_extern_macros)]

macro_rules! vec {
    ($a:tt) => { "no I'm not a vec :p" }
}

fn main() {
    println!("{:?}", vec!(4));
    println!("{:?}", std::vec!(4));
}
// std::vec! is fine, so why not std::error!?
```

----

BTW, if users of most other languages can live without a dedicated print-to-stderr function, I don't see how we are escalating to a crisis here 😄. I don't hear people saying Python / Go / C# / Swift have bad ergonomics because you don't have a simple function to write to stderr.

| Rust | print[ln]! | "eprint[ln]!" | write[ln]!(*f*) |
|---|---|---|---|
| C | printf | fprintf(stderr) | fprintf(*f*) |
| C# | Console.Write[Line] | Console.Error.Write[Line] | *f*.Write[Line] |
| C++ | std::cout << | std::cerr << | *f* << |
| D | writef[ln] | stderr.writef[ln] | *f*.writef[ln] |
| Go | fmt.Printf | fmt.Fprintf(os.Stderr) | fmt.Fprintf(*f*) |
| Haskell | print | hPrint stderr | hPrint *f* |
| Java | System.out.print[ln] | System.err.print[ln] | *f*.print[ln] |
| JavaScript | console.[debug/log/info] | console.[warn/error] | — |
| Lua | print / io.write | io.stderr:write | *f*:write |
| PHP | echo / print | error_log | fwrite(*f*) |
| Python | print | print(file=sys.stderr) | print(file=*f*) |
| Ruby | p / puts | warn | *f*.write |
| Swift | print | [¯\\\_(ツ)\_/¯](http://stackoverflow.com/questions/24041554/how-can-i-output-to-stderr-with-swift/27118822) (or NSLog) | print(to: &*f*) |

ExpHP:
> LOCAL_STDERR ≠ stderr() (though I think it's not a real motivation)

Eh?  To me this is the primary motivation!... yet by that same token, I still ultimately agree with your conclusion.  There should be more ways to write capturable output than through a macro; since macros cannot be passed around like files can, and are impossible to abstract over except through more macros. (well, okay, you can do some amount of abstraction with callbacks, but only so much!)

---

To be honest, I'm not sure I understand why rust's `std{out,err}()` don't return the local variants to begin with.  But of course, I am coming from the (very different) mindset of python, where `sys.{stdout,stderr,stdin}` can be easily swapped out by simple assignment. (the global streams, if one really needs them, remain available at `sys.__stdout__`)

bright-star:
> I am actually skeptical that we cannot use `error[ln]!` or `warn[ln]!` due to conflict with logging crates, because #1561 has been merged.

And in the discussion in #1561, one concern was about spurring use of the macro import/export system and demonstrating it was understandable and manageable. This seems like a perfect job for that (and I'm a fan of this syntax).

carols10cents:
> @carols10cents
> 
>     I would love if someone who feels strongly about this could read through the second edition of the book and file an issue or PR for all the cases that should change.
> 
> I will make time to do that if no one beats me to it. Where can I find the publication schedule for the book, so I can know when it needs to happen by?

Thank you @zackw! Basically as soon as this stabilizes... we don't have any hard dates yet, it's just "as soon as everything gets done". See [all the states the chapters have to go through and which chapter is in which state](https://github.com/rust-lang/book/projects/1) (be sure to scroll to the right), chapters 2, 3, 4, and 6 are supposed to be pretty much "frozen" right now with only small changes (I hope this to be a small change) since they'll be going through layout soon.

suhr:
I was about to propose `report[ln]!` too.

vi:
`report[ln]` idea is like `alert[ln]` idea version 2.

vi:
Can we reuse `log`'s macro names in a way that connecting `#[macro_use] extern crate log` automatically redirects the stderr console output to logs?  
Some macros (for no-newline printing/logging) may even be added to `log` to maintain symmetry.

liigo:
CAUTION
There is a very *weird* thing in this RFC: many people ask for a *very short* (and cryptic) name.
This is not common in the community.

We may create very short name, but only for very useful use cases. Per kennytm's [comment](https://github.com/rust-lang/rfcs/pull/1869#issuecomment-285956598) above, many other languages don't have short name for this use case (i.e. writing to stderr).

`stderr` is already a double-abbreviated name: standard=>std, error=>err. I don't think it can be even crazily shorten as `e`. `eprintln` is a bad & cryptic name.

dpc:
@liigo A program should print only a direct result to `stdout`, and errors, status updated, warnings to `stderr`. Since usually there's only one place when the result is printed out, and many, many potential errors that can happen along the way, code printing to stderr is actually way more common than one printing to stdout. At least if you want to do stuff with best practices in mind. Because of that people don't want it to be too noisy and cumbersome to type. I'd say that if it wasn't a breaking change, I wouldn't mind `println` to print go to stderr in the first place, and `println_stdout` be `stdout` version.

alexcrichton:
Ok the libs team convened today and we discussed this RFC. Overall there definitely seems to be broad agreement to the *functionality* of these macros and the question was the name. Of the many many names proposed here we've decided to stick with the original proposal, `eprint` and `eprintln`.

Thanks again for the RFC @zackw!

alexcrichton:
Tracking issue: https://github.com/rust-lang/rust/issues/40528

SRGOM:
I know this is a done deal but I'd like to very briefly present my thoughts against it-

Given that stdin/stdoout/stderr are unix file descriptors, including this in standard rust library with very specific use is not a  "rustic" way of handling things. I personally think the right way would be that we have multiple streams: Out, trace, debug, warn, error and at the start of a program the runtime initializes (or user manually) each based on what file descriptors are provided by the OS/shell. `println!` would just be an alias to `streamln!( 1, ... )`, and `errln!` to `streamln( 2, ... )`, `traceln!( 3, .. )` (Or we could start over and give them better nubers, out -> 1, trace -> 2, debug -> 3). 

So on windows, where there is only one output stream available- all these would direct to "0". On Unix, which guarantees stdout and stderr, all but "println!" would direct to "1". 

I'm not trying to start a flame war and I'm typing this on Unix myself but let's face it, these streams were designed when there was no GUI. The reason rust works is because rust has thrown out all "conventional" wisdom and starts with 0 baggage. I think a 0 baggage here would be having an option to have multiple streams. 

I haven't given too many details/some inconsistent details but I can flesh out more in the off chance you guys might be interested in thinking about it.

dpc:
@SRGOM: I've seen this in Powershell, and I dislike it. It really complicates things, for no good reason. `stderr` and `stdout` are meant to be outputs, not for logging levels. Rarely any program produces more than one actual output. The `stderr` is a complication so that there's a stream that is not an output that was intended, but something else. "something else" could mean different things: typically errors, or progress information. From the perspective of different programs, there's rarely any need to route different logging levels to different places, so putting them in different streams is not useful at all.

PR #1883, opened 2017-02-03
Move metadata into import libraries for MSVC

Diggsey:
[Rendered](https://github.com/Diggsey/rfcs/blob/f86a574b9bd54e94ba6b38d0674f6cabb7aff4b6/text/0000-metadata-in-import-libs.md)

### DIFF
--- a/text/0000-metadata-in-import-libs.md
+++ b/text/0000-metadata-in-import-libs.md
@@ -1,0 +1,83 @@
+- Feature Name: metadata_in_import_libs
+- Start Date: 2017-02-03
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+When targetting the MSVC toolchain and building a rust `dylib`, embed rust-specific metadata into the static import library
+instead of the DLL.
+
+This RFC is entirely specific to the MSVC toolchain.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


aturon:
cc @alexcrichton @brson

alexcrichton:
The precise location of metadata is largely just an implementation detail of the compiler itself, thankfully! In that sense this sounds like a great idea to me with the wins here. I don't think that we need an RFC here as well, we can just implement this and review the implementation (or at least that's what I would think)

Also cc @vadimcn

vadimcn:
:-1: (for being MSVC-only).  If we care about dylib size, we should move metadata to a sidecar file on all platforms.

retep998:
@vadimcn MSVC is the only platform where we already have an existing secondary file we can use. On all other platforms you'd have to add a new file to hold the metadata, and some build systems might not be prepared for that which could cause failures.

briansmith:
> 👎 (for being MSVC-only). If we care about dylib size, we should move metadata to a sidecar file on all platforms.

I agree with @retep998. Let's fix it for MSVC where the solution is obvious. Then we can look at fixing it for other platforms, where it may or may not be reasonable.

BTW, I'm not sure an RFC is needed for the MSVC change given what @alexcrichton said; it's just an implementation detail of the MSVC toolchain.

Diggsey:
OK, I'll close this if no RFC is needed for the change.

PR #1927, opened 2017-02-24
Reduce the number of constraints repeated when writing `T: SomeTrait`

sgrif:
[Rendered](https://github.com/sgrif/rfcs/blob/sg-visible-where-clause/text/0000-visible-where-clause.md)

### DIFF
--- a/text/0000-visible-where-clause.md
+++ b/text/0000-visible-where-clause.md
@@ -1,0 +1,178 @@
+- Feature Name: visible_trait_where_clause
+- Start Date: (2017-02-24)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Today when you write `T: SomeTrait`, you are also implicitly allowed to rely on
+`T: AnySuperTraits`. However, if `SomeTrait` has additional constraints besides
+supertraits, they are must be repeated every time `T: SomeTrait` is written.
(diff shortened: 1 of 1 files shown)


nikomatsakis:
There is a rustc bug on this issue as well:

https://github.com/rust-lang/rust/issues/20671

I am generally in favor of this change. I have to give the RFC a detailed read. I am not sure if there are backwards compatibility concerns -- if so I would expect them to arise only in corner cases, but still worth considering.

durka:
Why leave the several exceptions where supertrait bounds would not be inherited? It seems less confusing to inherit everything.

In particular I don't understand the difference between `Vec<Self>: Bar` and `Result<Self, T>: Bar`.

soltanmm:
When it says that `Self` appearing in the target means that the constraint will be inherited (elaborated?), does that include associated type projections (e.g. `Self::Foo: Trait`)? Either way, could that be explicitly mentioned?

aturon:
@nikomatsakis Any thoughts on this? Are you inclined to subsume this under a more general implied bounds RFC? Can we mentor someone to write that?

nikomatsakis:
I am going to close this RFC in favor of https://github.com/rust-lang/rfcs/pull/2089, which proposes a more general variant of the design described in here! But please, @sgrif or others, let me know if you see anything that this RFC covered that #2089 does not!

PR #1967, opened 2017-04-05
Shuffle around text to avoid > at beginning of line (block quote)

aidanhs:
Compare the third bullet point (shift operations) under

 - (original) https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md#arithmetic-operations-with-error-conditions
 - (new) https://github.com/aidanhs/rfcs/blob/aphs-fix-integer-overflow-markdown/text/0560-integer-overflow.md#arithmetic-operations-with-error-conditions


alexcrichton:
Thanks!

PR #1968, opened 2017-04-10
Allow any Displayable type for expect.

clarcharr:
Formal RFC for #1922.

[Rendered](https://github.com/clarcharr/rfcs/blob/expect_display/text/0000-expect-display.md)

### DIFF
--- a/text/0000-expect-display.md
+++ b/text/0000-expect-display.md
@@ -1,0 +1,80 @@
+- Feature Name: expect_display
+- Start Date: 2016-04-09
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Extend `expect` to take `&T` where `T: Display` instead of just `&str`.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


mark-i-m:
I am not sure I understand how this saves on ergonomics... Wouldn't you still need to wrap the displayed value of the argument in text to get a coherent error message? Alternately, you could define `Display` yourself for some type, which seems to dwarf the overhead of simply writing `expect(|e| panic!(...))` unless you are printing the same messages in many places.

Have I missed something?

aturon:
Thanks for the RFC! However, I personally feel like the benefits here don't outweigh the added complexity of the API signature.

This *is* a case where method-style macros would be nice, to allow for something more akin to `format!`. But sadly method-style macros have numerous issues.

In any case, this feels quite niche to me, and I think the current ways of achieving the mentioned goals suffice, so I'm going to move to close.

@rfcbot fcp close

aturon:
Ping @BurntSushi @Kimundi @brson @sfackler

alexcrichton:
Ok looks like not much more discussion happened during FCP, so I'm going to close this. Thanks again though for the RFC @clarcharr!

PR #1983, opened 2017-04-27
Amend RFC 1242 to require an RFC for deprecation of crates from the nursery

jethrogb:
[Rendered](https://github.com/jethrogb/rfcs/blob/nursery-deprecation/text/0000-nursery-deprecation.md)

### DIFF
--- a/text/0000-nursery-deprecation.md
+++ b/text/0000-nursery-deprecation.md
@@ -66,8 +66,8 @@ not a major commitment.
 
 Acceptance into the nursery could be hindered if it is believed it could be 
 hard to reverse course later due to the required RFC being percieved as an 
-obstacle. On the other hand, RFCs with broad concensus do not generally impose 
-a large procedural burden, and if there is no concensus it might be too early 
+obstacle. On the other hand, RFCs with broad consensus do not generally impose 
+a large procedural burden, and if there is no consensus it might be too early 
 to deprecate a nursery crate anyway.
 
 # Alternatives
--- a/text/0000-nursery-deprecation.md
+++ b/text/0000-nursery-deprecation.md
@@ -1,0 +1,82 @@
+- Feature Name: N/A
+- Start Date: 2017-04-26
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Amend [RFC 1242] to require an RFC for deprecation of crates from the 
+rust-lang-nursery.
+
+[RFC 1242]: https://github.com/rust-lang/rfcs/blob/master/text/1242-rust-lang-crates.md
+
+# Motivation
+[motivation]: #motivation
+
+There are currently very ubiquitous crates in the nursery that are being used 
+by lots and lots of people, as evidenced by the crates.io download numbers (for 
+lack of a better popularity metric):
+
+| Nursery crate | Downloads |
+| ------------- | --------- |
+| bitflags      |    3,156k |
+| rand          |    2,615k |
+| log           |    2,417k |
+| lazy-static   |    2,108k |
+| tempdir       |      934k |
+| uuid          |      759k |
+| glob          |      467k |
+| net2          |      452k |
+| getopts       |      452k |
+| rustfmt       |       80k |
+| simd          |       14k |
+
+(numbers as of 2017-04-26)
+
+[RFC 1242] currently specifies that
+
+> The libs subteam can deprecate nursery crates at any time
+
+The libs team can of course be trusted to be judicious in making such 
+decisions. However, considering that many of the nursery crates are depended on 
+by big fractions of the Rust ecosystem, suddenly deprecating things without 
+public discussion seems contrary to Rust's philosophy of stability and 
+community participation. Involving the Rust community at large in these 
+decisions offers the benefits of the RFC process such as increased visibility, 
+differing viewpoints, and transparency.
+
+# Detailed design
+[design]: #detailed-design
+
+The exact amendment is included as a change to the RFC in this PR.
+[View the amended text](1242-rust-lang-crates.md).
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+N/A
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Requiring an RFC for deprecation might impose an undue burden on the library 
+subteam in terms of crate maintenance. However, as [RFC 1242] states, this is
+not a major commitment.
+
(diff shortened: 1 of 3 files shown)


burdges:
I think this generally comes under the [1.0-level crates objective](https://github.com/rust-lang/rust-roadmap/issues/11) of the [2017 roadmap](https://blog.rust-lang.org/2017/02/06/roadmap.html) or maybe the high quality crates goal rooted in the [Rust platform discussion](https://internals.rust-lang.org/t/proposal-the-rust-platform/3745/).

An approach might be to open a well informed "What do we want for 1.0 discussion?" issue on each of the nursery crates named in this RFC, and link them from the 1.0-level crates objective issue.  Is a `log` vs `slog` discussion needed?  I know `rand` has numerous open issues and pull requests, but maybe the existing functionality could be declared "stable enough", even if not 1.0.  etc.  Also, I think declaring `rustfmt` to be 1.0 sounds really hard, but it has an RFC process and zero risk of depreciation, so maybe just skip that one. 

As an aside, I'd suspect the name "nursery" would be inappropriate for crates the language developers view as stable and ship with the compiler.  There is no breaking change in merely moving a crate from the nursery to another github url though.

text/0000-nursery-deprecation.md
```diff
@@ -0,0 +1,82 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Requiring an RFC for deprecation might impose an undue burden on the library 
+subteam in terms of crate maintenance. However, as [RFC 1242] states, this is
+not a major commitment.
+
+Acceptance into the nursery could be hindered if it is believed it could be 
+hard to reverse course later due to the required RFC being percieved as an 
+obstacle. On the other hand, RFCs with broad concensus do not generally impose
```
ranma42: typo: s/concensus/consensus/

text/0000-nursery-deprecation.md
```diff
@@ -0,0 +1,82 @@
...
+[drawbacks]: #drawbacks
+
+Requiring an RFC for deprecation might impose an undue burden on the library 
+subteam in terms of crate maintenance. However, as [RFC 1242] states, this is
+not a major commitment.
+
+Acceptance into the nursery could be hindered if it is believed it could be 
+hard to reverse course later due to the required RFC being percieved as an 
+obstacle. On the other hand, RFCs with broad concensus do not generally impose 
+a large procedural burden, and if there is no concensus it might be too early
```
ranma42: typo: s/concensus/consensus/

jethrogb:
> I know `rand` has numerous open issues and pull requests, but maybe the existing functionality could be declared "stable enough", even if not 1.0. etc.

I've been hearing rumors about a "big rewrite/redesign" for quite a while now, but nothing seems to be happening on that front.

aturon:
cc @rust-lang/libs: TL;DR, require an RFC to deprecate nursery crates. I agree with the rationale here; since the whole point of rust-lang crates is to put these libraries under full community governance, that should certainly apply to deprecation too.

@rfcbot fcp merge

sfackler:
What happens if an RFC to deprecate a nursery crate is rejected? Does the libs team have some kind of SLA on maintenance of a library none of them want to maintain?

sfackler:
@jethrogb a rand overhaul is blocked on someone with the time and interest stepping up and owning the design and implementation.

aturon:
@sfackler Keep in mind that it's the libs team that would make the decision on such RFCs. So it's really more about consulting the broader community and opening up a discussion, ensuring that it's not a "libs team against the world" type situation. I would imagine these RFCs would be pretty short, but that in at least some cases discussion would lead to discovery of people who do want to maintain the library (in the case that it's not being deprecated in favor of something else).

sfackler:
So the only real change here is adding a ~month to the time it takes from our decision to do this to actually making the move? I don't really see a world in which there's some huge outcry that we have to continue maintaining something.

aturon:
@rfcbot fcp cancel

We discussed this RFC in libs team triage, and it has led us to want to re-think the rust-lang-* orgs altogether (in light of the [Blitz](https://blog.rust-lang.org/2017/05/05/libz-blitz.html)). @sfackler and I are going to work together to write up our thoughts as a counter-proposal.

aturon:
I'm going to propose to close this RFC, for two reasons:

- Almost all of the crates listed are undergoing work toward 1.0 as part of the libz blitz or other efforts this year.

- The libs team no longer plans to introduce new crates through the nursery process, instead focusing on supporting the decentralized ecosystem.

In any case, any move to deprecate any of the existing crates will certainly involve public discussion.

@rfcbot fcp close

jethrogb:
> The libs team no longer plans to introduce new crates through the nursery process, instead focusing on supporting the decentralized ecosystem.

Has there been any discussion on this change? I understand that the libs team is very busy and the community can't just expect the libs team to do everything we throw at them. Therefore it's entirely reasonable that the libs team changes the way they operate if they think that will balance their workload. I do think it's important that if the libs team wants to make those changes that there is some public discussion about this, so that the community can see if there's any slack left that needs to be picked up elsewhere.

One main question I have is: what will be the process to move things into `std`? 

Please elaborate https://twitter.com/fitzgen/status/889980459953692674 with regards to your earlier statement.

> In any case, any move to deprecate any of the existing crates will certainly involve public discussion.

This is basically all that this RFC proposes to change, so why not just merge?

dtolnay:
Is there a counter-proposal (https://github.com/rust-lang/rfcs/pull/1983#issuecomment-300295429) yet?

I would be inclined to accept the RFC. So far the main argument against sounds like "we will never again deprecate anything from the nursery" which I don't find compelling.

> So the only real change here is adding a ~month to the time it takes from our decision to do this to actually making the move?

I don't expect a practical difference between how people treat a deprecation announcement without this RFC, vs a deprecation RFC per this RFC. People who care will take either as an indication to look for an alternative library—or raise concerns that a suitable alternative does not exist, which the RFC provides a place to do.

@rfcbot concern let's do it

aturon:
@rfcbot fcp cancel

aturon:
Alright, after talking last week in the libs team, we're just gonna go forward with this.

Several of the nursery crates are being tackled as part of the Blitz, and are expected to head toward rust-lang proper. The others, we'll probably propose deprecation for in a single RFC.

@rfcbot fcp merge

aturon:
(I ticked everyone's boxes based on the discussion in the meeting.)

aturon:
This RFC has now been merged. Thanks @jethrogb!

PR #1985, opened 2017-04-28
Tiered browser support policy for Rust's web content

carols10cents:
[Rendered!](https://github.com/integer32llc/rfcs/blob/tiered-browser-support/text/0000-tiered-browser-support.md)

### DIFF
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -122,12 +122,12 @@ last 1 UCAndroid version
 
 Support:
 
-- We add automated testing for each of these as much as possible (and work on
-  adding automated testing to those web contents that aren't currently tested,
-  such as rustdoc output).
+- We add automated testing of functionality in a variety of browsers through a
+  service such as [BrowserStack][] for each of these as much as possible (and
+  work on adding this type of automated testing to those web contents that
+  aren't currently tested, such as rustdoc output).
 - Bugs affecting the functionality of the sites in these browsers are
   prioritized highly.
-
 
 ### Unsupported browsers
 
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -128,35 +128,8 @@ Support:
 - Bugs affecting the functionality of the sites in these browsers are
   prioritized highly.
 
-### Tier 2
 
-Goal: Have some support of our web content, without additional time, effort, or
-monetary cost, for about 90% of users.
-
-Browsers:
-
-- Chrome up to 3 releases back and 1 release ahead of the current stable
-- Firefox up to 2 releases back and 2 releases ahead of the current stable, one
-  previous ESR
-- Safari up to 2 releases back from the current stable
-- UC Browser 1 previous stable release
-- Edge current stable and 2 previous releases
-- IE 11 until its EOL
-- Android Webview current stable and 1 previous release
-- Opera latest stable
-
-Support:
-
-- No automated testing for these.
-- Bug reports for these browsers are left open but given a low priority, and
-  closed if they are still open when a browser is old enough to drop out of
-  this tier.
-- Pull requests to fix functionality for these browsers are accepted.
-
-### Tier 9999
-
-Otherwise known as "deliberately not supporting". Not tier 3, to avoid
-conflating with the Rustc tier 3 that still has some support.
+### Unsupported browsers
 
 Goal: Avoid spending large amounts of time and code complexity debugging and
 hacking around quirks in older or more obscure browsers.
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -161,6 +161,12 @@ policy we end up with:
   through color also conveyed through an icon or text.
 - Follow best practices for making content usable from mobile devices with a
   variety of screen sizes.
+- Render content without requiring JavaScript (especially on
+  [crates.io][noscript]). Additional functionality beyond reading (ex: search,
+  follow/unfollow crate) may require JavaScript, but we will attempt to use
+  links and forms for progressive enhancement as much as possible.
+
+[noscript]: https://github.com/rust-lang/crates.io/issues/204
 
 Please comment if you think any of these should **not** be assumed, but rest
 assured it is not the intent of this RFC to get rid of these kinds of support.
@@ -267,8 +273,5 @@ reproduce it.
 [unresolved]: #unresolved-questions
 
 - Am I missing any official web content that this policy should apply to?
-- [Should we have web content that can function without JavaScript?][noscript]
 - Is it possible to add browser tests to rustdoc or would that just make the
   current situation of long, flaky rustc builds worse?
-
-[noscript]: https://github.com/rust-lang/crates.io/issues/204
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -98,17 +98,27 @@ effective way, the browser support tiers would be defined as:
 
 [actual usage metrics]: #google-analytics-browser-usage-stats
 
-### Tier 1
+Browsers are listed in [browserslist][] format.
+
+[browserslist]: https://github.com/ai/browserslist
+
+### Supported browsers
 
 Goal: Ensure functionality of our web content for 80% of users.
 
 Browsers:
 
-- Chrome current stable and one previous release
-- Firefox current stable and the latest ESR
-- Safari current stable, one previous minor release, and the current version on
-  iOS
-- UC Browser current stable
+```
+last 2 Chrome versions
+last 1 Firefox version
+Firefox ESR
+last 1 Safari version
+last 1 iOS version
+last 1 Edge version
+last 1 UCAndroid version
+```
+
+[On browserl.ist](http://browserl.ist/?q=last+2+Chrome+versions%2C+last+1+Firefox+version%2C+Firefox+ESR%2C+last+1+Safari+version%2C+last+1+iOS+version%2C+last+1+Edge+version%2C+last+1+UCAndroid+version)
 
 Support:
 
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -155,6 +155,8 @@ policy we end up with:
 - Follow best practices for accessibilty, fix bug reports from blind users,
   reach out to blind users in the community about how the accessibility of the
   web content could be improved.
+  - This would include supporting lynx/links as these are sometimes used with
+    screen readers.
 - Follow best practices for colorblindness, such as have information conveyed
   through color also conveyed through an icon or text.
 - Follow best practices for making content usable from mobile devices with a
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -95,6 +95,8 @@ almost-official Rust web content:
 Based on [actual usage metrics][] and with a goal of spending our time in an
 effective way, the browser support tiers would be defined as:
 
+[actual usage metrics]: #google-analytics-browser-usage-stats
+
 ### Tier 1
 
 Goal: Ensure functionality of our web content for 80% of users.
@@ -185,7 +187,6 @@ but our audience (developers) isn't the same as the general public.
 [CanIUse.com]: http://caniuse.com/usage-table
 
 ## Google analytics browser usage stats
-[actual usage metrics]: #actual-usage-metrics
 
 We have Google Analytics on crates.io and on rust-lang.org. The entire data set
 of the usage stats by browser, browser verison, and OS are available [in this
--- a/text/0000-tiered-browser-support.md
+++ b/text/0000-tiered-browser-support.md
@@ -1,0 +1,287 @@
+- Feature Name: tiered_browser_support
+- Start Date: 2017-04-25
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Official web content produced by the Rust teams for consumption by Rust users
+should work in the majority of browsers that Rust users are visiting these
+sites in. The Rust compiler only supports [a finite number of targets][forge],
+with varying degrees of support, due to the limits on time, expertise, and
+testing resources. Similarly, we don't have enough time, expertise and testing
+resources to be sure that our web content works in every version of every
+browser. We should have a list of browsers and versions in various tiers of
+support.
+
+[forge]: https://forge.rust-lang.org/platform-support.html
+
+# Motivation
+[motivation]: #motivation
+
+[This pull request to remove JQuery from rustdoc's output][jquery-pr] had
+discussion about what we could and could not do because of browser support.
+This is a discussion we haven't yet had as a community.
+
+[jquery-pr]: https://github.com/rust-lang/rust/pull/41307
+
+Crates.io doesn't display correctly in browsers without support for flexbox,
+such as [Windows Phone 8.1][win-phone], a device that is no longer supported. I
+made the decision that it wasn't worth it for the community to spend time
+fixing this issue, did I make the correct tradeoff for the community?
+
+[win-phone]: https://github.com/rust-lang/crates.io/issues/56
+
+Supporting all versions of all browsers with the same behavior is impossible
+with the small number of people who work on Rust's web content. Crates.io is
+not currently doing any cross-browser testing; there are some JavaScript tests
+of the UI that run in [PhantomJS][], a headless WebKit. Since we're not
+testing, we don't actually know what our current web support even is, except
+for when we get bug reports from users.
+
+[PhantomJS]: http://phantomjs.org/
+
+In order to fully test on all browsers to be sure of our support, we would
+either need to have all the devices, operating systems, browsers, and versions
+available and people with the time and inclination to do manual testing on all
+of these, or we would need to be running automated tests on something like
+[BrowserStack][]. BrowserStack does appear to have a free plan for open source
+projects, but it's unclear how many parallel tests the open source plan would
+give us, and we'd at least be spending time waiting for test results on the
+various stacks. [BrowserStack also doesn't support every platform][bs-support],
+Linux on the desktop being a notable section of our userbase missing from their
+platforms.
+
+[BrowserStack]: https://www.browserstack.com/pricing
+[bs-support]: https://www.browserstack.com/support
+
+# Detailed design
+[design]: #detailed-design
+
+## Rust web content
+
+Officially produced web content includes:
+
+- rust-lang.org
+- blog.rust-lang.org
+- play.rust-lang.org
+- crates.io
+- Rustdoc output
+
+Explicitly not included:
+
+- Content for people working on Rust itself, such as:
+  - [The Rust Forge][]
+  - [rusty-dash][]
+
+[The Rust Forge]: https://forge.rust-lang.org/
+[rusty-dash]: https://rusty-dash.com/
+
+Things that are not really under our control but are used for official or
+almost-official Rust web content:
+
+- GitHub
+- docs.rs
+- Discourse (used for [urlo][] and [irlo][])
+- [mdBook][] output (used for the books and other documentation)
+
+[urlo]: https://users.rust-lang.org/
+[irlo]: https://internals.rust-lang.org/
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Seems like it's a bit short. Only newer versions (to put it simply) is a bit unrealistic for a lot of companies, which means we potentially exclude any new project in rust for these companies if they can't have access to documentation.

carols10cents:
> Seems like it's a bit short. Only newer versions (to put it simply) is a bit unrealistic for a lot of companies, which means we potentially exclude any new project in rust for these companies if they can't have access to documentation.

Yup, that's the downside. Which is why I think it's interesting that GitHub (who also sells GitHub Enterprise into large companies) only supports the current version of Chrome, Firefox, Safari, and Edge.

GuillaumeGomez:
It's up to debate. I globally agree on the rest, just have a problem on this.

carols10cents:
> It's up to debate. I globally agree on the rest, just have a problem on this.

For the purposes of debating in one place (we've been discussing this in IRC), could you post the versions of browsers and which tiers of support that you'd propose that we use instead?

whitequark:
crates.io is also completely unusable without JavaScript, which always struck me as a strange decision.

carols10cents:
> crates.io is also completely unusable without JavaScript, which always struck me as a strange decision.

Yup, I mention that under "unresolved questions".

carols10cents:
Hi @kennytm, I see your "confused" reaction, what's confusing?

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+We should call this "Rust Browser Support", and we should have the tiers listed
+on the Rust Forge in a similar way to the tiers of Rust platforms supported.
+
+We should link to the tiered browser support page from places where Rust web
+content is developed and on the [Rust FAQ][].
+
+[Rust FAQ]: https://www.rust-lang.org/en-US/faq.html
+
+# Drawbacks
```
nagisa: Should mention that it is *unknown* what portion of users are unaccounted for by the tracking we use due to e.g. having disabled javascript, using privacy addons etc.

carols10cents: I do mention that under the analytics section, do you think it should be repeated here?

aturon:
I'm assigning to the core team, since this is a cross-cutting policy issue.

kennytm:
@carols10cents Hi sorry I'm just confused why this even needs to go through the RFC process, seems very bureaucratic for me 😄. 

I think making sure it works on the current Chrome (Desktop + Mobile), Firefox, Safari (Desktop + Mobile) and IE 11, and avoid abusing fancy stuff (graceful degradation), should be enough to be portable in most places.

dlight:
Regarding crates.io needing Javascript: I'm uncomfortable with any site that needs Javascript merely to display static data. This also happens with Google Groups and a lot of other sites.

It's okay to require Javascript for user interaction, even though it's preferably to have a fallback like plain HTML forms, if at all possible.

What happened to progressive enhancement?

GuillaumeGomez:
So for the web browser versions I was speaking of, it's as follow:

 * firefox 24
 * chrome 30
 * IE 11

steveklabnik:
@kennytm:

> @carols10cents Hi sorry I'm just confused why this even needs to go through the RFC process, seems very bureaucratic for me 😄.

So, stuff like this might not *have* to, but the RFC process is how we build community consensus; with something that affects everyone broadly like this, it's better safe than sorry, basically.

steveklabnik:
@dlight @whitequark see https://github.com/rust-lang/crates.io/issues/204

bluetech:
A small suggestion: it might be nice to specify the tiers in terms of [browserslist](https://github.com/ai/browserslist) queries.

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+| UC Browser      | 11              | 530      | 1.02          | 77.40        |
+| Chrome          | 58              | 520      | 1.00          | 78.40        |
+| Safari (in-app) | (not set) (ios) | 500      | 0.97          | 79.37        |
+| Firefox         | 54              | 472      | 0.91          | 80.28        |
+
+Interesting to note: Firefox 45 is the latest
+[ESR](https://www.mozilla.org/en-US/firefox/organizations/all/) (Firefox 52
+will also be an ESR but it was just released). Firefox 52 was the current major
+version for most of this past month; I'm guessing the early adopters of 53 and
+54 are likely Mozilla employees.
```
nrc: iirc, Firefox nightly has ~1 million users, so more than just employees, but still kind of a niche interest

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+| Safari          | 10.1 (macos)    | 1,592    | 3.08          | 67.74        |
+| Safari          | 10 (ios)        | 1,437    | 2.78          | 70.52        |
+| Safari          | 10.0.3 (macos)  | 851      | 1.64          | 72.16        |
+| Firefox         | 53              | 767      | 1.48          | 73.65        |
+| Chrome          | 55              | 717      | 1.39          | 75.03        |
+| Firefox         | 45              | 693      | 1.34          | 76.37        |
+| UC Browser      | 11              | 530      | 1.02          | 77.40        |
+| Chrome          | 58              | 520      | 1.00          | 78.40        |
+| Safari (in-app) | (not set) (ios) | 500      | 0.97          | 79.37        |
+| Firefox         | 54              | 472      | 0.91          | 80.28        |
```
nrc: Perhaps it is worth looking at some more generic measure of market share as well - we want to optimise for the users we want to have, not just the users we already have

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+Goal: Ensure functionality of our web content for 80% of users.
+
+Browsers:
+
+- Chrome current stable and one previous release
+- Firefox current stable and the latest ESR
+- Safari current stable, one previous minor release, and the current version on
+  iOS
+- UC Browser current stable
+
```
nrc: I think we should include Edge on this list - MS has big market share and we want to expand on Windows. Currently this feels like more of treating windows users second class.

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+- Android Webview current stable and 1 previous release
+- Opera latest stable
+
+Support:
+
+- No automated testing for these.
+- Bug reports for these browsers are left open but given a low priority, and
+  closed if they are still open when a browser is old enough to drop out of
+  this tier.
+- Pull requests to fix functionality for these browsers are accepted.
```
nrc: IMO we should not have this tier - I don't see a meaningful difference between this and tier 9999

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Orthogonal but related non-proposals
+
+The following principles are assumptions I'm making that we currently follow
+and that we should continue to strive for, no matter what browser support
+policy we end up with:
+
+- Follow best practices for accessibilty, fix bug reports from blind users,
+  reach out to blind users in the community about how the accessibility of the
+  web content could be improved.
```
nrc: + 100 - I think making screen reader support (and other accessibility issues) a tier 1 (or possibly tier 1.5) issue is more important than supporting any particular browser.

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,289 @@
...
+- Follow best practices for colorblindness, such as have information conveyed
+  through color also conveyed through an icon or text.
+- Follow best practices for making content usable from mobile devices with a
+  variety of screen sizes.
```
nrc: This seems like a big thing to me - supporting a browser is one thing, supporting it on every possible form factor is another. We should probably be explicit about this - while I think it is worth making r-l.o and the blog mobile friendly, we shouldn't commit to making rustdoc 100% mobile friendly for now (I'd like this as a goal, but not as a promise right now).

carols10cents:
> small suggestion: it might be nice to specify the tiers in terms of browserslist queries.

Neat! I didn't know that existed! I will update soon :)

text/0000-tiered-browser-support.md
carols10cents: I'd be up for adding Edge to tier 1, especially if we drop tier 2.

text/0000-tiered-browser-support.md
carols10cents: If we drop tier 2, I might change the labels to be "Supported browsers" and "Unsupported browsers" then. I'd like to see what people think though, I'd be fine with a simpler policy but I added tier 2 to add some definition around whether we'd accept patches for a particular version or not.

text/0000-tiered-browser-support.md
carols10cents: Most of these are goals rather than promises at the moment-- such as automated testing of rustdoc :)

text/0000-tiered-browser-support.md
carols10cents: [Caniuse](http://caniuse.com/usage-table) has global browser usage stats, but I still think there's going to be large differences between the general population and a population of developers. I'm not sure where to find browser stats for developers rather than browser stats for rust developers; I'll do some investigating but if anyone knows of any resources I'd love to hear about them!

clarcharr:
Adding a bit from the original thread about removing the `jQuery` dependency, which I brought up: my main justification was that `jQuery` is a very large library that shouldn't be included statically with every generated rustdoc. It's around 2 MiB minimised, and because it's copied every time that rustdoc is rendered, it can potentially take up a lot of space if you have a lot of crates that use it. This is especially important because rustdoc doesn't actually make use of jQuery a lot, and almost all of the stuff in it could be replaced with a modern equivalent (as @GuillaumeGomez did).

I agree that there becomes a point where you have to stop caring about really old implementations (does anyone remember how IE6 couldn't even display transparent PNGs?) but also I think that documentation size should also be taken into account when making decisions on what platforms to support.

Also huge thumbs up to @carols10cents for making this RFC because while I believe that this is a decision that could easily be done without an RFC, I do really appreciate the community discussion. :)

(This isn't actually going against anything said here, I'm just adding my probably-more-than-two-cents)

o01eg:
What about console text-mode browsers like lynx, links?

GuillaumeGomez:
That's not even part of the supported browsers in the RFC.

mgattozzi:
I was at RustFest this pass weekend and there was a blind Rust Developer there and he uses lynx/links in order to access many of the sites. I think from an accessibility point of view it might be a good idea to support them, even if it's a bit harder compared to the rest of them.

clarcharr:
Something that I think is also important to comment on is screen reader support.

clarcharr:
I think that screen reader support is something we should definitely comment on at the very least. Ideally, all of Rust's content would be accessible as possible.

carols10cents:
> I was at RustFest this pass weekend and there was a blind Rust Developer there and he uses lynx/links in order to access many of the sites. I think from an accessibility point of view it might be a good idea to support them, even if it's a bit harder compared to the rest of them.

---

> I think that screen reader support is something we should definitely comment on at the very least. Ideally, all of Rust's content would be accessible as possible.

---

Screen reader support is addressed in the [Orthogonal but related non-proposals](https://github.com/integer32llc/rfcs/blob/tiered-browser-support/text/0000-tiered-browser-support.md#orthogonal-but-related-non-proposals) section of the RFC. TL;DR screen reader support/accessibility is an important thing we should definitely be working on. If lynx/links is one of the ways people do that, we should support those browsers, I didn't know about that and I will edit the RFC to add lynx/links under this section.

carols10cents:
Does anyone have any thoughts on @nrc's suggestion that [tier 2 be removed](https://github.com/rust-lang/rfcs/pull/1985#discussion_r114086923)? 👍 / 👎 on this comment if you don't have any thoughts beyond general in favor/not in favor?

carols10cents:
I've gone ahead and removed tier 2 so yinz can see what this RFC would look like without it, but I could be convinced to put it back, so please let me know.

clarcharr:
Adding onto the comments about screen readers: I think this is something that goes a bit beyond Rust's general web content, and more about the general goals behind Rust's documentation, but I think that these posters provide a pretty nice list of basic accessibility things and it'd be nice if we somehow integrated them in with our general goals for accessibility in Rust docs:

https://github.com/UKHomeOffice/posters

(obviously some of these things go beyond web content but they are useful in general)

I mean, most of these are things that we're already doing (like trying to make documentation/etc. as understandable/approachable as possible) but I think that it'd be cool if there were some sort of official list of accessibility criteria the Rust project is trying to uphold, so that it can be referenced when changes are made.

carols10cents:
@rfcbot fcp merge

text/0000-tiered-browser-support.md
```diff
@@ -0,0 +1,277 @@
...
+last 1 UCAndroid version
+```
+
+[On browserl.ist](http://browserl.ist/?q=last+2+Chrome+versions%2C+last+1+Firefox+version%2C+Firefox+ESR%2C+last+1+Safari+version%2C+last+1+iOS+version%2C+last+1+Edge+version%2C+last+1+UCAndroid+version)
+
+Support:
+
+- We add automated testing for each of these as much as possible (and work on
+  adding automated testing to those web contents that aren't currently tested,
+  such as rustdoc output).
```
GuillaumeGomez: Rustdoc output is tested.

carols10cents: In what way is rustdoc output tested? What I mean here is running the javascript parts in multiple browsers via BrowserStack to ensure browser compatibility, and I didn't think that's something currently being done.

GuillaumeGomez: Ok, I see your point this time. Then it might be worth it to specify it. So that we all agree on what testing the output is. :)

wesleywiser:
I think that we should strongly consider supporting some version of Internet Explorer. While Edge is the default on Windows > 8, Edge is not even available on Windows 7 and, according to some sites, Windows 7 is still the most popular Windows version [StatCounter](http://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide).

I also think that the "static" web sites such as rust-lang.org and blog.rust-lang.org should not be a part of this policy. For many people, these sites are their first introduction to Rust and so they should be available in as many browsers as possible. Furthermore, since the content of these sites is largely static, it should be much easier to maintain them in older browsers than other pieces of web content like Rustdoc output.

dbrgn:
> Edge is not even available on Windows 7

But Chrome and Firefox are.

> I also think that the "static" web sites such as rust-lang.org and blog.rust-lang.org should not be a part of this policy.

I agree on that part though.

GuillaumeGomez:
> But Chrome and Firefox are.

All users don't necessarily use them. They could prefer IE (I don't judge).

carols10cents:
In my experience developing for IE, and given the browser stats generated by our audience (developers, not the general population), the amount of effort to support IE is not worth the reward.

I agree that it'd be nice to have the homepage and blog render in as many browsers as possible, and since their layout doesn't change, this might be a one-time cost of testing. I'm trying to define where we should spend our finite amount of time on testing with this RFC, however, so we need to draw the line somewhere. Note that the status quo right now is no testing.

wesleywiser:
I agree, the line must be drawn somewhere but I think it would unfortunate for it not to include *some* version of Internet Explorer. It seems to me that it would be easiest to support IE 11. The linked pull request in this RFC [even mentions](https://github.com/rust-lang/rust/pull/41307#issuecomment-294457125) that IE 11 would support the native browser APIs used.

According to [html5test.com](https://html5test.com/compare/browser/ie-11/safari-10.2.html), supporting IE 11 doesn't seem much harder to me than supporting the latest version of Safari. Most of the differences seem to me to be things that we don't need. The biggest difference is ES6 support, but we aren't currently using ES6 anyway.

GuillaumeGomez:
Also, we're talking docs rendering in here mainly. We don't need any fancy CSS feature, just a nice rendering, easy to read and interact with. This is why I think we're focusing way too much on too recent web browsers.

carols10cents:
> Also, we're talking docs rendering in here mainly.

Actually, I'm trying to talk about *testing resources* in here mainly.

wesleywiser:
If we don't have the resources to test IE 11, then I can accept that. However, the RFC as currently written goes further than simply stating that unsupported browsers will be untested; it says that any bugs for unsupported browsers will be WONTFIX'd and closed. While it's not explicitly stated, I think a safe assumption is that a PR which introduces known regressions in an unsupported browser will still be accepted. The net result is that our web content will stop working in unsupported browsers very quickly.

I would much prefer something like the original version of this RFC which had different tiers of support:

- "Gold" tier browsers
  - Modern browsers (the "Supported Browsers" list in this RFC)
  - Tested
  - Rust's web content is designed for these browsers
- "Bronze" tier browsers
  - Older browsers
    - IE 11
    - Perhaps older versions of the Android browser?
    - Perhaps lynx?
  - Not tested, we rely on bug reports from users
  - PRs are not allowed to cause known regressions to these browsers
  - Rust's web content should be functional but may have a degraded experience or small visual glitches
- Unsupported browsers
  - All other browsers
  - Bug reports will be closed as WONTFIX
  - PRs to fix issues can be accepted provided they don't impose a significant maintenance burden

carols10cents:
I'm interested to hear what the rest of @rust-lang/core thinks; [this commit](https://github.com/rust-lang/rfcs/pull/1985/commits/9ebbc6d1010378e0f3586876c89f82630f72e9de) shows the diff of the two versions that have been proposed.

GuillaumeGomez:
> Actually, I'm trying to talk about testing resources in here mainly.

That's not as I understood it. :-/

alexcrichton:
I'm personally ok with the RFC as-written, without a second tier of support. Would we still accept PRs to increase compatibility? If we did then I imagine that the process of having and not having a second tier would look very similar, modulo whether bugs are left open.

aturon:
@carols10cents I could go either way on this question, but like @nrc I suspect that in practice they'll work out roughly the same. That is, closing as WONTFIX in this case means we're not committing to fixing the issue, but would be open to a PR that does.

I think I prefer to start with the RFC as it stands now, and see how it shakes out in practice (i.e., how often we see such issues).

I think the most important step here is getting *any* testing and assurance. I think that alone is going to make our web presence of higher quality across the board.

steveklabnik:
> @carols10cents I could go either way on this question, but like @nrc I suspect that in practice they'll work out roughly the same. That is, closing as WONTFIX in this case means we're not committing to fixing the issue, but would be open to a PR that does.

Seconded.

> I think the most important step here is getting any testing and assurance. I think that alone is going to make our web presence of higher quality across the board.

Yup, exactly: we can always add more support later. I think of this RFC as being going from nothing to something at all, not "what is the perfect set of things to test." If it turns out a lot of people need something we don't support, it's easy enough to modify the guarantees later, and by the same token, I'm sure we'll eventually drop support for browsers after a few years.

carols10cents:
>> Actually, I'm trying to talk about testing resources in here mainly.
> 
> That's not as I understood it. :-/

Can you suggest improvements to the summary to make it clearer?

carols10cents:
@alexcrichton 

> Would we still accept PRs to increase compatibility? If we did then I imagine that the process of having and not having a second tier would look very similar, modulo whether bugs are left open.

It would depend on the complexity/maintenance burden of what the PRs were adding :-/

GuillaumeGomez:
The feature name is literally : `tiered_browser_support`. It's not much about testing but about web browsers support, which aren't the same thing at all. ;)

carols10cents:
> The feature name is literally : tiered_browser_support. It's not much about testing but about web browsers support, which aren't the same thing at all. ;)

What about the summary is unclear though?

steveklabnik:
@GuillaumeGomez I think you're _really_ nitpicking here; "support" in this context means "things we guarantee support for", and the mechanism by which we do that is through tests.

buosseph:
This may be nitpicky, but I feel Chrome for Android should be included in the browserlist. While I'm not aware of any major differences between the desktop and Android versions of Chrome, it would be good to be explicit and catch any cases where there are differences.

aturon:
Huzzah! This RFC has now been merged. Follow up, e.g. regarding Chrome for Android, should go on the [tracking issue](https://github.com/rust-lang/rust/issues/43035).

PR #2002, opened 2017-05-18
Broken links for core team

AndyGauge:
The mozilla/rust repository is a redirect to rust-lang and rust-lang has no wiki.  The best link I could find for the core team was on www.rust-lang.org


steveklabnik:
Yup, that's the right place; thank you!

PR #2005, opened 2017-05-22
Match Ergonomics Using Default Binding Modes

cramertj:
[Rendered.](https://github.com/cramertj/rfcs/blob/match-ergo-take-2/text/0000-match-ergonomics.md)

### DIFF
--- a/text/0000-match-ergonomics.md
+++ b/text/0000-match-ergonomics.md
@@ -1,0 +1,454 @@
+- Feature Name: pattern-binding-modes
+- Start Date: 2016-08-12
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Better ergonomics for pattern-matching on references.
+
+Currently:
+
+```
+match *x {
+    Foo(ref x) => { ... }
+    Bar(ref mut y, z) => { ... }
+}
+```
+
+Proposed:
+
+```
+match x {
+    Foo(x) => { ... }
+    Bar(y, z) => { ... }
+}
+```
+
+This is accomplished through automatic dereferencing and the introduction of
+default binding modes.
+
+# Motivation
+[motivation]: #motivation
+
+Rust is usually strict when distinguishing between value and reference types. In
+particular, distinguishing borrowed and owned data. However, there is often a
+trade-off between [explicit-ness and ergonomics](https://blog.rust-lang.org/2017/03/02/lang-ergonomics.html),
+and Rust errs on the side of ergonomics in some carefully selected places.
+Notably when using the dot operator to call methods and access fields, and when
+declaring closures.
+
+The match expression is an extremely common expression and arguably, the most
+important control flow mechanism in Rust. Borrowed data is probably the most
+common form in the language. However, using match expressions and borrowed data
+together can be frustrating: getting the correct combination of `*`, `&`, and
+`ref` to satisfy the type and borrow checkers is a common problem, and one which
+is often encountered early by Rust beginners. It is especially frustrating since
+it seems that the compiler can guess what is needed but gives you error messages
+instead of helping.
+
+For example, consider the following program:
+
+```
+enum E { Foo(...), Bar }
+
+fn f(e: &E) {
+    match e { ... }
+}
+
+```
+
+It is clear what we want to do here - we want to check which variant `e` is a
+reference to. Annoyingly, we have two valid choices:
+
+```
+match e {
+    &E::Foo(...) => { ... }
+    &E::Bar => { ... }
+}
+```
+
+and
+
+```
+match *e {
+    E::Foo(...) => { ... }
+    E::Bar => { ... }
+}
+```
+
+The former is more obvious, but requires more noisey syntax (an `&` on every
+arm). The latter can appear a bit magical to newcomers - the type checker treats
+`*e` as a value, but the borrow checker treats the data as borrowed for the
+duration of the match. It also does not work with nested types, `match (*e,)
+...` for example is not allowed.
+
+In either case if we further bind variables, we must ensure that we do not
+attempt to move data, e.g.,
+
+```
+match *e {
+    E::Foo(x) => { ... }
+    E::Bar => { ... }
+}
+```
+
+If the type of `x` does not have the `Copy` bound, then this will give a borrow
+check error. We must use the `ref` keyword to take a reference: `E::Foo(ref x)`
+(or `&E::Foo(ref x)` if we match `e` rather than `*e`).
+
+The `ref` keyword is a pain for Rust beginners, and a bit of a wart for everyone
+else. It violates the rule of patterns matching declarations, it is not found
+anywhere outside of patterns, and it is often confused with `&`. (See for
+example, https://github.com/rust-lang/rust-by-example/issues/390).
+
+Match expressions are an area where programmers often end up playing 'type
+Tetris': adding operators until the compiler stops complaining, without
+understanding the underlying issues. This serves little benefit - we can make
+match expressions much more ergonomic without sacrificing safety or readability.
+
+Match ergonomics has been highlighted as an area for improvement in 2017:
+[internals thread](https://internals.rust-lang.org/t/roadmap-2017-productivity-learning-curve-and-expressiveness/4097)
+and [Rustconf keynote](https://www.youtube.com/watch?v=pTQxHIzGqFI&list=PLE7tQUdRKcybLShxegjn0xyTTDJeYwEkI&index=1).
+
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC is a refinement of
+[the match ergonomics RFC](https://github.com/rust-lang/rfcs/pull/1944). Rather
+than using auto-deref and auto-referencing, this RFC introduces _default binding
+modes_ used when a reference value is matched by a non-reference pattern.
+
+In other words, we allow auto-dereferencing values during pattern-matching.
+When an auto-dereference occurs, the compiler will automatically treat the inner
+bindings as `ref` or `ref mut` bindings.
+
+Example:
+
+```rust
+let x = Some(3);
+let y: &Option<i32> = &x;
+match y {
+  Some(a) => {
+    // `y` is dereferenced, and `a` is bound like `ref a`.
+  }
+  None => {}
+}
+```
+
+Note that this RFC applies to all instances of pattern-matching, not just
+`match` expressions:
+
+```rust
+struct Foo(i32);
+
+let foo = Foo(6);
+let foo_ref = &foo;
+// `foo_ref` is dereferenced, and `x` is bound like `ref x`.
+let Foo(x) = foo_ref;
+```
+
+
+## Definitions
+
+A reference pattern is any pattern which can match a reference without
+coercion. Reference patterns include bindings, wildcards (`_`),
+`const`s of reference types, and patterns beginning with `&` or `&mut`. All
+other patterns are _non-reference patterns_.
+
+_Default binding mode_: this mode, either `move`, `ref`, or `ref mut`, is used
+to determine how to bind new pattern variables.
+When the compiler sees a variable binding not explicitly marked
+`ref`, `ref mut`, or `mut`, it uses the _default binding mode_
+to determine how the variable should be bound.
+Currently, the _default binding mode_ is always `move`.
+Under this RFC, matching a reference with a _non-reference pattern_, would shift
+the default binding mode to `ref` or `ref mut`.
+
+## Binding mode rules
+
+The _default binding mode_ starts out as `move`. When matching a pattern, the
+compiler starts from the outside of the pattern and works inwards.
+Each time a reference is matched using a _non-reference pattern_,
+it will automatically dereference the value and update the default binding mode:
+
+1. If the reference encountered is `&Pat`, set the default binding mode to `ref`.
+2. If the reference encountered is `&mut Pat`: if the current default
+binding mode is `ref`, it should remain `ref`. Otherwise, set the current binding
+mode to `ref mut`.
+
+```
+                        Start                                
+                          |                                  
+                          v                                  
+                +-----------------------+                     
+                | Default Binding Mode: |                     
+                |        move           |                     
+                +-----------------------+                     
+               /                        \                     
+Encountered   /                          \  Encountered       
+  &mut Pat   /                            \     &Pat
+            v                              v                  
++-----------------------+        +-----------------------+    
+| Default Binding Mode: |        | Default Binding Mode: |    
+|        ref mut        |        |        ref            |    
++-----------------------+        +-----------------------+    
+                          ----->                              
+                        Encountered                           
+                            &Pat                              
+```
+
+Note that there is no exit from the `ref` binding mode. This is because an
+`&mut` inside of a `&` is still a shared reference, and thus cannot be used to
+mutate the underlying value.
+
+Also note that no transitions are taken when using an explicit `ref` or
+`ref mut` binding. The _default binding mode_ only changes when matching a
+reference with a non-reference pattern.
+
+The above rules and the examples that follow are drawn from @nikomatsakis's
+[comment proposing this design](https://github.com/rust-lang/rfcs/pull/1944#issuecomment-296133645).
+
+## Examples
+
+No new behavior:
+```rust
+match &Some(3) {
+    p => {
+        // `p` is a variable binding. Hence, this is **not** a ref-defaulting
+        // match, and `p` is bound with `move` semantics (and has type `&i32`).
+    },
+}
+```
+
+One match arm with new behavior:
+```rust
+match &Some(3) {
+    Some(p) => {
+        // This pattern is not a `const` reference, `_`, or `&`-pattern,
+        // so this is a "non-reference pattern."
+        // We dereference the `&` and shift the
+        // default binding mode to `ref`. `p` is read as `ref p` and given
+        // type `&i32`.
+    },
+    x => {
+        // In this arm, we are still in `move`-mode by default, so `x` has type
+        // `&Option<i32>`
+    },
+}
+
+// Desugared:
+match &Some(3) {
+  &Some(ref P) => {
+    ...
+  },
+  x => {
+    ...
+  },
+}
+```
+
+`match` with "or" (`|`) patterns:
+```rust
+let x = &Some((3, 3));
+match x {
+  // Here, each of the patterns are treated independently
+  Some((x, 3)) | &Some((ref x, 5)) => { ... }
+  _ => { ... }
+}
+
+// Desugared:
+let x = &Some(3);
+match x {
+  &Some((ref x, 3)) | &Some((ref x, 5)) => { ... }
+  None => { ... }
+}
+```
+
+Multiple nested patterns with new and old behavior, respectively:
+```rust
+match (&Some(5), &Some(6)) {
+    (Some(a), &Some(mut b)) => {
+        // Here, the `a` will be `&i32`, because in the first half of the tuple
+        // we hit a non-reference pattern and shift into `ref` mode.
+        //
+        // In the second half of the tuple there's no non-reference pattern,
+        // so `b` will be `i32` (bound with `move` mode). Moreover, `b` is
+        // mutable.
+    },
+    _ => { ... }
+}
+
+// Desugared:
+match (&Some(5), &Some(6)) {
+  (&Some(ref a), &Some(mut b)) => {
+    ...
+  },
+  _  => { ... },
+}
+```
+
+Example with multiple dereferences:
+```rust
+let x = (1, &Some(5));
+let y = &Some(x);
+match y {
+  Some((a, Some(b))) => { ... }
+  _ => { ... }
+}
+
+// Desugared:
+let x = (1, &Some(5));
+let y = &Some(x);
+match y {
+  &Some((ref a, &Some(ref b))) => { ... }
+  _ => { ... }
+}
+```
+
+Example of new mutable reference behavior:
+```rust
+match &mut x {
+    Some(y) => {
+        // `y` is an `&mut` reference here, equivalent to `ref mut` before
+    },
+    None => { ... },
+}
+
+// Desugared:
+match &mut x {
+  &mut Some(ref mut y) => {
+    ...
+  },
+  &mut None => { ... },
+}
+```
+
+Example using `let`:
+```rust
+struct Foo(i32);
+
+// Note that these rules apply to any pattern matching
+// whether it be in a `match` or a `let`.
+// For example, `x` here is a `ref` binding:
+let Foo(x) = &Foo(3);
+
+// Desugared:
+let &Foo(ref x) = &Foo(3);
+```
+
+
+## Backwards compatibility
+
+In order to guarantee backwards-compatibility, this proposal only modifies
+pattern-matching when a reference is matched with a non-reference pattern,
+which is an error today.
+
+This reasoning requires that the compiler knows if the type being matched is a
+reference, which isn't always true for inference variables.
+If the type being matched may
+or may not be a reference _and_ it is being matched by a _non-reference
+pattern_, then the compiler will default to assuming that it is not a
+reference, in which case the binding mode will default to `move` and it will
+behave exactly as it does today.
+
+Example:
+
+```rust
+let x = vec![];
+
+match x[0] { // This will panic, but that doesn't matter for this example
+
+    // When matching here, we don't know whether `x[0]` is `Option<_>` or
+    // `&Option<_>`. `Some(y)` is a non-reference pattern, so we assume that
+    // `x[0]` is not a reference
+    Some(y) => {
+
+        // Since we know `Vec::contains` takes `&T`, `x` must be of type
+        // `Vec<Option<usize>>`. However, we couldn't have known that before
+        // analyzing the match body.
+        if x.contains(&Some(5)) {
+            ...
+        }
+    }
+    None => {}
+}
+```
+
+# How We Teach This
+[how_we_teach_this]: #how_we_teach_this
+
+This RFC makes matching on references easier and less error-prone. The
+documentation for matching references should be updated to use the style
+outlined in this RFC. Eventually, documentation and error messages should be
+updated to phase-out `ref` and `ref mut` in favor of the new, simpler syntax.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The major downside of this proposal is that it complicates the pattern-matching
+logic. However, doing so allows common cases to "just work", making the beginner
+experience more straightforward and requiring fewer manual reference gymnastics.
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. We could only infer `ref`, leaving users to manually specify the `mut` in
+`ref mut` bindings. This has the advantage of keeping mutability explicit.
+Unfortunately, it also has some unintuitive results. `ref mut` doesn't actually
+produce mutable bindings-- it produces immutably-bound mutable references.
+```rust
+// Today's behavior:
+let mut x = Some(5);
+let mut z = 6;
+if let Some(ref mut y) = *(&mut x) {
+    // `y` here is actually an immutable binding.
+    // `y` can be used to mutate the value of `x`, but `y` can't be rebound to
+    // a new reference.
+    y = &mut z; //~ ERROR: re-assignment of immutable variable `y`
+}
+
+// With this RFC's behavior:
+let mut x = Some(5);
+let mut z = 6;
+if let Some(y) = &mut x {
+    // The error is the same as above-- `y` is an immutable binding.
+    y = &mut z; //~ ERROR: re-assignment of immutable variable `y`
+}
+
+// If we modified this RFC to require explicit `mut` annotations:
+let mut x = Some(5);
+let mut z = 6;
+if let Some(mut y) = &mut x {
+    // The error is the same, but is now horribly confusing.
+    // `y` is clearly labeled `mut`, but it can't be modified.
+    y = &mut z; //~ ERROR: re-assignment of immutable variable `y`
+}
+```
+Additionally, we don't require `mut` when declaring immutable reference bindings
+today:
+```rust
+// Today's behavior:
+let mut x = Some(5);
+// `y` here isn't declared as `mut`, even though it can be used to mutate `x`.
+let y = &mut x;
+*y = None;
+```
+Forcing users to manually specify `mut` in reference bindings would
+be inconsistent with Rust's current semantics, and would result in confusing
+errors.
+
+2. We could support auto-ref / deref as suggested in
+[the original match ergonomics RFC.](https://github.com/rust-lang/rfcs/pull/1944)
+This approach has troublesome interaction with
+backwards-compatibility, and it becomes more difficult for the user to reason
+about whether they've borrowed or moved a value.
+3. We could allow writing `move` in patterns.
+Without this, `move`, unlike `ref` and `ref mut`, would always be implicit,
+leaving no way override a default binding mode of `ref` or `ref mut` and move
+the value out from behind a reference.
+However, moving a value out from behind a shared or mutable
+reference is only possible for `Copy` types, so this would not be particularly
+useful in practice, and would add unnecessary complexity to the language.
--- a/text/0000-match-ergonomics.md
+++ b/text/0000-match-ergonomics.md
@@ -218,7 +218,8 @@ No new behavior:
 match &Some(3) {
     p => {
         // `p` is a variable binding. Hence, this is **not** a ref-defaulting
-        // match, and `p` is bound with `move` semantics (and has type `&i32`).
+        // match, and `p` is bound with `move` semantics
+        // (and has type `&Option<i32>`).
     },
 }
 ```


petrochenkov:
There are examples of nested patterns, but no examples with nested/multiple references.
Is something like this work supposed to work?
```rust
match &&&Some(0) {
    Some(x)
}

=>

match &&&Some(0) {
    &&&Some(ref x) // typeof(x) == &i32
}

or

match &&&Some(0) {
    &&&Some(/* equivalent to */ ref ref ref x) // typeof(x) == &&&i32
}
```

From some details in the text (e.g. "we don't know whether `x[0]` is `Option<_>` or `&Option<_>`") it seems that only one layer of references is permitted (seems reasonable).

nikomatsakis:
> There are examples of nested patterns, but no examples with nested/multiple references.

My assumption is the following:

```
match &&&Some(0) {
    Some(x)
}

// Equivalent to:

match &&&Some(0) {
    &&&Some(ref x)
}
```

That is, we will dereference any number of `&T` or `&mut T` values during pattern matching, but we do not "reproduce" those references on the other side by creating a corresponding number of indirections. (That doesn't seem like a particularly useful behavior, and in fact would be quite limiting in terms of the maximum lifetime of those bindings, since they may have to point into the stack in the general case.)

cramertj:
@petrochenkov If the compiler can determine that the value being matched is a reference, and it is being matched by a non-reference pattern, it will dereference it and shift the default binding mode. This process should repeat until there are no references left, or the type is successfully matched.

Note that this does _not_ mean that it will pile up `ref`s, ala `ref ref ref...`. Dereferencing a shared reference switches the binding mode to `ref`, regardless of the existing binding mode.

Example:
```rust
match &&&Some(0) {
    Some(x) => {},
    None => {},
}

// will desugar to
match &&&Some(0) {
    &&&Some(ref x) => {},
    &&&None => {},
}
```

Similarly, I would expect a pattern with too few references to also compile:

```rust
// If the user writes this `match`:
match &&&Some(0) {
    &Some(ref x) => {},
    &None => {},
}

// the compiler sees this:
match &(&(&(Some(0)))) {
    &(Some(ref x)) => {},
    &(None) => {},
}

// and will desugar to
match &&&Some(0) {
    &(&&Some(ref x)) => {},
    &(&&None) => {},
}
```

When it sees the first `&` in the pattern, it dereferences the value being matched and continues inwards where it finds a non-reference pattern (`Some(ref x)`). The value being matched is still a reference (`&&Some(0)`), so it is dereferenced and the default binding mode is shifted to `ref`. This process repeats once more until the value is successfully matched.

nikomatsakis:
@cramertj 

This reads great! I think there is one sort of "ambiguity" in the RFC text, and the examples don't cover it. In particular, what sort of types are "auto-dereferenceable"? 

I see two interesting examples; one where the auto-deref occurs at the *outermost* level, and another with nested cases.

```rust
// EXAMPLE A -- deref at outermost level
//
// This could be made to work relatively easily, I think.

// If this works, I would expect `y` to be
// a `ref` binding.
let x: Rc<i32>;
match &x {
    Some(y) => ...,
    None => ...,
}

// More explicit form:
match &*x {
    &Some(ref y) => ...,
    &None => ...,
}
```

and:

```rust
// EXAMPLE B -- deref within
//
// I do not expect this to work.

let x: Option<Rc<Option<i32>>>;
match x {
    Some(Some(y)) => ...,
    None => ...,
}
```

My expectation is that we **could** support the first, but trying to support the second will be controversial and challenging. For one thing, it would require executing user-defined code (the `Deref` impl) during pattern matching; this may have side-effects (via `Cell`, `println!`) and there have been some concerns about that. 

I think I'd be in favor of supporting the first but not the second, but it does introduce a sort of asymmetry.

phaylon:
I'd like to make a case for allowing explicitly specifying `move`: The RFC talks about this being only useful with `Copy` types and thus "not ... particularly useful in practice". However, this is quite common in my code, for primitives, shared references, and my own `Copy` types. As far as I understand, to get at the normal copying behavior I'd have to specify "Foo(x) => *x" even when the variant type is "Foo(i32)".

cramertj:
@phaylon For copying inner `Copy` types, you could manually dereference them, call `clone()`, or write exactly what you do today:
```rust
struct Foo<T>(T);
let foo = Foo(10);
let foo_ref = &foo;
let &Foo(foo_inner_copy) = foo_ref; // Copies `10`
```

phaylon:
Ah, so as long as I don't leave off the `*` on the topic or `&` on the cases the match topic I won't trigger the auto-binding.

cramertj:
@phaylon Correct, and that's an important point-- this RFC doesn't change the behavior of _any_ existing code. It only allows for code that would not have compiled previously.

nikomatsakis:
I think that @withoutboats in the past specifically **wanted** to include `move` as something that the user could type. I think that @nrc, in contast, was not keen. @cramertj, quickly skimming the RFC, I realized I wasn't sure whether you were proposing that the following would or would not be legal:

```rust
fn main() {
    match &Some(3) {
        Some(move v) => // v is copied out
        None => ...,
    }
}
```

I think that I personally **favor** allowing `move` on bindings, specifically for pedagogical purposes. I'd like to be able to explain the binding modes just as you did, and show code with all modes fully elaborated, and then show how the defaults work when they are elided.

nikomatsakis:
(FWIW, I still think `move` is a non-perfect keyword, since "moving" a copy type is kind of a "copy", but I tend to teach this as 'when you reference a copy type, it is implicitly copied and everything is relative to that, so you are "moving the copy"', and I find that works "ok". Still it seems a bit more unfortunate than normal here in that `move` is **only** useful (in patterns) on copy types, since everywhere else that it is legal, it's also the default.)

nikomatsakis:
Beyond pedagogy, I think I find `Some(move v)` clearer than `&Some(v)` in terms of communicating that it is important that I am **copying `v` out from the option** here. (That said, it's often not that important, it's just needed to make types work out.)

cramertj:
@nikomatsakis I left a note about this at the very end of the RFC under alternatives. Personally, I think that `move` is more confusing than helpful because it only works for `Copy` types, which are already a point of confusion for newcomers to Rust.

I might be amenable to a different keyword. None of the entries on the existing [keyword list](https://doc.rust-lang.org/grammar.html#keywords) seem good to me. In a Rust 2.0/epoch world where we can add new keywords, I think `copy` would work nicely.

text/0000-match-ergonomics.md
```diff
@@ -0,0 +1,454 @@
...
+[comment proposing this design](https://github.com/rust-lang/rfcs/pull/1944#issuecomment-296133645).
+
+## Examples
+
+No new behavior:
+```rust
+match &Some(3) {
+    p => {
+        // `p` is a variable binding. Hence, this is **not** a ref-defaulting
+        // match, and `p` is bound with `move` semantics (and has type `&i32`).
```
rkruppe: Should this say "and has type `Option<&i32>`"?

cramertj: Yes, it should.

nikomatsakis:
> Personally, I think that move is more confusing than helpful because it only works for Copy types, which are already a point of confusion for newcomers to Rust.

I think by "only works for `Copy` types" you mean "if you have something where the default mode is not `move`, then using `move` will only work for copy types", right? This seems true, though one could also use it (redundantly) in cases where you don't need it, for pedagogical purposes.

I agree to some uncertainty. I think the argument in favor of it is that I would **rather** teach people about `Copy` types, and what a "move" means in that context, than to teach them about `&` patterns. The former has independent value: they have to learn about copy types anyway, and closure syntax means that they will encounter this `move` keyword in other places, where it has the same meaning ("move a copy"). But `&` patterns are just a way to get "move" mode by default under this proposal.

cramertj:
@nikomatsakis 
> I would rather teach people about Copy types, and what a "move" means in that context, than to teach them about & patterns.

That seems reasonable, but I probably wouldn't do either-- I'd just have them dereference the value:
```rust
struct Foo<T>(T);
let foo_ref = &Foo(10);
let Foo(inner) = foo_ref; // This gives an `&i32`, they want an `i32`.

// In simple, non-nested cases, just do this:
let Foo(inner) = *foo_ref;

// In complex, nested cases, this will always work:
let Foo(inner_ref) = foo_ref;
let inner = *inner_ref;
```

repax:
I suppose, and hope, that auto-dereferencing could work on user `Deref`-impls at some point in the future if they could be marked as *pure*, or free of side effects.

nrc:
> I suppose, and hope, that auto-dereferencing could work on user Deref-impls at some point in the future

This is an interesting question. I think the answer is that we should work with `Deref` as well as built-in references and there doesn't need to be a purity requirement, e.g., the following should work:

```
fn foo(x: Rc<Option<T>>) {
    match &x {
        Some(x) => ..., // x: &T
        None => {}
    }
}
```

In the previous RFC I proposed this should follow the rules for deref coercions and only deref if the discriminant is already borrowed. Some people felt we could follow the dot operator and deref all types.

burdges:
I'm slightly nervous about the new mutable reference behavior, but it's what `for` does already, so consistency wins out.  I do think any book that covers style should explain that, due to this convenience, functions that return mutable references, and data structures that contain mutable references, should indicate this fact with their suffix, like `_mut` or `Mut`.  

It's reasonable that Rust should only be responsible for making mutability explicit with basic operations, and in `std`, but if the programmers starts creating data structures that involve mutable references then they become responsible for making mutability explicit.  It's good to tell people though.

cramertj:
I'm working on gathering my thoughts with respect to the custom `Deref` question, and I'll post something in more detail soon. However, here's one interesting question I came across: how do we want to support types that implement both `Deref` and `DerefMut`? The obvious answer seems to be to prefer `DerefMut` since it allows for more flexibility, but I'm not so sure what that would actually mean in practice.

The following code seams reasonable:
```rust
let mut x = Box::new(Some(5));
match x {
    // The `DerefMut` impl of `Box` is used to get `&mut Option<i32>`,
    // which is then matched against `Some(5)`
    Some(5) => {},

    // Here, the `DerefMut` impl of `Box` is used to get `&mut Option<i32>`
    Some(x) => { 
        *x = 6;
    },

    // The `DerefMut` impl of `Box` is used to get `&mut Option<i32>`,
    // which is then matched against `None`
    None => {},
}
```

But what should happen in the case when `x` (our `Box<Option<i32>>`) isn't declared as `mut`? It can't be mutably borrowed for the `DerefMut` call, so it uses `Deref` instead.

```rust
let x = Box::new(Some(5)); // no `mut`
match x {
    // What happens here? We would use the `DerefMut` impl, but `x` is immutable.
    // So instead we could infer to use `Deref` here based on the mutability of `x`.
    Some(5) => {},

    // Here, we again see that `x` is immutable, and so use the `Deref` impl of `Box`
    // to get `&Option<i32>`
    Some(x) => {
        *x = 6; // Error: cannot assign to immutable borrowed content
    },

    // The `Deref` impl of `Box` is used to get `&Option<i32>`,
    // which is then matched against `None`
    None => {},
}
```

This obviously isn't an insurmountable issue, I just thought it was interesting that we would potentially have to dispatch to different methods (`Deref` vs. `DerefMut`) based on the mutability of the value being matched.

nikomatsakis:
@cramertj 

> That seems reasonable, but I probably wouldn't do either-- I'd just have them dereference the value:

That's an option, but it does lead to longer than necessary borrows, which in the general case can be a problem (although with NLL, less so). I sort of prefer to move the deref into the pattern match myself.

nikomatsakis:
@cramertj 

Hmm, the `Box` example you showed didn't work like I would have expected. That is:

```rust
let mut x = Box::new(Some(5));
match x {
    Some(y) => ... 
    None => ...
}
```

I do not expect `y` to be a reference here, as we have not passed through any `&` patterns. I would expect this to work only once we have some solution for `DerefMove`. For now, my assumption is that to trigger a `Deref` impl of any kind, the outermost type must be a reference, just as with a deref coercion. Therefore, I would expect *this* to work:

```rust
let mut x = Box::new(Some(5));
match &mut x {
    Some(y) => ... // y: &mut i32
    None => ...
}
```

cramertj:
@nikomatsakis Oh, I see. I was thinking that you were suggesting that things with `Deref<Target=T>` and `&T` be treated as equivalent for the purposes of pattern matching. That is, I was thinking that dereferencing a `Box<T>` would shift the default binding mode to `ref` or `ref mut`.

Your exact question was:
> In particular, what sort of types are "auto-dereferenceable"?

But I had mentally linked "auto-dereference" with "auto-dereference and shift the binding mode."

If we only auto-dereference custom types _without_ shifting the default binding mode, then yes, we can choose `DerefMove` vs. `Deref` vs. `DerefMut` based on the binding mode.

nikomatsakis:
@cramertj so, thinking about it more, I think we could leave the "autoderef" questions for another RFC. I'd really like to see the "core" ideas land. If we leave that out, do you feel like there are major unresolved questions for this RFC? I feel like it's ready to go to FCP, myself.

cramertj:
@nikomatsakis I agree, after thinking about it for a while I have a lot of questions about exactly how autoderef for custom types should work, and I think it'd be better to leave that feature for a future RFC.

The other major unresolved question is whether or not to introduce `move` patterns. As I said above, I'd prefer to leave them out, at least for the time being. `move` feels nice from a conceptual perspective, but its  usefulness is limited and IMO not worth the additional complexity it brings to the language.

nikomatsakis:
@cramertj 

> The other major unresolved question is whether or not to introduce move patterns. As I said above, I'd prefer to leave them out, at least for the time being. move feels nice from a conceptual perspective, but its usefulness is limited and IMO not worth the additional complexity it brings to the language.

Ah, yes, that's important to resolve. I think that I could go either way, but I lean towards having them. I think that it will be helpful for didactic purposes, and it doesn't really add any particular complexity to the language (in some sense, I feel like it removes complexity, in that there is no special form). But we already had this [back](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-303183127)-[and](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-303190298)-[forth](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-303219975), I think, so I guess I'm just re-treading ground.

I'm curious what the rest of the @rust-lang/lang team thinks. I am pretty sure that @pnkfelix would also want them, since they always favor an explicit form -- @withoutboats is also on record as being in favor. @nrc, I think, would prefer to leave them out. That just leaves @aturon and @eddyb to weigh in. =)

I suppose an alternative would be to seek for a different keyword or something like that, but it seems like `move` is really the *obvious* one here, given closures.

nikomatsakis:
(It's also true that if we do not include `move`, we can always add it at a later date.)

withoutboats:
I'd like to include `move`, but I'd also like to include explicit `ref` and `ref mut` on closures. Then a parallel can be made between these.

nikomatsakis:
@withoutboats 

> I'd also like to include explicit ref and ref mut on closures. Then a parallel can be made between these.

As part of this RFC, do you mean? Or just in general?

(Any ideas as to the syntax?)

withoutboats:
I don't mean as part of this RFC. As to the syntax, I just mean `ref |x| x + y` and `ref mut |x| y += x`, the same way that `move` can be applied today (`ref |x| y += x` would be an error).

aturon:
@cramertj, I finally had a chance to carefully read the RFC and the comment thread -- really fantastic work! This is basically how my mind always *wants* pattern matching to work, before I remind myself to go sprinkle in `ref`s.

Regarding some of the open questions:

- I'm in favor of punting user-defined `Deref` implementations and thus only supporting `&` and `&mut` to start. That's already a huge step, and we'll surely learn a lot along the way.

- I prefer not to introduce `move` *ever*, but certainly not at the outset. To restate what's been said: the only time you would *need* move is to reset the default binding mode to owned, which means you're moving out of a reference, which means you must have a `Copy` type. Of course, there are cases where this can be annoying (`&Option<i32>` giving you `&i32`), **but I'd prefer to resolve these cases by making `Copy` types more ergonomic in general**. That is, I'd like to pursue something like a `&T` to `T` coercion for `Copy` types.

  - re: @pnkfelix and having an explicit form, I'd consider the ability to dereference in the body of the arm a sufficient *explicit* form for "moving out" `Copy` data.

  - re: @withoutboats, can you elaborate on why you see the need for `move`?

I think that basically means I'm happy with the RFC as-is. I'm going to go ahead and propose to go to FCP **with the current state**, just so we can get the ball rolling, but of course those who strongly want to see `move` or `Deref` at the outset should withhold consent :-)

@rfcbot fcp merge

nikomatsakis:
So I agree that `&` suffices as an explicit form (in fact @cramertj and I were saying the same thing to each other over IRC just now). I am satisfied with this.

I am not *entirely* sure if I agree that just making `&Copy` more ergonomic removes the need for an explicit form, but it's a moot point, since we have one. The case I was thinking of was when using a `ref` binding would cause a loan that then fails the borrow checker (which would not be helped by `&Copy` ergonomics):

```rust
let mut x = Some(22);
loop {
    match &x {
        Some(p) => x = Some(p + 1), // illegal: `x` is borrowed here
        None => ...
    }
}
```

However, this is obviously artificial, since you could write it equally well as `match x { Some(p) ...` or `match &x { &Some(p) ...`. I'm not sure whether this will arise in other cases -- I feel like it does, but I couldn't come up with a convincing example.

withoutboats:
> re: @withoutboats, can you elaborate on why you see the need for move?

I don't see it as a need per se, more of an opportunity. :) So if other people don't feel its a simplifying force, I don't feel compelled to fight for it.

Right now we have `ref` and `ref mut` patterns on the one hands, and `move` closures on the other. Both of these are rather confusing features for new users in one way or another. My feeling is that we can unify these two concepts into a single notion of "ownership modes" which have certain defaults in different cases.

I also think `&Some(x)` is a less clear pattern than `Some(move x)` even considering that the move can only be applied to `Copy` types (I prefer a mental model  in which copies are just moves you get to keep using, ymmv).

cramertj:
> My belief is this is invalid, but I'm not sure.
@withoutboats Yes, that is valid under the current design. Each match arm is analyzed independently.

nikomatsakis:
> I also think &Some(x) is a less clear pattern than Some(move x) even considering that the move can only be applied to Copy types (I prefer a mental model in which copies are just moves you get to keep using, ymmv). 

I agree with this, but one mitigating factor is that, at the top-level, `&x` will behave as expected, whereas `move` will not:

```rust
let x = &3;
match x {
    move y => ... // `y` has type `&i32` here
}
```

(This is because we have not (yet) "skipped through" any `&` binders, so we are still in "move-by-default" mode, basically.)

nikomatsakis:
> Right now we have ref and ref mut patterns on the one hands, and move closures on the other. Both of these are rather confusing features for new users in one way or another. My feeling is that we can unify these two concepts into a single notion of "ownership modes" which have certain defaults in different cases.

Ah so I sort of remember now why I found `move` a more compelling answer for "explicit syntax" than the existing `&` and `ref` -- basically I had hoped we can get rid completely of `ref` bindings, if we have `move`. So if you imagine converting the following match (which you would write today):

```rust
match some_big_struct {
    SomeStruct { ref foo, bar } => { ... } // bar has `Copy` type
}
```

Into something like this:

```rust
match &some_big_struct {
    SomeStruct { foo, move bar } => { ... } // bar has `Copy` type
}
```

In that case, `move` would be the one thing that you learn, and it always has a consistent meaning -- take ownership of the value (sometimes a copy of it, if it is a `Copy` type).

However, there are a few flaws:

- move is an imperfect keyword, for reasons previously discussed
- there is the complication of the top-level, which might not behave as people expect
- if, as @aturon suggested, we could make it unnecessary to make `bar` "by value" anyway (i.e., because the `&Copy` behaves sufficiently close) then we could do even better and have no modifiers at all

So yeah, still happy to move forward without `move` and revisit later...

joshtriplett:
I'd really like to see some expansion on the "How We Teach This" section of this RFC.  The current RFC seems to assume that this approach will feel so much more natural that merely switching examples over to it will suffice.  However, I'd like to see some discussion of how this affects people's mental models of patterns and matching, both for new Rust programmers and those already comfortable with the language.

Right now, there's a kind of mental topological match that you can do, where you might imagine the shape of the value and the shape of the pattern, and let the pattern pull out bits of the value. (A diagram of that for a non-trivial expression would go really well with the discussion of match.) Things like `ref` act like an additional operator that applies when pairing up the two, taking a reference instead of a copy. (`ref` effectively acts like the *opposite* of `&`, adding a level of reference rather than removing one. Sometimes I wonder if `ref` should be spelled `*`.)

Prior to this RFC, no implicit coercions take place, so you always match the value and pattern precisely. What rigorous mental model should people follow after this RFC? And how can we convey that model?

withoutboats:
> @withoutboats Yes, that is valid under the current design. Each match arm is analyzed independently.

Oh, this concerns me. It seems confusing to allow the patterns to deconstruct the value differently like this.

petrochenkov:
@withoutboats 
The granularity is even smaller, each arm (`ARM1 | ARM2 | ARM3`) ([soon](https://github.com/rust-lang/rfcs/pull/1882) arm-subpattern) can deconstruct the value differently.

aturon:
@withoutboats 

> Oh, this concerns me. It seems confusing to allow the patterns to deconstruct the value differently like this.

I felt the same way at first, but after going carefully through the examples in the RFC text, I now can't imagine it working any other way.

aturon:
@withoutboats 

> I also think `&Some(x)` is a less clear pattern than `Some(move x)` even considering that the move can only be applied to `Copy` types

So, I certainly agree with that comparison, but I want to avoid both: just write `Some(x)` and have `x` coerce to an owned value as needed, due to being `Copy`.

withoutboats:
@aturon 

> So, I certainly agree with that comparison, but I want to avoid both: just write Some(x) and have x coerce to an owned value as needed, due to being Copy.

In favor of this (though still in favor of 'explicit move' for the conceptual unification I mentioned previously). But I also wonder if given those rules, we couldn't apply these rules in places that today we've said would be breaking? That is:

```rust
let x: &Option<String>;

match x {
    &Some(s) => { s: &String }
}
```

The only thing this breaks is Copy types becoming by reference, but we're saying we're solving that automatically.

Basically, I wonder if it wouldn't be better to just make the `&` optional here, instead of giving you a different ownership mode depending on whether the reference has been included or not.

aturon:
> Basically, I wonder if it wouldn't be better to just make the `&` optional here, instead of giving you a different ownership mode depending on whether the reference has been included or not.

I think what I'm advocating for is a world in which none of `&`, `ref`, their mutable variants, or `move` are needed for patterns. If we have coercions for `Copy` types (together with `*` if you want/need to be explicit), can you see any case you *need* one of these constructs to express?

In this world, for pattern matching you think of references not as giving you interesting type structure, but only affecting ownership.

withoutboats:
@aturon I agree with that goal. I'm just worrying about what happens when users do the unnecessary things, and how those explorations work together to form a whole mental model of how the system works.

I don't feel firmly about either of these, but I think there are two ways that the system could possibly be easier to understand:

* Have the same behavior when matching a reference whether you include the `&` or not.
* `move`, though unnecessary, is how to say to explicitly take ownership instead of by reference.

I'm also a little concerned about how we handle "inner references" - that is, references that aren't destructured at all, (so `Some(x)` matching `Option<&T>` or just `x` matching `&T`).

withoutboats:
But also AFAIK none of these would be breaking changes to adapt, which is why I've already checked my box.

aturon:
@withoutboats 

> Oh, this concerns me. It seems confusing to allow the patterns to deconstruct the value differently like this.

I've been thinking more about this, and want to conjecture: if we deprecate `&` patterns, this becomes a non-issue. Yet more fuel for that fire?

aturon:
@joshtriplett 

> What rigorous mental model should people follow after this RFC? And how can we convey that model?

Thanks for raising this concern; it'd be great to flesh out the section on teaching.

I've tried a couple times in the previous RFC to articulate my personal mental model, but let me try again.

- We continue to think of pattern matching as destructuring layers of type structure, where you generally write the construction form to produce the match.
- However, when it comes to reference types, we *do not consider this part of the type structure* for the purposes of matching. Instead, we see references purely as information about *ownership*. So `&Option<String>` and `Option<&String>` both have the same match-structure and ultimately ownership considerations; `Option<String>` has the same type structure but different ownership level. 
- This assumes that, with this change, we deprecate `ref`, `ref mut`, `&` and `&mut` in patterns, so you literally *can't* treat it as type structure. (With the `Copy`-based coercions we've been discussing, there should never be a need.)

This mental model coincides with a few other important things in Rust:

- Iteration via `for` (aka `IntoIterator`), where you can think of references as purely playing a role about ownership (which reappears in the iterated-over items).
- Deref coercions, which treat the *number* of layers of `&` (or other `Deref` types) as irrelevant, and are only concerned with owned vs borrowed. That's another case where we aren't really thinking of `&` as providing important type structure, but rather only ownership information.

To put it all really simply, when pattern matching, you match against the structure of the type with any `&`s removed; the variables bound in the match will then be wrapped with `&` or `&mut` as appropriate.

joshtriplett:
@aturon I'm having a really hard time translating that into a mental model that doesn't have a lot of "just ignore it and assume the right thing will happen".

As an example of the right thing not happening, see https://medium.com/@robertgrosse/how-copying-an-int-made-my-code-11-times-faster-f76c66312e0f ; in that case, the compiler's helpful inference caused a serious performance issue, invisibly.

I'm not trying to suggest that I have a well-founded concern here; I might well not. However, things like this give me that same feeling.

petrochenkov:
@joshtriplett 
>"just ignore it and assume the right thing will happen"

Hey, that's called ["post-rigorous" stage](https://github.com/rust-lang/rfcs/pull/1951#issuecomment-287482347) and means that you achieved enlightenment.

aturon:
@joshtriplett 

I'm not saying that you *ignore* layers of references and hope for the best; I'm saying that you don't think of them as interesting type structure for matching, but instead *as vital information about ownership*.

The point is that, if I have an `&Option<T>` in hand, I can think of that as "an `Option<T>` that is borrowed". From that perspective, I write a match that corresponds to the meaningful type structure (`Option`), and I fully expect -- and can predict -- that the variables in the arms will be borrowed if and only if the original value was. The exact same reasoning works for `Option<&T>`, `&&Option<T>`, and `Option<T>`. In every case, it will be totally clear that:

- The match arms should correspond to the relevant type structure, `Option`. That is, *for the purposes of which arms to write*, you strip out `&` and `&mut`.
- The variables in each arm will be borrowed if, and only if, they appeared nested within a `&` somewhere in the type (and mutably borrowed with `&mut` etc).

There's no magic, nor any complex constraint solving or interaction with type inference. Instead, there's a completely mechanical way to go from a given type to the match you should write, and the borrowing that will result from it.

Notably, there's little to no interaction with type inference here; the expression I'm matching against will have some type, inferred from other context, and the match is written completely predictably against that type.

Does that help? Can you try to think of some concrete examples where you feel like it's hard to understand what's going on, and you have to "just ignore it and assume the right thing will happen"?

Nemo157:
> there's little to no interaction with type inference here

This just made me have a thought, are there any ways in which removing explicit references from matches will cause type inference failures? And, if there are, are they common enough to care?

Without thinking about this in any detail, my presupposition is that there are some cases where this will cause type inference failure, but in those cases removing explicit references from the match and adding an explicit type declaration will be a net readability win.

text/0000-match-ergonomics.md
```diff
@@ -0,0 +1,455 @@
...
+Encountered   /                          \  Encountered       
+  &mut Pat   /                            \     &Pat
+            v                              v                  
++-----------------------+        +-----------------------+    
+| Default Binding Mode: |        | Default Binding Mode: |    
+|        ref mut        |        |        ref            |    
++-----------------------+        +-----------------------+    
+                          ----->                              
+                        Encountered                           
+                            &Pat
```
krdln: The "Encountered &Pat" labels in this flowchart are confusing for me. As I understand the RFC, it means "matched type is `&Pat`, but encountered pattern of type `Pat`", which is something different than this flowchart is suggesting.

Veedrac:
`move` seems misguided to me. If we have something like

    match &Some(3) {
        Some(v) => // v: &i32
        None => ...,
    }

Then `v` is an `&i32`, so the following looks like it should be valid

    match &Some(3) {
        Some(&v) => // v: i32
        None => ...,
    }

IOW, we just act like references follow the distributive property. `&Foo(a, b)` can go to `Foo(&a, &b)`, and perhaps `&&a` can go to `&a`.

jimmycuadra:
Possibly similar to Josh's reaction, I'm a little concerned with how these changes will be taught to people who already know the language.

This RFC and the discussion in the comments seem motivated by making the language more likely to do the right thing for new users without having to learn the details. But as a user of Rust for a few years, the changes here seem like a more complicated system to me. The only lightbulb moment I had on this topic when I was learning Rust was to understand that `match *x` means "the provided patterns will be evaluated against the value `x` is referencing, and `x` will continue to be a reference after the match block." This RFC feels like it trades that one insight for having to know/remember these "binding modes" when reading code that uses them. That's a net loss for me. Maybe what is needed is a better "how we teach this" targeting existing Rust users.

Interestingly, the change here that actually does interest me is about adding the explicit `move` to patterns. That's the only thing here that made things feel clearer rather than more complicated. I think it would be valuable for the same reasons that Niko mentioned (pedagogical) and that withoutboats mentioned (a unified model between patterns and closures.)

withoutboats:
> Possibly similar to Josh's reaction, I'm a little concerned with how these changes will be taught to people who already know the language.

I think the best way to teach this to people who already know the language is to compare it to `for` loops, which already work how this will work.

steveklabnik:
> Possibly similar to Josh's reaction, I'm a little concerned with how these changes will be taught to people who already know the language.

My personal experience is "write what this RFC says, get a compile error, sprinkle on some `ref`s, move on with life." I expect that with this RFC I'll just never have steps 2 and 3.

jimmycuadra:
> My personal experience is "write what this RFC says, get a compile error, sprinkle on some refs, move on with life." I expect that with this RFC I'll just never have steps 2 and 3.

Again, that's about _writing_ code. I'm more concerned with _reading_ existing code. Right now it's obvious whether something is a move or a reference because references are explicit.

burdges:
Interesting point.  It's true `for` patterns are harder to actually read and understand than `match` and `let` patterns, partially because they shift the burden of knowing where the references live from the writer to the (careful) reader. 

In `for`, these issues are partially addressed `mut` remains visible because `std::iter` obeys the convention that anything that returns a `&mut` has a `_mut` suffix.  I'd think this RFC fixed the previous version's bug where [mutability could be created ex nihilo using rvalues](https://github.com/rust-lang/rfcs/pull/1944#issuecomment-286873298) by requiring a mutable value. 

In `for`, you must write `for mut x in ..` to create newly mutable value from the actual rvalue returned by the iterator.  Analogously, I think this RFC requires `match f() { Some(mut x) => ... }` to get a mutable value when 'f()' does not return a `&mut Option<T>` or `Option<&mut T>`.

I think `std` mostly obeys this `_mut` suffix convention, but it gets tougher around `DerefMut`, so not sure how much clippy could enforce much of this convention for crates.

withoutboats:
@burdges If you have an `&mut [i32]` (or `&mut Vec<i32>` or `&mut VecDeque<i32>`, etc), the elements when you iterate over it are `&mut i32`, even without declaring them `mut`. This would follow the same situation - if you already have a mutable reference, you still have a mutable reference when you deconstruct it.

johncf:
I can see that this RFC improves "writability" as in it might reduce one or two compile-correct cycles, but not always. Here's an example:

```rust
let o = &Some(3);
match o {
    Some(x) => { ... },
    _ => (),
}
```

According to RFC, the `match` block desugars to:

```rust
match o {
    &Some(ref x) => { ... },
    &None => (),
}
```

But, since 3 is a `Copy` type, desugaring it without `ref` would also have been valid:

```rust
match o {
    &Some(x) => { ... },
    &None => (),
}
```

So a user who wrote the code expecting `x` to be `{integer}` would need an additional correct-compile cycle due to an "expected type `{integer}` found type `&{integer}`" error. (Ideally, the compiler should suggest to use `move` in the pattern, but I'd expect it to be hard(?))

Besides this loss of "writability-gain", there could be plain loss of readability too:

```rust
let o = &Some((3, 4));
match o {
    Some((ref x, y)) => { ... },
    _ => (),
}
```

This, according to my understanding, desugars to:

```rust
match o {
    &Some((ref x, ref y)) => { ... },
    &None => (),
}
```

This is simply confusing to a reader --- the fact that different patterns for `x` and `y` getting desugared to the same pattern for both. Of course, you could blame the author for writing such code, but the problem remains that the compiler happily accepts this.

Here's another code that I would not like to see (again, an unlikely one too):
```rust
match res {
    &Ok(x) => { ... }
    Err(y) => { ... }
}
```
Here's the question -- what's the type of `x` and `y`? If the type of `res` is `&Result<T, E>`, then `x: T` and `y: &E`, but if the type of `res` is `&&Result<T, E>`, then `x: &T` and `y: &E`. Desugaring in both cases would not be as straightforward to a reader as (in our plain old Rust):
```rust
match res {
    &Ok(x) => { ... }
    &Err(ref y) => { ... }
}
```

Sorry, but I really hope we don't go forward with this RFC.

johncf:
From a high-level perspective, the goal of this RFC is to move the responsibility of detailing type-specifics from the user to the compiler. The claim probably goes like "The less details the user is concerned with, the more productive she would be." Or the goal might be to "Reduce the landing time for the beginners."

Both of those goals seems very reasonable to me. But let me share some of my Rust-learning experience. (\<rant>) One of the things that took me a long time to understand is the `Deref` trait and the magic surrounding the dot and star operators. Especially in the following example:
```rust
let s = String::from("hello");
let r: &str = &*s;
```
To intuitively understand this, I used a mental model which basically says "`String`, just like `&str`, is a pointer to `str` too."

Once I understood the magic surrounding star and dot operators, I began to see things in a new light. It was an "of course it should be so" moment! But I would very much prefer for that magic to be restricted to those operators. I believe many developers, like myself, would prefer a clear understanding of things going on under the hood than thinking "It is magic." _(Side note: I suspect this to be one of the main reasons why tokio is not taking off as quick as it should be -- a lack of quality docs on internals and mechanics.)_ And investing time to understand such magic was significant for me. This investment of time is usually made after one passes the "beginner" level of expertise, and often people at that level are not considered when making design decisions (because it is somewhat vague). (\</rant>)

From what I can tell, every new magic feature added to the language takes away time-investments from "beginners", and introduces time-investments to "post-beginners". So it should be thoroughly investigated whether there is a significant net time-gain before introducing any "sugary" feature.

phaylon:
@lxrec: My concern is simply retaining the ability to stay explicit, and being able to continue using `ref` and `ref mut`. As an example, just today I wrote this:
```
let Select {
    ref selector,
    ref stream,
    ref mut builder,
    ref mut state,
    ref mut restrict,
} = *self;
```
to operate on the different parts individually. I value the documentation effect of using `mut` selectively very highly.

I seem to end up dabbling with a lot of complex state machines, with nested matches on function call return values, and I find the current explicit nature makes reading the code later easier.

johncf:
> try to articulate a more specific reason for that than "it feels too magical" (I've been unable to find any in this entire comment thread)

You're right about that regarding my second comment. But there, I was talking more generally (probably shouldn't have). See my [first comment](#issuecomment-307306060) for specifics.

Basically, most concerns raised in this thread boils down to one thing: Explicitness. I thought it was one of Rust's core philosophy to be explicit, which (at least for me) greatly increased readability. I feel that this RFC goes very much against that philosophy.

petrochenkov:
@johncf 
>Basically, most concerns raised in this thread boils down to one thing: Explicitness. I thought it was one of Rust's core philosophy to be explicit, which (at least for me) greatly increased readability. I feel that this RFC goes very much against that philosophy.

The whole ergonomic initiative is mostly against this philosophy, as practice has showed, but it's being pushed for by the teams and upvoted by language users, so maybe the old philosophy is not here anymore.

What the "explicit"-school people can do is to consistently go through the "ergonomics" RFCs and try to minimize harm, in case of this RFC this minimization was pretty successful, since the old totally magical RFC was replaced with this RFC proposing very straightforward desugaring (simpler than autoderef!). It can be made even more straightforward, e.g. if [multiple dereferences](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-303046368) are prohibited (they can be introduced later if practice shows they are more useful than harmful).
Seems like something called "compromise".

johncf:
> try to minimize harm

A suggestion to that end would be to replace `move` keyword with type annotations and [`Copy` coercions](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-305338105). Here's an example:

```rust
let o = &Some(3);
match o {
    Some(x as i32) => { ... },
    _ => (),
}
```
This would not only clarify what type it is, but would also be identical to when `o = Some(&3)`.

This would be adding a separate meaning to the keyword `as`, which might be confusing, since a `usize` cannot be `as i32`-ed within patterns.

petrochenkov:
@johncf 
>type annotations

For reference, there's a postponed issue about type ascription in patterns - https://github.com/rust-lang/rfcs/issues/354 (e.g. the syntax is `Some(x: i32)`).

withoutboats:
@Ixrec thankyou for your thorough comment! I just want to recognize how much work you put into challenging your own viewpoint. That's a very valuable thing to do both for clarifying your own views and for the insightful commentary it can generate for everyone else.

---

> For completeness, there is one thing I believe the RFC actually does need to clarify. Neither the previous RFC nor this one states whether or not the ref keyword is expected to be deprecated and removed from the language in some future epoch, or if it's expected to hang around as an advanced feature useful for pedagogical purposes or obscure corner cases.

I think we're punting on this for further learning once we can actually use this feature. Certainly this RFC alone doesn't mean `ref` is going away.

---

> I thought it was one of Rust's core philosophy to be explicit, which (at least for me) greatly increased readability. I feel that this RFC goes very much against that philosophy.

Its become a common belief that explicitness is one of Rust's core values, but I don't think it is. Rust's goal is to help you avoid errors; while there's always an argument to be made that being explicit is a way to do that, we only choose explicitness as a tool toward the end of avoiding errors, and I don't think we'd say explicitness is a virtue in itself.

I would say what's often been taken as "explicitness" in Rust is really better characterized as "honesty." Converting Paths to Strings doesn't require error handling because of a value for explicitness, but because it would be *dishonest* to claim that they can be converted to UTF-8 Strings. These ergonomics initiatives are not "dishonest" in that way; we still uphold all the invariants we promise to, they just aren't as in-your-face.

I also would be remise not to mention this blog pots by aturon, which outlines a more nuanced criteria for evaluating features than just saying explicitness is or is not good: https://blog.rust-lang.org/2017/03/02/lang-ergonomics.html

joshtriplett:
@withoutboats 
> I think the best way to teach this to people who already know the language is to compare it to for loops, which already work how this will work.

Can you elaborate on this? This seems like an interesting insight, and I'm not entirely clear on how those patterns are different. (The only difference I know of is "given an IntoIterator, call into_iter()".)

withoutboats:
> Can you elaborate on this? This seems like an interesting insight, and I'm not entirely clear on how those patterns are different. (The only difference I know of is "given an IntoIterator, call into_iter()".)

Because we've implemented `IntoIterator` for types like `&mut [T]` and `&HashMap<K, V>`, if your iterable is a reference, the items you'll get are also references. You get the same surface behavior when destructuring because of this RFC - if you had `&`, you get an `&`, if you had `&mut`, you get an `&mut`.

glaebhoerl:
First off, I think the RFC is good to go as-is!

As for where we should go from here in the future, I share the vision that @aturon described in [their comment](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-305347905). Basically, I think we should try to deprecate `&`, `&mut`, `ref`, *and* `ref mut` in patterns, and not introduce `move`, and have a world where the amount of special, strange, or funny syntax in patterns is zero. That makes the language a lot cleaner, and pedagogically, doing adjustments in the RHS of the `match` arm is *way* easier to explain than doing them in the pattern. For one thing, because the inverted meanings of things in a pattern context can be hard to grok (and once grokked, it's weird if something *isn't* inverted, so you can't win either way); and for another, because the RHS only involves things that are *already known* such as the `*` operator to dereference, or `&` to get a shared borrow if the `match` produced an `&mut`.

I also think we should move towards allowing the default binding mode to "pass through" user-defined `Deref` and `DerefMut` types in the case where they were already borrowed, as in @nikomatsakis's [comment](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-303565666), so you could `match` an `&Box<Foo>`, but not a `Box<Foo>`, and get an `&Foo`. This would further extend the "no strange syntax in patterns" story by obviating the need to add [`box` patterns](https://github.com/rust-lang/rust/issues/29641).

glaebhoerl:
(Woo! I made it in before @rfcbot!)

glaebhoerl:
Ah, I remembered that @nikomatsakis had a comment raising concerns about the feasibility of deprecating `ref`/`ref mut` on the previous RFC, but strangely couldn't find it yesterday. Turns out that's because it was actually [in a different discussion](https://internals.rust-lang.org/t/lang-team-minutes-elision-2-0/5182/61). Is there any case where reborrowing as shared with `&` on the RHS wouldn't work when you want these "mixed mode" borrows? That might be a bit more verbose, but as long as it comes up relatively rarely and there's no loss in actual expressivity, it seems like it might be worth taking the hit for the overall simplification of the language.

withoutboats:
@glaebhoerl I think your sentiment about the eventual state of patterns is shared by other people (definitely seems like what @aturon wants to do), and I'm sympathetic to it. I'd like to see how things look after we've implemented this RFC, and then revisit whether we want to go the "eliminate ref/ref mut/&/&mut" route or the "add move mode, maybe type ascriptions, but usually you won't use them" route.

nikomatsakis:
I also think that NLL is an ingredient in all of this. That is, if you *have* NLL, then you can do something like this:

```rust
let mut x = Some(3);
match &x {
    Some(v) => { // v: &i32
        let v1 = *v; // lifetime of `v` ends here
        x = Some(5); // so I can mutate `x` again
    }
    None => { }
}
```

Otherwise, I'm not sure how you could write that code without using either an `&Some(v)` pattern *or* a `Some(move v)` pattern.

nikomatsakis:
Anyway, I agree with @withoutboats -- let's see what develops.

nikomatsakis:

To track further discussion, subscribe to the tracking issue here: https://github.com/rust-lang/rust/issues/42640

petrochenkov:
[meta] Please, make sure RFC texts are updated with discussion results before they are merged.
For this specific RFC everything mentioned in this thread in the last 23 days is not reflected in the RFC text, including resolutions for multiple derefs and custom derefs.
This happens all the time and creates bugs and confusion.

nikomatsakis:
@petrochenkov thanks for pointing that out.

burdges:
I kinda doubt `&Some(v)` or `Some(move v)` could deprecated using NLL.  In fact, one could imagine either wanting to temporarily deactivate NLL or wanting more explicit binding modes as part of teaching a course on Rust.

samoylovfp:
"Rendered" link is broken

Gankro:
Not sure if this is part of @petrochenkov's comment, but the current example at the start of the RFC is incredibly confusing, as it's not at all what the RFC is supposed to do.

winsut:
Currently (and with this RFC), if one matches on a "mut" Option variable for Some(x), and wants the "x" to be mut, they have to specify "mut" in the pattern.

However, with this RFC, if one matches on a &mut Option for Some(x), then the x automatically becomes an &mut, even without specifying "mut" in the pattern.

Is this what we want?

The alternative is to have the binding mode only be "ref mut" if "mut" is specified in the pattern, and be "ref" otherwise.

cramertj:
@winsut: Good question! I addressed this issue in [the alternatives section](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md#alternatives). Additionally, the RFC's approach matches existing conventions already present [in things like `for` loops](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-307265785) and `let` bindings.

For example, the following all introduce new `&mut` variables without specifying `mut`:
```rust
let mut v = vec![];

let mut_ref = &mut v;
let mut_ref_2 = mut_ref; // `mut_ref_2` is `&mut _`

for x in mut_ref_2 { // `x` is `&mut _`
    ...
}
```

aturon:
When we first went through this RFC, I missed an [interesting comment](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-307602419) by @phaylon. The crux is that if you have an `&mut SomeStruct`, today you can use matching to destructure that into references to its fields, some of which are mutable, and some of which are not. Just like with the `mut` keyword, this can give you more insight into mutation in the code that follows.

@cramertj @nrc @nikomatsakis I'm curious what you think about this use case. Can you see another way to achieve a similar result if we eventually fully deprecate `ref`? Might this be a reason to keep `ref` around?

glaebhoerl:
Copying @phaylon's example just for easy reference:

```rust
let Select {
    ref selector,
    ref stream,
    ref mut builder,
    ref mut state,
    ref mut restrict,
} = *self;
```

At *worst*, you could express it as:

```rust
let Select {
    selector,
    stream,
    builder,
    state,
    restrict,
} = self;
let (selector, stream) = (&*selector, &*stream);
```

So you're "explicitly discarding" your privileges on `selector` and `stream`. I wonder if there's a nicer way to write that than `&*`?

nikomatsakis:
Hmm, to be honest, I think that the use case of "intermingled" `ref` and `ref mut` is *less* interesting than the use case of *just* using `ref`. If you only use `ref` bindings, then you wind up with a shared reborrow of the `&mut`, which you can do multiple times. So for example this would type-check:

```rust
let pair: &mut (i32, i32) = &mut (0, 1);
let x = match *pair { (ref x, _) => x };
let p: &(i32, i32) = &*pair;
```

while the following "erasure" does not:

```rust
let pair: &mut (i32, i32) = &mut (0, 1);
let x = match pair { (x, _) => x };
let p: &(i32, i32) = &*pair;
```

To do something equivalent, you would have to explicitly use a `&` borrow at the outermost level:

```rust
let pair: &mut (i32, i32) = &mut (0, 1);
let x = match &pair { (x, _) => x };
let p: &(i32, i32) = &*pair;
```

I'm not sure this is *bad*, but it is not something I had fully realized before. It seems related to the reasons that it is useful to have `last()` and `last_mut()` -- i.e., so that if you have an `&mut` reference, you can explicitly choose, via the name of the method, how you wish to be using it at a particular time. (In comparison to a hypothetical "generic" `last` that would be mut if `self` is mut.)

nikomatsakis:
One could imagine requiring the `mut` keyword to create the equivalent of today's `ref mut`:

```
let pair = &mut (0, 1);
match pair { (mut x, _) => { ... } }
// equivalent to: `match *pair { (ref mut x, _) => { ... } }`
```

Right now I think that `mut x` creates a variable that one can reassign (just as it does in any other context). This seems more "intuitive" to me.

joshtriplett:
@nikomatsakis The implicit modification of the original pair still seems *very* confusing to me. Having "mut" there says "this is modifiable", not "touching this will modify something else".

nikomatsakis:
@joshtriplett what do you mean by "implicit modification"? You mean that `x` has type `&mut i32`? This implies (then) that `x` could be used to modify the original by doing something like `*x += 1`. I think what you are saying is that, looking at `match pair { (x, _) => ... }`, you know that the code in the `...` cannot use `x` to mutate `pair`? I can certainly see this POV; I'm not sure how confusing it would be in practice. I don't yet see the best way to signal that you want `x` to be a "mutable" borrow. As you say, `mut x` doesn't feel ideal to me, since it seems to be saying that `x` is mutable, not that `*x` is mutable.

joshtriplett:
> As you say, mut x doesn't feel ideal to me, since it seems to be saying that x is mutable, not that *x is mutable.

This is an accurate summary of my concern, yes.

When I see something like `match pair { (mut x, _) => ... }`, I don't see anything that suggests that *pair* is modifiable. When I see a `&` or a `ref`, that possibility comes into play.

withoutboats:
I think @phaylon's example is why I was personally in favor of requiring `mut` to make an `&mut` binding.

joshtriplett:
@withoutboats I definitely wouldn't want `(x, _)` to create a `&mut` binding where there wasn't one previously; I just also don't think `(mut x, _)` is enough to make that clear enough, either.

cramertj:
@joshtriplett 
>  I definitely wouldn't want (x, _) to create a &mut binding where there wasn't one previously

It doesn't create one where there wasn't one previously-- if you write `if let Some(x) = &mut Some(5) { ... }`, `x` will be an `&mut i32`. It's just matching the reference type of the value being matched.

withoutboats:
@joshtriplett I think your concerns were already expressed in the RFC thread above. rather than re-open general concerns about the RFC I'd like it if we could focus on the specific concerns @aturon bumped the thread about (originally raised by @phaylon) about splitting an object into mutable and immutable parts.

Is this something we want to support easily under this RFC? What are the ways to do that? To me it seems like "requiring `mut` to get an `&mut`" is the most obvious way. I don't really agree with @nikomatsakis's concern about this being surprising. I guess technically it means there isn't a way to support mutating the value without reassigning the reference, but those two things really seem to go hand in hand to me. In fact, it seems like a stumbling block to require `mut` to reassign when you don't require `mut` to call `&mut self` methods.

cramertj:
IMO requiring `mut` here is inconsistent with other parts of the language:
```rust
let x = &mut 5; // `x` is `&mut`
let (x, y, z) = (&mut 1, &mut 2, &mut 3); // `x, y, z` are all `&mut`
for i in &mut vec![1, 2, 3] { ... } // `i` is `&mut`
```

It's probably worth discussing if we want to interpret the above examples differently in a new epoch/version, but for the time being, I think it's better to stay consistent.

I personally don't think the mixed `&mut`/`&` case is very compelling-- I can't imagine a scenario where it would cause a change in behavior to mark _some_ bindings as immutable (where they would have otherwise defaulted to `&mut`). Marking one or two references as immutable seems like more of a defensive move to ensure you're not mutating some items. It'd be nice to enable this (as it aligns with general best practices WRT "immutable by default") but, as I said above, I don't think it's consistent with our current handling of mutable references.

burdges:
I think this RFC creates only the second place where `rustc` plus `std` produce a named `&mut` binding without either some composite data type involving a `&mut` or some function returning one. 

There is a convention throughout most of `std` that `_mut` appear on functions that directly produce such objects.  I [believe](https://github.com/rust-lang/rfcs/pull/2005#issuecomment-307265785) the only current exception are the `impl<'a, T> IntoIterator for &'a mut F<T>` where `F` is `[]`, etc.  You need `iter_mut()` if you want an `Iterator<Item=&mut ..>` but do not have `&mut [T]` or `[&mut T]` or similar.  See: http://play.integer32.com/?gist=8a3899b7ab57fee0198755b04ecb6bfc&version=undefined 

Avoiding the `mut` here is only aiming for consistency with those `IntoIterator`s for `&mut`s.  It's looks incomparable to anything `rustc` itself requires.

glaebhoerl:
I can recognize the value there; at the same time, I also feel inclined to ask: if the language had *started out* with "default binding modes", and without the ability to express that, would we consider it worthwhile to add `ref` and `ref mut` to the language in order to cover that use case? Does the feature carry its weight?

phaylon:
@glaebhoerl Yes, I would. Similar to if `let` were always `mut` I'd welcome an immutable/mutable separation, with immutable being the default.

burdges:
I've definitely hit points where default binding modes hide the distinction between copies and `&mut` borrows, resulting in necessary side effects being lost.  I've noticed that _error[E0384]: cannot assign twice to immutable variable_ represents a tame version, which gives constructive messages:
```text
help: consider making this binding mutable
|
68 | let EvictionMask { mut pk, apk, deck_delta } = itr.next()?;
| +++
help: to modify the original value, take a borrow instead
|
68 | let EvictionMask { ref mut pk, apk, deck_delta } = itr.next()?;
| +++++++
```
The nasty version is when some hidden copy wrecks your `Iterator<Item=&mut T>` or similar.

It's maybe addressable by having deeper check for unused mutability.

PR #2013, opened 2017-06-01
Fix a typo about desiring unbounded lookahead.

tomprince:
I'm fairly certain the desire is to not require anything unbounded.

### DIFF
--- a/text/1624-loop-break-value.md
+++ b/text/1624-loop-break-value.md
@@ -305,7 +305,7 @@ let x = for thing in things default "nope" {
 There are two things to bear in mind when considering new syntax:
 
 *   It is undesirable to add a new keyword to the list of Rust's keywords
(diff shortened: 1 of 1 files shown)


text/1624-loop-break-value.md
```diff
@@ -305,7 +305,7 @@ let x = for thing in things default "nope" {
 There are two things to bear in mind when considering new syntax:
 
 *   It is undesirable to add a new keyword to the list of Rust's keywords
-*   It is strongly desirable that unbounded lookahead is required while syntax
+*   It is strongly desirable that unbounded lookahead *not* is required while syntax
```
Keats: the not should go after the is?

tomprince: Yes.

aturon:
Thanks!

PR #2020, opened 2017-06-05
Implement Add for OsString

Rufflewind:
Add `Add` to `OsString`.

[Rendered](https://github.com/Rufflewind/rfcs/blob/os-string-add/text/0000-os-string-add.md)

### DIFF
--- a/text/0000-os-string-add.md
+++ b/text/0000-os-string-add.md
@@ -1,0 +1,45 @@
+- Feature Name: `os_string_add`
+- Start Date: 2017-06-04
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Implement `Add` for `OsString` analogous to `Add` for `String`.
+
+# Motivation
+[motivation]: #motivation
+
+Make concatenating `OsString`s (e.g. in paths and/or command-line arguments) a tiny bit more convenient, which in turn makes writing small command-line utilities easier.
+
+# Detailed design
+[design]: #detailed-design
+
+Basically this:
+
+```rust
+impl Add<&OsStr> for OsString { type Output = OsString; ... }
+```
+
+It is a thin wrapper over `.push`.  The main advantage of this over `push` is that `push` necessitates having a mutable (in/out) variable and is hard to nest.  A typical usage of `push` would require at least 2 extra lines: one for the variable definition, and one for the `push`, neither of which are necessary for `+`.
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+Users already familiar with `String` shouldn't be surprised by the presence of this implementation.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Not aware of any besides adding a tiny bit of bloat to the std API surface.
+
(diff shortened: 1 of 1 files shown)


ubsan:
I don't like the `+` implementation for `String`, but as long as it exists it seems reasonable to have these implementations for `OsString`.

BurntSushi:
I think this RFC is lacking detail. Here are some questions that I expect to see answered in the RFC itself:

* An implementation on `Add` seems to encourage allocating new `OsString` values over mutating existing ones. How important is this draw back?
* What examples from real code benefit from this change?
* The "How We Teach This" section is basically non-existent. I don't think "users already know it so we don't need to teach it" is an acceptable answer. For example, do we start encouraging people to use `+` for `OsString` as an idiom? How do we balance it with the trade offs of using `push`?
* Why aren't we also implementing `AddAssign`?
* If we add `Add` impls for `OsString`, does that mean we should add `Add` impls for `PathBuf` too?
* If we don't like the `Add` impls on `String`, does it still make sense to add new `Add` impls for other string-like types?

Rufflewind:
> Unfortunately, I think this RFC is lacking detail.

It was intentionally short, given that:

- it is 7 lines of code
- it mirrors the existing `String` API
- it was presumed that the absence of `Add` was only an omission
- it does not impose any new semantics on `OsString` because `push` already exists

I was not previously aware of the controversy over `Add` for `String`.

> An implementation on Add seems to encourage allocating new OsString values over mutating existing ones. How important is this draw back?

`Add` does not encourage allocation any more than `.push` does, as the former is no more than a syntactic sugar over the latter that avoids the need for an intermediate mutable variable.

> Why aren't we also implementing AddAssign?

It can be included if there is demand.

burdges:
I'd say deprecate the `+` for `String` and add a binary `++` operator called `MonoidAdd` or something.

eddyb:
Why `MonoidAdd` when `Concat` or `Append` work quite nicely?

burdges:
Yeah `Append` would be better, anyone who knows what a moniod is can figure out that they should be using it.

le-jzr:
I'm partial to `Concat` over `Append`, but +1 either way.

With `Concat` that takes a slice of arbitrary size, instead of just two arguments, you could efficiently compile expressions like `a ++ b ++ c ++ d`, for which a regular binary operator would be very inefficient and deter users in favor of function calls.

eddyb:
@le-jzr I'd use some sort of `HList` (tuples, ideally) to allow heterogenous concats (`Option<T>` and/or `Option<&[T]>` comes to mind).

aturon:
@BurntSushi can we drive this toward a conclusion?

BurntSushi:
Given the inactivity on this RFC, I move to close it. In general, I think whether this RFC is appropriate or not depends on whether we want to see the `Add` impl on `String` take over on other string types. I lean towards no. I think this change just doesn't carry its weight.

@rfcbot fcp close

vks:
Does this imply `+` is going to be deprecated for `String` in some future epoch?

Centril:
Closing since FCP with a motion to close is now complete.

PR #2027, opened 2017-06-10
Tweak object safety rules to allow static dispatch

withoutboats:
[Rendered](https://github.com/withoutboats/rfcs/blob/object-safe-for-dispatch/text/0000-object_safe_for_dispatch.md)

### DIFF
--- a/text/0000-object_safe_for_dispatch.md
+++ b/text/0000-object_safe_for_dispatch.md
@@ -1,0 +1,132 @@
+- Feature Name: object_safe_for_dispatch
+- Start Date: 2017-06-10
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Tweak the object safety rules to allow using trait object types for static
+dispatch, even when the trait would not be safe to instantiate as an object.
+
+# Motivation
+[motivation]: #motivation
+
+Because Rust features a very expressive type system, users often use the type
+system to express high level constraints which can be resolved at compile time,
+even when the types involved are never actually instantiated with values.
+
+One common example of this is the use of "zero-sized types," or types which
+contain no data. By statically dispatching over zero sized types, different
+kinds of conditional or polymorphic behavior can be implemented purely at
+compile time.
+
+Another interesting case is the use of implementations on the dynamically
+dispatched trait object types. Sometimes, it can be sensible to statically
+dispatch different behaviors based on the name of a trait; this can be done
+today by implementing traits (with only static methods) on the trait object
+type:
+
+```rust
+trait Foo {
+    fn foo() { }
+}
+
+trait Bar { }
+
+// Implemented for the trait object type
+impl Foo for Bar { }
+
+fn main() {
+    // Never actually instantiate a trait object:
+    Bar::foo()
+}
+```
+
+However, this can only be implemented if the trait being used as the receiver
+is object safe. Because this behavior is entirely dispatched statically, and a
+trait object is never instantiated, this restriction is not necessary. Object
+safety only matters when you actually create a dynamically dispatched trait
+object at runtime.
+
+This RFC proposes to lift that restriction, allowing trait object types to be
+used for static dispatch even when the trait is not object safe.
+
+# Detailed design
+[design]: #detailed-design
+
+Today, the rules for object safey work like this:
+
+* If the trait (e.g. `Foo`) **is** object safe:
+    - The object type for the trait is a valid type.
+    - The object type for the trait implements the trait; `Foo: Foo` holds.
+    - Implementations of the trait can be cast to the object type; `T as Foo`
+    is valid.
+* If the trait (e.g. `Foo`) **is not** object safe:
+    - Any attempt to use the object type for the trait is considered invalid
+
+After this RFC, we will change the non-object-safe case to directly mirror the
+object-safe case. The new rules will be:
+
+* If the trait (e.g. `Foo`) **is not** object safe:
+    - The object type for the trait **does not** implement the trait;
+    `Foo: Foo` does not hold.
+    - Implementations of the trait **cannot** be cast to the object type,
+    `T as Foo` is not valid
+    - **However**, the object type is still a valid type. It just does not meet
+    the self-trait bound, and it cannot be instantiated in safe Rust.
+
+This change to the rules will allow trait object types to be used for static
+dispatch.
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+This is just a slight tweak to how object safety is implemented. We will need
+to make sure the the official documentation is accurate to the rules,
+especially the reference.
+
+However, this does not need to be **highlighted** to users per se in the
+explanation of object safety. This tweak will only impact advanced uses of the
+trait system.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is a change to an existing system, its always possible it could cause
+regressions, though the RFC authors are unaware of any.
+
+Arguably, the rules become more nuanced (though they also become a more direct
+mirror).
+
+This would allow instantiating object types for non-object safe traits in
+unsafe code, by transmuting from `std::raw::TraitObject`. This would be
+extremely unsafe and users almost certainly should not do this. In the status
+quo, they just can't.
+
+# Alternatives
+[alternatives]: #alternatives
+
+We could instead make it possible for every trait to be object safe, by
+allowing `where Self: Sized` bounds on every single item. For example:
+
+```rust
+// Object safe because all of these non-object safe items are constrained
+// `Self: Sized.`
+trait Foo {
+    const BAR: usize where Self: Sized;
+    type Baz where Self: Sized;
+    fn quux() where Self: Sized;
+    fn spam<T: Eggs>(&self) where Self: Sized;
+}
+```
+
+However, this puts the burden on users to add all of these additional bounds.
+
+Possibly we should add bounds like this in addition to this RFC, since they
+are already valid on functions, just not types and consts.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How does this impact the implementation in rustc?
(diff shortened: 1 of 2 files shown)


withoutboats:
cc @nikomatsakis @eddyb

text/0000-object_safe_for_dispatch.md
```diff
@@ -0,0 +1,132 @@
...
+safety only matters when you actually create a dynamically dispatched trait
+object at runtime.
+
+This RFC proposes to lift that restriction, allowing trait object types to be
+used for static dispatch even when the trait is not object safe.
+
+# Detailed design
+[design]: #detailed-design
+
+Today, the rules for object safey work like this:
```
ranweiler: `s/safey/safety`

Manishearth:
The documentation should mention fixing up https://doc.rust-lang.org/error-index.html#method-has-no-receiver

withoutboats:
What about that error message should change? You still can't have methods from the trait on the trait object type, just static methods from other traits (or inherent impls)

Manishearth:
Oh, nvm, I see now.

withoutboats:
This raises the question, if Foo is not object safe, can you impl Foo for Foo?

ranweiler:
> This would allow instantiating object types for non-object safe traits in unsafe code, by transmuting [...]

Is there _ever_ a context in this would be sensible? If not, it would be nice if we could modify the proposed design to continue to make this inexpressible, even in `unsafe` code.

Manishearth:
I'm overall not very fond of "let's make this inexpressible even in unsafe code", it _invariably_ ends up coming back to bite us. It's _very hard_ to know a priori what legitimate use cases there are for an API. I'm all for making things hard to do, but when we make things impossible it's often presumptuous. 

In fact, I'd argue for there to be a defined API with a stabilization path for constructing trait objects (`fn create_object<A, T: ?Sized>(x: *mut A) -> *mut T`) instead of needing to frob `std::raw` which is probably not going to stabilize (since the repr could change) and doesn't help you fetch the appropriate vtable pointers.

eddyb:
You can write that function today, just add an `A: Unsize<T>` bound and be warned that it's not limited to traits, but other than that it's just a refular unsizing coercion.

ranweiler:
> I'm overall not very fond of "let's make this inexpressible even in unsafe code", it invariably ends up coming back to bite us. It's very hard to know a priori what legitimate use cases there are for an API.

That makes sense to me! In this case, it seems like it might be worth thinking about a bit, since it is a question of preserving the status quo or not (vs. simply introducing new restrictions).

Suppose `Tr` is a trait that is _not_ object safe. The status quo is: we cannot— even with `transmute`— obtain a trait object for `Tr`. With this proposal, such a transmuting becomes newly possible, so we are expanding the API.

This is either _possibly_ sensible or it isn't, wrt what it is supposed to mean to be "object safe". If it _is_ possibly sensible, and the rules for object safety are _too strict_ right now, then this RFC adds a feature beyond just support for static dispatch: now you can (with some contortion) obtain a trait object that was previously unobtainable (at least after RFC 255). That's great! It also makes me ask: should there be a safe API for those cases?

If this is provably never sensible— e.g. it permits a _new_, previously-inexpressible violation of a contract that is _always_ supposed to hold— then to me, it seems there's a small amount of value in maintaining the status quo. This is definitely open to questioning, of course, since prior to RFC 255, any trait could be used to make a trait object. Maybe we lost something there, that we'd regain with this RFC. If so, it would be cool (but not mandatory) to have examples, if they exist.

Of course, since we only recover the ability to create trait objects from non-object safe traits in `unsafe` code, maybe this sort of "correctness" concern is just really out of place. Naively, it seems like something we should never be able to do this, by definition of "object safety". If there are valid use cases, that might suggest a need to modify the definition of object safety further, to accommodate those cases in safe code.

I definitely don't know, hence my interest in concrete examples.

withoutboats:
@ranweiler I think it depends on what you mean by "valid use cases." There are two different ways to interpret that:

* Is there a *good reason* to want to make a trait object from an object unsafe trait? The answer is that I don't know of any, but its impossible to affirmatively say there are none.
* Can you do this *without writing an invalid program*? That is, does this always just crash or lead to undefined behavior? Here, I have more certainty. You can do this without crashing; you can instantiate it with a null vtable pointer and just never try to access any items on it.

I still don't know why you would want to do that, but it would be a valid program AFAIK.

However, I tend to want to take a "we are all adults here" attitude toward unsafe code; you can do all sorts of inadvisible things. And I don't think restricting this would be very easy to implement, since we'd have to check object safety every time we learn the type of a value, rather than only when seeing these specific casts.

withoutboats:
Actually, I have a (contrived) example of something you could do.

Implementing *methods* on object-unsafe traits seems fine to me. You can never actually dereference their vtable or data pointers in those methods, because the object type doesn't implement the trait type, so it would be badly typed to call any methods from the vtable.

But that means you could totally define a method on an object unsafe trait, instantiate the object, and then call it:

```rust
trait ObjectSafe {
     fn safe(&self) -> u32;
}

trait NotObjectSafe {
     fn not_safe<T: Default>(&self) -> T;
}

trait FooBar {
     fn foobar(&self) -> u32;
}

impl FooBar for ObjectSafe {
     // performs dynamic dispatch through `self.safe()`
     fn foobar(&self) -> u32 { self.safe() }
}

impl FooBar for NotObjectSafe {
    fn foobar(&self) -> u32 {
        // NotObjectSafe does not implement NotObjectSafe, so `self.not_safe()` cannot be called
        0
    }
}

fn main() {
     unsafe {
         let obj: &'static NotObjectSafe = mem::transmute(TraitObject { data: ptr::null(), vtable: ptr::null() });
         obj.foobar()
    }
}

**Comment by withoutboats** [comment]
This RFC would basically let you use object unsafe traits analogously to opaque types; whatever data they carry you can't get to, but you can still operate on them.

**Comment by nikomatsakis** [comment]
Some thoughts:

A. I think that the rules proposed in this RFC are sensible and a reasonable "baby step" towards better object safety rules.

B. I would very much like to preserve the invariant that `Trait: Trait` holds so long as `Trait` is a valid type. This implies that "virtual calls" are kind of compartmentalized in an "automatic" impl that the compiler generates which handles virtual dispatch, rather than being a first-class thing.

This contributes to keeping MIR very simple. I might be overvaluing this, but I feel like there is value in having a simple "core language", and I'd like to see if we can hold the line there.

C. At present, we have this compromise, where a trait `Trait` can have non-object-safe methods and yet still be object-afe, but only if those methods require `Self: Sized`. This does invalidate invariant B because the type `Trait` does not implement `Sized`, and hence -- even in a generic context where we just know that we have some `T: Trait + ?Sized`, those methods cannot be used. Effectively this allows you to designate a "core group" of methods that are virtually dispatched, along with a bunch of others that must be statically dispatched. If you then combine this with an impl of `Trait` for `Box<Trait>` (or `&mut Trait` etc), then effectively get "full-featured" trait objects.

D. I think it's orthogonal from this RFC, to some extent, but I do like the idea of people being able to provide *their own* `impl Trait for Trait` impl in place of the compiler's automatic, virttual impl. This might allow you to make arbitrary traits 'object-safe'. You could imagine that you are allowed to skip specifying object-safe methods (which will then use virtual dispatch) but you must give definitions for all non-object-safe methods (which would then be statically run). This would presumably mean that they have some implementation that is defined in terms of the core, object-safe methods.

Because `Iterator` uses `fn(self)` methods, this wouldn't quite work in this case, but let me give it as a (broken) example anyhow, just to sketch out (and also because I've got finish this comment!). Imagine an alternate universe where `Iterator` had all the same methods it has today, but it did not have the `where Self: Sized` bounds on them. In that case, you might imagine supplying a impl like this:

```rust
impl<T> Iterator for Iterator<Item = T> {
    type Item = T;

    // No body means: delegate to a virtual call. Only possible
    // if the method is object-safe.
    fn next(&mut self) -> Option<T>; 

    fn map(self, ...) -> Map<Self, ...>{
        // this would presumably e more-or-less the same as the default impl;
        // but here you see the problem, that we can't do that because we need
        // to introduce some kind of pointer type for the trait object.
        //
        // if this were an `&mut self` method, it could be made to work.
    }
}
```

withoutboats:
> A. I think that the rules proposed in this RFC are sensible and a reasonable "baby step" towards better object safety rules.

> B. I would very much like to preserve the invariant that Trait: Trait holds so long as Trait is a valid type. 

These statements seem to be at odds, but I'm not certain that we mean the same thing. Under this RFC, `Trait: Trait` would not hold unless `Trait` is object safe, but `Trait` would *always* be a valid type. Are we using these terms differently from one another?

nikomatsakis:
@withoutboats 

> These statements seem to be at odds, but I'm not certain that we mean the same thing. Under this RFC, Trait: Trait would not hold unless Trait is object safe, but Trait would always be a valid type. Are we using these terms differently from one another?

You're correct. What I meant really was that `Trait: Trait` holds if you can create a trait object (and here I'm not referring to sorcery like `transmute`). Basically, I don't want you to be able to invoke a virtual method `m` except by doing `Trait::m(object)`. Right now, that suffices, and it also does under your RFC.

withoutboats:
@nikomatsakis Yes, I agree with that restriction

withoutboats:
@rfcbot fcp merge

As @nikomatsakis elucidated, there's a lot more we could do to make object safety more ergonomic (e.g. `where Self: Sized` on all items, `impl Trait for Trait` to fill in unvirtualizable items), but none of that conflicts with this RFC. There's very little to discuss here as far as I can see.

text/0000-object_safe_for_dispatch.md
```diff
@@ -0,0 +1,132 @@
...
+
+* If the trait (e.g. `Foo`) **is** object safe:
```
liigo: Are you means "If it is safe then...", or "It is safe if ..."?

withoutboats: the first

withoutboats:
I've merged this PR as RFC 2027, tracking issue is at https://github.com/rust-lang/rust/issues/43561

PR #2040, opened 2017-06-23
RFC - Zero-Sized References

EpicatSupercell:
[Rendered](https://github.com/EpicatSupercell/rfcs/blob/master/text/0000-zero-sized-references.md)

### DIFF
--- a/text/0000-zero-sized-references.md
+++ b/text/0000-zero-sized-references.md
@@ -56,7 +56,7 @@ Therefore, I propose to assume that whenever you find a self reference (or multi
 decide the reference is not Zero-Sized, since there most likely WILL be other data somewhere in the chain.
 This is most relevant to unions, which could have self-references and be instantiated at the same time.
 
-### *const and *mut pointers
+### `*const` and `*mut` pointers
 [pointers]: #pointers
 
 It is possible to convert a reference to a pointer. Currently, a reference to ZST points to an arbitrary location,
--- a/text/0000-zero-sized-references.md
+++ b/text/0000-zero-sized-references.md
@@ -1,6 +1,6 @@
 - Feature Name: zero_sized_references
 - Start Date: 2017-06-23
-- RFC PR: (leave this empty)
+- RFC PR: [#2040](https://github.com/rust-lang/rfcs/pull/2040)
 - Rust Issue: (leave this empty)
 
 # Summary
@@ -74,7 +74,7 @@ Converting in the other direction, the value of the pointer will be silently dro
 # How We Teach This
 [how-we-teach-this]: #how-we-teach-this
 
-For most rust users, this change will be invisible. Thier code will just become a tiny bit smaller.
+For most rust users, this change will be invisible. Their code will just become a tiny bit smaller.
 
 Users of unsafe rust might encounter this case. Therefore there should probably be a note in the nomicon that references might be optimized away for ZST.
 
@@ -94,6 +94,10 @@ The system that is now works well, and does not have to be changed.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
+
+### Definition of `&`
+
+Is the definition of `&` state that we guarantee it's a pointer? Or do we only promise you can use it to access the data?
 
 ### FFI + Escape mechanism
 
@@ -131,6 +135,9 @@ Safe rust code should be affected positively by this change. However unsafe code
 
 FFI is especially vulnurable to this change, as shown above.
 Are there better ways to deal with these errors without user involvement?
+
+(Unsafe code might break - are there examples of VALID code that breaks? Or does only invalid uses of references break?
+And what is our stance on breaking invalid code, assuming there is a large amount of it?)
 
 ### Specific examples - Pro
 
--- a/text/0000-zero-sized-references.md
+++ b/text/0000-zero-sized-references.md
@@ -1,0 +1,142 @@
+- Feature Name: zero_sized_references
+- Start Date: 2017-06-23
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+References to Zero-Sized Types (both shared and mutable) have been historically the size of `usize`.
+The proposed change is to make them also ZST.
+
+# Motivation
+[motivation]: #motivation
+
+References to any type in rust are represented as a pointer. Usually the pointer is smaller and faster to move around.
+However for Zero-Sized Types that only have a single value (for example `()` ) moving around is a no-op, and can be optimized away.
+Reading and writing the value is a no-op since it has only a single value anyway and therefore it carries no extra information.
+However, currently the compiler can't optimize away the pointer from data structures.
+
+Zero-Sized Types are useful for functions, lifetime guarantees and destructors,
+and references to them can be used to show these types "exist" (for references) or "you are the only one using it" (for mutable references).
+The actual value is meaningless and the representation should be optimized to be of size 0 as well.
+
+In addition, references to Zero-Sized Types often appear in polymorphic code, where they handle non-ZST as well.
+
+In both of these cases, it will be an advantage for the references to be Zero-Sized.
+
+# Detailed design
+[design]: #detailed-design
+
+## Calculating size
+[calculating-size]: #calculating-size
+
+Disclaimer: The writer of this RFC is not familiar with the inworkings of the compiler.
+
+
+Finding if a reference points to a ZST or not may not always be trivial.
+A struct with a reference to itself will not know it's size until after it knows the size to the reference.
+
+However, a couple of notes:
+
+```rust
+struct A<'a> (&'a A<'a>);
+```
+This struct cannot be instantiated, because the very first instance of it requires an instance of it to already exist.
+
+```rust
+enum A<'a> {
+    ZeroSized,
+    SelfRef (&'a A<'a>),
+}
+```
+The moment an enum has more than a single value, it cannot be Zero-Sized. Otherwise it isn't different than a struct.
+
+Therefore, I propose to assume that whenever you find a self reference (or multiple types referencing in a loop),
+decide the reference is not Zero-Sized, since there most likely WILL be other data somewhere in the chain.
+This is most relevant to unions, which could have self-references and be instantiated at the same time.
+
+## *const and *mut pointers
+[pointers]: #pointers
+
+It is possible to convert a reference to a pointer. Currently, a reference to ZST points to an arbitrary location,
+and when converting to a pointer the pointer recieves that arbitrary location.
+
+After this change, the reference will not hold any data. I propose that whenever a ZST reference is converted to a pointer,
+a warning/error be issued ("Warning: taking the address of a Zero-Sized Type is meaningless") and the pointer will recieve an address
+with the same algorithm that assigned an address to the reference in our current implementation.
+
+The purpose is to not break current code that might do this. We probably don't want to assign Null since for pointers it has
+a meaning that the value doesn't exist, which is different than "exist but no data".
+
+Converting in the other direction, the value of the pointer will be silently dropped - that value never had a meaning in the first place.
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+For most rust users, this change will be invisible. Thier code will just become a tiny bit smaller.
+
+Users of unsafe rust might encounter this case. Therefore there should probably be a note in the nomicon that references might be optimized away for ZST.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Breaking code
+
+Any place that assumed a reference holds a pointer might introduce bugs.
+
+FFI might behave differently than now, breaking code (see Unresolved questions).
+
+# Alternatives
+[alternatives]: #alternatives
+
+The system that is now works well, and does not have to be changed.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## FFI + Escape mechanism
+
+Do we need a way to remove the optimization? For example, if we have
+```rust
+struct ExternalStruct;
+
+let x: &mut ExternalStruct = ffi_function();
+other_ffi_function(x);
+```
+We might want to represent a pointer we got from FFI and checked it's not null as a reference to an object.
+Since that object is not accessable directly we represent it as an empty struct.
+However then the references to it are optimized away and we lose the pointer and then can't call the second function.
+
+Note: This break will not be silent. The safe wrapper for other_ffi_function will convert the reference back to pointer before
+using it, raising the warning/error of converting references to pointers.
+
+For that case, we might want to mark `ExternalStruct` as Non-Zero-Sized. Possibly
+```rust
+impl !Sized for ExternalStruct;
+```
+and maybe it should require `unsafe`.
+
+## Errors of conversion
+
+Do we want to give an error or a warning for converting references to pointers?
+
+The code above is an example where we break working code, so an error might be needed to show the significance.
+
+However, some conversions might be meaningless and wouldn't affect the execution, so the programmer might allow the conversion.
+
+## Mitigating breakage
+
+Safe rust code should be affected positively by this change. However unsafe code might break.
+
+FFI is especially vulnurable to this change, as shown above.
+Are there better ways to deal with these errors without user involvement?
+
+## Specific examples - Pro
+
+The RFC isn't well-justified until it has at least one detailed use case where it helps.
+Please share specific examples of code where Zero-Sized references are useful.
+
+## Specific examples against
+
+If you have specific examples where this change is detrimental, please share them.
--- a/text/0000-zero-sized-references.md
+++ b/text/0000-zero-sized-references.md
@@ -28,7 +28,7 @@ In both of these cases, it will be an advantage for the references to be Zero-Si
 # Detailed design
 [design]: #detailed-design
 
-## Calculating size
+### Calculating size
 [calculating-size]: #calculating-size
 
 Disclaimer: The writer of this RFC is not familiar with the inworkings of the compiler.
@@ -56,7 +56,7 @@ Therefore, I propose to assume that whenever you find a self reference (or multi
 decide the reference is not Zero-Sized, since there most likely WILL be other data somewhere in the chain.
 This is most relevant to unions, which could have self-references and be instantiated at the same time.
 
-## *const and *mut pointers
+### *const and *mut pointers
 [pointers]: #pointers
 
 It is possible to convert a reference to a pointer. Currently, a reference to ZST points to an arbitrary location,
@@ -81,7 +81,7 @@ Users of unsafe rust might encounter this case. Therefore there should probably 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-## Breaking code
+### Breaking code
 
 Any place that assumed a reference holds a pointer might introduce bugs.
 
@@ -95,7 +95,7 @@ The system that is now works well, and does not have to be changed.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-## FFI + Escape mechanism
+### FFI + Escape mechanism
 
 Do we need a way to remove the optimization? For example, if we have
 ```rust
@@ -117,7 +117,7 @@ impl !Sized for ExternalStruct;
 ```
 and maybe it should require `unsafe`.
 
-## Errors of conversion
+### Errors of conversion
 
 Do we want to give an error or a warning for converting references to pointers?
 
@@ -125,18 +125,18 @@ The code above is an example where we break working code, so an error might be n
 
 However, some conversions might be meaningless and wouldn't affect the execution, so the programmer might allow the conversion.
 
-## Mitigating breakage
+### Mitigating breakage
 
 Safe rust code should be affected positively by this change. However unsafe code might break.
 
 FFI is especially vulnurable to this change, as shown above.
 Are there better ways to deal with these errors without user involvement?
 
-## Specific examples - Pro
+### Specific examples - Pro
 
 The RFC isn't well-justified until it has at least one detailed use case where it helps.
 Please share specific examples of code where Zero-Sized references are useful.
 
-## Specific examples against
+### Specific examples against
 
 If you have specific examples where this change is detrimental, please share them.


SimonSapin:
Is it possible to do this in a library with `PhantomData<&'a ()>`?

EpicatSupercell:
The main purpose is to help with polymorphic code, that usually doesn't deal with ZST.

bugaevc:
[Rendered](https://github.com/EpicatSupercell/rfcs/blob/master/text/0000-zero-sized-references.md)

eddyb:
IIRC we haven't pursued this idea in the past because of the potential for breakage involved.

text/0000-zero-sized-references.md
```diff
@@ -0,0 +1,142 @@
...
+
+The purpose is to not break current code that might do this. We probably don't want to assign Null since for pointers it has
+a meaning that the value doesn't exist, which is different than "exist but no data".
+
+Converting in the other direction, the value of the pointer will be silently dropped - that value never had a meaning in the first place.
+
+# How We Teach This
+[how-we-teach-this]: #how-we-teach-this
+
+For most rust users, this change will be invisible. Thier code will just become a tiny bit smaller.
```
oli-obk: Thier -> Their

text/0000-zero-sized-references.md
```diff
@@ -0,0 +1,142 @@
...
+
+Users of unsafe rust might encounter this case. Therefore there should probably be a note in the nomicon that references might be optimized away for ZST.
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
oli-obk: It might also mess with monomorphization, because you can't know the size of a reference anymore without knowing the type. I'm not sure of the implications for generic functions. Currently `transmute<&T, usize>` works inside of generic functions. But we had the same issue with transmuting fn pointers, so there's some precedent for that.

text/0000-zero-sized-references.md
```diff
@@ -0,0 +1,142 @@
...
+
+Do we need a way to remove the optimization? For example, if we have
+```rust
+struct ExternalStruct;
+
+let x: &mut ExternalStruct = ffi_function();
+other_ffi_function(x);
+```
+We might want to represent a pointer we got from FFI and checked it's not null as a reference to an object.
+Since that object is not accessable directly we represent it as an empty struct.
```
oli-obk: iirc that is not how external structures should be represented, so breaking that at compile-time sounds like a win to me.

text/0000-zero-sized-references.md
```diff
@@ -0,0 +1,142 @@
...
+
+Do we want to give an error or a warning for converting references to pointers?
+
+The code above is an example where we break working code, so an error might be needed to show the significance.
+
+However, some conversions might be meaningless and wouldn't affect the execution, so the programmer might allow the conversion.
+
+### Mitigating breakage
+
+Safe rust code should be affected positively by this change. However unsafe code might break.
```
oli-obk: That's a dangerous statement. Unsafe code isn't any less stable than safe code. I think there should be some review of existing code that might break before deciding whether to do this. If all the unsafe code that breaks is some code that should never have been written in the first place because there's better ways to do it, then imo that would be ok.

text/0000-zero-sized-references.md
```diff
@@ -0,0 +1,142 @@
...
+
+FFI is especially vulnurable to this change, as shown above.
+Are there better ways to deal with these errors without user involvement?
+
+### Specific examples - Pro
+
+The RFC isn't well-justified until it has at least one detailed use case where it helps.
```
oli-obk: My opinion is that it's enough justification that this will solve all the issues around addresses and zsts once and for all.

ubsan:
I disagree completely with this change. `&` pointers are still pointers, and people trust that they don't change address.

text/0000-zero-sized-references.md
RalfJung: But that's already the case: `&mut Trait` has a different size than `&mut i32`.

eddyb: We do it based on sized-ness and the kind of "maybe-DST tail" when that is unknown.

ubsan:
This would also make references not `extern "C"` compatible, which is a major breaking change; if someone uses a ZST to represent an opaque type, for example.

RalfJung:
Shouldn't `extern "C"` use pointers anyway? C doesn't have references the way Rust does...

dlight:
"Breaking code" is a drawback big enough to make this Rust 2.0 material

SimonSapin:
Even in an hypothetical Rust 2.0 (which by the way I hope won’t happen for many years!), breaking code that uses `&ZST` as a way to implement `&Opaque` is disruptive enough that it shouldn’t be done.

A zero-size reference-like thing can be useful, but I think it should be a separate type. Something like:

```
// Does not implement Deref
pub struct ZeroSizedRef<'a, T>(std::marker::PhantomData<&'a T>);
```

EpicatSupercell:
A couple points I want to add (a bit meta):

- I am not as much pushing this RFC, as opening it for debate. I have heard strong opinions in both directions, and would like to have a discussion to come to an agreement. We could then point people towards this discussion whenever questions arise.

- The discussion of this RFC is expected to drift towards the definition of & and it's valid / invalid uses. That is intended and should be counted as on-topic.
---
Also, clarification about breaking code:

I want to find a solution that will not break valid unsafe code. I have no strong opinion about breaking non-valid unsafe code. However, since I have little knowledge about much of the unsafe code in the wild, the specifics of these should be discussed here.

glaebhoerl:
AFAICT there's two basic use cases here, which are both valid, but in tension:

 * If data structures with custom allocators store references to their allocators, and you define some global allocator with a ZST just to identify it, it would be unfortunate if data structures instantiated with it as their allocator had to physically store pointers which don't actually point to anything useful. (N.B. I'm not sure I have the details of this right, but there are probably analogous cases even if I don't.)

 * FFI code wants to use ZSTs (or similar) as stand-ins for foreign types, with `&ZST` containing no useful information on the Rust side, but faithfully preserving the memory address for when it is passed back into foreign code. (The `extern type` or `OpaqueData` RFCs might provide a not-technically-ZST option, but in any case we can't break code using the currently-existing one.)

How about an opt-in approach where authors can tag *specific* zero-sized types with "references to this type should also be ZSTs"? That satisfies the allocator use case (with a bit of manual annotation) without affecting any existing code. (My suspicion is that generic code would not pose a problem either because Rust implements it exclusively with monomorphization, but I haven't thought about it very deeply!)

eddyb:
You can transmute from/to `&T` *with a generic `T`*. So you'd need a trait that types opt out of and is present as a bound by default (i.e. like `Sized`), or some code will still break.

glaebhoerl:
@eddyb Is there actually generic code in the wild which does it? (I'm not doubtful, simply ignorant.)

eddyb:
Yes, we had to add a special-case because there were about a dozen (IIRC) crates which transmuted between pointers to `T: ?Sized`. I never actually checked how many would break if we pretended we didn't know the layout of pointers to `T: Sized`, but if the standard library doesn't abuse it then we could experiment with a small check for "does this type contain any type parameters?" (instead of "is it sized?").

text/0000-zero-sized-references.md
cramertj: Is sized-ness any easier to track than zero-sizedness?

text/0000-zero-sized-references.md
glaebhoerl: afaik that's currently the best way to represent external structures, see the discussion on the `extern type` RFC

ubsan:
@RalfJung no. If you want more info, hit me up on irc, I don't want to write it up here.

RalfJung:
@ubsan I would argue that the `extern "C"` function should have C-level types (raw pointers, no `Box` etc.), and then one writes a wrapper around it using more Rustic types. This avoids making assumptions about the layout if the Rust types.

But maybe that doesn't actually work so well, and anyway it's enough for some people to write libraries the way you propose to make this a problem.

glaebhoerl:
@eddyb Hmm, what if both pointers and references to such opted-in types were ZSTs, with a fixed address of 0xAlign? Would that solve the `transmute` problem, and would it introduce any others? It seems like it would avoid any kind of FFI issues unless people deliberately aim at their own feet.

eddyb:
> ZSTs, with a fixed address of 0xAlign

What does that mean? A by-value coercion? Then this will still silently break code which *doesn't* use `transmute`.

glaebhoerl:
@eddyb Can you give an example of what kind of code you're thinking of? I just mean that both `*const ZST` and `&ZST` would themselves be ZSTs, and whenever the actual value of the address is required (e.g. `*const ZST as usize`), it would be 0x1 (alignment was a silly thought, given that ZSTs don't really have differing alignments).

Hmm... that would break `ptr::null()` though. And making the value always be 0x0 instead would probably break `&T`, which is supposed to be non-zero. So I guess this doesn't really work out...

EpicatSupercell:
@glaebhoerl We really shouldn't touch `*const ZST`, since it carries extra data (null or non-null) it can't be ZST itself.

---

> there were about a dozen (IIRC) crates which transmuted between pointers to T: ?Sized.

@eddyb Pointers to `T: ?Sized` are not affected by this though? Since we only affect the references to structs of known size, references to trait objects are unaffected. References transmuted between each other will be affected only if the target is of different sizes... which IIRC would be an error anyway? And no sort of pointer would be affected. And I don't thik people transmute between `*const ZST` and `&ZST` since they could just use `as` for that. Are there other cases I didn't think of?

---

@ubsan The specific examples you showed didn't use a ZST. For examples with ZST, the next comes to mind:
```rust
struct OpaqueStruct;

extern "C" {
  fn some_fn(s: &OpaqueStruct);
}
```
but I think it could be enough to allow writing the next
```rust
unsafe impl !Sized for OpaqueStruct {}
```
and give an error that hints of the fix, something like `Can't use references to ZST in extern "C", consider adding "unsafe impl !Sized for OpaqueStruct {}"`.
Would this much be enough? The code will not silently break, and the fix is given + straightforward.

Note that writing the `unsafe impl !Sized` would probably disallow instantiating the struct on rust's side, but I can't even come up with a single case where you would send a pointer to a REAL ZST to C code.

eddyb:
@EpicatSupercell My point was that *there were even* transmutes between `T: ?Sized` pointers, so we had to support those, *on top of `T: Sized` pointers* (which have a known size).

EpicatSupercell:
@eddyb Is it valid to transmute between pointers to different known sizes? So transmute `&u8` to `&u16` for example.

eddyb:
@EpicatSupercell It has always been valid. Transmute only cares about *immediate* size, which is the same for all pointers to sized types.

> And I don't think people transmute between `*const ZST` and `&ZST` since they could just use `as` for that.

Strictly speaking, `transmute` was a mistake for almost everything it's commonly used for nowadays and I wish I hadn't seen some of the usage in the wild - the compiler should complain more.

Sadly https://github.com/rust-lang/rust/pull/32939 doesn't seem to link to the original crater report, so I can't show you the examples for `T: ?Sized` that I was mentioning earlier.

glaebhoerl:
> We really shouldn't touch `*const ZST`, since it carries extra data (null or non-null) it can't be ZST itself.

...Yes. That's what I said.

EpicatSupercell:
@eddyb Actually since `&ZST` and `*const ZST` ARE of different immediate size, the real question is whether we care about breaking the code that uses it like that. Honestly I'd hate to do that, but I feel like if we care TOO much about maintaining backwards compatibility, we'll end up worse than C++ in no time. If the use was never correct, AND it's in unsafe block, AND there are better solutions anyway, AND we give an error message instead of silently breaking, it might be fine to break it?

eddyb:
@EpicatSupercell If you can always detect it (which you can't because you can emulate `transmute` in stable code with raw pointers), then you can evaluate the impact *right now*, without waiting.

The typical usage scenario that we enabled was actually a transmute between two pointers with the *same* unknown potentially-unsized trailing type, so that would take over when working with `Sized` pointees, if their size cannot be computed statically, but only in some cases.

If you really feel confident enough that this scheme can work, I can implement the check (just the `transmute` compile errors), it's really just making the layout of `&T` depend on that of `T`.

djzin:
I suppose more generally, a reference is isomorphic to the type itself, meaning for any type (even implementing Drop) with sizeof(T) <= sizeof(*const T) and no interior mutability, any reference could be stored more efficiently as the value itself.

At least that's what I'd like to be able to say. There are plenty of reasons this won't work, which is a shame.

- `&7 as *const i32` is perfectly safe code which cannot be implemented if `&i32` is stored by value
- `&T as *const T as *const u8` is used often in unsafe code to copy bytes from one type to another
- `&T as *const T as usize` is valid today - something to think about
- A generic function with transmute will compile today but fail if even zero-sized references are introduced
- `&T as &Trait` is valid today - this could be worked out but of course you don't benefit from reducing the size of the type here - also some unsafe code is probably going to break if `&Trait` contains an inlined value instead of a reference

EpicatSupercell:
Let's see how these examples will change after the change!

- `&7 as *const i32` is unaffected, because not ZST
- `&T as *const T as *const u8` allows you to copy all 0 bytes of the ZST... from a random location (lets say 0x1). Since it doesn't matter where you copy 0 bytes from, no change here.
- `&T as *const T as usize` will now give the address assigned right now (0x1) instead of a different arbitrary address of a ZST.
- Generic function with transmute - between which two types? It matters.
- `&T as &Trait` is converting a normal internal pointer to a fat pointer. Now it generates a fat pointer from a new address (0x1) and the correct V-table. So, still correct.

djzin:
> `&T as *const T as usize` will now give the address assigned right now (0x1) instead of a different arbitrary address of a ZST.

This one I worry about. e.g. the following code would break:

```rust
let x = ();
let _spacer = 0;
let y = ();
assert!(&x as *const () != &y as *const ())
```

Aside - if you take the references first, you get the opposite behaviour, with the pointers equal to `0x1`

```rust
let x = &();
let _spacer = 0;
let y = &();
assert!(x as *const () == y as *const ())
```

djzin:
Also have to deal with the following behaviour:

```rust
let tuple = ((), ());
assert!(&tuple.0 as *const () == &tuple.1 as *const ());
```

EpicatSupercell:
Actually that code never was guaranteed to work, since LLVM is allowed to reorder variables how it sees fit.

djzin:
Ok if that's the case it's not the end of the world (for ZSTs at least) - just pointing out that there are some legitimate purely-safe-code programs that could in theory fail from this change - but if those programs were never guaranteed to work then that's just fine.

RalfJung:
> whenever the actual value of the address is required (e.g. *const ZST as usize), it would be 0x1 (alignment was a silly thought, given that ZSTs don't really have differing alignments).

They do: `[T; 0]` has the alignment of `T`.

Stebalien:
To provide a concrete example, relying on the address of `&ZST` is already a broken assumption because:

1. The standard library itself will *already* cook up arbitrary pointers to ZSTs at address 0x1.
2. [`ref_slice`](https://github.com/rust-lang/rust/issues/27774) was deemed legal but useless.

One can already throw away the address of a ZST reference in safe code by calling, `ref_slice(zst_ref).iter().next().unwrap()`. Any FFI code relying on the value of `&ZST as *const SomeCType` is already unsafe.

eddyb:
@Stebalien Fair enough, but that's not what *I personally* question, rather the behavior of generic code *assuming the size of `&T` for `T: Sized`*.

ubsan:
@RalfJung why write a wrapper around it if you don't need to? Also, again, types are better documentation than documentation. What's the difference between these three functions?

```
extern "C" {
  fn foo1(x: *mut T);
  fn foo2(x: *mut T);
  fn foo3(x: *mut T);
}
```

can you tell now?

```
extern "C" {
  fn foo1(x: Option<&mut T>);
  fn foo2(x: &Cell<T>);
  fn foo3(x: CBox<T>);
}
```

ubsan:
Also, I hate the assumption here that somehow, unsafe code is less important (and it's less of a bad thing to break it). Unsafe code is *far more* bad to break, because *they have assumptions that, if broken, will cause undefined behavior*.

RalfJung:
@ubsan: You are making lots of assumptions about data layout here, which I find very worrying. See e.g. <https://github.com/rust-lang/rfcs/pull/1758> -- there's no guarantee currently that transmuting between `T` and `Cell<T>` makes any sense, and yet that's exactly what you are doing.

ubsan:
@RalfJung I'm making exactly the assumptions that the standard library makes.

eddyb:
@RalfJung Isn't there an accepted `&mut T -> &Cell<T>` RFC? That pretty much guarantees it.

RalfJung:
@ubsan
> Also, these aren't transmutes, they're pointer-based type punning.

I don't see the distinction. Transmute could well be implemented using pointer type casting.

@eddyb 
> Isn't there an accepted &mut T -> &Cell<T> RFC? That pretty much guarantees it.

Ah, right, there is.

ubsan:
@RalfJung no it couldn't be, due to alignment requirements. pointer-based type punning is different from memcpy-based type punning.

Stebalien:
Personally, I'd love to just make `*const ZST` zero sized. This will cause lots of existing code to not compile but, IMO, that code is already wrong. The "correct" way to do this is to define an opaque type , as documented in the [nomicon](https://doc.rust-lang.org/nightly/nomicon/ffi.html) is `enum Opaque {}`.

However, given that we're living in the real world with people who will complain if their code breaks even if they're relying on UB, we probably can't do this :(.

---

But, to be pedantic, I'm going to try to redirect this discussion because it's coming to the right conclusions for the right reasons by the wrong arguments.

@ubsan 

>  Unsafe code is far more bad to break, because they have assumptions that, if broken, will cause undefined behavior.

There's no such thing as "causing undefined behavior". The behavior of unsafe code that makes invalid assumptions is *always* undefined, regardless of whether or not it happens to work and we *cannot* be scared of breaking code that relies on undefined behavior lest we never optimize/fix bugs. The questions to ask are:

1. Are we breaking *defined* behavior?
2. If we are changing undefined behavior, should we introduce a deprecation period/warning?
3. If we are changing *entrenched* undefined behavior, should just define it?

In this case, the only truly defined behavior I see this breaking is `size_of::<Option<&T>>() == size_of::<&T>()` (which we do specify). However, this assumption is only directly useful in FFI where using `&ZST`s are already provably broken.

The other possibly defined behavior that this would break is that `size_of::<&ZST>() == size_of::<*const ZST>()` (or even just `size_of::<&ZST>() > 0`).  However, I can't find anything documenting this behavior. On the other hand, `size_of::<&ZST>() > 0` could be implied by `size_of::<Option<&T>>() > size_of::<&T>()` (`&T` *can't* be a ZST bcause `Option<&T>` can't).

At the end of the day, I think this RFC will likely fall into category 3. Therefore, if we decide to close this RFC, we'll have a golden opportunity to actually define some currently undefined behavior (the assumptions we've decided are too important to break).

---

@eddyb 

> Fair enough, but that's not what I personally question, rather the behavior of generic code assuming the size of &T for T: Sized.

Is that actually defined anywhere? That is, isn't that why we haven't specified any of this stuff (so we have room to change and optimize representation)?

RalfJung:
> The "correct" way to do this is to define an opaque type , as documented in the nomicon is enum Opaque {}.

Oh no, the nomicon says that? I didn't know :( . I think that's horribly wrong. That opaque type states that it is *empty*. Passing an element of `&Opaque` around is super-dangerous. That's a whole separate discussion though, see <https://internals.rust-lang.org/t/recent-change-to-make-exhaustiveness-and-uninhabited-types-play-nicer-together/4602> and <https://github.com/rust-lang/rfcs/pull/1861>.

Ironically, the usual advice here (with which I agree) is to use an uninstantiable ZST instead:
```rust
mod opaque {
  struct Opaque { _priv: () }
}
use opaque::Opaque;
```

RalfJung:
> There's no such thing as "causing undefined behavior". The behavior of unsafe code that makes invalid assumptions is always undefined, regardless of whether or not it happens to work and we cannot be scared of breaking code that relies on undefined behavior lest we never optimize/fix bugs. 

We haven't even figured out what exactly these assumptions are yet, so you cannot blame anyone for violating them.

> Is that actually defined anywhere? That is, isn't that why we haven't specified any of this stuff (so we have room to change and optimize representation)?

The reason we haven't specified any of this is that we just don't know what the spec should be. Also see <https://internals.rust-lang.org/t/next-steps-for-unsafe-code-guidelines/3864>.

ubsan:
@Stebalien uh, no. That's not how undefined behavior works. In fact, we've defined that a reference is just a pointer in the past, so this would be making previously defined behavior different.

UB has a very specific definition. Unsafe code making invalid assumptions does not constitute.

ubsan:
(although unsafe code making invalid assumptions can *cause* UB. for example, relying on drop order can *result* in UB, but is not necessarily undefined behavior in and of itself.)

ubsan:
That there is what's known as "unspecified behavior", which is *very* different from UB.

Stebalien:
@RalfJung

> We haven't even figured out what exactly these assumptions are yet, so you cannot blame anyone for violating them.

We absolutely can "blame" (not really the right word) code for making assumptions about currently undefined behavior. If we haven't defined it, you can't rely on it.

> The reason we haven't specified any of this is that we just don't know what the spec should be.

You're assuming that there exists a golden spec out there that all rust code obeys, we just haven't figured out how to write it down yet; that's not the case. There's plenty of rust code that works today that will probably break tomorrow because we decide that something that code does is invalid (this happened, e.g., when we enabled field reordering).

---

@ubsan 

I was reacting to the debate about changing *unspecified* assumptions somehow introducing UB where it previously didn't exist but I clearly wasn't clear about that.

One *can* "cause" UB if we change specified behavior (modulo the semantics of  "cause UB" versus "cause programs to invoke UB"). I should have said one can't cause (programs to invoke) UB by changing *unspecified* behavior because programs relying on such unspecified behavior to avoid UB already invoke UB. Many of the assumptions I've seen discussed in this thread are, to the best of my knowledge, relying on unspecified behavior.

> In fact, we've defined that a reference is just a pointer in the past, so this would be making previously defined behavior different.

By "we've defined that a reference is just a pointer", I assume you mean that `&T` should have the same representation as `*const T` (not that `&T` is literally just a pointer because that's obviously not the case for DSTs). If so, this is already broken with or without this RFC given that stdlib already plays it fast-and-loose with the value of `&ZST`. However, I agree that that's a problem. As I've stated, I believe that the correct solution would be to make `*const T` zero sized but that would break lots of things.

> That there is what's known as "unspecified behavior", which is very different from UB.

You're right, I need to get my terms straight before correcting others. So, I'll try this again...

---

There's no such thing as "causing undefined behavior" *in programs that already rely on unspecified behavior to avoid invoking undefined behavior*. The behavior of unsafe code that makes invalid assumptions *(to avoid UB)* is always undefined, regardless of whether or not it happens to work and we cannot be scared of breaking code that relies on <s>undefined</s> *unspecified* behavior lest we never optimize/fix bugs. The questions to ask are:

1. Are we ~breaking defined~ *changing specified* behavior?
2. If we are changing ~undefined~ *unspecified* behavior, should we introduce a deprecation period/warning?
3. If we are changing entrenched ~undefined~ *unspecified* behavior, should just ~define~ *specify* it?

In this case, the only truly ~defined~ *specified* behavior I see this changing is `size_of::<Option<&T>>() == size_of::<&T>()` (which we do specify). However, this assumption is only directly useful in FFI where using `&ZST`s are already provably broken.

The other possibly ~defined~ *specified* behavior that this would break is that `size_of::<&ZST>() == size_of::<*const ZST>()` (or even just `size_of::<&ZST>() > 0`). However, I can't find anything documenting this behavior. On the other hand, `size_of::<&ZST>() > 0` could be implied by `size_of::<Option<&T>>() > size_of::<&T>()` (`&T` can't be a ZST because `Option<&T>` can't).

At the end of the day, I think this RFC will likely fall into category 3. Therefore, if we decide to close this RFC, we'll have a golden opportunity to actually ~define~ *specify* some currently ~undefined~ *unspecified* behavior (the assumptions we've decided are too important to break).

ubsan:
@Stebalien I'd argue that it's at least an implicit guarantee, if not an explicit one. Especially given that I use this in my own code, and it's likely being used in others' code...

Manishearth:
@RalfJung FWIW in Stylo we regularly use `&T` in FFI for the reason @ubsan mentioned; it's self documenting and also gives some lifetime bounds. It's been very useful. `&ZST` is a very common pattern in Rust FFI.

We also use `Option<&T>` for "nullable but valid" pointers, though we understand that that's not an optimization we are allowed to rely on (there are tests ensuring that it still works). I think we make similar assumptions about UnsafeCell (hence the `repr(transparent)` RFC), so that we can safely use aliased C++ types without noalias making things UB.

----

Given that Rust's unsafe code semantics are _currently_ unspecified, IMO it's not very productive to look at it from a "blame" point of view.  It is very easy to abdicate all responsibility for "incorrect" unsafe code, however correctness is a very fuzzy concept right now.

A more useful metric is "would it be reasonable to assume this works?". References being the same as a raw pointer internally is IMO a reasonable assumption, more so given the fact that the nomicon actually recommends using ZSTs for void (and this has been a common pattern within the community far predating the nomicon). OTOH, for example, expecting `Option<&T>` to always be a single word in size may not be as reasonable an assumption; which is why stylo takes all responsibility for this breaking and has tests.

Ultimately, it may be the _fault_ of the writers of unsafe code that their code breaks. However, if this is a more widespread pattern, we should still take it into account. I agree with @ubsan that if anything this is one of the places where we should apply more scrutiny, because this has the potential to cause UB if not handled right.

Not really saying anyone is doing so yet, but this is reminding me of the discussions on the drop order bug where many folks were ignoring the impact on existing code with "but well it's their _fault_ so it's not our problem", and I'd like to preempt that 😄

RalfJung:
> References being the same as a raw pointer internally is IMO a reasonable assumption,

That would not be violated by "also make pointers to ZST be themselves ZST". (Not that I think that's actually an option.)
I am curious though -- raw pointers to unsized types are already quite different from C pointers; I assume this means people have to be careful not to use pointers to unsized types in their FFI?

> given the fact that the nomicon actually recommends using ZSTs for void

Actually, [it does not](https://github.com/rust-lang-nursery/nomicon/issues/29) :(

> Not really saying anyone is doing so yet, but this is reminding me of the discussions on the drop order bug where many folks were ignoring the impact on existing code with "but well it's their fault so it's not our problem", and I'd like to preempt that 😄

Fully agreed.

rkjnsn:
Given current FFI uses, it's clear this can't be implemented without at the very least introducing and stabilizing some sort of opaque FFI type and having a transition period during which the compiler complains with increasing severity when a ZST is used with FFI. That seems both doable and desirable, given the many issues that ZSTs can have in an FFI context, despite currently being one of the least-bad alternatives.

That leaves the question of how much non-FFI code exists that relies on preserving and passing/storing the underlying pointer value for references to ZSTs, which I would imagine is not much.

Another issue brought up is generic code expecting `&T` to be a consistent size for all `T: Sized`. One thought I had here (which I admittedly have not fully thought through), is that if we say all ZSTs live at `0x1` and that creating an `&T` with a pointer value other than that is undefined behavior, then I *think* we can treat `&ZST` as semantically pointer sized, but still avoid passing/storing it in many cases.

That said, if we can find a way to make `&ZST` semantically zero sized, I feel it would be a cleaner solution.

nikomatsakis:
I do find the ideas in this RFC very appealing, but it seems to me that there are some pretty strong backwards compatibility concerns:

- There is tons of precedent that a `&T` and a `*const T` are representation-compatible, and that `&ZST` (or `*const ZST`) is a reasonable encoding of an opaque type.
- As @eddyb pointed out, we permit transmutes between `&T` and `&U` so long as `T: Sized` and `U: Sized`. The intention of these rules is to guarantee statically that the types have the same size. Clearly these rules would either have to be changed (causing code to stop compiling) or would become far less useful (causing weird errors).

In short, there seems to be a lot of precedent in favor of the assumption that `&T` (where `T: Sized`) is always a pointer, even if it it's sometimes a dummy one. And while I do think we have some amount of license to specify previously unspecified behaviors, which may cause breakage in unsafe code in the wild, we have to be pragmatic about it, and this change doesn't seem to offer enough upsides. If we did want to do this change, it seems like some more explicit opt-in would be better (e.g., a library type or something).

I therefore move to close this RFC (with thanks to the author for the enlightening discussion). 

@rfcbot fcp close

djzin:
I kind of agree. Probably should mark this down as a rust-2.0 wishlist idea. Then we could consider e.g. storing immutable reference to bool as a single bit/byte and other extensions if they make sense. Get that bit-twiddling efficiency out of the programmer's mind and into zero-cost abstractions ;)
null

SimonSapin:
We avoid significant breakage because it would be harmful to the ecosystem, not just to satisfy semver lawyers. “Rust 2.0”, if it ever happens, is not a license to introduce any and every kind of breakage.

djzin:
Of course - we don't want to just break everything - but in this case the breakage is in (specific) unsafe code only - the surface area of which I feel is acceptable for major version breakage. Basically any nontrivial change in e.g. how trait objects work, how nested enums are represented, extending the null pointer optimization - would require the same kind of breakage. So I feel the total benefits of allowing just this kind of breakage would be worth it. 
On a side note - IMHO it feels very "wrong" the way zsts are used for ffi. It seems like it breaks the assumptions of many potential optimizations anyway, leading to potential unsoundness down the line. There should probably at least be a better way to represent opaque types even if this particular optimization is never implemented.
null

le-jzr:
> There should probably at least be a better way to represent opaque types even if this particular optimization is never implemented.

https://github.com/rust-lang/rfcs/pull/1861 is already being merged, thankfully.

EpicatSupercell:
With the approaching resolution of #1861 we get a tool for FFI to solve some concerns. The other concerns are about unspecified behavior. I am not in favor of breaking code, so right now this RFC is inapplicable.

However, with the additions of epochs, a new possibility opens for us - allowing only references from a newer epoch to be ZST, while keeping all current code working. This RFC sounds fitting for a test run of epochs (assuming it's accepted in the first place). I'm not sure if now is a fitting time for it or not.

On a side note, due to personal circumstances, I didn't have the time to manage this RFC at all this past month :(

RalfJung:
A requirement for epoch's is that "if your code compiles in the previous epoch without warnings, then it will keep working in the next epoch". So we'd need to be able to reliably warn for any code that relies on the data layout of `&ZST`. That seems like a tough call...

EpicatSupercell:
It sounds possible to me. You make &ZST be Zero-Sized in the new epoch (= new code) and non Zero-Sized in the old epoch. When calling a function from old epoch or using a struct from old epoch you generate a value for the reference so it's no longer Zero-Sized, and when you receive a value from one you drop it so it's Zero-Sized again. Old code keeps working because it's never ZST, new code will not allow you to do illegal operations anyway.

Manishearth:
That's still not epoch-compatible -- you can't just "generate a value for the reference", the value has to still be correct if you don't want to break backwards compatibility. Remember, using `&ZST` for holding references to opaque C-managed types is a common, _recommended_ pattern for Rust FFI. Generating fake values will not fix this situation, it will only make it compile.

And the struct thing won't work at all, we can't have different representations for the same struct across crates; it would not allow them to be shared.


Epoches allow for breaking changes that are not in the core language, as defined in the RFC. I'd argue that this is definitely a change to the core language since it observably changes representations.

EpicatSupercell:
> different representations for the same struct 

Different representation depending on where it is defined, not used. Therefore a use of an epoch 1 struct in epoch 2 code will follow conventions of epoch 1 struct, while checking the size of an epoch 2 struct containing &ZST will return size 0 even in epoch 1 code.

> using &ZST for holding references to opaque C-managed types is a common, recommended pattern for Rust FFI

I come with the assumption that #1861 becomes the standard and '& extern type' becomes the recommended pattern. In that case &ZST stops being recommended and becomes an anti-pattern (and possibly warning/error).

le-jzr:
> I come with the assumption that #1861 becomes the standard and & extern type becomes the recommended pattern. In that case &ZST stops being recommended and becomes an anti-pattern (and possibly warning/error).

Antipattern or not, you still can't change the behavior in current epoch.
Rather, the compiler would have to emit warnings wherever code would not work for ZST references. This includes FFI, casts to/from usize, various transmutes, raw pointers of ZST types, etc.

The problematic part, however, is that generic code couldn't do those things on any generic references at all, unless a new auto-trait for non-ZSTs is added first (or ZSTs were removed from the `Sized` bound, which again, wouldn't be backwards compatible).

le-jzr:
With respect to interaction between epochs, imagine a specific case where epoch 2 code uses an epoch 1 FFI library that, for whatever irrelevant reason, passes around opaque payload as type `&()`. Epoch 2 code cannot just handle that `&()` from epoch 1 as a zero-sized reference, because it would conflict with the requirement that epoch 1 crates can be used in epoch 2 projects.

So for one thing, epoch 2 would need to understand its own `&()` as a different type from `&()` in epoch 1 functions. It would also need a way to use epoch 1 `&()` values, and name that legacy type in function signatures etc.

In other words, it would be possible, but it would be no small feat.

EpicatSupercell:
Guess it could be possible in that case to create a type called `Epoch_1::LegacyZST` (which is equal to legacy `&()`) and have the compiler complain that you try to assign a value of type `LegacyZST` to a type `&()`. Can't say it's very elegant, but is relatively simple.

My main reason I try and find a way to make this work is because I fear that if we decide a feature as simple (in my opinion) as this can't be used with Epochs, we will never be able to add any serious features without a major version (Rust 2.0). And with the collective desire to not go that route, I fear the language will stagnate even faster than C.

rkjnsn:
I have some thoughts. I'm afraid I have no conclusions. It seems to me there are two main ways to try to remove the overhead of references to ZSTs: make them ZSTs themselves (as proposed here), or try to eliminate them through optimization.

# Making them ZSTs

The former option is a breaking change to the language. I agree that it could probably be accomplished in a linking-compatible way by using something like a pair of `V1ZstRef<()>` and `V2ZstRef<()>` types. I imagine both would be usable from code in any version, with the first being equivalent to the original `&()`, and the latter to the new zero-sized `&()`.

The next question is, is it worth it? Even with a mechanism for making breaking changes to the language, the desire is still to keep them small and rare. We would also be committing to keeping both types of references in the compiler for the foreseeable future. Therefore, I would want to see concrete benefits demonstrated before going down that road, such a real work application where it saves a significant amount of memory or code size. I do agree that it seems reasonable to assume that using ZST references for FFI would become deprecated once we have a better alternative.

When I first came to Rust, I actually expected ZST references to also be ZSTs, and was surprised when they weren't. My preference would still be for them to be, I'm just not sure it's worth the breaking change. Aside from backward compatibility, I would like to understand better what generic patterns would be complicated, given that containers et al already have to special case ZSTs in most cases. Are there examples of code that currently works with ZSTs without special casing, but would have to deal with them specially after this change? Is there any code that would silently break?

# Eliminating ZST references as an optimization

This can be done either conservatively and aggressively.

## Conservatively

This would work by eliminating ZST references when the compiler can prove the address isn't observed, such as when a function never converts the reference to a raw pointer. This could be accomplished through metadata flags, and, like `impl Trait`, would mean that some changes to a function would require all users of it to also get recompiled. The easiest target for this optimization would be eliminating bare reference, such as `&self`, but with extended analysis, it would be possible to eliminate them from crate-private structs as well.

I imagine this approach would be able to avoid passing/storing ZST references here and there, but it wouldn't be reliable, and a change to a leaf function could force a ZST reference to be stored through most of the call graph.

## Aggressively

I see this approach as kind of a hybrid. ZST references would take up no space in a lot more cases, but code could still treat ZST references as pointer-sized in generic code. To make this work, the language would specify that for any `reference: &ZST`, `reference as *const ZST as usize == std::mem::align_of<ZST>`. That is to say, all instances of a given ZST are specified to reside at the lowest non-zero address consistent with its alignment. The language would further specify that attempting to create/modify a ZST reference to point at a different location would be undefined behavior. Importantly, this undefined behavior can easily be tested for by a sanitizer, which is desired for any UB in Rust.

This would allow the compiler to always eliminate passing and returning a bare `&ZST`, and would also enable compiler to avoid storing them in repr(rust) structs. Structs would then just take up the space necessary for there other members. The exception (to avoid unexpected ZSTs in generic code), is that structs containing nothing except ZST references and ZSTs would act just like bare `&ZST`s: they would act pointer sized, but the compiler could eliminate passing them or storing them.

This approach obviously requires that using `&ZST`s in FFI and other usages of ZSTs and opaque types to be deprecated and eventually removed in favor of #1861. I think this is desirable anyway, because using ZSTs as opaque types have a number of other disadvantages (which, indeed, served as the motivation for that RFC). This approach also wouldn't eliminate space usage in all places that making `&ZST` truly zero-sized would. E.g., unless optimized to take advantage of the new rules, `Vec<&ZST>` would still store a pointer-sized value for each reference, even though they're guaranteed to be the same.

Finally, this could break some non-FFI code. The most obvious example I can think of is something like using https://github.com/Diggsey/rust-field-offset with a zero-sized field. In debug mode, calculating the offset would underflow and panic. In release mode, applying the offset to a given struct would result in a pointer to a random location. Now, dereferencing a random point to a ZST is unlikely to actually cause any trouble, since there will never be an actual read and write to that location. Further the address would effectively get normalized back to the proper one when `apply` returned, since the compiler would have optimized out actually returning anything.

text/0000-zero-sized-references.md
pnkfelix: Sized-ness is "easier" to track than zero-sizedness, because a type parameter is assumed sized unless one explicitly declares otherwise via the `Sized?` marker. Once you have `<T: Sized?>`, then the generic code cannot make assumptions about the sized-ness of `T`.

aturon:
Thanks all for the discussion, and @EpicatSupercell for the RFC! I'm going to close as per discussion and FCP, but there are still alternative avenues to explore here!

kevincox:
> E.g., unless optimized to take advantage of the new rules, Vec<&ZST> would still store a pointer-sized value for each reference

Can someone explain why this is? What is preventing &ZST from being considered a ZST itself in this context?

SimonSapin:
@kevincox `&T` has been “non-null `*const T` with a lifetime” forever and some programs rely on it carrying a memory address.

For example they might use ZST’s to represent opaque types that are only manipulated through pointers or references (say, C++ objects whose memory layout might change across platforms). Using references over pointers is useful to get lifetime tracking. Then when it’s time to do something with that reference, it is cast back to a pointer and passed to some FFI function and needs to have preserved its address.

kevincox:
@SimonSapin That sounds like the general reason this RFC was denied, but I don't see how it translates to the `Vec` case with the "proposed" changes in that comment. That "proposal" would set `reference as *const ZST as usize == std::mem::align_of<ZST>` which would remove the ability of storing data in the reference.

vporton:
It is indeed possible to do zero-sized references without breaking existing code:
```rust
struct S<'a> {
    #[allow_zero_sized]
    pub e: &'a Empty,
}
```

Then `S` should be zero-sized.

Please implement this feature.

PR #2057, opened 2017-07-05
Add replace and swap functions to RefCell

notriddle:
[Rendered](https://github.com/notriddle/rust-rfcs/blob/refcell_replace/text/0000-refcell-replace.md)

### DIFF
--- a/text/0000-refcell-replace.md
+++ b/text/0000-refcell-replace.md
@@ -24,7 +24,7 @@ One of the most important parts of the ergonomics initiative has been reducing
 "type tetris" exactly like that `&mut *`.
 
 It also seems weird that this use-case is so much cleaner with a plain `Cell`,
-even though plain `Cell` is strictly a less popular abstraction.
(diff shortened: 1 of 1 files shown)


TimNN:
Would `try_replace` and `try_swap` be added as well, analogous to `try_borrow` and `try_borrow_mut`?

notriddle:
That's not part of the RFC.

Since the only way you can make the swap and replace functions trigger is by doing them at the same time as a borrow, I don't think you need the try variants that often.

aturon:
This looks great! I'm going to go ahead and propose to merge.

@rfcbot fcp merge

dtolnay:
@rfcbot reviewed

alexcrichton:
Ok looks like not much new discussion came up during FCP, so merging! I've opened a tracking issue at https://github.com/rust-lang/rust/issues/43570. Thanks again for the RFC @notriddle!

PR #2075, opened 2017-07-24
Add note of the RFC 2044 licensing process to README and update template

est31:
Also, add LICENSE-MIT and LICENSE-APACHE files to the repo.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
+   and distribution as defined by Sections 1 through 9 of this document.
+
+   "Licensor" shall mean the copyright owner or entity authorized by
+   the copyright owner that is granting the License.
+
+   "Legal Entity" shall mean the union of the acting entity and all
+   other entities that control, are controlled by, or are under common
+   control with that entity. For the purposes of this definition,
+   "control" means (i) the power, direct or indirect, to cause the
+   direction or management of such entity, whether by contract or
+   otherwise, or (ii) ownership of fifty percent (50%) or more of the
+   outstanding shares, or (iii) beneficial ownership of such entity.
+
+   "You" (or "Your") shall mean an individual or Legal Entity
+   exercising permissions granted by this License.
+
+   "Source" form shall mean the preferred form for making modifications,
+   including but not limited to software source code, documentation
+   source, and configuration files.
+
+   "Object" form shall mean any form resulting from mechanical
+   transformation or translation of a Source form, including but
+   not limited to compiled object code, generated documentation,
+   and conversions to other media types.
+
+   "Work" shall mean the work of authorship, whether in Source or
+   Object form, made available under the License, as indicated by a
+   copyright notice that is included in or attached to the work
+   (an example is provided in the Appendix below).
+
+   "Derivative Works" shall mean any work, whether in Source or Object
+   form, that is based on (or derived from) the Work and for which the
+   editorial revisions, annotations, elaborations, or other modifications
+   represent, as a whole, an original work of authorship. For the purposes
+   of this License, Derivative Works shall not include works that remain
+   separable from, or merely link (or bind by name) to the interfaces of,
+   the Work and Derivative Works thereof.
+
+   "Contribution" shall mean any work of authorship, including
+   the original version of the Work and any modifications or additions
+   to that Work or Derivative Works thereof, that is intentionally
+   submitted to Licensor for inclusion in the Work by the copyright owner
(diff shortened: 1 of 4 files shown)


0000-template.md
```diff
@@ -1,3 +1,14 @@
+<!---
+Copyright <fill in year> The Rust Project Developers. See the COPYRIGHT
```
joshtriplett: Please don't use "Copyright YYYY The Rust Project Developers. That isn't a legal entity. In the absence of copyright assignments (which don't seem like a good idea, and which the Rust project doesn't use in general), the copyrights would be held by the individual developers.

If you don't want to have people list individual copyright notices, you could just leave this part out and just keep the second paragraph. But please don't create a legally ambiguous situation by listing "The Rust Project Developers" as the copyright holder.

(You might, in an appropriate top-level file, note that copyrights are retained by the individual contributors. Listing every copyright holder is not strictly necessary from a legal perspective, though.)

eddyb: I mean, they probably copied it from `rust-lang/rust`, e.g. https://github.com/rust-lang/rust/blob/master/LICENSE-MIT#L1. Is that wrong too?

ubsan: @eddyb yes, probably :P

joshtriplett: @eddyb Yes, it is. I had a conversation about that with @brson a while back when a new such instance was added, and I *thought* some of them were addressed at that time (by switching to a notice that said something about copyrights being retained by contributors). But we need to address the rest of those notices.

est31: Yes, I've copied it over (I've literally typed `cp ../rust/LICENSE-* .` in my shell :p).

I don't think that saying "The Rust Developers" implies that this is some legal entity of some sort, but rather a collection of individuals of which each one continues to be copyright owner. Maybe I could remove the capitalization so that this becomes more obvious. What do you think?

Personally I think its better to say "The Rust Developers" than providing and maintaining a list of individuals who have made copyright protected changes to a work. Often these lists are terribly inaccurate.

joshtriplett: @est31 I *absolutely* agree that we shouldn't maintain an individual list. That's why I'm suggesting just saying "Copyrights are retained by the individual contributors.".  Legally, copyright notices are *not* required (though they must be preserved if included); copyright occurs whether you have a notice or not. But a notice of the form "Copyright YYYY Some Name" is a legally recognized copyright notice, and it implies that the named entity owns the copyright. I'm suggesting avoiding that form, which has a specific legal meaning.

(There's nothing wrong with phrases like "The Rust Developers"; the problem is using them in a copyright notice.)

est31: @joshtriplett hmmm maybe we could switch to a phrase similar to the one that servo uses (with adjustments ofc):

```text
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```

Generally I'd prefer to be consistent with `rust-lang/rust` headers, so maybe this could be discussed separately? We can change the header later on to whatever we agreed on to use in `rust-lang/rust`.

joshtriplett: @est31 That would be entirely fine with me. I'd just like to get this sorted out before we further propagate it.

est31: > That would be entirely fine with me.

Cool!

> I'd just like to get this sorted out before we further propagate it.

I'd prefer to have this PR merged sooner rather than later, as until its merged, I'll have to add every new contributor to the list of people requried to sign off (I already had to add killercup as his RFC got merged after I made the initial list). And some people can't contribute as long as this PR is unmerged because they insisted on only agreeing to licensing their past (not past and future) contributions under the license...

joshtriplett: @est31 I'm poking some folks right now about the rust-lang/rust repository. Meanwhile, would you have any objection to just dropping the first paragraph entirely, and only including the second, like this?


```
Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
<LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
option. This file may not be copied, modified, or distributed
except according to those terms. 
```

aturon:
I wonder whether we should have a COPYRIGHT file like the main Rust repo, rather than (or in addition to) putting that information in the README. Thoughts?

cc @rust-lang/core

est31:
@aturon when doing the PR I've wondered about that as well and decided on no because the COPYRIGHT file in the Rust repo consists of some notes about the copyright, and a list of licenses of third party code. As there is no third party stuff in this repo I've thought the COPYRIGHT file would be pretty empty, and my PR is already putting notes about copyright in the README, so it is documented.

joshtriplett:
Personally, I think the note in README suffices. rust-lang/rust needs COPYRIGHT mostly because of the various third-party licenses. In the absence of those third-party licenses, rust-lang/rust could just put the rest in the README.

alexcrichton:
As someone who's pretty ignorant to what all these files mean, I'm personally with the repo so long as it "look standard" in the sense that if you quickly glance at the repo you'll see a "LICENSE"-related thing at the top, so this seems sufficient for me at least!

PR #2123, opened 2017-08-24
implement FromIterator for tuples

debris:
[Rendered](https://github.com/debris/rfcs/blob/from_iterator_tuple/text/0000-from-iterator-tuple.md)

### DIFF
--- a/text/0000-from-iterator-tuple.md
+++ b/text/0000-from-iterator-tuple.md
@@ -1,0 +1,125 @@
+- Feature Name: from-iterator-tuple
+- Start Date: 2017-08-26
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Simplify conversion of `Vec<(_, _)>` and `HashMap<_, _>` to `(Vec<_>, Vec<_>)`.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


text/0000-from-iterator-tuple.md
```diff
@@ -0,0 +1,125 @@
...
+
+fn finished_processing_foo(f: Vec<ProcessedFoo>);
+
+fn finished_processing_bar(b: Vec<ProcessedBar>);
+
+fn old_usage() {
+	let foobar: Vec<_> = (0..10)
+		.into_iter()
+		.map(fetch)
+		.map(|(f, b)| (foo(f), bar(b)))
```
oli-obk: You can already write this as

```rust
let (processed_foo, processed_bar) = (0..10)
.into_iter()
.map(fetch)
.map(|(f, b)| (foo(f), bar(b)))
.unzip();
```

debris: I totally missed it! What do you think about `collect` falling back to `unzip`?

oli-obk: `FromIterator` is used for collections so far. I'm not sure how I feel about changing it to also work for tuples of collections. I'd rather generalize `unzip` in the future over all tuples if we ever get variadic generics.

debris: ok, thanks for the review! :)

PR #1935, opened 2017-02-28
Tuple-Based Variadic Generics

cramertj:
[Rendered](https://github.com/cramertj/rfcs/blob/variadic-generics/text/0000-variadic-generics.md)

cc @eddyb @sgrif

PR #2133, opened 2017-08-28
RFC - compiler-generated Clone impls for Arrays and Tuples

arielb1:
[rendered](https://github.com/arielb1/rfcs/blob/all-the-clones/text/0000-all-the-clones.md).

cc @eddyb @nikomatsakis @aturon

### DIFF
--- a/text/0000-all-the-clones.md
+++ b/text/0000-all-the-clones.md
@@ -1,0 +1,100 @@
+- Feature Name: all_the_clones
+- Start Date: 2017-08-28
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add compiler-generated `Clone` implementations for tuples and arrays with `Clone` elements of all lengths.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, the `Clone` trait for arrays and tuples is implemented using a [macro] in libcore, for tuples of size 11 or less and for `Copy` arrays of size 32 or less. This breaks the uniformity of the language and annoys users.
+
+Also, the compiler already implements `Copy` for all arrays and tuples with all elements `Copy`, which forces the compiler to provide an implementation for `Copy`'s supertrait `Clone`. There is no reason the compiler couldn't provide `Clone` impls for all arrays and tuples.
+
+[macro]: https://github.com/rust-lang/rust/blob/f3d6973f41a7d1fb83029c9c0ceaf0f5d4fd7208/src/libcore/tuple.rs#L25
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Arrays and tuples of `Clone` arrays are `Clone` themselves. Cloning them clones all of their elements.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Make `clone` a lang-item, add the following trait rules to the compiler:
+
+```
+n number
+T type
+T: Clone
+----------
+[T; n]: Clone
+
+T1,...,Tn types
+T1: Clone, ..., Tn: Clone
+----------
+(T1, ..., Tn): Clone
+```
+
+And add the obvious implementations of `Clone::clone` and `Clone::clone_from` as MIR shim implementations, in the same manner as `drop_in_place`. The implementations could also do a shallow copy if the type ends up being `Copy`.
+
+Remove the macro implementations in libcore. We still have macro implementations for other "derived" traits, such as `PartialEq`, `Hash`, etc.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The MIR shims add complexity to the compiler. Along with the `derive(Clone)` implementation in `libsyntax`, we have 2 separate sets of implementations of `Clone`. 
+
+Having `Copy` and `Clone` impls for all arrays and tuples, but not `PartialEq` etc. impls, could be confusing to users.
+
+# Rationale and Alternatives
+[alternatives]: #alternatives
+
+Even with all proposed expansions to Rust's type-system, for consistency, the compiler needs to have at least *some* built-in `Clone` implementations: the type `for<'a> fn(Foo<'a>)` is `Copy` for all user-defined types `Foo`, but there is no way to implement `Clone`, which is a supertrait of `Copy`, for it (an `impl<T> Clone for fn(T)` won't match against the higher-ranked type).
+
+The MIR shims for `Clone` of arrays and tuples are actually pretty simple and don't add much complexity after we have `drop_in_place` and shims for `Copy` types.
+
+## The array situation
+
+In Rust 1.19, arrays are `Clone` only if they are `Copy`. This code does not compile:
+```Rust
+fn main() {
+    let x = [Box::new(0)].clone(); //~ ERROR
+    println!("{:?}", x[0]);
+}
+```
+
+The reason (I think) is that there is no good way to write a variable-length array expression in macros. This wouldn't be fixed by the first iteration of const generics.
+
+OTOH, this means that making non-`Copy` arrays `Clone` is less of a bugfix and more of a new feature. It's however a nice feature - `[Box<u32>; 1]` not being `Clone` is an annoying and seemingly-pointless edge case.
+
+## Implement `Clone` only for `Copy` types
+
+As of Rust 1.19, the compiler *does not* have the `Clone` implementations, which causes ICEs such as rust-lang/rust#25733 because `Clone` is a supertrait of `Copy`.
+
+One alternative, which would solve ICEs while being conservative, would be to have compiler implementations for `Clone` only for *`Copy`* tuples of size 12+ and arrays, and maintain the `libcore` macros for `Clone` of tuples (in Rust 1.19, arrays are only `Clone` if they are `Copy`).
+
+This would make the shims *trivial* (a `Clone` implementation for a `Copy` type is just a memcpy), and would not implement any features that are not needed.
+
+When we get variadic generics, we could make all tuples with `Clone` elements `Clone`. With the right trick, we could also make all arrays with `Clone` elements `Clone`.
+
+## Use a MIR implementation of `Clone` for all derived impls
+
+The implementation on the other end of the conservative-radical end would be to use the MIR shims for *all* `#[derive(Clone)]` implementations. This would increase uniformity by getting rid of the separate `libsyntax` derived implementation. However:
+
+1. We'll still need the `#[derive_Clone]` hook in libsyntax, which would presumably result in an attribute that trait selection can see. That's not a significant concern.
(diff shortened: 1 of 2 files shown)


comex:
Perhaps add to Alternatives the option of waiting until the language features needed to make a proper generic implementation are available.  For arrays, that would presumably be [RFC 2000](https://github.com/rust-lang/rfcs/pull/2000); who knows about tuples.

cramertj:
@comex Tuples fall under #1935, but that RFC certainly won't land before the impl period.

text/0000-all-the-clones.md
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+}
+```
+
+OTOH, this means that making non-`Copy` arrays `Clone` is less of a bugfix and more of a new feature. It's however a nice feature - `[Box<u32>; 1]` not being `Clone` is an annoying and seemingly-pointless edge case.
+
+## Implement `Clone` only for `Copy` types
+
+As of Rust 1.19, the compiler *does not* have the `Clone` implementations, which causes ICEs such as rust-lang/rust#25733 because `Clone` is a supertrait of `Copy`.
```
kennytm: For convenience, please make these real links.

kennytm:
https://github.com/rust-lang/rfcs/pull/2132#issuecomment-325489815 refers me back to here 😂: You may want to add rules for function types and function pointers (with arity > 12) as well. These types already impl Copy, and thus should also impl Clone.



```
T1, ..., Tn types 
R type
"ABI" abi
(may need n > 12 || "ABI" ∉ {"C", "Rust"} to avoid existing impl)
----------------------------------------------------
extern "ABI" fn(T1, ..., Tn) -> R: Clone
extern "ABI" fn(T1, ..., Tn, ...) -> R: Clone
unsafe extern "ABI" fn(T1, ..., Tn) -> R: Clone
unsafe extern "ABI" fn(T1, ..., Tn, ...) -> R: Clone


f  fn item
----------------------------------------------------
f: Clone
```

text/0000-all-the-clones.md
```diff
@@ -0,0 +1,113 @@
...
+```Rust
+fn main() {
+    let x = [Box::new(0)].clone(); //~ ERROR
+    println!("{:?}", x[0]);
+}
+```
+
+~~The reason (I think) is that there is no good way to write a variable-length array expression in macros. This wouldn't be fixed by the first iteration of const generics.~~ Actually, this can be done using a for-loop (`ArrayVec` is used here instead of a manual panic guard for simplicity, but it can be easily implemented given const generics).
+```Rust
+impl<const n: usize; T> Clone for [T; n] {
```
kennytm: Missing a `T: Copy` bound?

arielb1: A `T: Clone` rule

arielb1:
> You may want to add rules for function types and function pointers (with arity > 12) as well. These types already impl `Copy`, and thus should also impl `Clone`.

These were non-controversial, so I didn't include them in the RFC. Maybe I should.

text/0000-all-the-clones.md
arielb1: Make what real links?

kennytm: @arielb1 Change 

```markdown
... ICEs such as rust-lang/rust#25733 because ...
```

to

```markdown
... ICEs such as [rust-lang/rust#25733] because ...

[rust-lang/rust#25733]: https://github.com/rust-lang/rust/issues/25733
```

Similar for references to "RFC #XXXX".

arielb1:
@kennytm Done.

arielb1:
@comex 

I don't think we are going to stay with the ICE forever - we need to do something about higher-ranked fn pointers, and in that case there's no good language reason not to also do `Copy` arrays and tuples.

est31:
Any reason on why `Clone` should get special treatment? There are many other traits which this RFC doesn't fix like `PartialEq`, `PartialOrd`, `Ord`, `Default`, `Debug`, `IntoIterator`.

Also, what to do once we have RFCs 2000 and 1935 merged, should the special case in the compiler be removed in favour of an implementation in libcore, and if yes, why not?

I'd like if the proposal got extended to all traits which are currently lang items. The situation with this RFC won't be perfect (e.g. `serde::ser::Serialize` will still not work for arrays with more than 32 elements), but it would be a much better situation than we have now.

It might look ugly to add so much manual stuff to the compiler but if the RFC is intended as temporary solution its worth it IMO.

kennytm:
@est31 `Clone` gets special treatment because of [`Copy`](https://github.com/rust-lang/rfcs/blob/master/text/0019-opt-in-builtin-traits.md#matching-and-coherence-for-the-builtin-types-copy-and-sized).

#2000 can get rid of the special case for arrays, and #1935 for tuples and simple function pointers, but the special cases must still stay for higher-ranked function pointers, function types, and closures if #2132 is merged.

est31:
@kennytm I see.

Either way, I still think it would be a good idea to implement this for other lang items as well, just so that we don't have to wait on const generics.

aturon:
Thanks, @arielb1, for this RFC! This nicely resolves a papercut, and doesn't seem to have any serious downsides. Let's start review for FCP.

@rfcbot fcp merge

scalexm:
@kennytm `Clone` indeed gets special treatment because of `Copy`, however as written in the RFC, the `Clone` impls for all `Copy` types will be done independently of this RFC, because of the ICEs (well actually the whole RFC was already implemented in [#43690](https://github.com/rust-lang/rust/pull/43690), somehow by mistake). So indeed I agree with @est31 that we should consider all lang items if possible.

starkat99:
Omg yes please, I ran headfirst into this inconsistency just this week (manually newtyped around it). This is definitely something that needs to be done!

eddyb:
@rfcbot concern why-only-clone

See https://github.com/rust-lang/rfcs/pull/2133#issuecomment-325685422 and https://github.com/rust-lang/rust/pull/44045#issuecomment-325294065 and the ensuing discussion.

nikomatsakis:
So, as for the "Why only `Clone`?" question, I do think there is a concern there, but I don't think it has to hold up this RFC, because I think to some extent we are already committed to a disparity between `Clone` and the other traits. This RFC moves the line where the disparity shows up, but doesn't resolve the disparity. I would love it if we evolved (eventually) a story to completely solve the disparity -- we've got parts of this story, but not the whole thing. Let me explain in detail, see what you think.

**How we already have a disparity:** We already say that `Copy: Clone`. We also say that all function pointers are `Copy`, as are all tuples types that consist only of copy things, fixed-length array types with copyable element type, etc. Those types must all be `Clone` -- at present they are not, and so we get an ICE. If we fix **just those cases**, it will already be true that `Clone` works for types like `[u8; 61412]` where `PartialEq` does not, so we will already have a disparity between clone and partial-eq.

**How this RFC changes the line:** However, under this RFC, we would have a somewhat bigger disparity, in that `Clone` would work for types like `[String; 61412]` as well. The complete list of things where the disparity would exist is as follows, I believe:

- Fixed-length arrays of large length (e.g., `[String; 61412]`);
- Tuples of large arity (e.g., `(String, ..., String)`);
- Function pointer types with bound regions (e.g., `fn(&u8)`).

**How we could resolve this disparity:** We have a number of ideas floating that can help. Const generics should enable one to write a `PartialEq` impl that applies to arbitrary length arrays. Variadics should resolve tuples of large arity. However, we don't have a plan that I know of around function pointer types with bound regions -- the only idea I've seen might be to have some built-in (and fundamental) trait `FunctionPointer` that is true for all function pointer types, so one can write a `impl<T: FunctionPointer> PartialEq for T` impl. This would probably require various improvements to specialization to make work.

**What are our options?** So it seems to me that we are stuck with a disparity between `Clone` and `PartialEq` for now. We can choose for that disparity to be limited to `Copy` types, or it can be extended to `Clone` types. The latter is strictly more convenient, obviously. I don't see a good reason to pick the former, personally.

(I feel like teaching etc would not be severely impacted by this: either way, in some cases you have to explain that `Clone` is a bit special, and the other traits haven't caught up yet; this is not something I would expect users to hit very early on, with the possible exception of arrays of large length, which thankfully is the one with the most solid plan for being resolved.)

Diggsey:
It seems like what's really wanted is a blanket `impl<T: Copy> Clone for T` - at the risk of opening a new can of worms, could this problem be addressed via changes to the coherence rules to make that possible? Since it's in core, it could rely on an unstable feature, like `#[fundamental]`.

scalexm:
@Diggsey it would be a solution to fix the ICEs indeed (and one that @nikomatsakis wanted to explore at some point) although another one was preferred (for the moment), but it would not address the main topics of the RFC, that is types like `[String; 2]` or `(String, String, String, String, String, String, String, String, String)` do no implement `Clone`.

nikomatsakis:
@Diggsey see my long series of blog posts on the topic ;) regardless, I do still believe we can enable that eventually through specialization, yes, though things are .. complex.

That said, I don't actually think it has all that much to do with this RFC. Or, to put it another way, that's effectively an **implement detail** -- @scalexm has implemented something that behaves analogously already, but it also went further, and implemented `Clone` for a broader set of types beyond those that are just `Copy`. We do have plans for how that might be achieved in a more general way for all traits, but those plans are not fully realized. Hence this RFC and ensuing discussion.

nikomatsakis:
So I talked to @eddyb on IRC, and it seemed like we reached an agreement that he was OK with this RFC, in combination with #2132, so long as:

- We "frame it" as: the compiler supplies `Clone` impls for compiler-synthesized types. Or, equivalently, for structural types + closures (which are *almost* structural). In other words, his primary concern that kicked off a lot of this discussion was that the compiler was growing code for generating impls in MIR that it didn't really need, but if we are to accept #2132, that code is needed, and he wants to be sure that there is logical way to draw the boundaries for what code **the compiler** generates and what code requires hand-written impls. 
- We are confident that the auto-generated impl will be equivalent to what we can do with const generics. I think we decided that we are (@scalexm and @arielb1 can confirm).

eddyb:
Also I should say that with variadic generics (or even today!), I like the possibility of reusing tuple impls for closures - if `(T, U, V): Hash` then so should a closure with 3 captures (cc @cramertj).

eddyb:
@rfcbot resolved why-only-clone

nikomatsakis:
@eddyb woah -- i.e., imagining that the compiler generates a bridge impl of that kind? That's a really interesting idea.

cramertj:
+1 for the bridge impl (I'm imagining in `let x = 5; let y = || println!("{:?}", x);`, the definition of `y` is something like `struct SomeAnonClosure((&i32)); impl Clone for SomeAnonClosure { fn clone(&self) -> Self { self.0.clone() } }`).

However, I don't want to limit the number of captures in a clone-able closure, so then the closure stuff is blocked on this change (making tuples of all sizes `Clone`/`Copy`able).

aturon:
This RFC has been **merged**! [Tracking issue](https://github.com/rust-lang/rust/issues/44496).

Thanks @arielb1!

vitiral:
@arielb1 the rendered link is now broken

PR #2147, opened 2017-09-11
Allow autoderef and autoref in operators

arielb1:
[Rendered]

[Rendered]: https://github.com/arielb1/rfcs/blob/df42b1df220d27876976b54dc93cdcb0b592cad3/text/0000-eye-of-sauron.md

### DIFF
--- a/text/0000-eye-of-sauron.md
+++ b/text/0000-eye-of-sauron.md
@@ -62,7 +62,7 @@ Allowing method-style autoref could make operator uses as clean as methods and f
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-Like methods, operators and indexing support automatic referencing and dereferencing. When you use an operator, the compiler will automatically add `&`, `&mut` and `*` operators to match the signature of the operator.
+Like methods, operators and indexing support automatic referencing and dereferencing. When you use an operator, the compiler will automatically add `&` and `*` operators to match the signature of the operator.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -248,6 +248,37 @@ We might want to allow mutable autorefs in more cases. That would be more consis
 ### Lifetime limits
 
 [RFC 2111] limits the lifetime of implicit autorefs to the containing expression. We might also want to do that here to avoid confusing escaping autorefs.
+
+### Improving impls
+
+Instead of adding autoderef for operators, we could try adding enough trait impls for operators.
+
+For example, with lattice specialization, we could have a trait with the following impls:
+```Rust
+trait MyAutoderef<T> {}
+impl<T> MyAutoderef for T {}
+impl<'a, U, V> MyAutoderef<U> for V where
+    V: Deref,
+    V::Target: MyAutoderef<U> {}
+// this is the special "lattice join" impl that is needed
+// to make the compiler shut up.
+impl<'a, T> MyAutoderef<T> for T
+    where T: Deref,
+          T::Target: MyAutoderef<T>
+{}
+```
+
+And then we could have impls
+```Rust
+impl<T: MyAutoderef<u32>, V: MyAutoderef<u32>> Add<U> for V {
+   // ...
+}
+```
+
+However:
+A. Lattice specialization requires that specialization will work well, which might take quite a bit of time to figure out.
+B. As written, the operator impls will all conflict with each-other and all other impls of operators and wildly break type inference. For every pair of types `U` and `V`, you'll have to convince the compiler that a type couldn't be both `MyAutoderef<U>` and `MyAutoderef<V>`, which would be non-trivial, or find some way to prioritize the impls, which will add more complexity and won't work cross-crate.
+C. Without autoref, types that are not `Copy` will be moved. For example, with this RFC you can index a `HashMap` with a `String` *without moving the `String`*, which can't be done by an impl.
 
 ### General Coercions
 
--- a/text/0000-eye-of-sauron.md
+++ b/text/0000-eye-of-sauron.md
@@ -96,7 +96,7 @@ Adjustment lists for all other operands (including the RHS of indexing operator,
 
 The adjustment lists selected are the lexicographically first pair of adjustment lists `(lhs_adjust, rhs_adjust)` (or with an unary op, just the `lhs_adjust`) such that
 A1. Both adjustment lists match the relevant regular expressions
-A2. Both adjustment lists must be valid to apply to their operand types. If, due to the presence of inference variables, it can't be determined whether these adjustment lists would be valid to apply, and we didn't find a smaller adjustment list that would apply, that is a compilation error (this is the "can't autoderef because of inference variables" case). 
+A2. Both adjustment lists must be valid to apply to their operand types. If, due to the presence of inference variables, it can't be determined whether these adjustment lists would be valid to apply, and we didn't find a smaller adjustment list that applies and might match the operator trait, that is a compilation error (this is the "can't autoderef because of inference variables" case). 
 A3. After applying both adjustment lists, the adjusted operand types are a potential match for the operator trait (if there is an ambiguity because of inference variables, it is counted as a match).
    A3.1. NOTE: the operator trait for overloaded indexing is `Index`, not `IndexMut`, even if indexing is done in a mutable context. rustc 1.20 is inconsistent in that regard.
    
--- a/text/0000-eye-of-sauron.md
+++ b/text/0000-eye-of-sauron.md
@@ -74,35 +74,33 @@ Like methods, operators and indexing support automatic referencing and dereferen
 
 Operator type-checking behaves similarly to method type-checking. It works as follows:
 
-## S1. Subexpression checking
+## Step 1 - Subexpression checking
 
 Both the LHS and the RHS (if binary) of the operator are first type-checked with no expected type.
 
 This differs from rustc 1.20, in which an expected type was sometimes propagated from the LHS into the RHS, potentially triggering coercions within the RHS. I should probably come up with an example in which this matters.
 
-## S2. Adjustment selection
+## Step 2 - Adjustment selection
 
 Afterwards, an adjustment list is selected for both operands as follows:
 
-Adjustment lists for the LHS of an indexing operation are selected from these matching the following regular expression:
+Adjustment lists for the LHS of an indexing operator (the `X` in `X[Y]`) are selected from these matching the following regular expression:
 ```
 "Deref"* "Autoref(Immutable)" "ConvertArrayToSlice"?
 ```
 
-Adjustment lists for all other operands (including the RHS of indexing operations) are selected from these matching the following regular expression
+Adjustment lists for all other operands (including the RHS of indexing operator, as well as all operands of all other operators) are selected from these matching the following regular expression
 ```
 "Deref"* ( "Autoref(Immutable)" "ConvertArrayToSlice"? )?
 ```
 
 The adjustment lists selected are the lexicographically first pair of adjustment lists `(lhs_adjust, rhs_adjust)` (or with an unary op, just the `lhs_adjust`) such that
 A1. Both adjustment lists match the relevant regular expressions
-A2. Both adjustment lists must be valid to apply to their operand types.
+A2. Both adjustment lists must be valid to apply to their operand types. If, due to the presence of inference variables, it can't be determined whether these adjustment lists would be valid to apply, and we didn't find a smaller adjustment list that would apply, that is a compilation error (this is the "can't autoderef because of inference variables" case). 
 A3. After applying both adjustment lists, the adjusted operand types are a potential match for the operator trait (if there is an ambiguity because of inference variables, it is counted as a match).
    A3.1. NOTE: the operator trait for overloaded indexing is `Index`, not `IndexMut`, even if indexing is done in a mutable context. rustc 1.20 is inconsistent in that regard.
-
-If the smallest adjustment can't be determined because of the presence of inference variables (because it is not obvious whether an adjustment list would be valid to apply), this is a compilation error.
    
-## S3. Fixups
+## Step 3 - Fixups
 
 After adjustments are selected, the following fixups are made. They do not affect adjustment selection.
 
--- a/text/0000-eye-of-sauron.md
+++ b/text/0000-eye-of-sauron.md
@@ -149,7 +149,7 @@ The picked assignment is the first assignment (in lexicographic order) from the 
 
 If no such assignment exists, it is a compilation error.
 
-A candidate might apply to an assignment unless subtyping the candidate's dispatchable argument types with the assignment's respective adjusted dispatchable argument types proves that one of the candidate's predicates can't hold (if the subtyping can't be done, that vacuously proves that the predicates can't hold).
+A candidate might apply to an assignment unless subtyping the candidate's dispatchable argument types with the assignment's respective adjusted dispatchable argument types, proves that one of the candidate's predicates can't hold (if the subtyping can't be done, that vacuously proves that the predicates can't hold).
 
 #### EXAMPLE 2. (Arithmetic)
 
@@ -559,7 +559,20 @@ We might want to allow for more general coercions than autoref and autoderef. Fo
 
 ### Appendix A. Method Dispatch
 
-This is supposed to describe method dispatch as it was before this RFC. 
+This is supposed to describe method dispatch as it is before this RFC.
 
-TBD
+Method dispatch does method-style lookup, with the following parameters:
+1. The only dispatchable argument is the receiver, which must be the first argument.
+2. Adjustments match the following regular expressions:
+    ```
+    "Deref"* ( ( "Autoref(Immutable)" | "Autoref(Mutable)" ) "ConvertArrayToSlice"? )?
+    ```
+3. The applicable candidates are described below.
+    3.1. All (non-static) methods with the right name from traits in scope are normal-priority candidates.
+    3.2. All inherent impl methods with the right name are high-priority candidates.
+    3.3. Let the "autoderef list" be the types that can be received by repeatedly autoderef-ing the receiver (there can be no inference variiables causing ambiguity there, because these are a part of the adjustment regular expression). Then, the following methods are added as high-priority candidates: **NOTE: now that I've wrote these out loud, I see that they might warrant changing in a post-Chalk universe**
+        3.3.1. For each type parameter `T` in the autoderef list, for each where-clause or supertrait of a where-clause in the parameter environment that is equivalent to one of the form `for<...> T: Trait<...>`, every method from that trait with the right name.
+        3.3.2. For each trait object `Trait<..>` in the autoderef list, for each bound on that trait object and each supertrait of it, every method from that bound or supertrait of a bound with the right name.
+        
+If one of the candidates added by rule 3.3 is selected, the respective where-clause or trait-object-bound is unified with the method's trait reference. If this fails, this is a compilation error.
 
--- a/text/0000-eye-of-sauron.md
+++ b/text/0000-eye-of-sauron.md
@@ -168,20 +168,70 @@ And have the obvious effect on values. Adjustments are ordered using the standar
 # Drawbacks
 [drawbacks]: #drawbacks
 
+### Inference Complexity
+
 The 2-argument inference adds more complexity to type-checking. It probably has some "interesting" and unexpected uses in some situations, and we need to gain some experience with it before stabilization.
 
+### Unexpected References
+
 The automatic references can make it less obvious when values are passed by reference or by value. Because operator traits are very general, the created autorefs can easily escape outside of the operator.
+
+First, I don't feel this is worse than the situation with method calls - they can also create unexpected references under pretty much the same situations.
+
+Second, and similarly to method calls, this is somewhat alleviated by the compiler picking the by-value implementation if the type matches, possibly causing move errors even if autoref would have worked:
+
+```Rust,ignore
+struct Bignum(..);
+// fair enough
+impl<'a> Add for &'a Bignum {
+    type Output = Bignum;
+    // ... impl, allocating
+}
+impl<'a> Add<&'a Bignum> for Bignum {
+    type Output = Bignum;
+    // ... impl, not allocating
+}
+
+let a = bignum1 + bignum2; // moves bignum1
+let b = bignum1 + 1; //~ ERROR use of moved value
+
+let c = &bignum3 + bignum4; // allocates, does not use bignum3.
+let d = bignum3 + 1; // works
+```
+
+However, this actually might tempt people to write only by-ref allocating impls for their types to avoid "unexpected" moves. We might want to gather more experience here.
 
 # Rationale and Alternatives
 [alternatives]: #alternatives
 
 The "Eye of Sauron" case is very annoying for mathematical code, and therefore is something we want to fix. This fix feels rather elegant and is contained to binary operators.
 
+## Alternative Solutions
+
 We need to use special coercion logic rather than the "standard" expected type propagation because we are coercing on the basis of trait impls. Therefore, options such as [RFC 2111] will not help in all cases. We could instead try to add trait-impl-based propagation, but that would not solve the "binary" nature of operator traits.
 
-[RFC 2111]: https://github.com/rust-lang/rfcs/pull/2111
+The essential reason we need to consider the "binary" nature is because of the `v.iter().filter(|x| x > 1)` case. If we did method-style inference, we would pick 0 dereferences for `x` - using the `PartialOrd<&&u32> for &&u32` impl combination - and will then fail when trying to coerce the `1`.
+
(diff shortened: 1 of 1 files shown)


aturon:
cc @cramertj

joshtriplett:
So, as the drawbacks section says:
> The automatic references can make it less obvious when values are passed by reference or by value. Because operator traits are very general, the created autorefs can easily escape outside of the operator.

Passing something by value when it was expected to be passed by reference can lead to severe performance issues. Passing something by reference when it was expected to be passed by value can *also* lead to severe performance issues. And both can make the behavior of code less transparent to the person reading it.

(Please note: I acknowledge that having to add `&` or `*` can *also* make code less transparent to the person reading it. I don't want to suggest that this won't ever help. I do want to observe that there's code that will become far less clear, in potentially critical ways, because of this.)

withoutboats:
@joshtriplett Can you give a concrete example? The RFC provides a concrete example of code that is made significantly less comprehensible by not having this feature, what is an example of something that would be made less comprehensible?

As far as I can see, this would only really trigger if you only have by-ref impls. Its true, you won't know if an operation using a type you don't know is by ref or by value, but why is that so critical to know that you can't look it up?

withoutboats:
I'd be a little more concerned about `&mut` coercions if it weren't for the fact that mutating a value in any of these positions would just be pathologically unclear code.

arielb1:
> I'd be a little more concerned about `&mut` coercions if it weren't for the fact that mutating a value in any of these positions would just be pathologically unclear code.

If we allow them, they will only be used in pathologically unclear code. If we don't allow them, no code that is not pathologically unclear would miss them.

Well, except for the "DSL maker" school that tries to overload syntax in every way possible. I'm not sure what's our best option there.

text/0000-eye-of-sauron.md
```diff
@@ -168,21 +168,71 @@ And have the obvious effect on values. Adjustments are ordered using the standar
...
+
+### Different set of operators
 
 Indexing is included in this feature basically because it behaves like the other operators. It's possible that this should not be done.
 
+### Different way to pick the adjustment list
+
+Lexicographic ordering feels like the right way to pick the adjustment list, but it might have unexpected edge cases which might justify a more complicated ordering.
+
+### Extended arithmetic fiixup
```
joshtriplett: s/fiixup/fixup/

withoutboats:
@arielb1 I guess someone could make `cin` and `cout`!

arielb1:
>  I guess someone could make `cin` and `cout`!

Not sure whether that's an argument for or against

> Does that just mean the bindings involved need to be `mut`?

It means that all the overloaded derefs in the middle need to implement `DerefMut`/`IndexMut`. e.g. if you have `x: Rc<Vec<u32>>`, then `x[0]` picks the `Index`/`Deref` impls, and then errors out because `Rc` does not implement `DerefMut`.

arielb1:
I'll try to clarify things tomorrow.

leodasvacas:
Your knowledge of this stuff is impressive!

The way this compares to #2111 should be clarified. Although the RFCs share similar motivations they only overlap in obscure cases and solve orthogonal issues.

What do you think of the alternative of waiting for intersection impls so that `Borrow` is both reflexive and has this impl like `AsRef` has which makes it go through references:
```rust
impl<T: Borrow<U>, U> Borrow<U> for &T
```
Then people would implement their operators like:
```rust
impl<T: Borrow<i32>, U: Borrow<i32>> Add<T> for U
```
This would solve many common cases. I wonder if it would also work for indexing.

burdges:
Just fyi there are some [analysis and benchmarks](https://github.com/isislovecruft/curve25519-dalek/issues/57) of using `#[inline(always)]` wrappers for the bignum cases.

text/0000-eye-of-sauron.md
```diff
@@ -0,0 +1,256 @@
...
+let a = (-A) * (one + nrr).invert();
+let z = u * (u.square() + A * u + one);
+```
+
+Allowing method-style autoref could make operator uses as clean as methods and fields, making code that uses them intensively far more ergonomic. It's also a fairly non-invasive extension.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Like methods, operators and indexing support automatic referencing and dereferencing. When you use an operator, the compiler will automatically add `&`, `&mut` and `*` operators to match the signature of the operator.
```
kennytm: Do you mean to include `&mut` here? The alternatives section mentions "Mutable autorefs ... is a bad idea".

text/0000-eye-of-sauron.md
```diff
@@ -0,0 +1,256 @@
...
+    
+    // I would like to write...
+    println!("{}", x[s]);
+    // But instead, I have to write...
+    println!("{}", x[&s]);
+```
+
+In this case, these are merely annoying papercuts. In some other cases, they can be a much worse problem.
+
+One of thim is Isis Lovecruft's "Eye of Sauron" case, which is a problem when using non-`Copy` bignums:
```
kennytm: `s/thim/them/`?

BTW I think linking to https://twitter.com/isislovecruft/status/839333846415998976 may provide some context what is meant by *Isis Lovecruft's "Eye of Sauron" case*.

text/0000-eye-of-sauron.md
```diff
@@ -0,0 +1,256 @@
...
+- The "standard" unary operators: `!`, `-`. This does *NOT* include the dereference operator `*`.
+- The indexing operator `...[...]`
+
+Operator type-checking behaves similarly to method type-checking. It works as follows:
+
+## S1. Subexpression checking
+
+Both the LHS and the RHS (if binary) of the operator are first type-checked with no expected type.
+
+This differs from rustc 1.20, in which an expected type was sometimes propagated from the LHS into the RHS, potentially triggering coercions within the RHS. I should probably come up with an example in which this matters.
```
oli-obk: This one's probably a crazy example, but the RHS is definitely dependent on the LHS and the case is rather degenerate:

```rust
    let byte = 0u8;
    let word = 1u16;

    let x = word + &byte.into();
    println!("{}", x);
```

This needs the `&` for *$reasons*

Found in https://github.com/rust-lang-nursery/rust-clippy/issues/2042

text/0000-eye-of-sauron.md
arielb1: This was left over from an old version of the RFC.

text/0000-eye-of-sauron.md
arielb1: > This one's probably a crazy example, but the RHS is definitely dependent on the LHS and the case is rather degenerate:

That's a trait-system issue. Operator overloading does succeed in this case. Without the autoref, you get these trait bounds:
```Rust
u16: Add<$0>
u8: Into<$0>
```

Because the compiler only considers each trait bound alone, it can't see that the only solution is `$0 = u16`.

If you add a reference, instead yoou get
```Rust
u16: Add<&'a $0>
u8: Into<$0>
```

And on the first bound, only the `impl Add<&u16> for u16` impl matches, which allows a solution.

arielb1:
> This would solve many common cases. I wonder if it would also work for indexing.

This won't solve the "Eye of Sauron" case with strings etc., because you need an autoref for that.

text/0000-eye-of-sauron.md
leodasvacas: This also won't coerce: `&&1 + 2`. As soon as you have two different implementations for the RHS it stops working. That coercion only fires in such rare situations that I wonder if it's existence is by design or an accident of implementation.

leodasvacas:
@arielb1 I don't follow, we have `Borrow<str> for String`. What example do you have in mind?  An advantage of that alternative is that only the cases the library wishes to solve are solved, in std [there is no consesus on String + String](https://internals.rust-lang.org/t/implement-add-for-string-string/4088) which this RFC would allow. It would be good to mention String + String in the RFC.

arielb1:
@leodasvacas 

But without autoref, operators would still take their operands by value, and `hashmap[string]` would still move `string`

arielb1:
Also, with this RFC, `x + y` would move `x` but not `y`, even if both are `String`s. You could still write `x + y.clone()` if you insisted on it, but the error messages won't lead you to it.

text/0000-eye-of-sauron.md
arielb1: In any case it's orthogonal to this RFC.

leodasvacas:
@arielb1 I get it now, thanks! For types that are not `Copy` this RFC has the big advantage of not moving them.

text/0000-eye-of-sauron.md
```diff
@@ -74,35 +74,33 @@ Like methods, operators and indexing support automatic referencing and dereferen
...
 ```
 
-Adjustment lists for all other operands (including the RHS of indexing operations) are selected from these matching the following regular expression
+Adjustment lists for all other operands (including the RHS of indexing operator, as well as all operands of all other operators) are selected from these matching the following regular expression
 ```
 "Deref"* ( "Autoref(Immutable)" "ConvertArrayToSlice"? )?
 ```
 
 The adjustment lists selected are the lexicographically first pair of adjustment lists `(lhs_adjust, rhs_adjust)` (or with an unary op, just the `lhs_adjust`) such that
 A1. Both adjustment lists match the relevant regular expressions
```
nikomatsakis: fwiw this list does not render well in markdown. Perhaps add some bullets?

nikomatsakis:
I'm basically in favor of this. But I want to raise a question about possible programming patterns just to make the argments in favor more specific.

When I work with `Rc` or `Arc`, I tend to follow a particular pattern:

- instances of that type that are stored in structs or other collections are by-value
- all local variables and parameters are references

So now if you want to "escape" a local variable or parameter (e.g., to store in a field), you do `self.foo = foo.clone()`, but otherwise you get to just pass around `foo: &Arc<u32>`. This makes rc and arc "feel like" they are copy, even though they are not. This really minimizes the number of times I have to explicit clone, and in particular makes it match a predictable pattern (whenever you are "escaping" a value outside of the current stack frame, essentially).

Do you think that this pattern would apply to bignums? Where are the painpoints when you do this? Presumably one would implement `Add` and so forth over references -- but I guess then doing things like `self.count + other_count` would actually be `&self.count + other_count`, to start.

In any case, as a historical note, the `+` and so forth in Rust used to *always* auto-ref, much like `==`. We chose to make them by-value to *permit* more efficient implementations: but at that time, I think we expected either that one would pass around `&`-references exclusively (and define ones impls over `&` references) or else that we would wind up adding autoref back at some point. It seems to me that by and large the jury is out here, and not having autoref is annoying. =)

@joshtriplett I am definitely wary of adding too much autoref sugar in general and finding that there are performance pathologies created, or just introducing plain confusion for users (as an example, when I teach ownership transfer, I *always* get a question about why `format!("{}", foo)` doesn't consume `foo`, since there is no `&`... this just goes to show you that people see the pattern and pick up on it).

That said, I do think that the damage here is somewhat mitigated by the fact that one must provide the impls to be used in the first place -- if you think a type is too expensive to pass by reference, or should *only* be passed by reference -- that is within your power to enforce (using newtypes, if necessary), no?

nikomatsakis:
I guess that one important question is to drill in on the backwards compatibility argument. I think that the reason that this would be considered backwards compatible is that, basically, we will try the current setup today, and only fallback to auto-ref in those cases where you would get an error today, is that right? A slight unknown to me then is how often this will cause confusion of the same kind that we get when coercions don't fire (because types weren't fully known at the time) but could have done -- but I think that's an orthogonal problem that we have to address regardless.

nikomatsakis:
@burdges 

> Just fyi there are some analysis and benchmarks of using #[inline(always)] wrappers for the bignum cases. tl;dr It might work, but it requires considerable care, which makes writing math libraries harder. Also if autojazz works reasonably then it might leave more room for optimization.

Can you provide more context? What performance loss is almost recovered? What is "autojazz"?

text/0000-eye-of-sauron.md
```diff
@@ -0,0 +1,286 @@
...
+
+Like methods, operators and indexing support automatic referencing and dereferencing. When you use an operator, the compiler will automatically add `&` and `*` operators to match the signature of the operator.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+"Operators" here refers to several typeck operators:
+- The "standard" eager binary operators: `+`, `-`, `*`, `/`, `%`, `^`, `&`, `|`, `<<`, `>>`, `==`, `<`, `<=`, `!=`, `>=`, `>`.
+- The "standard" unary operators: `!`, `-`. This does *NOT* include the dereference operator `*`.
+- The indexing operator `...[...]`
```
kennytm: What about compound assignment operators (`+=` etc)?

joshtriplett:
@nikomatsakis For clarification, is this *only* applied to operators which *exclusively* accept references and *not* owned values? In particular, if I have `foo + bar` where foo and bar are, say, `i32` and `&i32`, would this turn foo into a reference, or give an error that would lead the user to say "oh, I should fix my types"?

(Situation I have encountered multiple times: "huh, that's an odd error, wait, how have I been accidentally using double-references this whole time and everything worked?". I *want* that error, and I'd rather see it more often.)

withoutboats:
> (Situation I have encountered multiple times: "huh, that's an odd error, wait, how have I been accidentally using double-references this whole time and everything worked?". **I want that error, and I'd rather see it more often.)** [emphasis added]

Why do you want that? There's no difference semantically. For performance reasons, I expect the optimizer to usually be able to remove unnecessary references, and if it really matters, you need to be profiling anyway.

This happens all the time, for me most often when I wrote something like this:

```rust
vec.iter().filter(|x| x > 0);
```

Here, `x` has the type `&&i32`, and I usually have to add some `*` or `&` or whatever. This is just frustrating to me.

There's nothing I could even have done to eliminate the double reference here; the Iterator API was designed with the knowledge that the optimizer would remove these references anyway, and the only remaining downside is that I get blocked by an error and have to litter my code with symbols that don't provide actionable info.

withoutboats:
@arielb1 I have no objections to this RFC and I'm quite excited about it, but I wonder if you could add a "walkthrough" of how trait resolution will work in some of the motivating examples.

In particular, I think both the "uneven int references" and "only a by ref impl exists for this bignum" cases would be good to provide a clear walkthrough of how it will resolve to the correct impl.

e.g.

We have `vec.iter.filter(|x| x > 0);` and we know `vec: Vec<i32>`, so we know `x: &&i32`. We have these impls that are relevant (list the relevant impls), this is how this algorithm will correctly insert references/dereferences and resolve this.

We have `x + y` where x and y are both `Bignum` and an `Add<&Bignum> for &Bignum` impl. This is how the algorithm with correctly insert references/dereferences and resolve this.

This would just help demonstrate, without having to follow through your description of the algorithm and then performing it onesself, that this is a sufficient solution to get what I think we want from this RFC which is "operator arguments act like method receivers."

arielb1:
Yeah this RFC is basically "operators are slightly modified method receivers" + a description of half of method dispatch. I'll add walkthroughs.

joshtriplett:
@withoutboats I regularly end up with such errors that tell me I've ended up with types like `&String`, or `&&str`. Type inference makes it easy for this to happen without realizing. And I can often refactor the code so that I don't have such types. For instance, I might replace a `.iter()` with an `.into_iter()`, or match away a reference earlier so that I don't end up with double-references later.

burdges:
@nikomatsakis  I'd suggested the operators being `#[inline(always)]` wrappers around by reference calls, which worked in many situations, but maybe not all and even if it could work in all situations one obviously needed to be careful.  It'd be great if the compiler could make optimal autoref/autoderef decisions somehow.

aturon:
@joshtriplett It'd be super helpful to get two or three *concrete examples* (i.e. code snippets) where this arises and you feel it's important to be alerted.

text/0000-eye-of-sauron.md
burdges: Just to nit pick, there are no bignums in curve25519, only 256bit numbers, which do not count as bignums because they do not use `Box<>`.  Also, those numbers could easily be `Copy` so if they are not it's presumably Isis and Henry found that 32 bytes is already big enough for `Copy` to cause a performance penalty

burdges:
I'd imagine that `Sized` situations, including non-RSA crypto, permit the compiler to make an optimal choice between by-value and by-reference calls based purely upon the size of the type, but this choice might technically depend upon the target architecture, ala cache line size, register usage, etc.  

If otoh you're working with non-crypto bignums, then your type might looks roughly like
```
enum Bignum {
    Small(u64),
    Big(Vec<u64>)
}
```
so you'd want to pass by value. 

If you are working with crpytographic bignums for RSA, then you'll want roughly either a `Box<[u64]>` in your type, which you'd always pass by value, or monomorphize with `[u64; n]` for `n=256,512,1024,2048`, which you'd always pass by reference. 

As an aside, `Vec` already consumes 24 bytes, assuming `usize` is 8 bytes, while these values curve25519-dalek makes such a big deal to place behind references are only 32 bytes, but maybe the allocation eats so much that it doesn't matter for `Vec`.

arielb1:
Now with method dispatch documented. Turns out to be ugly and short.

withoutboats:
> I regularly end up with such errors that tell me I've ended up with types like &String, or &&str. Type inference makes it easy for this to happen without realizing. And I can often refactor the code so that I don't have such types. For instance, I might replace a .iter() with an .into_iter(), or match away a reference earlier so that I don't end up with double-references later.

This happens to me also. But you haven't really explained why you *want* to do this. With this RFC, like the match RFC, the end goal is really to get to a place where you just don't have to go through this edit-compile-debug cycle over errors that are not semantically significant to your program.

Its especially worth keeping in mind, with this RFC, that we're just applying a subset of the coercions we apply to method receivers to operators. That is, today you can write:

```rust
let foo: i32 = 100;
foo.is_positive()
```

We don't make you write:

```rust
let foo: i32 = 100;
(&foo).is_positive()
```

Even though `is_positive` is a by-reference method.

In fact, the `is_positive` method provides a pretty clear example of the benefit of this. These two pieces of code are equivalent:

```rust
vec.iter().filter(|x| x.is_positive())
vec.iter().filter(|x| x > 0)
```

But the second will not compile, and instead you'll have to do something like:

```rust
vec.iter().filter(|x| **x > 0)
```

This seems quite silly to me.

joshtriplett:
@withoutboats The difference is "not semantically significant" in the same sense that the difference between `&String` and `&str` is not semantically significant, or the difference between `HashMap` and `BTreeMap` is not semantically significant. For many programs, you'll get the same results, sure. But the performance could be drastically different. And I expect, in a Rust program, to actually be able to predict performance, and not have unnecessary indirections.

I managed to find the blog post I was looking for: https://medium.com/@robertgrosse/how-copying-an-int-made-my-code-11-times-faster-f76c66312e0f

*This* is the kind of problem that I think autoderef and autoref will make far more prevalent.

withoutboats:
@rfcbot fcp merge

(We'll keep talking through this issue of course @joshtriplett, but I wanted to prep this with a disposition befoer the lang team meeting tomorrow.)

withoutboats:
> The difference is "not semantically significant" in the same sense that the difference between &String and &str is not semantically significant, or **the difference between HashMap and BTreeMap** is not semantically significant. For many programs, you'll get the same results, sure. But the performance could be drastically different. And I expect, in a Rust program, to actually be able to predict performance, and not have unnecessary indirections.

> This is the kind of problem that I think autoderef and autoref will make **far more prevalent.**

(emphasis added) Both of the bolded portions of your comment seem exaggerated to me. Let me try to organize several interwoven lines of argument for why I think this will not have a significant negative performance impact.

### The micro-optimization opportunities are infrequent

First, let's define the scenario concretely. You have code like this:

```rust
vec.iter().filter(|x| x > 0)
```

In addition:

1. You never use `vec` again after this iterator is run.
2. Iterating by value would be significantly more performant than iterating by reference.

In other words, there's two very narrowing parameters here: you *could* perform the micro-optimization to remove the indirection, and it actually impacts performance to do it.

I'm more likely to, by necessity or laziness, just write this:

```rust
vec.iter().filter(|&&x| x > 0)
```

In other words, the code the compiler would have inserted for me *anyway*.

### If you knew you should perform this optimization, you're likely to do it regardless

The scenario you're describing is that because you received an error message, you realize that you can perform a micro-optimization. But this seems a little farfetched to me: most importantly, how do you know that its a micro-optimization, instead of a micro-pessimization?

If you *already know* that removing the indirection will be faster, why did you introduce it in the first place? I see two scenarios in which you might know about the performance impact:

1. You know the domain extremely well, and have a clear sense of the performance impact. In that case, I would expect you to be cognizant of this when deciding whether to iterate over that vector by value or by reference, and have done the right thing already.
2. You are profiling. In that case, you don't need an error message to call out the right thing to do, your profiling results will do that.

You might, today, remove indirection when you get errors like this, and that's fine. But are you really doing that for good reason, or based on a hunch that it will *probably* be faster?

It just seems bad to me to introduce a re-edit cycle in order to get you to think about the performance of an indirection. A re-edit cycle has a huge cost, and if the performance impact of something like this really matters, you'll know some other way.

### The compiler can optimize many cases already

I believe that in a lot of these cases the difference between the possibilities (especially when you mention something like removing a reference through a match) is going to be a wash for the compiler. I believe these dereferences are largely reorderable, removable, and optimizable, and LLVM will only get better at doing this for you over time.

### The scenarios impacted by this RFC are rare

Lastly, as I discussed previously, *we already do this kind of autoref and autoderef*. We currently do it for method receivers, this only extends that behavior to most operators. 

This means the only time you get these kind of "prompting errors" today that you wouldn't after this RFC is when you use a binary operator on the indirected value. If instead you called methods on it, you get no error, and if you used it in other ways that give you an error today, that won't change as a result of this RFC. So talk about "far more prevalent" seems like a large exaggeration.

---

Now, to take a contrary approach and try to lay out reasons why I think this is not only *not bad* but *very good*. :-)

### We have carefully benchmarked code in the wild that will benefit from this RFC

Today, curve25519-dalek has concrete examples of nigh-unreadable code because they have found that passing their FieldNumber type by value has a serious deleterious performance impact. We have a clear example of production code that's trying to be competitive with handwritten assembly - the absolute high end for Rust performance - and which is hurt by our current semantics.

If they weren't competing with handwritten assembly, the authors of curve25519-dalek might've been satisfied with the performance of by-value operations on their field numbers & not taken on the high syntactic burden of performing operations by reference.

After this RFC, people writing high performance math code will be more likely to actually benchmark the difference between by-ref and by-value code than they are today (they'll just have to change the impls to get the different codegen, rather than throwing syntactic salt into every use site). As a result, this RFC could actually improve the performance in crates where it really counts.

### Users who don't need to care don't get bothered

On the other end of the spectrum, as I've demonstrated there are many cases where this "prompting error" is not helpful:

* The only way to fix it might be exactly the code the compiler is going to generate (I think this is the majority of cases).
* Even if they have the choice, the performance different will not be significant to the user (I think this is the majority of the remaining cases).

So, today, we have an error being introduced that users can usually fix only one way, and even if they could fix it the other way, it won't matter to them - either because the performance difference is not significant, or because performance is not really critical in their application (especially if they're still learning Rust, and systems programming through Rust).

This error is just annoying to advanced users (which is bad), but to new users it can really be the end of their Rust experience. A lot of the "borrowing frustration" especially new users run against is not necessarily complex lifetime situations, but just this kind of "type tetris" juggling. Mitigating it can be a high impact way of easing the onboarding experience for people coming to Rust, which is probably our biggest user complaint.

djc:
I really think this can help, but I also have a concern that's been growing on me recently, which is that it seems the various RFCs (like [Match Ergonomics](https://github.com/rust-lang/rfcs/pull/2005)) are adding different ways of doing coercion in different places, and I wonder if these amount to local optimizations that make it hard for users to maintain a global understanding of what coercion is generally available.

My recent example is this:

https://users.rust-lang.org/t/troubles-creating-generic-function-interface-for-iterables-their-references/12595

It looks like this RFC will make "operators" more like the `.` operator in terms of coercion; should we instead aim for an RFC that makes all possible coercion sites/methods as similar as possible?

burdges:
Yes, I'd assume the current RFCs will incur a much longer path to stabilization, due to there being so many of them.  Ideally, any doing funny business with `&mut`, like match ergonomics and this one, should be held back from stabilization until the existing [unsoundness bugs](https://github.com/rust-lang/rust/issues?q=is%3Aopen+is%3Aissue+label%3AI-unsound) get fixed, although obviously plenty of people know vastly more about those bugs than me.

aturon:
(Signed off for @cramertj, who is on vacation)

aturon:
As per the commentary here and in the related [coercions RFC](https://github.com/rust-lang/rfcs/pull/2111#issuecomment-329522487), the lang team has decided to **close this RFC in favor of experimentation**.

That is, while we think this design is very plausible, there are a number of related changes being considered, and we'd like to land these all behind feature gates and gain experience with them, before coming back after the impl period with fresh RFCs.

Thanks @arielb1 for writing this so well and so quickly; I suspect much of the text will wind up in the final RFC.

aturon:
[Experiment tracking issue](https://github.com/rust-lang/rust/issues/44619).

PR #2174, opened 2017-10-15
Begin documenting initialization checking

arielb1:


text/2094-nll.md
```diff
@@ -1674,40 +1764,49 @@ fn access_legal(lvalue, is_shallow, is_read) {
...
+2. the loan path is, or has a **supporting prefix** `<base>.<field1>` that accesses a field of a union, and `lvalue` has a prefix of the form `<base>.<field2>` for a *different* field with the same base union.
+  - so: if `a.b` is a union with distinct fields `x` and `y`, deeply accessing a path `a.b.x.c` is illegal if `a.b.y`, `a.b.y.s`, or (in contrast with shallow accesses, and as long as both dereferences are of a `&mut T`) `*(*a.b.y.s).t` is borrowed. There `<base>` is `a.b`, `<field1>` is `y` and `<field2>` is `x`.
+3. `lvalue` is a **supporting prefix** of the loan path
   - supporting prefixes were defined earlier
   - so: reading a path like `a` is illegal if `a.b` is mutably
     borrowed, but -- in contrast with shallow accesses -- reading `a` is also
     illegal if `*a` is mutably borrowed
-    
+
+And the accessed data and all children in its lvalue tree must be definitely initialized.
```
Ixrec: Is this sentence fragment supposed to be connected to the "3." bullet point immediately above, or something higher up?

arielb1:
cc @pnkfelix

arielb1:
cc @nikomatsakis

text/2094-nll.md
```diff
@@ -1592,7 +1592,49 @@ our example, this would be the case if the type of `list` had been
...
+- after a consuming use of the lvalue `lv`, if the type of `lv` does not implement `Copy`, all lvalues in the lvalue subtree of `lv` are definitely uninitialized
+- after a MIR `drop(lv)` or `storagedead lv` statement, all lvalues in the lvalue subtree of `lv` are definitely uninitialized
+- after an `lv = <rvalue>` statement or an `lv <- <rvalue>` (drop+replace) statement, all lvalues in the lvalue subtree of `lv` are definitely initialized
+- after a non-panic return from an `lv = call <fn>(<args>)` statement, all lvalues in the lvalue subtree of `lv` are definitely initialized
+
+NOTE: it is possible for all the fields of a struct or tuple to be initialized without the struct/tuple itself being initialized, for example:
+```Rust
+let x: (u32, u32);
+x.0 = 4;
+x.1 = 5;
```
nikomatsakis: In the current borrow checker, [we do not accept this program](https://play.rust-lang.org/?gist=9f5087a9c82fbf836013615c5d9f2429&version=stable). Are there corner cases that I am overlooking here? 

Presuming there are not, I think that the current borrow checker basically enforces a rule like "Assigning to an lvalue LV requires that all prefixes of LV are definitely initialized", right?

Maybe we should just adopt this rule for now?

arielb1: This is indeed a difference from the current borrow checker. I see no problem with this rule.

nikomatsakis:
I think I'd like to move these changes over to https://github.com/nikomatsakis/nll-rfc/ -- I still view that as the 'official home' of the NLL RFC, at least until stabilization.

aturon:
@nikomatsakis What do you want to do with this RFC?

pnkfelix:
I'm going to try to follow @nikomatsakis 's request and turn this PR into a PR against  https://github.com/nikomatsakis/nll-rfc/ ; after I have created that second PR, I will close this one.

PR #2177, opened 2017-10-16
Update README.md to remove the mentioned shepherd

saschagrunert:
I guess the "shepherd" can be removed since he/she does not exist any more. :+1:


cramertj:
@saschagrunert A number of teams have shepherds. See [the Rust Team page.](https://www.rust-lang.org/en-US/team.html)

saschagrunert:
Alright, then I guess we have to re add the _role of the shepherd_ or just a minimum description about what a shepherd_ is, right? It seems a bit out of context here.

aturon:
The change is accurate regardless.

PR #2185, opened 2017-10-21
RFC: movable array iterators

ishitatsuyuki:
[Rendered](https://github.com/ishitatsuyuki/rfcs/blob/array_into_iter_move/text/0000-array_into_iter_move.md)

### DIFF
--- a/text/0000-array_into_iter_move.md
+++ b/text/0000-array_into_iter_move.md
@@ -1,0 +1,66 @@
+- Feature Name: array_into_iter_move
+- Start Date: 2017-10-21
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Implement into_iter on array types (not reference, so it moves).
+
+# Motivation
+[motivation]: #motivation
+
+Arrays are particularly useful in `flat_map`, but currently a Vec is required as a movable
+iterator because arrays only have slice iterator semantics. Obviously this is not optimal.
+
+Now with `ManuallyDrop` implemented, it should be possible to implement moving into_iter on arrays.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+You can convert an array into an iterator with move, just like Vec:
+
+```rust
+for x in [1, 2] {
+    // x is {integer} instead of &{integer}
+}
+```
+
+```rust
+let v: Vec<_> = (0..5).flat_map(|x| [x, x*2]).collect();
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Proposed implementation:
+
+- a type `array::IntoIter` is created. The `array::into_iter` method is implemented.
+- Inside the method:
+  * Transmute the array so it has `ManuallyDrop<T>` signature. This should be memory safe.
+  * Create a `IntoIter` struct with contents moved.
+- Inside the iterator:
+  * Keep track of valid range (index) and move items out as `next()` is called.
+  * Don't forget to drop the items if the iterator itself is dropped in middle.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is not 100% backwards compatible as it changes the signature of into_iter (only if directly
+called like below). A crater run is required.
+
+```rust
+[1, 2].into_iter();
+// This was originally yielding references, but now values.
+```
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+TBD
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+TBD
--- a/text/0000-array_into_iter_move.md
+++ b/text/0000-array_into_iter_move.md
@@ -36,15 +36,18 @@ let v: Vec<_> = (0..5).flat_map(|x| [x, x*2]).collect();
 
 Proposed implementation:
 
-- a type `array::IntoIter` is created. The `array::into_iter` method is implemented.
+- a type `array::IntoIter<T>` is created. The `array::into_iter` method is implemented.
+- 0..32 variant is first implemented using macros, then migrated to const generics once it lands.
 - Inside the method:
   * Wrap the array with `ManuallyDrop`.
   * Create a `IntoIter` struct with contents moved.
 - Inside the iterator:
   * Keep track of valid range (index) and move (`ptr::read`) items out as `next()` is called.
   * Don't forget to drop the items if the iterator itself is dropped in middle. This should be done with `drop_in_place`.
-  
-We should add a lint for redundant Vec in clippy to promote the use of this.
+
+Before this lands, a deprecation warning should be implemented in either rustc or clippy to minimize the breakage. (See below for compatibility issues.)
+
+We should also add a lint for redundant Vec usage (instead of array) in clippy to promote the use of this.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -65,4 +68,4 @@ TBD
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-- What should the type signature look like? Possibly utilizing const generics.
+TBD
--- a/text/0000-array_into_iter_move.md
+++ b/text/0000-array_into_iter_move.md
@@ -38,11 +38,13 @@ Proposed implementation:
 
 - a type `array::IntoIter` is created. The `array::into_iter` method is implemented.
 - Inside the method:
-  * Transmute the array so it has `ManuallyDrop<T>` signature. This should be memory safe.
+  * Wrap the array with `ManuallyDrop`.
   * Create a `IntoIter` struct with contents moved.
 - Inside the iterator:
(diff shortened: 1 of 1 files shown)


SimonSapin:
I very much want this, but what is return type of `[T; N]::into_iter`? I think it should be `IntoIter<T, N>`, with [const generics](https://github.com/rust-lang/rust/issues/44580). So this needs to wait for that to be implemented.

Or maybe `IntoIter<A> where A: Unsize<[T]>` might work, but stabilizing with `Unsize` itself unstable doesn’t feel right.

----

> Transmute the array so it has ManuallyDrop<T> signature. This should be memory safe.

I believe that in a method taking `self`, the array can be safely moved into `ManuallyDrop::new(…)` without transmute. Hopefully this move optimizes to a no-op?

ishitatsuyuki:
Please prefer line comments so I can respond without copying :)

> I believe that in a method taking self, the array can be safely moved into ManuallyDrop::new(…) without transmute. Hopefully this move optimizes to a no-op?

I think there's a need to be able to drop each elements individually (in other words, only in the range that is not moved out). So we need ManuallyDrop on each element, and trasmuting should be the easiest so far.

SimonSapin:
You can call `drop_in_place(…)` selectively on individual items or a sub-slice inside of `ManuallyDrop<[T; N]>`.

ishitatsuyuki:
Good. I will update the text later.

text/0000-array_into_iter_move.md
```diff
@@ -0,0 +1,68 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is not 100% backwards compatible as it changes the signature of into_iter (only if directly
+called like below). A crater run is required.
+
+```rust
+[1, 2].into_iter();
+// This was originally yielding references, but now values.
+```
```
bluss: This is pretty much ensured to break actual code in many places, both in crater (that we can test) and lots of code we can't test. Maybe we can begin warning for this long before the new iterator is implemented?

ishitatsuyuki: Personally speaking, I don't think people would use `into_iter` despite the fact it yields references, and they would just `iter` instead. Thus I think fixing after a crater should be sufficient.

bluss: We don't reach all rust code in the world with crater. If the break is wide spread, we have to be more careful.

main--: And what about `for &x in [1,2,3]`, isn't that just `into_iter`?

bluss: That does not compile today I think so it's not a backwards company concern.

main--: Oh I see. `[1].into_iter()` currently resolves to the impl for `&[_; _]` which the new `[_; _]` impl would override, right?

cristicbz:
@SimonSapin 

> I very much want this, but what is return type of [T; N]::into_iter? I think it should be IntoIter<T, N>, with const generics. So this needs to wait for that to be implemented.

This could be done without stabilising `Unsize`, by doing:
```rust
impl<T> Iterator for IntoIter<[T; 1]> { ... }
impl<T> Iterator for IntoIter<[T; 2]> { ... }
impl<T> Iterator for IntoIter<[T; 3]> { ... }
// ...
impl<T> Iterator for IntoIter<[T; 32]> { ... }
``` 

with all the impls calling to some shared unsafe method to avoid duplication. Has a nice migration story for when const generics land,  just replacing with

```rust
impl<T, const N> Iterator for IntoIter<[T; N]> {}
```

text/0000-array_into_iter_move.md
bluss: Yes!

ishitatsuyuki:
Please take a look, I think I have addressed the concerns.

SimonSapin:
This looks great to me, thanks for writing it up!

sinkuu:
[`Drop` impls cannot be specialized](https://github.com/rust-lang/rust/pull/23638), so we may have to use `Unsafe` or `FixedSizeArray` if we implement this RFC before const generics.

```rust
struct IntoIter<T> { ... }
impl<T> Drop for IntoIter<[T; 0]> { ... }
impl<T> Drop for IntoIter<[T; 1]> { ... } // error: Implementations of Drop cannot be specialized
...

// I expect this will work
struct IntoIter<T, const N: usize> { ... }
impl<T, const N: usize> Drop for IntoIter<T, N>

// This works today
struct IntoIter<T, A> where A: FixedSizeArray<T> { ... }
impl<T, A> Drop for IntoIter<T, A> where A: FixedSizeArray<T> { ... }
```

eddyb:
@sinkuu What do you mean by `Unsafe`?

ishitatsuyuki:
Presumably Unsize.

ishitatsuyuki:
Should we make multiple IntoIter structs until const generic lands? Like this: `IntoIter32<T>`.

I think this should resolve the specialization problem, alongside reducing the template complexity, so we can store pointers (or unsafe slices) inside the struct.

SimonSapin:
> `IntoIter32<T>`

I want this feature, but I don’t think I want to stabilize such an API.

It’d be harder to generalize once const generics land, too.

> so we can store pointers (or unsafe slices) inside the struct

I don’t understand what you mean here. The point of `into_iter` is that the items move when the iterator moves. Pointers should not be involved.

ishitatsuyuki:
> but I don’t think I want to stabilize such an API.

I'm afraid that even `IntoIter<T>` has to change into `IntoIter<T, const N>` so signature change is unavoidable.

> I don’t understand what you mean here.

We need to save the range we have consumed with the iterator. It can be either usize-indexed or just start-end pointer.

SimonSapin:
Changing the signature of stable APIs is not acceptable, so I’m afraid that this implies we should not stabilize anything here without const generics.

Self-referential pointers would become invalid when the iterator is moved. It has to be indices. (Two of them, or a `Range`, to support `DoubleEndedIterator`.)

ishitatsuyuki:
I've come up with a partial stabilization plan, please let me know if this is possible.

As the signature of the iterator struct is subject to change, we will be keeping that struct(s) unstable until const generics land. I will call this struct `UnstableIntoIter` below.

`IntoIterator` is implemented like below, using the length of 32 as an example:

```rust
impl<T> IntoIterator for [T; 32] {
    type Item = T;
    type IntoIter = UnstableIntoIter32<T>;

    fn into_iter(self) -> T { ... }
}
```

I have no idea if this implementation will be accepted as stable in the compiler, so we can use it without the ability to annotate the type.

kennytm:
Instead of a specific `IntoIter` or `UnstableIntoIter` method, could it just return an existential type `impl Iterator<Item=T>`?

scottmcm:
Would it be compatible to replace
```rust
struct IntoIter0<T>{ .. };
..
struct IntoIter32<T>{ .. };
```
with
```rust
type IntoIter0<T> = IntoIter<T, const 0>;
..
type IntoIter32<T> = IntoIter<T, const 32>;
```
once const generics land?

Then the eventual type would be [the nice one](https://github.com/rust-lang/rfcs/pull/2185#issuecomment-338375424), and the residual cruft is just a bunch of deprecated type aliases, which should be easy to maintain and ignore in the docs.

bluss:
We can do one thing before we even start the implementation: Work on deprecating & future warning the existing `array_value.into_iter()` calls.

ishitatsuyuki:
I experimented with an implementation of deprecation warning, but it seems that I can't avoid one particular false positive: https://play.rust-lang.org/?gist=8b69b43bc947bf112c0f20d0e58b3add&version=stable

Basically, the differences between the below cannot be detected without changing the signature:

```
a.into_iter();
(&a).into_iter();
```

sfackler:
We discussed this at the @rust-lang/libs meeting today, and feel like this won't really be actionable until integer generics are implemented.

@rfcbot fcp postpone

bluss:
@sfackler What do you think about starting to warn / soft deprecate `[1, 2, 3].into_iter()` in favour of `[1, 2, 3].iter()` ?

ishitatsuyuki:
I have mentioned that if we don't stabilize the iterator types, we can implement it with macro generated iterators and partially stabilize the into_iter() function only.

sfackler:
@bluss 

Not sure - we haven't really done anything like that before. @aturon thoughts?

@ishitatsuyuki 

I am personally not a fan of the current "up to size 32" approach, and don't really want to expand that further.

ishitatsuyuki:
@sfackler - what we only can do currently is that "up to size 32" approach; it should not be a blocker to improve array usability.

I originally thought that the main concern is backwards compatibility, which needs a lint to be developed. As I emphasized above, const generics is not a blocker for this, just a wishlist item. As const generics is an approved RFC, we can expect the stabilization at the same time.

aturon:
@bluss I'd be in favor.

Centril:
Closing since FCP with the proposal to postpone is now complete.

PR #2189, opened 2017-10-25
Add a RFC for TryClone trait

achanda:
[Rendered](https://github.com/achanda/rfcs/blob/307e2c78d94d78bbe8d0b1ab5c46fefab56345c7/text/0000-try-clone-trait.md)


text/0000-try-clone-trait.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Alternatives
+The status quo, keep everything as-is.
+
+# Unresolved questions
+Do we need this functionality? It is not clear under what conditions can cloning a type
+fail. In case we do not need this, implementing the RFC will produce extra maintenance
+overhead.
+
+[1] https://github.com/search?l=Rust&q=TryClone&type=Code&utf8=%E2%9C%93
```
Centril: Nitpick: Prefer inline links instead of footnotes.

text/0000-try-clone-trait.md
```diff
@@ -0,0 +1,47 @@
...
+of those crates.
+
+# Detailed design
+This trait should be added in `std` and should be called `TryClone`. It should look similar
+to `TryFrom`
+
+```
+pub trait TryClone<T> {
+    type Error,
+    fn try_clone() -> Result<Self, Self::Error>
```
Centril: Should:
```rust
fn try_clone_from(&mut self, source: &Self) -> Result<(), Self::Error> { ... }
```
be added to mirror `Clone`?

text/0000-try-clone-trait.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Summary
+A number of types in the standard library has a `try_clone` method that returns a 
+`Result<T>`. The semantics for conversion is similar in all these cases, thus it
+will be useful to add a new `TryClone` trait and have the types implement that trait.
+
+# Motivation
+A number of crates have defined `TryClone` as a trait and implemented it on standard
+types [1]. Including this in the standard library will reduce some clutter in the code
+of those crates.
```
sfackler: What code? Can you give an example of real-world code that would be improved here?

ExpHP:
To me, it does *not* look like `try_clone` is the standard library's name for a clone function that can fail.  I mean, the `try_clone` methods in the standard library are an odd bunch; notice that **all of their descriptions** are of the form with _Creates a new independently owned handle to the underlying \__, despite the general sounding name.

If anything, their naming says more to be about the *ancestry* of these types than anything else.

> A number of crates have defined `TryClone` as a trait and implemented it on standard types. https://github.com/search?l=Rust&q=TryClone&type=Code&utf8=%E2%9C%93

What I see in that link are:

* A trait that [entirely ignores](https://github.com/lrs-lang/lib/blob/42b958a32161ec418bceffd340259b14274a3294/src/base/conv/clone.rs#L25-L41) the standard library's `try_clone` functions.
* Three cases ([1](https://github.com/bfil/exar-db/blob/5a801fbeb177b6c2a5c34be63d9e667a7a72a453/exar-net/src/stream.rs#L64-L71) [2](https://github.com/jmpspace/websocket-protocol/blob/master/src/protocol.rs#L42-L47) [3](https://github.com/Centril/tor_control/blob/8a3c3f894a89571b9c641c990f3067f8d2c06f2c/src/lib.rs#L434-L436)) of what appear to me to be premature abstractions.  They each wrap precisely one of the standard library's `try_clone` methods, and it's clear that none of them were ever intended to support any of the other types.  AFAICT all of these traits merely exist to allow the structs in these crates to wrap types written by wrappers around the standard library type in client code (who can implement the trait for compatibility).

Evrey:
> This trait should be added in `std` [...]

You meant `core::clone` there. I'm not that excited for a second `HashMap`, i.e. an item that *could* be in `core` *(or rather `alloc` in this case)* but isn't.

> It is not clear under what conditions can cloning a type fail.

- [On integer overflow](https://doc.rust-lang.org/src/alloc/arc.rs.html#671-718)
- [On more integer overflow](https://doc.rust-lang.org/src/alloc/rc.rs.html#806-826)
- [On yet even more integer overflow](https://doc.rust-lang.org/src/alloc/rc.rs.html#1230-1247)
- [On conflicting borrows](https://doc.rust-lang.org/src/core/cell.rs.html#842-847)
- [On members failing to clone](https://doc.rust-lang.org/src/std/thread/mod.rs.html#913)
- On any other kind of invariant-breaking edge case, where keeping shared data like reference counts or dirty flags is a thing.

Note that those are also candidates for an `impl TryClone` and that there are probably many more, especially when [fallible allocations](https://github.com/rust-lang/rfcs/pull/2116) are a thing.

All in all, I do like the proposal. I probably won't need it any time soon, but I like having it around, just in case.

One more thing, though:

Besides quite some typos, you forgot about `Clone::clone_from`, so you might want to expand `TryClone` like this:

```rust
pub trait TryClone: Sized {
    type Error;

    fn try_clone(&self) -> Result<Self, Self::Error>;

    #[inline]
    fn try_clone_from(&mut self, src: &Self) -> Result<(), Self::Error> {
        *self = src.try_clone()?
    }
}
```

text/0000-try-clone-trait.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Detailed design
+This trait should be added in `std` and should be called `TryClone`. It should look similar
+to `TryFrom`
+
+```
```
bjorn3:
````
```rust
````

For syntax highlighting

Comment (bjorn3) [reply text/0000-try-clone-trait.md]: Oops, someone has already requestes this.

**Comment by achanda** [comment]
Thanks, everyone. Pushed a new version with all comments addresses.

File: text/0000-try-clone-trait.md [hunk 5]
```diff
@@ -0,0 +1,72 @@
...
+
+# Motivation
+Failing to clone an object is fairly common, for example, the system runs out of resources to represent the underlying type.
+Implementations have worked around this in two ways. Some types in the standard library has a `try_clone` method that returns a 
+`Result<T>`. A number of [crates](https://github.com/search?l=Rust&q=TryClone&type=Code&utf8=%E2%9C%93) have defined `TryClone` as a trait and implemented it on standard
+types.
+
+One example is [Clone](https://doc.rust-lang.org/src/alloc/arc.rs.html#710) for `std::alloc::arc`. This is better represented
+by returning a `Result<Arc>`. Doing that will also make the unsafe block unnecessary. Also, this will enable
+the caller to handle these errors idiomatically instead of aborting the process.
```
sfackler: This error is only encountered if your process leaks 2 billion copies of the Arc, after which you can never clone it again. How is that situation supposed to be "idiomatically handled"?

Evrey: Talking about 2 billion copies... An `Arc` takes 4 bytes on 32-bit architectures plus the data behind that 4 bytes pointer. If, in theory, you could use every single byte of your program to allocate a copy of an `Arc`, then you would have a maximum reference count of `0x40000000_u32`, which is `< u32::max_value()`. And now reduce that number further, because you also have to fit the actual data, your whole program, and the OS kernel into your 4GiB address space.

As a result, it is absolutely impossible to overflow the reference counters of `Arc` and `Rc`, ever, on any CPU architecture. So, in a way, you could optimise `alloc` instead by not doing overflow checks.

sfackler:
```rust
let foo = Arc::new(0);
loop {
    mem::forget(foo.clone());
}
```

Evrey: Now, that is just evil. ;)

pornel:
I see it as useful for wrapping C libraries. Most have functions for copying/duplicating their objects, but they're allowed to fail (`NULL` is always an option), so a non-panicking `Clone` sounds like a good fit for these.

Centril:
Do we need a `try_` version for [`std::borrow::ToOwned`](https://doc.rust-lang.org/nightly/std/borrow/trait.ToOwned.html)?

burdges:
Just fyi, `Cow::into_owned` needs to take `self` by value, so if you push this enough you'd need to return the `Cow` along with your error, which breaks `?`.
```
TryCow::try_into_owned(self) -> Result<<B as TryToOwned>::Owned,(Self,<B as TryToOwned>::Error)>
```

dtolnay:
One important thing was not clear to me from the RFC and the discussion so far -- the text seems to frame TryClone as an improvement over ad hoc `try_clone()` methods on individual types, but why is this an improvement? What are some concrete uses cases that this enables that are inconvenient or not possible today?

PR #2278, opened 2018-01-04
Fix small typo in template.

nelsonjchen:


Centril:
Thanks! Seems like a straightforward fix =)

PR #2286, opened 2018-01-10
Add note to RFC 505 that it has been amended by RFC 1574

lukaramu:
RFC 1574 amended RFC 505. This change adds a note to the beginning of RFC 505, noting the amendment and linking to RFC 1574 and the combined version of the conventions contained in Appendix A of RFC 1574.


Centril:
Thanks, I'll take a look at this one in a bit =)

PR #2342, opened 2018-02-20
Allow `if` and `match` in constants

oli-obk:
[Rendered](https://github.com/oli-obk/rfcs/blob/const-control-flow/text/0000-const-control-flow.md)

### DIFF
--- a/text/0000-const-control-flow.md
+++ b/text/0000-const-control-flow.md
@@ -1,0 +1,136 @@
+- Feature Name: const-control-flow
+- Start Date: 2018-01-11
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Enable `if` and `match` during const evaluation and make them evaluate lazily.
+In short, this will allow `if x < y { y - x } else { x - y }` even though the
+else branch would emit an overflow error for unsigned types if `x < y`.
+
+# Motivation
+[motivation]: #motivation
+
+Conditions in constants are important for making functions like `NonZero::new`
+const fn and interpreting assertions.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+If you write
+
+```rust
+let x: u32 = ...;
+let y: u32 = ...;
+let a = x - y;
+let b = y - x;
+if x > y {
+    // do something with a
+} else {
+    // do something with b
+}
+```
+
+The program will always panic (except if both `x` and `y` are `0`) because
+either `x - y` will overflow or `y - x` will. To resolve this one must move the
+`let a` and `let b` into the `if` and `else` branch respectively.
+
+```rust
+let x: u32 = ...;
+let y: u32 = ...;
+if x > y {
+    let a = x - y;
+    // do something with a
+} else {
+    let b = y - x;
+    // do something with b
+}
+```
+
+When constants are involved, new issues arise:
+
+```rust
+const X: u32 = ...;
+const Y: u32 = ...;
+const FOO: SomeType = if X > Y {
+    const A: u32 = X - Y;
+    ...
+} else {
+    const B: u32 = Y - X;
+    ...
+};
+```
+
+`A` and `B` are evaluated before `FOO`, since constants are by definition
+constant, so their order of evaluation should not matter. This assumption breaks
+in the presence of errors, because errors are side effects, and thus not pure.
+
+To resolve this issue, one needs to eliminate the intermediate constants and
+directly evaluate `X - Y` and `Y - X`.
+
+```rust
+const X: u32 = ...;
+const Y: u32 = ...;
+const FOO: SomeType = if X > Y {
+    let a = X - Y;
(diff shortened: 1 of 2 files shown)


Centril:
The semantics of `match` in constants is not described at all, so it feels a bit thin atm. Could you add some details on this in the RFC?

oli-obk:
@Centril I don't see `match` and `if` as separate commands semantically. The following can be converted

```rust
match enum_value {
    Variant0Pattern => {},
    Variant1Pattern => {},
    _ => {
        // something else
    },
}
```

to some very unsafe code:

```rust
if std::intrinsics::discriminant(enum_value) == 0 {
     let Pattern: EnumVariant0 = transmute(enum_value);
} else if std::intrinsics::discriminant(enum_value) == 1 {
     let Pattern: EnumVariant1 = transmute(enum_value);
} else {
     // something else
}
```

Which is almost how MIR handles matches.

Centril:
@oli-obk Still... I feel lang RFC texts should be as unambiguous as possible... This makes it easier to review changes proposed and define the spec / reference of the language =)

Havvy:
Part of the issue with this is that `if` and recursion allows for looping in a very unidiomatic way, so I've been told that we've been waiting until we have a good const story for looping. Could be solved naively by disallowing recursion by making it a compiler error when it happens?

oli-obk:
> Could be solved naively by disallowing recursion by making it a compiler error when it happens?

The allowed recursion depth isn't very high. Also `const fn` is unstable. You won't be able to do many interesting loops this way.

nox:
> @Centril I don't see match and if as separate commands semantically. The following can be converted

Then I would describe the semantics of `match` instead of `if`, and explain that `if` is just sugar for boolean matches. It seems weird to me to do the opposite.

petrochenkov:
@nox @oli-obk 
>and since if is simply match on a boolean, ...

Fun fact: there are actually differences in `if cond { ... } else { ... }` and `match cond { true => ..., false => ... }` with regards to how long `cond` lives! (At least without NLL.)
I tried to desugar all `if`s into `match`es as part of https://github.com/rust-lang/rfcs/pull/2260#issuecomment-367158854 and found some pretty real breakage from borrow checker during `rustc` bootstrap.

withoutboats:
@rfcbot fcp merge

eddyb:
@petrochenkov `match { let cond = cond; cond } { ... }` might work - or we could even make "terminating" scopes explicit in HIR instead of leaving them implicit.

gnzlbg:
This is required to implement `const fn` constructors for portable packed SIMD vector types in `stdsimd`, and is tangetially related to fixing https://github.com/rust-lang/rust/issues/48745 .

rkruppe:
@gnzlbg Can you elaborate how? I don't see why those functions (both for portable vector types and arch-specific ones) would have to branch or match, and indeed I can't see anything other than simple data movement when I look at the current implementations.

gnzlbg:
@rkruppe wrong link, I've edited the post, but this explains it in a nutshell:

Boolean vector constructors' take `bool`s as arguments but use internally a signed integer type to store their values. The conversion is:

```rust
#[inline]
/* const */ fn bool_to_internal(x: bool) -> $elem_ty {  // $elem_ty is, e.g., i32
    if x  {
        !(0 as $elem_ty)
    } else {
        0 as $elem_ty
    }
}
```

The `new` and `splat` constructors need to perform these conversions: 

```rust
#[inline]
pub /*const*/ fn new($($elem_name: bool),*) -> Self {
    $id($(Self::bool_to_internal($elem_name)),*)
}
```

and currently cannot be `const fn`. The constructors of the non-boolean SIMD vector types are, however, `const fn`.

rkruppe:
Ah, so this is *only* about the boolean consts? And probably only the platform specific types?

In any case note that this RFC just gives you lazy evaluation of the branches. In simple cases like selecting between `0` and `!0`, you can already use [arrays and indexing](https://play.rust-lang.org/?gist=c0e40e50c9baa165c7b1b11c8f00233a&version=nightly) today.

gnzlbg:
> Ah, so this is only about the boolean consts? And probably only the platform specific types?

This is about the portable boolean vector types.

> In simple cases like selecting between 0 and !0, you can already use arrays and indexing today.

Oh, I did not know about that! Thanks! That should work for `stdsimd` :)

gnzlbg:
@rkruppe so I can't give this a try here because rustc does not allow using bools as simd vector types yet :/

eddyb:
> In simple cases like selecting between `0` and `!0`, you can already use arrays and indexing today.

Why not just cast the boolean to an integer (giving you either `0` or `1`) and then negate the integer (giving you either `0` or `-1` aka `!0`)?

This is usually seen in branchless selects, where `c ? a : b` becomes `(-c & a) | (~-c & b)`.

Centril:
**Huzzah!** The RFC has been **accepted**.

Tracking issue: https://github.com/rust-lang/rust/issues/49146

PR #2344, opened 2018-02-22
Allow `loop` in constant evaluation

oli-obk:
[Rendered](https://github.com/oli-obk/rfcs/blob/const_looping/text/0000-const-looping.md)

### DIFF
--- a/text/0000-const-looping.md
+++ b/text/0000-const-looping.md
@@ -82,11 +82,15 @@ indefinitely, we count the number of terminators processed and whenever we reach
 a fixed limit, we report a lint mentioning that we cannot guarantee that the
 evaluation will terminate and reset the counter to zero. This lint should recur
 in a non-annoying amount of time (e.g. at least 30 seconds between occurrences).
+This means that there's an internal deterministic counter (for the terminators) and
+a timestamp of the last (if any) loop warning emission. Both the counter needs to reach
+its limit and 30 seconds have to have passed since the last warning emission in order
+for a new warning to be emitted.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-* Infinite loops will hang the compiler if the lint is not denied
+* Infinite loops will cause the compiler to never finish if the lint is not denied
 
 # Rationale and alternatives
 [alternatives]: #alternatives
--- a/text/0000-const-looping.md
+++ b/text/0000-const-looping.md
@@ -9,16 +9,18 @@
 Allow the use of `loop`, `while` and `while let` during constant evaluation.
 `for` loops are technically allowed, too, but can't be used in practice because
 each iteration calls `iterator.next()`, which is not a `const fn` and thus can't
-be called within constants.
+be called within constants. Future RFCs (like
+https://github.com/rust-lang/rfcs/pull/2237) might lift that restriction.
 
 # Motivation
 [motivation]: #motivation
 
-Any iteration is expressible as a recursion. Since we already allow recursion
+Any iteration is expressible with recursion. Since we already allow recursion
 via const fn and termination of said recursion via `if` or `match`, all code
-enabled by const recursion is already legal now. Writing loops with recursion is
-very tedious and can quickly become unreadable, while regular loops are much
-more natural in Rust.
+enabled by const recursion is already legal now. Some algorithms are better
+expressed as imperative loops and a lot of Rust code uses loops instead of
+recursion. Allowing loops in constants will allow more functions to become const
+fn without requiring any changes.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
@@ -73,8 +75,8 @@ const fn fib(n: u128) -> u128 {
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-A loop in MIR is a cyclic graph of BasicBlocks. Evaluating such a loop is no
-different from evaluating a linear sequence of BasicBlocks, except that
+A loop in MIR is a cyclic graph of `BasicBlock`s. Evaluating such a loop is no
+different from evaluating a linear sequence of `BasicBlock`s, except that
 termination is not guaranteed. To ensure that the compiler never hangs
 indefinitely, we count the number of terminators processed and once we reach a
 fixed limit, we report an error mentioning that we aborted constant evaluation,
--- a/text/0000-const-looping.md
+++ b/text/0000-const-looping.md
@@ -78,18 +78,15 @@ const fn fib(n: u128) -> u128 {
 A loop in MIR is a cyclic graph of `BasicBlock`s. Evaluating such a loop is no
 different from evaluating a linear sequence of `BasicBlock`s, except that
 termination is not guaranteed. To ensure that the compiler never hangs
-indefinitely, we count the number of terminators processed and once we reach a
-fixed limit, we report an error mentioning that we aborted constant evaluation,
-because we could not guarantee that it'll terminate.
+indefinitely, we count the number of terminators processed and whenever we reach
+a fixed limit, we report a lint mentioning that we cannot guarantee that the
+evaluation will terminate and reset the counter to zero. This lint should recur
+in a non-annoying amount of time (e.g. at least 30 seconds between occurrences).
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-* Loops are not guaranteed to terminate
-    * We catch this already by having a maximum number of basic blocks that we
-      can evaluate.
-* A guaranteed to terminate, non looping constant might trigger the limit, if it
-  has too much code.
+* Infinite loops will hang the compiler if the lint is not denied
 
 # Rationale and alternatives
 [alternatives]: #alternatives
@@ -98,3 +95,9 @@ because we could not guarantee that it'll terminate.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
+
+* Should we add a true recursion check that hashes the interpreter state and
+  detects if it has reached the same state again?
+    * This will slow down const evaluation enormously and for complex iterations
+      is essentially useless because it'll take forever (e.g. counting from 0 to
+      `u64::max_value()`)
--- a/text/0000-const-looping.md
+++ b/text/0000-const-looping.md
@@ -1,0 +1,98 @@
+- Feature Name: const_looping
+- Start Date: 2018-02-18
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow the use of `loop`, `while` and `while let` during constant evaluation.
+`for` loops are technically allowed, too, but can't be used in practice because
+each iteration calls `iterator.next()`, which is not a `const fn` and thus can't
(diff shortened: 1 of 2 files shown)


text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+- Start Date: 2018-02-18
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow the use of `loop`, `while` and `while let` during constant evaluation.
+`for` loops are technically allowed, too, but can't be used in practice because
+each iteration calls `iterator.next()`, which is not a `const fn` and thus can't
```
Centril: Linking #2237 which (or a modified version of which) will fix that and let you use `iter.next()` in `const fn`.

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+be called within constants.
+
+# Motivation
+[motivation]: #motivation
+
+Any iteration is expressible as a recursion. Since we already allow recursion
```
Centril: Small possible improvement: s/a recursion/with recursion.

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+via const fn and termination of said recursion via `if` or `match`, all code
+enabled by const recursion is already legal now. Writing loops with recursion is
```
Centril: Does this hold given recursion depth limits?

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+very tedious and can quickly become unreadable, while regular loops are much
+more natural in Rust.
```
Centril: This feels like a somewhat **bold** assertion that seems to suggest that functional languages are less readable, while I'd argue that the opposite is the case, recursion is often more readable than loops - and iteration combinators using `map`, `filter`, `fold` even more so (tho combinators have nothing to do with recursion).

I agree that regular loops are more used in Rust, but I believe this has more to do with the lack of guaranteed TCO (Tail Call Optimization, https://en.wikipedia.org/wiki/Tail_call, https://stackoverflow.com/questions/310974/what-is-tail-call-optimization) which RFC #1888 attempts to address.

I don't think any of these assertions are necessary to motivate loops in `const fn` - feature parity alone and the feeling that the language is uniform when you switch from `fn -> const fn` should be enough.

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+If you previously had to write functional code inside constants, you can now
+change it to imperative code. For example if you wrote a fibonacci like
```
Centril: Who doesn't love a fibonacci example =P This section feels very nicely written!
Stylistically I'ld do: `s/like/like:` (and before the other sections as well..)

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+        i += 1;
+    }
+    b
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A loop in MIR is a cyclic graph of BasicBlocks. Evaluating such a loop is no
```
Centril: s/BasicBlocks/`BasicBlocks`

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+different from evaluating a linear sequence of BasicBlocks, except that
+termination is not guaranteed. To ensure that the compiler never hangs
+indefinitely, we count the number of terminators processed and once we reach a
+fixed limit, we report an error mentioning that we aborted constant evaluation,
```
Centril: Perhaps a word about if and how you can control that fixed limit? (I assume the `#[recursion_limit = "<limit>"]` attribute?)

text/0000-const-looping.md
```diff
@@ -0,0 +1,98 @@
...
+because we could not guarantee that it'll terminate.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Loops are not guaranteed to terminate
```
Centril: Nicely written.

In the **far future** it might be interesting to have a `total` effect in the language with a totality checker that guarantees termination.

text/0000-const-looping.md
oli-obk: yea, if the loop takes too many iterations you'll hit the limit, but that's not much of an issue imo, since you can always work around that with unrolling or similar tricks.

text/0000-const-looping.md
oli-obk: there's no such attribute

Centril: Not this? https://github.com/rust-lang/rust/blob/master/src/librustc/middle/recursion_limit.rs

oli-obk: That's a different recursion. Recursing in const eval does not actually cause recursion in the compiler, just more elements in the virtual stack `Vec`.

text/0000-const-looping.md
Centril: Ah, thanks for explaining =)

Still... might it be good to allow users to raise the limit from the default so that it is not so fixed and so that you can fix cases wherein you know that termination is guaranteed but rustc complains nonetheless?

text/0000-const-looping.md
joshtriplett: Do we have a mechanism to adjust the amount of compile-time evaluation allowed, similar to the macro limits?

oli-obk: Nope, but it's trivial to add (a single line next to the recursion limit attr code)

scottmcm:
Just counting all terminators worries me slightly that a tweak to MIR optimizations or HIR->MIR lowering will result in code that happened to be right on the edge breaking in an update.

Is there some tweak to the rule that's more resilient to otherwise-invisible changes, and where it's easier to understand by looking at your code how many of whatever it's counting you're taking?  Like maybe only counting backedges taken.  (Though even that changes for loop unrolling, I guess.)

Come to think of it, I wonder if the recursion limit means that the MIR inliner should add a new marker statement to the MIR so that heuristic changes don't change how many times a function recurses...

oli-obk:
Well.... the current limit is riddiculously high. 1 million terminators take a while to evaluate.

We could also hash the entire state at every terminator and detect true loops this way. We could even only do that once we reach the limit so we don't slow down the regular evaluations.

oli-obk:
We could emit a warning for long evaluations so ppl know rustc isn't hanging, just busy

burdges:
Also test builds could report how much time was spent in rustc itself vs const evaluation, so slow crates get blamed quickly.

Centril:
> Well.... the current limit is riddiculously high. 1 million terminators take a while to evaluate.

Do you have some guess as to how much *a while* might be? Just so we're on the same page =)

oli-obk:
While that depends on the code you're executing, experience with when the limit was on statements was several minutes

nagisa:
We should *not* have any error on constant evaluation by default. Not time and not terminator count based. Instead, there should exist a lint of-sorts that would warn every so often that such-and-such constant item is still being evaluated, and that’s it. People who want to make sure stuff doesn’t get out of hand can then #[deny] it and people who know they gonna do heavy compile-time computations can #[allow] it so the compiler doesn’t nag at them too much.

So basically what [this](https://github.com/rust-lang/rfcs/pull/2344#issuecomment-368227609) says.

joshtriplett:
@nagisa I'd be concerned about people running automated builds and finding that they take forever, in that case. CI, for instance.

nagisa:
@joshtriplett Sure, hence `#[deny]`. Of course we still would need to provide a way to configure actual limit. Most importantly, with this approach breaking code which uses `#[deny(long_const_eval)]` would be about as breaking as breaking code which specifies `#[deny(warnings)]` (we do not consider it to be a breaking change).

clarcharr:
Considering how Rust's type system is turing complete, infinite compile loops are already possible. This just makes them a bit easier.

oli-obk:
To the best of my knowledge you can't do infinite loops in the type system without hitting the recursion limit. True recursions would trigger an error of the form "needed X to compute X"

oli-obk:
They are different, because now there can be repetitive recursions. I think going with the warning and no limit is the way to go. I'll change the RFC accordingly

oli-obk:
Long running const eval doesn't cause errors anymore: https://github.com/rust-lang/rust/issues/49980

text/0000-const-looping.md
```diff
@@ -0,0 +1,103 @@
...
+termination is not guaranteed. To ensure that the compiler never hangs
+indefinitely, we count the number of terminators processed and whenever we reach
+a fixed limit, we report a lint mentioning that we cannot guarantee that the
+evaluation will terminate and reset the counter to zero. This lint should recur
+in a non-annoying amount of time (e.g. at least 30 seconds between occurrences).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Infinite loops will hang the compiler if the lint is not denied
```
pnkfelix: Nit: Even just a `#[warn]` on the lint will cause warning output to occur (and repeat ad nauseaum) in response to an infinite loop, right? Usually when I hear "hang" I think of "no output visible"... so I would personally have written "if the lint is `#[allow]`'ed" here. But that might just be my own personal interpretation.

text/0000-const-looping.md
```diff
@@ -0,0 +1,103 @@
...
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A loop in MIR is a cyclic graph of `BasicBlock`s. Evaluating such a loop is no
+different from evaluating a linear sequence of `BasicBlock`s, except that
+termination is not guaranteed. To ensure that the compiler never hangs
+indefinitely, we count the number of terminators processed and whenever we reach
+a fixed limit, we report a lint mentioning that we cannot guarantee that the
+evaluation will terminate and reset the counter to zero. This lint should recur
+in a non-annoying amount of time (e.g. at least 30 seconds between occurrences).
```
pnkfelix: So just to be clear: the const-eval state will include both a terminator-counter and a timestamp; When the terminator-counter both 1. exceeds the (here unspecified) fixed limit, and 2. the current time delta from the timestamp is non-annoying, then we report the lint, reset the terminator-counter *and* timestamp, and finally resume const-eval execution?

(I'm just checking that there's not a typo here and that we are planning on using both pieces of state for the filter here, something of which I am in favor.)

pnkfelix:
@rfcbot fcp merge

(I usually construct a comment thread summary before issuing the rfcbot command, but in this case I *think* all of the relevant comments were addressed, usually via incorporation into the RFC text, so I'm able to just recommend that people read the RFC.)

text/0000-const-looping.md
oli-obk: The current implementation just has a MIR terminator counter for the warning. There's no host time check. Shouldn't be too hard to add.

Centril:
**Huzzah!** This RFC is now **merged!**

Tracking issue: https://github.com/rust-lang/rust/issues/52000

PR #2352, opened 2018-02-26
typo in rfc1216

dmitryvm:
Was searching some info about exclamation mark that I noticed first time recently. Found lots of answers to my questions in the RFC. Thanks for the great work!

### DIFF
--- a/text/1216-bang-type.md
+++ b/text/1216-bang-type.md
@@ -191,7 +191,7 @@ So why do this? AFAICS there are 3 main reasons
     `Result<T, E>` to satisfy a trait but we know that the method will always
     succeed.
 
-    For example, here's a saner implementation of `FromStr` for `String` that
+    For example, here's a saner implementation of `FromStr` for `String` than
     currently exists in `libstd`.
 
     ```rust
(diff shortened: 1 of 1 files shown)


text/1216-bang-type.md
```diff
@@ -191,7 +191,7 @@ So why do this? AFAICS there are 3 main reasons
     `Result<T, E>` to satisfy a trait but we know that the method will always
     succeed.
 
-    For example, here's a saner implementation of `FromStr` for `String` than
+    For example, here's a saner implementation of `FromStr` for `String` that
     currently exists in `libstd`.
```
scottmcm: I think this one is correct as-is, as the RFC is written from the perspective of the feature not existing.  (And the current version still uses `ParseError` because back-compat.)

Centril: Yes, I agree. The proposed change here is semantic and not just a typo fix.

dmitryvm: Oops, indeed. I'm sorry, I should have read that more attentively, before considering it as a mistake. I changed it back to the original.

Centril:
Looks all good now, thanks!

PR #2365, opened 2018-03-22
Fix pseudocode typo in RFC 1951

Phlosioneer:
I believe this `Whatever` was accidentally left out.


Centril:
Yep, looks like it. Thanks!

PR #2379, opened 2018-04-01
Shorten the `cargo` command

mark-i-m:
[Rendered](https://github.com/mark-i-m/rfcs/blob/46a74a9a204ac3879accecdb702cd27266d53a2b/text/0000-cargogo.md)

### DIFF
--- a/text/0000-cargogo.md
+++ b/text/0000-cargogo.md
@@ -1,0 +1,65 @@
+- Feature Name: cargogo
+- Start Date: 2108-04-01
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes shortening the name of the `cargo` command for ergonomic benefit from 5 letters down to 2 letters long.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, the `cargo` command is very long to type: 5 whole letter! Compare this to other more ergonomic commands like `ls` or `rg`, and you can see that there are clearly gains to be made. For this reason, this RFC proposes shortening the command down to 2 letters from 5 letters.
+
+Careful consideration is given to avoid breaking scripts and other tooling.
+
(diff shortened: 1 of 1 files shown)


Centril:
Clearly the new command should be named `cg`!

crlf0710:
Can `cargo` work independently of toolchain? I mean, can we just 1. rename `cargo` to something else. 2. rename `rustup` to `cargo` ?

warlord500:
I am definitely against this! while cargo is long for its high usage but I would like to remind ever one of common command used just like cargo, the great make command! in most modern operating systems, you have support for either alias and/or symbolic links and/or hard links and/or scripts making this change near useless.

for the hypothetical case that you are not  running on system with that support, you can likely cross compile to that machine using one of the above methods.

in addition, this change could lead to large amounts of confusion. people have to figure out if some one is referring to go(the language), (the compiler), (rust build tool for package management), and possibly more uses of "go". 

any shorter word will likely have the same problem as "go".

est31:
I think we should call it ar instead. Go is already taken.

gerryxiao:
Should be 'car', more pleasure to use really.

nox:
Cars pollute and make people lazy, I think it should be named bike, which is still a 20% size improvement of the name. Then we can rename rustup to shed, which is a 33% size improvement too.

sgrif:
I think we can all agree that there's no possible downside to using `go` for this command. Clearly this needs to be nominated for the next release

Centril:
@nox In honor of Haskell Curry, I propose we rename `cargo` to `B` and `rustup` to `S`; And so then we have b & s.

bluetech:
This is missing the [Prior Art](https://github.com/rust-lang/rfcs/blob/master/text/2333-prior-art.md) section. Of course, there isn't any.

sgrif:
April 1 was a fun day. :smile:

PR #2380, opened 2018-04-01
Deprecate unsafe

xfix:
[Rendered](https://github.com/xfix/rfcs/blob/deprecate-unsafe/text/0000-deprecate-unsafe.md)

### DIFF
--- a/text/0000-deprecate-unsafe.md
+++ b/text/0000-deprecate-unsafe.md
@@ -1,0 +1,123 @@
+- Feature Name: deprecate_unsafe
+- Start Date: 2018-04-01
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Currently Rust allows using `unsafe`. This RFC proposes to deprecate
+`unsafe` keyword.
+
+# Motivation
+[motivation]: #motivation
+
+`unsafe` keyword is the first cause of segfaults in Rust, compiler bugs
+being second. Removing this feature would make Rust significantly safer.
+Additionally, writing `unsafe` code is tricky, and there are plenty of
+[traps to aware of](https://doc.rust-lang.org/nomicon/).
+
+This will make Rust much easier to learn, as it will remove a big part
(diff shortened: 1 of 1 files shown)


mark-i-m:
> unsafe keyword is the first cause of segfaults in Rust, compiler bugs being second

I think we should try to minimize churn over the epoch. To this end, I think a more conservative alternative would be to remove the bugs from the compiler.

spacekookie:
`rustfix` could be used to turn current Rust code into Java code

le-jzr:
> unsafe keyword is the first cause of segfaults in Rust

The obvious solution is to use `rustfix` to convert all unsafe code to C.

Centril:
Why translate to Java when you can translate to [Iota](https://en.wikipedia.org/wiki/Iota_and_Jot#Iota)?

varkor:
It'd surely be simpler just to make `rustfix` replace all `unsafe` blocks with `mem::uninitialized` and make `mem::uninitialized` safe (I think it must be `unsafe` for legacy reasons — I can't see any way anything could go wrong when using it).

mark-i-m:
Unfortunately I don't think you can do that. Deciding if two program are equivalent is undecidable...

warlord500:
@xfix, before making this RFC, did you ever ask yourself why unsafe exists in rust?
I am sure this is a joke, but for anybody who does take this seriously!

there are plenty of things that unsafe rust makes possible that if implemented correctly are perfectly safe
but can not be implemented without unsafe keyword or would be horribly slow/bloated in safe code! 

discouraging unsafe is perfectly reasonable position but outright removal would be disastrous! 
there are all ready tools that exist without changing the compiler or cargo to help eliminate unnecessary
unsafe code. 

put ``rust #[deny(unsafe_block)] `` at the top of crate or module and you have guaranteed that the module doesn't have any unsafe code. If we are to add that to every module automatically we are back where we started. ``rust #[allow(unsafe_block] `` become effectively the equivalent to the unsafe keyword.

xfix:
Closing, because this would prevent using `unsafe` in strings which seems like a bad idea.

PR #2382, opened 2018-04-02
Fix typo in 2349-pin.md

sinkuu:
This double-negation ("not implement `!Unpin`") looks like a typo. `PinBox` is useful for immovable types (which do not implement `Unpin`), right?

### DIFF
--- a/text/2349-pin.md
+++ b/text/2349-pin.md
@@ -231,7 +231,7 @@ impl<T: ?Sized> PinBox<T> {
 ```
 
 These APIs make `PinBox` a reasonable way of handling data which does not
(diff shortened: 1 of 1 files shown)


kennytm:
Technically "does not implement `!Unpin`" is a triple-negation 😛

Nemo157:
> Technically "does not implement `!Unpin`" is a triple-negation :stuck_out_tongue:

It may seem that way at first glance, but `Unpin` is not actually a negative, rather it refers to the fact a type is allowed to be "unpinned" even after it has been pinned:

> to unfasten or loosen by or as if by removing a pin; detach

PR #2387, opened 2018-04-03
Unapprove placement RFCs

aidanhs:
r? @nikomatsakis 

See https://github.com/rust-lang/rust/issues/27779#issuecomment-378416911 for reference.

### DIFF
--- a/text/0809-box-and-in-for-stdlib.md
+++ b/text/0809-box-and-in-for-stdlib.md
@@ -3,6 +3,12 @@
 - RFC PR: [rust-lang/rfcs#809](https://github.com/rust-lang/rfcs/pull/809)
 - Rust Issue: [rust-lang/rust#22181](https://github.com/rust-lang/rust/issues/22181)
 
+# This RFC has been **unapproved**
+
+For details see the [summary comment].
+
+[summary comment]: https://github.com/rust-lang/rust/issues/27779#issuecomment-378416911
+
 # Summary
 
   * Change placement-new syntax from: `box (<place-expr>) <expr>` instead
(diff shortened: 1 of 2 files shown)


JustAPerson:
At first I was confused to see this, but after reading the linked threads I'm glad that you and others remain motivated in seeking perfection in this area! Thank you! We must be competitive with C++ :)

aidanhs:
Is that common language for taking something accepted out of circulation? I tend to associate it with abandoning a proposal that hasn't been approved yet. I'd be happy using the language that another community/body/organisation users, but I couldn't think of any examples to draw from.

clarcharr:
Could go with "retroactively closed/postponed." I'd say that these have been retroactively postponed to indicate that while the RFCs were originally accepted, in hindsight it makes more sense to act as if they were postponed, because they need better design to satisfy the guarantees.

mark-i-m:
Repealed?

clarcharr:
Question: do we need an FCP to revert an FCP?

Ixrec:
Personally I think "retract" is the best term, though it's by a very small margin. Nitpicky arguments:

"Repeal" sounds too legalistic to me. It makes me imagine RFCs as proposals to modify the standardese text of the ISO Rust specification, such that repealing one in the future would imply making the opposite modification to the specification text. I think "Retract" manages to strike the ideal balance of sounding professionally formal without quite crossing into legalistically, only-comprehensible-for-wizards formal.

"Retroactively closed/postponed" sounds like the RFC should have been closed/postponed a long time ago, but for whatever reason it fell through the cracks and we simply forgot to do so. Since we accepted and implemented this RFC, then removed the implementation, that doesn't quite sound right to me.

"Unaccepted" would be my #2 choice. I like it slightly less than "retract" only because it sounds similar to "unacceptable", which is a somewhat judgmental word. But that's the only tiebreaker nitpick I can come up with.

nikomatsakis:
We discussed in the @rust-lang/lang meeting and decided to resist the alluring bikeshed and just merge as is. (Note that FCP on this decision has already occurred.)

PR #2389, opened 2018-04-04
Clarify we mean UpperCamelCase, not lowerCamelCase

shepmaster:
<s>Because it looks like a camel with a single hump in the middle: 🐪</s>

### DIFF
--- a/text/0430-finalizing-naming-conventions.md
+++ b/text/0430-finalizing-naming-conventions.md
@@ -20,7 +20,7 @@ official.
 
 ## General naming conventions
 
-In general, Rust tends to use `CamelCase` for "type-level" constructs
+In general, Rust tends to use `UpperCamelCase` for "type-level" constructs
 (types and traits) and `snake_case` for "value-level" constructs. More
 precisely, the proposed (and mostly followed) conventions are:
 
@@ -28,9 +28,9 @@ precisely, the proposed (and mostly followed) conventions are:
 | ---- | ---------- |
 | Crates | `snake_case` (but prefer single word) |
 | Modules | `snake_case` |
-| Types | `CamelCase` |
-| Traits | `CamelCase` |
-| Enum variants | `CamelCase` |
+| Types | `UpperCamelCase` |
+| Traits | `UpperCamelCase` |
+| Enum variants | `UpperCamelCase` |
 | Functions | `snake_case` |
 | Methods | `snake_case` |
 | General constructors | `new` or `with_more_details` |
@@ -38,12 +38,12 @@ precisely, the proposed (and mostly followed) conventions are:
 | Local variables | `snake_case` |
 | Static variables | `SCREAMING_SNAKE_CASE` |
 | Constant variables | `SCREAMING_SNAKE_CASE` |
-| Type parameters | concise `CamelCase`, usually single uppercase letter: `T` |
+| Type parameters | concise `UpperCamelCase`, usually single uppercase letter: `T` |
 | Lifetimes | short, lowercase: `'a` |
 
 ### Fine points
 
-In `CamelCase`, acronyms count as one word: use `Uuid` rather than
+In `UpperCamelCase`, acronyms count as one word: use `Uuid` rather than
 `UUID`.  In `snake_case`, acronyms are lower-cased: `is_xid_start`.
 
 In `snake_case` or `SCREAMING_SNAKE_CASE`, a "word" should never


shepmaster:
Corresponding PR has been made to the [api-guidelines repo](https://github.com/rust-lang-nursery/api-guidelines/pull/163)

petrochenkov:
Wait, what about this guy?
default

At least [wiki](https://en.wikipedia.org/wiki/Camel_case) says both `camelOne` and `CamelTwo` are camel case and we also have long-stable `non_camel_case_types` lint - something that we can't change.

shepmaster:
> Wait, what about this guy?

That would be equivalent to `reallyLongIdentifier`; two humps in the middle: 🐫. There are no emoji for higher-order-camels, so please don't ask about that case 😉 

> At least wiki says both `camelOne` and `CamelTwo` are camel case

Yes, many people misuse the terms, so this is an uphill battle on my part. However, I hope we'd agree that one of those is an idiomatic Rust type name and the other is not. We could choose to say "no, not *that* camel case, use the *other* camel case", or we could choose to use a distinct term that only means the one we want.

One interesting thing that I've been mulling about regards WASM. In the JS ecosystem, `camelCase` and `PascalCase` are both used (in different contexts). If more JS developers come to Rust and we tell them to use "camel case", they might be needlessly confused.

> we also have long-stable `non_camel_case_types` lint

A good point, but we've previously deprecated / renamed lints that had names we no longer felt were appropriate. I hadn't thought of this case before, but I believe it could fit into the same category. We certainly wouldn't ever get *rid* of support for the existing lint name.

Zarenor:
In terms of prior art, I think it's also worth mentioning that the [C# capitalization guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)  also make the distinction between PascalCase and camelCase. Oracle's [Java recommendations](http://www.oracle.com/technetwork/java/codeconventions-135099.html) dodge the issue by not naming the conventions.

I think it's thoroughly more clear to have more-distinct names, too. I realize that we're not recommending the camelCase option for any identifiers in Rust, but it's certainly out there in other languages. I think having distinct terms that are shared across ecosystems is a win.

Ixrec:
FWIW, I personally prefer to call these "lower camel case" and "upper camel case" since it's hard to imagine how anyone could manage to misinterpret those terms, and the term "PascalCase" is significantly less famous than camel case.

loganfsmyth:
I'll just throw in that when I came across `SomeWord` being described as camel case, I legitimately thought it was a typo/mistake. Explicitly calling out upper camel vs lower camel would certainly go a long way to helping that. That said, I was entirely unaware that anyone considered capitalized-first-word as a potential form of camel case. I don't know that I would have known the word `PascalCase` off the top of my head, but it's quite easy to search for, so it seems like the most practical description in my mind.

sgrif:
We should probably get @killercup's permission

mark-i-m:
> There are no emoji for higher-order-camels

My favorite out-of-context quote for the month.

eddyb:
> FWIW, I personally prefer to call these "lower camel case" and "upper camel case"

I've seen them called `lowerCamelCase` and `UpperCamelCase` for many years in both C++ and JS projects before *a single* `PascalCase` reference (which confused me at first). YMMV of course.

mark-i-m:
Personally, my vote is for :camel: case, which has been what I have always heard it called...

sgrif:
@alexcrichton @BurntSushi This is a pretty low effort RFC to nominate, either of you care to bring it up at the next libs team meeting?

joshtriplett:
I don't think we should call this "PascalCase"; informally, I'd say that I don't think that term has *nearly* the name recognition. And even "UpperCamelCase" doesn't seem clear except insofar as being self-documenting. We should just explain and/or define the terms we use in detail the first time we use them, which will make it clear even to people who haven't heard any of the terms before. We give examples in a table, but I don't think we have an explanation anywhere that actually says "capitalize the first letter of each word".

shepmaster:
I *may* have forgotten about this PR for a while. I've removed all `PascalCase` and replaced with `UpperCamelCase`. When this is merged, I'll fix the api-guidelines repo to match.

nodingneu:
Rust is the first time I've heard of UpperCamelCase in 500 years of coding. I'm sad it's not called PascalCase here too.

shepmaster:
@nodingneu that's wonderful! I'm glad that we got to be a small part of your learning today. I hope that when I am 500 years old I am still able to learn something new every day as well. Have an enriching learning journey!

PR #2408, opened 2018-04-19
Fix tracking issue in defaulted type params RFC

kamalmarhubi:
It previously pointed to one closed as a duplicate


aturon:
Thanks!

PR #2417, opened 2018-04-24
widening_mul

strake:
[View](https://github.com/strake/rfcs.rs/blob/carrying_mul/text/0000-widening-mul.md)

### DIFF
--- a/text/0000-carrying-mul.md
+++ b/text/0000-carrying-mul.md
@@ -1,0 +1,64 @@
+- Feature Name: carrying_mul
+- Start Date: 2018-04-24
+- RFC PR: 
+- Rust Issue: 
+
+# Summary
+[summary]: #summary
+
+Add an inherent method to integral types which does a double-wide multiplication.
+
+# Motivation
+[motivation]: #motivation
+
+Double-wide multiplication is a prerequisite of arbitrary-precision multiplication. Many machine architectures specify an instruction for this operation, and it is cumbersome to otherwise define. It is also a prerequisite to completely define a `u2size` type of double the size of a machine word which, in the author's experience, is useful at times (e.g. modular arithmetic on offsets into large cyclic arrays of unusual length, smoothsort).
+
+For an integral type of known width, other than the widest (now `u128`), one can define the operation in terms of a wider type, but `usize` has unknown width, and Rust may want to support 128-bit architectures (e.g. RV128) in future.
+
+As the author writes, the "num-bigint" crate merely uses `u32` as its word type to avoid this difficulty, which is likely suboptimal on 64-bit architectures.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+`pub fn carrying_mul(self, other: Self) -> (Self, Self)`
+
+Returns the low and high words of the product of `self` and `other`.
+
+One can, for example, define 
+Explain the proposal as if it was already included in the language and you were teaching it to another Rust programmer. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how Rust programmers should *think* about the feature, and how it should impact the way they use Rust. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing Rust programmers and new Rust programmers.
+
+For implementation-oriented RFCs (e.g. for compiler internals), this section should focus on how compiler contributors should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+`pub fn carrying_mul(self, other: Self) -> (Self, Self)`
+
+Returns the low and high words of the product of `self` and `other`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None known
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+The alternative is to not define this method, which means to do a double-wide multiplication, the user must use inline asm (unstable) or do an awkward dance of shifts and multiplications.
+
+# Prior art
+[prior-art]: #prior-art
+
+This feature is already in many assembly languages (e.g. "mul" on x86, "mulh" on RISC-V M).
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- What should we call the method?
+- What should the order of return values be?
--- a/text/0000-carrying-mul.md
+++ b/text/0000-carrying-mul.md
@@ -49,7 +49,8 @@ None known
 # Rationale and alternatives
 [alternatives]: #alternatives
 
-The alternative is to not define this method, which means to do a double-wide multiplication, the user must use inline asm (unstable) or do an awkward dance of shifts and multiplications.
+- We could not define this method, which means to do a double-wide multiplication, the user must use inline asm (unstable) or do an awkward dance of shifts and multiplications.
+- We could define a `mul_high` method which merely returns the high word.
 
 # Prior art
 [prior-art]: #prior-art
--- a/text/0000-carrying-mul.md
+++ b/text/0000-carrying-mul.md
@@ -20,10 +20,6 @@ As the author writes, the "num-bigint" crate merely uses `u32` as its word type 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-`pub fn carrying_mul(self, other: Self) -> (Self, Self)`
-
-Returns the low and high words of the product of `self` and `other`.
-
 One can, for example, define 
 Explain the proposal as if it was already included in the language and you were teaching it to another Rust programmer. That generally means:
 
--- a/text/0000-carrying-mul.md
+++ b/text/0000-carrying-mul.md
@@ -20,16 +20,19 @@ As the author writes, the "num-bigint" crate merely uses `u32` as its word type 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-One can, for example, define 
-Explain the proposal as if it was already included in the language and you were teaching it to another Rust programmer. That generally means:
+In general, the product of an m-bit and an n-bit number has (m+n) bits. If one wishes to define arbitrary-precision arithmetic, one (usually) chooses a word size, and defines the multiple-precision operations in terms of primitive operations on this type. The `carrying_mul` function allows one to do so conveniently, for example:
 
-- Introducing new named concepts.
-- Explaining the feature largely in terms of examples.
-- Explaining how Rust programmers should *think* about the feature, and how it should impact the way they use Rust. It should explain the impact as concretely as possible.
-- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
-- If applicable, describe the differences between teaching this to existing Rust programmers and new Rust programmers.
+```rust
+pub struct u2size { msw: usize, lsw: usize }
 
-For implementation-oriented RFCs (e.g. for compiler internals), this section should focus on how compiler contributors should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
+impl Mul for u2size {
+    fn mul(self, other: Self) -> Self {
+        let (lsw, c) = self.lsw.carrying_mul(other.lsw);
+        u2size { lsw, msw: c + self.msw * other.lsw
+                             + self.lsw * other.msw }
+    }
+}
+```
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
--- a/text/0000-widening-mul.md
+++ b/text/0000-widening-mul.md
@@ -1,4 +1,4 @@
-- Feature Name: carrying_mul
+- Feature Name: widening_mul
 - Start Date: 2018-04-24
 - RFC PR: 
 - Rust Issue: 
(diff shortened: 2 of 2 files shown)


le-jzr:
I think we could come up with a better name for the method, but otherwise this is something I'd very much like to see added. Given that on some architectures (x86 to name one), multiplication always produces result twice the width of the operands, it seems rather inefficient to not provide a way to use the higher half.

strake:
@le-jzr We can also choose an efficient implementation on architectures where it doesn't and save users the bother.

Some name ideas:
- `carrying_mul`
- `double_mul`
- `double_wide_mul`

strake:
For reference, here is the generic definition from my [u2size crate](https://crates.io/crates/u2size):
```rust
#[inline(always)]
fn carrying_mul(x: usize, y: usize) -> (usize, usize) {
    let n = 0usize.count_zeros() >> 1;
    let halves = |x| (x >> n, x & (!0 >> n));
    let ((x1, x0), (y1, y0)) = (halves(x), halves(y));
    let (z2, z0) = (x1 * y1, x0 * y0);
    let z1 = (x1 + x0) * (y1 + y0) - z2 - z0;
    let (w, c) = usize::overflowing_add(z0, z1 << n);
    (z2 + z1 >> n + c as u32, w)
}
```

The emitted machine code for this is quite heinous.

le-jzr:
Personally, I'd go with `wide_mul`.
I don't think consistency with `wrapping` and `overflowing` is helpful. Intuitively, it seems to me to be in a completely different category.

le-jzr:
After a bit of though, I see the logic behind `widening_mul`. It could work.

tspiteri:
@strake I know it's not really the point, but couldn't you do something like this?
```rust
#[inline(always)]
fn carrying_mul(x: usize, y: usize) -> (usize, usize) {
    let usize_bits = 0usize.count_zeros();
    if usize_bits == 64 {
        // assuming carrying_mul_64 is implemented
        let (a, b) = carrying_mul_64(x as u64, y as u64);
        (a as usize, b as usize)
    } else if usize_bits == 32 {
        // assuming carrying_mul_32 is implemented
        let (a, b) = carrying_mul_32(x as u32, y as u32);
        (a as usize, b as usize)
    } else {
        // fallback
        let n = usize_bits >> 1;
        let halves = |x| (x >> n, x & (!0 >> n));
        let ((x1, x0), (y1, y0)) = (halves(x), halves(y));
        let (z2, z0) = (x1 * y1, x0 * y0);
        let z1 = (x1 + x0) * (y1 + y0) - z2 - z0;
        let (w, c) = usize::overflowing_add(z0, z1 << n);
        (z2 + z1 >> n + c as u32, w)
    }
}
```
You could also add a case for `usize_bits == 16` if that is necessary. I don't think `usize` can actually be 8 bits wide anywhere, currently or in the future.

devonhollowood:
Maybe instead of returning a tuple, it would make sense to return a struct? So e.g. `u32::carrying_mul()` would return a `carried_product<u32>`. This struct would have `high` and `low` fields. This makes it hard to confuse the fields, which would otherwise be an error-prone and hard to lint against. (Feel free to bikeshed on all the names here.)

eddyb:
I have *only* seen this called "widening multiply", but mostly in low-level (compiler/ISA) contexts.

joshtriplett:
I do wonder why this is a `Mul` impl on `u2size` rather than a mul of two `usize` values that returns a `u2size` (or equivalent such as a 2-tuple of `usize`, which I'd prefer over a new type).

nagisa:
Despite what the drawbacks section says, `a as u128 * b as u128` results in
ideal code for widening u64 multiplication (with or without optimisations,
provided overflow checks are disabled.) Very similar results are obtained
if result is split into two parts afterwards.

tspiteri:
Regarding output tuple order, I think it could be made more obvious using the method name. For example `fn mul_low_high(self, multiplier: Self) -> (Self, Self)` makes it more obvious that `ret.0` is the low word and `ret.1` is the high word, and `let (high, low) = a.mul_low_high(b);` sets off alarm bells.

strake:
@joshtriplett the `Mul` impl is merely an example of use; the actual method is `fn(usize, usize) -> (usize, usize)`.

@nagisa yes, it does for me too, but i as a crate author worry whether it will remain so on other targets — will it generate such code on a 32-bit machine? a 128-bit machine? is Rust willing to preclude ever supporting 128-bit targets (e.g. RV128I)? I would prefer to have a method which means efficient double-wide multiplication rather than rely on the compiler to figure out what i mean and myself to not mess it up every time.

@rkruppe thanks for verbalizing these points — yes, the notion is to have a canonical form so neither the compiler nor a future reader needs to guess what is going on.

clarcharr:
Personally I think that the return type should be the next size up of an integer type. I wouldn't mind creating `u256` and `i256` types that only allow bit operations and casting.

JonSeverinsson:
As the order of the high and low half of the result tuple is not obvious suggests to me that a tuple is probably not the best return type. The obvious solution would be a struct with two fields named "high" and "low", but I'm wondering if simply returning a wider integer type is not simpler (though that means i128/u128 will unfortunately lack the method).

I.e. add the following functions (and the equivalent for unsigned integers) to the standard library.
```
impl i8 {
    pub fn widening_mul(self, rhs: Self) -> i16 { ... }
}
impl i16 {
    pub fn widening_mul(self, rhs: Self) -> i32 { ... }
}
impl i32 {
    pub fn widening_mul(self, rhs: Self) -> i64 { ... }
}
impl i64 {
    pub fn widening_mul(self, rhs: Self) -> i128 { ... }
}
impl isize {
    #[cfg(target_pointer_width = "16")]
    pub fn widening_mul(self, rhs: Self) -> i32 { ... }
    #[cfg(target_pointer_width = "32")]
    pub fn widening_mul(self, rhs: Self) -> i64 { ... }
    #[cfg(target_pointer_width = "64")]
    pub fn widening_mul(self, rhs: Self) -> i128 { ... }
}
```

eddyb:
@JonSeverinsson Not sure there is much of a point in doing that because you can simply cast to the wider type and perform the multiplication.

le-jzr:
Returning twice as wide type instead of a tuple has the issue that it wouldn't work on `usize`, which is the canonical use case for this kind of operation (at least until someone adds `ireg`/`ureg`).

strake:
I'd be fine with returning either a tuple or a struct. Returning twice-as-wide type is impossible for `usize` (unless we define `u2size` in core, which i consider far out of scope of this RFC).

kennytm:
Does it make sense to define `widening_mul` on signed integers?

```rust
fn widening_mul(self, other: isize) -> (isize, usize);
```

clarcharr:
@kennytm absolutely imho.

scottmcm:
> Double-wide multiplication is a prerequisite of arbitrary-precision multiplication

I took [a stab](https://play.rust-lang.org/?gist=d27121df379ceac785b6d2876e3c154b&version=stable&mode=debug) at the simplest version of that using the proposed method, and I think it ends up being something like this:

```rust
fn scalar_mul_assign(big: &mut Vec<u8>, scalar: u8) {
    let mut carry = 0;
    for i in big.iter_mut() {
        let (low, high) = i.wide_mul(scalar);
        let (low, overflowed) = low.overflowing_add(carry);
        *i = low;
        carry = high + if overflowed { 1 } else { 0 };
    }
    if carry > 0 {
        big.push(carry);
    }
}
```

That's not great, and I doubt LLVM will do a great job with it.

So I think it'd be good to have the carry built-in, so it can be something like

```rust
pub fn scalar_mul_assign2(big: &mut Vec<u8>, scalar: u8) {
    let mut carry = 0;
    for i in big.iter_mut() {
        let (x, c) = i.mul_with_carry(scalar, carry);
        *i = x;
        carry = c;
    }
    if carry > 0 {
        big.push(carry);
    }
}
```

Since one can always pass 0 as the carry, which LLVM will fold away easily.

Pratyush:
+1, would make writing certain cryptographic code *very* simple

clarcharr:
One interesting use case for this might be a more efficient version of the `multiply_and_divide` function that `num-integer` uses internally:

```
/// Calculate r * a / b, avoiding overflows and fractions.
///
/// Assumes that b divides r * a evenly.
fn multiply_and_divide<T: Integer + Clone>(r: T, a: T, b: T) -> T {
    // See http://blog.plover.com/math/choose-2.html for the idea.
    let g = gcd(r.clone(), b.clone());
    r/g.clone() * (a / (b/g))
}
```

I mean, `num-integer` specifically almost certainly won't use this as it's not backwards compatible, but it is a nice use case.

nagisa:
I just found an ideal use case for `widening_mul` that is not really handled by anything else in the language currently, ignoring bare, and unstable, assembly.

The use case in question is more checked multiplication, especially for `i128/u128`, solving https://github.com/rust-lang/rust/issues/53023.

The implementation essentially would be:

```rust
fn checked_mul(self, other: i128) -> Option<i128> {
    let (high, low) = self.widening_mul(other);
    if high != 0 {
        None
    } else {
        Some(low)
    }
}
```

which results in a very significantly better, although still not ideal, code than what we do currently (division over i128s).

<details> 
<summary>x64 machine code</summary>

```asm
	pushq	%r15
	pushq	%r14
	pushq	%rbx
	movq	%rcx, %r10
	movq	%rdx, %r9
	movq	%r9, %rcx
	sarq	$63, %rcx
	movq	%r8, %rbx
	sarq	$63, %rbx
	movq	%rsi, %r14
	imulq	%rbx, %r14
	movq	%rsi, %rax
	mulq	%rbx
	movq	%rax, %r11
	addq	%r14, %rdx
	imulq	%r9, %rbx
	addq	%rdx, %rbx
	movq	%rcx, %r15
	imulq	%r8, %r15
	movq	%rcx, %rax
	mulq	%r10
	movq	%rax, %r14
	addq	%r15, %rdx
	imulq	%r10, %rcx
	addq	%rdx, %rcx
	addq	%r11, %r14
	adcq	%rbx, %rcx
	movq	%r10, %rax
	mulq	%rsi
	movq	%rdx, %r15
	movq	%rax, %r11
	movq	%r8, %rax
	mulq	%rsi
	movq	%rdx, %rbx
	movq	%rax, %rsi
	addq	%r15, %rsi
	adcq	$0, %rbx
	movq	%r10, %rax
	mulq	%r9
	movq	%rdx, %r15
	movq	%rax, %r10
	addq	%rsi, %r10
	adcq	%rbx, %r15
	setb	%al
	movzbl	%al, %esi
	movq	%r8, %rax
	mulq	%r9
	addq	%r15, %rax
	adcq	%rsi, %rdx
	addq	%r14, %rax
	adcq	%rcx, %rdx
	xorl	%ecx, %ecx
	cmpq	$1, %r10
	sbbq	$0, %rax
	sbbq	$0, %rdx
	movl	$0, %eax
	jb	.LBB0_2
	movq	%r11, 16(%rdi)
	movq	%r10, 24(%rdi)
	xorl	%eax, %eax
	movl	$1, %ecx
.LBB0_2:
	movq	%rcx, (%rdi)
	movq	%rax, 8(%rdi)
	movq	%rdi, %rax
	popq	%rbx
	popq	%r14
	popq	%r15
	retq
```

</details>

sm-Fifteen:
I'd like to point out that the [apfloat component of the official librustc has had a `widening_mul` function for well over a year](https://github.com/rust-lang/rust/blame/a62af858e003e4077f5f4554c3235a4c82636b7e/src/librustc_apfloat/ieee.rs#L2600-L2619), so there's some precedent for this in official rust projects, just not in the language itself.

TheIronBorn:
The `rand` library uses [`wmul`](https://github.com/rust-random/rand/blob/6add2fd439f959989ee371b3eab94607fe173ece/src/distributions/utils.rs#L19)

eddyb:
@sm-Fifteen It is no coincidence I wrote that function as a non-allocating bigint multiplication helper when I ported APFloat from C++ *and* https://github.com/rust-lang/rfcs/pull/2417#issuecomment-384118898 - so don't count my "vote" for `widening_mul` twice, heh.

strake:
Consensus for name seems to be `widening_mul`; modified RFC.

fstirlitz:
What's the rationale for having `u2size` at all? Given that `usize` is supposed to be used for pointers, offsets and memory sizes, if multiplication involving it overflows, it means the result represents an amount of memory you cannot possibly allocate or refer to. You shouldn't even care what the value is: all that really matters is that it's too large to handle.

clarcharr:
@fstirlitz One use case would be intermediate results that may overflow, e.g. rationals. While the end result of x*y/z might fit in a `usize`, the product `x*y` may not.

lu-zero:
Would be good to have the variants that return the wider type directly, since it has direct uses for a number of common dsp operations.

strake:
@lu-zero `u2size` is not in core/std; i merely cite it as an example use of this operation. For fixed-width types, you can merely cast to the wider type before multiplying, and the emitted code will use the machine instruction for widening multiplication.

rkruppe:
I think it's sad that this petered out. This is a small, useful API addition which apparently isn't too controversial or prone to vigorous bikeshedding (except the name, but that's now resolved). Maybe we should just skip ahead to implementing it? Evidently the RFC process is too inert to keep this RFC PR moving.

Centril:
> Maybe we should just skip ahead to implementing it?

That's usually a good idea for small libs features.

clarfonthey:
So I found myself needing this again today, and it looks like the overall decision was to just implement this? I would be willing to help do that, but I honestly have no idea how -- I'm not sure what LLVM intrinsics are used to do wide-multiplication, and I couldn't find the docs for it. If we at least add some intrinsics for this it would make implementing a working solution easier, and we can figure out the exact API later.

scottmcm:
@clarfonthey there's no intrinsic for it; it represents it with `n[su]w` multiplication on the wider integer type.

Demo: https://rust.godbolt.org/z/YhdKnj4Tb

(I do think that people writing Rust shouldn't need to know that, though, and I could imagine the best way to implement it in cranelift could be different, which would be another reason to have an exposed specialized method for it.)

scottmcm:
This was accepted for nightly without the RFC, tracking issue https://github.com/rust-lang/rust/issues/85532

https://doc.rust-lang.org/nightly/std/primitive.u32.html#method.widening_mul
https://doc.rust-lang.org/nightly/std/primitive.u32.html#method.carrying_mul

As such, I'm going to close this on the assumption that it's no longer needed.  Feel free to reopen if so desired.

PR #2439, opened 2018-05-13
RFC for DerefMove

alercah:
[Rendered](https://github.com/alercah/rfcs/blob/deref-move/text/0000-deref-move.md)

Part of #997.

### DIFF
--- a/text/0000-deref-move.md
+++ b/text/0000-deref-move.md
@@ -1,0 +1,322 @@
+- Feature Name: `deref_move`
+- Start Date: 2018-05-05
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add a new `DerefMove` trait that allows consuming a smart pointer to move its
+contents, as in `let x = *p;`.
+
+For more background and discussion, see #178 and #997.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, two smart pointer traits provide access to values contained in smart
+pointers: `Deref` and `DerefMut`. These traits allow overloading of the
+dereferencing (unary `*`) operator, and by extension participate in autoderef in
+a number of places, such as method calls and index expressions.
+
+These two traits, however, only dereference references to produce references to
+the contained value. They do not offer the option to dereference by value,
+consuming the smart pointer to produce the owned value inside. As a special
+case, `Box<T>` has compiler support allowing it to be dereferenced (explicitly
+or implicitly) to produce the inner `T`. Because this is special-cased in the
+compiler, this functionality is not available for other smart pointer types.
+
+Other smart pointers may support this consuming operation, but it must be done
+explicitly due to the lack of language support. For instance, in the stdlib, the
+following methods both consume a smart pointer to produce the inner value:
+
+* `ManuallyDrop<T>::into_inner()`
+* `binary_heap::PeakMut<'a, T>::pop()`
+* `Cow<'a, B>::into_owned()`
+
+I have not made any attempt to survey into the larger ecosystem for more
+examples, but there are probably many more.
+
+Note that removing `FnBox` is explicitly *not* part of the motivation for this
+RFC. It has been suggested in a few places, such as in discussion of
+rust-lang/rust#28796, that `DerefMove` would allow `Box<FnOnce()>` to be called
+without needing the `FnBox` workaround. This is not the case, however: the inner
+value is dynamically-sized and thus cannot be moved onto the stack. Indeed, one
+can call a `Box<T>` where `T` is a concrete type implementing `FnOnce()`.
+`DerefMove` would allow `Box<FnOnce()>` to continue not working, just without a
+special case in the compiler to allow it to do so.
+
+This will not remove all of `Box`'s special cases from the compiler, but
+represents progress towards that goal.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The following trait is defined in `std::ops`:
+
+```rust
+trait DerefMove : DerefMut {
+    fn deref_move(self) -> Self::Target;
+}
+```
+
+This trait allows moving out of the result of a derefence. This can happen
+either explicitly, as in `let x = *p;`, or implicitly, as in `p.into_foo()`.
+`Box<T>` from the standard library implements `DerefMove`, allowing you to move
+out of it.
+
+```rust
+let b = Box::new("s".to_owned());
+let s = *b;
+// ERROR: b has already been moved on the previous line.
+// let t = *b;
+```
+
+If the `Target` type is `Copy`, then `DerefMove` will not be used, and instead
+`Deref` will be used and the resulting referenced value will be copied:
+
+```rust
+let b = Box::new(0i32);
+// Since i32 is Copy, this is equivalent to i = b.deref().
+let i = *b;
+// That makes this legal, since b was not moved.
+let j = *b;
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+`DerefMove` is a lang item that, like `Deref` and `DerefMut`, plays a role in
+dereferencing, both explicitly with the unary `*` operator and implicitly with
+autoderef.
+
+`deref_move` is only called in contexts where a value is needed to be moved out.
+This can occur either because an explicit dereference is used in a context where
+a value is to be moved, or because autoderef selects a method taking `self` (or
+`Box<self>`) as a parameter. In both cases, when evaluating whether to call
+`deref_move`, the compiler will look at whether `Target` implements `Copy`. If
+it does not, then `deref_move` is called to move the target out of the smart
+pointer. If it does, then instead the compiler will only call `deref` and then
+copy out of the returned reference. For generic types, the compiler only checks
+if a `Copy` bound is available at the use site; it is not done during
+monomorphization.
+
+For explicit dereference, the changes are straightforward. Any explicit
+dereference of a `DerefMove` type becomes a movable expression, just as
+dereference of `Box` is today.
+
+For autodereference: 
+
+* In a function call expression, the callee operand may be dereferenced and
+  moved if the `Target` implements `FnOnce` and no reference-based call trait.
+* In a method call expression, the receiver operand may be dereferenced and
+  moved if the function selected by overload resolution takes `self` or
+  `Box<self>` as a parameter (in the latter case, this would require a smart
+  pointer containing a `Box<Self>`, of course).
+* In a field access expression, the left-hand operand may be dereferenced and
+  moved if the result of the expression is moved. In this case, the move is
+  always entire; partial move is not supported when a dereference is necessary.
+* An index expression can never move its left operand, so it is unaffected.
+* A borrow expression can never move its operand, so it is unaffected.
+
+No change is made to overload resolution of method calls. This means that calls
+which are invalid today due to preferring `T` over `&T` may now be legal,
+selecting to move out of the smart pointer rather than giving an error.
+
+There is an ambiguity in a method call expression when the `Target` type of the
+receiver is a reference, and that reference type has a method taking `self`. In
+this case, autodereference could either dereference via `deref_move` or via
+`deref` or `defer_mut`. In this case, it will prefer the latter (for `deref`, it
+is a special case of the `Copy` rule above, but for `deref_mut` it is not) to
+avoid unnecessarily moving out of the receiver.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Nobody understands autoderef already and this makes it worse
+
+Or at least, it would if `Box` wasn't already complicating it in the same way.
+
+## Copy behaviour is unintuitive in generic cases
+
+If a type has nontrivial/side-effectful `defer_move`, it may be slightly
+confusing why it is not used for `Copy` types in non-generic code, but is used
+in generic code.
+
+## Breaking change to an edge case of `Box<&mut T>`
+
+Currently, a `Box<&mut T>` variable [can be
+dereferenced](https://play.rust-lang.org/?gist=5b3f046ecf29da8f00fde3bd1d2a3df4&version=nightly&mode=debug)
+in at least some cases into a `&mut T` without either requiring that the
+variable be mutable (as would be the case for any other `DerefMut` type), or
+moving out of the `Box`. This behaviour would not be maintained if its magical
+dereferencing were replaced with `DerefMove`, as `DerefMut` requires a mutable
+variable.
+
+This is an edge case, as it requires `Box<&mut T>`, which is an unlikely type,
+and in most cases it can be trivially fixed by making the variable mutable. It's
+possible there are more complex cases (such as if the `Box` itself is stored in
+another smart pointer) where this is not easily fixed, but I would be
+unsurprised if there were few, if any, instances of this in extant code.
+Moreover, the behaviour is arguably a bug because it is taking a mutable borrow
+of an immutable variable (and one can modify the above example to get an error
+message implying exactly that), though the semantics are more like the unique
+immutable borrows used by closures.
+
+As a result, I propose that this breaking change be accepted after verifying its
+effect on Cargo packages.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+## Copy handling
+
+The handling of copies is the real complexity of `DerefMove`. Currently, one can
+copy out of a `Box` without consuming it, so this desirable behaviour needs to
+be preserved. The proposal is to do so by special casing in the compiler.
+However, it could also be done via, for instance, a separate method in the Deref
+trait definition:
+
+```rust
+   fn deref_copy(&self) -> Self::Target where Self::Target : Copy {
+       return self.deref();
+   }
+```
+
+This still requires special-casing in the compiler, but it is less magical. One
+could even envision a `DerefCopy` trait with similar purpose.
+
+These approaches would, however, make diverging implementations (that is,
+implementations for which the behaviour of the various deref traits is
+unintuitive because they don't actually refer to the same underlying value) more
+difficult to write, and we want to avoid that. And they would not fundamentally
+solve the special casing in the compiler; to do that, we would really need a
+more advanced type system allowing us to condition the type of `self` in
+`deref_move` on whether or not `Target : Copy`.
+
+With negative trait bounds, we could also bound `DerefMove` on `Target : !Copy`,
+eliminating the problem, at the cost of requiring every implementation to also
+bound on `!Copy`.
+
+## Status quo
+
+The status quo is disfavoured because it requires special-casing `Box` in the
+compiler, privileging it over other smart pointers. While there's no obvious
+call in the stdlib for `DerefMut` other than `Box` (see below), it prevents
+other library authors from writing similar code.
+
+## `IndexMove` trait
+
+One option considered in the early discussion was an `IndexMove` trait, similar
+to `DerefMove`, which would consume a collection and return a single element. I
+have not included this in this RFC because no compelling use case was presented,
+and the behaviour of `let s = v[0];` actually consuming `v` seems likely to add
+cognitive overhead to the language.
+
+Compared to other methods of moving out of a collection, `IndexMove` may provide
+slight micro-optimization in some cases, but it's not clear without benchmarks
+that this is actually meaningfully better than using, say,
+`.into_inter().nth(n)`. Given that the `IndexMove` was proposed largely from
+parallel construction from `DerefMove` without any specific use cases, it does
+not seem worthwhile to move further with it at this time.
+
+## Weakening `DerefMut` bound
+
+The `DerefMut` bound on `DerefMove` and, in particular, the `Deref` bound that
+it implies, constrain implementors somewhat. It requires that `Deref` be
+implementable, so the implementation is not allowed to overload dereferencing in
+such a way that it produces a new value each time. This would not be useful for
+non-`Copy` types, since they would be consumed, but a `Copy` type could in this
+fashion allow itself to be copied and then produce a new value on each
+derefence. This would be useful to avoid redefining methods on the outer type.
+
+This very much goes against the intent of the `Deref` traits to be restricted
+smart pointers, however. `Deref` is currently the only way for methods of one
+type to be callable on another, but it is deliberately limited in scope. For
+instance, one might be tempted to implement `Deref` on a newtype for this
+purpose, but this is considered unidiomatic. Additionally, this is a bandaid for
+the fact that newtypes would also often like to be able to inherent trait
+implementations, and inhereting methods is not sufficient for that.
+
+The additional requirement is to avoid complication of the following case:
+
+```rust
+struct RefMethod();
+
+trait Consume : Sized {
+    fn consume(self) {}
+}
+
+impl<'a> Consume for &'a mut RefMethod {}
+
+struct BadDeref<'a> (&'a mut RefMethod);
+
+impl<'a> Deref for BadDeref<'a> {
+    type Target = RefMethod;
+    fn deref(&self) -> &Self::Target { &self.0 }
+}
+
+impl<'a> DerefMove for BadDeref<'a> {
+    fn deref_move(self) -> Self::Target { self.0 }
+}
+
+fn main() {
+    let mut r = RefMethod{};
+    let b = BadDeref(&mut r);
+    b.consume();
+    b.consume();
+}
+```
+
+In this case, the first `b.consume()` is legal, but the second `b.consume()` is
+not since `b` was already moved out of. But if `BadDeref` also implements
+`DerefMut`, then the first `b.consume()` becomes illegal because `deref_mut`
+becomes preferred and `b` is immutable. By requiring that `DerefMove :
+DerefMut`, this edge case cannot be encountered: this code will always be
+illegal unless `b` is made mutable.
+
+# Prior art
+[prior-art]: #prior-art
+
+The compiler special case for `Box` is the most compelling prior art. This RFC
+is written with that behaviour in mind; one of the most important criteria here
+is that `Box` should remain completely backwards-compatible. In case of any
+conflict between current `Box` behaviour and the behaviour specified in this
+RFC that isn't explicitly called out as being different, it should be considered
+a bug in the RFC and either this RFC should be corrected or more discussion
+should ensue.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## Implementors
+
+It's not clear to me which, if any, of the smart pointers above should implement
+`DerefMove` other than `Box`. It requires discussion of whether inadvertently
+moving out of the pointer could cause issues. I think that it's reasonable to
+start with only `Box` for now:
+
+* `Cow` does not implement `DerefMut`, so it is ineligible.
+* `PeekMut`'s `pop` has side effects on the containing collection, and so it
+  seems best to require it to be moved explicitly.
+* Moving out of a `ManualDrop` causes the newly-moved value to regain normal
+  drop semantics. This seems like the least dangerous, given the mechanics of
+  `ManualDrop`, but it still probably deserves separate consideration.
+
+Additional implementations could, of course, be done as separate proposals for
+the libs team.
+
+## Preferring to copy the pointer
+
+Some earlier discussion proposed that when a type implements both `DerefMove`
+and `Copy`, then the special case of a `Copy` target does not apply. In other
+words, given `let x = *p;`, if both `p` and `p`'s target type are `Copy`, then
+rather than calling `deref` and copying the result, this assignment would call
+`deref_mut`, copying `p` to use as the receiver.
+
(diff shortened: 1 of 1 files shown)


comex:
👍, but note that `Cow::into_owned` is probably not a good candidate for`DerefMove` since it potentially performs an expensive operation (cloning).

alercah:
Right, that's why it doesn't implement `DerefMut` either, I believe.

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+examples, but there are probably many more.
+
+Note that removing `FnBox` is explicitly *not* part of the motivation for this
+RFC. It has been suggested in a few places, such as in discussion of
+rust-lang/rust#28796, that `DerefMove` would allow `Box<FnOnce()>` to be called
+without needing the `FnBox` workaround. This is not the case, however: the inner
+value is dynamically-sized and thus cannot be moved onto the stack. Indeed, one
+can call a `Box<T>` where `T` is a concrete type implementing `FnOnce()`.
+`DerefMove` would allow `Box<FnOnce()>` to continue not working, just without a
+special case in the compiler to allow it to do so.
```
Gankro: These last two sentences are jumbled up, and I'm not actually 100% clear what they're trying to express. Maybe just:

> Although `DerefMove` *would* allow smart pointers containing a *concrete* `FnOnce` instance to be called. 

And it may be worth noting that if by-value DSTs ever get implemented (which is theoretically possible and has long been desireable), then the combination of that feature with DerefMove would in fact be sufficient to make `Box<FnOnce()>` work.

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+
+Add a new `DerefMove` trait that allows consuming a smart pointer to move its
+contents, as in `let x = *p;`.
+
+For more background and discussion, see #178 and #997.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, two smart pointer traits provide access to values contained in smart
```
Gankro: nix the first instance of "smart pointer", I think ("two traits provide...")

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+pointers: `Deref` and `DerefMut`. These traits allow overloading of the
+dereferencing (unary `*`) operator, and by extension participate in autoderef in
+a number of places, such as method calls and index expressions.
+
+These two traits, however, only dereference references to produce references to
+the contained value. They do not offer the option to dereference by value,
```
Gankro: This sentence is a mouthful. Perhaps:

> Unfortunately, these two traits only provide borrowed access to the contained value.

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+With negative trait bounds, we could also bound `DerefMove` on `Target : !Copy`,
+eliminating the problem, at the cost of requiring every implementation to also
+bound on `!Copy`.
+
+## Status quo
+
+The status quo is disfavoured because it requires special-casing `Box` in the
+compiler, privileging it over other smart pointers. While there's no obvious
+call in the stdlib for `DerefMut` other than `Box` (see below), it prevents
+other library authors from writing similar code.
```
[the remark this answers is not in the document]
alercah: Ah yes, will fix.

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+No change is made to overload resolution of method calls. This means that calls
+which are invalid today due to preferring `T` over `&T` may now be legal,
+selecting to move out of the smart pointer rather than giving an error.
+
+There is an ambiguity in a method call expression when the `Target` type of the
+receiver is a reference, and that reference type has a method taking `self`. In
+this case, autodereference could either dereference via `deref_move` or via
+`deref` or `defer_mut`. In this case, it will prefer the latter (for `deref`, it
+is a special case of the `Copy` rule above, but for `deref_mut` it is not) to
+avoid unnecessarily moving out of the receiver.
```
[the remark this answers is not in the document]
alercah: Yeah, it's a bit confusing. If the Target is `&T`, and there's a `self` method on `&T`, then since `&T` is `Copy`, the rules above will mean that we use `deref` and then copy. But with `Target = &mut`, there are two options: either call `deref_mut` getting an `&mut &mut T` and then dereference that, or just call `deref_move` to get a `&mut T`; the former should be preferred.

I'll try to clarify this paragraph.

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+`DerefMove` is a lang item that, like `Deref` and `DerefMut`, plays a role in
+dereferencing, both explicitly with the unary `*` operator and implicitly with
+autoderef.
+
+`deref_move` is only called in contexts where a value is needed to be moved out.
+This can occur either because an explicit dereference is used in a context where
+a value is to be moved, or because autoderef selects a method taking `self` (or
+`Box<self>`) as a parameter. In both cases, when evaluating whether to call
```
Gankro: I forget, is `Box<self>` magic or would this also work for `MyBox<self>`? (should this be made explicit?)

bstrie:
I find the `DerefMut` bound interesting because it means that the Deref traits will operate in exactly the opposite, er, direction from the Fn traits: DerefMove implies DerefMut implies Deref, whereas Fn implies FnMut implies FnMove. Not sure if this is somehow profound or just an interesting symmetry. :)

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Nobody understands autoderef already and this makes it worse
+
+Or at least, it would if `Box` wasn't already complicating it in the same way.
+
+## Copy behaviour is unintuitive in generic cases
+
+If a type has nontrivial/side-effectful `defer_move`, it may be slightly
```
Gankro: `deref_move` typo

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+confusing why it is not used for `Copy` types in non-generic code, but is used
+in generic code.
+
+## Breaking change to an edge case of `Box<&mut T>`
+
+Currently, a `Box<&mut T>` variable [can be
+dereferenced](https://play.rust-lang.org/?gist=5b3f046ecf29da8f00fde3bd1d2a3df4&version=nightly&mode=debug)
```
Gankro: I would prefer if you inlined this example, so that these documents are less prone to bitrot.

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+
+With negative trait bounds, we could also bound `DerefMove` on `Target : !Copy`,
+eliminating the problem, at the cost of requiring every implementation to also
+bound on `!Copy`.
+
+## Status quo
+
+The status quo is disfavoured because it requires special-casing `Box` in the
+compiler, privileging it over other smart pointers. While there's no obvious
+call in the stdlib for `DerefMut` other than `Box` (see below), it prevents
```
Gankro: typo: `DerefMove`

text/0000-deref-move.md
```diff
@@ -0,0 +1,322 @@
...
+other library authors from writing similar code.
+
+## `IndexMove` trait
+
+One option considered in the early discussion was an `IndexMove` trait, similar
+to `DerefMove`, which would consume a collection and return a single element. I
+have not included this in this RFC because no compelling use case was presented,
+and the behaviour of `let s = v[0];` actually consuming `v` seems likely to add
+cognitive overhead to the language.
```
Gankro: Oh i just thought of a use-case; arrays. IndexMove is necessary to get e.g. a Box out of an Array. But in that case into_iter or slice patterns are probably better.

text/0000-deref-move.md
alercah: By-value DSTs are sufficient and necessary for `Box<FnOnce()>`, even in absence of `DerefMove`. As you noted, concrete `FnOnce` instances can be called with `DerefMove` and they can be called from a `Box` today.

I'll try to clarify this.

text/0000-deref-move.md
alercah: `Box<self>` is magic as a receiver type, and I'm not proposing to eliminate that here. It's only mentioned here for completeness.

text/0000-deref-move.md
comex: Well, there's definitely a use case for an Index trait which *returns* things by value.  For example, consider a type that wants to expose an API similar to `Vec<bool>`, but is implemented as a bit set, so indexing into it can't return an existing `&bool` reference.  However, you wouldn't usually want to have `self` be by value – but afaik that provides the most flexibility, since you could always do `impl<'a> DerefMove for &'a MyBitSet`.  (By itself, this would only allow reading values, not writing them, but #997 also proposed an `IndexSet` that would support writing.)

Gankro: IndexGet is the intended solution to that problem, but yes it could just be provided through IndexMove.

text/0000-deref-move.md
```diff
@@ -0,0 +1,325 @@
...
+This will not remove all of `Box`'s special cases from the compiler, but
+represents progress towards that goal.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The following trait is defined in `std::ops`:
+
+```rust
+trait DerefMove : DerefMut {
```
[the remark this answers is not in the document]
alercah: The rationale is below, but I had not considered this use case. The ergonomic drawbacks may be worth it if this use case is worth supporting.

text/0000-deref-move.md
alercah: I initially worked on all versions proposed by #997, but quickly abandoned that when I realize that the `IndexGet` and `IndexSet` make more sense as a separate RFC. Note that you couldn't impl `DerefMove` on a reference since it would have to have the same `Target` as the provided `Deref` instance; I don't think that the same restriction would necessarily apply to `IndexMove`, however.

text/0000-deref-move.md
```diff
@@ -0,0 +1,325 @@
...
+
+```rust
+let b = Box::new("s".to_owned());
+let s = *b;
+// ERROR: b has already been moved on the previous line.
+// let t = *b;
+```
+
+If the `Target` type is `Copy`, then `DerefMove` will not be used, and instead
+`Deref` will be used and the resulting referenced value will be copied:
```
pythonesque: I think understand why this is necessary after staring at it for a while (since otherwise existing code that expects to be able to copy out of `*x` would be broken as it would move out of `b` incorrectly).  But as I noted above: it's not clear to me why the `Target` of `DerefMove` has to be the same as that for `Deref` since the rule seems to be that if the target for `Deref` is `Copy` we ignore `DerefMove`.  You could always lint for implementations of `DerefMove` with targets that are known to be `Copy` at definition time.

text/0000-deref-move.md
pythonesque: This is sort of getting fixed with arbitrary self types, right?  Or is this specifically about the `Box<self>` syntax?

text/0000-deref-move.md
```diff
@@ -0,0 +1,325 @@
...
+   }
+```
+
+This still requires special-casing in the compiler, but it is less magical. One
+could even envision a `DerefCopy` trait with similar purpose.
+
+These approaches would, however, make diverging implementations (that is,
+implementations for which the behaviour of the various deref traits is
+unintuitive because they don't actually refer to the same underlying value) more
+difficult to write, and we want to avoid that. And they would not fundamentally
```
pythonesque: Do you mean easier to write?

text/0000-deref-move.md
```diff
@@ -0,0 +1,325 @@
...
+    b.consume();
+}
+```
+
+In this case, the first `b.consume()` is legal, but the second `b.consume()` is
+not since `b` was already moved out of. But if `BadDeref` also implements
+`DerefMut`, then the first `b.consume()` becomes illegal because `deref_mut`
+becomes preferred and `b` is immutable. By requiring that `DerefMove :
+DerefMut`, this edge case cannot be encountered: this code will always be
+illegal unless `b` is made mutable.
```
pythonesque: I sort of understand why this is a bad edge case, but as was pointed out above there are useful situations where you can't feasibly implement `DerefMut`; however, this would still be a bit surprising in those cases.  I think what this edge case actually suggests is that `Box` gets it "right" mostly by accident; if it is important, the rules for `mut` bindings should be tweaked around `DerefMove` somehow (possibly that `&mut` methods on a target of `DerefMove` should require a `mut` binding, though I'm not sure how you justify that).  In any case, allowing this edge case still seems preferable to me to ruling out `DerefMove` on `!DerefMut` types.

text/0000-deref-move.md
```diff
@@ -0,0 +1,325 @@
...
+a bug in the RFC and either this RFC should be corrected or more discussion
+should ensue.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## Implementors
+
+It's not clear to me which, if any, of the smart pointers above should implement
+`DerefMove` other than `Box`. It requires discussion of whether inadvertently
```
pythonesque: `Vec` should likely have an implementation; I expect it can't right now because of sizedness issues.  Actually, that suggests to me that allowing by-value DSTs is pretty important to making `DerefMove` work well.

text/0000-deref-move.md
alercah: Yes, once arbitrary receiver types are accepted then `Box` won't be magical. But that shouldn't affect this RFC one way or another---it's just that I felt I needed to mention that not all by-value receivers are `Self`. Overload resolution will continue the way it normally is (note: I said "autoderef selects" above, but I really meant overload resolution).

text/0000-deref-move.md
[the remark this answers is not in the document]
alercah: The behaviour of ignoring `DerefMove` when `Target : Copy` for `Deref` would be complicated if the `Target` types were different; that would possibly create scenarios where a generic smart pointer would fail for a non-`Copy` `Target` because `DerefMove` is ignored but `Deref` has a different `Target`.

This would also complicate autoderef and in particular its relationship with overload resolution, probably unbearably so. I think it would be so hard to use a type with different `Deref` and `DerefMove` `Target`s that I think that, if we decided to make the two independently implementable, the compiler must enforce that the `Target`s are identical.

text/0000-deref-move.md
alercah: Yes, sorry. See my above comment that if we did this we may actually need to enforce that they match.

text/0000-deref-move.md
alercah: The behaviour of `Box` here is the same as the unique immutable reference that closures use to get around the difficulty of an immutable variable holding a mutable reference, which can be expressed in MIR but not in Rust. We could consider making `deref_mut` take `&uniq self`, if that type existed, and then this problem would not exist. Alternatively, simply dropping the notion of immutable variables would also solve this.

text/0000-deref-move.md
alercah: That's a good point. I will revise the RFC to discuss the interaction with by-value DSTs more, since that seems very fundamental.

text/0000-deref-move.md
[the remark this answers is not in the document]
scottmcm: Hmm, I suppose something like `struct TaggedPtr(usize);` is a case where it could plausibly be `DerefMove` but not `Deref` -- the move version is cheap, just masking away the extra bits in the alignment, and conceptually reasonable, since it's still a pointer -- but it can't be `Deref` because it doesn't actually have the real pointer to borrow.

That would need that "split a trait into two" RFC, I think, to introduce `DerefTarget` or something that just provides the associated type.  But it might also be too much of a stretch to bother supporting even if we could...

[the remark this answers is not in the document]
clarcharr: Small nit: this trait requires a `Sized` super trait and `where Self::Target: Sized` bound.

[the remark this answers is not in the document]
pythonesque: @scottmcm The `TaggedPtr` case could still be `Deref`, I think, since `usize` is `Copy`, as long as the target was the pointed-to value, not the pointer itself.  It would require unsafe code to cast the masked `usize` to an actual pointer with a lifetime, but that would require unsafe code anyway--it would still (presumably) be safe.

text/0000-deref-move.md
pythonesque: @alercah 

> The behaviour of ignoring DerefMove when Target : Copy for Deref would be complicated if the Target types were different; that would possibly create scenarios where a generic smart pointer would fail for a non-Copy Target because DerefMove is ignored but Deref has a different Target.

Can you elaborate on what you mean by "fail"?  I assume all that would happen is that the Deref target and implementation would be used on autoderef, which seems no different than what would happen if the targets were the same.

> This would also complicate autoderef and in particular its relationship with overload resolution, probably unbearably so.

Wouldn't it always be unambiguous which implementation to use in any given context, in the same way that `Copy` itself is?  I may be missing something.  Or is the problem that the two types could have methods with the same name, and you couldn't tell which one to call or prioritize one over the other?  If that's your concern I definitely see why you would want the types to match.

text/0000-deref-move.md
```diff
@@ -0,0 +1,325 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+The compiler special case for `Box` is the most compelling prior art. This RFC
+is written with that behaviour in mind; one of the most important criteria here
+is that `Box` should remain completely backwards-compatible. In case of any
+conflict between current `Box` behaviour and the behaviour specified in this
+RFC that isn't explicitly called out as being different, it should be considered
+a bug in the RFC and either this RFC should be corrected or more discussion
+should ensue.
```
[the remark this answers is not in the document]
matthewjasper: Also moving from `b.0` doesn't drop the rest of `*b` until `b` is dropped.

```rust
struct D(i32);

impl Drop for D {
    fn drop(&mut self) {
        println!("drop({})", self.0);
    }
}

fn main() {
    let mut b = Box::new((D(0), D(1)));
    drop(b.0);
    println!("Hi!");
    b = Box::new((D(2), D(3))); // old value of b is dropped.
    println!("Hi again!");
}
```

Output:

```text
drop(0)
Hi!
drop(1)
Hi again!
drop(2)
drop(3)
```

text/0000-deref-move.md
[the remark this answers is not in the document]
scottmcm: @pythonesque Good point!  I forget that pointers get to cheat 🙂

mikeyhew:
I'm glad to see someone is pushing forward on this again.

As others have mentioned, I think we will want `DerefMove` to support DST someday. Since the `deref_move` method proposed here returns the target type by value, it can't work if `Target: ?Sized`, even with the Unsized Rvalues RFC, which does not support DST return values. 

That being said, this version of `deref_move` works right now for `Sized` types, and since we don't have the language features yet that would make it work for DST (either `&move` references, or some way to return DSTs by-value), I think it's fine to use this version, and stabilize the effects of it for Box and whatever other types implement it, but keep the trait itself unstable so that it can be changed to support DST in the future.

DDOtten:
How about the name `IntoDeref` instead of `DerefMove`. We already have at least two kinds of names for taking ownership with things like `IntoIter` and `FnOnce`. I think it would be good use one of these for learnability.

clarcharr:
Now that I think of it, I definitely agree with @DDOtten. The convention for `Iterator` should be clear enough that `IntoDeref` is the right choice.

CryZe:
Another alternative would be DerefInto, which has the benefit of using the
Into as well, keeps Deref as a prefix like DerefMut and has a better
grammatical flow to it. However to my knowledge into has always been a
prefix, so that‘s slightly suboptimal.

Clar Roʒe <notifications@github.com> schrieb am Fr. 18. Mai 2018 um 07:03:

bjorn3:
I vote for the name `DerefOnce` in parallel to `FnOnce`

Deref | Fn
-----|------
DerefMut | FnMut
**DerefOnce** | FnOnce

CryZe:
Deref{Into/Move/Once} doesn't imply that you can only deref it a single time though, as it also implies DerefMut and transitively Deref. However FnOnce doesn't imply FnMut or Fn, and thus the trait as a trait bound itself, only allows you to call the function once. So while Once sounds nice on a surface level, the dependencies are quite a bit different.

bjorn3:
@CryZe hadn't thought about that.

mikeyhew:
The name `DerefInto` also makes a bit more sense for this by-value version of `DerefMove`, compared to previous proposals involving `&move` references. Naming it `DerefInto` would allow us to add an `&move` version later and name it `DerefMove`. One could be implemented in terms of the other to avoid breaking backward compatibility:

```rust
// potential trait hierachy

trait DerefInto: DerefMut {
    fn deref_into(self) -> Self::Target where Self: Sized;
}

trait DerefMove: DerefInto {
    fn deref_move(&move self) -> &move Self::Target;
}

impl<T> DerefInto for T
where
    T: DerefMove
{
    fn deref_into(self) -> Self::Target where Self: Sized {
        self.deref_move().into_inner()
    }
}
```

Thoughts?

joshtriplett:
Name bikeshedding aside (I can see a case for either `DerefInto` or `DerefMove`), this seems like a good idea. :+1:

The ability to call `self` methods on a wrapper type seems reasonable.

alercah:
I apologize I haven't come back to this. I got stumped by @matthewjasper's examples. After some more thinking today, I think it might be doable, but we would need something like `&move` in MIR.

Ericson2314:
I still think we'll need a `&move` to do `DerefMove` right. But temporary making an *unstable* `DerefMove` like this is a good first step. The compiler support for `Box` is really an abomination and we should strive to get rid of it.

mikeyhew:
I didn't notice @matthewjasper's comments before. It looks like `DerefInto` wouldn't be backward compatible with `Box` because of the order in which things are dropped.

To get the right drop order, I'm pretty sure either a cleanup RAII type is needed, or `deref_move` needs to take a closure. Here's both versions using `&move` references:

```rust
trait DerefMove: DerefMut {
    type Cleanup;

    // explicit lifetimes for clarity
    fn deref_move<'a>(&'a move self) -> (&'a move Self::Target, Self::Cleanup);
}

trait DerefMove: DerefMut {
    // this also works with unsized rvalues:
    // `fn deref_move<F: FnOnce(Self::Target) -> O, O>(self, f: F) -> O`
    // however, it's a bit sketchy — see below
    fn deref_move<F, O>(&move self, f: F) -> O
    where
        F: FnOnce(&move Self::Target) -> O;
}
```

The closure version looks cool, but is kind of sketchy – what if the implementation doesn't call `f` and just returns `Self::Target`?

alercah:
Yeah, trying to manage partial is pretty hacky. But it might be doable if we can accept some magic. I have an idea of making `deref_move`'s ABI be magical and special, to work like a hypothetical `&move` might, using pass-by-pointer instead of pass-by-value. Some of these ideas come from [this prior RFC](https://github.com/arielb1/rfcs/blob/missing-derefs/text/0000-missing-derefs.md).

At the language level:

1. `deref_move` is, like `drop`, not callable explicitly. Also, like `Drop`, `DerefMove` prevents partial moves.
1. As an exception to the above rule, it is legal to partially move out of `self` inside of `deref_move` 
1. You can partially move out of `*b` where `b` implements `DerefMove`. Subsequent accesses to the partially-moved object do not call `deref_move` again.
1. `self` is not dropped at the end of `deref_move`.
1. When a value of a type implementing `DerefMove` is dropped, first, `deref_move` is called if it wasn't already. Regardless, its return value, or what remains of it, is dropped. `drop` is called, then any remaining fields are moved out of.

The idea here is that `deref_move` is basically "phase 1" of a bigger deref-and-drop operation which works out well. But it runs into recursion issues like those discussed in [this thread](https://internals.rust-lang.org/t/moves-from-self-during-the-drop-hook/1536/18).

As an unrelated note, I think I realized another problem with the proposal: how does `Box<T>` return a moved value from heap? I think that without additional compiler support magic it would be required to manually copy the data onto the stack, which is not at all ideal. So we need some sort of `&move`-like magic *somewhere* to let you get a `T` from a value on the heap without copying from it. It might end up being `deref_move` specific, like you can return `unsafe { *p }` for `p: NonNull<T>` or something, and the compiler understands that this

mikeyhew:
@alercah 

> an unrelated note, I think I realized another problem with the proposal: how does Box<T> return a moved value from heap? I think that without additional compiler support magic it would be required to manually copy the data onto the stack, which is not at all ideal. So we need some sort of &move-like magic somewhere to let you get a T from a value on the heap without copying from it. It might end up being deref_move specific, like you can return unsafe { *p } for p: NonNull<T> or something, and the compiler understands that this

Ya, I think what really needs to happen is `&'a move T` needs to take over from `Box<T>` and do all the magic. Then `Box<T>` will still have all the magical capabilities, it'll just be available to other types too, if they implement `DerefMove`.

alexreg:
What's happening with this RFC lately? 😊

alercah:
I am planning to write up an internals post asking if we want to work on this as it is, or just hold off for `&move`.

alexreg:
@alercah Couldn't the two features interact with each other in a sane way? Anyway, fair enough.

alercah:
As it is, we basically need a special-case of `&move` to make this work, as I discussed above. So it may not be worth implementing unless we're happy to implement `&move` generally.

alexreg:
Okay, sure. Best to discuss it on internals/Discord then, and then come back and edit this RFC accordingly, I guess.

alexreg:
What's the latest on this? I'd definitely be interested in pushing it towards acceptance, even if that means working on the code.

Ericson2314:
It means `&move`. Now that unsized stack variables have landed, I think it would be an especially fun time to implement `&move`.

alexreg:
@Ericson2314 Yeah, sounds good. Does that need a separate RFC, or can people start work on it now? (Indeed, do you have plans to?)

qnighy:
FWIW I published an experimental crate for `&move`: https://crates.io/crates/refmove

Ericson2314:
@alexreg I'm not pushing for that right now. (Allocators and stdlib Cargo I think need me in particular more, without even going into Haskell and NixOS things). So happy to pass the baton on this one. Hopefully with NLL here, the type of passes we'd need are a lot closer to the status quo.

alexreg:
@Ericson2314 

Fair enough. Maybe @alercah would be more interested in moving forwards `&move` syntax, with some help?

(And at the risk of going a little off-topic... what's stdlib Cargo?)

Ericson2314:
(The overall goal is making standard library crates...not special. Stdlib-aware Cargo (oops forgot the "-aware", sorry) is the Cargo aspect of that, meaning being able to . https://github.com/rust-lang/rfcs/pull/1133 was my RFC that essentially died of old age. The plan I think (hope!) is to return to such things after the 2018 epoch.)

alercah:
I'm going to be focusing a bit on documentation in the next little while, I think, as the edition comes round. I'll think about picking this up afterward. In the meanwhile, I'm going to withdraw this PR.

LinuxMercedes:
This would be pretty handy to have for newtypes, as I just ran into a somewhat confusing error message. Example:

```rust
enum Action {
  DoThing(String),
  // etc.
}

struct Actions(Vec<Action>);

impl Deref for Actions {
  /* ... */
}

impl DerefMut for Actions {
  /* ... */
}

fn perform(actions: Actions) {
  for action in actions.into_iter() { // <-- error[E0507]: cannot move out of borrowed content
    // do stuff
  }
}
```

Of course, laid out like this it's perhaps obvious, but if these parts are scattered across a larger project and you're thinking about it as "well `Actions` is basically just a `Vec`, I should be able to consume it with `into_iter()`", it's a bit less obvious what's going on.

alexreg:
@alercah With the 2018 edition release firmly out of the way now, have you considered picking this up again? :-)

alercah:
No, I do not have the time right now.

alexreg:
Okay, does anyone else want to pick this up then? I can help.

LinuxMercedes:
I'd be interested, although based on my other responsibilities I think it'll be about a month or so to put together a new RFC based on the comments here.

alexreg:
@LinuxMercedes That's no problem. I have several things keeping me busy too. Feel free to take this and create a new RFC, making a new PR whenever you're ready. I'm available for chat on Discord/Zulip in the meanwhile, if you like.

runiq:
@LinuxMercedes Hey, friendly reminder, no pressure: have you had the chance to look at this issue again? :)

LinuxMercedes:
hahahahaha no :frowning_woman: 

soon though!

TotalKrill:
I am a tad bit sad, that my venture into Newtype to avoid Orphan Rule led me here. It was a rollercoaster ride, where every solution, had a problem, with another solution, but sadly. This seems to be where it ended.

Progdrasil:
Same as @TotalKrill here, this would be especially useful for functions consuming NewTypes.

viliml:
> hahahahaha no 🙍‍♀️
> 
> soon though!

@LinuxMercedes has there been any progress?

clarfonthey:
Discussions on a new RFC should probably happen in a separate Internals or Zulip thread, since this thread is old and followed by a lot of people who probably don't want to be pinged by random messages directed at specific people.

PR #105, opened 2014-06-03
RFC: Add unboxed, abstract return types

aturon:

PR #2444, opened 2018-05-21
Add RFC undo-universal-impl-trait.

phaazon:
[Rendered](https://github.com/phaazon/rfcs/blob/undo-universal-impl-trait/text/0000-undo-universal-impl-trait.md)

### DIFF
--- a/text/0000-undo-universal-impl-trait.md
+++ b/text/0000-undo-universal-impl-trait.md
@@ -127,11 +127,11 @@ Now consider:
 fn add_self(x: impl Add<Output = impl Copy> + Copy)
 ```
 
-You can see the duplication here and the overwhelming confusion that both the `impl Trait` will
-resolve to the same type, even though they don’t have the same contract, which is impossible to
-guess from the interface while it could. This is legal because we only talk about two contracts here
-and the function will pick a type at the union (it must be `Add + Copy` and `Copy`) but you also
-have that weird `Output` contract as well.
+You can see the duplication here and the fact that both the `impl Trait` will resolve to the same
+type, even though they don’t have the same contract, which is impossible to guess from the interface
+while it could. This is legal because we only talk about two contracts here and the function will
+pick a type at the union (it must be `Add + Copy` and `Copy`) but you also have that weird `Output`
+contract as well.
 
 Even weirder:
 
--- a/text/0000-undo-universal-impl-trait.md
+++ b/text/0000-undo-universal-impl-trait.md
@@ -1,0 +1,238 @@
+- Feature Name: undo-universal-impl-trait
+- Start Date: 2018-05-21
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+This RFC undoes [RFC 1951], which gave Rust the `impl Trait` syntax in function argument position.
+
+# Motivation
+[motivation]: #motivation
+
+## Brief history
+
+`impl Trait` was initially introduced in [RFC 1522]. That feature was completely new to Rust as it
+added a very interesting and powerful feature: being able to provide the caller of a function with a
+type contract by providing *an anonymized type* in return location of the function and by letting
+the callee pick the type.
+
+    // the caller only knows _properties_ on the returned value, it doesn’t know its real type
+    fn zero_to_ten() -> impl Iterator<Item = u32>
+
+That feature was a complete success and was quickly adopted. The syntax, `impl Trait`, was then
+quickly refered as “the syntax to pick the type in the callee” and was very simple to get our
+fingers around. This is a use of *existential quantification* and the single one currently available
+in Rust. With the right education – and the Rust team made a wonderful job with the book – it’s easy
+to teach newcomers about the feature.
+
+Then came [RFC 1951] and [RFC 2071]. The latter generalizes the use of `impl Trait` to `let`
+bindings, `const` and `static` declarations and is not of our matter here – because they also act as
+*existentials*, which is great. However, the former, [RFC 1951], introduced `impl Trait` in argument
+position. There were a lot of debates about the semantics that the syntax should have but
+eventually, the consensus came to a decision. The following syntax:
+
+    fn foo(x: impl Debug);
+
+Is akin to this:
+
+    fn foo<T: Debug>(x: T)
+    fn foo<T>(x: T) where T: Debug
+
+And so the feature was stabilized.
+
+This RFC wants to undo this.
+
+## Why is `impl Trait` in argument position wrong?
+
+People out there didn’t really realize what was going on. Complains and confusions on IRC, reddit
+and the Rust forum started to appear, after the stabilization, by different kind of people –
+confusion among newcomers, advanced proficient Rustaceans, etc.. Here are some – non-exhaustive –
+links:
+
+  - https://internals.rust-lang.org/t/how-is-impl-display-different-from-foo-t-display-x-t-t/7566/3
+  - https://www.reddit.com/r/rust/comments/8jfn7z/what_is_the_advantage_of_impl_trait_in_argument/
+  - https://www.reddit.com/r/rust/comments/8io4q6/impl_trait_is_here_now_what/dyu5dlm/
+  - https://github.com/rust-lang/rust/issues/44721#issuecomment-330935175
+
+[RFC 1951] is organized with *arguments* that we’re basing the current document against.
+
+### Argument from learnability
+
+> Argument [here](https://github.com/rust-lang/rfcs/blob/master/text/1951-expand-impl-trait.md#argument-from-learnability)
+
+This argument explains that people who have an understanding of generics would try to use them in
+return position for anonymized types (existentials). This argument then stated it’s confusing
+because the only way to do this with current Rust is to use a complete different mechanism – being
+`impl Trait`. By contrast, if they have first learned `impl Trait` in argument position, trying to
+use it in return position *will just work as intended*.
+
+This argument is twisted along the lines of subjective opinion about learnability. Rust is a
+language with a fabulous set of learning resources (the first and most important one being
+[The Book]). It seems sound to state that **if** people are introduced first to `impl Trait` in
+argument position, it’s likely that they’ll understand the use in return location. However, what if
+they don’t want to express it? Existentials in return locations are already a very specific use case
+of a type system. Having a complete different mechanism is actually a feature, because we are
+talking about something that has no alternative elsewhere in Rust – i.e. pick the type in the
+callee. The initial assumption that having a different mechanism is confusing is wrong because
+`impl Trait` carries its own semantics and because anonymization in return type is not a trivial
+feature people will use without actually wanting it.
+
+Also, people are used to generics. People coming from C++, C#, Java etc. actually know about
+generics. The links listed above prove how confusing it can get to show them a complete, different
+way to use generics (`impl Trait` in argument position is a use of hidden generics, that is,
+anonymized type variables with explicit trait bounds). For people who don’t come from such
+languages or for completely newcomers, we are assuming that generics are too hard to understand at
+first but people will eventually need to learn them, as `impl Trait` in argument position is
+strictly less powerful than type variables and trait bounds. They will have to read the book –
+perhaps several times. It then confusing to be introduced a concept that is less powerful than
+generics and yet used in a complete, separate way that has no other alternative (return location).
+
+Finally, advanced Rustaceans are also quite confused, because they’ve been using `impl Trait` in
+return location for a while now (mostly on nightly) and the initial RFC was well accepted among the
+community and was a success. Those developers obviously mapped `impl Trait` to *callee-picked*
+types, because at that time it was its sole semantics. When a function with an argument annotated
+with `impl Trait` appears to such a developer, they get confused because that would mean *an argument
+which type is chosen by the callee*, which doesn’t quite make sense.
+
+### Argument from ergonomics
+
+> Argument [here](https://github.com/rust-lang/rfcs/blob/master/text/1951-expand-impl-trait.md#argument-from-ergonomics)
+
+In this argument, [RFC 1951] states that having `impl Trait` in argument position eases how our
+mental model about a function works. We don’t have to read the type variables and rembember their
+trait bounds anymore to understand the contract of a function.
+
+This argument seems sound but again is wrong. Consider:
+
+```
+fn foo<T>(a: T, b: T, c: impl Debug) where T: Add<Output = T> + Copy
+```
+
+This function has three type variables to substitute to be completely monomorphized. However, it’s
+very hard to see it at first because one of the variables is hidden. Worse, the ergonomics argument
+doesn’t stand here because you can see you **have** to read the type variables in order to get what
+you can do with the variables.
+
+Also, consider this function that takes an argument and add it to itself:
+
+```
+fn add_self<T>(x: T) -> T where T: Add<Output = T> + Copy
+```
+
+Now consider:
+
+```
+fn add_self(x: impl Add<Output = impl Copy> + Copy)
+```
+
+You can see the duplication here and the overwhelming confusion that both the `impl Trait` will
+resolve to the same type, even though they don’t have the same contract, which is impossible to
+guess from the interface while it could. This is legal because we only talk about two contracts here
+and the function will pick a type at the union (it must be `Add + Copy` and `Copy`) but you also
+have that weird `Output` contract as well.
+
+Even weirder:
+
+```
+fn add_self(x: impl Add<Output = impl Sized> + Copy) -> impl Sized {
+  x + x
+}
+```
+
+### Argument from familiarity
+
+> Argument [here](https://github.com/rust-lang/rfcs/blob/master/text/1951-expand-impl-trait.md#argument-from-familiarity)
+
+This argument is quite the same as the one for ergonomics and yet has this sentence:
+
+> In Rust, this is just a syntactic easement into a unitary polymorphism system which is
+> fundamentally one idea: parametric polymorphism with trait constraints.
+
+The example given in that statement is itself a proof that *unitary polymorphism* isn’t that common
+since the `U` type variable *needs* to be explicitely written. The familiarity with Java or C# might
+hold but shouldn’t drive the design decision and even more important, the consistency of the Rust
+syntax.
+
+### Summary of arguments
+
+This document has shown that `impl Trait` in argument position is confusing for newcomers and
+already advanced proficient Rustaceans. Among the issues:
+
+  - Two syntactic schemes to express parametric polymorphism (universal):
+    + `<T: Bound>` / `<T>` + `where: Bound`
+    + `impl Trait`
+  - `impl Trait` is strictly less powerful than type variables plus traits, so very limited.
+  - Even some examples requiring only one place type variable cannot be expressed with `impl Trait`
+    (`T: Add<Output = T>`) without a lot of confusion.
+  - It’s very hard to tell what are the variables to monomorphize a function, especially if you mix
+    both the style.
+  - Confusing because people are used to `impl Trait` as existential.
+
+This RFC was written with respectful discussions in mind. If something can be done, it can also be
+undone and this RFC strongly thinks it would be better for the community to undo [RFC 1951] for all
+the reasons listed above.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+  - For newcomers, nothing to teach since we’re removing a syntax construct.
+  - For others, the book must be updated with a deprecated section and explain why the feature was
+    removed if we reach a consensus.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+If though the syntax is removed, `rustc` must still be able to recognize it so that it can lead
+people to type variables and trait / where clauses.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The main drawback is that people are already starting to write Rust code with this new feature. It
+will then have an impact on the existing codebase. However, as shown in the current document, moving
+away from `impl Trait` in argument position is straight-forward and lossless.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+# Prior art
+[prior-art]: #prior-art
+
+Haskell is a good candidate to base this RFC against. Rust trait bounds can be seen as Haskell
+typeclass constraints in functions. For instance:
+
+```
+-- a function that takes a value which type is an instance of the Num typeclass (i.e. you can add,
+-- multiply, etc.) and that adds this value to itself
+addSelf :: (Num a) => a -> a
+addSelf a = a + a
+```
+
+The following:
+
+```
+addSelf :: Num a -> Num a
+```
+
+Would violate the type checker because a function expects *types* but the implementation uses
+*constraints*. The `ghc` error is:
+
+```
+…6:12: error:
+    • Expected a type, but ‘Num a’ has kind ‘Constraint’
+    • In the type signature: addSelf :: Num a -> Num a
+```
+
+Haskell lists the constraints an expression is limited with in the head part of its declaration (the
+part in between `()` in the above example). This makes it easy to see the contract of the function,
+as it’s easy to read the same in Rust with `where` clauses.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+[RFC 1522]: https://github.com/rust-lang/rfcs/blob/master/text/1522-conservative-impl-trait.md
+[RFC 1951]: https://github.com/rust-lang/rfcs/blob/master/text/2071-impl-trait-type-alias.md
+[RFC 2071]: https://github.com/rust-lang/rfcs/blob/master/text/2071-impl-trait-type-alias.md
+[The Book]: https://doc.rust-lang.org/book/second-edition/index.html
--- a/text/0000-undo-universal-impl-trait.md
+++ b/text/0000-undo-universal-impl-trait.md
@@ -110,7 +110,7 @@ This argument seems sound but again is wrong. Consider:
 fn foo<T>(a: T, b: T, c: impl Debug) where T: Add<Output = T> + Copy
 ```
 
-This function has three type variables to substitute to be completely monomorphized. However, it’s
+This function has two type variables to substitute to be completely monomorphized. However, it’s
 very hard to see it at first because one of the variables is hidden. Worse, the ergonomics argument
 doesn’t stand here because you can see you **have** to read the type variables in order to get what
 you can do with the variables.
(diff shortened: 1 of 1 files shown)


roblabla:
Aren't universal impl trait stable since 1.26 ? This would be a breaking change.

GuillaumeGomez:
I like the idea.

gmorenz:
> Also, consider this function that takes an argument and add it to itself:
>
> ```rust
> fn add_self<T>(x: T) -> T where T: Add<Output = T> + Copy
> ```
>
> Now consider:
>
> ```rust
> fn add_self(x: impl Add<Output = impl Copy> + Copy)
> ```
> 
> You can see the duplication here and the overwhelming confusion that both the `impl Trait` will
resolve to the same type, even though they don’t have the same contract, which is impossible to
guess from the interface while it could. This is legal because we only talk about two contracts here
and the function will pick a type at the union (it must be `Add + Copy` and `Copy`) but you also
have that weird `Output` contract as well.
>
> Even weirder:
>
> ```rust
> fn add_self(x: impl Add<Output = impl Sized> + Copy) -> impl Sized {
>  x + x
> }
> ```

The `impl trait` versions here are better code, since they give a better specification of what the function actually does. The equivalent non impl trait version to the best impl trait version would be

```rust
fn add_self<T: Add<Output = O> + Copy, O: Sized>(x: T) -> impl Sized {
  x + x
}
```

We can make a non-impl trait version that is even better, since we can specify that the `Add::Output` type is equal to the return type of the function

```rust
fn add_self<T: Add<Output = O> + Copy, O>(x: T) -> O {
  x + x
}
```

We can mix this with impl trait for what is my preferred version

```rust
fn add_self<O>(x: impl Add<Output=O> + Copy) -> O {
  x + x
}
```

What I get from this whole line of reasoning though is that universal impl trait encourages better code. It makes you realize what types need to be equal, and what ones don't.

Boscop:
I'm very much in favor of removing `impl Trait` for arg types:

1. `impl Trait` for arg types is non-orthogonal, which makes it harder to decide which way of universal quantification to use, and this question will come up **over and over and over** just because there are now 2 ways.
1. It's not actually easier to learn than generics. It looks very different from generics syntaxes that most programmers are familiar with.
1. It's easier for learning Rust to NOT conflate universal and existential quantification.
People should be aware of the difference, so that's another argument to keep only `<T: Trait>` for arg types and `impl Trait` only for existential quantification (including for other cases of existential quantification later when `impl Trait` will be allowed for const/let and existential types).
1. The argument that "having `impl Trait` for arg types allows postponing introducing generics until later in the book" doesn't count, because newbies will have to read the Rust book multiple times anyway. 
1. Most newbies will be familiar with another language that has similar generics syntax, e.g. C++, C#, Java, D, Scala etc.
1. There is **NO** evidence that `impl Trait` makes generics/Rust easier to learn, quite the contrary, it seems to confuse people, and:
1. From personal experience teaching Rust to complete newbies (no previous programming experience), and to some who only knew C or Python: 
Universally quantifying generics are not a difficult part of Rust (but references/lifetimes in generics). For anyone coming from any other language with universally quantifying generics they will quickly grasp the Rust syntax, but `impl Trait` is completely foreign to them.
1. The argument that `impl Trait` for arg types should be introduced to mirror `impl Trait` for return types is invalid: We Rust programmers lived without `impl Trait` for return types for years, it can be postponed until the end of the book, until newbies have understood universally quantifying generics! Newbies can write **a lot** of code before ever even needing `impl Trait` for return types.
1. There is no reason to ever choose to write `impl Trait` for arg types as a non-newbie, so why should newbies learn this way, if they have to unlearn it very quickly anyway?
1. I disagree with the [dialectical ratchet](https://github.com/rust-lang/rfcs/pull/2071#issuecomment-329026602) in this case. 
It may apply to references/lifetimes but **NOT** to `impl Trait` for arg types. 
The concepts of references and lifetimes complement each other, they are used together, not just by newbies. These are orthogonal concepts!
But having "`impl Trait` for args" as "the newbie-way to express universally quantified generics" means that newbies have to learn something that they have to unlearn later!
Which **WILL** confuse them because there are 2 overlapping ways to do the same thing and the one they learned is stricly less powerful, which demotivates them because they now have to unlearn it and learn the fully powerful way! And then they will ask themselves: "why did i even have to do the detour through "`impl Trait` for args", this `<T: Trait>` syntax is more familiar and more powerful"..
1. We should not compromise orthogonality of the language to introduce a strictly less powerful syntactic sugar, **ESPECIALLY** because it will only confuse people more.
1. The Rust language contains a lot of concepts, everyone will have to read the book a couple times. Making Rust easy to **use** and easy to **remember** is at least as important as making it easy to learn, and the bottleneck for learnability is the docs, not the language design itself. The docs about generics can always be improved without affecting Rust's design, we shouldn't compromise Rust for this! 
1. What makes a language easy to use and remember is the consistency and orthogonality of the concepts, so that there aren't overlapping concepts, and so that in every situation there is one obvious way to do things. Not 2 ways, where one is strictly less powerful. (It's like having traits + OOP inheritance, makes it harder to choose which one to use.)
1. In the future, we should **really** do some A/B testing to determine which change to Rust makes it actually easier to learn, before stabilizing that change. And when features are introduced that are controversial or unorthogonal, the debate thread should be more visible and open for longer. I only noticed that `impl Trait` was also being stabilized for args when it was too late to comment on it, because it was kind of hiding in the shadow of `impl Trait` for return types.

(Sorry if this sounds like a rant, I just feel very strongly about this :)

Pzixel:
When you write an RFC "Add feature X" they upvote you.
When featurer is added and you write an RFC "Remove feature X" they upvote you.

You think "Maybe I don't understand what's inside people minds"

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+[The Book]). It seems sound to state that **if** people are introduced first to `impl Trait` in
+argument position, it’s likely that they’ll understand the use in return location. However, what if
+they don’t want to express it? Existentials in return locations are already a very specific use case
+of a type system. Having a complete different mechanism is actually a feature, because we are
+talking about something that has no alternative elsewhere in Rust – i.e. pick the type in the
+callee. The initial assumption that having a different mechanism is confusing is wrong because
+`impl Trait` carries its own semantics and because anonymization in return type is not a trivial
+feature people will use without actually wanting it.
+
+Also, people are used to generics. People coming from C++, C#, Java etc. actually know about
```
sgrif: You've followed

> This argument is twisted along the lines of subjective opinion about learnability

With an equally subjective argument.

sgrif: People come to Rust from languages other than C++, C#, and Java. Not to mention that the entire point of `impl Trait` in argument position is that it works much closer to Java/C#. In java, you don't write void `<T extends SomeInterface> method(T arg)`, you write `void method(SomeInterface arg)`.

Restioson:
I completely agree with this PR. I didn't understand why we had two superfluous syntaxes for the same thing (although impl Trait is less useful as you can't turbofish it).

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+generics. The links listed above prove how confusing it can get to show them a complete, different
+way to use generics (`impl Trait` in argument position is a use of hidden generics, that is,
+anonymized type variables with explicit trait bounds). For people who don’t come from such
+languages or for completely newcomers, we are assuming that generics are too hard to understand at
+first but people will eventually need to learn them, as `impl Trait` in argument position is
```
sgrif: This is no different than lifetime elision. `fn foo(&self)` is strictly less powerful than `fn foo<'a>(&'a self)`.

logannc:
Does anybody actually think in terms of 'caller' vs 'callee' chosen types? Are we really confusing people aware of the type theory involved for more than a few moments?

`impl Trait` is an anonymized type known to the compiler to implement `Trait`. With that definition, the distinction between 'universal' vs. 'existential' goes away and, for symmetry, it makes sense to have it available in both argument and return positions.

vorot93:
What we see here is a huge failure of public review process. Simply put, universal-impl-trait escaped wide public review by coat racking onto conservative-impl-trait PR. We all waited and prepared for the latter and thus the former was not noticed before too late.

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+mental model about a function works. We don’t have to read the type variables and rembember their
+trait bounds anymore to understand the contract of a function.
+
+This argument seems sound but again is wrong. Consider:
+
+```
+fn foo<T>(a: T, b: T, c: impl Debug) where T: Add<Output = T> + Copy
+```
+
+This function has three type variables to substitute to be completely monomorphized. However, it’s
```
sgrif: I think you meant two type variables

Restioson:
@logannc Yes, we are confusing people who understand what the difference between universal impl Trait and existential impl Trait. I researched the topic extensively to understand it, and after a few hours I was still sort of uneasy as to why we had it do two perpendicular things. 

> impl Trait is an anonymized type known to the compiler to implement Trait. With that definition, the distinction between 'universal' vs. 'existential' goes away
No, this distinction does *not* go away. As a universal bound there are multiple types you can call the method on (i.e `fn x(a: impl T)`), but as an existential bound there is exactly one (i.e `fn x() -> impl T`). This makes it different for monomorphisation and for normal programming too -- you could expect `fn x() -> impl Iterator<Item=u32>` to return any given type that implements `Iterator<Item=u32>` (similar to how collect returns any type implementing `FromIterator`).

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+Also, consider this function that takes an argument and add it to itself:
+
+```
+fn add_self<T>(x: T) -> T where T: Add<Output = T> + Copy
+```
+
+Now consider:
+
+```
+fn add_self(x: impl Add<Output = impl Copy> + Copy)
```
sgrif: This signature is not equivalent. The only output type you could have here is `impl Copy` or `()`

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+You can see the duplication here and the overwhelming confusion that both the `impl Trait` will
```
sgrif: This argument is entirely subjective. I personally don't find this confusing. I'm taking "some type that can be added and copied. The result of the addition is some type that can be copied".

"overwhelming confusion" feels like unnecessary rhetoric to me.

text/0000-undo-universal-impl-trait.md
sgrif: Is the number of variables being monomorphised something people frequently care about? Is it a case we need to optimize for over readability?

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+> fundamentally one idea: parametric polymorphism with trait constraints.
+
+The example given in that statement is itself a proof that *unitary polymorphism* isn’t that common
+since the `U` type variable *needs* to be explicitely written. The familiarity with Java or C# might
+hold but shouldn’t drive the design decision and even more important, the consistency of the Rust
+syntax.
+
+### Summary of arguments
+
+This document has shown that `impl Trait` in argument position is confusing for newcomers and
```
sgrif: This document hasn't really shown *anything* from newcomers. It's pretty hard for us to gauge how this affects newcomers at this point, since it does not appear in the book or anywhere in the documentation.

Instead this document seems to be claiming every other argument is subjective, but this one is somehow objective.

Perhaps the source of confusion is the fact that it's a brand new, undocumented feature that people are still just starting to use.

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,238 @@
...
+    fn foo<T: Debug>(x: T)
+    fn foo<T>(x: T) where T: Debug
+
+And so the feature was stabilized.
+
+This RFC wants to undo this.
+
+## Why is `impl Trait` in argument position wrong?
+
+People out there didn’t really realize what was going on. Complains and confusions on IRC, reddit
```
sgrif: This feels like a straw man to me. The feature is not yet documented. There is always going to be confusion until that gets remedied. There's also a lot of confusion around how reborrowing works (another undocumented feature), and that feature's been around for much longer.

text/0000-undo-universal-impl-trait.md
Restioson: How so?

gmorenz: `fn foo(&self, x:  &str) -> &str` vs `fn foo<'a, 'b>(&'a self, x: &'b str) -> &'b str`. You can use the second in more places, e.g. with `'a=lifetime of some object` and `'b = 'static` (or just the lifetime of some longer lived object).

text/0000-undo-universal-impl-trait.md
```diff
@@ -0,0 +1,309 @@
...
+    have to learn**, should it be through the book or any other tutorials. Making Rust easy to
+    **use** and easy to **remember** is as important as making it easy to learn and making so is via
+    the documentation, tutorials and the book, not via the language design itself. We can improve
+    the documentation about generics without affecting the current design.
+  - Overlapping concepts harm the consistency and orthoganality of a language. Having several ways
+    to do something forces people to know all the ways – even if they always prefer using one –
+    because they will maintain and contribute to codebases they haven’t read nor written code for
+    before.
+  - Turbofishing is impossible with `impl Trait`, forcing you to use type ascription in other
+    places.
```
petrochenkov: This is fixable.
Technically, nothing prevents supporting turbofish and making arg position `impl Trait` purely a sugar for named type parameters.

text/0000-undo-universal-impl-trait.md
phaazon: Interfaces in Java are not akin to universal quantification, because those are dynamic dispatched variables. The equivalent in Rust would be `&Trait` or `dyn Trait`.

xgalaxy:
Excuse the stupid question.

How is impl Trait in argument position any different than say.. C# where a function or method takes an interface instead of a concrete type? Can someone explain that to me. Because in C# I don't feel like its confusing at all.

text/0000-undo-universal-impl-trait.md
phaazon: You’re right, it’s too much subjective. I’m removing it.

text/0000-undo-universal-impl-trait.md
phaazon: I’ve linked a few links in the top of the document. Also, reactions and the fact it’s been so opinionated is somehow a hint that there’s a massive confusion, right?

phaazon:
I added a bunch of comments from here. Thank you all for such massive participation.

sgrif:
> How is impl Trait in argument position any different than say.. C# where a function or method takes an interface instead of a concrete type?

In terms of program behavior it's equivalent to passing an interface in C#. Arguably the closer equivalent would be taking a "trait object", but this is something that only matters if you care about very specific performance trade offs that you don't normally need to care about.

> I'm still learning Rust but when I saw impl Trait in argument position I just equated it to doing the same thing I would do in C# with interfaces.

That was the goal behind the feature.

text/0000-undo-universal-impl-trait.md
Restioson: It's slightly more arcane, though. Often when doing something like `fn x<A: TraitA, B: TraitB<A>>(b: B)` I write it with the thing the 'main' generic depends on first (I don't know if you are allowed to put it second, but either way it feels 'wrong' to do it in reverse). This might not be so with impl Trait turbofishing, or at any rate is more difficult to figure out.

phaazon:
@xgalaxy to me, it’s different because of *dynamic dispatch* vs. static dispatch. The Rust equivalent would be:

    fn foo(x: &Debug)
    fn foo(x: dyn Debug) // with the new syntax

bowbahdoe:
Perhaps this is the wrong time and place to ask this, but what is turbofishing and why does it matter that it is impossible with `impl Trait`?

text/0000-undo-universal-impl-trait.md
Restioson: I mean, personally I was confused as to why we had the same syntax for the same thing, but that's just me I guess.

phaazon:
@bowbahdoe turbofishing is a way to force monomorphization of a function by providing it with the type variables substituted. You already have done it for sure. See:

```
let x = (0..10).collect::<Vec<_>>(); // the syntax ::<_>() is turbofishing
```

text/0000-undo-universal-impl-trait.md
phaazon: It’s not just you. All of this is very confusing and opinionated.

text/0000-undo-universal-impl-trait.md
sgrif: Yup, but the entire point is that you generally don't care about the tradeoffs between a monomorphised function and a dynamically dispatched one. Similarly, you generally don't need to care about the differences between returning `Box<Foo>` and `impl Foo`. This is a "do what I mean" feature. Similar to lifetime elision, it hides details that you usually don't care about. If you end up in a situation where you do need to care about it (`T: Add<Output = T>` is a good example), or want to care about it (you're in a situation where binary size is very important), then you learn the underlying mechanisms.

text/0000-undo-universal-impl-trait.md
sgrif: > Also, reactions and the fact it’s been so opinionated is somehow a hint that there’s a massive confusion, right?

People having strong opinions does not mean they're confused.

vorot93:
@sgrif Are you referring to `impl Trait` in general or `universal-impl-trait` in particular. We all discussed `conservative-impl-trait`, but `impl Trait` in argument position? Not so much.

And being available in the nightly means nothing. There are a lot of experimental and incomplete features that can be turned on via feature gate and may go away at any time. Just look at non-existant trait aliases for example.

sgrif:
@phaazon There's an important detail that you're leaving out, which is that turbofishing is almost never required for type variables used in an argument position. The compiler can infer it from the type of the argument you're passing. The *only* time I've ever needed to turbofish an argument parameter is when passing a generic `extern "C"` function pointer.

In fact, I hope the way this ends up getting resolved is to allow type parameters to be explicitly passed, but no way to explicitly state the type of `impl Trait`, as it's a strength of the feature. Take for example this function in Diesel:

```
fn debug_query<DB, T>(query: &T) -> DebugQuery<T, DB>
where
    DB: Backend,
    T: QueryFragment<DB>,
```

The type of `DB` can never be inferred. The type of `T` is always inferred. 100% of the usage of this function looks like `debug_query::<Pg, _>(&query)`. Instead we could write it as:

```
fn debug_query<DB: Backend>(query: &impl QueryFragment<DB>) -> DebugQuery<impl QueryFragment<DB>, DB>
```

which would remove the pointless `, _` from the caller.

text/0000-undo-universal-impl-trait.md
phaazon: > You generally

Are you sure about that? :)

sgrif:
Another example from Diesel that would be improved by `impl Trait` in argument position:

```
fn transaction<T, E, F>(&self, f: F) -> Result<T, E>
where
    F: FnOnce() -> Result<T, E>,
    E: From<diesel::Error>,
```

in typical usage, `T` and `F` are inferred, and `F` is a closure (so cannot be named). `E` however, is usually explicitly given, since the body of the closure will likely use `?`. So it's normally invoked as `conn.transaction::<_, diesel::Error, _>(|| stuff()?; Ok(()))`

This case is a little more interesting, since we can remove the `F` parameter, but we can't remove the `T` (unless `FnOnce` were stabilized in a form where the arguments are an associated type). However, we could change the signature to:

```
fn transaction<T, E>(&self, f: impl FnOnce() -> Result<T, E>) -> Result<T, E>
```

which would change the invocation to `conn.transaction::<_, diesel::Error>(|| stuff()?; Ok(()))`

sgrif:
@Pzixel that is incorrect.

Phlosioneer:
@sgrif Yes, but how do you teach that to a newbie? The difference here goes back to the caller vs callee distinction. You need to know about that distinction to be able to understand why those two functions are different.

Pzixel:
@sgrif 
> that is incorrect.

Well, could you elaborate a while then, please? IIRC each input `impl Trait` becomes a generic parameter. If you say that `output` parameter would be tied to the `input` one then how compiler could figure out if they are tied? How could I express then that they are different? This one shows that I'm right

```rust
fn foo(value: &impl std::fmt::Debug) -> impl std::fmt::Debug {
    42
}

fn main() {
    foo(&"42");
}
```

sgrif:
@Pzixel It's equivalent to `fn debug_query<DB, T>(query: &T) -> DebugQuery<opaque type you don't know anything about and don't choose, DB>`. Whether or not the return type is the same as the input type is unimportant to this API. If I wanted to explicitly state that they were the same type, I'd give it a name and make that clear (e.g. make it a type parameter)

Phlosioneer:
@Pzixel only the input `impl Trait`s can be replaced by named type parameters. Output ones are not expressible by the named-type-parameter typesystem, because the _function_ decides their type based on the inputs, rather than the _caller of the function_ declaring a type for `U`.

In your example:
```
fn foo(value: &impl std::fmt::Debug) -> impl std::fmt::Debug {
    42
}

fn main() {
    let dbg1 = foo(&"42");
    let dbg2 = foo(Some(3));
    let debugs = vec![dbg1, dbg2];
    println!("all debugs: {:#?}", debugs);
}
```
This works with type parameters (if you declare the returning type to be u8 or something like that, to resolve the ambiguous integer type). It does NOT work with impl Trait - at no point does the function guarantee it returns the _same_ output type for those two _different_ input types.

sgrif:
@Phlosioneer I think you're over-emphasizing the importance of knowing/caring about callee chosen vs caller chosen. A better comparison to make would be the difference between `fn foo(x: &Display) -> &Display` and `fn foo(x: impl Display) -> impl Display`, as they're much closer semantically.

Boscop:
@sgrif IIRC, in the D language, turbofish allows omitting type params (from the end), they don't have to be substituted by `_`. If we had this in Rust, you wouldn't need `impl Trait` for arg types to satisfy your use case :)

sgrif:
@Boscop Yup, that's a direction we could go as well. As with all things, it's a tradeoff.

Pzixel:
@sgrif 
> It's equivalent to fn debug_query<DB, T>(query: &T) -> DebugQuery<opaque type you don't know anything about and don't choose, DB>. Whether or not the return type is the same as the input type is unimportant to this API. If I wanted to explicitly state that they were the same type, I'd give it a name and make that clear (e.g. make it a type parameter)

Yes, you're right, thank you.

However, I still think that it leads to confusion. There should be only exact one way to do some thing.

The main reason against `impl Trait` in input parameter is that it procides hard-to-spot monoporphisation.  As you pointed out, `fn foo(x: &Display) -> &Display` and `fn foo(x: &impl Display) -> impl Display` look similar a lot, while they are different. Some generic salt is a good here, imho. I can easily spot that  `fn foo<T: Display(x: &T) -> impl Display` will have a static dispatch, and I still can write `fn foo<T: Display(x: &T) -> T` if I want to express relationship between input and output types.

text/0000-undo-universal-impl-trait.md
Boscop: @sgrif 
Wouldn't the actual equivalent of 
```java
void foo(SomeInterface arg)
``` 
be 
```rust
fn foo(arg: Rc<RefCell<dyn SomeTrait>>)
```
?

Because the instance is GC'ed and mutable.

So when translating Java code (or thinking) to Rust, one wouldn't just be able to write `fn foo(arg: impl SomeTrait)` instead of the java method anyway, because that would be different semantics (it takes ownership and can't mutate). One would have to re-architect the code in the translation process to use references/lifetimes and to please the borrow checker anyway. 
So `impl Trait` for args isn't really the equivalent (or what one wants) in most cases.

sgrif:
> However, I still think that it leads to confusion. There should be only exact one way to do some thing.

Are you of the opinion that lifetime elision should also be removed? These two features have similar uses and drawbacks. You also can't provide a lifetime explicitly via turbofish when lifetime elision is being used. Sometimes you need to explicitly name them (either to show equivalence or some relationship between multiple lifetimes), but it's pretty rare. Same for `impl Trait`. Often you don't need to give a type parameter a name. It's probably more common to need to name a type (either to show equivalence or some relationship between multiple type parameters), but it's still somewhat rare.

I think it's important that we have a way to drop down to a lower level, either because you need to use a turbofish (extremely rare for argument parameters) or because you need to name a type. But this feature gives us a great parity between `fn foo(x: &Trait) -> &Trait` and `fn foo(x: impl Trait) -> impl Trait`.

Pzixel:
> Are you of the opinion that lifetime elision should also be removed? These two features have similar uses and drawbacks. You also can't provide a lifetime explicitly via turbofish when lifetime elision is being used. Sometimes you need to explicitly name them (either to show equivalence or some relationship between multiple lifetimes), but it's pretty rare. Same for impl Trait. Often you don't need to give a type parameter a name. It's probably more common to need to name a type (either to show equivalence or some relationship between multiple type parameters), but it's still somewhat rare.

No, lifetime elision is fine, because it's the only way to specify object lifetimes. There is no "easier but more fragile syntax" that allows you to write the very same thing but using a bit less characters.

I love `impl Trait`, but Rust already have a very nice and powerful generics. Having single way to write "argument of type T" have much higher value than possibility to sometimes inline type declarations. This is why I love `impl Trait` in return position: it's not possible to express is via existing language features. So it's a strict language expansion, while `universal-impl-trait` is substitution.

aturon:
We are going to take the unusual step of immediately closing and locking this RFC.

While there are clearly a lot of feelings on all sides of the issue, the fact is that this feature has shipped on the stable channel and is not going to be removed. This is both because of our basic stability guarantees but also because the decision was made in good faith and pursuant to our process, and we stand by that. 

After multiple years of RFC and tracking issue discussions (the first one was [RFC #105](https://github.com/rust-lang/rfcs/pull/105) in 2014!), the Rust Language Design Team ultimately reached a decision to ship this feature. This decision was not reached lightly. We discussed in depth a number of alternatives, including limiting `impl Trait` to return position as well as using an alternate syntax. Each has its pros and cons, and ultimately a judgement call had to be made. This is the nature of the language design process — indeed, any decision process. Few decisions are clear cut, which is why our process includes a number of points where feedback can be given, including a number of final-comment-period advertisements and the like. As a community, we have made a deliberate choice to slow down development to ensure thorough vetting and input into the process.

To be clear: we understand that there are downsides to this feature, and that some people find those downsides concerning. All of us care deeply about Rust, and it can be distressing to see people in power moving things in a direction you dislike. But, at the end of the day, we have to be able to make — and stick with — decisions, striking a balance between long-running feedback and shipping. Rust 2018 will ship with `impl Trait` in argument position.

-- @nikomatsakis and @aturon, on behalf of the Rust Language and Core Teams

PR #2487, opened 2018-06-23
RFC: auto-bound.

phaazon:
[Rendered link :P](https://github.com/phaazon/rfcs/blob/auto-bound-ergonomics-initiative/text/0000-no-bound-ergonomics.md)

### DIFF
--- a/text/0000-no-bound-ergonomics.md
+++ b/text/0000-no-bound-ergonomics.md
@@ -1,0 +1,99 @@
+- Feature Name: auto-bound
+- Start Date: 2018-06-24
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+This document adds a new ergonomics syntax to Rust to express type variables which *bounds* or
+*bounds combination* – are infered from the implementation – in order to *mirror* the already
+existing `impl Trait` in return position that infers the type on the implementation.
+
+```rust
+fn foo(x: &impl _) { // x’s bounds will be deduced from the implementation
+  let y = x.clone(); // we want Clone here
+}
+```
+
+In the case of an ambiguity, the bound can be added the regular way:
+
+```rust
+fn foo(x: &impl _ + SomeTrait) { // x bounds will be deduced from the implementation
+  let y = x.clone(); // we want Clone here
+  x.some_trait_method();
+}
+```
+
+# Motivation
+[motivation]: #motivation
+
+The main motivation is that we want to have more people coming from C++ and Java / JavaScript easily
+get their feet wet with Rust. A new initiative has started to do this: the *ergonomics initiative*.
+The idea is that the community really likes the *safety* of the language, its low *overhead* and its
+*explicitness*, but it also seems to love *ergonomics* and *implicit* features to catch and factor
+away typical patterns.
+
+C++ programmers are used to what is called [duck typing](https://en.wikipedia.org/wiki/Duck_typing),
+a *very good* idea introduced decades ago. This feature enables them to write something like this:
+
+```cpp
+template <T> // T can be anything
+void add(T a, T b) {
+  return a + b;
+}
+```
+
+This code compiles and will adapt to any kind of input’s type.
+
+> We took a C++ example, but this is also the case in JavaScript, for instance.
+
+This is called duck typing and it’s very likely that programmers used to that will want to use the
+same feature in Rust – and we also want that very much, right!
+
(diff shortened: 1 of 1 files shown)


text/0000-no-bound-ergonomics.md
```diff
@@ -0,0 +1,99 @@
...
+# Summary
+[summary]: #summary
+
+This document adds a new ergonomics syntax to Rust to express type variables which *bounds* or
+*bounds combination* – are infered from the implementation – in order to *mirror* the already
+existing `impl Trait` in return position that infers the type on the implementation.
+
+```rust
+fn foo(x: &impl _) { // x’s bounds will be deduced from the implementation
+  let y = x.clone(); // we want Clone here
```
sfackler: What happens if there exists a second trait that has a method called `clone`? Why wouldn't that infer no bounds on the type at all, since `&T` is already Clone for any T?

text/0000-no-bound-ergonomics.md
```diff
@@ -0,0 +1,99 @@
...
+fn foo(x: &impl _ + SomeTrait) { // x bounds will be deduced from the implementation
+  let y = x.clone(); // we want Clone here
+  x.some_trait_method();
+}
+```
+
+# Motivation
+[motivation]: #motivation
+
+The main motivation is that we want to have more people coming from C++ and Java / JavaScript easily
```
sfackler: Java?

text/0000-no-bound-ergonomics.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Summary
+[summary]: #summary
+
+This document adds a new ergonomics syntax to Rust to express type variables which *bounds* or
+*bounds combination* – are infered from the implementation – in order to *mirror* the already
+existing `impl Trait` in return position that infers the type on the implementation.
+
+```rust
+fn foo(x: &impl _) { // x’s bounds will be deduced from the implementation
```
sfackler: How is someone using this API supposed to know what they can pass to it?

text/0000-no-bound-ergonomics.md
```diff
@@ -0,0 +1,99 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+This design interacts with the `impl Trait` in argument position and the bound mechanism. `rustc`
+will be more permissive on which types it accepts by completely removing bounds, enabling people to
+*just use what they are used to from other languages like C++ and Javascript*.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This RFC *might* induce *confusion* and some people might have to *yell* because they don’t like
```
sfackler: This paragraph reads as being deeply sarcastic.

text/0000-no-bound-ergonomics.md
Centril: From the source code itself it is of course impossible. Outside of that, there are two principle ways that come to mind for knowing what they can pass in:

+ By reading the rustdoc, which can specify the actual bounds.
+ With IDE + RLS support which can tell you what the actual bounds are.

I'm reserving judgement on whether this is sufficient or a good idea for Rust in general.
I mostly use similar features in Haskell as a useful rapid prototyping development aid,
but I always specify the types after I've asked `ghci` what the types are.

Nonetheless, we don't have a proper REPL *yet*; so I don't know how valuable this would be until then.

text/0000-no-bound-ergonomics.md
```diff
@@ -0,0 +1,99 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC adds the new syntax `impl _`. Several good points:
+
+  - No new keyword added, so no breakage.
+  - The feature has no known overlapping syntax, so we can add the feature as a minor, non-breaking
+    change.
+  - It uses two keywords used everywhere in Rust: `impl` and `_`. The syntax even looks like the
+    *awesome* `impl Trait` in argument position (universal). Replacing the `Trait` with `_` is akin
```
Centril: You previously, https://github.com/rust-lang/rfcs/pull/2444, argued that `arg: impl Trait` was a bad idea and should have been undone. I somehow doubt you've suddenly changed your mind that much. I therefore think this RFC is insincere.

Centril:
Since this RFC was opened insincerely, per statements in `#rust-fr` on IRC,
I'm going to save everyone the time, energy and confusion, by closing this PR outright.

PR #2493, opened 2018-07-02
Rename 2071-impl-trait-type-alias.md to 2071-impl-trait-existential-types.md

jethrogb:
Renaming to match feature name


Centril:
I prefer not to do this as there may be links in the wild to the document which would 404. If you leave a file in place with a link to the new file address then I would be more positively inclined.

jethrogb:
Did that, but it looks like GitHub doesn't support symlinks very well.

Centril:
Try a file with an actual hyperlink instead :)
It takes the reader one more click but at least it will work.

Centril:
Works for me :) Thanks!

Centril:
Works for me; Thanks!

PR #2510, opened 2018-08-01
update ManuallyDrop RFC to reflect how the implementation now looks

RalfJung:
This reflects https://github.com/rust-lang/rust/pull/52711. I have no idea what the correct process is here.

### DIFF
--- a/text/1860-manually-drop.md
+++ b/text/1860-manually-drop.md
@@ -93,14 +93,12 @@ impl<T> Deref for ManuallyDrop<T> {
 impl<T> DerefMut for ManuallyDrop<T> {
     // ...
 }
-
-// Other common impls such as `Debug for T: Debug`.
 ```
 
 The lang item will be treated specially by the compiler to not emit any drop
 glue for this type.
 
-Let us apply this union to a somewhat expanded example from the motivation:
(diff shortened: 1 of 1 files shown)


text/1860-manually-drop.md
```diff
@@ -96,6 +97,9 @@ impl<T> DerefMut for ManuallyDrop<T> {
 // Other common impls such as `Debug for T: Debug`.
```
eddyb: This is unnecessary now.

text/1860-manually-drop.md
```diff
@@ -96,6 +97,9 @@ impl<T> DerefMut for ManuallyDrop<T> {
...
 ```
 
+The lang item will be treated specially by the compiler to not emit any drop
+glue for this type.
+
 Let us apply this union to a somewhat expanded example from the motivation:
```
eddyb: Still says `union` - you probably want to search the whole file.

RalfJung:
Fixed nits.

cramertj:
We don't really have a process for updating RFCs, as they're generally meant to express design and intent rather than serve as a reference. That said, it seems fine to just fix this to more correctly describe a (stable) feature the language.

PR #2531, opened 2018-08-26
Create 2505-foreach-style.md

ccgauche:


Centril:
Hi there. Thank you for your interest in improving Rust!

Unfortunately, this idea does not seem fully formed per our [RFC template](https://github.com/rust-lang/rfcs/blob/master/0000-template.md) so I'm going to go ahead and close this. If you would like to brainstorm your ideas or need help with formulating your proposal, you can do that more effectively at our [internals forum](https://internals.rust-lang.org/).

With respect to your ideas, it seems to me like it is already possible to do this.
For example, we may write ([playground](https://play.rust-lang.org/?gist=ff645742ff34feef907c615cceff1641&version=stable&mode=debug&edition=2015)):

```rust
fn main() {
    let list = vec!["hello", "world.", "Nice", "to", "meet", "you!"];

    for x in list {
        println!("word: {}", x);
    }
}
```

If you also need the index, you can write this with ([playground](https://play.rust-lang.org/?gist=8d9b5c4f326fc7848643502e1ded89c0&version=stable&mode=debug&edition=2015)):

```rust
fn main() {
    let list = vec!["hello", "world.", "Nice", "to", "meet", "you!"];

    for (i, x) in list.iter().enumerate() {
        println!("word #{}: {}", i, x);
    }
}
```

As for iterating over a HashMap, you can write (see [HashMap::iter](https://doc.rust-lang.org/nightly/std/collections/struct.HashMap.html#method.iter)):

```rust
use std::collections::HashMap;

let mut map = HashMap::new();
map.insert("a", 1);
map.insert("b", 2);
map.insert("c", 3);

for (key, val) in map.iter() {
    println!("key: {} val: {}", key, val);
}
```

ccgauche:
Sorry it's my first RFC. 
Best regards ccgauche.

PR #2579, opened 2018-10-27
RFC: DynTrait and VTable

SimonSapin:
> Add a `DynTrait` trait and a `VTable` type to provide access to the components of a fat pointer to a trait object `dyn SomeTrait`, and the ability to reconstruct such a pointer from its components.

[HTML view](https://github.com/SimonSapin/rfcs/blob/0bb757ee33001b26f7ad122d5228b2f43a41c04e/text/0000-vtable.md)

This RFC is based on @Gankro’s proposal at https://gist.github.com/Gankro/b053cb4d1cb3bcaec070de89734720f7. By the time I got it into a submittable state however, I because increasingly convinced that it would be better replaced by a more general design that would apply not just to trait objects but all pointers, with an associated type for (fat) pointer metadata. I’m submitting this as a PR anyway for posterity and comparison, but I’ll be closing it immediately.

### DIFF
--- a/text/0000-vtable.md
+++ b/text/0000-vtable.md
@@ -1,0 +1,308 @@
+- Feature Name: `vtable`
+- Start Date: 2018-10-24
+- RFC PR:
+- Rust Issue:
+
+# Summary
+[summary]: #summary
+
+Add a `DynTrait` trait and a `VTable` type to provide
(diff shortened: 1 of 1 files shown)


SimonSapin:
Closing in favor of https://github.com/rust-lang/rfcs/pull/2580.

PR #2624, opened 2019-01-13
Amend RFC 1242 to specify what “minimal maintenance” means for r-l-deprecated crates

jethrogb:
[Rendered](https://github.com/jethrogb/rfcs/blob/deprecation-maintenance/text/0000-deprecation-maintenance.md)

### DIFF
--- a/text/0000-deprecation-maintenance.md
+++ b/text/0000-deprecation-maintenance.md
@@ -1,0 +1,200 @@
+- Feature Name: N/A
+- Start Date: 2019-01-13
+- RFC PR: (leave this empty)
+- Rust Issue: N/A
+
+# Summary
+[summary]: #summary
+
+Amend [RFC 1242] to specify what “minimal maintenance” means for [rust-lang-deprecated] crates.
+
+[RFC 1242]: https://github.com/rust-lang/rfcs/blob/master/text/1242-rust-lang-crates.md
+[rust-lang-deprecated]: https://github.com/rust-lang-deprecated/
+
+# Motivation
+[motivation]: #motivation
+
+From time to time, official rust-lang crates may be deprecated according to [RFC 1242]:
+
+> At some point a library may become stale -- either because it failed to make it out of the nursery, or else because it was supplanted by a superior library.
+
+Especially in the latter case, use of the original crate may still be pervasive throughout the crates.io ecosystem. This evidenced by the crates.io download numbers (for lack of a better popularity metric):
+
+| Deprecated crate | Recent downloads | All-time downloads |
+| ---------------- | ---------------- |------------------- |
+| time             |             789k |             5,346k |
+| rustc-serialize  |             391k |             5,558k |
+| tempdir          |             280k |             2,616k |
+| hexfloat         |              258 |              3,027 |
+
+(numbers as of 2019-01-13)
+
+The community moving away from recently-deprecated crates will take time.
+
+Generally, dependencies on deprecated crates are not consciously added by crate authors. The deprecation messaging is well-recieved and new functionality is generally implemented with non-deprecated alternatives. However, crates can easily depend on deprecated crates *transitively*. To further drive this point home, `time` is used by such popular crates as `hyper`, `chrono`, `cookie`. `rustc-serialize` by `rust-crypto` and the RLS. `tempdir` by `cc` and `semver`.
+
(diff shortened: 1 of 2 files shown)


KodrAus:
Having a better understanding of what it means to deprecate a nursery crate seems like a good goal to me!

The main concern I have right now is I don't think this RFC sufficiently acknowledges our real lack of bandwidth for some nursery crates, let alone those in the deprecated organisation. I think bandwidth is the main reason some of these libraries receive no maintenance. Reviewing, merging, and publishing even small PRs in a repository that you aren't immersed in has a lot of peripheral context building effort in addition to that needed to review, merge, and publish. Having a policy for maintenance doesn't create bandwidth for what is, unfortunately, a bit of a thankless task.

On the other hand though, if there are folks interested in supporting these libraries then having a framework for helping them decide what level of contribution to consider might help them onboard.

jethrogb:
In a sense, I think the question of who should be performing the maintenance is orthogonal to this RFC. The status quo already calls for maintenance. This RFC just specifies what's included in that. If there are currently deprecated crates without maintainers, that situation should be resolved by the crate owner/libs team/community ASAP.

Perhaps you are implicitly advocating for *no* maintenance. As mentioned in the drawbacks/alternatives section of this RFC, I don't think that's a defensible position.

jethrogb:
> Reviewing, merging, and publishing even small PRs in a repository that you aren't immersed in has a lot of peripheral context building effort in addition to that needed to review, merge, and publish.

I don't agree with the “peripheral context building effort” part. Looking at the most recent merges of time and rustc-serialize, these could've been easily reviewed by anyone familiar with Rust but not at all familiar with the crate itself.

KodrAus:
> In a sense, I think the question of who should be performing the maintenance is orthogonal to this RFC.

I get the sentiment here, and while it doesn't seem unreasonable I think reality undermines the goals of this RFC. So I have to disagree that the question of who should perform the maintenance is orthogonal. It's fundamental. Without addressing that root problem we're really just elaborating on a process that doesn't actually exist outside the text in this repository.

That being said, I'm not opposed to this, I just think we need to understand the picture it fits into better, manage our expectations around how maintenance on deprecated repositories is likely to be performed, and look at how changes/clarifications to the deprecation process could result in an increase in bandwidth for these kinds of tasks.

Ericson2314:
We should really look at the dependency graph. Maybe there's just a few popularish crates still using these that bump up the download counts.

varkor:
I think setting out explicit guidelines as to how much maintenance a crate will receive is valuable. Declaring a crate to be "unmaintained", even if that's an undesirable stance, is better than claiming it's "minimally maintained" at the moment, when even minimal pull requests are not accepted.

If it's actually not enough bandwidth to maintain these crates, it's better to be up front about this. Part of the frustration with the status quo is setting incorrect expectations.

(To be clear, I'm not advocating one alternative or the other here; just that being clear about what deprecation *does* entail is one of the most important factors.)

text/0000-deprecation-maintenance.md
```diff
@@ -0,0 +1,200 @@
...
+[2018 Roadmap]: https://blog.rust-lang.org/2018/03/12/roadmap.html
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+In evaluating a PR for a deprecated crate meeting the popularity threshold, the reviewer could ask the following questions:
+
+* Does this PR fix a bug in the crate? Alternatively, are these the minimal changes required to make the crate function properly on a new platform?
+
+* Does this PR keep the public API of the crate the same? (This is a bellwhether for changes that are not bug fixes).
```
dwijnand: Cool word!  But I think it's misspelt.

```suggestion
* Does this PR keep the public API of the crate the same? (This is a bellwether for changes that are not bug fixes).
```

gnzlbg:
>  If there are currently deprecated crates without maintainers, that situation should be resolved by the crate owner/libs team/community ASAP.

@jethrogb How ? Finding maintainers even for non-deprecated crates is hard. The crate owner in this case would be the `rust-lang` org, and beyond asking whether someone wants to maintain a particular crate, there is not much that can be done AFAICT.

jethrogb:
@gnzlbg I'm not convinced that this is going to be an issue. If someone wants something to change about a crate without a maintainer, they are incentivized to find a maintainer. It could be that same person as well.

PR #2646, opened 2019-02-26
Add more examples for | patterns in let in RFC 2175

Manishearth:
This introduces no semantic changes to the RFC, and merely clarifies some things.

I wanted to better understand how `let x | y` was supposed to work and found the RFC rather confusing, especially since it doesn't have much of a guide-level section.

Added some examples to clarify things.

As `let x | y` isn't implemented yet this is the only place to find documentation about this proposed feature, so we should ensure it's good

r? @Centril.


Centril:
Thanks!

PR #2670, opened 2019-03-30
Bump mdbook: 0.1 -> 0.2.

frewsxcv:
Confirmed locally that 0.2 builds the book fine.


steveklabnik:
The major thing that changed was inter-page links, but we don't really use those here, so seems good. Thanks!

PR #2677, opened 2019-04-04
Add the Close trait

czipperz:
[Rendered](https://github.com/czipperz/rfcs/blob/close-trait/text/0000-close-trait.md)

### DIFF
--- a/text/0000-close-trait.md
+++ b/text/0000-close-trait.md
@@ -203,7 +203,15 @@ explicitly importing `Close`.
 
 We could implement `Close` for all `Drop` types by always succeeding.  Since
 this change can be done retroactively later without breaking backwards
-compatibility, I choose to leave it out of this RFC.
+compatibility, I choose to leave it out of this RFC.  Example implementation is
+included in case this is desirable in the future:
+
+    default impl<T> Close for T {
+        type Error = !;
+        fn close(self) -> Result<(), Self::Error> {
+            Ok(())
+        }
+    }
 
 ### Named `TryDrop` instead
 
--- a/text/0000-close-trait.md
+++ b/text/0000-close-trait.md
@@ -185,6 +185,29 @@ We could implement `Close` for all `Drop` types by always succeeding.  Since
 this change can be done retroactively later without breaking backwards
 compatibility, I choose to leave it out of this RFC.
 
+### Named `TryDrop` instead
+
+We could name this trait `TryDrop` instead, and the method `try_drop`.  This
+seems well and good as `Close` allows us to generalize the idea of `Drop`
+similarly to `TryFrom` and `TryInto`.  I do not prefer this so long as `close`
+takes `self` by value because I do not think it is intuitive to have different
+input parameters for semantically similar functions.
+
+    fn close(self) -> Result<(), Error>;
+    fn drop(&mut self) -> ();
+
+### Taking `self` by `&mut`
+
+`Close` has the signature `fn close(self) -> Result<(), Error>` whereas `Drop`
+has `fn drop(&mut self) -> ()`.  These signatures aren't identical, but they
+could be made to be.  We could rework the method to be `fn try_drop(&mut self)
+-> Result<(), Error>` and have a similar wrapper function `std::mem::try_drop`
+that took it by value.  This seems very reasonable at first glance, but becomes
+much more difficult when considering how to implement `std::mem::try_drop`.
+This function would probably have to have some compiler magic to recursively
+`drop` each member after `try_drop` is called without calling the `drop`
+instance on the overall `struct`.
+
 # Prior art
 [prior-art]: #prior-art
 
--- a/text/0000-close-trait.md
+++ b/text/0000-close-trait.md
@@ -90,9 +90,10 @@ As it currently stands, there shall be no default implementation.
 ## File
 
 This trait will be implemented for `fs::File` with `Error` specified as
-`io::Error`.
-
-In `src/libstd/fs.rs`:
+`io::Error`.  Similarly to how other methods are implemented on the inner `File`
+implementations, `close` takes `self.inner` by reference (see below).  Thus we
+must `forget` the `File` to prevent a double free.  To do this we place this
+impl block in `src/libstd/fs.rs`:
 
 ```rust
 impl Close for File {
--- a/text/0000-close-trait.md
+++ b/text/0000-close-trait.md
@@ -229,16 +229,30 @@ input parameters for semantically similar functions.
 `Close` has the signature `fn close(self) -> Result<(), Error>` whereas `Drop`
 has `fn drop(&mut self) -> ()`.  These signatures aren't identical, but they
 could be made to be.  We could rework the method to be `fn try_drop(&mut self)
--> Result<(), Error>` and have a similar wrapper function `std::mem::try_drop`
-that took it by value.  This seems very reasonable at first glance, but becomes
-much more difficult when considering how to implement `std::mem::try_drop`.
-This function would probably have to have some compiler magic to recursively
-`drop` each member after `try_drop` is called without calling the `drop`
-instance on the overall `struct`.
+-> Result<(), Error>`.  This yields a few implementation ideas:
 
-The pros of taking `&mut self` are that it would allow for `dyn Close` objects
-to be usable.  It also offers more compatibility with external libraries that
-already use `&mut self` for similar methods.
+1. Treat `try_drop` like `drop`.  Thus we have a similar wrapper function
+   `std::mem::try_drop` that took it by value.  Make the `try_drop` method
+   uncallable.  This would require a lot of new language features just to
+   achieve this simple library feature.  This function would have to have some
+   compiler magic to recursively `drop` each member after `try_drop` is called
+   without calling the `drop` instance on the overall `struct`.
+
+2. Have this method prepare the object for destruction, effectively making it
+   unusable.  This would allow for `dyn Close` objects to be usable.  It also
+   offers more compatibility with external libraries that already use `&mut
+   self` for similar methods.  However, it also brings some down sides.  It
+   renders the object into a unusable state, where every method called either
+   performs undefined behavior or returns an error.  The first case is bad and
+   goes against Rust's core principles.  The second is bad because it forces the
+   object to provide a special case specifically to handle implementing `Close`.
+   For this reason, I believe implementers will choose not to implement this
+   trait when it is so much easier to not worry about it.
+
+### As a method of Write
+
+I believe this is a bad idea because it would remove backwards compatibility.
+It also prevents us from closing reading resources.
 
 # Prior art
 [prior-art]: #prior-art
@@ -260,6 +274,19 @@ or
 [`java.io.Writer`](https://docs.oracle.com/javase/7/docs/api/java/io/Writer.html).
 This RFC is very similar in implementation to this interface except that
 instances of `Read` and `Write` are not required to implement `Close`.
+
+The Rust library Tokio has a similar concept.  It defines the method
+[`shutdown`](https://docs.rs/tokio/0.1.18/tokio/io/trait.AsyncWrite.html#tymethod.shutdown)
+as part of their trait `AsyncWrite`.  As described in the documentation there, this allows
+for handling errors in shutting down the stream before the object is dropped.
+It does this by taking `&mut self` and rendering `self` unusable after it is
+called.  A discussion of the tradeoffs of this are made above.
+
+The Rust library futures-preview has a similar concept as Tokio, defining the
+method
+[`poll_close`](https://docs.rs/futures-preview/0.3.0-alpha.13/futures/io/trait.AsyncWrite.html#tymethod.poll_close)
+as part of their trait `AsyncWrite`.  The semantics are the same as Tokio, but
+the name more resembles that of `Close`.
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
--- a/text/0000-close-trait.md
+++ b/text/0000-close-trait.md
@@ -263,6 +263,8 @@ instances of `Read` and `Write` are not required to implement `Close`.
 What module should this go in?  It seems logical to go in `std::io`, but it
 might be more appropriate for `std::mem` beside `std::mem::drop`.
 
+Should this be named `Close` or `TryDrop`?
+
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
--- a/text/0000-close-trait.md
+++ b/text/0000-close-trait.md
@@ -1,0 +1,209 @@
+- Feature Name: close-trait
+- Start Date: 2019-04-03
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


Nemo157:
For some prior-art in the Rust ecosystem Tokio has `AsyncWrite::shutdown` and Futures 0.3 has `AsyncWrite::poll_close` that have a similar function. One way that these differ from the proposed trait is that they are explicitly only associated with the writer side of a potentially duplex IO stream. This becomes interesting when you consider something like `TcpStream` where you may want to close your outgoing data but continue receiving data until the other side also closes their stream.

CryZe:
I think this could also be generalized to a general `TryDrop` similar to `TryFrom` and `TryInto`.

czipperz:
@CryZe As it stands, `Close` is a generic way to try dropping an object.  I thought about renaming it to `TryDrop`, but decided on `Close` because it is more consistent with other programming languages' features.  However, we could have a method `close` on `File` and other resources that simply `try_drop`s them.

czipperz:
`TryDrop` would also seem to be in favor of defining a default implementation for any type implementing `Drop` by simply returning `Ok(())`.

czipperz:
`Close` also has the signature `fn close(self) -> Result<(), Error>`.  But `Drop` takes `&mut self`.  That is, it isn't a direct generalization from `Drop` to `Close` because the signatures are different.  I'm proposing a library solution here, not a language solution.

A language solution could take `&mut self` and call the `try_drop` with `?`, cascading errors.  This would, however, require a large breaking change to the language and I am not in favor of this.

WiSaGaN:
What is the benefit of adding a new trait rather than just adding as an inherent method?

Ixrec:
@WiSaGaN The RFC gives the examples of close() methods on `BufReader<T>` and `BufWriter<T>` that would not be possible to write if they couldn't be generic over the closeable T they're buffering.

czipperz:
This was at the end of the first paragraph of motivation.  Do you have any other concerns @WiSaGaN ?

> This will also allow us to `close` wrapper types, such as `BufReader` or `BufWriter` when they wrap any type implementing `Close`.

vi:
Does it need to be `self`? Is `&mut self` not enough to close a resource?

pnkfelix:
@vi if you want to statically ensure no one can use the resource after it’s closed, then you need to take `self`, not `&mut self`

casey:
Nice RFC!

I've wanted something like `Close` in the past, and one thing that I also would have used is an annotation that tells the compiler that it should emit a warning or error if a type that implements `Close` is not manually closed, i.e. something like `MustClose`. In my particular case, I had a resource that might throw an error on cleanup, and I wanted to make sure that I always cleaned up instances of those resources and handled the errors appropriately. Having the compiler warn me whenever it was inserting automatic drops fo these resources (and thus not handling errors) would have been very handy.

I personally would find it a bit confusing if all types that implemented `Drop` automatically got implementations of `Close`, since it would look like cleaning up those resources was fallible. However, it might be useful for generic code, and it might have type `Result<(), !>`, indicating that it never fails. Still, I would hold off until there's a need.

I think I like `TryDrop` a bit more than `Close`. I'm imagining resources other than files, things like debugging aids, benchmarking aids, etc, and `close()` doesn't necessarily make sense for those.

Since I think it generalizes to non `io` resources, it would be intuitive to me somewhere other than `io`, perhaps `mem`.

vi:
> @vi if you want to statically ensure no one can use the resource after it’s closed, then you need to take self, not &mut self

@pnkfelix How do I use `Close::close` from `tokio_io::AsyncWrite::shutdown` then? Also how do I close the `Box<dyn TraitThatExtendsClose>`?

czipperz:
@vi It already does have it listed in the alternatives.  I'll add the `dyn` problem to it though

czipperz:
@vi how do we deal with this issue now with just normal Drop implementations? It would have to do something like call `ptr::drop_in_place`, which is unsafe.

czipperz:
Mmh. It seems `shutdown` and the like prepare the object for `Drop` to be called.  I'll put that in the prior art section

matklad:
cc https://github.com/rust-lang/rust/issues/32677

burdges:
If I understand, there are many designs suitable for different settings, but this one fits with `std::io`, right?  

We'd manually propagate errors with this by calling `.close() ?`, right?  I suppose you considered automatically propagating errors with roughly
```
pub trait Drop
    type Error = !;
    try_drop(?? self) -> Result<(),Error> { self.drop(); Ok(()) }
    drop(&mut self);
}
```
Actual drop glue would call `Drop::try_drop(..) ?` resulting in type errors if you lack the correct conversion `impl`s though, making this a massive breaking change, which sounds unacceptable.  It goes against Rust's explicitness goal too I guess.

I'd think transactional data structures require multiple `&mut self` functions that confirm or above transactions, but those sound custom and unlikely to fit anything `std` provides.  I doubt ending session types could be homogenized enough to fit into `std` but.. 
```
/// Try ending a session type
pub trait TryEnd {
    type Error;
    fn close(self) -> Result<(), (Error,Option<Self>)>;
}
impl<T: TryEnd> Drop for T {
    drop(&mut x) { panic!() }
}
```

I think this RFC's design makes sense, but the documentation should explain that `Close` exists primarily for `std::io`, and that it may or may not fit other uses cases, so users should not attempt to shoehorn every type into `Close`.

czipperz:
@burdges 

>```
>pub trait Drop
>    type Error = !;
>    try_drop(?? self) -> Result<(),Error> { self.drop(); Ok(()) }
>    drop(&mut self);
>}
>```

This won't work for the following reasons:

1. `drop` cannot be manually called.  Thus `try_drop` isn't writable.
2. `try_drop` shouldn't be manually called in this instance for the same reason
3. How do we recursively `try_drop` while we can't manually call `try_drop`?  For instance, how does a `BufReader` implement this?

text/0000-close-trait.md
```diff
@@ -0,0 +1,303 @@
...
+
+fn main() -> std::io::Result<()> {
+    let mut file = File::create("foo.txt")?;
+    file.write_all(b"Hello world!")?;
+    Ok(())
+}
+```
+
+Although this checks if there were errors in writing the data to the file, there
+sometimes are spurious failures that aren't caught here.  Our program is
```
scooter-dangle: For someone like myself who might not be familiar with them, can you give some examples of what these failures might be? Are there any that `File::sync_all` wouldn’t surface?

xfix:
About alternatives, I'm wondering about adding a `close` method to `std::io::Write` trait.

```rust
fn close(mut self) -> io::Result<()> where Self: Sized {
    self.flush() // Or `Ok(())` I suppose, this is arguable
}
```

The RFC argues this is a bad idea as "it would remove backwards compatibility", but I don't see it myself being a case. A default implementation would mean that the already existing code would work with a no-op `close` method implementation. One interpretation I could think of is that code explicitly implementing `close` would not work in Rust versions before `close` introduction, but so would be the case whatever we would go with. Could this be clarified?

It also argues that "It also prevents us from closing reading resources", which frankly seems like non-issue, what does `close` failing for a structure implementing `Read` and not `Write` would even mean - there were no changes, so surely an error could be ignored. In theory, `close` can always fail, but I'm not sure what reasonable implementation would return an error other than `EINTR` (which is not really an error) on `close` if a file was never written to.

As for other traits where closing can error, I think it makes sense for those traits to promise their own `close` methods. Having Rust defined `Close` trait isn't really advantageous compared to traits providing their own `close` method.

Keep in mind that having `Close` trait would mean changing `<T: Write>` to `<T: Write + Close>` would break backwards compatibility, as not every `Write` implementation will implement `Close` (if say, due to being written before `Close` existed).

text/0000-close-trait.md
czipperz: For specifically `File`, that is correct.  `sync_all` allows us to handle these errors.  Maybe a better example is something like a `TCPStream` where you have to send a finish packet to signal to the other side to close the connection.

czipperz:
@xfix Thanks for the feedback!

I was arguing that it would be backwards incompatible if we didn't have a default implementation.  That is what Tokio and futures-preview does (see the comment by Nemo above).  That is, they require an implementation of `shutdown` or `poll_close`.

I think it makes sense just to flush and auto drop.  I'll add this in.

>`dyn Write`

To be perfectly honest, it's unclear to me what we should do about this case.  Because when we close a resource we should take it by value and dispose of it.  But at the same time, dynamic types can't work with this.  The approach taken by both Tokio and futures-preview is to take `&mut self`.  I don't particularly like this implementation because it puts the stream into an invalid state and it is unclear what to do besides drop it at that point.

xfix:
Whatever we would do, you could have dynamic dispatch closable `Write` implementations. It's not pretty, but hey, it works. My issue is that there is an implementation of `Write` for `Box<T> where T: ?Sized + Write`, which would do the wrong thing by default.

```rust
use std::io;

trait Write {
    fn close(self) -> io::Result<()>;
}

struct Ex;

impl Write for Ex {
    fn close(self) -> io::Result<()> {
        println!("Closing...");
        Ok(())
    }
}

trait CloseWrite: Write {
    fn close(self: Box<Self>) -> io::Result<()>;
}

impl<T: Write> CloseWrite for T {
    fn close(self: Box<T>) -> io::Result<()> {
        (*self).close()
    }
}

fn main() {
    let x: Box<dyn CloseWrite> = Box::new(Ex);
    x.close();
}
```

It's possible to have a method like this to workaround this issue:

```
#[unstable(feature = "write_internals",
            reason = "this method should not be exposed publicly",
            issue = "0")]
#[doc(hidden)]
pub fn __close_boxed(self: Box<Self>) -> Result<()> {
    (*self).close()
}
```

But sufficient to say, this is a bad idea.
fn __close_boxed

xfix:
Question, let's say `close` method gets added to `Write`, what should be the implementation for `&mut T where T: ?Sized + Write`? Specifically, should it be flushing or not. I can imagine a popular use-case of `close` will be to properly clean up a value instead of calling `drop`, in which case flushing is probably unnecessary.

Nemo157:
IMO if `Write::close` exists (and is defined over `&mut Self`) then `impl Write for &mut impl Write + ?Sized` should be proxying it through correctly, if you want to pass a borrowed `Write` implementation to a function and not have it close it (or not flush it) you can wrap the `&mut Write` in a type that will proxy through all other functions and ignore calls to `close` (or `flush`).

xfix:
@Nemo157 `fn close(self)` takes ownership, so `&mut T` implementation cannot redirect to the inner `close` implementation.

In my opinion, doing nothing (returning `Ok(())` is reasonable here).

Nemo157:
@xfix it doesn't _have_ to take `self` by value, e.g. `TcpStream::shutdown` takes `&self` so you can call that and acquire a `Write` instance that is already closed.

> I'm assuming people may want to update their generic APIs to explicitly call `close` instead of dropping the value, so it makes sense to avoid flushing.

That depends a lot on the specific API, if it's just a helper API then it shouldn't be calling `close` (or `flush`). If it is an API that expects to take full ownership and manage the lifetime of the data stream, including closing it once complete, then not flushing it could cause issues, e.g. if it sends a final small message, closes the writer, then waits for an ack on a reader; if closing the writer doesn't flush out the final message (i.e. the writer were a `&mut BufWriter`) this would sit waiting on the ack forever.

tanriol:
@czipperz Please note that both Tokio's `shutdown` and Futures's `poll_close` have one significant difference in semantics from the suggested `Close` trait. Both are `poll`-style async, so if they return `NotReady` you're supposed to call them again after they're ready to make progress, possibly more than once.

This is actually a known [problem](https://github.com/rust-lang-nursery/futures-rs/issues/1278) that completion-based IO operations may have no way to cancel them synchronously and thus `Drop` does not work with them... `Close`, as defined now, will not work either. This does not feel good, especially taking into account that these are just the kind of operations where one can expect an error during shutdown.

xfix:
Closing asynchronous I/O should be out of scope for this RFC. It's so different to regular closing that even if we somehow made `Close` work for both synchronous and asynchronous code, the bounds for functions using `Close` trait would be so different that there would be no reason for doing so.

mzabaluev:
@xfix
>     * `File` - `self.sync_all()`

That's not what the OS `close` typically does, and it may be unnecessary in many cases. The generic `close` for I/O objects should only do the minimal work to finalize the object's pending operations and resources for which failures may need to be observed.

mzabaluev:
Following the design of a session-ending trait thrown in by @burdges, `Close`/`TryDrop`, too should return any information that can be used for recovery in the `Err` variant. For `BufWriter` this could be:

```rust
impl<W: Write + Close> Close for BufWriter<W> {
    type CloseError = buf_writer::CloseError<W>;
    fn close(self) -> Result<(), Self::CloseError> {
        let flush_res = self.flush();
        if let Err(e) = flush_res {
            return buf_writer::CloseError::incomplete_flush(e, self);
        }
        self.inner.close()?;  // The error variant returned here features W::CloseError
    }
}
```

pdolezal:
I recently hit the need of some kind of `Close` as well and I got reference to this PR. Perhaps I could refresh it here a bit with the [current discussion on Rust Internals](https://internals.rust-lang.org/t/question-closing-resources-like-std-write/13169) which I opened a few days ago (not knowing about this PR yet).

czipperz:
I'm not using Rust anymore so if someone else wants this PR to go through feel free to fork my PR.

3point2:
I came across this PR while working on BurntSushi/ripgrep#1766. Just want to add that during my research I found https://crates.io/crates/io-close, which implements a Close trait. It works by taking ownership, flushing, using unsafe {} to close the underlying fd/handle/socket, and returning any error.

yaahc:
Hey, we discussed this RFC in last week's [libs team meeting](https://hackmd.io/7mVQp2OgSlS_LomurCWy1g#Nominated). There are some additional notes in the agenda but the gist of the discussion is that we're unsure about the specific motivating use cases, and we're unsure if this should be a trait. As an alternative we were wondering if an inherent `File::close` method would be sufficient.

Could you update the RFC to include some example use cases that motivate / justify this being a Trait? Assuming you feel it needs to be a trait.

pdolezal:
> Hey, we discussed this RFC in last week's [libs team meeting](https://hackmd.io/7mVQp2OgSlS_LomurCWy1g#Nominated). There are some additional notes in the agenda but the gist of the discussion is that we're unsure about the specific motivating use cases, and we're unsure if this should be a trait. As an alternative we were wondering if an inherent `File::close` method would be sufficient.

I can point out to the [IRLO discussion](https://internals.rust-lang.org/t/question-closing-resources-like-std-write/13169) where the example, which led to the discussion, was mentioned. Briefly, I wanted to implement a bitwise compression algorithm that is supposed to terminate the output, so it has to emit the termination when the output is going to finish. I wanted to implement it as a `Write` decorator and I wanted yet to have the ability to indicate that emitting the terminator failed, so that the user of the decorator could react rather than, e.g., getting a corrupted file silently. I don't see any obvious way to make such a generic decorator without having a generic way to learn that a final operation of some sort failed.

I mentioned in the discussion that a more generic trait, let's call it `TryDrop`, might be available instead. Closing a file seems like an important, but still special case. I can imagine streams of different elements than raw bytes to be pushed into a (buffered) sink, which may, e.g., return an error (that might include yet unprocessed elements) on a premature drop attempt.

czipperz:
I no longer use Rust so am going to close this issue.  If anyone is interested in moving forward with this proposal feel free to fork it.

PR #2693, opened 2019-04-24
Add Unicode Annex 31 methods to `char`

notriddle:
[Rendered](https://github.com/notriddle/rust-rfcs/blob/patch-2/text/0000-char-uax-31.md)

### DIFF
--- a/text/0000-char-uax-31.md
+++ b/text/0000-char-uax-31.md
@@ -70,7 +70,7 @@ even though Unicode places plenty of requirements on that process, too.
 
 The other problem is that a lot of languages aren't defined in terms of Annex 31 anyway,
 like Swift and HTML, which simply spell out the set of allowed code points themselves,
-so this isn't necessarily useful to allw of the language implementers.
+so this isn't necessarily useful to all of the language implementers.
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
--- a/text/0000-char-uax-31.md
+++ b/text/0000-char-uax-31.md
@@ -21,7 +21,7 @@ a standardized set of code point categories for defining computer language synta
 
 This is being used in production Rust code already.
 Rust's own compiler already has functions to check against Annex 31 code point categories in the lexer,
-[but not everyone who works on the compiler knows about them](https://internals.rust-lang.org/t/for-await-loops/9819/16),
+[but not everyone who works on the compiler knows about them](https://internals.rust-lang.org/t/do-we-need-unicode-whitespace/9876),
 and since they're not in the standard library,
 not everyone who works on Rust-related tooling has access to them.
 I'm not asserting that putting these in libstd would've avoided that bug,
--- a/text/0000-char-uax-31.md
+++ b/text/0000-char-uax-31.md
@@ -7,7 +7,8 @@
 [summary]: #summary
 
 Add functionc to the standard library for testing a `char` against [UAX TR31](https://unicode.org/reports/tr31/) ("Unicode Annex 31")
-`Pattern_White_Space`, `Pattern_Syntax`, `XID_Start`, `ID_Nonstart`, and `XID_Continue`.
+`Pattern_White_Space`, `Pattern_Syntax`, `XID_Start`, `ID_Nonstart`, and `XID_Continue` (the XID ones are already in the standard
+library, but are unstable; this RFC proposes to stablize them).
 
 # Motivation
 [motivation]: #motivation
@@ -45,17 +46,10 @@ These functions are also exposed as methods on the `char` type.
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-## `fn char::is_xid_start(self) -> bool`
-
-Check if `self` is a member of Unicode Annex 31's `XID_Start` code point category.
-
-## `fn char::is_xid_continue(self) -> bool`
-
-Check if `self` is a member of Unicode Annex 31's `XID_Continue` code point category.
-
 ## `fn char::is_id_nonstart(self) -> bool`
 
 Check if `self` is a member of Unicode Annex 31's `ID_Nonstart` code point category.
+This function is defined as `self.is_xid_continue() && !self.is_xid_start()`.
 
 ## `fn char::is_pattern_syntax(self) -> bool`
 
--- a/text/0000-char-uax-31.md
+++ b/text/0000-char-uax-31.md
@@ -1,5 +1,5 @@
 - Feature Name: `char_uax_31`
-- Start Date: 2019-24-04
+- Start Date: 2019-04-24
 - RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
 - Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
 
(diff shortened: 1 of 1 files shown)


clarfon:
I think this could probably be done as a PR to Rust directly rather than an RFC.

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
+- Feature Name: `char_uax_31`
+- Start Date: 2019-24-04
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Add functionc to the standard library for testing a `char` against [UAX TR31](https://unicode.org/reports/tr31/) ("Unicode Annex 31")
```
mibac138:
```suggestion
Add functions to the standard library for testing a `char` against [UAX TR31](https://unicode.org/reports/tr31/) ("Unicode Annex 31")
```

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
...
+
+The big problem, that has always made designing the text APIs hard,
+is that it's not clear how much of Unicode we want to include in libstd.
+The standard library certainly doesn't want a hashtag parser, even though Annex 31 describes one in section 6,
+and libstd certainly doesn't want a character shaping algorithm,
+even though Unicode places plenty of requirements on that process, too.
+
+The other problem is that a lot of languages aren't defined in terms of Annex 31 anyway,
+like Swift and HTML, which simply spell out the set of allowed code points themselves,
+so this isn't necessarily useful to allw of the language implementers.
```
mibac138:
```suggestion
so this isn't necessarily useful to all of the language implementers.
```

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
...
+
+As a systems language, Rust is heavily used for parsing.
+As a progressive, forward-thinking language that accepts anyone,
+Rust supports Unicode and makes the definitive string types UTF-8.
+At the intersection of these needs sits *UAX #31: Unicode Identifier and Pattern Syntax* ("Annex 31"),
+a standardized set of code point categories for defining computer language syntax.
+
+This is being used in production Rust code already.
+Rust's own compiler already has functions to check against Annex 31 code point categories in the lexer,
+[but not everyone who works on the compiler knows about them](https://internals.rust-lang.org/t/for-await-loops/9819/16),
```
Manishearth: I think this link is incorrect

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
...
+[motivation]: #motivation
+
+As a systems language, Rust is heavily used for parsing.
+As a progressive, forward-thinking language that accepts anyone,
+Rust supports Unicode and makes the definitive string types UTF-8.
+At the intersection of these needs sits *UAX #31: Unicode Identifier and Pattern Syntax* ("Annex 31"),
+a standardized set of code point categories for defining computer language syntax.
+
+This is being used in production Rust code already.
+Rust's own compiler already has functions to check against Annex 31 code point categories in the lexer,
```
Manishearth: This is for the unstable, old, non_ascii_idents feature, which [we have already RFCd to change](https://github.com/rust-lang/rfcs/blob/master/text/2457-non-ascii-idents.md). The change will still need _some_ function like this, but it may need tailoring for bidi characters. This is true for other implementors too, the XID functions often need tailoring.

They're not super hard to tailor, though, so we could still expose these functions to match spec and let those tailoring just suffix the calls with `|| ch == ... && ch != ...`

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
...
+
+Check if `self` is a member of Unicode Annex 31's `Pattern_Syntax` code point category.
+
+## `fn char::is_pattern_white_space(self) -> bool`
+
+Check if `self` is a member of Unicode Annex 31's `Pattern_White_Space` code point category.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
```
Manishearth: Matching unicode versions is also a big issue here. This function will be inaccurate half the time as we don't always update our data files immediately (it's not always straightforward). Even if we do, the behavior of this function will change every year, and while we don't have a guarantee on stdlib behavior stability, this does mean that older compilers will lead to different results on code that compiles. This further makes me feel like this should be a versioned crate.

is_whitespace already opened the doors to this issue, but I don't want to make it worse. is_whitespace is a small relatively stable list whereas XID expands all the time.

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
...
+[summary]: #summary
+
+Add functions to the standard library for testing a `char` against [UAX TR31](https://unicode.org/reports/tr31/) ("Unicode Annex 31")
+`Pattern_White_Space`, `Pattern_Syntax`, `XID_Start`, `ID_Nonstart`, and `XID_Continue` (the XID ones are already in the standard
+library, but are unstable; this RFC proposes to stablize them).
+
+# Motivation
+[motivation]: #motivation
+
+As a systems language, Rust is heavily used for parsing.
```
Manishearth: To me this reads as motivation for why Rust needs such functions _somewhere_, but inclusion in the stdlib is a much higher bar, _especially_ when this gets us tangled up with unicode versioning.

Yes, we already have is_whitespace, but that's an old API that was grandfathered in, and it has fewer unicode stability issues than XID.

Manishearth: One argument for it being in std is discoverability, but if you're parsing some grammar that grammar will likely tell you about XID.

text/0000-char-uax-31.md
```diff
@@ -0,0 +1,103 @@
+- Feature Name: `char_uax_31`
+- Start Date: 2019-24-04
```
mzji:
```suggestion
- Start Date: 2019-04-24
```

text/0000-char-uax-31.md
notriddle: You're right.

text/0000-char-uax-31.md
notriddle: The XID part is currently unstable. But stable Rust does respect `Pattern_White_Space`, so there's already committed-to Annex 31-based syntax.

text/0000-char-uax-31.md
notriddle: Agreed. Added it.

text/0000-char-uax-31.md
matklad: As a concrete case of tailoring, Rust actually uses `XID_Start | '_'` and not a plain `XID_Start`. I've actually got bitten by that when implementing lexer for IntelliJ Rust a while ago :)

matklad:
cc https://github.com/rust-lang/rust/pull/62848 which proposes to move in the **opposite** direction: remove these methods from `libcore` and use `unicode_xid` crate in the compiler.

scottmcm:
https://github.com/rust-lang/rust/pull/62848 appears to have reached consensus for the opposite direction; does that amount to an agreement to close this one?

notriddle:
Yeah, I will definitely rescind this whole thing, if that's going to be the solution.

matklad:
@notriddle `is_xid_start` and `is_xid_continue` were removed from libcore. rustc now uses [unicode-xid](https://crates.io/crates/unicode-xid) crate as the source of truth for these definitions. And, if one is interested specifically in Rust definition of identifiers, there's [rustc_lexer](https://crates.io/crates/rustc_lexer). 

Based on the [comment above](https://github.com/rust-lang/rfcs/pull/2693#issuecomment-533759615) I am going to optimistically close this issue and remove the nominated label, but please reopen if I misunderstood your intentions.

PR #2721, opened 2019-07-11
build: Improve CI time

lzutao:


Mark-Simulacrum:
The current CI time appears to be about ~10 minutes -- and given that RFCs are merged rarely, I don't really want to spend the time it'll take me to figure out whether this is a correct change, sorry. If CI times become a problem at some point we can revisit. Thank you!

PR #2724, opened 2019-07-14
Add link points to non_exhaustive feature

lzutao:


KodrAus:
Hm, if anybody has a different opinion I'd be all ears, but I feel like this would be better as an addition to the existing text mentioning the postponed RFC, rather than a replacement.

clarfon:
Honestly, why not just update the original RFC for private enum variants and reference the RFC for the `non_exhaustive` feature? There are so many places that reference it and it would be a bit easier to do that than to update every reference to it.

lzutao:
Don't know if is still needed.

PR #2786, opened 2019-10-17
avoid unreasonable growth of the repository size

matklad:
At the moment, the rfcs repo is pretty huge. The big history of
gh-pages is to blame, as each gh-pages commit contains a 50MB search.js
file. There's little value in preserving history of gh-pages branch,
so let's avoid creating new commits.

Note that someone should manually squash existing commits to one on the gh-pages branch. Alternatively, we can merge a one-off change to .travis.yml for squashing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,8 @@ script:
   # of our files (e.g. ignoring directories with a leading '_').
   - touch .nojekyll
   - git add -A .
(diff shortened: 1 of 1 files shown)


fbstj:
this is "fixing" #2750 ?

matklad:
Ah indeed, the approach in #2751 seems better!

PR #2796, opened 2019-10-28
RFC: Subslice-offset - Get the offset of references into a slice.

m-ou-se:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/2796)*

[Rendered](https://github.com/rust-lang/rfcs/blob/3e5e3bbbd5678a39175ad2d409f4131bb4fdd943/text/0000-subslice-offset.md)

(This idea was mentioned in the 'Future possibilities' section of https://github.com/rust-lang/rfcs/pull/2791)

### DIFF
--- a/text/0000-subslice-offset.md
+++ b/text/0000-subslice-offset.md
@@ -1,0 +1,344 @@
+- Feature Name: `subslice_offset`
+- Start Date: 2019-10-27
+- RFC PR: None yet
+- Rust Issue: None yet
+
+# Summary
+[summary]: #summary
+
+Add functions to the slice primitive to get the offset/index of an element or
+subslice based on their memory addresses:
+
+```rust
+impl<T> [T] {
+    pub fn index_of(&self, element: &T) -> Option<usize>;
+    pub fn range_of(&self, subslice: &[T]) -> Option<Range<usize>>;
+}
+```
+
+# Motivation
+[motivation]: #motivation
+
+In many cases, one might end up with a reference to some element in&mdash;or
+some part of&mdash;a slice.
+For example, after using `split` or `chunks` on a slice, using some searching
+algorithm from a external crate, etc.
+
+Some of these, like `matches`, have a counterpart that also gives the indices of
+the parts, like `match_indices`. However, many don't.
+
+The proposed functions make it easy to safely get indices in these cases,
+without having to resort to manual bookkeeping or (unsafe) pointer math.
+
+For example, for `split`: instead of adding a separate `[T]::split_indices`,
+`range_of` could be used to get the indices of the split results.
+
+Even when all standard library functions giving references to elements or
+subslices would have a variant giving the indices as well, we can't expect all
+crates to do this. `index_of` and `range_of` provide a safe way to get an index
+when all you have is a reference (and the original slice).
+
+Also, in some cases, using this functionality would allow for more efficient
+code, as the best alternative is often manual bookkeeping: keeping track of the
+offset/index in addition to the reference, which may take extra time and space.
+
+#### Use Case Example 1: Avoid Pointer Math
+
+In the implementation of `PathBuf::set_extension`,
+[pointer math is used][offset_example] to calculate the index of the end
+of the `Path::file_stem()` in the path, to find the position where the new
+extension should be added. Right now this involves casting raw pointers to
+`usize`s and subtracting them, making the code non-trivial and error-prone:
+
+```rust
+let end_file_stem = file_stem[file_stem.len()..].as_ptr() as usize;
+let start = os_str_as_u8_slice(&self.inner).as_ptr() as usize;
+(...)
+v.truncate(end_file_stem.wrapping_sub(start));
+```
+
+With the proposed functions, it'd look like this:
+
+```rust
+let file_stem_range = os_str_as_u8_slice(&self.inner)
+    .range_of(file_stem)
+    .expect("file_stem() gave a slice outside of the path");
+(...)
+v.truncate(file_stem_range.end);
+```
+
+[offset_example]: https://github.com/rust-lang/rust/blob/18ae175d60039dfe2d2454e8e8099d3b08039862/src/libstd/path.rs#L1372-L1375
+
+#### Use Case Example 2: Increase Efficiency
+
+For this example, take this struct representing some token in a tokenized source
+file:
+
+```rust
+struct Token<'a> {
+  (...)
+  source: &'a str, // the exact source of this token
+  offset_in_source_file: usize, // for error reporting
+}
+```
+
+If the `source` always points into the full `String` contents of the file,
+keeping `offset_in_source_file` is a waste of space, as that information is
+already encoded in the memory address stored in `source`:
+`file_contents.range_of(token.source)`.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## index_of
+
+```rust
+impl<T> [T] {
+    pub fn index_of(&self, element: &T) -> Option<usize>
+}
+```
+
+Returns the index of the element that a reference refers to.
+
+If the given reference points inside the slice, returns the index of
+the element it refers to. If the reference does not refer to within the
+slice, `None` is returned instead.
+
+If an index is returned, it is less than `self.len()`.
+
+Note that this does not look at the value, but only at the reference.
+If you want to find the index of an element equal to a given value, use
+`iter().position()` instead.
+
+```rust
+let a = [0; 5];
+
+assert_eq!(a.index_of(&a[2]), Some(2));
+assert_eq!(a.index_of(&3), None);
+```
+
+## range_of
+
+```rust
+impl<T> [T] {
+    pub fn range_of(&self, subslice: &[T]) -> Option<Range<usize>>
+}
+```
+
+Returns the range referred to by a subslice.
+
+If the given slice falls entirely within this slice, returns the range
+of indices the subslice refers to. If the given slice is not a subslice
+of this slice, `None` is returned.
+
+If a range is returned, both ends are less than or equal to `self.len()`.
+
+Note that this does not look at the contents of the slice, but only at
+the memory addresses.
+
+```rust
+let a = [0; 5];
+
+assert_eq!(a.range_of(&a[2..5]), Some(2..5));
+assert_eq!(a.range_of(&[7, 8, 9]), None);
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The functions below are added to `impl<T> [T] { .. }` in `src/libcore/slice/mod.rs`,
+with doc comments based on the [Guide-level explanation][guide-level-explanation] above.
+
+```rust
+    pub fn index_of(&self, element: &T) -> Option<usize> {
+        let element = element as *const _;
+        let range = self.as_ptr_range();
+        if range.contains(&element) {
+            unsafe { Some(element.offset_from(range.start) as usize) }
+        } else {
+            None
+        }
+    }
+
+    pub fn range_of(&self, subslice: &[T]) -> Option<Range<usize>> {
+        let range = self.as_ptr_range();
+        let subrange = subslice.as_ptr_range();
+        if subrange.start >= range.start && subrange.end <= range.end {
+            unsafe {
+                Some(Range {
+                    start: subrange.start.offset_from(range.start) as usize,
+                    end: subrange.end.offset_from(range.start) as usize,
+                })
+            }
+        } else {
+            None
+        }
+    }
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- It might encourage users to use this in places where better alternatives
+  are available. For example: `.iter().find()` in combination with `.index_of()`
+  instead of `.iter().position()`.
+
+- In most parts of the library, two references to different but equal objects
+  behave the same. The memory address itself being significant instead of the
+  value it refers to is somewhat uncommon.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This functionality is already available through calculations with raw pointers.
+Often, people prefer to avoid that, which in many cases leads to complicating
+the code in other ways.
+
+The main consideration is which things should have a safe and ergonomic option,
+and which are best left to raw pointer manipulation.
+
+The proposed functions put the line between references and pointers:
+the functions only work on references; as soon as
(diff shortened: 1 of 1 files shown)


text/0000-subslice-offset.md
```diff
@@ -0,0 +1,344 @@
...
+
+The proposed functions put the line between references and pointers:
+the functions only work on references; as soon as
+you're using raw pointers, you'll have to do things manually instead.
+
+The alternative is to try to cover some cases for raw pointers as well:
+
+## Raw Pointers
+
+By requiring a `&T` or a `&[T]` in these functions, we needlessly restrict them
```
rkruppe: This restriction to dereferenceable pointers side-steps potential problems with pointer comparisons: compiler optimizations can change the results of pointer comparisons to be different from the order of the runtime addresses when the pointers are dangling. See for example section 4.6 of [this paper](https://people.mpi-sws.org/~jung/twinsem/twinsem.pdf).

It is not immediately apparent to what degree LLVM performs such optimizations, but it seems risky to assume that it never will. LLVM docs say [`icmp` on pointers works "as if [the pointers] were integers"](https://llvm.org/docs/LangRef.html#icmp-instruction) but this is imprecise and apparently contradicted by the fact that [some comparisons](https://godbolt.org/z/NWGoGt) which can be true at runtime under -O0 are [folded to `false` at -O2](https://godbolt.org/z/x50w1j).

cc @rust-lang/wg-unsafe-code-guidelines

m-ou-se: Thanks for the feedback!

I mention somthing similar briefly, but it's a bit hidden at the end of 'option 4':

>  The downside is that we lose the guarantees of a &[T], and can no longer make assumptions such as start <= end or about the maximum size of a slice (which is needed to safely use pointer::offset_from).

It's part of the reason why I'm proposing the versions with references, and only mention the pointers as alternatives. I should probably mention this more clearly right at the start of the raw pointers section.

rkruppe: I did miss that part, thanks for pointing it out.

Those problems you list seem of a different nature, though. Not knowing that `start <= end` and that the slice spans no more than `isize::MAX` bytes requires a different, slightly less efficient implementation (e.g., some more comparisons). In contrast, it is not clear to me whether the problem of non-deterministic pointer comparison can be circumvented at all, since e.g. ptr->int casts [do not currently stop the LLVM optimization I showcased earlier](https://godbolt.org/z/ur5Gjt). That may have to change anyway to make LLVM's memory model consistent, but at present I don't have confidence the raw pointer version can be implemented correctly, at least according to the naive specification of comparing the pointer addresses numerically.

jeekobu:
I feel these would be useful.  I've implemented the following less-generic analog before and was surprised it didn't exist.

```rust
fn offset_within(parent: &str, child: &str) -> Option<usize> { ... }
```

text/0000-subslice-offset.md
```diff
@@ -0,0 +1,344 @@
...
+
+The functions below are added to `impl<T> [T] { .. }` in `src/libcore/slice/mod.rs`,
+with doc comments based on the [Guide-level explanation][guide-level-explanation] above.
+
+```rust
+    pub fn index_of(&self, element: &T) -> Option<usize> {
+        let element = element as *const _;
+        let range = self.as_ptr_range();
+        if range.contains(&element) {
+            unsafe { Some(element.offset_from(range.start) as usize) }
```
HeroicKatora: This implementation panics for ZSTs (but would otherwise be unsound anyways). The requirements of `offset_from` state:

>  Both the starting and other pointer must be either in bounds or one byte past the end of the same allocated object. Note that in Rust, every (stack-allocated) variable is considered a separate allocated object.
> [...]
> Panics
> This function panics if T is a Zero-Sized Type ("ZST").

References to ZSTs are always dangling pointers and thus `offset_from` never valid as they never point into any allocation. The non-ZST case could be improved with an explicit comment detailing soundness justifications.

Also note that even if having a special case implementation, slices of ZST will return `Some(0)` in many more cases than the caller might realize, sometimes depending on optimization levels.

```
let not_an_allocation = vec![(); 16];
let some_new = &();
// Entirely indetermined, might hold or not depending on compiler
assert!(not_an_allocation.index_of(&some_new).is_some());
```

text/0000-subslice-offset.md
```diff
@@ -0,0 +1,344 @@
...
+        } else {
+            None
+        }
+    }
+
+    pub fn range_of(&self, subslice: &[T]) -> Option<Range<usize>> {
+        let range = self.as_ptr_range();
+        let subrange = subslice.as_ptr_range();
+        if subrange.start >= range.start && subrange.end <= range.end {
```
HeroicKatora: Again, this does not consider the ZST case. As outlined above, pointers may compare equal even though they logically refer to different regions in the slice. A sensible choice if the operation should succeed could be to assume a subslice starting at index `0` and having a length of `subslice.len()`.

This shows that the implementation is entirely non-trivial. This is a major motivation for having such an interface in `std`, in my opinion.

197g (commented):
The feature might be confusing for slices of ZSTs. In particular, all elements are virtually indistinguishable from their pointers alone as they have the same address. The only useful return values of `index_of` seem `Some(0)` or `None`.

text/0000-subslice-offset.md
HeroicKatora: Also, it is unsound in general.

Consider that the subslice may be empty. In that case, the start and end pointer can fulfil the pointer comparison but not in fact be part of the same allocation. Note that this may be the case even for standard/non-ZST types. Then the `offset_from` calls are UB and do not panic. The crucial difference is that unlike the single element case even the `start` pointer may be a one-past-the-end pointer.

text/0000-subslice-offset.md
RalfJung: @rkruppe unfortunately I don't think restricting to dereferencable pointers entirely side-steps the problems here when ZST are involved. ZST pointers are nominally "dangling" as far as LLVM is concerned (or at least, they might be). But then, as @HeroicKatora points out, there are larger problems with ZST.

RalfJung: > That may have to change anyway to make LLVM's memory model consistent, but at present I don't have confidence the raw pointer version can be implemented correctly, at least according to the naive specification of comparing the pointer addresses numerically.

Indeed LLVM is currently in a sad state where it is not possible to reliably compare the integer addresses of two pointers. Casting to integers first and then comparing *should* work but LLVM incorrectly "optimizes" that to comparing at pointer type instead. We could conceivably hack around that by "obfuscating" the comparison enough, which however would likely be catastrophic for performance. Also see these LLVM bugs: [[1]](https://bugs.llvm.org/show_bug.cgi?id=34548), [[2]](https://bugs.llvm.org/show_bug.cgi?id=35229).

text/0000-subslice-offset.md
RalfJung: Agreed -- if `subslice` is empty this should short-circuit and return `None` (or maybe `Some(0..0)`?).
Otherwise it should use `index_of(subrange.start)` and then add `subslice.len` to get the end index.

steffahn:
There’s also UB in the proposed implementation, if it is called with slices [like this (playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7b663101329adc698e7b527c4af771b3).

RustyYato:
Note: The UB comes from 

> If any of the following conditions are violated, the result is Undefined Behavior:
> 
>    * Both the starting and other pointer must be either in bounds or one byte past the end of the same allocated object. Note that in Rust, every (stack-allocated) variable is considered a separate allocated object.
>
>    * ***Both pointers must be derived from a pointer to the same object.*** (See below for an example.)
>
>    * The distance between the pointers, in bytes, cannot overflow an isize.
>
>    * The distance between the pointers, in bytes, ***must be an exact multiple of the size of T.***
>
>    * The distance being in bounds cannot rely on "wrapping around" the address space.

Emphasis mine. The first point fundamentally can't be checked, the second point could be checked, but isn't right now.

This UB can be alleviated by using `wrapping_offset_from` instead, if that existed! (I'm surprised it doesn't exist). You'll have to convert the pointers to integers and do the arithmetic yourself to be safe.

text/0000-subslice-offset.md
```diff
@@ -0,0 +1,344 @@
...
+# Summary
+[summary]: #summary
+
+Add functions to the slice primitive to get the offset/index of an element or
+subslice based on their memory addresses:
+
+```rust
+impl<T> [T] {
+    pub fn index_of(&self, element: &T) -> Option<usize>;
+    pub fn range_of(&self, subslice: &[T]) -> Option<Range<usize>>;
```
SimonSapin, line 15: Returning a range instead of just the start offset seems not strictly necessary since its length should always be `subslice.len()`, but on further thought it’s nice to have and makes the meaning of the return value obvious. So :+1:

SimonSapin:
Here is another proposed implementation based on the feedback here. The main changes are explicit handling of ZSTs, and "manually" doing arithmetic on pointer addresses instead of using `<*const T>::offset_of` to avoid its potential UB.

In order to handle the case mentioned https://github.com/rust-lang/rfcs/pull/2796#issuecomment-772914452 (reproduced in code comments) we need to compute the distance in bytes and check its modulo. At that point we’ve already done half of the job of `offset_of` so we might as well do the division on that integer.

Doc comments are to be added to those already specified in the RFC. Maybe something should also be added about potentially returning `Some` in unexpected cases with ZSTs? https://github.com/rust-lang/rfcs/pull/2796#discussion_r340943150

<details>
<summary>Proposed implementation</summary>

```rust
use std::mem;
use std::ops::Range;

trait SliceExt<T> {
    fn index_of(&self, element: &T) -> Option<usize>;
    fn range_of(&self, subslice: &[T]) -> Option<Range<usize>>;
}

impl<T> SliceExt<T> for [T] {
    /// # Zero-sized types
    ///
    /// When `size_of::<T>() == 0` the memory address of `&T` is not meaningful,
    /// so `index_of` can only return `None` or `Some(0)`.
    fn index_of(&self, element: &T) -> Option<usize> {
        let size_of = mem::size_of::<T>();
        if size_of > 0 {
            let range = as_address_range(self);
            let element_address = element as *const T as usize;
            if range.contains(&element_address) {
                let bytes_offset = element_address - range.start;
                if bytes_offset % size_of == 0 {
                    Some(bytes_offset / size_of)
                } else {
                    // This could happen with arguments like this:
                    //
                    // ```
                    // use bytemuck::cast_slice;
                    // let x: &[[i32; 2]] = &[[1,2],[3,4]];
                    // let y: &[[i32; 2]] = cast_slice(&cast_slice::<_, i32>(x)[1..3]);
                    // let index = x.index_of(&y[0]);
                    // dbg!(x, y, index);
                    // ```
                    //
                    // … which prints:
                    //
                    // ```
                    // x: [[1,2],[3,4]]
                    // y: [[2,3]]
                    // index: None
                    // ```
                    None
                }
            } else {
                None
            }
        } else {
            // All items of a slice of ZST share the same address
            if element as *const T == self.as_ptr() {
                Some(0)
            } else {
                None
            }
        }
    }

    /// If a range is returned, its length is always `subslice.len()`.
    ///
    /// # Zero-sized types
    ///
    /// When `size_of::<T>() == 0` the memory address of `&T` is not meaningful,
    /// so `index_of` can only return `None` or `Some(0..subslice.len())`.
    fn range_of(&self, subslice: &[T]) -> Option<Range<usize>> {
        let size_of = mem::size_of::<T>();
        if size_of > 0 {
            let range = as_address_range(self);
            let subrange = as_address_range(subslice);
            if subrange.start >= range.start && subrange.end <= range.end {
                let bytes_offset = subrange.start - range.start;
                if bytes_offset % size_of == 0 {
                    let start = bytes_offset / size_of;
                    let end = start + subslice.len();
                    Some(start..end)
                } else {
                    // This could happen with arguments like this:
                    //
                    // ```
                    // use bytemuck::cast_slice;
                    // let x: &[[i32; 2]] = &[[1,2],[3,4]];
                    // let y: &[[i32; 2]] = cast_slice(&cast_slice::<_, i32>(x)[1..3]);
                    // let range = x.range_of(y);
                    // dbg!(x, y, range);
                    // ```
                    //
                    // … which prints:
                    //
                    // ```
                    // x: [[1,2],[3,4]]
                    // y: [[2,3]]
                    // range: None
                    // ```
                    None
                }
            } else {
                None
            }
        } else {
            // All items of a slice of ZST share the same address
            if subslice.as_ptr() == self.as_ptr() {
                Some(0..subslice.len())
            } else {
                None
            }
        }
    }
}

fn as_address_range<T>(slice: &[T]) -> Range<usize> {
    let ptr_range = slice.as_ptr_range();
    let start = ptr_range.start as usize;
    let end = ptr_range.end as usize;
    start..end
}
```

</details>

SimonSapin:
I’ve edited my proposed doc-comment additions to cover ZSTs and empty subslices.

Rua:
I was going to propose `index_of` myself, so I'm happy to see this is already proposed and being worked on.

kpreid:
[I think this RFC is a good idea](https://internals.rust-lang.org/t/proposal-get-range-of-sub-slice/16556/8?u=kpreid), but the specific choice of name `index_of()` concerns me, because it sounds too much like an innocuous value-comparison operation, not one which rightfully has the warning “Note that this does not look at the value, but only at the reference” in its documentation. Generally Rust keeps pointer-equality in items with `ptr` in the name, like `Arc::ptr_eq()`, and everything else is by value.

Unfortunately, these particular methods _don't take pointers_ but putting `ptr` in the name would sound like they do. Maybe `index_of_ref()` and `range_of_ref()` would be adequate to flag “something unusual is here”? Or more verbosely, `index_of_contained_ref()`?

m-ou-se:
Unfortuantely, I won't have the time to update this RFC any time soon. If someone wants to push this feature forward without waiting for me to update it, please feel free to open a new RFC taking whatever you need from this one. :)

joshtriplett:
This came up in a recent discussion.

Potential alternative: a function on a slice that takes another slice and computes the offset from one to the other. So, for instance, a function on a `&str` that accepts another `&str` and returns a `usize`. This would be useful, for instance, after splitting a string, to get the offset of the split-out string in the original string.

m-ou-se:
https://github.com/rust-lang/libs-team/issues/382 has been accepted, tracked in https://github.com/rust-lang/rust/issues/126769, and those methods have been stabilized. Hurray.

PR #2811, opened 2019-11-10
RFC: Generic parameters in derive

mzabaluev:
Add ability to pass generic parameters of the impl to the derive macros, greatly increasing the flexibility of the `derive` attribute.

### DIFF
--- a/text/0000-generic-derive.md
+++ b/text/0000-generic-derive.md
@@ -1,0 +1,269 @@
+- Feature Name: `generic_derive`
+- Start Date: 2019-11-09
+- RFC PR: 
+- Rust Issue: 
+
+# Summary
+[summary]: #summary
+
+Add ability to pass generic parameters of the impl to the derive macros,
+greatly increasing the flexibility of the `derive` attribute.
+
+# Motivation
+[motivation]: #motivation
+
+Derive macros are a very convenient way to generating trait impls based on
+the definition item of a type. However, the ability to use `#[derive(Trait)]`
+is denied when the impl must have generic parameters that need to be defined
+and bound in a more customized way than what the derive macro could generate
+automatically based on the definition item of the `Self` type.
+
+Consider The Most Annoying Problem of `#[derive(Clone)]`:
+
+```rust
+#[derive(Clone)]
+pub struct WaitingForGodot<T> {
+    // ...
+    _phantom_godot: PhantomData<T>
+}
+```
+
+The use of `derive` here is often a convenient pitfall that generates this impl:
+
+```rust
+impl<T: Clone> Clone for WaitingForGodot<T> {
+    //  ^---- Oops, did not really need this bound
+    // ...
+}
+```
+
+This can be easily solved by customizing the impl parameter:
+
+```rust
+#[derive(<T> Clone)]
+pub struct WaitingForGodot<T> {
+    // ...
+    _phantom_godot: PhantomData<T>
+}
+```
+
+More traits could be made conveniently derivable with custom generics than
+is feasible now:
+
+```rust
+use derive_unpin::Unpin;
+
+#[derive(<St: Unpin, F> Unpin)]
+pub struct MyFold<St, F> {
+    #[unsafe_pinned]
+    stream: St,
+    #[unsafe_unpinned]
+    op: F,
+}
+```
+
+In tandem with more elaborate helper attributes, it could be even more powerful:
+
+```rust
+// A not-yet-written library providing the derive macro
+use async_state_machine::Future;
+use futures::future::{TryFuture, IntoFuture, MapOk};
+
+#[derive(
+    <
+        Fut1: TryFuture,
+        Fut2: TryFuture<Error = Fut1::Error>,
+        F: FnOnce(<Fut1 as TryFuture>::Ok) -> Fut2,
+    > Future
+)]
+enum AndThen<Fut1, Fut2, F> {
+    First(MapOk<Fut1, F>),
+    #[after(First)]
+    #[future(output)]
+    Then(IntoFuture<Fut2>),
+}
+```
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The trait name in a `derive` attribute can be adorned with generic parameters
+that specify the generics of the generated `impl` item:
+
+```rust
+#[derive(<T: Bound1, U: Bound2> Frob<T>)]
+struct Foo<U> {
+    // ...
+}
+```
+
+The derive macro for `Frob` is expected to generate an implementation item
+with these generic parameters:
+
+```rust
+impl<T: Bound1, U: Bound2> Frob<T> for Foo<U> {
+    // ...
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The syntax of an item in the `derive` attribute is extended to a subset of the
+language that can occur in a trait implementation item between the keywords
+`impl` and `for`:
+
+> _DeriveItem_ :\
+> &nbsp;&nbsp; _Generics_<sup>?</sup> _TypePath_
+
+The procedural macro can optionally support generic parameters to `derive` by
+defining an entry point annotated with the `proc_macro_derive_with_generics`
+attribute:
+
+```rust
+extern crate proc_macro;
+use proc_macro::TokenStream;
+
+#[proc_macro_derive_with_generics(Frob)]
+pub fn derive_frob_with_generics(
+    generics: TokenStream,
+    trait_args: Option<TokenStream>,
+    item: TokenStream,
+) -> TokenStream {
+    // ...
+}
+```
+
+Invoked in the example above, the function will receive the token stream of
+`<T: Bound1, U: Bound2>` as the first argument, a `Some` value with the token
+stream of `<T>` as the second argument, and the token stream with the
+`struct Foo` item as the third.
+
+If the compiler does not find a matching `proc_macro_derive_with_generics`
+symbol in the procedural macro crate that it has resolved for a `derive` item
+that features generics, an error is reported stating that the macro does not
+support generics. A plain old `derive` item can be processed with
+a function annotated as `proc_macro_derive_with_generics` if no function
+is annotated as `proc_macro_derive` for the same trait, otherwise the other
+function gets called.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This extension complicates the syntax of the `derive` attribute.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Extending `derive` this way, we can solve its current shortcomings and
+open it to more uses and experimentation. The proposed syntax should be
+familiar to the developers, as it forms a part of the syntax of the intended
+trait impl item.
+
+An [earlier proposal][rust-lang/rfcs#2353] to control generic bounds on
+derived items introduces two attributes used on the generic parameters of
+the type definition item, the whole item, or its fields. Using separate
+attributes, however, visually distances the declaration from its effect
+on the behavior on the `derive` attribute, and in many cases would be
+more verbose. It also splits the solution across multiple attributes, whereas
+the extended `derive` syntax proposed here is holistic, consistent with the
+syntax of the generated impl item to the extent of being a literal
+subsequence of it, and may allow further extension also in holistic ways.
+The extension proposed here is opted into by the macro authors if and when
+they wish to do so, while the solution proposed in RFC 2353 expects all
+macro authors to implement support for the new attributes "so that a consistent
+experience is maintained in the ecosystem".
+
+[rust-lang/rfcs#2353]: https://github.com/rust-lang/rfcs/pull/2353
+
+An alternative has been proposed in the pre-RFC discussion to enable custom
+bounds by trait-specific inert attributes. This has some disadvantages of
+the alternative above, furthermore, it bifurcates the solution into mostly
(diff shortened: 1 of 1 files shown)


nox:
So with this proposal, `Clone` can never allow opting out from bounds, right? Given new macro need to opt-in into this new system.

mzabaluev:
@nox The proc macro for `Clone` will be updated to allow unbounded (or differently bounded) parameters. The first code example of the RFC details how it would look like in code.

nox:
> @nox The derive macro for `Clone` will be updated to allow unbounded (or differently bounded) parameters. The first code example of the RFC details how it would look like in code.

Wait, so you write down the names of the type params for which you *don't* want a trait bound? That seems counter-intuitive. Let's say you have `Foo<T, U>`, what happens to the `U` trait bound if you write `#[derive(<T> Clone)]`?

nox:
And how do you handle trait bounds on field types without repeating those field types?

Lythenas:
TL;DR: I think this is a shortcoming of the derive implementation for `Clone` (and probably others) and not something we should have syntax for.

If I understand correctly the Problem you're trying to solve is that for

```rust
struct X<T> {
  x: PhantomData<T>
}
```

`#[derive(Clone)]` generates

```rust
impl<T> Clone for X
where
    T: Clone
{
  fn clone(&self) -> Self {
    Self {  x: x.clone() }
  }
}
```

where it should actually generate:

```rust
impl<T> Clone for X
where
    PhantomData<T>: Clone // this is implemented for every T
{
  fn clone(&self) -> Self {
    Self { x: x.clone() }
  }
}
```

I think this is more a problem of how the derive for Clone (and others) is implemented. It should generate the bound from the field types (`PhantomData<T>`) instead of the type parameters (`T`). Because it is actually calling `.clone()` on the type of the field (`PhantomData<T>`) and not the generic type parameter (`T`).

If you were to actually ignore the type parameter when deriving `Clone` what would you call for a field of type `T`. You still have to produce a value for every field when cloning. Even if it gets erased by the compiler in the case of `PhantomData<T>`.

It might be useful to have uniform way to give arguments to derives but I think the information you want to provide to the derive implementation is so varied that the way it's currently done is good (custom attributes).

mzabaluev:
@nox
> Let's say you have `Foo<T, U>`, what happens to the `U` trait bound if you write `#[derive(<T> Clone)]`?

That won't work; you need to list all of the parameters of the impl item:

```rust
#[derive(<T, U: Clone> Clone)]
struct Foo<T, U> {
    // ...
}
```

The idea is that the derive macro can paste these parameters verbatim after the `impl` keyword and expect the generated impl to work with the bounds given there. This may be more repetitive than field-specific directive attributes (though purely character for character, this way could still be more concise), but it has a "worse is better" virtue: it does not require from the code authors and readers to apply rules in their heads to figure out what parameter bounds will the generated impl item get.

> And how do you handle trait bounds on field types without repeating those field types?

The answer is that you use non-parameter types in the bounds _only if you have to_ (in the `where` clause, which is now officially part of the proposal). I don't expect this to be needed often. First off, a trait impl for a public type should never leak types of private fields which are not directly parameters. Second, with a typical recursively implemented, derivable trait, all fields must have an impl of this trait for their type, so you only need to fit the bounds that satisfy those field impls, it's not imperative to work the field types themselves into the outer impl's bounds.

mzabaluev:
@Lythenas 
> I think this is more a problem of how the derive for Clone (and others) is implemented. It should generate the bound from the field types (`PhantomData<T>`) instead of the type parameters (`T`). Because it is actually calling `.clone()` on the type of the field (`PhantomData<T>`) and not the generic type parameter (`T`).

I think you are conflating the field type with the generic bound that needs to be satisfied so that `.clone()` works on that field. `derive` should not automatically expose field types in signatures of generated impls, certainly not for private fields. But derivation of the minimal necessary bounds that satisfy the field's impl is not something that a proc macro can do in current Rust, it should be left to the macro user to define. After all, the whole feature is meant to reduce boilerplate, but assignment of impl bounds may sometimes be a design decision rather than a completely mechanical task.

nox:
> The answer is that you apply bounds to non-parameter types _only if you have to_ (in the `where` clause, which is now officially part of the proposal). I don't expect this to be needed often. First off, a trait impl for a public type should never leak types of private fields which are not directly parameters. Second, with a typical recursively implemented, derivable trait, all fields must have an impl of this trait for their type, so you only need to fit the bounds that satisfy those field impls, it's not imperative to work the field types themselves into the outer impl's bounds.

There are *many* types in Stylo, Servo's styling system, where we can't just have general `T: Foo` bounds on type params and must put those bounds on field types instead. Repeating the field types *and* the type parameters for each derive just does not scale.

mzabaluev:
> There are _many_ types in Stylo, Servo's styling system, where we can't just have general `T: Foo` bounds on type params and must put those bounds on field types instead.

I'm interested to look at the examples and the design considerations that led to a proliferation of non-uniform impls for derivable traits.

nox:
> I'm interested to look at the examples and the design considerations that led to a proliferation of non-uniform impls for derivable traits.

We have generic types that implement a given trait (let's say `ToCss`) only for some combination of type parameters, we then reuse one of those types (let's say `Position<T>`) in another generic structure for which we want to derive `ToCss`. If the derive system generates a `T: ToCss` bound, the derived code won't compile.

mzabaluev:
> We have generic types that implement a given trait (let's say `ToCss`) only for some combination of type parameters

That sounds like a major inconvenience for developing any kind of regular trait coverage. But yes, for this kind of design the bounds would need to be verbose.

nox:
> That sounds like a major inconvenience for developing any kind of regular trait coverage.

Things are this way because that's the most natural way for them to be, and all of our `derive` stuff has something similar to `derive_field_bounds` etc like in the other RFC.

JelteF:
> I think this is more a problem of how the derive for Clone (and others) is implemented. It should generate the bound from the field types (PhantomData<T>) instead of the type parameters (T). Because it is actually calling .clone() on the type of the field (PhantomData<T>) and not the generic type parameter (T).

I think this is the only correct and most elegant solution to this problem, not having to enter bounds in the derive itself. This solution was actually implemented in my `derive_more` crate for the `Display` derive there: https://github.com/JelteF/derive_more/pull/95 (see issue for more discussion: https://github.com/JelteF/derive_more/issues/93)

I think it might actually be possible to implement this without breaking backwards compatibility for the current derives, because it should only fix derives that are generating invalid code now.

nox:
@JelteF If the field types are private and the type for which we derive stuff is public, last I checked it will not pass privacy checks.

JelteF:
@nox changing the outer type to `pub` in the tests in my library I indeed get the following warning, but no error:
```
warning: private type `generic::complex_type_field_enumerator::Struct<T>` in public interface (error E0446)
   --> tests/display.rs:283:22
    |
283 |             #[derive(Display)]
    |                      ^^^^^^^
    |
    = note: `#[warn(private_in_public)]` on by default
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #34537 <https://github.com/rust-lang/rust/issues/34537>
``` 

Maybe this should be a reason to reconsider phasing this behaviour out (or modifying it so that this type of case is allowed). It seems to me that this should be the way you would want to implement an implementation like this. As far as I can tell there's no reason for this type of trait implementation to be disallowed.

JelteF:
Actually, in the tracking issue in that for that lint there is the following comment that basically describes this exact issue:
https://github.com/rust-lang/rust/issues/34537#issuecomment-468161091
It also says that that code should be fixed once https://github.com/rust-lang/rust/issues/48054 is implemented, so it sounds like this code should (eventually) pass the new privacy rules too.

mzabaluev:
Private-in-public is only part of why leaking field types in bounds is problematic. Even if the type of a private field is public (as e.g. `PhantomData` is), leaking it in an impl bound for the containing structure breaks encapsulation, as you can no longer freely change that field's type without affecting the public signature of the impl.

JelteF:
Do you mean that someone can implement Clone for PhantomData<TheirType>. Without implementing clone for TheirType?
And that changing it the field from PhantomData<T> to T would result in API breakage.

mzabaluev:
> We have generic types that implement a given trait (let's say `ToCss`) only for some combination of type parameters

I see only six instances in [this list](https://doc.servo.org/style_traits/values/trait.ToCss.html#implementors) where non-parameter bounds were needed.

Let's take a simple one:

```rust
impl<H, V, NonNegativeLengthPercentage> ToCss for Circle<H, V, NonNegativeLengthPercentage> where
    GenericPosition<H, V>: ToCss,
    NonNegativeLengthPercentage: ToCss + PartialEq,
```

Why can't there be a generic `impl<H: ToCss, V: ToCss> ToCss for GenericPosition<H, V>`? Where are the specific impls defined?

mzabaluev:
@JelteF `Clone` [is implemented](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#impl-Clone) for `PhantomData` regardless of the parameter type.

What I mean is that changing this:

```rust
use some_public_api::Bar;

pub struct Foo<T> {
    inner: Bar<T>
}

impl<T> Clone for Foo<T> where Bar<T>: Clone {
    // ...
}
```

to this:

```rust
use some_public_api::Baz;

pub struct Foo<T> {
    inner: Baz<T>
}

impl<T> Clone for Foo<T> where Baz<T>: Clone {
    // ...
}
```

is a breaking change for users of `Foo`, and the bounds like that are best avoided because they expose an implementation detail of `Foo`.

mzabaluev:
> I see only six instances in [this list](https://doc.servo.org/style_traits/values/trait.ToCss.html#implementors) where non-parameter bounds were needed.

As further analyzed in https://github.com/servo/servo/issues/24733, of these only two have a reason, albeit questionable, for having a non-parameter bound, and a third is caused by those two.

mzabaluev:
A legitimate _reason_ for having bounds on public field types might be maintainability: like in the Servo example, you are certain, due to the design constraints on the subject functionality, that a certain struct has to have a field of a certain public type, so it makes sense to define the bounds on the struct's impl through those of the field.

This looks like a design decision to me, and as such, it deserves some more explicit syntax.

JelteF:
My what I meant with my question was: in what case does it actually break
when these bounds change? For instance changing from Arc to Box doesn't,
because they both implement Clone for all T. And in this breaking case,
what Clone implementation would you want to have?

I'm asking, because in the end this where statement is the most minimal
automatic bound you can give. It sounds like on one hand you want the
minimal viable bounds, but on the other you hand you don't. (the first case
is explained in the RFC, but the second not really)

mzabaluev:
@nox 
> and all of our `derive` stuff has something similar to `derive_field_bounds` etc like in the other RFC.

It should be noted that any custom helper attributes to further reduce boilerplate, or even a systematic solution like #2353, are not incompatible with this proposal. It only makes them needed in fewer cases.

mzabaluev:
> My what I meant with my question was: in what case does it actually break when these bounds change?

Consider a type in another crate that embeds `Foo<T>` and implements `Clone`. It has to specify its own bounds for `Clone` so that the bound on `impl<T> Clone for Foo<T> ...` is satisfied. Before the change, those bounds have to include `Bar<T>: Clone`, but nothing required that impl to satisfy `Baz<T>: Clone`, so after the change, it no longer compiles.

JelteF:
> Consider a type in another crate that embeds `Foo<T>` and implements `Clone`. It has to specify its own bounds for `Clone` so that the bound on `impl<T> Clone for Foo<T> ...` is satisfied. Before the change, those bounds have to include `Bar<T>: Clone`, but nothing required that impl to satisfy `Baz<T>: Clone`, so after the change, it no longer compiles.

Just to be clear, you mean something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f13f8658e57d192d3c4e74a4514d3c61

My question is, what Clone implementation would you want in this case then?

mzabaluev:
> Just to be clear, you mean something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f13f8658e57d192d3c4e74a4514d3c61

I'd prefer a less contrived example because there is no good reason for `MyType` not to implement `Clone` here.

JelteF:
> I'd prefer a less contrived example because there is no good reason for `MyType` not to implement `Clone` here.

This is just your example written from your comment written out as far as I can tell. So if you think it's too contrived (I agree it is), that's because your original example already was. I cannot think of a non contrived example, that's why I was asking you for it.

mzabaluev:
> This is just your example written from your comment written out as far as I can tell. So if you think it's too contrived (I agree it is), that's because your original example already was.

My example did not have this irregular trait impl:

```rust
impl Clone for Bar<MyType> {
    fn clone(&self) -> Self {
        Self {
            data: MyType{}
        }
    }
}
```

Sure this forces a bound on `Bar<T>: Clone` to be able to cover `MyType`, but is there a non-contrived reason why `MyType` should not implement `Clone` itself and therefore be covered by the derived `impl<T: Clone> Clone for Bar<T>`?

JelteF:
I'm just trying to understand your example, but afaict you just effectively
said that it's incomplete. So I don't know the answer to your question.
Seems like you should think of that given it's your example

mzabaluev:
@JelteF As far as I understand, your addition is meant to justify a bound on a private field type. Unfortunately, it does not go far enough in showing what real purpose would this design be useful for, because you can just derive `Clone` for `MyType` instead: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a7c2026817494c6376c47b435cac0650

PR #2814, opened 2019-11-12
Adding is_zero() to core::time::Duration

InTVyWOC:
This PR is for adding new function `is_zero()` to `core::time::Duration`.

Related issue: <https://github.com/rust-lang/rfcs/issues/2809>

### DIFF
--- a/text/0000-duration-is-zero.md
+++ b/text/0000-duration-is-zero.md
@@ -1,0 +1,121 @@
+- Feature Name: `Duration::is_zero()`
+- Start Date: 2019-11-12
+- RFC PR: N/A
+- RFC Issue: [rust-lang/rfcs#2809](https://github.com/rust-lang/rfcs/issues/2809)
+
+# Summary
+[summary]: #summary
+
+Adding new function `is_zero(&self) -> bool` to `core::time::Duration`.
+
+# Motivation
+[motivation]: #motivation
+
+- Why are we doing this?
+
+    Personally I found many times the need to compare a duration to zero. What I would do is:
+    
+    ```rust
+    if some_duration == Duration::from_secs(0) {
+        ...
+    }
+    ```
+
+- What use cases does it support?
+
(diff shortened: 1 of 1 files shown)


jhpratt:
Sort of prior art, albeit in progress and unreleased: I'm rewriting the time crate, and included [this exact method](https://time-rs.github.io/time/time/struct.Duration.html#method.is_zero). It is slightly different in that the implementation allows for both positive and negative durations, and there are corresponding methods.

I fully support, though! Even though I don't think an RFC is necessary for something like this :)

joshtriplett:
A couple of notes:

- I would not in general expect `== time::ZERO` to generate more code than the `is_zero` method, especially if `==` gets inlined. I do think that `.is_zero()` seems more readable, though, and I still think we should have an `.is_zero()` method, but I don't think we need it for performance reasons specifically.

- Another alternative: we could have a constant function `Duration::zero()` that generates a zero duration. Worth mentioning as an alternative, I think.

LukasKalbertodt:
:+1: On adding `is_zero`. Seems like a small and useful addition to me. 

@joshtriplett mentioned `Duration::zero()` as alternative. I would like to propose it as a potential addition. I.e. let's just add `Duration::is_zero` and `Duration::zero`. 

And as mentioned in the issue you opened, it would have probably been fine to directly create a PR on the main repo. It's a really small change.

InTVyWOC:
@LukasKalbertodt 

Thanks. I've updated the file.

I will create new PR if this RFC would be changed to "final comment period" state (like in guideline). It would prevent a lot of changes from time to time, while this RFC is being updated.

jonas-schievink:
This really doesn't have to go through the full RFC process, a PR on rust-lang/rust is just fine. Closing in favor of that.

jhpratt:
@jonas-schievink #2809 can be closed as well, then, I believe.

PR #2816, opened 2019-11-13
FnPinMut: fully general stackless semicoroutines

CAD97:
cc @samsartor 

[Rendered](https://github.com/CAD97/rust-rfcs/blob/fn-pin-mut/text/0000-fn-pin-put.md)

I think the building blocks for putting this together as a proper proposal are here; let's put them together!


CAD97:
ugh this was supposed to be a self-pr github plz (sorry)

PR #2818, opened 2019-11-15
A macro to get the current function name.

Lokathor:
[rendered](https://github.com/Lokathor/rfcs/blob/fn-name/text/0000-fn_name_macro.md)

### DIFF
--- a/text/0000-fn_name_macro.md
+++ b/text/0000-fn_name_macro.md
@@ -40,7 +40,7 @@ The exact text is considered "debug" information and not subject to Rust's stabi
 
 Previous discussion was within an issue of the RFCs repo: https://github.com/rust-lang/rfcs/issues/1743
 
-In summary, this is _extremely_ useful for debug purposes, and has been highly desired for nearly two years.
+In summary, this is useful for debug purposes, and has been desired for nearly two years.
 
 Alternative: we could call it `fn_name!` instead.
 
--- a/text/0000-fn_name_macro.md
+++ b/text/0000-fn_name_macro.md
@@ -1,0 +1,61 @@
+- Feature Name: fn_name_macro
+- Start Date: 2019-11-14
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+<!-- https://github.com/rust-lang/rfcs/issues/1743 -->
+
+# Summary
+[summary]: #summary
+
+This RFC adds an additional macro, `function!`, to the `core` crate. When invoked, the macro expands to the name of the function that contains the call site.
+
+# Motivation
+[motivation]: #motivation
+
+This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `col!` and `module_path!`.
+
+For most people the name of the function is a much more immediate way to understand the context of a message than file name and line number.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For debug information about what's happening within the program there's several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Use of the `function!` macro expands to the compiler's internal name for the function. This will generally be the name that the user wrote into the file but in the case of a closure or similar it will be something like the function's name with a unique suffix.
+
+The exact text is considered "debug" information and not subject to Rust's stability guarantee.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* Doing this adds another macro to `core`.
+* This macro in particular cannot be implemented in Rust itself, it requires special support from `rustc`, as well as from all potential alternative compilers for the Rust language.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Previous discussion was within an issue of the RFCs repo: https://github.com/rust-lang/rfcs/issues/1743
+
+In summary, this is _extremely_ useful for debug purposes, and has been highly desired for nearly two years.
+
+Alternative: we could call it `fn_name!` instead.
+
+# Prior art
+[prior-art]: #prior-art
+
+* C99 has a `__func__` pre-processor macro that expands to the current funciton name. [link](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2004/n1642.html)
+* C# has a `nameof` operator which can be used on functions
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+None, other than perhaps bikeshed on the name.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+None. This is a small change that's pretty open and shut.
--- a/text/0000-fn_name_macro.md
+++ b/text/0000-fn_name_macro.md
@@ -13,7 +13,7 @@ This RFC adds an additional macro, `function!`, to the `core` crate. When invoke
 # Motivation
 [motivation]: #motivation
 
-This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `col!` and `module_path!`.
+This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `column!` and `module_path!`.
 
 For most people the name of the function is a much more immediate way to understand the context of a message than file name and line number.
 
--- a/text/0000-fn_name_macro.md
+++ b/text/0000-fn_name_macro.md
@@ -20,7 +20,7 @@ For most people, the name of the function is a much more immediate way to unders
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-For debug information about what's happening within the program there's several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
+For debug information about what's happening within the program there are several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
--- a/text/0000-fn_name_macro.md
+++ b/text/0000-fn_name_macro.md
@@ -15,12 +15,24 @@ This RFC adds an additional macro, `function!`, to the `core` crate. When invoke
 
 This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `column!` and `module_path!`.
 
-For most people, the name of the function is a much more immediate way to understand the context of a message than file name and line number.
+For many people, the name of the function is a much more immediate way to understand the context of a message than file name and line number. Often when discussing a log message or panic message with others (in a chat, forum, github issue, etc) the source code might not be immediately available.
+
+As an example, it's a lot easier to understand what's going on when a message comes from "f32::floor" instead of from "src/libcore/f32.rs:45".
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-For debug information about what's happening within the program there are several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
+For debug information about what's happening within the program there are several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function.
+
+```rust
+// this prints "entering my_func" before doing the work
+fn my_func() {
+  println!("entering {}", function!());
+  // work here
+}
+```
+
+If used outside of a function it causes a compilation error.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -44,18 +56,35 @@ In summary, this is useful for debug purposes, and has been desired for nearly t
 
 Alternative: we could call it `fn_name!` instead.
 
+Alternative (centril): We could introduce a much more general `item_path!()` macro that is useful in non-function contexts.
+
+```rust
+mod foo { // item_path!() ==> foo
+    const X: &str = item_path!(); // foo::X
+    struct A;
+    impl A {
+        const Y: &str = item_path!(); // foo::A::Y
+        fn fun() {
+            // item_path!() ==> foo::A::fun
+        }
+    }
+     type T = Stuff<{item_path!()}> // Stuff<"foo::T">
+}
+```
+
 # Prior art
 [prior-art]: #prior-art
 
 * C99 has a `__func__` pre-processor macro that expands to the current funciton name. [link](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2004/n1642.html)
-* C# has a `nameof` operator which can be used on functions
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-None, other than perhaps bikeshed on the name.
+* Do type methods and type associated functions include their type in the printed name? eg: `Vec::new`
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
-None. This is a small change that's pretty open and shut.
+None that are specifically improvements of this macro.
+
+We could of course add additional debugging macros.
--- a/text/0000-fn_name_macro.md
+++ b/text/0000-fn_name_macro.md
@@ -15,7 +15,7 @@ This RFC adds an additional macro, `function!`, to the `core` crate. When invoke
 
 This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `column!` and `module_path!`.
 
-For most people the name of the function is a much more immediate way to understand the context of a message than file name and line number.
+For most people, the name of the function is a much more immediate way to understand the context of a message than file name and line number.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation


text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For debug information about what's happening within the program there's several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Use of the `function!` macro expands to the compiler's internal name for the function. This will generally be the name that the user wrote into the file but in the case of a closure or similar it will be something like the function's name with a unique suffix.
```
kennytm: What about a closure in a static?

```rust
static F: &(dyn Fn() -> &'static str + Sync) = &|| "1";
```

(the fully-qualified name is currently `playground::F::{{closure}}`)

Lokathor: @eddyb know this one (i hope)

comex:
👍

Pie-in-the-sky: It would be nice to have a more flexible feature where you could have a function automatically receive its *caller's* file/line/function name/etc., like C++ recently added with `std::source_location` (although the design C++ picked is terrible).  But of course there's no need for that to block adding `function!` to the existing set of macros.

Lokathor:
That actually was requested earlier today: https://github.com/rust-lang/rfcs/issues/2815

CryZe:
> It would be nice to have a more flexible feature where you could have a function automatically receive its caller's file/line/function name/etc.

@comex This is already being implemented https://github.com/rust-lang/rust/pull/65881

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+[motivation]: #motivation
+
+This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `column!` and `module_path!`.
+
+For most people the name of the function is a much more immediate way to understand the context of a message than file name and line number.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For debug information about what's happening within the program there's several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
```
pcpthm: I think returning an `Option<&'static str>` or some special name like `"(top-level)"` is more useful than a compilation error. It is surprising when some macro cannot be used outside a function.

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+[summary]: #summary
+
+This RFC adds an additional macro, `function!`, to the `core` crate. When invoked, the macro expands to the name of the function that contains the call site.
+
+# Motivation
+[motivation]: #motivation
+
+This is a useful extension of Rust's existing debug reporting: `file!`, `line!`, `column!` and `module_path!`.
+
+For most people the name of the function is a much more immediate way to understand the context of a message than file name and line number.
```
Centril:
```suggestion
For most people, the name of the function is a much more immediate way to understand the context of a message than file name and line number.
```

text/0000-fn_name_macro.md
Centril:
```suggestion
For debug information about what's happening within the program there are several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
```

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For debug information about what's happening within the program there's several useful macros that you might use. One of them is the `function!` macro, which expands to the name of the current function. If used outside of a function it causes a compilation error.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
```
Centril: This section needs some elaboration to consider the various cases (e.g. like the one @kennytm raised) and give examples of code and what the result is.

Also, looking at the implementation of `module_path!` in `libsyntax_ext` https://github.com/rust-lang/rust/blob/9e8c4e6fb1c952048fb823e59f4c9c6487bf9a58/src/libsyntax_ext/source_util.rs#L65-L73 it looks like the information is readily available. In this case however, the information is not available in `cx` or reachable fields. Some elaboration on the implementation would be good. cc @petrochenkov since they are the most likely T-compiler reviewer for this RFC. Also, this probably has little to do with "debuginfo" -- that's a different part of the compiler that comes into play much later in the compilation process.

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+
+Use of the `function!` macro expands to the compiler's internal name for the function. This will generally be the name that the user wrote into the file but in the case of a closure or similar it will be something like the function's name with a unique suffix.
```
pcpthm: An alternative is to return the name of the syntactic first ancestor function containing the macro expression so
```rust
fn process_items(items: &[Item]) {
    process_items.iter().for_each(|item| {
        dbg_with_function_name!(item);
        process(item);
    });
}
```
prints `[process_items:xxx]: Item { ... }` regardless of whether `for` is used or iterator is used.

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+
+* Doing this adds another macro to `core`.
+* This macro in particular cannot be implemented in Rust itself, it requires special support from `rustc`, as well as from all potential alternative compilers for the Rust language.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Previous discussion was within an issue of the RFCs repo: https://github.com/rust-lang/rfcs/issues/1743
+
+In summary, this is _extremely_ useful for debug purposes, and has been highly desired for nearly two years.
```
Centril:
```suggestion
In summary, this is useful for debug purposes, and has been desired for nearly two years.
```

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Use of the `function!` macro expands to the compiler's internal name for the function. This will generally be the name that the user wrote into the file but in the case of a closure or similar it will be something like the function's name with a unique suffix.
+
+The exact text is considered "debug" information and not subject to Rust's stability guarantee.
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
Centril: It's another blow to incremental compilation (well it doesn't get worse than `column!()`, but this adds more...).

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,61 @@
...
+
+<!-- https://github.com/rust-lang/rfcs/issues/1743 -->
+
+# Summary
+[summary]: #summary
+
+This RFC adds an additional macro, `function!`, to the `core` crate. When invoked, the macro expands to the name of the function that contains the call site.
+
+# Motivation
+[motivation]: #motivation
```
Centril: Some elaboration on the common use case would be good. In the issue you mention:
> I wouldn't normally think to pair func_name!() with line!() and file!() because usually when i want it I want "emergency println debugging" and i'm just throwing in a few printers and i delete them once i've sorted out what's wrong.

and I'm wondering whether you expect the macro to be used in e.g. `dbg!(...)` (which does print out `file!()` and `line!()`).

text/0000-fn_name_macro.md
Centril: Also I'm not sure I agree with this, especially not when the claim is "much more". In particular, the `dbg!()` output is structured in such a way that e.g. in VSCode's integrated terminal, the output becomes a clickable link that immediately takes you to the source location. That seems hard to beat...?

Centril:
Who is going to implement this RFC by the way? I think we should have someone available to do that, with mentoring if necessary (if so a mentor is required, e.g. @petrochenkov), before accepting the RFC.

text/0000-fn_name_macro.md
mathstuf: Another use case I'll be using it for: testing. In many of the tests I've written (and this reflects the kinds of crates I've been writing too), temporary directories and other resources are named according to the name of the test function. Not having to keep them in sync manually would be great.

Centril: Not sure why you wrote this here, but your use case sounds like you need guarantees wrt. the output (or else your tests might suddenly start failing on an upgrade) and this RFC explicitly disclaims such guarantees.

da-x:
I have a reference implementation of this feature.

https://github.com/da-x/rust/pull/new/function-macro-1.38.0

It working, but would probably need some changes to conform to the RFC.

text/0000-fn_name_macro.md
mathstuf: Maybe I should have put it on the section header line. It's more that Github only allows a single conversation per source line.

I don't need stability guarantees, just uniqueness (it also helps to disambiguate log lines in the face of rust-lang/rust#42474).

Centril:
> https://github.com/da-x/rust/tree/function-macro-1.38.0

Hmm; it feels fairly invasive to add `&Option<ast::Path>` in so many places. Looking over that, I think something like `item_path!()` which is the concatenation of the `module_path!()` as well as e.g. `StructName::function_name` would be less invasive and strictly more general.

da-x:
@Centril I agree regarding having `&Option<ast::Path>` added in the internal API, and so I've originally also had [a change](https://github.com/da-x/rust/commit/590f73f7a1f45b3f3b0da8c3a51291f868d27202) to extend `ExtCtxt` with a new field instead. However 1) it's for a much earlier `rustc` version and 2) it uses `Rc` as in the new `Option<Rc<ast::Path>>` field which I opt to avoid. Need to see how it fits into the scheme of things now.

comex:
> @comex This is already being implemented [rust-lang/rust#65881](https://github.com/rust-lang/rust/pull/65881)

What!  Now that's the kind of pleasant surprise that comes rarely.

Hmm... but if that *is* already being implemented, might it make sense to not add a new macro, but instead just add the function name to `Location`?

It seems like `Location::caller()` is already enough to get you the current location – making it equivalent to `file!()`/`line!()`/etc. – if the current function is *not* `#[track_caller]`.  But we should probably add something that always gets the current location even if the current function is `#[track_caller]`; I'll comment in the tracking issue.

On the other hand, it may be worth adding a macro just to avoid confusing people who see the macros but don't know about `Location`.

text/0000-fn_name_macro.md
joshtriplett: Definitely not an Option, that would make it much more painful to use.

text/0000-fn_name_macro.md
joshtriplett: Seems much less likely to cause such problems than column.

Centril: Yeah (I said as much), but it should still be noted in the text to provide a more full examination of the trade-offs.

text/0000-fn_name_macro.md
Centril: Well if we generalize this to items rather than to functions then there is always an enclosing item wherever this is used except directly in top level modules so erroring becomes much less common.

text/0000-fn_name_macro.md
Lokathor: That assumes that you are looking at the message in the VS Code integrated terminal. One of my machines doesn't even run VS Code.

Often I see panic or log messages that are pasted into Discord/Zulip/Riot/Github, where you also cannot click the `file:line` text.

text/0000-fn_name_macro.md
Lokathor: Good point. I will add an example usage to the guide level explanation.

text/0000-fn_name_macro.md
Lokathor: I'm going to add centril's "item_path!" idea, which covers this case.

text/0000-fn_name_macro.md
Lokathor: How is incremental compilation particularly negatively affected? It seems like this would be a lot _less_ sensitive to file changes than `line!` and `column!` are.

Lokathor:
Updated based on current feedback.

I'm extremely convinced by the `item_path!()` idea, to the point where if Centril had simply suggested it _yesterday_ instead of today I wouldn't have even opened this RFC.

text/0000-fn_name_macro.md
Centril: Yeah it is less sensitive than `line!()` certainly. It's probably the case that `function!()` or `item_path!()` is as bad as `module_path!()` is; only if you rename the item or parent module does the macro output change, so there's probably not a lot of impact on incremental after all.

text/0000-fn_name_macro.md
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+     type T = Stuff<{item_path!()}> // Stuff<"foo::T">
+}
+```
+
+# Prior art
+[prior-art]: #prior-art
+
+* C99 has a `__func__` pre-processor macro that expands to the current funciton name. [link](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2004/n1642.html)
```
jonas-schievink: (Luckily) This is not a pre-processor macro, just a predefined identifier provided by the C compiler

joshtriplett:
I'm in favor of `item_path!` as well, as a generalization of this that seems quite useful.

Lokathor:
This time you write it Josh ;3

nagisa:
I wonder if specifying this in terms of `type_name::<$thisfunction>()` is better and whether this could be more composable and general if we instead added a macro to produce `$thisfunction` rather than `type_name::<$thisfunction>()`.

Centril:
@nagisa That would presumably need to account for substitutions to polymorphic functions as well as considering nested `fn`s in `fn`s, no? To me, `item_path!` seems more general.

eddyb:
If we're bikeshedding the name, something like `def_path!` feels a bit more general.
(although both it and `item_path!` feel a bit weird wrt closures)

Diggsey:
Also, `scope!`/`scope_name!`/`scope_path!`/`path!` are options for the more general version.

Silentdoer:
this is in wip?

Lokathor:
No, this is pretty dead I'd have to say.

The idea isn't fundamentally bad, it's just we all got busy with other things.

I'll close this for now.

ArekPiekarz:
In the meantime we can use `function_name!()` from stdext crate - https://docs.rs/stdext/0.2.1/stdext/macro.function_name.html

brainstorm:
@ArekPiekarz Is the latest version (0.3.0) still working for that purpose? I'm getting the following while trying it out:

```rust
% cargo build
   Compiling stdext v0.3.0
error[E0391]: cycle detected when caching mir of `num::integer::<impl at /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:274:9: 440:10>::BITS` for CTFE
   --> /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:277:13
    |
277 |             const BITS: u32 = Self::BITS;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
445 | impl_integer!(u8, u16, u32, u64, u128);
    | --------------------------------------- in this macro invocation
    |
    = note: ...which requires normalizing `<u8 as num::integer::Integer>::BITS`...
note: ...which requires simplifying constant for the type system `num::integer::<impl at /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:274:9: 440:10>::BITS`...
   --> /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:277:13
    |
277 |             const BITS: u32 = Self::BITS;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
445 | impl_integer!(u8, u16, u32, u64, u128);
    | --------------------------------------- in this macro invocation
note: ...which requires simplifying constant for the type system `num::integer::<impl at /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:274:9: 440:10>::BITS`...
   --> /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:277:13
    |
277 |             const BITS: u32 = Self::BITS;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
445 | impl_integer!(u8, u16, u32, u64, u128);
    | --------------------------------------- in this macro invocation
note: ...which requires const-evaluating + checking `num::integer::<impl at /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:274:9: 440:10>::BITS`...
   --> /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:277:13
    |
277 |             const BITS: u32 = Self::BITS;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
445 | impl_integer!(u8, u16, u32, u64, u128);
    | --------------------------------------- in this macro invocation
    = note: ...which again requires caching mir of `num::integer::<impl at /Users/rvalls/.cargo/registry/src/github.com-1ecc6299db9ec823/stdext-0.3.0/src/num/integer.rs:274:9: 440:10>::BITS` for CTFE, completing the cycle
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error

For more information about this error, try `rustc --explain E0391`.
error: could not compile `stdext`

To learn more, run the command again with --verbose.
```

ArekPiekarz:
@brainstorm I'm not the author of that crate, but I just tested that it still works in a simple program. Can you share code which exhibits the problem?

brainstorm:
@ArekPiekarz I don't have much time to cook a proper small reproducer like I used to in https://github.com/brainstorm/dotenv-codegen-rustc-hang, but here's a 2 line start if you want to have a crack at it: 

```shell
git clone https://github.com/umccr/aws-benchmarks && cd aws-benchmarks
git checkout 560e531b9bd263f73bbf547eb27f7e2485461c3c && cargo build
```

... maybe some interaction of the other crates I'm pulling in? Don't know 🤷🏻‍♂️

ArekPiekarz:
@brainstorm I have built your project with your instructions without any errors. Here is the log - https://pastebin.com/m86af0pz
I'm on Manjaro 21.0.7 with Rust 1.53.

brainstorm:
Yep, that was it! I was running with `rustc 1.52.1 (9bc8c42bb 2021-05-09)`, then did rustup update to 1.53 and everything worked, thx!... `stdext` should  have a MSRV to that, I'll notify upstream accordingly.

AsteriskZuo:
The lack of official support is a bit of a shame.

silverweed:
So...has Rust just given up on this? Are we doomed to depend on a separate crate for this trivial functionality as well?

PR #2823, opened 2019-11-28
adding RFC process-handle-for-async

frehberg:
[Rendered](https://github.com/frehberg/rfcs/blob/process-handle/text/0000-process-handle-for-async.md)

### DIFF
--- a/text/0000-process-handle-for-async.md
+++ b/text/0000-process-handle-for-async.md
@@ -1,0 +1,101 @@
+- Feature Name: `process_handle_for_async`
+- Start Date: 2019-11-28
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Introduce process descriptor handles for each process in `'std::process'` for async polling of process-termination.
+
+# Motivation
+[motivation]: #motivation?
+
+Currently `std::process` is managing processes (on Posix/Linux systems) either in synchronous way using the wait-function-family, or asynchronously via signal-handlers (SIGCHLD). 
+
+The problem is that such signals may be delivered to any thread running within the parent-process. So, dealing with the termination of child-processes in async manner will require a shared collection containing the process-status of all child-processes.
+
+This RFC proposes a portable alternative to signal-handling to deal with the termination of child-processes in async manner, and introducing strict ownership for child-processe using a process-handle (aka process descriptor)
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Depending on the underlying operating system, sub-processes may be associated with a process-id (pid) and most oeprating systems provide  so called process-handles (process descriptors). Such handles may be used to perform async polling or sending signals to the corresponding process. Such process handles may be used to establish strict opwnership for child-processes in Rust.
+
+The process-handle feature shall permita portable API for various operating systems. The following listing shows the process-handle features being available for different operating systems, and might be the base for a generalized Rust-API in `std::process`.
+
+- As with Windows, the fucntion `CreateProcess` provides the HANDLE for the newly created child-process.
+- As with OpenBSD Unix provides process handles 
+- As with FreeBSD Unix the function `libc::pdfork(..)` provides a process descriptor for the newly forked child-process.
+- As with Linux, the `pidfd`-familiy provides funtionality to convert a pid int a file-descriptor (merged into release Linux-5.4)
+- Older Linux-releases or Posix-Systems may use the `forkfd`-concept to establsih a descriptor/handle for a child-process
+ 
+As for the latter case when the underlying OS does not provide a suitable naticve process-handle, the forkfd concept may be used. This is based on so called death-pipes: An anonymous pipe is established between parent-process (RX end) and the child-process (TX end). Here the parent-process may use RX end to poll for read-events (file descriptor) in async manner, for example using a future.
+Now, if the child-process terminates the TX end is being destroyed and the RX-end in the parent process will receive and EOF read-event.
+
+All of the listed process-handle flavors share the common concept to perform async polling for process-termination and might be the base for a generalized process-handle concpet in Rust `std::process`, leveraging a strong ownership-concpet for child-processes. The handling of signal SIGCHLD and sharing of child-process meta-data would no longer be required.
+
+Pros:
+
+- permits strict ownership and async-handling of termination of child-processes
+- no shared collection required to deal with SIGCHLD signal and exit-status of child-processes
+- on Posix/VxWorks etal the process-handle will be a file-descriptor, and can be handled by async/await framework.
+- on Windows the process-handle will be the HANDLE-object provided by function CreateProcessA(..).
+
+Cons:
+
+- on Posix/VxWorks etal each child-process will cause an additional file-descriptor in parent process and a non-referenced file-descriptor in child-process.
+- legacy signal-handlers dealing with SIGCHLD might consume the exit-status before the owning async task got a chance to read the exit-status, therefore the SIGCHLD signal-handling should be disabled in the parent-process, or just be used for legacy code.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The following patch is a first sketch of the concept (under construction)
+Branch https://github.com/frehberg/rust/tree/process_handle
+Patch frehberg/rust@337690e
+
+The API impl of `std::process::Process` shall be extended to provide access to the handle for the corresponding child-process. The ProcHandle API shall abstract from underlying native process-handle feature.
+
+```rust
+// Module std::process
+
+impl Process {
+...
+ /// Provides access to the underlying process handle for async polling for child-process-termination
+ pub fn handle(&self) ->  &ProcHandle;
+...
+}
+```
+This ProcHandle shall provide access to the underlying native process-handle, permitting integration into the async polling framework using futures and for example tokio.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+In case the underlying operating system does not provide a native concept for process handles, this functionality may introduce an overhead per child-process.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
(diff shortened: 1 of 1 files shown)


sfackler:
How does this differ from the `AsRawHandle` trait for Windows and `id` method for Unix? https://doc.rust-lang.org/std/os/windows/io/trait.AsRawHandle.html

frehberg:
@sfackler  Windows-Handles are the perfect way, the Unix-PIDs are a nightmare, citing from tokio-process https://github.com/alexcrichton/tokio-process/blob/master/src/unix/mod.rs

`//! Unix handling of child processes
//!
//! Right now the only "fancy" thing about this is how we implement the
//! `Future` implementation on `Child` to get the exit status. Unix offers
//! no way to register a child with epoll, and the only real way to get a
//! notification when a process exits is the SIGCHLD signal.
//!
//! Signal handling in general is *super* hairy and complicated, and it's even
//! more complicated here with the fact that signals are coalesced, so we may
//! not get a SIGCHLD-per-child.
//!
//! Our best approximation here is to check *all spawned processes* for all
//! SIGCHLD signals received. To do that we create a `Signal`, implemented in
//! the `tokio-signal` crate, which is a stream over signals being received.
//!
//! Later when we poll the process's exit status we simply check to see if a
//! SIGCHLD has happened since we last checked, and while that returns "yes" we
//! keep trying.
//!
//! Note that this means that this isn't really scalable, but then again
//! processes in general aren't scalable (e.g. millions) so it shouldn't be that
//! bad in theory...`

This breaks the ownership concept and relies heavily on signal-hander setups. So, instead of using the hairy signal concept of legacy Unix, Rust should rely on the newer Unix APIs for process handling, such as pidfd, pdfork, anf forkfd, providing now a similar functionality as Windows-Handle. QProcess demonstrates an API, being portable for all kind of OS.

joshtriplett:
I'd love to see this change, but I don't know if we can reasonably handle the non-pidfd case or the POSIX case without placing requirements on the caller that we don't want to place.

@thiagomacieira, can you weigh in regarding the Qt code for the non-pidfd compatibility case? We talked about the requirements the self-pipe and SIGCHLD approach had, to avoid disrupting any other code in the same program (given all the issues with signal handling).

thiagomacieira:
The code I have at https://github.com/qt/qtbase/tree/dev/src/3rdparty/forkfd works pretty well for all Unix operating systems. The pidfd version of it will come soon, as soon as I can find the time to test on a Linux 5.4.0 to make sure things still run (change is at https://codereview.qt-project.org/c/qt/qtbase/+/108456). Here's what you need to know:

* implementation works on all Unix I could get my hands on, including an older broken version of macOS
* implementation works with Bionic, MUSL and uClibc on Linux too
* implementation does catch multiple children exiting and won't miss them due to queued SIGCHLD
* implementation works in single- and multi-threaded environments
* regular operation is thread-safe, lock-free and requires no extra threads. It strictly obeys `man signal-safety`

But:
* you can only listen for child processes. You can't open an already-running process like you can on Windows.
* you can pass the file descriptor via Unix SCM_CREDENTIALS to another process, but it's not adviseable
* installing the signal handler is not thread-safe. There's no way to implement this thread-safely. Some environments may want to install the handler early on, while there's a single thread running to avoid an issue.
* removing the signal handler is not possible. There's no way to safely do this (including in single-threaded mode). Therefore, it's highly adviseable to never unload the signal handler code.
* handling the signal depends on cooperation with any other libraries installing SIGCHLD handlers. Specifically:
  * requires any overriding handlers to call our handler if they didn't handle the child in question
  * requires any overriders not to uninstall our handler if they attempt to remove their handler (they shouldn't remove, actually)

Glib's GSpawn code is guilty of both cooperation shortcomings. Therefore, if there's a chance that GSpawn will be used in the same process, the forkfd signal handler should be installed *after* Glib's.

thiagomacieira:
An alternative implementation is to have a babysitter, intermediary process. Instead of the process that is the final objective being the direct child of the main process, it's a grandchild, with the intermediate child stopped at `waitpid`. When the grandchild dies, the babysitter wakes up, reaps and writes to the pipe. At that point, the main process `waitpid`s the intermediary process too.

Advantages:
* does not depend on cooperation with other frameworks in the same process
* depending on how it's coded, the babysitter may outlive the main process and therefore passing its pipe via SCM_CRED to another process is possible

Disadvantages:
* if the babysitter is a fork-no-exec child, then it could be holding a lot of resources from the parent process. Closing all non-relevant file descriptors is easy, but that won't solve other resources like memory-mapped deleted files. At a minimum, it will still have the entire memory footprint of the original process when forked, thus causing copy-on-write faults on the parent process.
* if the babysitter instead exec's another process, then we have the problem of needing this file to exist in the filesystem somewhere (`execveat` of a memfd executable that was stored inside the parent process' binaries is possible, but then won't share memory with other babysitters). Plus there's a start-up time cost.

The minimum cost of an intermediary babysitter, for a very, very tiny, statically-linked executable, is 1 shared page + 2 pages per process thus launched (1 for .data/.bss and one for the stack, which could be bigger since the environment is stored there).

thiagomacieira:
Finally, a third alternative is to dedicate a thread of the main process to `waitpid` on each child. That means you have one thread per child in the main process. You remove the disadvantages of a separate process (all the memory is shared), the signal handler cooperation and consuming a file descriptor.

But threads aren't free. You do consume at least one page per thread, for the thread's stack. Like the babysitter case, this means you have more Linux processes[*] running at any point in time, which means you could get much closer to your fork limit.

[*] "process" here is the kernel term: any userspace thread.

frehberg:
As already stated in the associated issue-ticket: https://github.com/rust-lang/rfcs/issues/2817

As for the question, how to deal with child-processes on older Linux/Posix-Systems (the non-pidfd compatibility case), the following code illustrates the way, the QProcess implementation is working in this legacy environment successfully.
https://gist.github.com/frehberg/58cac82a332febd6aff0e188c825b699

The code can be executed in two modes 1) Not-Qt-Mode, keeping the read-end inparent-process and write-end in child-process. 2) Qt-Mode, closing both ends in forked child-process and keeping both ends in parent-process.

Compile the code using "gcc -o a.out test_dath_pipe.c"

And execute as "./a.out not-qt-mode" or "./a.out qt-mode"

It shows, that in both modes the parent-process receives the termination signal as POLLIN-event.

The way QProcess has been implemented is favourable, indeed, as it does not leak file-descriptors to sub-sub-processes.

I hope, this short code snippet illustrates the mechanisms, implementing the death-pipe on top of basic POSIX API in case the underlying OS does not provide a suitable native functionality.

text/0000-process-handle-for-async.md
```diff
@@ -0,0 +1,101 @@
...
+Currently `std::process` is managing processes (on Posix/Linux systems) either in synchronous way using the wait-function-family, or asynchronously via signal-handlers (SIGCHLD). 
+
+The problem is that such signals may be delivered to any thread running within the parent-process. So, dealing with the termination of child-processes in async manner will require a shared collection containing the process-status of all child-processes.
+
+This RFC proposes a portable alternative to signal-handling to deal with the termination of child-processes in async manner, and introducing strict ownership for child-processe using a process-handle (aka process descriptor)
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Depending on the underlying operating system, sub-processes may be associated with a process-id (pid) and most oeprating systems provide  so called process-handles (process descriptors). Such handles may be used to perform async polling or sending signals to the corresponding process. Such process handles may be used to establish strict opwnership for child-processes in Rust.
```
the8472:
```suggestion
Depending on the underlying operating system, sub-processes may be associated with a process-id (pid) and most operating systems provide  so called process-handles (process descriptors). Such handles may be used to perform async polling or sending signals to the corresponding process. Such process handles may be used to establish strict ownership for child-processes in Rust.
```

text/0000-process-handle-for-async.md
```diff
@@ -0,0 +1,101 @@
...
+
+The process-handle feature shall permita portable API for various operating systems. The following listing shows the process-handle features being available for different operating systems, and might be the base for a generalized Rust-API in `std::process`.
```
the8472:
```suggestion
The process-handle feature shall permit a portable API for various operating systems. The following listing shows the process-handle features being available for different operating systems, and might be the base for a generalized Rust-API in `std::process`.
```

text/0000-process-handle-for-async.md
```diff
@@ -0,0 +1,101 @@
...
+
+- As with Windows, the fucntion `CreateProcess` provides the HANDLE for the newly created child-process.
+- As with OpenBSD Unix provides process handles 
+- As with FreeBSD Unix the function `libc::pdfork(..)` provides a process descriptor for the newly forked child-process.
+- As with Linux, the `pidfd`-familiy provides funtionality to convert a pid int a file-descriptor (merged into release Linux-5.4)
+- Older Linux-releases or Posix-Systems may use the `forkfd`-concept to establsih a descriptor/handle for a child-process
+ 
+As for the latter case when the underlying OS does not provide a suitable naticve process-handle, the forkfd concept may be used. This is based on so called death-pipes: An anonymous pipe is established between parent-process (RX end) and the child-process (TX end). Here the parent-process may use RX end to poll for read-events (file descriptor) in async manner, for example using a future.
```
the8472:
```suggestion
As for the latter case when the underlying OS does not provide a suitable native process-handle, the forkfd concept may be used. This is based on so called death-pipes: An anonymous pipe is established between parent-process (RX end) and the child-process (TX end). Here the parent-process may use RX end to poll for read-events (file descriptor) in async manner, for example using a future.
```

text/0000-process-handle-for-async.md
```diff
@@ -0,0 +1,101 @@
...
+Now, if the child-process terminates the TX end is being destroyed and the RX-end in the parent process will receive and EOF read-event.
+
+All of the listed process-handle flavors share the common concept to perform async polling for process-termination and might be the base for a generalized process-handle concpet in Rust `std::process`, leveraging a strong ownership-concpet for child-processes. The handling of signal SIGCHLD and sharing of child-process meta-data would no longer be required.
```
the8472:
```suggestion
All of the listed process-handle flavors share the common concept to perform async polling for process-termination and might be the base for a generalized process-handle concept in Rust `std::process`, leveraging a strong ownership-concept for child-processes. The handling of signal SIGCHLD and sharing of child-process meta-data would no longer be required.
```

PR #2840, opened 2019-12-29
Non-ASCII identifiers: NFC normalization is performed by lexer, not parser

petrochenkov:
cc https://github.com/rust-lang/rust/issues/55467#issuecomment-569484220


Centril:
@petrochenkov Can you elaborate a bit on your comment over at the tracking issue wrt. the trade-offs?

Centril:
Also cc @dtolnay

petrochenkov:
@Centril 
It's not even a trade-off, just something new - we didn't previously lose any information about tokens before passing them to macros.
Requiring macros to perform NFC normalization by themselves would be kind of insane (we require them to perform string unescaping by themselves though).

petrochenkov:
NFC normalization itself is a tradeoff - identifier no longer directly corresponds to a range in the source map, so any code doing span arithmetic assuming e.g. `ident.len() == ident.span.len()` turns into a goldmine for fuzzers.

Centril:
Alright. Thanks.

@rust-lang/lang I'll go ahead and merge this without FCP since this feels like sufficiently a detail and I'll defer to @petrochenkov's judgement here (although the explanation given makes a lot of sense).

PR #2842, opened 2020-01-01
Update unicodeset link

Manishearth:
The syntax changed and the links broke.

r? @centril


Manishearth:
Also wondering if I should update the RFC to use the term `uncommon_codepoints` (the RFC says `less_used_codepoints` but it left that up as an unresolved question and the tracking issue has settled on `uncommon_codepoints`), but my understanding is that we typically don't update RFCs to reflect decisions made later? (@nikomatsakis ?)

Centril:
> but my understanding is that we typically don't update RFCs to reflect decisions made later? (@nikomatsakis ?)

We don't require it, but I *personally* wouldn't mind accepting & reviewing such fixes and clarifications.

PR #2871, opened 2020-02-28
RFC: Clarifying UnwindSafe

Diggsey:
[Rendered](https://github.com/Diggsey/rfcs/blob/clarify_unwind_safe/text/0000-clarifying-unwind-safe.md)

### DIFF
--- a/text/0000-clarifying-unwind-safe.md
+++ b/text/0000-clarifying-unwind-safe.md
@@ -1,0 +1,150 @@
+- Feature Name: clarifying_unwind_safety
+- Start Date: 2020-02-27
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+The [`UnwindSafe`](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html) marker trait and associated machinery were stabilized in Rust 1.9 with the intent to help prevent a class of logic errors which might otherwise be introduced by the addition of the [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) function.
+
+However, the exact requirements for an `UnwindSafe` type have never been formally stated, and there is much confusion on the topic. Many types which probably should be considered `UnwindSafe` do not implement the trait, resulting in many false positives when `catch_unwind` is used. As a result, it is often regarded by Rust developers as simply a nuisance to be bypassed by liberal use of `AssertUnwindSafe`.
+
+# Motivation
+[motivation]: #motivation
+
+With this RFC I hope to:
+- Attain consensus on what precisely the `UnwindSafe` trait means.
+- Agree a direction for improving the current state of `UnwindSafe`.
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+As this is a rather atypical RFC, this section will explain the current situation and propose a set of more formal rules for when a type should be considered `UnwindSafe`. Potential improvements will be discussed in the reference-level explanation.
+
+## Background
+
+The `UnwindSafe` trait was originally introduced under the name `PanicSafe` in [RFC 1236](https://github.com/rust-lang/rfcs/blob/master/text/1236-stabilize-catch-panic.md). That RFC gives many examples of how `UnwindSafe` might be used, but falls short of providing any formal definition of what is required for a type to be `UnwindSafe`. Since then there has been no further clarification of what this trait actually means. As a result, documentation has been written about this trait whose facts do not come from any accepted RFC, and which is (in my view) incorrect.
+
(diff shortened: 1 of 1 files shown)


KrishnaSannasi:
On the note of  poisoning and atomics implementing `RefUnwindSafe` [in this section](https://github.com/Diggsey/rfcs/blob/clarify_unwind_safe/text/0000-clarifying-unwind-safe.md#poisoning)

Poisoning prevents people from seeing the contents of a `Mutex` in an erroneous state. Mutexes only poison if they witness a panic while locked. Viewed from this way, since there is no way to panic while an atomic is "locked" (doing an operation), they are trivially `RefUnwindSafe`.

Diggsey:
@KrishnaSannasi That could be one way to interpret it, however by that logic `Cell` should also implement `RefUnwindSafe`, yet it explicitly opts out:
https://doc.rust-lang.org/std/cell/struct.Cell.html#impl-RefUnwindSafe

KrishnaSannasi:
It's not explicitly opted out, it just inherited that from `UnsafeCell`, that may just be an oversight.

glaebhoerl:
> Viewed from this way, since there is no way to panic while an atomic is "locked" (doing an operation), they are trivially `RefUnwindSafe`.

Here is another way to look at it: in concurrent code, because threads may run in parallel or pre-empt each other at any time, it is necessary to explicitly mark those parts of the code where invariants may be temporarily violated before being restored ("critical sections"), to prevent other threads from observing the inconsistent state. (We also need it to avoid low-level data races, but if that were the only issue, then we could solve it by just using atomics for everything, and it isn't.) In single-threaded code, these temporary violations of invariants are still there, but since there is no danger of concurrency or preemption, they are just implicit -- we're just careful not to call any functions which could problematically witness the inconsistent state before we make it consistent again, and that's it.

This is similar to how if you're writing a single-threaded program with its own database, you can get away for a while without really thinking about transactions (the database's version of critical sections), because it's not possible for e.g. another thread to run a SELECT in between two of your INSERTs which really should've happened atomically (because only one of the tables having been updated and not the other is a temporary invariant violation). Once you add another thread, you need to start marking these regions of temporary invariant violations, which were there all along, explicitly.

But there's a way that the absence of explicit transactions can bite you even in single-threaded database code: if your program crashes or gets terminated, the computer freezes or the power goes out, *etc.*, maybe your program will have finished doing the first INSERT but not the second. And now, the next time your program starts, its database will be in an inconsistent state, and the program will misbehave.

This is totally analogous to "unwind-unsafety" in single-threaded code: you have two `Cell`s which need to get updated together, and you do just that, no other thread can touch them, it's fine. Until one day, the code in between the two updates panics (say, because of an integer overflow). You didn't want a panic to take down your whole program, so you have a `catch_unwind` at an outer scope to restart the task. But now it witnesses an inconsistent state, and the program misbehaves.

(This ties into the subject of mutex poisoning: the *ideal* thing to do, if a thread panics while holding a locked mutex, would be to *restore the contents of the mutex to its original state*, from the beginning of the critical section, before unlocking the mutex and continuing, so that it behaves transactionally. But making a copy at the beginning would be costly in the normal case where the code *doesn't* panic, and may also be impossible for move-only types. So poisoning the mutex is a good next-best compromise.)

glaebhoerl:
> A type should be `RefUnwindSafe` iff any of the following are true:
> 
> 1) The type is not internally mutable.
> 2) The type is `Sync`.
> 3) The type implements poisoning.
> 4) The type is `AssertUnwindSafe`.

My vague recollection is that `&mut` can also be used to witness inconsistent state post-panic. (Which is mutation, but not 'internal'.) Am I mistaken?

Diggsey:
Ah yes, `&mut` is a bit of an edge case - it's not internally mutable in the sense that you can't mutate it behind a `&` reference, but on the other hand, it does allow you to mutate things through an immutable `let` binding.

KrishnaSannasi:
> other hand, it does allow you to mutate things through an immutable let binding

Only if you go through some other form of interior mutability. Otherwise you can't mutate values stored in immutable let bindings.

Lokathor:
I think they mean
```rust
let gen = &mut AnRNGType::new();
gen.next_u32(); // immutable binding to a mutable reference allows mutation of pointed-to data
```

glaebhoerl:
An interesting assumption we use in reasoning about the unwind-safety of mutexes and atomics, which is worth calling out explicitly, is that types which *can* be used in a multi-threaded way, *will be*. Since unwind-safety-checking is only "lint-like" in the first place, this is probably fine. But there's no reason in principle why you couldn't use `AtomicUsize` in single-threaded code in the same way as a `Cell<usize>`, and in the case of `Cell` we said it's *not* unwind-safe. Similarly, you can also use a `Mutex<T>` in the same way as a `Cell<T>`, only ever performing a single read or write inside the critical section, while only accessing it from a single thread. We just *expect* (with some justification) that people won't actually do this in practice.

Diggsey:
@glaebhoerl The way I would like people to think about it is this: the fact that a type implements `Sync` is the "warning sign" that it might be used concurrently, and so whenever you are dealing with `Sync` types you should be concerned about unwind safety. This is why I listed them as an "escape hatch" from the `UnwindSafe` mechanism.

Amanieu:
My personal preference would be to get rid of `UnwindSafe` entirely.

PR #2884, opened 2020-03-17
RFC: Placement by return

PoignardAzur:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/2884)*

[Rendered](https://github.com/PoignardAzur/placement-by-return/blob/placement-by-return/text/0000-placement-by-return.md).

---

**Glossary:**

- **GCE:** [Guaranteed Copy Elision](https://stackoverflow.com/questions/38043319/how-does-guaranteed-copy-elision-work).
- **NRVO:** [Named Return Value Optimization](https://shaharmike.com/cpp/rvo/).
- **DST:** [Dynamically-Sized Type](https://doc.rust-lang.org/reference/dynamically-sized-types.html).

PR #2888, opened 2020-03-24
RFC: `ForbiddenValues` trait to enable more optimizations

tommit:
[Rendered](https://github.com/tommit/rfcs/blob/forbidden-value-trait/text/0000-forbidden-value-trait.md)


shepmaster:
I'm no compiler implementation expert, but this RFC feels pretty light in the details department. It seems to focus mostly on a surface syntax of how this would be used, but doesn't give any time to *how* the implementation using that information would happen.

KrishnaSannasi:
This looks like it only permits 1 forbidden value, which is rather limiting. Note: If you're type has any padding bytes, you can get `255` forbidden values by simply replacing the padding byte with

```
#[repr(u8)]
enum ZeroPadding {
    Padding = u8
}
```

For enums are much harder to layout optimize, usually we can merge the discriminants together, this is what allows `Option<Result<(), ()>>` to be a single byte. So I don't see the value in adding a special (super-limited) trait. If you can make this more general than what we already have, then it may be worth consideration. But as it stands, we really don't need this added complexity.

tommit:
@shepmaster I know nothing about rustc, but here's my best guess about the implementation:
1) In rustc, there must code that decides whether or not to perform an optimization based on forbidden value(s) given some type(s) - I propose we change that code to also check whether or not the type(s) in question implements `ForbiddenValue`.
2) In rustc, there must be code that decides which forbidden value(s) to use when it performs the optimization - I propose we change that code to use `ForbiddenValue::FORBIDDEN_VALUE_BYTES` as the forbidden value where applicable.

@KrishnaSannasi The value of this RFC is for those cases where it's not convenient to use an enum as a part of your type. For example:
```rust
pub struct S {
    x: u32
}
```
If legal values for the field `x` are all except `u32::MAX`, then you wouldn't want to make an enum with 2^32-1 variants and use it for `x` in order to get the optimizations for `S`.

KrishnaSannasi:
So, similar to the currently unstable + internal `rustc_layout_scalar_valid_range_start` and `rustc_layout_scalar_valid_range_end` attributes?

tommit:
@KrishnaSannasi I wasn't aware of those attributes, so thank you for bringing them to my attention. The error message when `#![feature(rustc_attrs)]` isn't available says that those attributes will never be stable. But they seem to be one way of tapping into the forbidden-value-based optimizations which is exactly what this RFC is asking for.

clarfon:
Personally, I'd rather those internals be exposed as some sort of bounded integer types than something like this.

Lokathor:
That is much less easily usable to create type system guarantees.

shepmaster:
I don't think "bounded integer types" and "arbitrary niche specification" are exclusive features. Bounded integer types are a concrete usage, potentially with a different surface syntax, but could be implemented using arbitrary niches.

In my head, I view it akin to how `async`/`await` sits atop generators.

clarfon:
That's a pretty good analogy, honestly. The main reason why I mention bounded integer types is because it requires actual language support to allow coercing integer literals into bounded types, and allowing casting properly. Niche specialization can technically be done today in a very hacky way with enums, although I personally wouldn't mind requiring that someone make an enum to allow niches like disjoint ranges (e.g. a byte whose values can only be 0-10 or 20-100).

Evrey:
It would be extremely powerful if we could utilise `const impl` for this feature to turn the check into a `const fn is_forbidden(self) -> bool`. For `NonZeroX` that'd just do `self == 0`.

Alternatively, and with less reliance on `const fn` progress, the `FORBIDDEN_VALUE_BYTES` constant could instead be a `&[[u8; size_of::<Self>]]`. If that slice is empty, the trait impl has no effect. If it has just one entry, it equals the current proposal. If it has more than one entries, all of them are forbidden. If all possible values are included, then all possible states of `Self` are forbidden, oops. An equal entry appearing more than once has the same effect as an entry appearing just once.

pickfire:
How about a forbidden bit? I have seen it in http://troubles.md/abusing-rustc/

elichai:
Is this useful as long as repr(Rust) has unstable representation?
I don't want to trade one optimization for another.

Evrey:
@elichai Good question. As in my boolean function proposition, it would be good to just use `Self` instead of a byte array. So `const FORBIDDEN_VALUE: Self`. However, with a constant, one could abuse that by calling methods in invalid state. With a boolean function, there is no instance of an invalid `Self` lying around.

tommit:
> This looks like it only permits 1 forbidden value, which is rather limiting.

@RustyYato I changed this RFC to allow multiple forbidden values. Also, added an example of a `FastFloat` type that shows a case where a bounded integer type wouldn't be of any use.

text/0000-forbidden-value-trait.md
```diff
@@ -0,0 +1,70 @@
...
+```rust
+#![feature(const_transmute)]
+
+#[repr(transparent)]
+struct FastFloat(f32);
+
+unsafe impl ForbiddenValues<[[u8; 4]; 3], [u8; 4]> for FastFloat {
+    const FORBIDDEN_VALUES: [[u8; 4]; 3] = unsafe {
+        [
+            mem::transmute(f32::NAN),
```
Evrey, line 45: Note that `f32` has about 16.7 million different `NaN` values.

text/0000-forbidden-value-trait.md
```diff
@@ -0,0 +1,70 @@
...
+            mem::transmute(f32::INFINITY),
+            mem::transmute(f32::NEG_INFINITY),
+        ]
```
pickfire, line 48: How do we have a range of forbidden values? For example if we were to have `u8` that is able to store 55 numbers at most, then all the rest of the numbers as forbidden values.

text/0000-forbidden-value-trait.md
```diff
@@ -0,0 +1,70 @@
...
+This would increase the number of situations where the compiler can do the aforementioned, genuinely useful optimizations.
+
+# Detailed design
+
+Add the following trait somewhere in the standard library:
+```rust
+unsafe trait ForbiddenValues<A, B>
+where
+    A: FixedSizeArray<B>,
+    B: FixedSizeArray<u8>,
```
pickfire, line 20: Must this be `u8`?

pickfire:
I wonder if rust currently pack struct of bools into just a single `u8` or `usize`. It would be interesting to see multiple `bool` reuse the same memory but different bit.

pickfire:
In addition to multiple combination of numbers, I wonder if there are ways to specify a range of bits easily despite being less efficient. Like some masks
```rust
forbidden_bits(0b11110000)
```

text/0000-forbidden-value-trait.md
tommit: No problem. You just write a procedural macro to produce those 16.8 million lines of code and buy a supercomputer to compile it 😜

I concede that your suggestion for a trait with a `const fn is_forbidden(&self) -> bool` method would be strictly better than this RFC - it would make it (relatively) easy to specify large ranges of forbidden values. But I do wonder about its compiler implementation - without knowing much at all, I would imagine that the compiler would have to call `value.is_forbidden()` for each possible bit pattern of `value`.

jhpratt: Why would the compiler have to call `value.is_forbidden()` for each possible bit pattern?

If the compiler "needs" 5 values for a variety of optimizations, it would just have to find 5 forbidden values and use those.

clarfon:
> I wonder if rust currently pack struct of bools into just a single `u8` or `usize`. It would be interesting to see multiple `bool` reuse the same memory but different bit.

It can't pack them into one bite because `bool` is guaranteed to always be the least significant bit. There's no way to pack things smaller than the byte level at the moment, even though enums optimise this way because of monomorphism.

> In addition to multiple combination of numbers, I wonder if there are ways to specify a range of bits easily despite being less efficient. Like some masks
> 
> ```rust
> forbidden_bits(0b11110000)
> ```

The compiler uses value ranges internally which are IMHO a lot more natural to deal with.

text/0000-forbidden-value-trait.md
tommit: You would have to add all those 201 (2^8-55) forbidden values to the `FORBIDDEN_VALUES` array as `[u8; 1]` values.

clarfon: That seems extremely suboptimal. Why not just accept ranges? Individual values can be done as ranges too.

text/0000-forbidden-value-trait.md
clarfon: That would always have a worst case of going through every value, though. Which isn't scalable past a few bits.

text/0000-forbidden-value-trait.md
jhpratt: I have a struct that is packed and uses 30 bits. For obvious reasons it's stored as a 32 bit integer.

If we used `const fn(T) -> bool` or ranges, this would be trivial to write. With the proposal, I'd have to list out _over a billion values_. Obviously that's neither feasible nor fast.

text/0000-forbidden-value-trait.md
tommit: Yes, I think so. `B` is an array of bytes used for representing a forbidden bit pattern of `Self`. I believe that if `T` implements some `ForbiddenValues<A, B>`, then it could be undefined behaviour to have a value of type `T` with a bit pattern equal to one of the items in `FORBIDDEN_VALUES`.

text/0000-forbidden-value-trait.md
clarfon: I agree that a list is infeasible, but my point is that a predicate is also infeasible. Imagine that the forbidden values are in the higher bits, and that the compiler simply does a sweep of the values in order to find the forbidden ones. It'll still have to go through almost anything.

And if your argument is that the compiler should be able to look at the function and determine what it's doing, 1. the halting problem and 2. ranges are easier and way less jank :p

jhpratt: Slice of ranges? I think that would solve pretty much every situation, as a value can (of course) be represented as a range.

text/0000-forbidden-value-trait.md
tommit: @clarfon Do you mean ranges as in something like the following?
```rust
unsafe trait ForbiddenValues<A, B>
where
    A: FixedSizeArray<(B, B)>,
    B: FixedSizeArray<u8>,
{
    const FORBIDDEN_VALUES: A;
}
```
Where `B` is now interpreted as a fixed-size unsigned integer type with an arbitrary width, and the tuple `(B, B)` represents the inclusive start and end of the range of forbidden values.

text/0000-forbidden-value-trait.md
tommit: @clarfon Do you mean ranges as in something like the following?
```rust
unsafe trait ForbiddenValues<A, B>
where
    A: FixedSizeArray<(B, B)>,
    B: FixedSizeArray<u8>,
{
    const FORBIDDEN_VALUES: A;
}
```
Where `B` is now interpreted as a fixed-size unsigned integer type with an arbitrary width, and the tuple `(B, B)` represents the inclusive start and end of the range of forbidden values.

programmerjake: I think using a list of both ranges and bit masks to define forbidden values is probably the best solution, since it handles nearly all cases and doesn't have the halting problem and similar issues you'd run into with defining it using a `const fn`.

For the bit mask, it would match all values `v` that pass `(v & a) == b` where `a` and `b` are values of some sort (byte arrays? integers?).

As an example of why ranges alone are insufficient:
What if you have a type like:
```rust
enum E<T> {
    A(T),
    B(T),
    C(T),
    D(T),
}

type E2<'a'> = E<&'a u32>;
```

where you want Rust to pack the enum tag for `E2` into the 2 low bits of the pointer since the pointer is guaranteed to be some value with the two low bits are zeros due to references guaranteed alignment and `u32` having an alignment of 4? Rust currently doesn't pack the enum that way, but could. This kind of bit packing is used extensively throughout LLVM's codebase (implemented using custom C++ code), so is not that uncommon.

If you were to enumerate the forbidden values of a custom `T` type using ranges, you'd get something like:
```rust
const FORBIDDEN_VALUES: [...] = [
    0x0..=0x3, // include 0 for null
    0x5..=0x7, // repeats every 4 values
    0x9..=0xB,
    0xD..=0xF,
    0x11..=0x13,
    0x15..=0x17,
    0x19..=0x1B,
    0x1D..=0x1F,
    ... // another *billion* lines for 32-bit pointers
];
```

tommit: Something like the following to allow forbidding values both by ranges and by a bit mask?
```rust
unsafe trait ForbiddenValues<R, B>
where
    R: FixedSizeArray<RangeInclusive<B>>,
    B: FixedSizeArray<u8>,
{
    const FORBIDDEN_RANGES: R;
    const FORBIDDEN_BITS: B;
}
```

text/0000-forbidden-value-trait.md
Evrey: To eliminate dead code, all `rustc` needs to do is insert and constant-fold the mentioned `unreachable_unchecked` snippet. This is something it can do today.

Finding holes for compactly encoding `enum` variants, however, is more tricky.

Evrey:
If we don't want to go for functions, which is understandable, as reverse-finding invalid values is difficult to implement, then we need the following primitives to implement efficient packing of multiple invalid states:

- Pairs of bit masks and bit patterns, which rule out large chunks of values which may be non-continuous.
  - To elaborate, `(0b0_11111111_000_0000…, 0b0_11111111_000_0000…)` would say this:
    - `pair.0` fetches the exponent bits of an `f32`.
    - `pair.1` says »if the masked-off value equals this, then boo-boo«, which would be all `NaN`s.
- Simple bit masks, which rule out all values where `0 != (x & mask)`. A boolean is then an `u8` with the mask `0b11111110`.
- Ranges, which rule out large chunks of continuous values that don't have to hit powers of two.
  - However… ranges on byte arrays? How do you express this? `[0x00, 0x00, 0xD8, 0x00] ..= [0x00, 0x00, 0xDF, 0xFF]`, which would filter away surrogate pairs from a `char`. In Big Endian, by the way.
- A blacklist of individual values.

However, there's two big downsides:

- Expressing interdependencies of fields is still extremely unpractical, requiring billions of entries at worst. My example, again, would be a range type that guarantees that `start <= end`.
- As hinted at, what about endianness? This is where the raw bytes approach quickly falls apart.

I still favour the function approach, even though it's more complicated compiler work for enum optimisation.

pickfire:
> It can't pack them into one byte because bool is guaranteed to always be the least significant bit. There's no way to pack things smaller than the byte level at the moment, even though enums optimise this way because of monomorphism.

Is there a reason why must it be the least significant bit? Is accessing it faster? Can it be the second least significant bit?

text/0000-forbidden-value-trait.md
programmerjake: For bitmasks, just a single integer is insufficient, you'd need something like:

```rust
struct BitMask<T> {
    mask: T,
    value: T,
}

impl<T> BitMask<T> {
    fn matches<'a>(&'a self, value: T) -> bool
    where
        T: BitAnd<&'a T, Output = T> + PartialEq,
    {
        (value & &self.mask) == self.value
    }
}

unsafe trait ForbiddenValues<B>
where
    B: MemoryBits,
{
    const FORBIDDEN_RANGES: &'static [RangeInclusive<B>];
    const FORBIDDEN_BIT_MASKS: &'static [BitMask<B>];
}
```

This allows you to represent things such as an integer where top 4 and bottom 12 bits can't be `0x3` and `0xCAB` by including a bit mask `BitMask { mask: 0xF000_0FFFu32, value: 0x3000_0CABu32 }` assuming `u32` implements `MemoryBits`.

programmerjake:
> > It can't pack them into one byte because bool is guaranteed to always be the least significant bit. There's no way to pack things smaller than the byte level at the moment, even though enums optimise this way because of monomorphism.
> 
> Is there a reason why must it be the least significant bit? Is accessing it faster? Can it be the second least significant bit?

the reason is because `bool` is defined to be compatible with C and C++'s `bool`, which is a single byte with values 0 and 1 on basically every platform. If you had some struct `S` with two bools:
```rust
struct S {
    a: bool,
    b: bool,
}
```
Rust allows you to do something like:
```rust
extern "C" {
    fn my_external_function(a: *mut bool, b: *mut bool);
}

fn f(v: &mut S) {
    unsafe {
        my_external_function(&mut v.a, &mut v.b);
    }
}
```

where `my_external_function` is a C/C++ function that can read and write to both `bool`s, so they both are required to be the same as C++'s `bool`.

clarfon:
There's actually a simpler reason for bool: types must always have the same layout in memory to make borrows work. If `&bool` could refer to any bit of a given address then it no longer would be just a pointer. Sure, having a type like `bit<N>` would solve this but that's not bool.

tommit:
How about a bit different approach, using internal iteration:

```rust
unsafe trait ForbiddenValues<B>
where B: FixedSizeArray<u8>
{
    const COUNT: usize;

    fn forbidden_values<L>(loop_body: L)
    where L: FnMut(&B) -> bool;
}
```

Where:
* `ForbiddenValues::COUNT` is the number of different forbidden values the implementor has (which makes it easy for the compiler to quickly check if it can perform a certain optimization that requires at least N forbidden values)
* If `loop_body` returns true, it's a sign that the compiler wants to stop querying new forbidden values, and `forbidden_values` should return early
* If `loop_body` always returns false, then `forbidden_values(loop_body)` must call `loop_body` exactly `ForbiddenValues::COUNT` times
* Each argument passed to `loop_body` is a reference to an array of bytes that represent a bit pattern of a forbidden value
* `forbidden_values` must not pass the same argument to `loop_body` twice

One might implement `ForbiddenValues` like this:

```rust
unsafe impl ForbiddenValues<[u8; 4]> for f32 {
    const COUNT: usize = 5;

    fn forbidden_values<L>(mut loop_body: L)
    where L: FnMut(&[u8; 4]) -> bool {
        let mut forbidden_bytes: [u8; 4];

        macro_rules! call_loop_body {
            () => {
                if loop_body(&forbidden_bytes) {
                    return;
                }
            };
        }

        for i in 0..5 {
            forbidden_bytes = unsafe { mem::transmute(1u32 << i) };
            call_loop_body!();
        }
    }
}
```

With a future compiler (after [this bug](https://github.com/rust-lang/rust/issues/43408) is fixed), the definition could be made simpler and more robust:

```rust
unsafe trait ForbiddenValues {
    const COUNT: usize;

    fn forbidden_values<L>(loop_body: L)
    where L: FnMut(&[u8; mem::size_of::<Self>()]) -> bool;
}
```

RustyYato:
> How about a bit different approach, using internal iteration:

This is interesting, however I think a better solution would build off of `Iterator`

```
unsafe trait ForbiddenValues
{
    type ByteArray: FixedSizeArray<u8>;
    type Forbidden: Iterator<Item = Self::ByteArray>;
    
    fn forbidden_values() -> Self::Forbidden;
}
```

Then, Rust could just call `fobidden_values` when it wants a series of forbidden values, and store the iterator if it needs to get more `forbidden_values` without reusing old values.

tommit:
> This is interesting, however I think a better solution would build off of `Iterator`

@RustyYato  That did occur to me first, but then I thought it would be easier for the user to implement internal iteration than an (external) `Iterator` for this use case.

> ...and store the iterator if it needs to get more forbidden_values without reusing old values.

It didn't occur to me that external iterator would provide this flexibility to the compiler implementation. Although, I have no idea whether the compiler actually needs that flexibility or not.

Note that your trait definition can be made a bit simpler:
```rust
unsafe trait ForbiddenValues {
    type Forbidden: Iterator<Item: FixedSizeArray<u8>>;

    fn forbidden_values() -> Self::Forbidden;
}
```

pickfire:
> where my_external_function is a C/C++ function that can read and write to both bools, so they both are required to be the same as C++'s bool.

Couldn't the memory address be changed to accommodate the change of the least significant bit? I am not quite sure about this but I think if one can do `bits >> 1` to move the least significant bits from second level, I think it may be possible to use the second least significant bit using bit shift. Am I wrong?

> There's actually a simpler reason for bool: types must always have the same layout in memory to make borrows work. If &bool could refer to any bit of a given address then it no longer would be just a pointer. Sure, having a type like bit<N> would solve this but that's not one type.

Ah, but is there a reason why borrow must be on a byte level rather than on a bit level? This could enables more optimization and LLVM could pad those unused bits later (probably).

> ```rust
> unsafe trait ForbiddenValues: Sized {
>     type Forbidden: Iterator<Item = [u8; mem::size_of::<Self>()]>;
> 
>     fn forbidden_values() -> Self::Forbidden;
> }
> ```

This looks good. But do we need associated type? Can it be `fn forbidden_values() -> Iterator<Item = [u8; mem::size_of::<Self>()]>`?

> ```rust
> unsafe trait ForbiddenValues {
>     type ByteArray: FixedSizeArray<u8>;
>     type Forbidden: Iterator<Item = Self::ByteArray>;
>     
>     fn forbidden_values() -> Self::Forbidden;
> }
> ```

Is there any chance it that `ByteArray` would not be `FixedSizedArray<u8>`?

burdges:
> Ah, but is there a reason why borrow must be on a byte level rather than on a bit level? 

There are several bitfield crates so one might provide some `BitRef<'a,T>` and  `BitMut<'a,T>` borrows that stored a `usize` that shifts, although the latter might run into https://github.com/rust-lang/rfcs/issues/1403  It'd need either the backing type `T` to be atomic or else `BitRef`/`BitMut` must not be `Send`, so they must act as if `T: !Sync` even though `T: Sync`. 

You'll want some higher level interfaces normally I think.

RustyYato:
@tommit

> It didn't occur to me that an external iterator would provide this flexibility to the compiler implementation. Although, I have no idea whether the compiler actually needs that flexibility or not.

One example is `Option<Option<MyNicheType>>`. If we have qn external iterator, we could do something like

```
Layout: MyNicheType

MyNicheType.niches = MyNicheType::forbidden_values()

Layout: Option<MyNicheType>

Option<MyNicheType>.niches = MyNicheType.niches.clone()
none_repr = Option<MyNicheType>.niches.next()

Layout: Option<Option<MyNicheType>>

Option<Option<MyNicheType>>.niches = Option<MyNicheType>.niches.clone()
none_repr = Option<Option<MyNicheType>>.niches.next()
```

Altough this requires the iterator to be clone. I think this is a reasonable requirement.

Ixrec:
Somehow no one's touched on the deeper question about motivation so I guess I'll do it.

AFAIK, any exotic layout guarantees you could imagine are already _possible_ to implement today either with a bunch of bitshift operators or with some `unsafe` Rust, and the real question is whether crates already do the job or some of those use cases are painful or limited enough to benefit from some core language integration. Even for the concrete example of bitfields, this seems to be an ongoing debate, although at least to me it feels like (ignoring FFI issues involving C's bitfields) mostly the community is fine with the crates and having core language sugar just isn't that much of a win over macros in a crate. The only other concrete example I can think of is prefix layout guarantees for efficient single inheritance, which seems to have even lower demand as I've seen no discussion on it in years (unlike other aspects of inheritance). And in both of those examples, I can't recall anyone claiming they put such types into an enum and then suffered from a missing niche optimization.

So in that context, a fully general mechanism for customizing arbitrary niche optimization seems to be exceptionally lacking in motivation. Plus, the current text of the RFC doesn't even mention "hand-optimize your layouts with bitshifts/`unsafe`" as an alternative, despite that being precisely the status quo it needs to argue against. All this talk of whether the interface should use ranges or bit patterns or whatever seems kinda pointless until we've established why crates aren't good enough.

tommit:
@Ixrec Let's say I have created a `struct Foo { n: u32 }` and I know that the value in the field `n` is never equal to `u32::MAX`. Let's also say that my goal is to have `size_of::<Foo>() == size_of::<Option<Foo>>()`.

> AFAIK, any exotic layout guarantees you could imagine are already _possible_ to implement today either with a bunch of bitshift operators or with some `unsafe` Rust...

Are you saying that I can accomplish my goal today? And if so, how would I go about doing it?

> And in both of those examples, I can't recall anyone claiming they put such types into an enum and then suffered from a missing niche optimization.

...or are you saying that my goal is not worth it to warrant this RFC?

Ixrec:
I'm saying this RFC hasn't provided any compelling motivation, and I'm not aware of any myself. Of course I can't unilaterally declare that no good motivation exists, but if it does exist it needs to be written down in the RFC itself. The burden of proof is on the change proposal.

Your specific example is of course not possible today. It's trivial to come up with an example that technically isn't possible today without exactly the feature being proposed. The issue is showing a realistic and compelling example to convince people that actual code would actually benefit from this to a significant enough degree that cannot be achieved in other ways that it's worth adding the proposed feature. Is this `Option<NeverMaxU32>` example something that you ran into in practice? If so, can you describe more of the context and why none of the obvious alternatives were applicable there?

shepmaster:
> Are you saying that I can accomplish my goal today? And if so, how would I go about doing it?

[Can I use the “null pointer optimization” for my own non-pointer types?](https://stackoverflow.com/q/30414068/155423)

```rust
use std::{mem, num::NonZeroU32};

struct Foo {
    n: NonZeroU32,
}

impl Foo {
    pub fn new(v: u32) -> Self {
        Self { n: Self::convert(v) }
    }
    
    pub fn get(&self) -> u32 {
        u32::from(self.n).wrapping_sub(1)
    }

    pub fn set(&mut self, v: u32) {
        self.n = Self::convert(v)
    }
    
    fn convert(v: u32) -> NonZeroU32 {
        NonZeroU32::new(v.wrapping_add(1)).expect("Out of range")
    }
}

fn main() {
    let mut f = Foo::new(0);
    dbg!(f.get());

    f.set(1);
    dbg!(f.get());

    // f.set(u32::MAX); // panic
    // dbg!(f.get());

    assert_eq!(mem::size_of::<Foo>(), mem::size_of::<Option<Foo>>());
}
```

tommit:
> I'm saying this RFC hasn't provided any compelling motivation, and I'm not aware of any myself.

I assumed that people wouldn't have added the optimization that makes `size_of::<Option<bool>>() == 1` to the compiler if they didn't have some motivation for doing it. And if there was some motivation for doing that optimization for `bool`, then why not for other types?

> The issue is showing a realistic and compelling example to convince people that actual code would actually benefit from this to a significant enough degree that cannot be achieved in other ways that it's worth adding the proposed feature.

My actual use case has been saving a considerable amount of RAM with an array of `Option<T>` values, where `T` has forbidden values. As a workaround I have created a `struct Opt<T> { opt_value: T }` to mimic the functionality of `Option<T>` but it lacks the ergonomics that come with an actual enum (like pattern matching and the `?` operator).

burdges:
You should check if any proc macro crates for bitfields and/or packed enum create some `Guard` type ala `MutexGuard` that you explicitly borrow from the bitfield and/or packed enum.  I'd expect some `Guard` type would remain the only way to pack this type into 1 byte because you must handled mutated data when taking references: 
```
enum Foo {
    A(U8LessThan64>
    B(U8LessThan64>
    C(U8LessThan128>
}
```

tommit:
> [Can I use the “null pointer optimization” for my own non-pointer types?](https://stackoverflow.com/q/30414068/155423)
> 
> ```rust
> use std::{mem, num::NonZeroU32};
> 
> struct Foo {
>     n: NonZeroU32,
> }
> 
> impl Foo {
>     pub fn new(v: u32) -> Self {
>         Self { n: Self::convert(v) }
>     }
>     
>     pub fn get(&self) -> u32 {
>         u32::from(self.n).wrapping_sub(1)
>     }
> 
>     pub fn set(&mut self, v: u32) {
>         self.n = Self::convert(v)
>     }
>     
>     fn convert(v: u32) -> NonZeroU32 {
>         NonZeroU32::new(v.wrapping_add(1)).expect("Out of range")
>     }
> }
> 
> fn main() {
>     let mut f = Foo::new(0);
>     dbg!(f.get());
> 
>     f.set(1);
>     dbg!(f.get());
> 
>     // f.set(u32::MAX); // panic
>     // dbg!(f.get());
> 
>     assert_eq!(mem::size_of::<Foo>(), mem::size_of::<Option<Foo>>());
> }
> ```

It only now, much later, occurred to me that your solution here probably kind of is what @Ixrec meant by _implementing exotic layout guarantees with a bunch of bitshift operations_ (because I had no idea what he meant by that). With the exception that your example uses `wrapping_add|sub` instead of shifts because my example just so happened to have no extra bit to spare (which was not my intention... like I said, I had no idea).

So, if my options are to use `Option<Bar>` with extra shift/math operations when accessing `Bar` or use a trivial `struct Opt<T> { t: T }` wrapper type that has worse ergonomics than `Option` but has no size/performance overhead, then I would choose the latter.

Ixrec:
Sorry for the confusion. @burdges and @shepmaster's comments together are pretty much exactly what I had in mind with that comment. I thought this was well-known since that's basically how all bitflags/bitfield crates work, and those are by far the most prominent use case for the sort of thing this RFC is about.

pickfire:
> And in both of those examples, I can't recall anyone claiming they put such types into an enum and then suffered from a missing niche optimization.

@lxrec so this means one should not use the `bool` in standard library but instead use some more optimized `bool` for memory in some other crates? I think this would make more crates such as https://github.com/maciejhirsz/beef appear to be a more optimized version of the stuff available in standard library. Maybe it would be cool to see multiple copy of the standard library as a crate built for different reasons, such as `async-std` (built for async but it's already here) or built for speed or built for memory.

Ixrec:
There isn't really such a thing as a "more optimized `bool`". What you'd be optimizing is a larger data structure which semantically is made up of bools, e.g. because you want each "logical bool" to take up only one bit of actual memory. In other words, bitfields.

In general there will always be lots of crates offering variations on stuff in `std`, because there's a million ways to implement these things with various tradeoffs that are occasionally relevant to someone. However, async-std doesn't really have anything to do with this RFC. and IIUC beef is only tangentially relevant to this RFC since it's not so much exploiting niches in user-defined types as it is a) exploiting pointer layout guarantees and b) getting away with smaller types by exploiting the fact that you almost never need `String`/`Vec` lengths/capacities that go all the way up to `usize::MAX`. You could argue that a) should happen by default in the language, but raw pointers are already a primitive type, so that could be done with a far more limited proposal than this.

burdges:
There are parallels between bitfield access and `Cell` so maybe one should not do fancy `*Guard` types for bitfield types.

As an aside, we'd save more memory and better improve cache locality with NRVO ala https://github.com/rust-lang/rfcs/pull/2884

kennytm:
(#2884 did *not* propose NRVO, just plain RVO. Also you don't need #2884 for the optimizer to apply NRVO.)

tommit:
> You should check if any proc macro crates for bitfields and/or packed enum create some `Guard` type ala `MutexGuard` that you explicitly borrow from the bitfield and/or packed enum. I'd expect some `Guard` type would remain the only way to pack this type into 1 byte because you must handled mutated data when taking references:
> 
> ```
> enum Foo {
>     A(U8LessThan64>
>     B(U8LessThan64>
>     C(U8LessThan128>
> }
> ```

I admit I don't know what a `Guard` is nor how references are relevant here. But you mention that the size of `Foo` in your example could somehow be made 1 byte, and I can and want to say something about that.

First of all, as we can see from the code snippet below, the compiler doesn't optimize the size of `Bar` (a type similar to `Foo` in your example) to be 1 byte even when it could do so. This RFC doesn't suggest that the compiler should optimize the size of `Bar` nor do I think it should.

In order to optimize the size of `Bar`, the compiler would need to set some of the forbidden bits of `U8LessThan2` and `U8LessThan4`in a `Bar` variable to indicate which enum variant (`A`, `B`, or `C`) the variable contains. And then, in order to retrieve the value in the field of the variant, the compiler would need to mask those forbidden bits out.

This is clearly a tradeoff between size (of `Bar`) and performance (pessimizing performance by adding an extra bitmasking operation) and not a pure optimization (by which I mean something that makes things strictly better, and not worse in any way). I'm not (and neither is this RFC) interested in making these kinds of tradeoffs. This RFC is interested in getting the compiler to do more of those "pure" optimizations that it is already making with types like `Option<U8LessThan2>`.

```rust
#[repr(u8)]
enum U8LessThan2 {
    _0, _1,
}

#[repr(u8)]
enum U8LessThan4 {
    _0, _1, _2, _3,
}

enum Bar {
    A(U8LessThan2),
    B(U8LessThan2),
    C(U8LessThan4),
}

const _: () = assert!(mem::size_of::<Bar>() == 2);
```

clarfon:
So, I can't actually find any good examples -- would someone be able to provide a link to some example of what the weird unsafe bitshifting looks like, as a way of creating forbidden values?

pickfire:
@clarfon Is https://github.com/maciejhirsz/beef/blob/856988b38cafa8c8f8f44f6f0923c717ce3f518b/src/lean.rs#L24-L29 considered weird enough?

tommit:
> @clarfon Is https://github.com/maciejhirsz/beef/blob/856988b38cafa8c8f8f44f6f0923c717ce3f518b/src/lean.rs#L24-L29 considered weird enough?

AFAIK, that example uses a `u64` value to store two `u32` values in order to reduce the number fields in a struct from 3 to 2 (which boosts performance due to a specific aspect of calling convention). Then it uses bit masking and shifting to access those two `u32` values embedded in the `u64` value. But I don't see how any of that is relevant to this RFC or forbidden values.

pickfire:
Is forbidden values applied to existing `NonZero*` types?

tommit:
There have been multiple comments here that mention bitfields in various contexts, and I haven't understood why people are talking about bitfields. What do bitfields have to do with this RFC? I've been thinking that it's me who doesn't understand something here because that happens a lot, but now I'm beginning to think that maybe some people have misunderstood what this RFC is trying to accomplish. And that would be clearly my fault.

So, I'll now try to explain better what this RFC is all about:

1. The short version:

Give the compiler the ability to use the forbidden values of any applicable (opted-in) user-defined type to represent data-less enum variants.

2. The long version:

There are some types that should/must never hold certain (forbidden) value(s). Here are some examples of types with forbidden values that the compiler knows about and uses to represent data-less enum variants today:

* Primitive types:
`bool`, `char`, `()`, `&T`, `&mut T`

* Standard library types:
```rust
std::{
    boxed::Box,
    num::{NonZeroI8, NonZeroU8, NonZeroI16, /* ... */},
    ptr::NonNull,
}
```

* All unitary (i.e. data-less, c-like) enum types whose number of variants is less than the number of different values their internal representation can have. For example:
```rust
enum UnitaryEnum {
    A, B, C, D,
}
```
(I believe that list is not exhaustive)

Enum types can have both data-less variants and variants with data attached to them. Here's an example:
```rust
enum EnumType<T> {
    VariantWithData(T),
    VariantWithoutData1,
    VariantWithoutData2,
}
```

If you instantiate an `EnumType<Foo>` with type `Foo` that has at least two forbidden values that the compiler knows about, then the compiler doesn't create a separate tag (an integer) field inside `EnumType<Foo>` for indicating which variant it currently contains, but rather it stores only a `Foo` inside `EnumType<Foo>` and uses some two different forbidden values of `Foo` to indicate that `EnumType<Foo>` contains either `VariantWithoutData1` or `VariantWithoutData2`, and it uses all the non-forbidden values of `Foo` to indicate that it contains a `VariantWithData(foo_value)`. This optimization not only makes the size of `EnumType<Foo>` smaller than it would be with a separate tag, but it may result in better performance also due to the fact that the value of `Foo` that is being read is already in a register after the cpu has verified that `EnumType<Foo>` contains the `VariantWithData` variant.

Currently there are only two ways a user-defined type `Foo` can be such that the compiler knows it has forbidden value(s) (and what those values are) and then uses those values to perform the aforementioned optimization whenever possible: either `Foo` is a unitary enum type or it is a type that contains a field of some type the compiler already knows has forbidden value(s). The goal of this RFC is to be able to tell the compiler about forbidden values of user-defined types it currently doesn't can cannot know about.

programmerjake:
> There have been multiple comments here that mention bitfields in various contexts, and I haven't understood why people are talking about bitfields. What do bitfields have to do with this RFC? I've been thinking that it's me who doesn't understand something here because that happens a lot, but now I'm beginning to think that maybe some people have misunderstood what this RFC is trying to accomplish. And that would be clearly my fault.

The reason several people have mentioned bitfields and bitmasks is because some kinds of forbidden values are not one or a few ranges of values, instead, (like `&u32`) they have tons of forbidden values that are way easier to express using a bitmask:
For example, `&u32` has forbidden values for address `0` (the null pointer) and for every address that is not a multiple of `4` (because `u32` must be aligned to a multiple of `4` bytes). Expressing those forbidden values is easily done using:
* The value `0`
* The bit mask expression `v & 0x3 == 0x1`
* The bit mask expression `v & 0x3 == 0x2`
* The bit mask expression `v & 0x3 == 0x3`


Now that I think about it, both kinds of forbidden value specifications could be combined into a single kind:
Any value `v` is a forbidden value when `(v & mask) >= low && (v & mask) <= high` for at least one element in a list of `mask`, `low`, `high` triples.

```rust
// replace as necessary to avoid depending on const generics if that causes problems:
struct ForbiddenValue<const N: usize> {
    mask: [u8; N],
    low: [u8; N],
    high: [u8; N],
}

impl<const N: usize> ForbiddenValue<N> {
    const fn is_forbidden(&self, value: &[u8; N]) -> bool {
        let mut cmp_low = Ordering::Equal;
        let mut cmp_high = Ordering::Equal;
        // would need to switch for into a while loop since const for doesn't currently work
        for i in 0..N {
            // compare in little endian order -- could switch to big endian by reversing iteration order
            // or match native endian
            let masked_byte = self.mask[i] & value[i];
            match masked_byte.cmp(&self.low[i]) {
                Ordering::Equal => {}
                v => cmp_low = v,
            }
            match masked_byte.cmp(&self.high[i]) {
                Ordering::Equal => {}
                v => cmp_high = v,
            }
        }
        cmp_low != Ordering::Less && cmp_high != Ordering::Greater
    }
}

**Comment by tommit** [comment]
I changed the design (in the RFC) a bit by removing the generic parameters because they would allow the trait to be implemented multiple times for a specific type.

Then it occurred to me, that it is already possible with the current design to define a large number of forbidden values easily by using a `const` block like in the following: (notice that this wouldn't be possible if `ForbiddenValues::FORBIDDEN` was a slice)
```rust
unsafe trait ForbiddenValues {
    type Forbidden: FixedSizeArray<[u8; mem::size_of::<Self>()]>;

    const FORBIDDEN: Self::Forbidden;
}

unsafe impl ForbiddenValues for u32 {
    type Forbidden = [[u8; 4]; 50];

    const FORBIDDEN: [[u8; 4]; 50] = {
        let mut forbidden = [[0u8; 4]; 50];
        let mut item = 1u32;
        let mut i = 0;

        while i < 50 {
            array[i] = unsafe { mem::transmute(item) };
            item += 128;
            i += 1;
        }
        forbidden
    };
}
```

And I think it's unlikely that the compiler would need a huge number of forbidden values to perform its optimizations, so a reasonable number like 50 would probably suffice.

Another thing to note is that I believe that these (potentially large) `FORBIDDEN` arrays are completely compiled away from the final binary, as long as user doesn't access them.

programmerjake:
> And I think it's unlikely that the compiler would need a huge number of forbidden values to perform its optimizations, so a reasonable number like 50 would probably suffice.

There are some optimizations that could be done in the future that would need large numbers of forbidden values, such as packing values into bitfields, so, to prevent getting stuck with an API that is really unsuited for those use cases, I think we should plan ahead by using an API that can succinctly express all forbidden values, especially including those that require bitmasks to describe succinctly (such as the type `MyFakeReferenceType` described in https://github.com/rust-lang/rfcs/pull/2888#issuecomment-628789625)

tommit:
> There are some optimizations that could be done in the future that would need large numbers of forbidden values, such as packing values into bitfields

@programmerjake Could you elaborate a bit on what you mean by that?

And I'm going to anticipate your answer now, and comment a bit on it beforehand. Tell me if I'm wrong, but this is what I assume you meant by that:

There will be some bitfield-type that gives the user the impression (by doing bitwise operations behind the scenes) that one can store types of arbitrary bit widths inside it. In order to store a value of type `T` in, let's say, certain 6 bits of that bitfield-type, the compiler must verify that all the values of `T` fit in 6 bits (and which bits they are stored in `T`). You call that an optimization, but packing values into smaller types by doing bitwise operations is not an optimization, it is a tradeoff. Like I said before, I still maintain that bitfields have nothing to do with this RFC.

Another point I want to make, or rather suggest, is that we should assume that there will be a bounded integer type in the standard library at some point, and that the compiler will implicitly know about all the forbidden values of those types just like it does today with unitary enums. Such bounded integer types will remove a lot of the use cases for the `ForbiddenValues` trait in this RFC, including the use case for verifying that a (bounded integer) value fits in some "field" of some bitfield-type.

tommit:
> I think we should plan ahead by using an API that can succinctly express all forbidden values

@programmerjake I'd like to point out that the design you [described](https://github.com/rust-lang/rfcs/pull/2888#issuecomment-628789625) cannot express all forbidden values. For example:
```rust
// Contract:
// For each value `f` of type `Foo`, `f.start <= f.end`
struct Foo {
    start: u32,
    end: u32,
}
```

The [iterator based design](https://github.com/rust-lang/rfcs/pull/2888#issuecomment-627676454) is the one that can express all forbidden values (not succinctly though). And it has the caveat that it may not be able to use `std::iter::Iterator` due to the `impl const` requirement, so there would need to be some new alternative iterator trait, like:
```rust
pub trait AlternativeIterator {
    type Item;

    const LEN: usize;

    fn next(&mut self) -> Option<Self::Item>;
}
```

The current design is simple & not robust, and the iterator based design is complicated & robust.  I think they represent the two extremes of the design space. In order to navigate that space, it would be good to get the compiler to tell statistics about the maximum and the average number of forbidden values it uses to perform optimizations.

Evrey:
Just to add perspective: I'm not just interested in more efficient `enum` packing. I'm interested in statically proving things about my data. So if I can't exhaustively and conveniently express a large number of invalid states, this RFC is — for me — useless.

tommit:
> I'm interested in statically proving things about my data.

@Evrey  Can you elaborate on what you mean by that? I'm pretty sure "statically" means "at compile-time". But other than that, I just don't understand what you mean.

> So if I can't exhaustively and conveniently express a large number of invalid states, ...

I take "exhaustively" to mean that you need to be able to express all invalid states. You also use the word "conveniently" - do you consider the external iterator based design convenient enough?

jhpratt:
@Evrey Note that the trait is `unsafe`. The compiler will make assumptions based on what you tell it: it won't have to prove anything. Having to statically prove certain values aren't possible is extremely difficult if not impossible, let alone extraordinarily time consuming.

burdges:
We'll need compiler plugins for formal verification tools like refinement types because rustc performance would suffer if random crates started using those tools everywhere.  

I'm kinda curious just how much performance hit we already take from generic-array, which winds up being an _unecessary_ dependency of 500+ crates.

Evrey:
@tommit One example I repeatedly brought up is expressing `start <= end` as an invariant for ranges. But there's more, like `len <= cap` for a vector, etc.

As for the iterator: It still passes around byte buffers, which have the big endianess issue. And iterating over 16.8 million `NaN` values is… uhh… let's just say compile times will be horrible. Let's also imagine writing an `f32` wrapper that only allows finite values in `0.0 ..= 1.0`. Pretty useful for a lot of things. And now imagine the amount of invalid compared to valid state. And when iterating over all that, I'll need quite the complicated state machine, first iterating over out-of-range values, the infinities, and finally all the `NaN`s.

@jhpratt Well, yes, like with `NonZeroU32` and friends initialisation will be dynamically checked. But when extracting the value, the compiler is informed about the given invariants and optimises accordingly.

@burdges Very true. But on the other hand, whoever uses refinement types (and I see this RFC as an opportunity for pretty much »poor man's refinement types«) or even full formal verification of the code is already okay with long compile times.

jhpratt:
Yeah, it makes an assumption about the data because you said it could, basically. I think we're sort of saying the same thing.

text/0000-forbidden-value-trait.md
```diff
@@ -23,12 +23,13 @@ unsafe trait ForbiddenValues {
...
 
-To implement `ForbiddenValues` for type `T`, the following condition must be met:
+To implement `ForbiddenValues` for type `T`, the following conditions must be met:
 
-For each value `t` of type `T`, and for each item `f` in `T::FORBIDDEN`:
+1. For each value `t` of type `T`, and for each item `f` in `T::FORBIDDEN`:
 ```rust
 *(&raw const t as *const [u8; size_of::<T>()]) != f
 ```
+2. `T` must not be a type that has forbidden values the compiler already knows about (e.g. `char`)
```
pickfire: > e.g. `char`

How about other types, are there any other types that the compiler already knows about? Is it possible to lists them out?

Ixrec: Not sure how relevant this is now that the issue's closed, but https://doc.rust-lang.org/std/#primitives is a list of primitive types in the language. Which of these counts as "having forbidden values" depends on what you mean by that. For example, "uninitialized" is a forbidden / insta-UB-causing value for all of them, and `str` has to be valid UTF-8 within safe code but unsafe code can violate that without triggering UB, and I don't know if you'd want to include library types like `NonZero*`, and so on and so forth. But if we try to just ignore all of this scope vagueness, I think `bool` is the only other primitive type that is "like `char`" insofar as it clearly has to be implemented as one of the integer types with certain values excluded (in `bool`'s case, every value except 0 or 1).

PR #2895, opened 2020-04-01
RFC: Generic member access for dyn Error trait objects

yaahc:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/2895)*

[**Rendered**](https://github.com/yaahc/rfcs/blob/master/text/0000-dyn-error-generic-member-access.md)

Also, I've mocked up the proposal in this repository - https://github.com/yaahc/nostd-error-poc/

[Rendered](https://github.com/yaahc/rfcs/blob/master/text/0000-dyn-error-generic-member-access.md)


text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,335 @@
...
+# Summary
+[summary]: #summary
+
+This RFC proposes a pair of additions to the `Error` trait to support accessing
+generic forms of context from `dyn Error` trait objects, one method on the
+`Error` trait itself for returning references to members based on a given type
+id, and another fn implemented for `dyn Error` that uses a generic return type
+to get the type id to pass into the trait object's fn. These functions will act
+as a generalized version of `backtrace` and `source`, and would primarily be
+used during error reporting when rendering a chain of opaque errors.
```
anp, line 15:
```suggestion
This RFC proposes two additions to the `Error` trait to support accessing
generic forms of context from `dyn Error` trait objects, generalizing the
pattern used in `backtrace` and `source` and allowing ecosystem iteration
on error reporting infrastructure outside of the standard library. The two
proposed additions are a new trait method `Error::provide_context` which
offers `TypeId`-based member lookup and a new inherent fn
`<dyn Error>::context` which makes use of an implementor's `provide_context`
to return a typed reference directly.

These additions would primarily be useful in "error reporting" contexts where we
typically no longer have type information and may be composing errors from many
sources.
```

IMO the definitions themselves are very readable as an "intro" to what you're proposing if you put them together:

```rust
pub trait Error {
    /// Provide an untyped reference to a member whose type matches the provided `TypeId`.
    ///
    /// Returns `None` by default, implementors are encouraged to override.
    fn provide_context(&self, ty: TypeId) -> Option<&dyn Any> {
        None
    }
}

impl dyn Error {
    /// Retrieves a reference to `T`-typed context from the error if it is available.
    pub fn context<T: Any>(&self) -> Option<&T> {
        self.provide_context(TypeId::of::<T>())?.downcast_ref::<T>()
    }
}
```

anp (commented):
(didn't make it very far before my day filled up again, will return shortly)

yaahc:
@burdges this is a general restriction on Any and downcast on Error types, I don't think it's currently possible to downcast types that aren't : 'static

The second design with the a Provider miiight work with this, not sure.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+implementing the Error trait. For accessing `dyn Error` members it provides the
+`source` function, which conventionally represents the lower level error that
+caused the current error. And for accessing a `Backtrace` of the state of the
+stack when an error was created it provides the `backtrace` function. For all
+other forms of context relevant to an Error Report the error trait provides the
+`context`/`provide_context` functions.
+
+As an example of how to use these types to construct an error report lets
+explore how one could implement an error reporting type that retrieves the
+Location where each error in the chain was created, if it exists, and renders
```
anp, line 89: nit: backticks for Location

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+it as part of the chain of errors. Our end goal is to get an error report that
```
anp, line 90: I'd start the walkthrough with the motivating output, much better headline

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+
+```
+Error:
+    0: Failed to read instrs from ./path/to/instrs.json
+        at instrs.rs:42
+    1: No such file or directory (os error 2)
+```
+
+The first step is to define or use a Location type. In this example we will
+define our own but we could use also use `std::panic::Location` for example.
```
anp, line 101: Ah, I might be the only person in the world who finds this really confusing but this really tripped me up (I've been staring at core::panic::Location for months)

anp: Specifically what tripped me up was ending this sentence referencing std but then still defining our own. I'd pick one or the other and run with it.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+There are two additions necessary to the standard library to implement this
+proposal:
+
+Add a function for dyn Error trait objects that will be used by error
+reporters to access members given a generic type. This function circumvents
```
anp, line 192: s/generic//

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+Add a member to the `Error` trait to provide the `&dyn Any` trait objects to
+the `context` fn for each member based on the type_id.
+
+```rust
+trait Error {
+    /// ...
+
+    fn provide_context(&self, id: TypeId) -> Option<&dyn Any> {
```
anp, line 223: It occurs to me reading this that it seems a little more common to name this kind of thing `get_context` in Rust but IDK.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* The API for defining how to return types is cumbersome and possibly not
+  accessible for new rust users.
+    * If the type is stored in an Option getting it converted to an `&Any` will
+      probably challenge new devs, this can be made easier with documented
+      examples covering common use cases and macros like `thiserror`.
+```rust
```
anp, line 249: Is this misplaced?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+} else if typeid == TypeId::of::<SpanTrace>() {
+    self.span_trace.as_ref().map(|s| s as &dyn Any)
+}
+```
+* When you return the wrong type and the downcast fails you get `None` rather
```
anp, line 254: this could be multiple sentences or have more punctuation or...

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+  than a compiler error guiding you to the right return type, which can make it
+  challenging to debug mismatches between the type you return and the type you
+  use to check against the type_id
+    * The downcast could be changed to panic when it fails
```
anp, line 258: I'd call this an unknown rather than a drawback "how should the `context` fn handle failed downcasts?"

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+    * There is an alternative implementation that mostly avoids this issue
```
anp, line 259: ?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+   `source.provide` first which would prevent the error reporter from knowing
+   which error in the chain gathered each piece of context and might cause
+   context to show up multiple times in a report.
+
+# Prior art
+[prior-art]: #prior-art
+
+I do not know of any other languages whose error handling has similar
+facilities for accessing members when reporting errors. For the most part prior
+art exists within rust itself in the form of previous additions to the `Error`
```
anp, line 329: Java's Error casting can kinda do this, you can attempt to cast to the interface that has what you want and see what happens.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- What should the names of these functions be?
+    - `context`/`context_ref`/`provide_context`
+    - `member`/`member_ref`
+    - `provide`/`request`
+- Should we go with the implementation that uses `Any` or the one that supports
+  accessing dynamically sized types like traits and slices?
```
anp, line 340: Another question I have: should a "probe a trait object by TypeId and/or Symbol" be a language feature in itself? i.e. you're proposing a constrained form of reflection (if you squint) -- what are the arguments against a more general form of it for any `dyn Trait`?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+    - I bring this up specifically for the case where you want to use this
+      function to get an `Option<&[&dyn Error]>` out of an error, in this case
+      its unlikely that the error behind the trait object is actually storing
+      the errors as `dyn Errors`, and theres no easy way to allocate storage to
+      store the trait objects.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+I'd love to see the various error creating libraries like `thiserror` adding
```
anp, line 351: Prefer "the author" or "we" to "I" in something that'll become a normative text, IMO

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+support for making members exportable as context for reporters.
+
+Also, I'm interested in adding support for `Error Return Traces`, similar to
+zigs, and I think that this accessor function might act as a critical piece of
```
anp, line 355: It might be worth writing a blog post companion to this which sketches out the possibility?

anp (commented):
I focused my feedback more on "technical writing" as requested, but in general I'm quite excited to see this proposal move forward! There are a number of cases at work where it would be nice to specialize error presentation in this way. At a high level I think this proposal would benefit from a "why not more general, i.e. all trait objects?" section. I also recommend putting more examples closer to the top of the "error reports" that you'd like to enhance -- they're quite motivating IMO.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Example use cases this enables
+
+* using `backtrace::Backtrace` instead of `std::backtrace::Backtrace`
+* zig-like Error Return Traces by extracting `Location` types from errors
+  gathered via `#[track_caller]` or some similar mechanism.
+* error source trees instead of chains by accessing the source of an error as a
+  slice of errors rather than as a single error, such as a set of errors caused
+  when parsing a file
+* `SpanTrace` a backtrace like type from the `tracing-error` library
```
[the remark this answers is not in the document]
yaahc: I think that might be more detail than is necessary for the RFC, I've merged it with the "alternatives to std::backtrace" bullet and it has a docs.rs link

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,356 @@
...
+By adding a generic form of these functions that works around the restriction
+on generics in vtables we could support a greater diversity of error handling
+needs and make room for experimentation with new forms of context in error
+reports.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Error handling in rust consists mainly of two steps, creation/propogation and
+reporting. The `std::error::Error` trait exists to bridge the gap between these
```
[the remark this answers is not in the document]
yaahc: Not that I can think of, the only other purpose it has is downcasting, so if you wanted a type that could contain an open set of errors where the only thing you do with them is downcast them back to concrete types you'd just use `dyn Any`.

text/0000-dyn-error-generic-member-access.md
yaahc: Do you mean move this to the top of the guide level explanation? or to motivation / summary?

text/0000-dyn-error-generic-member-access.md
yaahc: no, but I wasn't really sure how to put a code snippet in a bullet list

text/0000-dyn-error-generic-member-access.md
yaahc: Nika has a suggestion below, I really don't know which one is preferrable, they're both good, nika's has serious advantages, but it's also more complicated.

yaahc: Maybe I should just bite the bullet and make Nika's the primary suggestion and leave the `Any` version as the alternate, because its much shorter.

text/0000-dyn-error-generic-member-access.md
yaahc: I'll look into this, ty for the heads up

text/0000-dyn-error-generic-member-access.md
yaahc: :thinking:

text/0000-dyn-error-generic-member-access.md
yaahc: Good idea :D

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,459 @@
...
+forms of context from `dyn Error` trait objects. This generalizes the pattern
+used in `backtrace` and `source`. This proposal adds the method
+`Error::get_context` to the error trait, which offers `TypeId`-based member
+lookup, and a new inherent fn `<dyn Error>::context`, which makes use of an
+implementor's `get_context` to return a typed reference directly. These
+additions would primarily be useful in "error reporting" contexts, where we
+typically no longer have type information and may be composing errors from many
+sources.
+
+_note_: This RFC focuses on the more complicate of it's two proposed solutions
```
mark-i-m, line 19:
```suggestion
_note_: This RFC focuses on the more complicated of it's two proposed solutions
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,471 @@
...
+## TLDR
+
+Add this method to the `Error` trait
+
+```rust
+pub trait Error {
+    // ...
+
+    /// Provides an object of type `T` in response to this request.
+    fn get_context<'r, 'a>(&'a self, request: Request<'r, 'a>) -> ProvideResult<'r, 'a> {
```
kennytm, line 32: Ref https://github.com/rust-lang/rfcs/pull/2895#discussion_r403736887 @anp,

I find it strange that we have both `context()` function (a getter, sort of) and a `get_context()` function (sounds like a getter, but it's not). Keeping it `provide_context()` maybe better.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,471 @@
...
+}
+
+/// Marker type indicating a request has been fulfilled.
+pub struct FulfilledRequest(PhantomData<&'static Cell<()>>);
+
+/// Provider method return type.
+///
+/// Either `Ok(Request)` for an unfulfilled request, or `Err(FulfilledRequest)`
+/// if the request was fulfilled.
+pub type ProvideResult<'r, 'a> = Result<Request<'r, 'a>, FulfilledRequest>;
```
kennytm, line 339: This looks really odd that `Err` is the "happy case" especially as a public type. If the purpose is just to use `?` chaining it is probably better to just create an `impl Try` type like [LoopState](https://github.com/rust-lang/rust/blob/2454a68cfbb63aa7b8e09fe05114d5f98b2f9740/src/libcore/iter/mod.rs#L375).

```rust
enum Answer<T, S> {
    Found(T),
    Continue(S),
}
```

kennytm:
One thing I'm uneasy with reusing the [`object-provider` API](https://docs.rs/object-provider/0.1.0/object_provider/) is the double lifetime:

```rust
fn get_context<'r, 'a>(&'a self, request: Request<'r, 'a>) -> ProvideResult<'r, 'a> {  ...  }
```

Having one explicit lifetime can still be managed, but having two or more correlated lifetimes becomes mentally challenging.

IMO we shouldn't need to copy the `object-provider` API entirely.

Firstly, the return type `ProvideResult<'r, 'a>` is unnecessarily complex in order to support `?`. This type actually only has 2 states: fulfilled or not. This is isomorphic to a bool. So why not just make it a `bool`?

```rust
fn get_context<'a>(&'a self, request: Request<'_, 'a>) -> bool {
    request.provide::<Backtrace>(&self.backtrace) ||
        request.provide::<SpanTrace>(&self.span_trace) ||
        request.provide::<dyn Error>(&self.source) ||
        request.provide::<Vec<&'static Location<'static>>>(&self.locations) ||
        request.provide::<[&'static Location<'static>]>(&self.locations)
}
```

But type safety maybe a concern (e.g. user might have used `&&` instead of `||`), so we could absorb that `bool` into the `Request` and use the builder pattern

```rust
fn get_context<'_, 'a>(&'a self, request: &mut Request<'_, 'a>) {
    request
        .provide::<Backtrace>(&self.backtrace)
        .provide::<SpanTrace>(&self.span_trace)
        .provide::<dyn Error>(&self.source)
        .provide::<Vec<&'static Location<'static>>>(&self.locations)
        .provide::<[&'static Location<'static>]>(&self.locations);
}
```

In fact, unless you're only want to keep the first `T1` in `request.provide::<T1>(&t11).provide::<T1>(&t12);`, even the boolean check is unnecessary, since LLVM is able to optimize that chain involving lots of `if TypeId::of::<T>() == self.type_id()` into a switch table.

Secondly, `Request<'r, 'a>` has two lifetimes essentially because it represents a `&'r mut &'a Any`. Perhaps we could shave off the outer reference? It is actually possible through an `extern type`. This further simplifies `get_request` into a single lifetime

```rust
fn get_context<'a>(&'a self, request: &mut Request<'a>) {
    ...
}
```

The diff is a bit long, I've pushed it to https://github.com/yaahc/nostd-error-poc/compare/yaahc:eaa45d4...kennytm:dbd5f03?w=1 for reference.

yaahc:
There was a specific soundness issue which caused @mystor to add the double lifetime, though I don't remember what it was. Overall I'm definitely in favor of removing the double lifetime if possible but we should check with nika first probably.

yaahc:
> Firstly, the return type `ProvideResult<'r, 'a>` is unnecessarily complex in order to support `?`. This type actually only has 2 states: fulfilled or not. This is isomorphic to a bool. So why not just make it a `bool`?

entirely in favor of simplifying the result type 

> In fact, unless you're only want to keep the first `T1` in `request.provide::<T1>(&t11).provide::<T1>(&t12);`, even the boolean check is unnecessary, since LLVM is able to optimize that chain involving lots of `if TypeId::of::<T>() == self.type_id()` into a switch table.

I do feel it is important to keep the explicit type parameter rather than manually doing the type id comparisons, assuming that's what you mean here. By using the explicit type parameter you get compile errors when they mismatch and it works as a coercion site which makes the API more convenient to use and less error prone overall.

kennytm:
@yaahc I agree that it's not a good idea to manually compare `TypeId`. What I mean is that, after all inlining, the *generated* code in `get_context()` will look like 

```rust
if request.type_id() == TypeId::of::<T1>() {
    /* code for T1 */
}
if request.type_id() == TypeId::of::<T2>() {
    /* code for T2 */
}
if request.type_id() == TypeId::of::<T3>() {
    /* code for T3 */
}
...
```

which LLVM will notice that all the `type_id` is immutable and optimize into

```
match request.get_type_id() {
    const { TypeId::of::<T1>() } => { /* code for T1 */ }
    const { TypeId::of::<T2>() } => { /* code for T1 */ }
    const { TypeId::of::<T3>() } => { /* code for T1 */ }
    ...
}
```

and thus the short-circuit property of `?` and `||` are not necessary to produce the optimal code.

mystor:
As I mention in that old comment, the `&mut` approach also requires a second lifetime, but it's hidden on the `&mut` itself. The original approach in the `object-provider` crate actually used a similar design.

The main drawback with the bare `&mut Request` type is unsoundness due to `mem::swap`. Take, for example, the following code, which can fulfill the outer request with a `NonZeroUsize`, despite it requesting a `&'static u8` (untested, but I believe something similar can be written):

```rust
let x: NonZeroUsize = ...;
let bad_pointer = Request::with::<&'static u8>(|ra| {
    Request::with::<NonZeroUsize>(|rb| {
        mem::swap(ra, rb);
    });
    ra.provide::<NonZeroUsize>(&x);
}).unwrap();
// `bad_pointer` is invalid
```

I believe this unsoundness can be avoided by using `Pin`, and some of the `PhantomData` shenanigans to make the `'out` lifetime invariant aren't necessary when it's behind a `&mut`, so it's at least a bit simpler to use.

```rust
#[repr(C)]
 pub struct Request<'out> {
    type_id: TypeId,
    _pinned: PhantomPinned,
    _marker: PhantomData<&'out ()>,
}

// ...
fn get_context<'a>(&'a self, request: Pin<&mut Request<'a>>) {
    ...
}
```

The remaining downside with this simplified `Pin`-based design is that it doesn't provide any sort of of single-use guarantee, but I think that's an acceptable trade-off for the perceived simplicity, especially as most call-sites will chain `.provide` method calls.

kennytm:
@yaahc  @mystor Yes there is an implicit second lifetime in `&mut Request<'a>`, but being implicit makes it easier to deal with — you could mostly ignore that lifetime when reading.

I think we could workaround the unsoundness by making `Request::with` *private* (`pub(crate)`), since `Request::with` is only used in `<dyn Error>::context`. This ensures there cannot never be two `Request` instances simultaneously in the same thread.

But if we have to keep `Request::with` public, I agree that the `Pin`-based design is the best solution.

(Or we should restart the progress on #1861 so we could add a [thin unsized tail](https://github.com/yaahc/nostd-error-poc/blob/dbd5f0318202e58c388aae90530cc6e5e2e9287a/fakecore/src/any.rs#L11-L13) to `Request`, which also properly makes `mem::swap` not callable.)

mystor:
Adding something like `core::any::Request`, and making it's soundness require that it is only used by `Error` seems unfortunate to me, so I'd greatly prefer the `Pin` solution, even if we choose not to initially stabilize a method like `Request::with`.

Adding a thin unsized tail would also be nice, and this exact sort of use-case is why I was interested in #1861 and #1993. I worry a bit about creating a dependency on solving thin unsized types for this RFC, though, as it has proven to be difficult to solve in a satisfying manner.

yaahc:
I lean towards preferring a design that makes `Request` generally usable, but I otherwise agree with all of your changes @kennytm, so I'm going to pull in your branch and add Pin and update the RFC to match.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -228,24 +226,31 @@ Here is the implementation for the proof of concept, taken from Nika Layzell's
...
+impl<'out> Request<'out> {
     /// Provides an object of type `T` in response to this request.
     ///
     /// Returns `Err(FulfilledRequest)` if the value was successfully provided,
     /// and `Ok(self)` if `T` was not the type being requested.
     ///
     /// This method can be chained within `provide` implementations using the
     /// `?` operator to concisely provide multiple objects.
-    pub fn provide<T: ?Sized + 'static>(self, value: &'out T) -> ProvideResult<'r, 'out> {
+    pub fn provide<T: ?Sized + 'static>(&mut self, value: &'out T) -> &mut Self {
```
kennytm, line 256:
```suggestion
    pub fn provide<T: ?Sized + 'static>(self: Pin<&mut Self>, value: &'out T) -> Pin<&mut Self> {
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -259,84 +264,66 @@ impl<'r, 'out> Request<'r, 'out> {
     ///
     /// This method can be chained within `provide` implementations using the
     /// `?` operator to concisely provide multiple objects.
-    pub fn provide_with<T: ?Sized + 'static, F>(mut self, cb: F) -> ProvideResult<'r, 'out>
+    pub fn provide_with<T: ?Sized + 'static, F>(&mut self, cb: F) -> &mut Self
```
kennytm, line 270:
```suggestion
    pub fn provide_with<T: ?Sized + 'static, F>(self: Pin<&mut Self>, cb: F) -> Pin<&mut Self>
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -228,24 +226,31 @@ Here is the implementation for the proof of concept, taken from Nika Layzell's
...
-/// reference.
-pub struct Request<'r, 'out> {
-    buf: NonNull<TypeId>,
-    _marker: PhantomData<&'r mut &'out Cell<()>>,
-}
-
-impl<'r, 'out> Request<'r, 'out> {
+/// `'out` is the lifetime of the requested reference.
+#[repr(transparent)]
+pub struct Request<'out>(RequestBuf<PhantomData<&'out Cell<()>>>);
```
kennytm, line 242: need to include a `PhantomPinned` somewhere (maybe inside `RequestBuf` for simplicity), to avoid this type becoming `Unpin`.

text/0000-dyn-error-generic-member-access.md
yaahc: this line no longer compiles when you apply this change.

```rust
            let this = unsafe { &mut *(self as *mut _ as *mut RequestBuf<Option<&'out T>>) };
```

do you wanna try applying these changes in https://github.com/yaahc/nostd-error-poc/ first and open a PR there, once it's merged there I'll copy the changes into the RFC.

mystor:
@kennytm @yaahc I actually wrote a branch for a potential `0.2.0` version of the `object-provider` crate which also uses `Pin<&mut T>` internally. I think there are a few behaviour differences between the two. https://github.com/mystor/object-provider/tree/pin_variant

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+  gathered via `#[track_caller]` or similar.
+* error source trees instead of chains by accessing the source of an error as a
+  slice of errors rather than as a single error, such as a set of errors caused
+  when parsing a file TODO reword
+* Help text such as suggestions or warnings attached to an error report
+
+## Moving `Error` into `libcore`
+
+Adding a generic member access function to the `Error` trait and removing the
+`backtrace` function would make it possible to move the `Error` trait to libcore
```
anp, line 130: I had forgotten that `Error::backtrace` was still unstable -- might be worth making that explicit.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+without losing support for backtraces on std. The only difference being that
+in places where you can currently write `error.backtrace()` on nightly you
+would instead need to write `error.context::<Backtrace>()`.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Error handling in Rust consists of three steps: creation/propagation, handling,
+and reporting. The `std::error::Error` trait exists to bridge the gap between
+creation and reporting. It does so by acting as a interface that all error
```
anp, line 140:
```suggestion
creation and reporting. It does so by acting as an interface that all error
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+types can implement that defines how to access context intended for error
+reports, such as the error message, source, or location it was created. This
+allows error reporting types to handle errors in a consistent manner when
+constructing reports for end users while still retaining control over the
+format of the full report.
+
+The `Error` trait accomplishes this by providing a set of methods for accessing
+members of `dyn Error` trait objects. It requires that types implement the
+display trait, which acts as the interface to the main member, the error
```
anp, line 149:
```suggestion
`Display` trait, which acts as the interface to the main member, the error
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+message itself.  It provides the `source` function for accessing `dyn Error`
+members, which typically represent the current error's cause. It provides the
+`backtrace` function, for accessing a `Backtrace` of the state of the stack
```
anp, line 152: Is it worth mentioning that backtrace is unstable?

anp: FWIW realizing that this could subsume `Error::backtrace` is waaaay cool.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+            ind += 1;
+            current_error = error.source();
+        }
+
+        Ok(())
+    }
+}
+```
+
+As you can see the `Error` trait provides the facilities needed to create error
```
anp, line 250: I think you could delete this sentence without losing your point, or maybe make it more specific, referencing the specific portions of the example that highlight this?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+reports enriched by information that may be present in source errors.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The following changes need to be made to implement this proposal:
+
+### Add a type like [`Request`] to core
```
anp, line 258: Is it possible to describe the type in terms of its role, rather than in terms of another type?

"Add a `Request` type to core for providing type-indexed context"? IDK

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+
+This type fills the same role as `&dyn Any` except that it supports other trait
```
anp, line 260: This seems like an addendum to a more "central" description of `Request`, which I imagine could take the form of Any's docs: "A trait to emulate dynamic typing."

What about "`Request` is a type to emulate nested dynamic typing." ?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+objects as the requested type.
+
+Here is the implementation for the proof of concept, based on Nika Layzell's
+[object-provider crate]:
+
+A usable version of this is available in the [proof of concept] repo under
```
anp, line 266: Would a WIP PR with a couple example tests be too much? Seems like the next step up from having made a functioning proof-of-concept. IIRC there are a few RFCs where a "companion" PR was uploaded to illustrate a PoC of an approach.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+
+    fn provide_context<'a>(&'a self, _request: Pin<&mut Request<'a>>) {}
+}
+```
+
+### Use this `Request` type to handle passing generic types out of the trait object
+
+```rust
+impl dyn Error {
+    pub fn context<T: ?Sized + 'static>(&self) -> Option<&T> {
```
anp, line 414: I think it would be really useful to have this in the TLDR

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+    /// Retrieve a reference to `T`-typed context from the error if it is available.
+    pub fn context<T: Any>(&self) -> Option<&T> {
+        self.provide_context(TypeId::of::<T>())?.downcast_ref::<T>()
+    }
+}
+```
+
+### Why isn't this the primary proposal?
+
+There are two big issues with using the `Any` trait that I believe justify the
```
anp, line 471:
```suggestion
There are two significant issues with using the `Any` trait that motivate the
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+more complicated solution.
+
+- You cannot return dynamically sized types as `&dyn Any`
+- It's easy to introduce runtime errors with `&dyn Any` by either comparing to
+  or returning the wrong type
+
+By making all the type id comparison internal to the `Request` type it is
```
anp, line 478:
```suggestion
By making all the `TypeId` comparison internal to the `Request` type it is
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+impossible to compare the wrong type ids. By encouraging explicit type
```
anp, line 479:
```suggestion
impossible to compare the wrong `TypeId`s. By encouraging explicit type
```

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+parameters when calling `provide` the compiler is able to catch errors where
+the type passed in doesn't match the type that was expected. So while the API
+for the main proposal is more complicated it should be less error prone.
+
+# Prior art
+[prior-art]: #prior-art
+
+I do not know of any other languages whose error handling has similar
+facilities for accessing members when reporting errors. For the most part,
```
anp, line 488: AIUI the analogous operation in a class-inheritance language like Java would be to attempt a cast to a subclass, or am I missing something? This seems like the mix-in version of that if I squint.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+prior art for this proposal comes from within rust itself in the form of
+previous additions to the `Error` trait.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* What should the names of these functions be?
```
anp, line 495: I'm most interested in the question raised above about `std::Error::context` conflicting with `failure::Fail::context` and `anyhow::Error::context`. Have you considered a different name for the "reporter-facing" API like `Error::from_context` or `Error::request_context`?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+    * `member`/`member_ref`
+    * `provide`/`request`
+* Should there be a by-value version for accessing temporaries?
+    * We bring this up specifically for the case where you want to use this
+      function to get an `Option<&[&dyn Error]>` out of an error, in this case,
+      it is unlikely that the error behind the trait object is actually storing
+      the errors as `dyn Error`s, and theres no easy way to allocate storage to
+      store the trait objects.
+* How should context handle failed downcasts?
+    * suggestion: panic, as providing a type that doesn't match the typeid
```
anp, line 506: What's the alternative? Map it to `None` as if no context was provided?

text/0000-dyn-error-generic-member-access.md
yaahc: it could also subsume `source` fwiw :D

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,527 @@
...
+        if let Some(buf) = self.as_mut().downcast_buf::<T>() {
+            // NOTE: We could've already provided a value here of type `T`,
+            // which will be clobbered in this case.
+            *buf = Some(cb());
+        }
+        self
+    }
+
+    /// Get the `TypeId` of the requested type.
+    pub fn type_id(&self) -> TypeId {
```
anp, line 318: If your goal is to use explicit type parameters as the public API to avoid the hazards of manually-compared TypeIds, should this still be public?

text/0000-dyn-error-generic-member-access.md
yaahc: I just feel like I need some sort of conclusion, otherwise the guide ends super suddenly and it feels bad. Coming up with a good conclusion is proving difficult :(

text/0000-dyn-error-generic-member-access.md
anp: True, but we can't *delete* `source`. I guess deprecation markers have been pretty successful in the ecosystem.

text/0000-dyn-error-generic-member-access.md
yaahc: oh thats a good idea. It wouldn't be able to show the "move to core" bits for the error trait because theres the minor issue of `downcast` having `Box` in its API and needing to be in the same crate that defines `Error`, but the rest of it could be done.

One downside to that though is its a little harder to pull and use, because presumably that would be a branch of `rust-lang/rust`, so the compile times would be a lot worse and you'd have to know how to use `x.py`, which isn't hard but its definitely _harder_.

I'll probably do it though, but still treat the proof-of-concept repo as the source of truth where changes are initially made.

text/0000-dyn-error-generic-member-access.md
anp: What if you move the example output from above down here and make the "intended output" above a bit more generic?

```
Error:
    0: ERROR MESSAGE
        at LOCATION
    1: PRIOR MESSAGE
    ...
```

text/0000-dyn-error-generic-member-access.md
yaahc: yea. This could probably be used for upcasting too, you just have to return `&self`, I'm not really sure if thats prior art but ill see about mentioning similarities.

text/0000-dyn-error-generic-member-access.md
yaahc: I'm not sure I like `from_context`, but `request_context` is a good option. Though I think I'd rather conflict honestly, to encourage downstream error libraries to change their API to match the mental model for the error trait that is proposed here. In those libraries they view the entire chain of errors as one error, where each time you wrap it you're adding context. This isn't consistent with the error trait though, with the API provided here each error is distinct and the saved sources are context. So while its definitely true that errors in a chain have a context-like relationship a bunch of the error handling libraries out there model it backwards and I think that's a source of confusion in the ecosystem and part of why people complain about error handling in rust so much.

text/0000-dyn-error-generic-member-access.md
yaahc: The problem isn't when they return None, its when they return the wrong type

```rust
if TypeId::of::<SpanTrace>() == id {
   Some(&self.backtrace as &dyn Any)
}
```

This is pretty clearly a program error imo, this is only a problem with the `Any` based approach though, so I should probably remove this one. With `Request` your only options are to either do `request.provide::<SpanTrace>(&self.backtrace)` which is caught at compile time or do `request.provide(&self.backtrace)` which will just correctly compare the typeID to a `Backtrace`, which might still be wrong if you were actually trying to return a SpanTrace, like if you do `&self.backtrace` twice and meant to do `&self.span_trace` the second time. It should probably panic if it is provided the same type twice.

text/0000-dyn-error-generic-member-access.md
yaahc: This is an open question but I lean towards no, I'll add it to the unresolved questions section.

text/0000-dyn-error-generic-member-access.md
yaahc: yea, i don't really think its necessary to delete source, at a certain point it is okay to optimize your API for the happy path, and sources are by far the most common kind of context, and its only adding a dependency on its own definition, so its presence not an issue as far as moving the Error trait around is concerned afaik.

text/0000-dyn-error-generic-member-access.md
mystor: I included `type_id` and `is` in the API largely to improve the flexibility of the API for use-cases other than the `Error` trait.

A use-case which would be difficult without the `type_id` and `is` methods would be one where an interface is optionally provided, but the decision of whether or not to provide it is expensive. The `provide_with` method is required to return `&T`. With `is`, it would be possible to write code like:

```rust
if req.is::<T>() && expensive_available_check() {
    req.provide::<T>(&whatever);
}
```

Theoretically we could provide something like `try_provide_with` which takes `FnOnce() -> Option<&T>`, but generally control-flow within closures is much more limited, so providing options seems good.

If we really don't want to allow `is` and `type_id`, we could potentially expose something similar to the `downcast_buf` method (probably with a better name/API) which would ensure the types match, but allow non-closure-based control flow. Perhaps something like this:

```rust
impl Request<'a> {
    fn downcast<T: ?Sized + 'static>(&mut self) -> Option<&mut Option<&'a T>>;
}
```

It might be a good idea to not stabilize any low level APIs in the first pass, and revisit these later as needs arise.

anp: This makes sense to me - I was actually wondering about being able to hoist the branch into the implementor code just like your example. Agreed it seems less important for the first pass.

Does making `is<T>()` public require doing the same for `type_id()`? The former seems obviously useful and hard to mess up, the latter seems like the actual source of footguns.

text/0000-dyn-error-generic-member-access.md
anp: Yeah it’s more “prior niche fulfillment” than “prior art”.

text/0000-dyn-error-generic-member-access.md
mystor: It would be totally possible/reasonable to only expose `is<T>()`, and not expose `type_id()` if we're worried that it won't be used correctly

text/0000-dyn-error-generic-member-access.md
```diff
@@ -354,7 +354,7 @@ impl<'a> Request<'a> {
     }
 
     /// Get the `TypeId` of the requested type.
-    pub fn type_id(&self) -> TypeId {
+    fn type_id(&self) -> TypeId {
```
mystor, line 357: Might be worth removing the method completely and accessing `self.type_id` at any call-sites.

kennytm:
Note that @withoutboats proposed to stabilize `Error::backtrace` in rust-lang/rust#72981.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,558 @@
...
+        while let Some(error) = current_error {
+            writeln!(fmt, "    {}: {}", ind, error)?;
+
+            if let Some(location) = error.context::<Location>() {
+                writeln!(fmt, "        at {}:{}", location.file, location.line)?;
+            }
+
+            ind += 1;
+            current_error = error.source();
+        }
```
gThorondorsen, line 259: That loop is a great use case for using `self.0.chain().enumerate()` (still unstable, cc rust-lang/rust#58520).

yaahc: This can also be done pretty well with `iter::successors(some(self.0.as_ref()), |s| s.source()).enumerate();`

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,554 @@
...
+
+impl<'a> fmt::Debug for Request<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Request")
+            .field("type_id", &self.type_id)
+            .finish()
+    }
+}
+
+// Needs to have a known layout so we can do unsafe pointer shenanigans.
```
tmccombs, line 399: If rust had an offset_of operator or macro, I don't think `#[repr(C)]` would be necessary here.

MikailBag: There is [`memoffset`](https://lib.rs/crates/memoffset) crate. AFAIU it will work on stable in sound way pretty soon, when `std::ptr::raw_ref` is stable.

Plecra:
Has there been any consideration of `core::task::Context` in this API? It has a similar need for user-defined types, and (if they were added) would ideally use the same mechanism as `Error`.

It would seem the major differences are that `Context` would typically use a `&mut` reference, and also have a greater need for types that borrow from the `Context`.

It'd be possible to express with a trait, at a cost to ergonomics:

```rust
/// Implemented on a type to enable it to be used with `context`
trait Request<'a>: 'static {
    /// The type returned by `context` calls
    type Out;
}
```

Would it be possible to add a `TypeId` that used the same id for types with different lifetimes? The implementation wouldn't need to be public API

yaahc:
> Has there been any consideration of `core::task::Context` in this API? It has a similar need for user-defined types, and (if they were added) would ideally use the same mechanism as `Error`.

Not particularly, and I'm sorry but I'm not really following you on how `Context` is related to this API or how it has similar needs for user-defined types. Could you explain what you mean? Is the `Context` in this case filling the same role as the `&dyn std::error::Error` type in the generic member access API?

Plecra:
Right. `Future`s are all tied to a specific reactor that manages their resources (Like sleeping and spawning other futures), and there have been a few proposals for giving them a generic way to communicate with the executor they're running in. The `Request` type added in this PR functions similarly to some of those proposals.

```rust
pub mod hook {
    pub struct Sleep {
        pub dur: Duration,
        marker: PhantomPinned,
    }
    impl Future for Sleep {
        type Output = ();
        fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
            cx.context::<dyn FnMut(Pin<&mut Sleep>) -> Poll<()>>().expect("no sleeping implementation")(self)
        }
    }
}
pub async fn sleep(dur: Duration) {
    hook::Sleep { dur, marker: PhantomPinned }.await
}
```

It would now be the caller's choice how to manage sleeping futures (allowing them to choose between designs like async-io, tokio, glommio, and easily mock the entire io stack).

There's no room in this PR for a sudden addition to `Context`, but I bring it up because the current `'static` bound would heavily restrict these executor APIs. A typical backend for `sleep` would be expected to queue the running `Future` up to run again after `Duration`, and would need mutable access to some form of task list. Without a lifetime, it will have to use something like `Arc<Mutex<Tasks>>` instead of simply borrowing from its scope. This would be especially difficult in embedded async.

All this is mostly out of the RFC's scope, and I only mentioned it out of curiosity. Sorry for the confusion 😜

yaahc:
> There's no room in this PR for a sudden addition to `Context`, but I bring it up because the current `'static` bound would heavily restrict these executor APIs. A typical backend for `sleep` would be expected to queue the running `Future` up to run again after `Duration`, and would need mutable access to some form of task list. Without a lifetime, it will have to use something like `Arc<Mutex<Tasks>>` instead of simply borrowing from its scope. This would be especially difficult in embedded async.

I see. I'm still a little confused on the reasoning for a `'static` bound on being too restrictive. For request that only applies to the type being accessed, but it should still be possible to get a `&mut T` so long as `T: 'static`. I'm under the impression that this is unavoidable because you cannot downcast to types with lifetimes in them. I guess, could you go into more detail on a specific example that won't work with the `Request` proposed and show how your alternative fixes it?

Plecra:
As far as I'm aware, the `'static` restriction only applies to `TypeId` and `Any` as a consequence of them only using runtime checks. `Context` has its own lifetime that can be used to ensure types live long enough at compile time.

To demonstrate *why* arbitrary lifetimes would be useful, here's an example. I've excluded a lot of the code that would be involved in a real-world implementation, but I hope this gets the idea across. [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1b2b23576d10243da644fd450a1a1cc5)

This shows how an executor might build up a record of the events it needs to wait for (in this case, `minimum_sleep`). To do this, it needs to be able to modify its local state - that is, it needs a `&'a mut _`. It's certainly possible to use `Arc`s and locks to achieve this at runtime, but that would come with a performance penalty and platform restrictions.

As for "fixing" this... the trait I mentioned in https://github.com/rust-lang/rfcs/pull/2895#issuecomment-735713486 would allow us to the the `TypeId` of `Self` to uniquely identify the type constructor for `Out` and build types with lifetimes in them. I think it *might* be possible extend this and avoid needing to add the type annotations like so:

```rust
pub trait Identified {
    type Id: 'static;
}
impl<'a, T: ?Sized + 'static> Identified for &'a T {
    type Id = Ref<T>;
}
pub struct Ref<T: ?Sized>(core::marker::PhantomData<*const T>);
```

I'll try to flesh this out with updated signatures for `provide_context` and `context`

yaahc:
> I'll try to flesh this out with updated signatures for `provide_context` and `context`

hype. Incase it helps, https://github.com/mystor/object-provider/pull/3 is the working copy of the changes in `object-provider` used for the new API.

yaahc:
@Plecra I talked to @mystor about your suggestion and it turns out she had suggested essentially the same design back in august and I'd shot it down because it seemed overly complicated at the time, I was definitely mistaken.

Hows this version look to you? https://discord.com/channels/414933673545039883/727251312986816523/784602437519933460

yaahc:
@Plecra I talked to @mystor about your suggestion and it turns out she had suggested essentially the same design back in august and I'd shot it down because it seemed overly complicated at the time, I was definitely mistaken.

Hows this version look to you? https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a4c9aa1c309fede3ffaf0357ce6dc233

Plecra:
That's pretty much it, yea! When trying to create the example, I hit the *small* snag that `Component` is an unsafe trait. The implementor basically has to ensure `transmute<OtherTypesWithThisIdType<'a>, Self<'a>>()` is valid. It might be possible to use a derive macro for most cases, but there's also the problem of how this system interacts with variance which honestly goes over my head. 

Written out like that, though, I'm surprised by how well it works 😁 I also like that it removes the distinction between value and reference types in the API.

yaahc:
> That's pretty much it, yea! When trying to create the example, I hit the _small_ snag that `Component` is an unsafe trait. The implementor basically has to ensure `transmute<OtherTypesWithThisIdType<'a>, Self<'a>>()` is valid. It might be possible to use a derive macro for most cases, but there's also the problem of how this system interacts with variance which honestly goes over my head.
> 
> Written out like that, though, I'm surprised by how well it works grin I also like that it removes the distinction between value and reference types in the API.

yea, nika originally had this version using an unsafe trait and a MBE macro to define the trait and add an assertion for invariance: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c42239eda4c3270c13397ad70ed4f71e. She went over it for a bit and was fairly confident that the unsafe trait and variance restrictions weren't necessary, and yea, the logic behind it goes way over my head too, for now... >:D

mystor:
I've put together a (super rough) simplified version of the core idea behind the proposal in https://github.com/rust-lang/rfcs/pull/2895#issuecomment-739123333 here: https://github.com/mystor/dyno.

The `Request` API which roughly maps to the existing `object_provider` crate is in the `provider` submodule, and is built on top of the core `Tag` and `Tagged` traits.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,755 @@
...
+    ///     }
+    /// }
+    /// ```
+    pub fn provide_ref_with<T: ?Sized + 'static, F>(&mut self, cb: F) -> &mut Self
+    where
+        F: FnOnce() -> &'a T,
+    {
+        if let Some(RefResponse(response @ None)) = self.downcast_ref_response::<T>() {
+            *response = Some(cb());
+        }
```
bstrie, line 431: Surely there's a typo here somewhere? `response @ None` can't be doing anything interesting, and it doesn't mean anything to try to deref it in the next line. Maybe the second line is meant to be `*self.response =`, but then I have no clue what the purpose of the `@` pattern is.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,755 @@
...
+impl<'a> Request<'a> {
+    /// Perform a checked downcast of `response` to `Option<&'a T>`
+    fn downcast_ref_response<'b, T: ?Sized + 'static>(
+        &'b mut self,
+    ) -> Option<&'b mut RefResponse<'a, T>> {
+        if self.is_ref::<T>() {
+            // safety: If `self.is_ref::<T>()` returns true, `response` must be
+            // of the correct type. This is enforced by the private `type_id`
+            // field.
+            Some(unsafe { &mut *(&mut self.response as *mut _ as *mut RefResponse<'a, T>) })
```
bstrie, line 341: Maybe my coercion-fu is lacking, but what is the `as *mut _` part in the middle achieving?

bstrie (commented):
It's a bit hard to follow the machinery here. The frequent indirection is probably the biggest culprit, but the rather vague `Response` and `Request` terminology is a bit too general to be of any help. Names that contained "Any" or "Dyn" or "Type" might be more self-explanatory.

text/0000-dyn-error-generic-member-access.md
Plecra: Mutable references (in this case `&mut R`) can only be cast to pointers to the same type. Here, that pointer is then cast from `*mut R` to `*mut RefResponse<'a, T>`.

text/0000-dyn-error-generic-member-access.md
Plecra: Does this help explain it?

```rust
struct Data(Option<()>);

let mut data = Data(None);
if let Some(Data(empty @ None)) = Some(&mut data) {
    *empty = Some(());
}
assert_eq!(data.0, Some(()));
```

bstrie: Oof, the default binding modes threw me for a loop here. I thought they only worked at the top level of each pattern, it didn't occur to me that `response @ None` could be desugaring to `response @ &mut None`.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,755 @@
...
+}
+
+fn read_instrs(path: &Path) -> Result<String, ExampleError> {
+    std::fs::read_to_string(path).map_err(|source| {
+        ExampleError {
+            source,
+            path: path.to_owned(),
+            location: Location {
+                file: file!(),
+                line: line!(),
```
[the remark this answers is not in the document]
yaahc: correct. It is already possible to do this with `track_caller`, so we could update this to include `track_caller`. I don't recall if there was a specific reason I avoided using track_caller here but I don't mind updating this example.

text/0000-dyn-error-generic-member-access.md
yaahc: All of this is out of date anyways, I need to update it to use https://docs.rs/dyno/ as the example

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,755 @@
...
+    * ~~We bring this up specifically for the case where you want to use this
+      function to get an `Option<&[&dyn Error]>` out of an error, in this case,
+      it is unlikely that the error behind the trait object is actually storing
+      the errors as `dyn Error`s, and theres no easy way to allocate storage to
+      store the trait objects.~~
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+This opens the door to supporting `Error Return Traces`, similar to zigs, where
```
yaahc, line 740:
```suggestion
This opens the door to supporting [`Error Return Traces`](https://ziglang.org/documentation/master/#toc-Error-Return-Traces), similar to zigs, where
```

Plecra:
Hiya! Seems like this RFC has stalled a little - what's blocking progress at the moment?

yaahc:
Just personal life events, had a surgery + vacation + just started a new job last week. I intend to update this RFC this week time permitting and I have the following changes on my todo list for that update before I'm ready to start pushing the libs team for review and approval.

- [ ] update downcast code to be based on `dyno` rather than `object-provider`
- [ ] add an example use case based on traits used in rustc's compiler error reporting and formatting framework

For the second one there's a specific struct/trait I recall @anp linking to me a little while back that I felt was basically serving the same role as generic member access, though I can't recall exactly what it was atm, so the first step will be digging up the forgotten context.

bstrie:
Do the parts regarding error reporting have any interaction with the proposed `{:!}` error formatting specifier? More generally, how does this RFC fit into the error handling blog post (https://github.com/rust-lang/blog.rust-lang.org/pull/799)? I assume this is one step on the "move Error into core" process?

yaahc:
> Do the parts regarding error reporting have any interaction with the proposed `{:!}` error formatting specifier? More generally, how does this RFC fit into the error handling blog post ([rust-lang/blog.rust-lang.org#799](https://github.com/rust-lang/blog.rust-lang.org/pull/799))? I assume this is one step on the "move Error into core" process?

Indirectly they could, depending on how we implement it. One idea for the reporting proposal is to have a fmting trait that is associated with `{:!}` and has a default impl for all types that implement `Error`, but then can be custom implemented for types that dont (e.g. `anyhow::Error` / `eyre::Report`). The default reporting would not exercise generic member access, and would only access the context needed for the common case, error messages, source, and maaaybe backtrace. The custom reporting types on the other hand would be able to hook in all of their application specific logic that looks up any odd context that is relevant to their usecases.

Amanieu:
This RFC feels very complex and I am a bit uncomfortable with adding so much API surface to std just for the `Error` trait.

I wonder if it is possible to offload some of the complexity of this API to crates.io, similar to how `proc_macro` only exposes a basic interface (just a token stream) while most of the functionality for building proc macros is located in the `syn` and `quote` crates which are not part of the standard library and can evolve independently of it.

Plecra:
> I wonder if it is possible to offload some of the complexity of this API to crates.io, similar to how `proc_macro` only exposes a basic interface (just a token stream) while most of the functionality for building proc macros is located in the `syn` and `quote` crates which are not part of the standard library and can evolve independently of it.

The new API is a fairly fundamental extension to the `Any` trait. The bulk of the API is to do with enabling dynamic downcasts including lifetimes.

There are some parts of the API that could be trimmed at the cost of ergonomics, but I think the new API should be thought of similarly to the `Pin` API. It's being added to enable an `std` API we want, and will likely be useful in implementing other crates in the ecosystem (more safe pinning/downcasts). We'd include the simpler `provide_...` methods for the same reason we have the `Deref` implementations on `Pin` and the `debug_struct` APIs (and on and on): It makes it easier for users to do the right thing

bstrie:
WRT "update RFC to be based on dyno design", can someone give a brief summary of the advantages of this new design over the old one?

yaahc:
> 
> 
> WRT "update RFC to be based on dyno design", can someone give a brief summary of the advantages of this new design over the old one?

I think [this reply](https://github.com/rust-lang/rfcs/pull/2895#issuecomment-739005994) summarizes it best but the main difference is `dyno` uses a proxy type that itself only has a phantom data, and so it is `'static` and gets a `TypeID` and a trait with an associated type that is the actual type being erased, where as the old design required the type being erased to itself be 'static. The new design allows for closures containing mutable references to be requested as shown in [this earlier comment](https://github.com/rust-lang/rfcs/pull/2895#issuecomment-736683549), which could be used in an executor to request a callback that then modifies state in the executor on the other side of the trait object.

yaahc:
> 
> 
> This RFC feels very complex and I am a bit uncomfortable with adding so much API surface to std just for the `Error` trait.
> 
> I wonder if it is possible to offload some of the complexity of this API to crates.io, similar to how `proc_macro` only exposes a basic interface (just a token stream) while most of the functionality for building proc macros is located in the `syn` and `quote` crates which are not part of the standard library and can evolve independently of it.

@Amanieu fwiw I'm already working on extracting the type tagged downcasting part of this RFC into it's own independent RFC, which should hopefully make both RFCs much easier to review.

CleanCut:
This looks fantastic! ❤️ 

What is the likely timeline for this to reach stable?  Given that it hasn't hit nightly yet, I'm guessing this will be gated behind a Rust 2024 edition?

yaahc:
> This looks fantastic! ❤️
> 
> What is the likely timeline for this to reach stable? Given that it hasn't hit nightly yet, I'm guessing this will be gated behind a Rust 2024 edition?

There's no timeline but it won't be gated being any edition. Everything in this RFC should be backwards compatible.

yaahc:
Reminder for myself: Once this lands I would like to update the example in the docs added in https://github.com/rust-lang/rust/pull/95356 to instead use generic member access to grab the `ExitCode` from the error.

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,707 @@
...
+This allows error reporting types to handle errors in a consistent manner
+when constructing reports for end users while still retaining control over
+the format of the full report.
+
+The `Error` trait accomplishes this by providing a set of methods for accessing
+members of `dyn Error` trait objects. It requires that types implement the
+`Display` trait, which acts as the interface to the main member, the error
+message itself.  It provides the `source` function for accessing `dyn Error`
+members, which typically represent the current error's cause. Via
+`#![feature(backtrace)]` it provides the `backtrace` function, for accessing a
```
waynr, line 158: This seems outdated. In master branch I don't see any `backtrace` function on the `Error` trait with or without that feature. Maybe the `backtrace` feature was already merged but left out the `backtrace` method on `Error`. This RFC makes it unnecessary anyway so I'll look into updating this section here.

Amanieu: There used to be an unstable `backtrace` function on `Error`, but it was removed in favor for the generic member access proposed in this RFC.

lygstate:
maybe this proposal need updated to reflect that fact?

text/0000-dyn-error-generic-member-access.md
```diff
@@ -0,0 +1,707 @@
...
+        Request::request_ref(|req| self.provide_context(req))
+    }
+
+    pub fn context<T: 'static>(&self) -> Option<T> {
+        Request::request_value(|req| self.provide_context(req))
+    }
+}
+```
+
+# Drawbacks
```
CobaltCause, line 599: A potential additional drawback is that requesting types dynamically like this can lead to surprising changes in runtime behavior when changing versions of (transitive) dependencies.

Given:

* A crate `foo` with versions 1.0.0 and 2.0.0, each of which provides a type `Foo`.
* A crate `bar` which depends on version 1.0.0 of `foo` and whose errors provide `Foo`.
* A crate `baz` which depends `bar` and on version 2.0.0 of `foo` and tries to request `Foo` from `bar`'s errors.

Then a situation is created in which `baz` will fail to request `Foo` from `bar`'s error because the requested `Foo` and provided `Foo` are actually distinct types because the transitive dependency `foo`'s versions don't match up.

This could even happen in a semver-compatible release of `bar` and `baz` because technically the API is not broken: the type can still be requested, but it will simply begin returning `None` instead of `Some(_)`. This would be even worse if `foo`, `bar,` and `baz` are all library crates depended on by an application crate and the semver-incompatible versions of `foo` are introduced by a semver-compatible change in `bar` or `baz` which would get pulled in by running `cargo update` on the application crate. The application crate has no recourse in this situation aside from waiting for `bar` and `baz` to agree on a semver-compatible version of `foo` again.

(I've been playing with an alternate design based on an associated type rather than generic member access which causes this situation to be a compile time error instead of a runtime behavior change, but that design requires traits with lifetime GATs to be dyn safe to be actually good, and associated type defaults for it to *maybe* be possible to retrofit that design onto the standard library's existing error trait.)

yaahc: I agree this is a footgun I'd prefer to avoid but as far as I can tell this is not a problem being uniquely introduced by generic member access. This is a problem with type erasure in general and the same situation could be encountered by having foo return `Foo` as a `Box<dyn Any>` in any of it's public API's. If `baz` tries to downcast using it's own path to `foo::Foo` it will run into the same type ID mismatch.

I avoid this in my crates currently by re-exporting types from crates that are part of my public API (https://docs.rs/color-eyre/latest/color_eyre/#reexports). Applying this to your example you could avoid this footgun if you downcast/request `Foo` via `bar::foo::Foo` which would correctly resolve to whatever version it is using. This of course gets a little silly when you want to be able to handle all the versions of `Foo`, because then presumably you have to try to request `Foo` once for each path through which you could possibly reach it, which is definitely a subpar user experience, so not necessarily a great solution.

That said, I was never able to figure out how to implement generic member access w/o type erasure and runtime checks. I'm curious to understand more about your potential alternative design, because I would definitely love to see a better solution but I don't see how it is possible given the constraints of the error trait and the goals of generic member access.

CobaltCause: > I agree this is a footgun I'd prefer to avoid but as far as I can tell this is not a problem being uniquely introduced by generic member access.

True.

> This is a problem with type erasure in general and the same situation could be encountered by having foo return `Foo` as a `Box<dyn Any>` in any of it's public API's.

However, it seems to me that this kind of thing is relatively rare currently, both in the standard library and ecosystem crates. The only standard library exception I can think of is `std::io::Error`, for which I found (part of?) the rationale for in https://github.com/rust-lang/rust/pull/24793#issuecomment-97153138, and I haven't personally encountered any ecosystem crates that make use of this and tell users in documentation that they can/should downcast the inner error to some concrete type.

So, I think making type erasure a central part of the error handling API of the standard library would probably be the main place this kind of problem would come up. I'm not sure the existing occasional cases of downcasting in public APIs can be used to justify a standard library API built entirely on downcasting.

> Applying this to your example you could avoid this footgun if you downcast/request `Foo` via `bar::foo::Foo` which would correctly resolve to whatever version it is using.

While true, I imagine another place this might become a problem is in a library whose purpose is to use generic member access to generate pretty error reports. This hypothetical library might hypothetically pull in the `backtrace` crate for handling backtraces. If any other crate in the dependency tree has errors that provide a backtrace from a different version of the `backtrace` crate, the hypothetical library will not be able to extract them. I don't think this can be solved with re-exports, because that would force the library to have a re-export of every crate that has errors that provide a `backtrace` crate backtrace.

Your next point about requesting the type through each path (including crate version) does provide a solution there, but also as you point out, would not be the most fun to maintain.

> I'm curious to understand more about your potential alternative design, because I would definitely love to see a better solution but I don't see how it is possible given the constraints of the error trait and the goals of generic member access.

I can't promise that this particular idea is any better, but for what it's worth, it goes like this:

```rust
trait Error {
    /// Type that contains details about this error.
    ///
    /// For example, this could be a struct containing a String with help text and a Backtrace.
    type Details<'a> = ()
        where Self: 'a;

    /// Get the details about this error.
    fn details(&self) -> Self::Details<'_>;
}
```

As written, this requires some language features that aren't implemented yet:

1. Associated type defaults. Necessary to make it possible to backport this onto `core::error::Error`, but luckily not required for defining a new error trait in a new crate.
2. `dyn`-safety for traits with lifetime GATs. In my current implementation, I've simply dropped the lifetime and am just returning an owned value, which has some overhead but at least allows experimenting with the rest of the design.

Another part of this design is a `map`-like function to change the type of `Details` when composing errors that have different concrete types.

To approximate generic member access to make it possible to implement a hypothetical library for pretty-printing errors like described above, I imagine a crate could define some traits like `trait GetBacktrace { fn get_backtrace(&self) -> Option<&Backtrace>; }` (or one trait with many methods, I suppose) and then adding these traits as bounds on `Error::Details` on the pretty-printing functions. The author of a binary crate could then implement these traits for their `Error::Details` type and use the library's pretty-printing code.

tmccombs: Wouldn't that still have a similar problem? To adapt your example from before:


- A crate foo with versions 1.0.0 and 2.0.0, each of which provides a trait Foo.
- A crate bar which depends on version 1.0.0 of foo and whose errors have a `Details` type that implements Foo 
- A crate baz which depends bar and on version 2.0.0 of foo and has uses a bound that requires `Error::Details` to implement Foo (from version 2.0.0 of foo)

baz still can't get the the `Foo` details from errors from bar.

CobaltCause: That's correct, but the difference is *how* it can't get the details. With the generic member access design, this would fail at run time: the generic member access functions would return `None`. With the associated type design, this would fail at compile time: a compilation error would be emitted about the trait bound being unsatisfied.

tmccombs: You can currently have a `Box<dyn Error + 'static>` (or reference, or some other lifetime). Associated types and dyn-safety for traits with lifetime GATs isn't enough. If you had a `Details` type that isn't equal to the default, your error won't be compatible with anything that types a dynamic error, without some other kind of mechanism. And there may be some benefit to a compile time check. But that also means you lose use cases where you might want to extract data dynamically without knowing if it exists at runtime.

CobaltCause: > You can currently have a `Box<dyn Error + 'static>` (or reference, or some other lifetime). Associated types and dyn-safety for traits with lifetime GATs isn't enough. If you had a `Details` type that isn't equal to the default, your error won't be compatible with anything that types a dynamic error, without some other kind of mechanism.

I guess a really important question is how many notable occurrences of `Box<dyn Error>` (and similar) are there in public APIs? The only one I can think of off the top of my head is `std::io::Error::{other, into_inner}`. In such a case, I believe the `map` function I alluded to before should be sufficient. To extract the original `Details`, you'd have to downcast to the concrete type first, which I think is not really a regression because that's how `std::io::Error` is currently intended to work. For internal-only APIs, the type could simply be changed to specify the desired `Details`.

> But that also means you lose use cases where you might want to extract data dynamically without knowing if it exists at runtime.

I suspect these cases could be covered for `Error`-processing functions by adding bounds on `Details` of traits that contain accessor functions that return `Option<T>` so that `Error` implementors have the, uh, option of returning `None`. Are there any situations where that would be insufficient?

tmccombs: > I guess a really important question is how many notable occurrences of Box<dyn Error> (and similar) are there in public APIs?

Well, I think returning a `Result<(), Box<dyn Error + 'static>` from main is a somewhat common pattern. And while you could maybe replace that with `Result<(), Box<dyn Error<Details=SomeType> + 'static>` , you would still need the same Details type for all errors, which gets messy if your errors come from different sources. 

And you have libraries like [anyhow](https://docs.rs/anyhow/latest/anyhow/) that make heavy use of `dyn core::error:Error`, such as in `root_cause()` and `chain()` and `from_boxed()` in anyhow. 

> To extract the original Details, you'd have to downcast to the concrete type first, which I think is not really a regression because that's how std::io::Error is currently intended to work

But that means you have to know what the concrete type is. And `Details` would need to implement a trait that has support for downcasting. And how do you handle combining multiple detail types into the a single error. Do you try casting to a struct that has the exact combination of details you expect? Do you use a collection of Details objects, and then try downcasting each one of those to what you expect?

CobaltCause: > Well, I think returning a Result<(), Box<dyn Error + 'static> from main is a somewhat common pattern.

The return type of `main` is not a public API, and the requirements of the `Termination` trait allow you to choose other types for the `Err` variant, so this would be easy to change to specify a details type.

> And you have libraries like [anyhow](https://docs.rs/anyhow/latest/anyhow/) that make heavy use of `dyn core::error:Error`, such as in `root_cause()` and `chain()` and `from_boxed()` in anyhow.

I don't think this really counts towards what I was asking either. For `anyhow` (and similar crates) to be relevant here we'd need to know how often other library crates have `anyhow::Error` in their public interface.

> But that means you have to know what the concrete type is.

Worth noting that this is not a regression from the status quo.

> And how do you handle combining multiple detail types into the a single error. Do you try casting to a struct that has the exact combination of details you expect? Do you use a collection of Details objects, and then try downcasting each one of those to what you expect?

You would use the `map` function to compose detail types. You could implement your details type to do any of those things if you wanted, but I think a simpler approach would be to `impl From<OtherDetails> for MyDetails` or so.

tmccombs:
> so this would be easy to change to specify a details type

But adding a Details type to existing errors would break existing code, which means this would either not be very useful, or would have massive churn.

And you would need all your errors to have the same details type, which likely means you would have to map over the details for all of them, which isn't very ergonomic. 


With the original proposal, the only places that have to worry about the context/details of an error are places that consume those details, and places that produce them. But with the associated type, everything else also has to worry about it to make sure the types are consistent, and the details are propagated up.


> I don't think this really counts towards what I was asking either. For anyhow (and similar crates) to be relevant here we'd need to know how often other library crates have anyhow::Error in their public interface

My point is that in order for anyhow to make use of this feature, it would have to make a breaking change, whoch would cause a lot of churn. And what's more, any errors that define a Details type wouldn't be compatible with anyhow, without first mapping to get rid of the details. And I suspect other crates would have similar issues, anyhow is just a prominent example.

> Worth noting that this is not a regression from the status quo.

Even if this was true, it makes it less useful than the original RFC.

But it *is* a regression, in a sense. Because you can't have something like ```Box<dyn Error<Details=dyn Any>>```. Any use of Error has to know, at compile time, the *concrete* type of the Details type. 

I'm not even really sure how downcasting would work here? Do you use a Details type of `Box<dyn Any>` and downcast that? But what if the details isn't the actual type you want, but some composite type that contains the type you want? Does the Details type have to implement a trait that has some way to "downcast" to another type, that may just be a component of the actual details object? Well, then you end up with something exactly like this RFC, or need something like a typemap.

> You would use the map function to compose detail types.

But map to what?

Let's say you are implementing a library,  where you have a fallible function that combines multiple different types of errors, say an  error from another library that includes some details and a user-defined error you know nothing about.  You don't know anything about the Details type on the user defined error (that is generated by user code), but the caller of your function may also want to make use of the details, but you don't know what traits it wants those details to implement. So you can't possibly implement the MySpecialContext details trait that your user wants, even though you might have that data, if the error came from user code.

It is also going to add a lot of complexity to intermediate code that is just handling combining different types of errors. Now you not only need an enum for the different types of errors, you also probably need an enum for different types of Details. And probably need enums to delegate various traits. And it needs to do that regardless of if it actually produces or consumes any detail information.

fogti:
One should also keep in mind that the purpose of the `Error` trait usually is abstracting over all sorts of errors, and rarely caring about the specific one (applications of downcasting should be relatively rare, they're basically an abstraction leak otherwise; the `details` we're talking about here are for supplying additional context which is otherwise hard to transport (although one could introduce additional traits and such to support them, they're at least "mostly boilerplate" to implement). Introducing a `Details` associated type pretty much defeats the purpose (e.g. of using `Box<dyn Error>`); albeit it might make sense in a separate trait, I don't think it would properly belong in the `Error` trait.

olivia-fl:
The previous two comments were probably intended for [this review comment thread](https://github.com/rust-lang/rfcs/pull/2895#discussion_r1894674526). Not sure if it's possible to move them now.

text/0000-dyn-error-generic-member-access.md
CobaltCause: FWIW I've just published [`derail` v0.2.0](https://docs.rs/derail/0.2.0/derail/) which uses an associated type like I describe in this thread, and can also represent errors with zero or many causes rather than just zero or one. There should be enough documentation to play with it and try to get some ideas for how to use it and how well it might or might not work in practice, but there isn't really any guide-level documentation currently (though I do plan to add some eventually).

PR #2897, opened 2020-04-04
result-missing-methods-for-err RFC

PonasKovas:
[Rendered](https://github.com/PonasKovas/rfcs/blob/ccf0b6486fcac5cea786915e06040820bd23b736/text/0000-result-missing-methods-for-err.md)

### DIFF
--- a/text/0000-result-missing-methods-for-err.md
+++ b/text/0000-result-missing-methods-for-err.md
@@ -1,0 +1,43 @@
+- Feature Name: `result_missing_methods_for_err`
+- Start Date: 2020-04-04
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Add these methods:
+ - `into_err`: does the same as `into_ok` but for `Err`,
+ - `map_err_or`: does the same as `map_or` but for `Err`,
+ - `map_err_or_else`: does the same as `map_or_else` but for `Err`,
+ - `unwrap_err_or`: does the same as `unwrap_or` but for `Err`,
+ - `unwrap_err_or_default`: does the same as `unwrap_or_default` but for `Err`,
+ - `unwrap_err_or_else`: does the same as `unwrap_or_else` but for `Err`.
+
+# Motivation
+[motivation]: #motivation
+
+These methods should be present because of logical reasons.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+No drawbacks.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+# Prior art
+[prior-art]: #prior-art
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+# Future possibilities
+[future-possibilities]: #future-possibilities


PonasKovas:
[Implementation](https://github.com/PonasKovas/rust)

mehcode:
I'm not on any teams so take my word with some rice but generally I see simple API additions like this just be a PR to Rust without an RFC.

programmerjake:
I would consider `map_err_or_else` to be superfluous since it does the exact same thing as `map_or_else` but with a different argument order. Other than that, looks reasonable at first glance!

kornelski:
Can you provide code examples where each of this methods is useful?

Just because they fit a pattern doesn't necessarily mean that all of them are needed.

PonasKovas:
Sure. I came to writing this RFC because I needed one of those methods myself but couldn't find it.
 - `into_err`: for example I could have a method that takes over the thread and runs infinitely, unless an error is encountered. It would make sense for that method to have the return type `Result<!, Error>` and you would want to convert it to `Error` right away with `into_err` because it's impossible for it to be `Ok`.
 - `map_err_or`: Honestly I can't think of a situation, but I can imagine people finding themselves needing it.
 - `map_err_or_else`: This one's pretty useless, yeah. When I wrote this RFC I wasn't sure whether I should add it or not, but I thought I would, just for the pattern. If you think that it shouldn't be here, sure, I don't really have a strong opinion.
 - `unwrap_err_or`: This one is pretty similar to `map_err_or`, I can't think of a logical situation where you would have default errors, but, well, some people write weird code.
 - `unwrap_err_or_default`: Same, default errors.
 - `unwrap_err_or_else`: If I have a function that returns `Result<u32, Error>`, and in case of success all I need to do with the `u32` is to, for example, print it to stdout and exit from the program, while error handling is a much more complex thing to do - there are lots of errors that could have happened and I need to handle all of them. In this situation it would make sense to do it like this:
```rust
match result.unwrap_err_or_else(|n| { println!("{}", n); std::process::exit(0); }) {
    // lots of code to handle errors...
}
```

I hope this convinces you

matthiasbeyer:
Adding these functions via extension traits in a dedicated crate for this seems to be a simpler solution, IMO. Such a crate would also be able to break the API if necessary, which the stdlib cannot. This seems to be an advantage in this case, TBH.

There are already some crates out there that implement such extensions (not sure whether these exact use-cases are covered, but I am pretty sure that there are crates for similar cases, like for example [resiter](https://github.com/matthiasbeyer/resiter/) (which is for iterators over `Result<_, _>` ; shameless self-promotion).

burdges:
I suppose `into_err` sounds plausibly useful.  It'll stay suck behind `#![feature(never_type)]` though anyways, so whatever.

If invoked directly, all the `map_err_*` and ``unwrap_err_*` only make Rust code less readable when compared with the equivalent `map_or_else`.  Among these, I think only `unwrap_err_or_default` supports closure usage ala `iter.map(Result::unwrap_err_or_default)`, but again it's clearer to write it out like
```
let unwrap_err_or_appropriate_error = |x| x.map_or_else(|_| E::AppropriateErrorVariant,|y| y)
.. iter.map(unwrap_err_or_appropriate_error) ..
``` 

As you say, we should not encourage default values for error types.  In fact, we already do not permit `?` to convert from `Option<T>::None` to `Result<T,E>::Err(Default::default())`.  If you want such behavior then you must explicitly `impl From<NoneError> for E` instead.

kornelski:
Drawbacks: "No Drawbacks" is too optimistic. The drawback is that there will be more methods, and it will be marginally harder to find useful methods. Users reading `Result` docs will need to think "Do I need this method or another one?". It increases chance of someone being confused that there's `map_err_or`, but not `map_ok_or`.

Another drawback is that if the best use of a function is for "some people who write weird code", then Rust will help writing weird code. It dilutes what idiomatic Rust is.

PonasKovas:
I thought of a situation where you might need `map_err_or`, `unwrap_err_or` or `unwrap_err_or_default`.
Imagine that you have a function that runs some code and returns a `Result<(), Error>`, where `Error` is an enum containing all possible errors. This is the main function of your application so you want to return a `0` exit code in case of `Ok`, and in case of an error, you may want to print it to stderr and the convert it to an i32 and return that as an exit code. In this situation, `map_err_or` would prove useful, and quite readable in my opinion:
```rust
let result: Result<(), Error> = run();
result.map_err_or(0i32, |e| {
  eprintln!("Error: {}", e);
  e as i32
})
```
As for `unwrap_err_or` and `unwrap_err_or_default`, imagine the same situation, but where instead of `Result<(), Error>` we have straight `Result<(), i32>` with the exit code right there in the result.
```rust
let result: Result<(), i32> = run();
result.unwrap_err_or(0i32)
```
```rust
let result: Result<(), i32> = run();
result.unwrap_err_or_default()
```
To me, this looks more readable than `result.err().unwrap_or(0i32)` and `result.err().unwrap_or_default()`

burdges:
I'm confident readers would fine `unwrap_err_of_*` much less readable.  Although rare, at least `.err()` sounds intuitive, especially with `.ok()` being used more frequently. 

I think rust devs learn `unwrap_or*` eventually, but readers commonly write minimal rust code, if any, so `unwrap_or*` already suck.  All important code gets read many many times, but written only once, so readers' experience matters far more.  Aside from lead devs doing code review, readers includes newer inexperienced devs, project managers, technical writers, auditors, and experts in algorithms, statistics, or other domains.  

Arguably, we should replace `r.map_or_else(f,g)` with `r.map(f).unwrap_or_else(g)`, because `map_or_else` makes the argument order confusing.  `map_err_or_else` makes this confusion much worse.  At minimum, we should explain `Result::map_or_else` via the `map` plus `unwrap_or_else`  mnemonic.

PR #2901, opened 2020-04-06
Add RFC for else-match.

Jezza:
[Rendered](https://github.com/Jezza/rfcs/blob/else-match/text/0000-else-match.md)

The goal would be to allow a trailing match expression following an else branch, permitting an else match construct.

```rust
let target = if ctx.close {
    None
} else match ctx.target {
    PerfTarget::None => None,
    PerfTarget::FrameTime(ft) => Some(ft),
    PerfTarget::Fps(fps) => Some(1000 / fps as usize),
};
```

### DIFF
--- a/text/0000-else-match.md
+++ b/text/0000-else-match.md
@@ -1,0 +1,113 @@
+- Feature Name: `else-match`
+- Start Date: 2020-04-06
+- RFC PR: -
+- Rust Issue: -
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


Jezza:
With some discussion in the Discord channel, this was determined as a step too far.

Without changes to match, the only place it would make sense would the final else in the chain, but why would match be any different?

Other than it being an expression, I feel like there wouldn't be too much difference between match in particular, and other expressions.

PR #2924, opened 2020-05-11
RFC: Deprecate FromStr in favor of TryFrom<&str>

malobre:
[Rendered](https://github.com/malobre/rfcs/blob/master/text/0000-deprecate-fromstr.md)

This is a work in progress.

### DIFF
--- a/text/0000-deprecate-fromstr.md
+++ b/text/0000-deprecate-fromstr.md
@@ -56,7 +56,7 @@ impl From<&str> for Dummy {
 struct Dummy<'a>(&'a str);
 
 // This doesn't compile
-impl<'a> FromStr for Dummy<'a> {
+impl<'a> std::str::FromStr for Dummy<'a> {
     type Err = core::convert::Infallible;
     
     fn from_str(s: &str) -> Result<Self, Self::Err> {
--- a/text/0000-deprecate-fromstr.md
+++ b/text/0000-deprecate-fromstr.md
@@ -11,7 +11,7 @@ Deprecate [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) in f
 # Motivation
 [motivation]: #motivation
 
-`FromStr` was created when `TryFrom` did not exist, `FromStr` is now superfluous.
+[`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) was created when [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) did not exist, [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) is now superfluous.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
@@ -56,9 +56,9 @@ assert_eq!(p.unwrap(), Point { x: 1, y: 2 })
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-- Mark `FromStr` as deprecated
-- Mark `str::parse()` as deprecated
-- Make `FromStr` implement `TryFrom<&str>`:
+- Mark [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) as deprecated
+- Mark [`str::parse()`](https://doc.rust-lang.org/std/primitive.str.html#method.parse) as deprecated
+- Make [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) implement `TryFrom<&str>` with [specialization](https://github.com/rust-lang/rfcs/pull/1210):
 ```rust
 impl<T> TryFrom<String> for T
 where
--- a/text/0000-deprecate-fromstr.md
+++ b/text/0000-deprecate-fromstr.md
@@ -11,9 +11,7 @@ Deprecate [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) in f
 # Motivation
 [motivation]: #motivation
 
-`FromStr` was created when `TryFrom` did not exist.
-
-As such, `FromStr` is now superfluous.
+`FromStr` was created when `TryFrom` did not exist, `FromStr` is now superfluous.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
--- a/text/0000-deprecate-fromstr.md
+++ b/text/0000-deprecate-fromstr.md
@@ -27,7 +27,7 @@ struct Point {
     y: i32,
 }
 
-impl From<&str> for Point {
+impl TryFrom<&str> for Point {
     type Error = ParseIntError;
 
     fn try_from(s: &str) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


text/0000-deprecate-fromstr.md
```diff
@@ -0,0 +1,81 @@
...
+```rust
+let p = Point::try_from("(1,2)");
+assert_eq!(p.unwrap(), Point { x: 1, y: 2 })
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+- Replace all `FromStr` implementation in `std` with the corresponding `TryFrom<&str>`
+- Rewrite `str::parse()` to use `TryFrom<&str>`
```
shepmaster, line 60: I don't think this is possible. This is stable code that must continue compiling:

```rust
fn my_parse<T: std::str::FromStr>(v: i32) -> T {
    match v.to_string().parse() {
        Ok(v) => v,
        Err(_) => panic!()
    }
}
```

text/0000-deprecate-fromstr.md
```diff
@@ -0,0 +1,81 @@
...
+- Mark `FromStr` as deprecated
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Backward compatibility will be broken
```
shepmaster, line 66: This... isn't a drawback; as far as I understand it it's a non-starter.

malobre: Ok, I couldn't find anything on the policy around backward compatibility, I'll close the PR.

shepmaster: > I couldn't find anything on the policy around backward compatibility

What places did you look? It could be good to have something more explicitly stated there for the future.

One reason it might not be easy to find is that stability is kind of a "core tenet" of the language and standard library and just permeates everything. One place I found was https://blog.rust-lang.org/2014/10/30/Stability.html

BurntSushi: I think the canonical source is still the API Evolution RFC: https://rust-lang.github.io/rfcs/1105-api-evolution.html

malobre: Thanks for the link.
I searched for "rust deprecation policy" and "rust backward compatibility" with DuckDuckGo which led me nowhere. However a search on Google got me plenty of results but not the link you sent.

This should probably be in the Readme.

malobre:
I updated the RFC to be backward compatible (this is still a wip).

BurntSushi:
I am strongly opposed to deprecating `parse`. It is widely used, and the churn that would be caused by deprecating that is unacceptable IMO.

I'd also rather we didn't deprecate `FromStr` either, since that is also pretty widely used.

I don't really see a lot of value in deprecating this stuff. This RFC is very light on details.

shepmaster:
I am appreciative of the strangeness between `FromStr` and `TryFrom`. I wonder if a future hypothetical world with specialization would allow implementing one in terms of the other as a blanket implementation.

malobre:
There has been some discussion about this [here](https://github.com/rust-lang/rfcs/issues/2143).

About the shortness of the RFC: I will add content. I just need some time to put my thoughts into writings.

shepmaster:
> I just need some time to put my thoughts into writings.

You may wish to re-read https://github.com/rust-lang/rfcs#before-creating-an-rfc

clarfon:
I strongly feel that `TryFrom` should always be used instead of `FromStr`, and deprecating would be very useful. However, I'm very much not in favour of adding a blanket implementation, as it either requires specialisation or seriously limits possible impls.

A few things that influence my feelings:

1. The term `parse` gives a much different connotation than a simple string conversion. I can wrap a string in an error struct and I'd say that makes conversion reasonable, but I wouldn't call that parsing.
2. The blanket impl for `From` makes infallible conversions much more idiomatic.
3. The `FromStr` trait limits lifetimes in a way that prevents borrowing the passed string. For example, you can't convert `&'a str` into `Wrapper<'a>`; you *must* copy the string into a `String` if you do `FromStr`.

Honestly, I feel that the churn is very overstated. Deprecation exists for cases like this where a very clear alternative is there, and this is an example.

That said, I do agree that the RFC itself could do much better to elaborate on all of these points.

clarfon:
To clarify: I think your enthusiasm in writing this RFC is wonderful, although I think that if you don't know a whole lot about the process it's much better to communicate with folks before submitting an RFC, and I know a lot of people would likely be willing to help you out.

The definition of breaking changes is probably one of the biggest things to know when it comes to writing an RFC and I'd say that you definitely want to read up on that first. I believe that there's actually an RFC for it, but as I'm on my phone I don't have the time to look up which one it is.

malobre:
Thanks for your opinion, this RFC was clearly prematurely created. My apologies.

I've [posted a topic](https://internals.rust-lang.org/t/pre-rfc-deprecate-fromstr-in-favor-of-tryfrom-str/12331?u=malobre) for further discussion.

a-m-joseph:
As a somewhat-new Rust user I experienced considerable confusion trying to figure out why both FromStr and TryFrom<&str> existed.  I spent several hours banging my head against the wall trying to figure out why the standard library had a FromStr trait!  I thought to myself: surely the Rust designers, having created such an amazing language, would not have included a superflous type like this for no reason.  There must be some secret super-power posessed by FromStr that I am not aware of.  I must learn this secret super-power!

To my dismay, FromStr has no such secret super-powers.  And I cannot recover the afternoon spent searching for them :(

I realize that deprecating FromStr and parse() will lead to lots of churn for experienced Rustaceans.  But please realize that leaving them in the standard library for the rest of eternity will instead create much confusion for newcomers.

In the meantime, please consider adding text to [(the docs](https://doc.rust-lang.org/std/str/trait.FromStr.html) explaining FromStr's historical provenance, its overlap with From<&str>, and the reason why FromStr still exists.  This was the first place I went looking for an answer to "why does FromStr exist when we have TryFrom<&str>?".

I hope this comment is not perceived as unconstructive.  Just wanted to provide some perspective from a new user.

RustyYato:
Adding docs to `FromStr` about it's overlap with `TryFrom<&str>` sounds like a fantastic idea! I'm surprised there are no such docs already.

malobre:
If anyone wishes to take over this RFC draft, be my guest.
I'm (unfortunately 😢) not using Rust for work and therefore can't currently spare much time contributing to the project.

clarfonthey:
Honestly considering it. Not sure if I'll have a chance to give it a shot before the weekend.

zbraniecki:
If I understand, for that to be possible, we'd need to cover all currently covered `FromStr` implementations in the stdlib with `TryFrom<&str>`. For example, at the moment `usize` implements `FromStr` but not `TryFrom<&str>`.

Getting that coverage could be done in parallel with this RFC, and would bring `TryFrom<&str>` up to par with `FromStr`.

Am I correct?

malobre:
FYI, as noted [here](https://internals.rust-lang.org/t/pre-rfc-deprecate-fromstr-in-favor-of-tryfrom-str/12331/6) some crates implements `FromStr` and `TryFrom<&str>` for different purposes. [Playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=303cfc71940f3ce6c86bdaa2cb3c5ee3).
But yes, this could be a separate RFC as this was more focused on deprecating `FromStr`.

PR #2932, opened 2020-05-23
Update 2394-async_await.md

Senrelio:
'block_on' function might have been moved from futures module to futures::executor module. I wonder if the dependency used there should be state explicitly?

### DIFF
--- a/text/2394-async_await.md
+++ b/text/2394-async_await.md
@@ -89,7 +89,7 @@ async fn print_async() {
 fn main() {
      let future = print_async();
      println!("Hello from main");
-     futures::block_on(future);
+     futures::executor::block_on(future);
 }
 ```
 


steveklabnik:
We don't really have a policy if RFCs get updated after acceptance. I'll let the lang team decide if they want to merge this or not, but in general, RFCs are not a spec, and you can't assume they're what's actually in the language.

r? @rust-lang/lang

joshtriplett:
I don't see any problem with merging this change, but I don't want anyone getting the idea that every past RFC will remain accurate in the future and should be kept up to date.

nikomatsakis:
I think we should merge it, though I agree that we don't have a general rule towards keeping RFCs fully up to date.

PR #2938, opened 2020-05-30
Readme fixes

jzaefferer:
This PR provides a few small fixes for the README:

* move global links to the end of the document
* clarify how the 0000 gets replaced with the PR issue number
* use https for all links
* adding missing ToC entries

For "clarify how the 0000 gets replaced with the PR issue number" I looked at how the rename was made in practice, since I couldn't find any instruction for it. I hope this captures it well enough.

I noticed these while adopting this RFC process for Nushell, under the MIT License: https://github.com/nushell/rfcs/pull/1 I've added an Acknowledgements section to reflect the copy we started with. If there's a problem with this approach, please let us know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,17 +252,17 @@ consensus and community norms, not impose more structure than necessary.
 
 
 [official Discord server]: https://discord.gg/rust-lang
-[developer discussion forum]: http://internals.rust-lang.org/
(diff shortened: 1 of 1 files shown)


jzaefferer:
I've rebased this and resolved the conflict.

jyn514 (approved):

PR #2949, opened 2020-06-24
Hierarchic anonymous life-time

redradist:


Ixrec:
This proposal was previously discussed at https://internals.rust-lang.org/t/simplification-reference-life-time/12224/8?u=ixrec, but unfortunately it looks like none of the problems raised there have been addressed. To avoid relitigating what those problems are, here's a quick summary:

- The motivation cited is iterative development, but it's well-known that when prototyping/iterating/refactoring, references in structs are best avoided in favor of cloning or boxing until the final design becomes clear.
- Lifetimes are semantically viral, cannot be encapsulated, or however one wishes to phrase that. Omitting the viral _syntax_ without any change to the semantics just creates structs that misleadingly appear to have no lifetime parameter even though they actually still have one. Historically this omission was allowed on return types, but that proved so confusing in practice we've already deliberately reintroduced a `'_` syntax for that (as even the RFC acknowledges).
- Some posts in the linked thread imply part of the motivation is a poor error message, but this change would actually make it harder to produce a good error message (for the reasons in the previous bullet point)
- This doesn't even work for structs with multiple lifetime parameters.
- While there are past suggestions/proposals for a `'self` lifetime, they all seem to be aiming at different problems (e.g. enabling self-referential structs), and on closer inspection all of them turned out to be incomplete or confused.
- Having to write `<'a>` two times on each struct in a composition hierarchy just isn't a big enough problem in practice to justify dedicated sugar anyway.

Ixrec:
FWIW, I could get behind a proposal to allow `struct Foo<'_> { b: Bar }` where `Bar` has an omitted lifetime parameter. That does cut the typing in your example from `<'a><'a>` to just `<'_>`, seems like a place we'd expect `'_` to work anyway (or at least I did; I had to go check that it doesn't already work), and doesn't hide the fact that `Foo` has a lifetime parameter. But that's as far as I could see us ever going with this sort of idea.

redradist:
@Ixrec 
> 
> This proposal was previously discussed at https://internals.rust-lang.org/t/simplification-reference-life-time/12224/8?u=ixrec, but unfortunately it looks like none of the problems raised there have been addressed. To avoid relitigating what those problems are, here's a quick summary:
> 
>     * The motivation cited is iterative development, but it's well-known that when prototyping/iterating, references in structs are best avoided in favor of cloning or boxing until the final design becomes clear.
> 
>     * Lifetimes are semantically viral, cannot be encapsulated, or however one wishes to phrase that. Omitting the viral _syntax_ without any change to the semantics just creates structs that misleadingly appear to have no lifetime parameter even though they actually still have one. Historically this omission was allowed on return types, but that proved so confusing in practice we've already deliberately reintroduced a `'_` syntax for that (as even the RFC acknowledges).
> 
>     * Some posts in the linked thread imply part of the motivation is a poor error message, but this change would only make it harder to produce a good error message (for the reasons in the previous bullet point)
> 
>     * This doesn't even work for structs with multiple lifetime parameters.

It will work with multiple life-times as well as without life-time:
```rust
struct CompositeObject<'a> {
    counter: &'a usize,
    obj: &'self SomeType,
}
struct BigObject<'a> {
    composite_obj: CompositeObject<'a>,
    count: i32,
}
struct Application<'a> {
   big_obj: BigObject<'a>,
}
```
This code will be translated to:
```rust
struct CompositeObject<'a, 'self> {
    counter: &'a usize,
    obj: &'self SomeType,
}
struct BigObject<'a, 'self> {
    composite_obj: CompositeObject<'a, 'self>,
    count: i32,
}
struct Application<'a, 'self> {
   big_obj: BigObject<'a, 'self>,
}
```
`'self` is implicitly added and it will allow even in iterative programming to use references ;)

> 
>     * While there are past suggestions/proposals for a `'self` lifetime, they all seem to be aiming at different problems (e.g. enabling self-referential structs), and on closer inspection all of them turned out to be incomplete or confused.

Why ? Inspection will show the line where object implicitly bound with some other object

> 
>     * Having to write `<'a>` two times on each struct in a composition hierarchy just isn't a big enough problem in practice to justify dedicated sugar anyway.

It is not two times ))) It is two times in example of code ... I have been writing library where it was a real issue ... Each time I wanted to test some new design and I had to change half of the code to test it and then better design was to change it little-bit and it was a nightmare

redradist:
@Ixrec 
> 
> 
> FWIW, I could get behind a proposal to allow `struct Foo<'_> { b: Bar }` where `Bar` has an omitted lifetime parameter, since `struct Bar<'_>{}` would still needs a `'_`. That does cut the typing in your example from `<'a><'a>` to just `<'_>`, seems like a place we'd expect `'_` to work anyway (or at least I did; I had to go check that it doesn't already work), and doesn't hide the fact that `Foo` has a lifetime parameter. But that's as far as I could see us ever going with this sort of idea.

It will work simlicly with `'_` anonymous life-time like this:
```rust
struct Foo<'_> {
    b: Bar
}

struct Bar<'_>{
   obj: &'self BigObj,
}
```
and it will translated to:
```rust
struct Foo<'_, 'self> {
    b: Bar<'_, , 'self>
}

struct Bar<'_, 'self>{
   obj: &'self BigObj,
}
```

zackw:
I want to highlight some bits of back-and-forth above that I think point right at the crux of why @redradist and everyone else are talking past each other:

@lxrec wrote:
> Lifetimes are semantically viral, cannot be encapsulated, or however one wishes to phrase that. Omitting the viral syntax without any change to the semantics just creates structs that misleadingly appear to have no lifetime parameter even though they actually still have one.

@redradist wrote:
> Each time I wanted to test some new design and I had to change half of the code to test it and then better design was to change it little-bit and it was a nightmare

It seems to me, @redradist, that you have not understood why Rust has lifetime parameters in the first place, and so you're trying to avoid thinking about the exact thing that lifetime parameters exist _to require you to think about_.

Your hypothetical pair of `Application` structs...

> ```rust
> struct Application<'a> { ... }
> struct Application { ... }
> ```

have to be _used_ in profoundly different ways.  I deleted the contents of the structs because _they don't matter_.   The presence or absence of the lifetime parameter is relevant to the _client_ of the API and that's why we want it to be explicit.  Specifically, this function

```
fn create_application() -> Application {
   Parameter param { ... };
   Application { ..., param, ... }
}
```

is _correct_ (and will compile, assuming Parameter is a Copy type) for `struct Application` with no lifetime parameter, but _incorrect_ for `struct Application<'a>` (and won't compile, even with `&param`) because you're returning an object containing a reference to another object that no longer exists after the function returns.

You're frustrated because you keep messing with the internals of `Application` in ways that change whether it needs a lifetime parameter and having to change every nested structure, and that frustration is Rust trying to show you that you're doing your API design backward.  You shouldn't _allow_ the internals of `Application` to dictate whether it has lifetime parameters.  You should consciously decide whether or not `Application`'s _external API_ ought to include lifetime parameters, based on what makes most sense for how `Application` should be used, and then propagate that decision downward into the internals.

burdges:
I think `struct Foo<'_> { b: Bar }` harms readability too much @Ixrec 

@redradist  You'll reduce refactorings by initially writing
```
struct Foo<B> { b: B, .. }
impl<B: Borrow<Bar>> Foo<B> { .. }
impl<B: BorrowMut<Bar>> Foo<B> { .. }
```

redradist:
> 
> 
> I want to highlight some bits of back-and-forth above that I think point right at the crux of why @redradist and everyone else are talking past each other:
> 
> @Ixrec wrote:
> 
> > Lifetimes are semantically viral, cannot be encapsulated, or however one wishes to phrase that. Omitting the viral syntax without any change to the semantics just creates structs that misleadingly appear to have no lifetime parameter even though they actually still have one.
> 
> @redradist wrote:
> 
> > Each time I wanted to test some new design and I had to change half of the code to test it and then better design was to change it little-bit and it was a nightmare
> 
> It seems to me, @redradist, that you have not understood why Rust has lifetime parameters in the first place, and so you're trying to avoid thinking about the exact thing that lifetime parameters exist _to require you to think about_.

Do not make fast conclusion ;) My background started from C/C++ where the not tracking life-time was an issue

On the other hand `Rust` has very cool and nice idea about life-times and by tracking them it can know when using some variable can 'cause and issue ... But as drawback `Rust` was designed in such way the it enforce developers to explicitly add named life-times that is not bad idea in some cases for example in dependent life-times:
```rust
#![feature(label_break_value)]

struct MyStruct<'a, 'b: 'a> {
    k: &'a i32,
    i: &'b u32,
}

fn main() {
    'a: {
        let i: u32 = 3;
        'b: {
            let k: i32 = 3;
            let struc = MyStruct { k: &k, i: &i };
        }
    }
}
```
But most of the time writing life-time is just boilerplate ... all structs bypass life-times like this:
```rust
struct MyStruct<'a> {
    k: &'a i32,
    i: MyStruct2<'a>,
}

struct MyStruct2<'a> {
    i: &'a u32,
}

fn main() {
    'a: {
        let i: u32 = 3;
        'b: {
            let k: i32 = 3;
            let struc = MyStruct { k: &k, i: MyStruct2 { i: &i } };
        }
    }
}
```
I do not think that named life-times it is bad in fact I like them, but in certain cases they unnecessary and just reduce code readability (

> 
> Your hypothetical pair of `Application` structs...
> 
> > ```rust
> > struct Application<'a> { ... }
> > struct Application { ... }
> > ```
> 
> have to be _used_ in profoundly different ways. I deleted the contents of the structs because _they don't matter_. The presence or absence of the lifetime parameter is relevant to the _client_ of the API and that's why we want it to be explicit. Specifically, this function
> 
> ```rust
> fn create_application() -> Application {
>    Parameter param { ... };
>    Application { ..., param, ... }
> }
> ```
> 
> is _correct_ (and will compile, assuming Parameter is a Copy type) for `struct Application` with no lifetime parameter, but _incorrect_ for `struct Application<'a>` (and won't compile, even with `&param`) because you're returning an object containing a reference to another object that no longer exists after the function returns.
> 
> You're frustrated because you keep messing with the internals of `Application` in ways that change whether it needs a lifetime parameter and having to change every nested structure, and that frustration is Rust trying to show you that you're doing your API design backward. You shouldn't _allow_ the internals of `Application` to dictate whether it has lifetime parameters. You should consciously decide whether or not `Application`'s _external API_ ought to include lifetime parameters, based on what makes most sense for how `Application` should be used, and then propagate that decision downward into the internals.

Okay, I got your point, but what is bad in mixing two solution: explicit life-time (for user) and implicit self (for developer)
```rust
fn make_app(config: &Config) -> Application<'_>; // 'self was added as hidden life-time due to internal struct has reference
```
It has all advantages as from user as from developer point of view ...

redradist:
Reading all the comments on pull-request and in discussion, I have figured out that I was wrong in name

I mostly mean some life-time that is implicitly added to structure declaration

The syntax also could be something like this:
```rust
struct CompositeObject<'a> {
    counter: &'a usize,
    obj: &'_ SomeType,
}
struct BigObject<'a> {
    composite_obj: CompositeObject<'a>,
    count: i32,
}
struct Application<'a> {
   big_obj: BigObject<'a>,
}
```
and translation to:
```rust
struct CompositeObject<'a, '_> {
    counter: &'a usize,
    obj: &'_ SomeType,
}
struct BigObject<'a, '_> {
    composite_obj: CompositeObject<'a, '_>,
    count: i32,
}
struct Application<'a, '_> {
   big_obj: BigObject<'a, '_>,
}
```

Maybe this syntax much better, because it does not confuse anyone ....

`'self` as concept could be implemented differently like this:
```rust
struct CompositeObject<'a> {
    counter: &'a usize,
    obj: &'self SomeType,
}
struct BigObject<'a> {
    composite_obj: CompositeObject<'a>,
    count: i32,
}
struct Application<'a> {
   big_obj: BigObject<'a>,
}
```
`'self` life-time is smaller or equal than life-time of `CompositeObject` object

If `'comp_obj` is life-time of `CompositeObject` than `'comp_obj: 'self`

redradist:
@Ixrec @zackw @burdges I have changed the name and the wording in RFC

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+```
+or
+```rust
+fn make_app(config: &Config) -> Application<'_>;
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Not known at the current time
```
pythoneer, line 126: There are several drawbacks mentioned in the corresponding thread on IRLO and the comments on this PR. It would be helpful to list them here to get a better overview.

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+   big_obj: BigObject<'anon>,               // 'anon is implicitly used here
+}
+```
+
+Take a look at example with multiple anonymose life-times:
+```rust
+struct CompositeObject {
+    obj0: &'_ SomeType,
+    obj1: &'_ SomeType,
+}
```
CryZe, line 86:
```suggestion
struct CompositeObject<'_> {
    obj0: &SomeType,
    obj1: &SomeType,
}
```
Hidden lifetimes are deprecated, you also don't ever want `'_` on references, as that's implied.

redradist: Why they deprecated ?

What if the `obj0` and `obj1` would have different life-times ? How it will work with single anonymous life-time in declaration ?

pickfire: @redradist Why not? Both of them could depend on the same lifetime and life as long as each other.

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
+- Feature Name: Hierarchic anonymous life-time
+- Start Date: 2020-06-24
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+
+New use of anonymous life-time '_ that implicitly added to current structure
```
pickfire, line 8:
```suggestion
New use of anonymous life-time '_ that implicitly added to current structure.
```

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+struct BigObject<'a> {
+    composite_obj: CompositeObject<'a>,
+    count: i32,
+}
+
+struct Application<'a> {
+   big_obj: BigObject<'a>,
+}
+```
+Everywhere in composition hierarchy I need to write 'a ... most of the times it is just boilerplate code ...
```
pickfire, line 45:
```suggestion
The developer need to write `'a` throughout the hierarchy.
```

text/0000-hierarchic-anonymous-life-time.md
redradist: > 
> 
> @redradist Why not? Both of them could depend on the same lifetime and life as long as each other.

@pickfire But what would be semantic of `struct CompositeObject<'_>` ? Does `'_` mean one life-time or it mean several different life-times ?

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+}
+
+struct BigObject<'anon> {                   // 'anon is implicitly added life-time
+    composite_obj: CompositeObject<'anon>,  // 'anon is implicitly used here
+    count: i32,
+}
+
+struct Application<'anon> {                 // 'anon is implicitly added life-time
+   big_obj: BigObject<'anon>,               // 'anon is implicitly used here
+}
```
pickfire, line 78: If the following code is generated, how will a reader knows that
```rust
struct BigObject {
    composite_obj: CompositeObject,
    count: i32,
}
```

have an implicit `CompositeObject<'anon>` by reading the documentation generated for `BigObject`?

text/0000-hierarchic-anonymous-life-time.md
pickfire: That's true even though it is uncommon.

text/0000-hierarchic-anonymous-life-time.md
redradist: > 
> 
> If the following code is generated, how will a reader knows that
> 
> ```rust
> struct BigObject {
>     composite_obj: CompositeObject,
>     count: i32,
> }
> ```
> 
> have an implicit `CompositeObject<'anon>` by reading the documentation generated for `BigObject`?

@pickfire Actually thinking ... what explicit life-time gives you ? You just need to know how to initialize `CompositeObject` ... all other job should be done by `CompositeObject`, it is level of abstraction

text/0000-hierarchic-anonymous-life-time.md
redradist: @pythoneer I will add ;)

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+
+struct BigObject {
+    composite_obj: CompositeObject,
+    count: i32,
+}
+
+struct Application {
+   big_obj: BigObject,
+}
+```
```
dlight, line 61: I think that you need some notation on each point of use to specify that a lifetime was elided. That is, `CompositeObject` isn't a type with no lifetime parameters, it's a type whose exact lifetime parameters were deemed unimportant. So in `struct BigObject`, when defining a field with this type, we could write `CompositeObject<'_>` to mark this.

But even this wouldn't be sufficient: this feature wants to abstract away the number of lifetime parameters too (we could have two fields with elided lifetimes). So maybe `CompositeObject<'_ ...>` works? Where `'_ ...` means any number of lifetime parameters (that gets implicitly added to `BigObject`'s own parameters). Likewise for `Application`.

Now, you perhaps need some notation in the struct definitions too to specify that there are N elided lifetimes. For symmetry, `struct BigObject<'_ ...>` works.

With those additions, the feature seems less appealing. But without them, it's confusing.

text/0000-hierarchic-anonymous-life-time.md
felix91gr: To me it has been quite helpful to see explicit lifetimes because then I know that there is a lifetime being tracked inside the structure. In that way, it is a very helpful kind of documentation. One that is also always up-to-date, which is not a common feature of documentation in general.

redradist: > 
> 
> To me it has been quite helpful to see explicit lifetimes because then I know that there is a lifetime being tracked inside the structure. In that way, it is a very helpful kind of documentation. One that is also always up-to-date, which is not a common feature of documentation in general.

@felix91gr You answer question: "... it is a very helpful kind of documentation ...", it is mix of multiple stuffs in one ... it breakage of lot of pattern like `SOLID` - Single Responsibility

Lets not mix life-time with good documentation ... from my point you should only know how to initialize `CompositeObject` that is all

Responsibility of proper handling life-time is responsibility of `CompositeObject` and developer that written it

Bypassing life-time from "low-level" structures to "high-level" structures (logically) is violation of layers responsibilities

Each abstract layer souls be responsible for its stuff

felix91gr: > it breakage of lot of pattern like SOLID - Single Responsibility

Those patterns were created as heuristics for Object-Oriented Programming, not as tips for general design.

In this case, having two responsibilities is good - it puts the information in exactly one place, instead of two. Keeping the documentation of what this lifetime is affecting becomes trivial once you have to keep it written, doesn't it?

redradist: > 
> 
> > it breakage of lot of pattern like SOLID - Single Responsibility
> 
> Those patterns were created as heuristics for Object-Oriented Programming, not as tips for general design.
> 
> In this case, having two responsibilities is good - it puts the information in exactly one place, instead of two. Keeping the documentation of what this lifetime is affecting becomes trivial once you have to keep it written, doesn't it?

No, those patterns could be applied to system that want to be designed good ;)

And if you want good documentation it is possible to extend rustdoc in way it read information form AST (in AST implicit life-times are available) and then generate good documentation

yvt: > Responsibility of proper handling life-time is responsibility of `CompositeObject` and developer that written it
>
> Bypassing life-time from "low-level" structures to "high-level" structures (logically) is violation of layers responsibilities

Actually, by storing references in `CompositeObject`, you are *explicitly* delegating the responsibility of managing the lifetime of the pointee to an upper layer (i.e., the owner of `CompositeObject`).

If exposing lifetime to an upper layer is not a desired option, you shouldn't store references in the first place.

redradist: > 
> 
> > Responsibility of proper handling life-time is responsibility of `CompositeObject` and developer that written it
> > Bypassing life-time from "low-level" structures to "high-level" structures (logically) is violation of layers responsibilities
> 
> Actually, by storing references in `CompositeObject`, you are _explicitly_ delegating the responsibility of managing the lifetime of the pointee to an upper layer (i.e., the owner of `CompositeObject`).
> 
> If exposing lifetime to an upper layer is not a desired option, you shouldn't store references in the first place.


Instead I need to use ugly hacks like `Borrow` or `BorrowMut` which even do not provide life-time ?
Seems like by such strategy language are enforing users to user `Borrow` and `BorrowMut` and all goodice of Borrow Checker disappear  (

Ixrec: Why are `Borrow`/`BorrowMut` relevant to any of this?

The problem here is one of fundamental soundness. If a struct stores a reference (directly, without `Box`ing), then all code which ever creates, manipulates or destroys instances of that struct _must_ be aware of that fact, or Rust the language (mostly the borrow checker in this case) can no longer guarantee "no UB in safe code". In particular, that means client code will fail to compile in some cases because of those stored references, and we want those compiler errors to make sense no matter how many layers of libraries there are in between.

When you really want to encapsulate a lifetime / make it an implementation detail clients don't need to care about, you simply need to make it live an arbitrarily long time by allocating, a.k.a. boxing it. That's true in every programming language; most of the others either autobox things by default or just let UB happen wherever. And yes, it's often a better design to pay the boxing cost to simplify things for your clients. But this whole proposal is (presumably) about cases where that boxing cost is unacceptable.

redradist: > 
> 
> Why are `Borrow`/`BorrowMut` relevant to any of this?
> 
> The problem here is one of fundamental soundness. If a struct stores a reference (directly, without `Box`ing), then all code which ever creates, manipulates or destroys instances of that struct _must_ be aware of that fact, or Rust the language (mostly the borrow checker in this case) can no longer guarantee "no UB in safe code". In particular, that means client code will fail to compile in some cases because of those stored references, and we want those compiler errors to make sense no matter how many layers of libraries there are in between.

Okay, agree about this statement, what if just as was suggested in this topic to do something like this:
```rust
struct CompositeObject<...'_> {
    obj0: &SomeType, # Deduce life-time for this reference
    obj1: &SomeType, # Deduce life-time for this reference
}
```
and generates the same code as I suggest ?

> 
> 
> When you really want to encapsulate a lifetime / make it an implementation detail clients don't need to care about, you simply need to make it live an arbitrarily long time by allocating, a.k.a. boxing it. That's true in every programming language; most of the others either autobox things by default or just let UB happen wherever. And yes, while "SOLID" is pretty C++-specific and not directly applicable to most projects, it's often a better design to pay the boxing cost to simplify things for your clients. But this whole proposal is (presumably) about cases where that boxing cost is unacceptable, and therefore encapsulation is simply not an option.

`SOLID` is not C++-specific, it could be applied to anything that work with objects (structs, classes and so on), in AST all parts is some kind of object ... Anyway it is a holy-war and does not make sense to disscuss it, please take a look what I suggested above, will it be appropriate ? Syntax `...'_`

pickfire: What happens if `CompositeObject` is not just `CompositeObject<'anon>` but `CompositeObject<'anon0, 'anon1>`? How can one specify the lifetime of which variable relates to which part of the lifetime?

redradist: @pickfire 
It will work as I described previously:
```rust
struct CompositeObject<...'_> {
    obj0: &SomeType, # Deduce life-time for this reference
    obj1: &SomeType, # Deduce life-time for this reference
}
```
to
```rust
struct CompositeObject<'anon0, 'anon1> {
    obj0: &'anon0 SomeType, # Deduce life-time for this reference
    obj1: &'anon1 SomeType, # Deduce life-time for this reference
}
```

I think that may this syntax would be more useful:
```rust
struct CompositeObject& {
    obj0: &SomeType, # Deduce life-time for this reference
    obj1: &SomeType, # Deduce life-time for this reference
}
```
and the compiler inference all life-times:
```rust
struct CompositeObject<'anon0, 'anon1> {
    obj0: &'anon0 SomeType, # Deduce life-time for this reference
    obj1: &'anon1 SomeType, # Deduce life-time for this reference
}
```
We would have goodness from all point of view ... User can see that `CompositeObject&` use some references inside, but at the same time user do not have to specify all life-times manually

pickfire: > We would have goodness from all point of view ... User can see that CompositeObject& use some references inside, but at the same time user do not have to specify all life-times manually

I understand that users can see some references, what I meant is what if the user wants to see all the references? Like checking if he/she could merge lifetime or something like that.

redradist: > 
> 
> > We would have goodness from all point of view ... User can see that CompositeObject& use some references inside, but at the same time user do not have to specify all life-times manually
> 
> I understand that users can see some references, what I meant is what if the user wants to see all the references? Like checking if he/she could merge lifetime or something like that.

Then author should specify directly life-time, but if user should not know about life-time at all, user will see just `CompositeObject&` - which means that class could contain some references, be attention !!

pickfire: class?

CryZe: >  but if user should not know about life-time at all

That can't ever be a thing. Lifetimes need to always be publicly known as they affect usage.

redradist: > 
> 
> > but if user should not know about life-time at all
> 
> That can't ever be a thing. Lifetimes need to always be publicly known as they affect usage.

It does not make sense !! You just want to see life-time, but there is no answer what you are doing with it ?

My answer - nothing, you just need to know that struct contains some references (from user point of view), and if there are some not trivial cases like this:
```rust
struct CompositeObject<'a, 'b: 'a> {
    obj0: &'a SomeType, # Deduce life-time for this reference
    obj1: &'b SomeType, # Deduce life-time for this reference
}
```
Developer still will not have other way, just to declare life-times

pickfire: Ah, now that is more understandable, so it means some cases are not covered. @redradist I think you should write that down in both the example (special case not supported) and in the drawbacks.

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+}
+
+struct BigObject {
+    composite_obj: CompositeObject,
+    count: i32,
+}
+
+struct Application {
+   big_obj: BigObject,
+}
```
pickfire, line 27:
```suggestion
struct Country {
    name: String,
}

struct World {
    country: Country,
    count: i32,
}

struct Application {
   country: Country,
}
```

Too many confusing terms are used here (`obj`, `object`), I believe it should could be something more understandable with a good example.

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+
+struct BigObject<'a> {
+    composite_obj: CompositeObject<'a>,
+    count: i32,
+}
+
+struct Application<'a> {
+   big_obj: BigObject<'a>,
+}
```
pickfire, line 43:
```suggestion
developer decides to make an inner item a reference:
```rust
struct City<'a> {
    name: &'a str,
}

struct State<'a> {
    city: Vec<City<'a>>,
    covid_deaths: i32,
}

struct Country<'a> {
   state: Vec<State<'a>>,
}
```

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+
+struct BigObject {
+    composite_obj: CompositeObject,
+    count: i32,
+}
+
+struct Application {
+   big_obj: BigObject,
+}
```
pickfire, line 60:
```suggestion
struct City {
    name: &'_ str,
}

struct State {
    city: Vec<City>,
    covid_deaths: u32,
}

struct Country {
   state: Vec<State>,
}
```

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+Code much simpler and more maintainable than fighting with named life-times in composite hierarchy
```
pickfire, line 63:
```suggestion
Code much simpler and easier to refactor throughout hierarchy of named lifetimes
by being less maintainable and more implicit
```

text/0000-hierarchic-anonymous-life-time.md
pickfire: Will the generated lifetime be in the docs or just used internally? If that is the case, maybe there is a term for it like "generated internally" or "syntactic sugar"?

text/0000-hierarchic-anonymous-life-time.md
redradist: Very funny )))))))

text/0000-hierarchic-anonymous-life-time.md
redradist: I do not think that generated life-time should be in docs due to that it does not have name ...
I think it should have just description that this structure has use references

CryZe: Yes, the lifetime needs to be in the docs. It is necessary for the user to know about the lifetime as it massively affects the usage of the struct. You need to design your code in a way that is compatible with the struct's lifetime semantics. Hiding those is incredibly dangerous in that it might mislead the developer into designing code that might not compile at all. Hiding lifetimes has already been deprecated, you are supposed to use at least `'_` nowadays, as it clarifies that a lifetime relationship is going on. So the absolute minimum that might make this RFC acceptable would be `struct Foo<'_>`. `<'_ ...>` probably is not working either, as that either just means the same thing as `<'_>` or it is incredibly fragile.

redradist: > 
> 
> Yes, the lifetime needs to be in the docs. It is necessary for the user to know about the lifetime as it massively affects the usage of the struct. You need to design your code in a way that is compatible with the struct's lifetime semantics. Hiding those is incredibly dangerous in that it might mislead the developer into designing code that might not compile at all. Hiding lifetimes has already been deprecated, you are supposed to use at least `'_` nowadays, as it clarifies that a lifetime relationship is going on. So the absolute minimum that might make this RFC acceptable would be `struct Foo<'_>`. `<'_ ...>` probably is not working either, as that either just means the same thing as `<'_>` or it is incredibly fragile.

Take a look at examples above with syntax like this:
```rust
struct CompositeObject& {
    obj0: &SomeType, # Deduce life-time for this reference
    obj1: &SomeType, # Deduce life-time for this reference
}
```
or even just notification that in future structure will use references:
```rust
struct CompositeObject& { // No references inside, but may will be in future
    obj0: SomeType,
    obj1: SomeType,
}
```

CryZe: I believe at this point the RFC is basically lifetime elision in type declarations with `&` instead of `'_` (which imo there's no reason to introduce new syntax when we already have `'_` for this. And additionally it introduces automatic elision for `'static` which is probably a bit more controversial (and should probably be a separate RFC afterwards).

shepmaster:
Using a global pandemic as example code is extremely poor judgement and only negatively reflects on this proposal.

text/0000-hierarchic-anonymous-life-time.md
redradist: > 
> 
> I believe at this point the RFC is basically lifetime elision in type declarations with `&` instead of `'_` (which imo there's no reason to introduce new syntax when we already have `'_` for this. And additionally it introduces automatic elision for `'static` which is probably a bit more controversial (and should probably be a separate RFC afterwards).

There is on big difference `'_` works and show only that in structure used only one reference with one life-time, but `CompositeObject&` shows that in structure used some references without explicitly declared life-time and number of fields could be different: 2, 3 ... whatever

CryZe: It doesn't, you can use a single lifetime but many references. Having more than one lifetime in a type's definition is only useful for disambiguating them for the user, but you can't do that with & anyway.

redradist: > 
> 
> It doesn't, you can use a single lifetime but many references. Having more than one lifetime in a type's definition is only useful for disambiguating them for the user, but you can't do that with & anyway.

Okay, will it be possible to use `'_` life-time without reference inside ?
```rust
struct CompositeObject<'_> { // No references inside, but may will be in future
    obj0: SomeType,
    obj1: SomeType,
}
```

felix91gr:
@shepmaster I think the covid stuff was @pickfire's thing, but I agree. That's bad taste.

text/0000-hierarchic-anonymous-life-time.md
```diff
@@ -0,0 +1,151 @@
...
+    population: u32,
+}
+
+struct State& {
+    cities: Vec<City&>,
+}
+
+struct Country& {
+   state: Vec<State&>,
+}
```
pickfire, line 60: This looks uglier than the original one. I don't quite like the trailing `&`, no taste IMHO.

nikomatsakis:
I'm going to go ahead and pre-emptively close this RFC. It's pretty clear from reading the text that it's not in a state to be accepted and still needs significant iteration. Moreover, we're in the process of moving to a new system (the [lang-team major change process](https://github.com/rust-lang/rfcs/pull/2936)) that is explicitly intended to capture 'early stage' thinking like this and try to decide if it's a direction we want to pursue, so I think this would be a better fit for that process. Thanks @redradist for the PR!

redradist:
@nikomatsakis Have you changed internal processes ? Should either I reopen it or continue discussion in https://internals.rust-lang.org/t/simplification-reference-life-time/12224/31 ?

PR #2978, opened 2020-08-29
Proposal for user namespaces on crates.io

jmillikin:


jmillikin:
cc @steveklabnik, who posted the original thread on crates.io package naming policy.

Diggsey:
The problem I have with scopes is that they don't really solve the problem they're claiming to solve and introduce a lot more problems, when you take into account the main constraint on crates.io:

- Crate identifiers should be permanent. If I depend on "foo/bar v1.0" or whatever, it should continue to refer to the same crate forever, modulo some *extremely* rare exceptions like legal issues.

If you introduce support for namespaces, you now have two new problems as a result of this constraint:

1) Namespaces are taken permanently, as soon as any crate is published under that namespace, so you still have the issue of exhausting the set of namespaces/picking an unused namespace. Also, because namespaces grant you so much power, there's a much stronger motivation for people to take all the short namespace names early and sit on them.

2) If you allow changing a namespace, or moving a crate between namespaces, you have to set up appropriate redirects from the old namespace. You now have the problem that multiple crate identifiers can refer to the same crate. Cargo needs to know whether two crates are the same or not, so now cargo needs an internet connection to do that dependency resolution...

If you tie those namespace to usernames, you make everything even worse:

- It's unclear how things like ownership transfer could happen when crates are tied to a specific user?
- I would rather not lock ourselves in to *github* usernames specifically...
- The need to change namespace names becomes much more common.

Taking a step back, the two motivations listed in the RFC are:

> it is often difficult to find a name that is short enough to be ergonomic when used as a symbol in Rust source code.

This could be solved by adjusting the default crate -> symbol name mapping. For example, by removing things before a `/`. It doesn't motivate tying the prefix directly to a username.

In any case, this would mean that users would need to override the symbol name in more cases, which could cause problems with macro expansions trying to refer to symbols under the wrong name. (I'm not sure this is a solved problem yet...)

> Introducing per-user namespaces will allow every crates.io user to upload packages without worrying about crate name conflicts.

This doesn't necessarily mean that we have to use the username as the scope. If this is really such a huge issue, crates.io could have a feature to generate an unused crate name based on a particular word.

Ultimately, the requirements of crate names being "meaningful", "unique" and "permanent" are always going to be in conflict, and the "unique" and "permanent" constraints are non-negotiable.

Personally, I would be happy to give up on the "meaningful" aspect - give all crates a "display name" which is distinct from their "identifier". Allow users to pick any "display name" they want, and make identifiers be randomly generated by crates.io.

jmillikin:
> Namespaces are taken permanently, as soon as any crate is published under that namespace, so you still have the issue of exhausting the set of namespaces/picking an unused namespace. Also, because namespaces grant you so much power, there's a much stronger motivation for people to take all the short namespace names early and sit on them.

Users already have to pick an unused username before they can register with an identity provider, which happens before they sign up for a crates.io account. I am not expecting a rush of people signing up for short GitHub names so that they can get "powerful" crates.io user namespaces.

> If you allow changing a namespace, or moving a crate between namespaces, you have to set up appropriate redirects from the old namespace. You now have the problem that multiple crate identifiers can refer to the same crate. Cargo needs to know whether two crates are the same or not, so now cargo needs an internet connection to do that dependency resolution...

I don't see why Cargo would need to know or care whether two crates.io packages are the same crate. It can't know that today, right? I can upload a single crate under many package names and Cargo would treat them all as separate dependencies.

The goal of redirects would be to avoid breaking previously registered package versions. Any other semantics can be decided by the crates.io maintainers such that it minimizes their support burden.

> It's unclear how things like ownership transfer could happen when crates are tied to a specific user?

I don't think ownership transfer -- removal of the original author's ownership -- would be possible for packages in a user namespace. If a package should be collectively owned, it should not be registered in a user namespace. If a user's package becomes so popular that they want it to become its own project, or if a user wants to hand off ownership to others, then they'll need to re-register it under a different name and communicate this to their users.

> I would rather not lock ourselves in to github usernames specifically...

Would the proposed alternative of explicit namespaces for identity providers solve this concern?

> In any case, this would mean that users would need to override the symbol name in more cases, which could cause problems with macro expansions trying to refer to symbols under the wrong name. (I'm not sure this is a solved problem yet...)

I don't think this would affect how often symbols need to be renamed, or macro expansions. Macros already must refer to their own crate with `$crate::` so that they can be independent of how their caller refers to their crate, and there's no need to rename symbols when build tools like Cargo have native support for separating crate names from package names.

> This doesn't necessarily mean that we have to use the username as the scope. If this is really such a huge issue, crates.io could have a feature to generate an unused crate name based on a particular word.

There's no need for crates.io to have that functionality. UUIDs can be valid crates.io package names, and tools such as https://www.uuidgenerator.net/ can generate UUIDs easily.

The goal of this proposal is to avoid the ergonomic issues that result from using generated package names.

> Ultimately, the requirements of crate names being "meaningful", "unique" and "permanent" are always going to be in conflict, and the "unique" and "permanent" constraints are non-negotiable.

We should be careful to distinguish "crates.io package" from "crate", because crates are consumed by `rustc` and have different naming requirements. There is no reason that a crate name must be _globally_ unique -- only unique within a single `[dependencies]` section or `rustc` invocation.

text/0000-crates.io-user-namespaces.md
```diff
@@ -0,0 +1,169 @@
...
+to apply to the crate name component of a namespaced package name.
+
+The syntax and allowed character set of the username portion is the same as for
+crates.io usernames.
+
+If a user requests that their username be changed (e.g. via a support request),
+then any existing package uploads under their old user namespace will be
+redirected to their new username. This RFC does not specify whether new versions
+would be available under the old package name, and recommends the crates.io
+team decide that based on expected support load.
```
pietroalbini, line 51: Unfortunately it can't work this way.

crates.io usernames are tied to the authentication provider, which is currently just GitHub (but that could change in the future), so username changes are handled by them. There is currently no involvement by the crates.io team when someone changes their username.

GitHub's policies are especially problematic for this proposal, as anyone can change their username as many times as they want by clicking a button, and the old username is instantly released into the available names pool. An example of how this could be problematic (and IMO a security vulnerability):

1. Andrea creates the `andrea` GitHub user, and publishes the `~andrea/foo` crate.
2. Andrea changes their username to `ferris-andrea`, freeing the `andrea` name.
3. An attacker creates the `andrea` GitHub user, and re-publishes `~andrea/foo` with malicious code in it.

The crates.io application solves this problem by referring internally to an user only by their GitHub account ID, treating the username as an arbitrary string that gets displayed to users. Because of that, changing username won't affect for example the ownership of your crates.

Also, what would happen on username changes should not be left at the crates.io team to decide, but it should be specified in the RFC itself.

text/0000-crates.io-user-namespaces.md
```diff
@@ -0,0 +1,169 @@
...
+used as a symbol in Rust source code.
+
+Introducing per-user namespaces will allow every crates.io user to upload packages without worrying about crate name conflicts.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+TODO: Is a guide-level explanation needed at all? There's already documentation
+on how to define dependencies in Cargo where the package name is different from
+the crate name, and this isn't adding any new capabilities.
```
pietroalbini, line 32: I would like to see what a person needs to do to add a user-namespaced dependency to their project, and how they would use it in their source code. I can roughly figure it out by reading the rest of the RFC, but I'm not sure it's exactly what you had in mind :)

text/0000-crates.io-user-namespaces.md
jmillikin: Would using a format like `~github:123456` work, then?

I really have no attachment to the details of how user IDs are represented, and if a particular identity provider does not provide stable user names then whatever stable identifier they support will have to do.

text/0000-crates.io-user-namespaces.md
jmillikin: Cargo supports specifying the package name, so something like this (copying from https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html):
```toml
[dependencies]
foo = { version = "0.1", package = "~someuser/foo" }
```

text/0000-crates.io-user-namespaces.md
pietroalbini: Yes, it would indeed work, but I'm not sure that's great from an user perspective: right now I can remember off the top of my head that if I want to serialize stuff I can depend on `serde` version `1`, but it's going to be hard to remember to which user ID it belongs to.

jmillikin: If a package author wants to let users depend on them without copy-pasting a `[dependencies]` line from the readme, then they'll have to register a non-namespaced crate.

pietroalbini:
> I don't think ownership transfer -- removal of the original author's ownership -- would be possible for packages in a user namespace. If a package should be collectively owned, it should not be registered in a user namespace. If a user's package becomes so popular that they want it to become its own project, or if a user wants to hand off ownership to others, then they'll need to re-register it under a different name and communicate this to their users.

I feel like this would prevent any project the author expects to possibly be successful from using user namespaces. Communicating the crate moved to the old users is not really feasible, as crates.io doesn't have a way to notify downstream users (and it can't add it, as we require no authentication to depend on crates).

> We should be careful to distinguish "crates.io package" from "crate", because crates are consumed by rustc and have different naming requirements. There is no reason that a crate name must be globally unique -- only unique within a single [dependencies] section or rustc invocation. And the mapping of crates.io package to crate is ~arbitrary, controlled entirely by the author of the build configuration.

While it's technically true that right now there is no enforced mapping between the crates.io crate name and the name you refer to in the source code, the mapping is still there by convention. I've yet to see a codebase where renamed dependencies are widely used, and when I see a crate name in the source code I implicitly expect it's the same name as the package.

Losing this property would be really bad IMO, as every project would have (for example) `http::Request` pointing to something different. Having to check the `Cargo.toml` every time I want to open docs.rs would be far from ideal for me.

Diggsey:
> I don't see why Cargo would need to know or care whether two crates.io packages are the same crate. It can't know that today, right? I can upload a single crate under many package names and Cargo would treat them all as separate dependencies.

At the moment, two packages are the same if they have the same name. Cargo needs to know whether two packages are the same in order to pick a compatible version.

It sounds like you are saying that when a package is moved, it's effectively a new package, with no compatibility at all with the previous package? It seems like that would make namespace renaming impossible in practice, because it would break all of your packages...

>  If a user's package becomes so popular that they want it to become its own project, or if a user wants to hand off ownership to others, then they'll need to re-register it under a different name and communicate this to their users.

Again, that's a lot of breakage for users of the package... Think about crates whose types are used in public APIs. For example, the UUID crate. Any time a new major version is published it causes a divide across the ecosystem.

> Would the proposed alternative of explicit namespaces for identity providers solve this concern?

It would certainly solve the specific concern of being tied to github.

> Macros already must refer to their own crate with $crate:: so that they can be independent of how their caller refers to their crate, and there's no need to rename symbols when build tools like Cargo have native support for separating crate names from package names.

I need to refresh my memory on this, but AIUI it's not that simple for procedural macros, and there are still cases today where procedural macros expect their crate to be at a particular path.

text/0000-crates.io-user-namespaces.md
pietroalbini: Could you add that to the guide level explaination? Something like "the user visits the crate's github page, copies this snippet from the readme [snippet], and uses the crate in the code like this [snippet]". I feel like showing the whole process a user needs to do to use a crate could help people form their opinion on the RFC.

jmillikin:
> I feel like this would prevent any project the author expects to possibly be successful from using user namespaces.

That seems fine. The goal is not to force use of namespaces, but to allow authors to upload crates with names that already exist as registered packages.

> Losing this property would be really bad IMO, as every project would have (for example) `http::Request` pointing to something different. Having to check the `Cargo.toml` every time I want to open docs.rs would be far from ideal for me.

If you find authors renaming crates in ways that don't make sense, that sounds like something to take up with those authors.

The primary alternative to this RFC for me is registering packages named with UUIDs, so that I don't have to worry about what might already be registered. This would work about as well for my workflow, which does not use Cargo, but I suspect that other people would be bothered by UUID names in crates.io and docs.rs.

Diggsey:
> The primary alternative to this RFC for me is registering packages named with UUIDs

I think UUIDs would be excessive: crates.io can act as the authority, so we can have shorter identifiers that are still unique. It could apply a short random suffix, eg. `foo/jqw`

pietroalbini:
> That seems fine. The goal is not to force use of namespaces, but to allow authors to upload crates with names that already exist as registered packages.

If that's the goal, what could prevent you from publishing today the crate `http` as `jmillikin-http`?

> If you find authors renaming crates in ways that don't make sense, that sounds like something to take up with those authors.

My concern is not with author choosing weird renames, but with having multiple crates from different authors being renamed as the same thing across multiple project. How would you personally handle depending on `~foo/http` in a codebase and on `~bar/http` in another unrelated codebase?

text/0000-crates.io-user-namespaces.md
pietroalbini: Hmm, I'm worried this could hurt the adoption of this feature. I think we should strive to find an ergonomic solution to the namespace name.

jmillikin:
> If that's the goal, what could prevent you from publishing today the crate `http` as `jmillikin-http`?

Please refer to previous discussions linked in the RFC for why ad-hoc name patterns do not solve the issue addressed by namespaces.

> My concern is not with author choosing weird renames, but with having multiple crates from different authors being renamed as the same thing across multiple project. How would you personally handle depending on `~foo/http` in a codebase and on `~bar/http` in another unrelated codebase?

Finding the definition of a symbol already requires inspecting `Cargo.toml` and `Cargo.lock` to find which package and version is being depended on, so nothing changes here. The question of whether crate names and package names could differ was resolved as soon as Cargo added that functionality.

jmillikin:
> It sounds like you are saying that when a package is moved, it's effectively a new package, with no compatibility at all with the previous package? It seems like that would make namespace renaming impossible in practice, because it would break all of your packages...

I don't think it's worth over-optimizing for the rare case of package renaming, which is currently not possible anyway.

pietroalbini:
> Finding the definition of a symbol already requires inspecting Cargo.toml and Cargo.lock to find which package and version is being depended on, so nothing changes here. The question of whether crate names and package names could differ was resolved as soon as Cargo added that functionality.

You're technically right, crates.io names and crate names inside the source code *could* differ. At least in my experience that's not true in any codebase I worked on. If I saw `use http::StatusCode;` I knew exactly which `http` crate we were talking about, and most of the times the documentation at `https://docs.rs/{crate}` (without specifying a version) returned useful results.

I did some grepping in the caches of a [Crater](https://github.com/rust-lang/crater) agent, and these were the renames I found:

* Out of 870 random GitHub repositories currently in the cache, none renamed dependencies.
* Out of 7000 crates.io dependencies currently in the cache (including multiple versions of the same dependency), there were 350 instances of a rename [(raw data)](https://gist.github.com/pietroalbini/6245ddb0d5c163f694cc4a2f45689f83). By glancing at the data, renames are used for:
  * `rustc-std-workspace-*`: those are needed for crates the standard library depends on, and that's not relevant with namespaces (it's just a trick needed to avoid the standard library build breaking). It also doesn't impact people wondering what crate they're using in the source code, as that's just an implementation detail of rustc's build system.
  * Renaming crates to work around limitations in Cargo's `--features`. Those renames follow the pattern of `{name}_crate`, so it's still easy to understand what they refer to.
  * Depending on multiple versions of the same crate by renaming the crate, for example, to `tokio01` and `tokio02`. It's still easy to understand what they refer to.
  * Simulating namespaces in big projects (like `gfx` or `parity`), where they rename `projectname-crate` to `crate`. This **is** indeed confusing.

Overral only less than 4.5% of the projects cached in that Crater agent used renames, and only a small portion of them used renames in a way I'd personally find confusing. For scale, the total amount of code I checked amounts to around 6.7% of the code downloadable by Crater, so while this is not a scientific result by any means it's not handwavy either.

I'm worried implementing this RFC as it's currently written could worsen the situation.

pietroalbini:
> I don't think it's worth over-optimizing for the rare case of package renaming, which is currently not possible anyway.

Hmm, currently you don't *have to* rename crates though, while with this RFC it would be required to remove the original author as an owner.

pietroalbini:
**Note:** while I'm a member of the crates.io team, this is just my opinion/wishlist on namespaces. Other team members might have a different opinion on this topic :)

---

While I'm not opposed to adding namespaces to crates.io, I feel like none of the proposals put forward over the years address all the concerns I'd have (and this is a situation [other team members are in too](https://internals.rust-lang.org/t/pre-rfc-user-namespaces-on-crates-io/12851/28?u=pietroalbini)).

The main requirements that come to my mind are:

* The proposal should be backwards compatible with the current situation.
* The namespaces should be stable and immutable like the rest of the registry, and it should not be possible to rename them. Once a namespace is created it should not be removed.
* Implementing namespaces shouldn't hurt the usability and ergonomics of crates.io, for example by having random or identifiers users aren't expected to remember.
* Namespaces should not restrict others' ability to publish packages without the namespace (for example,  by allowing people to reserve a `foo-*` prefix).
* Namespaced crates should be usable inside the source code without any renaming, and that should not create collisions.

The only proposal that comes close to satisfying all those efforts is automatically creating a namespace for every top-level package published. This would allow, for example, for the owners of the `foo` crate to publish `foo/bar`, while still allowing third parties to publish `foo-bar`. That proposal doesn't address using namespaced crates inside the source code, as that would probably require dedicated syntax in the language. Random examples that come to mind are:

```rust
use foo::crate::bar::Http;
use foo@bar::Http;
use foo:bar::Http;
use foo:::bar::Http;
```

None of these examples I thought of in the past 10 minutes are feeling "right" to me though, and I don't personally have the time to design what's best and gather consensus on it. Still, I'd be **personally** willing to accept an RFC that addresses that concern.

jhpratt:
@pietroalbini Regarding the syntax towards the end of your most recent comment, I don't believe it would be ambiguous to do something like `use foo/bar::Http`, unless I'm missing something? I feel that would also be the most natural way to import via a namespace.

Though as I'm typing this out, I also realized that whatever syntax is used has to work with fully qualified paths, such as `foo/bar::Http::new()`. That would presumably cause issues for most syntax proposals.

comex:
What about just `foo::bar::Http`?  (That would imply that `foo` itself should be purely a collection of sub-crates rather than a crate of its own, to avoid ambiguity with an item `bar` defined in `foo`.)

CAD97:
> Namespaced crates should be usable inside the source code without any renaming, and that should not create collisions.

I just want to note that the current no-namespace setup doesn't quite meet that requirement, either. It's perfectly valid today (though rare for one reason or another) to explicitly set `lib.name` such that the library's name is not the same as the package name.

Just as an example, I could publish a fork of `pest` as the package `cad97-pest` but with the library name of `pest`, so users could just replace `pest = "2"` with `cad97-pest = "2"` in their `Cargo.toml` and have no further changes to make.

I freely admit that namespaces could likely make the issue of name collisions more _common_, but I don't think it would make it _worse_, even if naïvely done. Packages can already provide library crates with arbitrary names completely unrelated to their package name.

You argue (rightly!) that "well-behaved" crates typically won't do that. I argue that a) presenting the falsehood that package name === crate name makes it _more_ confusing when it isn't the case, and b) it's not the "well-behaved" crates that would be running into name collisions, anyway, even in a world with any of the proposed namespace proposals.

jmillikin:
> Namespaced crates should be usable inside the source code without any renaming, and that should not create collisions.

This requirement feels like it's fundamentally incompatible with the idea of namespaces, which worries me. It essentially mandates that the crates.io package namespace be 1:1 with crate names, because if it were possible to have multiple packages containing a crate with the same name then there will be ambiguity when reading source that refers to _only_ the crate name.

I also feel like it's not reasonable to ascribe this to namespace when Cargo already provides that capability built-in. As you note, people are already experimenting with ad-hoc forms of namespacing via renames, and it doesn't seem to have caused any significant confusion in the community.

> Implementing namespaces shouldn't hurt the usability and ergonomics of crates.io, for example by having random or identifiers users aren't expected to remember.

I think that crates.io package names are already pretty close to arbitrary identifiers -- when working with Cargo I use dependency stanzas like `http = { package = "hyper" }` so I don't have to remember the codenames during normal day-to-day programming. If crates.io continues to follow the example of NPM then the situation will only move further in that direction.

pietroalbini:
> > Namespaced crates should be usable inside the source code without any renaming, and that should not create collisions.
>
>I just want to note (again) that the current no-namespace setup doesn't quite meet that requirement, either. It's perfectly valid today (though rare for one reason or another) to explicitly set lib.name such that the library's name is not the same as the package name.
>
> [...]
>
> You argue (rightly!) that "well-behaved" crates typically won't do that. I argue that a) presenting the falsehood that package name === crate name makes it more confusing when it isn't the case, and b) it's not the "well-behaved" crates that would be running into name collisions, anyway, even in a world with any of the proposed namespace proposals.

I probably should've worded that requirement a bit better. Sure, there is a small number of crates that does this today, and that's fine (those features were implemented for a reason!). Still, we shouldn't turn that into what *everyone* has to do.

> This requirement feels like it's fundamentally incompatible with the idea of namespaces, which worries me. It essentially mandates that the crates.io package namespace be 1:1 with crate names, because if it were possible to have multiple packages containing a crate with the same name then there will be ambiguity when reading source that refers to only the crate name.

Hmm, I don't think it's incompatible with namespaces. Being able to refer to the `Request` item inside `~jmillikin/http` in the code like this would satisfy that requirement (let's not bikeshed the syntax now):

```rust
use jmillikin::crate::http::Request;
```

If we do this (possibly with a less verbose syntax?) people who want to rename crates can still do that, but everyone else can just add `"~jmillikin/http" = "1"` to their `Cargo.toml` and start using it in their code.

> I also feel like it's not reasonable to ascribe this to namespaces when Cargo already provides that capability built-in. As you note, people are already experimenting with ad-hoc forms of namespacing via renames, and it doesn't seem to have caused any significant confusion in the community.

A very small number of project are using renames, not the community at large. This RFC as it's currently written would *force* everyone using namespaced packages to use renames though, and I'd personally prefer to avoid that.

> I think that crates.io package names are already pretty close to arbitrary identifiers -- when working with Cargo I use dependency stanzas like `http = { package = "hyper" }` so I don't have to remember the codenames during normal day-to-day programming. If crates.io continues to follow the example of NPM then the situation will only move further in that direction.

I'm glad renames are that useful for you! Still, we should consider the community at large when writing an RFC, and from what I gathered in the data earlier just a small amount of the community currently  relies on renames.

jmillikin:
> Hmm, I don't think it's incompatible with namespaces. Being able to refer to the Request item inside `~jmillikin/http` in the code like this would satisfy that requirement (let's not bikeshed the syntax now):
>
> ```rust
> use jmillikin::crate::http::Request;
> ```

That would be horrible, though. It further confuses the issues of crate names and package names, and it would require all users of a namespaced `http` crate to append a username (or, given the previous discussion a GitHub User ID) to all imports of that crate.

The goal of this RFC is to allow multiple crates named `http` to co-exist in the same registry, without having to change how they're referred to in code. Today, without that RFC, such a situation is only possible by using Git dependencies.

pietroalbini:
> That would be horrible, though. It further confuses the issues of crate names and package names, and it would require all users of a namespaced http crate to append a username (or, given the previous discussion a GitHub User ID) to all imports of that crate.

Well, people could choose to rename the crates on their own if they so wish, as they can do today. But yes, I think a namespaces RFC should encourage including the namespace in the import path as the default option, and provide a suitable syntax for that.

I understand it would be annoying for this current proposal, especially if we have to use IDs in the name, but it's not the only viable namespacing scheme. For example, if we were to grant a namespace for every top-level crate, `serde_derive` could be published as `serde/derive` and referred to in the code as:

```rust
use serde::crate::derive::Deserialize;
```

> The goal of this RFC is to allow multiple crates named http to co-exist in the same registry, without having to change how they're referred to in code. Today, without that RFC, such a situation is only possible by using Git dependencies.

Then I'm not sure I agree with the RFC :(

jmillikin:
> I understand it would be annoying for this current proposal, especially if we have to use IDs in the name, but it's not the only viable namespacing scheme. For example, if we were to grant a namespace for every top-level crate, serde_derive could be published as serde/derive and referred to in the code as:
>
> ```rust
> use serde::crate::derive::Deserialize;
> ```

That's not a namespace, though, that's just a longer and more complicated crate name syntax that is backwards-incompatible with the current Rust language.

What would be necessary to add _namespaces_, only, to crates.io so that it's possible for potential new contributors to register our packages on crates.io?

Diggsey:
What about leaning on the existing "registries" functionality?

The basic idea would be to allow users/organisations to create their own "sub-registries" that are hosted on subdomains or sub-paths of crates.io.

The rules for publishing to crates.io could be relaxed to allow packages in the main registry to depend on packages in these sub-registries.

You still have the issue that these sub-registries are permanent and cannot be renamed, so it may make sense to use IDs rather than names to identify them, but that's a separate question.

pietroalbini:
> That's not a namespace, though, that's just a longer and more complicated crate name syntax that is backwards-incompatible with the current Rust language. It has no benefits over giving every crate very_long_username_prefixed_names.

It is a namespace, and it does provide advantages: in the `serde/derive` example above, users know the crate is part of the "serde" project (and not a third party crate that happens to work with serde), and the authors of the serde project can publish new crates under their umbrella without worrying about naming conflicts.

It also allows you to create your own personal namespace by registering an empty "jmillikin" crate, which would grant you exclusive access to that namespace.

> What would be necessary to add namespaces, only, to crates.io so that it's possible for potential new contributors to register our packages on crates.io?

I already expressed my concern with *forcing* users of namespaced crates to refer to the crate ambiguously in the source code.

Aloso:
An alternative is to handle `/` the same way as `-`, by automatically substituting it with `_`. This can cause ambiguities when a crate depends on both `foo-bar` and on `foo/bar`, but these ambiguities will be very rare. In this case, the user can rename one of the crates:

```toml
[dependencies]
"foo/bar" = "1"
foo_bar_alt = { version = "1", package = "foo-bar" }
```

Usage:
```rust
use foo_bar::*;
use foo_bar_alt::*;
```

The renaming is inconvenient, but I believe it won't be necessary very often.

jmillikin:
> I already expressed my concern with _forcing_ users of namespaced crates to refer to the crate ambiguously in the source code.

They're not forced to refer to it ambiguously -- the crate name exposed to `rustc` continues to be defined by `Cargo.toml`, and they can use any value they want.

These are all fine:
```toml
[dependencies]
foo = { package = "~github:jdoe/foo", version = "1.0" }

[dependencies]
jdoe_foo = { package = "~github:jdoe/foo", version = "1.0" }

[dependencies]
com_github_jdoe_foo_v1_0 = { package = "~github:jdoe/foo", version = "1.0" }
```

The question of what crate name a package name gets mapped to is not under the control of the package registry, and the registry has no say in what downstream users choose to name their dependency references. This is true not only because of `Cargo.toml` supporting package renames, but also because of `extern crate foo as bar;` allowing arbitrary renames within the Rust source itself.

jmillikin:
Quick idea: what if the crates.io web UI defaulted to a long and detailed crate name, so that the default is the way you want (globally unique), but if `lib.name` is set explicitly then that's what gets shown instead?

## case 1: default behavior
```toml
[package]
name = "~jdoe/something"
```
is shown by crates.io as:
```toml
[dependencies]
jdoe_something = { package = "~jdoe/something" }
```

## case 2: lib.name override (no namespace)
```toml
[package]
name = "something-long-and-always-aliased"

[lib]
name = "something"
```
is shown by crates.io as:
```toml
[dependencies]
something = { package = "something-long-and-always-aliased" }
```


## case 3: lib.name override (with namespace)
```toml
[package]
name = "~jdoe/something"

[lib]
name = "something"
```
is shown by crates.io as:
```toml
[dependencies]
something = { package = "~jdoe/something" }
```

pietroalbini:
I'd still prefer that the `package` key is not used unless the person adding the dependency explicitly wants a different name.

> They're not forced to refer to it ambiguously -- the crate name exposed to rustc continues to be defined by Cargo.toml, and they can use any value they want.

Sorry, I should've been more clear. I was referring to breaking the convention that the crate name you refer to in the source code is *almost always* the name of the crates.io package. I totally understand you often rename your dependencies, but from the data I gathered earlier that's not how most of the community adds dependencies.

jmillikin:
Meta comment: I find this conversation extremely frustrating. I emailed help@crates.io asking whether the crates.io team was open to the idea of namespaces, _specifically_ so that I could avoid having an extended debate about whether the concept of namespaced crates.io packages is acceptable to the team. I did this because I'm not being paid to write design docs for Rust package registries, and at the current time I only have an hour or two per week available to spend on programming as a hobby.

The response I received from help@crates.io is that namespaces would be an option if the right design could be proposed, so I spent the time to draft and submit this RFC. If I had been told that one or more members of the crates.io team objected to the idea of namespaces in principle, then I could have short-circuited this wasteful exercise.

> Sorry, I should've been more clear. I was referring to breaking the convention that the crate name you refer to in the source code is almost always the name of the crates.io package.

I'm not sure this is a "convention", but rather a limitation of crates.io as it exists today. There is no way to register two crates with the same name, so of course the vast majority of crates will be referred to by a string that matches their crates.io package name.

The entire purpose of namespaces is to remove that limitation. It would allow more than one package to be registered with a particular crate name. In the vast majority of cases source code will continue to use crates exactly as they are named, but the crate name would be a subset of the crates.io package name.

I feel like arguing against package namespaces on the grounds of readability is similar to arguing against modules. It could be argued that `use foo::bar::baz; baz(); is less clear than requiring `foo_bar_baz()` to be spelled out each time, ala C, but this is not the direction that Rust went in and I would not want to use a language that enforced globally unique function names in the name of readability.

Similarly, I feel that it is a mistake for crates.io to enforce globally unique crate names as a readability issue. The benefits of allowing multiple people to register their libraries in a particular problem domain (whether it's HTTP clients, or XML parsers, or checksum algorithms, or so on) in my opinion far outweigh the minuscule cost of referring to `Cargo.toml` to disambiguate the source of a crate.

I'm going to close this proposal now because it doesn't feel like it's making progress, and we all have better ways to spend our time, but the experience has left me fairly sour on crates.io as a service. I remain hopeful that at some future point you will reconsider, and thereby allow me to register libraries so that Cargo users can depend on them.

CAD97:
I think the point of contention is what a "namespace" feature would mean in the first place.

To the crates team, it seems to *primarily* mean that a project can put multiple packages together under an umbrella, such that you know the packages are for-sure by the project.

To the people who feel most slighted by the crates team's approach here, namespaces *primarily* mean the ability to publish a crate with a desired name even if there's already a package published that provides a crate with that name, by putting the package into a different namespace such that the names do not clash.

If the latter party asks about "namespaces" and means the latter, and the team answers and means the former, you can see where the miscommunication enters, especially since the crates team has now communicated here the position that *generally*, the `package.name == lib.name` falsehood should not be made more false; i.e., the latter group's goal is an explicitly non-desired property from the crates team.

CAD97:
(separating this sentiment from the previous:)

To those who feel strongly about namespaces in the latter way described above: I suggest publishing your crates with a `package.name` with "ad-hoc namespacing" and an explicitly set `lib.name` that doesn't equal the `package.name`.

I'm personally of the camp that we will and should trend towards crate manifests declaring dependencies as `lib_name = { package = "...", ... }` in the future, specifically because of the fact that the `package_name = { ... }` form does not actually mean that the acquired library crate is `package_name`, because `lib.name != package.name`.

If the crates team feels strongly that the equivalence of the two names is a property worth keeping, it should be the case that crates *soon* deprecates and makes it an error to publish packages where that isn't true. It *already* is the case that the property doesn't hold, and the number of people using that fact can only grow. Imho, continuing to pretend the two names are the same while they aren't is harmful, because eventually users *will* run into a case where the two names aren't the same.

hoijui:
The reason I would want namespaces, is not so I can pick the name I want, but so I can publish crates and use the whole comfort that brings, even though I am still a newb, and some/many/all of the crates I publish will likely not be used by anyone but me, and I don't want to block common, simple names for that. The solution for that, which I was told in the rust newb chat, is to not publish, and rather define my inter-dependencies in Cargo.toml directly on the git URLs. That's what I am doing now, and it kind of gives me namespaces. ;-)
If there was some way to also register such projects in a registry (could be fediverse style too), so they could be efficiently discovered by others, I would not miss anything.

PR #2987, opened 2020-09-20
RFC 2582: fix implicit auto-deref of raw pointers

RalfJung:
The code as given wouldn't even compile.

### DIFF
--- a/text/2582-raw-reference-mir-operator.md
+++ b/text/2582-raw-reference-mir-operator.md
@@ -198,14 +198,14 @@ This has the side-effect of being able to entirely remove reference-to-pointer-*
 
 ```rust
 let x: *mut Struct = NonNull::dangling().as_ptr();
-let field: *mut Field = &mut x.field;
+let field: *mut Field = &mut (*x).field;
 ```
 
 The lint as described in this RFC would nudge people to instead write
 
 ```rust
 let x: *mut Struct = NonNull::dangling().as_ptr();
-let field: *mut Field = &raw mut x.field;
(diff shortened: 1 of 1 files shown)


text/2582-raw-reference-mir-operator.md
```diff
@@ -198,14 +198,14 @@ This has the side-effect of being able to entirely remove reference-to-pointer-*
...
-let field: *mut Field = &mut x.field;
+let field: *mut Field = &mut (*x).field;
 ```
 
 The lint as described in this RFC would nudge people to instead write
 
 ```rust
 let x: *mut Struct = NonNull::dangling().as_ptr();
-let field: *mut Field = &raw mut x.field;
+let field: *mut Field = &raw mut (x*).field;
```
SimonSapin, line 208:
```suggestion
let field: *mut Field = &raw mut (*x).field;
```

RalfJung: *oops* thanks!

RalfJung:
Who'd usually review such editorial changes to existing RFCs? @nikomatsakis?

pickfire:
Or maybe we should change the compiler such that it runs? I think the original code looks more ergonomic except that it doesn't work now.

RalfJung:
@pickfire raw pointer very deliberately do not auto-deref as dereferencing a raw pointer needs to take great care to avoid causing [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).

ehuss (approved):

PR #2994, opened 2020-09-28
Fix ABNF in RFC 2873

jethrogb:
`nostack` is specified further on in the RFC.

cc @Amanieu


lzutao:
The more up-to-date doc is in https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html .
Which may be pulled to Reference book when asm! is stabilized.

Amanieu (approved):

Amanieu:
Can you also make this change to the documentation in the unstable book in rust-lang/rust?

PR #2997, opened 2020-10-01
Add 0000-vecdeque-binary-search.md: Binary search fns for VecDeque

vojtechkral:


scottmcm:
> The benefit of this design is that it's very simple and just follows what's already established for slice and Vec.

I agree, and as such I think you can just submit it as a PR.  This would be an inherent method, thus can be stabilization gated, and there are no API questions since it's just following the existing precedent.

(The more general version would probably be setting new precedents -- for example, we only have sort on slices, not general iterators -- so that one would probably have more value in being an RFC.)

But of course I'm not on libs so you can always wait and see what they say.

PR #3026, opened 2020-11-29
Make generate-book.sh work on macOS.

ehuss:
The macOS `ln` command does not support the `-r` or `-t` flags.  This changes it so that it should work on all platforms.


ehuss:
Closing per #3215.

PR #3032, opened 2020-12-08
fix(text): update 0008-new-intrinsics.md

juzhiyuan:


text/0008-new-intrinsics.md
```diff
@@ -2,7 +2,7 @@
 - RFC PR: [rust-lang/rfcs#8](https://github.com/rust-lang/rfcs/pull/8)
 - Rust Issue: 
 
-** Note: this RFC was never implemented and has been retired. The
+**Note**: this RFC was never implemented and has been retired. The
 design may still be useful in the future, but before implementing we
 would prefer to revisit it so as to be sure it is up to date. **
```
ehuss, line 7: I think the intent was to make the whole thing bold.

```suggestion
**Note: this RFC was never implemented and has been retired. The
design may still be useful in the future, but before implementing we
would prefer to revisit it so as to be sure it is up to date.**
```

PR #3039, opened 2020-12-18
Fix typo in 1598-generic_associated_types.md

RichoDemus:


RichoDemus:
maybe this is actually correct english, nvm

PR #3042, opened 2020-12-25
Create 0000-target-feature-trait-methods.md

ejmahler:
[Rendered](https://github.com/ejmahler/rfcs/blob/target-feature-trait-methods/text/0000-target-feature-trait-methods.md)

### DIFF
--- a/text/0000-target-feature-trait-methods.md
+++ b/text/0000-target-feature-trait-methods.md
@@ -1,0 +1,251 @@
+- Feature Name: `#[target_feature(..)]` In Trait Methods
+- Start Date: 2020-12-25
+- RFC PR: none yet
+- Rust Issue: none yet
+
+[RFC 2045 (`target_feature`)]: https://github.com/rust-lang/rfcs/pull/2045
+[RFC 2396 (`target feature 1.1`)]: https://github.com/rust-lang/rfcs/pull/2396
+
+# Summary
+[summary]: #summary
+
+This RFC builds on [RFC 2045 (`target_feature`)] and [RFC 2396 (`target feature 1.1`)] by:
+* Allowing trait methods to have `#[target_feature(..)]` attributes
+* Expanding target_feature 1.1's relaxed safety rules to include calls to and from the newly-allowed `#[target_feature(..)]` trait methods.
+
+# Motivation
+[motivation]: #motivation
+
+Imagine we're writing an AVX library with the goal of being generic over f32 and f64.  We can start by creating a trait
+called `AvxVector` with some methods common to both f32 and f64 vectors:
+
+```rust
+pub trait AvxVector {
+    unsafe fn add(left: Self, right: Self) -> Self;
+}
+impl AvxVector for __m256 {
+    unsafe fn add(left: Self, right: Self) -> Self {
+        _mm256_add_ps(left, right)
+    }
+}
+impl AvxVector for __m256d {
+    unsafe fn add(left: Self, right: Self) -> Self {
+        _mm256_add_pd(left, right)
+    }
+}
+```
+The `RustFFT` crate implements a [real-world example of this](https://github.com/ejmahler/RustFFT/blob/simd/src/avx/avx_vector.rs#L16)
+
+Now, users of our library can compute a sum of f32 and f64 vectors with the same code:
+```rust
+#[target_feature(enable = "avx")]
+unsafe fn do_work() {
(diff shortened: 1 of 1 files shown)


joshtriplett:
This looks great.

One bit of feedback: I think it should be possible to omit the `#[target_feature]` on the trait impl, if and only if you're not actually *using* the `#[target_feature]` for that particular impl. It wouldn't be implicitly added; if you want it on your impl, you'd still have to write it.

The idea is that the `#[target_feature]` on the trait definition requires *callers* of the trait to either use `unsafe` or have the requisite target feature themselves. However, just because the caller meets that requirement doesn't mean the callee has to *use* that feature. So, the trait impl could leave off the `#[target_feature]` if it wants to.

text/0000-target-feature-trait-methods.md
```diff
@@ -0,0 +1,251 @@
...
+    fn add(left: Self, right: Self) -> Self;
+
+    fn sub(left: Self, right: Self) -> Self;
+
+    #[target_feature(enable = "avx")]
+    fn mul(left: Self, right: Self) -> Self;
+
+    #[target_feature(enable = "avx")]
+    fn div(left: Self, right: Self) -> Self;
+}
```
Nokel81, line 93: Lets say that you wanted all these methods to be `#[target_feature(enable = "avx")]`. Should it be possible to mark the `trait` itself as such? Perhaps this could be an "alternative" or "future work".

```rust
#[target_feature(enable = "avx")]
pub trait AvxVector {
    fn add(left: Self, right: Self) -> Self;

    fn sub(left: Self, right: Self) -> Self;

    fn mul(left: Self, right: Self) -> Self;
    
    fn div(left: Self, right: Self) -> Self;
}
```

ejmahler:
> 
> 
> This looks great.
> 
> One bit of feedback: I think it should be possible to omit the `#[target_feature]` on the trait impl, if and only if you're not actually _using_ the `#[target_feature]` for that particular impl. It wouldn't be implicitly added; if you want it on your impl, you'd still have to write it.
> 
> The idea is that the `#[target_feature]` on the trait definition requires _callers_ of the trait to either use `unsafe` or have the requisite target feature themselves. However, just because the caller meets that requirement doesn't mean the callee has to _use_ that feature. So, the trait impl could leave off the `#[target_feature]` if it wants to.

I've been thinking about this since you commented, and I have some thoughts on it.

On one hand, a disadvantage is that both the compiler implementation and the user's understanding of how this feature works would have to be more complicated. If you're writing a trait impl, you can't simply look at the top of the method and see which instruction sets are available. Additionally, in order to even begin taking advantage of this feature, there is more information for the programmer to internalize about which code is allowed and which code isn't.

On the other hand, an advantage is that if the trait is a part of a public API, adding target features to a trait method would no longer cause trait impls to fail to compile. Impls in another crate would still compile, they just won't use the new target features that the trait has added. It would still be a breaking change overall, because *callers* of that public trait would have to deal with the new feature being added.

If we could make it so that adding features wasn't a breaking change at all, I would say it would be worth it to go with your suggestion -- but I don't see how that's possible. And without that, the only benefit would be added convenience for code writers, at the expense of making the code a little harder to read. Which IMO is not a good tradeoff.

yanchith:
Sorry if this has been discussed elsewhere, but I wonder what can be done to improve the experience of `core::ops` with `#[target_feature]` and SIMD. For instance, `core::ops::Add` is not target-featured, so this would also disallow its impls to be target-featured, making it impossible to write:

```rust
#[derive(Clone, Copy)]
pub struct f32x8(__m256);

impl f32x8 {
    #[inline]
    #[target_feature(enable = "avx")]
    pub fn splat(a: f32) -> Self {
        let v = unsafe { _mm256_set1_ps(a) };
        Self(v)
    }
}

impl core::ops::Add for f32x8 {
    #[inline]
    #[target_feature(enable = "avx")]
    pub fn add(self, rhs: Self) -> Self {
        let v = unsafe { _mm256_add_ps(self.0, rhs.0) };
        Self(v)
    }
}

#[target_feature(enable = "avx")]
fn hello() {
    let a = f32x8::splat(1.23);
    let b = f32x8::splat(42.0);
    let c = a + b; // SAFE, because the trait impl for core::ops::Add enables AVX
}
```

I know that this is mostly wishful thinking, and that unless the trait itself forces additional constraints, instanced functions can't rely on the additional constraints, but still this is something that would be nice to have.

PR #3087, opened 2021-02-22
fix: share trait was renamed to sync

ibraheemdev:


scottmcm:
I'm torn on this one.  The RFC is so old that I'm not sure that updating all the references is the best way forward.  It's [expected](https://github.com/rust-lang/rfcs/#the-rfc-life-cycle) that RFCs are "point in time", in a sense, and the the reference is the up-to-date explanation of the current state.  And the trait name isn't the only thing out of date -- it also mentioned "default traits", which were renamed to OIBITs and now `auto trait`s.

Inspired by [the note I added to 2203](https://rust-lang.github.io/rfcs/2203-const-repeat-expr.html), I wonder if it might make since to instead just put a note at the top, something like
```md
> ⚠ Note to modern readers ⚠
>
> This is a very old RFC.  For context, what's now `Sync` was at the time known as `Share`, and "default traits" evolved into what's now known as `auto trait`s.
```

Thoughts, lang folks?

Havvy:
I agree with @scottmcm that we should avoid modernizing RFC language and that they are more of a historical artifact. Adding a preamble to resolve historical context is a good solution.

m-ou-se:
Closing this per the above comments.

PR #3099, opened 2021-03-31
RFC: `cargo`-`miri` integration

mark-i-m:
I've been meaning to write this RFC for years now, but the opportunities just keep passing me up... until now 🎉

### DIFF
--- a/text/0000-cargo-miri-integration.md
+++ b/text/0000-cargo-miri-integration.md
@@ -101,7 +101,7 @@ where `$PATH_TO_FOO` is either the path specified in `Cargo.toml` or `<sysroot>/
 # Drawbacks
 [drawbacks]: #drawbacks
 
-The slowness of `miri` may increase build times mildly. However, this expense is deemed small in comparison to vast might of safe-build-system-yness that will be unleashed on a UB-impoverished world.
+The slowness of `miri` may increase build times mildly. However, this expense is deemed small in comparison to the vast might of safe-build-system-yness that will be unleashed on a UB-impoverished world.
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
@@ -120,7 +120,7 @@ Seems obvious, right? `miri`.
 
 > - What is the impact of not doing this?
 
-It is highly likely that the world will end.
+It is highly likely that the world will end due to UB. Or at best my laundry might get eaten.
 
 # Prior art
 [prior-art]: #prior-art
@@ -143,4 +143,4 @@ Ken Thompson's excellent [Reflections on Trusting Trust](https://dl.acm.org/doi/
 The obvious extension would be to get rid of the system call interface of `corgi`, as this represents attack surface and added complexity.
 One way to do so would be to reimplement `Linux` (or really any OS) and run it inside of `corgi` with the rest of the build system.
 
-For added security, one can follow the [Defense in Depth](https://csrc.nist.gov/glossary/term/defense_in_depth) strategy: implement `corgi` inside `vim` on Linux inside `corgi`.
+For added security, one can follow the [Defense in Depth](https://csrc.nist.gov/glossary/term/defense_in_depth) strategy: implement `corgi` inside `vim` on Linux inside `corgi` inside `vim` on Linux inside `corgi` inside `vim` on Linux inside `corgi` ... until it's safe enough.
--- a/text/0000-cargo-miri-integration.md
+++ b/text/0000-cargo-miri-integration.md
@@ -1,0 +1,146 @@
+- Feature Name: `cargo_miri_integration`
+- Start Date: 2020-04-01
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


Lokathor:
Feels like you're a day early.

kennytm:
@Lokathor  The date is correct in Japan.

text/0000-cargo-miri-integration.md
```diff
@@ -0,0 +1,146 @@
...
+
+There have been many [discussions on IRLO about sandboxing code](https://internals.rust-lang.org/search?q=sandbox).
+
+Ken Thompson's excellent [Reflections on Trusting Trust](https://dl.acm.org/doi/10.1145/358198.358210) underscores the importance of trusting one's infrastructure. As we all know, if a program is UB-free, it is entirely trustworthy.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Are corgis really the best dog to name this tool after? The name at least sounds like `cargo` a bit...
+- What about all the cat-lovers in the rust community?
```
Kixiron, line 138: Who needs 'em? This is a non-issue, please remove it from the unresolved questions as it's quite self-evident

Lokathor:
I suppose technically so, but midnight-30 is still pretty early

mark-i-m:
Yeah, I wrestled with that too... problem is if I posted it when I wake up on April 1, it will already be over for most of the world...

mark-i-m:
I hate living on an ellipsoid :( Maybe somebody should post an RFC to change that....

kennytm:
```sh
corgi fetch -- ellipsoid
```

RDambrosio016:
Thank you for this RFC, i always feel that people use too many dependencies and cargo should make build times exponentially worse the more dependencies you use, up until the deno folks have to figure out how to run cargo for 3 weeks with their ~400 dependencies. Have you considered running everything on top of valgrind? then for every byte leaked you add 1 second to the build times just to discourage leaking? It would also make miri even slower, thereby increasing the effectiveness of this RFC.

mark-i-m:
> then for every byte leaked you add 1 second to the build times just to discourage leaking?

Actually, this seems like a good approach to bugs in general... for each bug in the program, add 1 second to the run time.

Ekleog:
I mean I've got a butterfly net so that allows me to detect some of the bugs

shadowtime2000:
@Ekleog How would we download the butterfly net to catch the bugs for everyone

shadowtime2000:
@mark-i-m I feel emails are too basic for such a task. 

We should create a server exactly at the center of the earth and have it digitally send butterfly nets through websocket connections that are opened by `cargo` whenever a command is run and never closes until the computer shuts down. That way everyone can have butterfly nets and we can also punish people for using `cargo`.

Why would we want to punish people for using `cargo`?
Reject Rust, return to Assembly.

mark-i-m:
> Why would we want to punish people for using cargo?
> Reject Rust, return to Assembly.

Hmm... assembly such a pleasure to work with, though, especially compared to rust... is it really a punishment?

rodrimati1992:
> @mark-i-m But aren't bugs... like bugs and u can't actually easily detect them?

But bugs add weight to the program, so obviously they will make it slower, even if by a little.

text/0000-cargo-miri-integration.md
RalfJung: :cat2:

text/0000-cargo-miri-integration.md
```diff
@@ -0,0 +1,146 @@
...
+[drawbacks]: #drawbacks
+
+The slowness of `miri` may increase build times mildly. However, this expense is deemed small in comparison to the vast might of safe-build-system-yness that will be unleashed on a UB-impoverished world.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+> - Why is this design the best in the space of possible designs?
+
+Seems obvious, right? `miri`.
```
oli-obk, line 111: Yes, this is the best reason, and anyone who argues otherwise should obviously be writing their build scripts in assembler

oli-obk:
> As a lead of the @rust-lang/miri team, I fully approve of this proposal. :)

🚢 it

text/0000-cargo-miri-integration.md
petrochenkov: >What about all the cat-lovers in the rust community?

[Miri is a cat](https://www.youtube.com/watch?v=d8zXlwEijN4) in the first place, so their needs are already covered.

mark-i-m:
So can we get an fcp merge?

shadowtime2000:
> > @mark-i-m But aren't bugs... like bugs and u can't actually easily detect them?
> 
> But bugs add weight to the program, so obviously they will make it slower.

So let's basically detect everything as a bug because everything is relatively slow but then once we make something slow that is a bug making us make it even slower basically creating a never ending program. Wonderful.

programmerjake:
maybe there could also be a feature where if compilation takes more than 5s, then it switches to bitcoin mining for 10min giving you time to find bugs by reading through your source code again! obviously bitcoin mining is chosen since your basically guaranteed to have a hashrate only sufficient to make a few cents per day worldwide (due to ASICs being waay better than cpus at sha256). the resulting bitcoin can be used once a year (each april 1st) to fund cat food treats for miri!

RalfJung:
I think it is time to close this. :)  Thanks all!

pickfire:
This rfc does not have a rendered link.

programmerjake:
@pickfire https://github.com/mark-i-m/rfcs/blob/cargo-miri/text/0000-cargo-miri-integration.md

PR #3110, opened 2021-04-16
Fix RFC 243 code blocks and add note

mbartlett21:
This fixes the code blocks that broke in my previous commit and adds a note about RFC 2046 `label-break-value`


ehuss (approved):
Thanks!  I'm usually reluctant to add new content to old RFCs. RFCs are not living documents, and represent the state when the RFC was accepted. This is small enough that I think it's fine to add, but I don't think it is something we should make too much of a habit on without a broader policy decision on how to manage such changes.

PR #3118, opened 2021-05-03
Fix 2-rfc 'Active RFC List' hyper links

8picomu:
Links was wrong, so I fixed it.

### DIFF
--- a/text/0002-rfc-process.md
+++ b/text/0002-rfc-process.md
@@ -83,7 +83,7 @@ Whomever merges the RFC should do the following:
 * Create a corresponding issue on [Rust repo](https://github.com/rust-lang/rust)
 * Fill in the remaining metadata in the RFC header, including links for
   the original pull request(s) and the newly created Rust issue.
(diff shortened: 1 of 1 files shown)


text/0002-rfc-process.md
```diff
@@ -83,7 +83,7 @@ Whomever merges the RFC should do the following:
 * Create a corresponding issue on [Rust repo](https://github.com/rust-lang/rust)
 * Fill in the remaining metadata in the RFC header, including links for
   the original pull request(s) and the newly created Rust issue.
-* Add an entry in the [Active RFC List] of the root `README.md`.
+* Add an entry in the [Active RFC List](https://rfcbot.rs/) of the root `README.md`.
```
kennytm, line 86: changing the link won't match the subsequent text though. 

as the entire point is outdated, but we don't update the content of an outdated RFC, imo we shouldn't touch this link at least.

8picomu: I see, I understand

PR #3121, opened 2021-05-06
Fix minor typo in const-ub RFC

teor2345:
It looks like "anything else" is what was meant here.


teor2345:
@RalfJung just a quick typo fix for one of your RFCs :-)

RalfJung:
Ah yes, this is one of the things that are hard for foreign speakers to get right. ;) When you use "everything" vs "anything" is very different in German and English. Thanks for catching this!

So, LGTM -- but I do not have PR merge powers in this repo.

Mark-Simulacrum (approved):

PR #3147, opened 2021-07-13
io-safety: add missing link reference definitions

wchargin:
This RFC contained some links that referred to Rust symbols but lacked a
Markdown link reference definition, so they were just parsed as normal
text.

Test Plan:
In the rendered document, the character `[` now only appears as part of
the phrase `&[u8]`.

wchargin-branch: io-safety-link-refs


ehuss (approved):
Thanks!

PR #3153, opened 2021-07-22
Update all references to my name

lilyball:
As mentioned in rust-lang/rust#87357, this repo has no `.mailmap`. Is there a reason for this?


jyn514 (approved):
I don't feel qualified to sign off on adding a mailmap here - @ehuss do you know if there's a team or group in charge of the RFCs repo?

ehuss:
I think it is essentially everyone, which means in practice that nobody is in charge.  Ultimately I think it is the core team, but they are not active here.

jyn514:
Gotcha. @lilyball if you want to make a PR adding a mailmap I'm happy to merge it :)

Issue #3152, opened 2021-07-20
Reserving "nobreak" as a keyword for Rust Edition 2021


JelteF:
The reason I'm opening this issue is that there has been some activity on the old issues, regarding returning values with `break` from `for` and `while` loops:
1. https://github.com/rust-lang/rfcs/issues/961
2. https://github.com/rust-lang/rfcs/issues/1767

Rereading those threads, the main issues why those were originally closed seem to be:
1. `else` being a confusing name for this block
2. not being able to add a new keyword that makes the meaning of the block clear
3. there was no experience with the `loop` version of the `break <value>` syntax

Number 1 is still valid (although some disagree). 3 is not valid anymore, this feature [was stabilized more than 4 years ago in 2017](https://github.com/rust-lang/rust/pull/42016). Number 2 seems much less of a problem these days, since we now have editions (which we didn't have when these original issues were opened).

That's why I'm proposing to reserve `nobreak` as a keyword in Rust Edition 2021. That way we at can add this feature in the future, using a clear keyword. 

I think `nobreak` clearly shows the meaning of the block. It's the block that you enter when there was no break. Code would then look like this:
```rust
let found_six = for val in [1, 4, 3, 2] {
    if val == 6 {
        break true;
    }
} nobreak {
    false
}
```

scottmcm:
I'm going to close this because based on https://github.com/rust-lang/rfcs/pull/2441#issuecomment-395256368, lang has decided not to do speculative reservations of *specific* keywords any more.

The new plan is to generally reserve space with https://github.com/rust-lang/rfcs/pull/3101, and use that reserved space as a transition plan (perhaps via something like https://github.com/rust-lang/rfcs/pull/3098) to allow the feature to be used on the latest edition with slightly-inferior syntax until the keyword can be fully reserved in a future one, *after* the RFC for the whole feature has been accepted.

(Note that this is neither a statement in opposition nor in support of the feature.  Just that the way forward is to RFC the whole feature, including the eventual keyword -- not to separate out the keyword first.)

PR #3163, opened 2021-08-19
Proposal: Else clauses for for and while loops

Starwort:
Did I do this right?

### DIFF
--- a/0000-else-clauses-for-and-while.md
+++ b/0000-else-clauses-for-and-while.md
@@ -1,0 +1,382 @@
+- Feature Name: Else clauses for `for` and `while` loops
+- Start Date: 2021-08-19
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Else clauses for `for` and `while` loops would clarify 'break with value' semantics for these loops, and allow for a more natural way of running code after a loop has completed without breaking.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC proposes the introduction of for/else and while/else loops; an extension to loop syntax that allows code to run if the loop is never broken. Using these loops simplifies the previous conversations around `for`/`while` break-with value semantics. It also allows for a more natural way of running code after a loop has completed without breaking.
+
+For example, the following code represents a very common pattern:
+
+```rs
+let mut found = false;
+
+for value in data {
+    if value == target {
+        found = true;
+        break;
+    }
+}
+
+if !found {
+    println!("Couldn't find {}", target);
+    return;
+}
+```
+
+This requires an additional variable, increasing visual complexity; and allows a future maintainer to add more code between the initial variable, the loop, and the variable checking code, which could change behaviour (e.g. if the checking code returns on failure, code added prior to the check would run unconditionally).
+
+This proposal introduces the following syntax, with the intent of preventing these issues:
+
+```rs
+for value in data {
+    if value == target {
+        break;
+    }
+} else {
+    println!("Couldn't find {}", target);
+    return;
+}
+```
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When using a `loop` block, it can return a value by `break`ing with that value. For example:
+
+```rs
+let x = loop {
+    if (...) {
+        break 42;
+    } else {
+        // do work
+    }
+};
+```
+
+This code sets x to 42, after doing some work.
+
+If we instead want to loop over some values, and either return one of these values or a default value, we can use a `for` block:
+
+```rs
+let x = for i in 0..10 {
+    if (...) {
+        break i;
+    }
+} else {
+    11 // default value
+};
+// x is now a usize; either a value 0-9, or 11 if our condition was never met
+```
+
+If we wanted to instead use a while loop, we could use a `while` block instead:
+
+```rs
+let x = while (...) {
+    // do work
+    if (...) {
+        break Some(value);
+    }
+} else {
+    Some(42) // default value
+};
+```
+
+The reason the `else` keyword is reused here is twofold:
+
+1. It is not a new keyword, meaning all code written before the implementation will still work identically (as currently break-with-value semantics are not allowed in for/while loops)
+2. It makes sense within the semantics of `for`/`while` desugaring; while loops desugar to code similar to the following:
+
+    ```rs
+    loop {
+        if (...) {
+            ... // while loop code
+        } else {
+            break {
+                ... // else clause code - for while loops without an else clause, this is an empty block
+            };
+        }
+    }
+    ```
+
+    For loops, on the other hand, desugar to code similar to the following:
+
+    ```rs
+    loop {
+        if let Some(value) = iterator.next() {
+            ... // for loop code
+        } else {
+            break {
+                ... // else clause code - for for loops without an else clause, this is an empty block
+            };
+        }
+    }
+    ```
+
+    As for loops are essentially a special case of while-let loops, it's easy to see how those desugar:
+
+    ```rs
+    loop {
+        if let (...) {
+            ... // while-let loop code
+        } else {
+            break {
+                ... // else clause code - for while-let loops without an else clause, this is an empty block
+            };
+        }
+    }
+    ```
+
+    Seeing how the constructs desugar, it is easy to see how the `else` clause can be used to add behaviour when the loop terminates without breaking.
+
+It is allowed to assign a for or while loop without an else clause to a variable, as long as it does not contain a break-with-value. The value of the variable is `()`.
+
+However, it is *not* allowed to assign a for or while loop (containing a break-with-value) without an else clause to a variable, as this would lead to an inconsistent type (as using break-with-value would then need to return `Option<T>`, which is incompatible with `()`).
+
+Sample error:
+```rs
+fn main() {
+    let x = for i in 0..2 { break 2; };
+    println!("{}", x);
+}
+```
+
+Leads to:
+
+```
+error[EXXXX]: `break` with value from a `for` loop missing an `else` clause
+ --> src/main.rs:2:13
+  |
+2 |     let x = for i in 0..2 { break 2; };
+  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^ for loops which break with a value must have an `else` clause
+  |             |
+  |             you can't `break` with a value in a `for` loop without an `else` clause
+  |
+help: add an `else` clause to this `for` loop
+  |1
+2 |     let x = for i in 0..2 { break 2; } else { 42 };
+  |                                        ^^^^^^^^^^^
+```
+
+It is also not allowed to create a for/else or while/else with inconsistent types.
+
+Sample error:
+```rs
+fn main() {
+    let x = for i in 0..2 { break 2; } else {"foo"};
+    println!("{}", x);
+}
+```
+
+Leads to:
+
+<!-- I created this mock error before realising that the `loop` with incompatible types case gives a simpler error message: Perhaps this could be used to improve that case?
+
+```
+error[E0308]: mismatched types
+ - -> src/main.rs:2:13
+  |
+2 |     let x = for i in 0..2 { break 2; } else {"foo"};
+  |         -                         -          ----- expected `usize` because of prior `break` with a value
+  |         |                         |
+  |         |                         found value type `usize`
+  |         cannot determine consistent type; can be either `usize` or `&'static str`
+``` -->
+
+```
+error[E0308]: mismatched types
+ --> src/main.rs:2:46
+  |
+2 |     let x = for i in 0..2 { break 2; } else {"foo"};
+  |                                              ^^^^^ expected integer, found `&str`
+```
+
+Similarly, incompatible break-with-values are disallowed, as with `loop` blocks.
+
+Sample error:
+```rs
+fn main() {
+    let x = for i in 0..2 {
+        if i == 0 {
+            break 2;
+        } else {
+            break "hello";
+        }
+    };
+    println!("{}", x);
+}
+
+```
+
+Leads to:
+
+```
+error[E0308]: mismatched types
+ --> src/main.rs:6:19
+  |
+6 |             break "hello";
+  |                   ^^^^^^^ expected integer, found `&str`
+```
+
+Loops are allowed to use an `else` clause even if they do not contain a break-with-value. This is useful for code that is only executed if the loop terminates normally. Loops which do not contain a break-with-value will not be used as an implicit return value (so they do not need to be followed with a semicolon).
+
+This loop does not contain a break-with-value, and will not be used as an implicit return value:
+
+```rs
+for i in 0..2 {
+    if i == 3 {
+        break;
+    }
+} else {
+    println!("`i` never reached 3.");
+}
+```
+
+This loop, however, contains a break-with-value, and will be used as an implicit return value, with the type `&str`:
+
+```rs
+for i in 0..2 {
+    if i == 3 {
+        break "found 3";
+    }
+} else {
+    "didn't find 3"
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+For/else and while/else largely do not interact with other features of the language, however:
+
+- For/else and while/else loops are valid expressions, and their type must be consistent.
+  - All `break` statements must contain a value of the same type, and the `else` clause must implicitly return the same type.
+  - Alternatively, all `break` statements must not contain a value, and the `else` clause must implicitly return `()`.
+- For/else and while/else loops which do not contain a break-with-value are not used as implicit return values, but may be assigned to variables (and return the unit type `()`, as for/while loops without else clauses do).
+- For and while loops without an else clause are still not allowed to contain a break-with-value.
+- Else clauses may not contain breaks (this is an unresolved decision). This is because it could be seen as ambiguous whether the break refers to the `else` clause or an enclosing loop.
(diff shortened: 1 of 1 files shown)


lebensterben:
This would be redundant given RFC https://github.com/rust-lang/rust/pull/63177

Starwort:
Perhaps I'm just tired - it's 11:30 PM here - but I fail to see the connection between that pull and this proposal

lebensterben:
See https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.try_find

> Applies function to the elements of iterator and returns the first true result or the first error.

Your motivating example is

>```rust
>let mut found = false;
>for value in data {
>    if value == target {
>        found = true;
>        break;
>    }
>}
>if !found {
>    println!("Couldn't find {}", target);
>    return;
>}
>```

With `try_find`, it returns early either on a match (when the closure returns an `Ok(true)`), or on a error.

For an `Iterator<Item = T>`, this returns `Result<Option<T>>`, which contains all the information you possibly need:
- Whether a failure occured
- Whether a match was found and if so the matched item

So you can write something like:
```rust
if let Ok(Some(target)) = hay.iter().try_find(|value| Ok(value == target)) {
  /* do something */
} else {
  /* do something else */
}
```

lebensterben:
The PR contains the following examples, which can all be written in a concise and readble way without proposed new syntax:

---

Proposed syntax:
```rs
let x = for i in 0..10 {
    if (...) {
        break i;
    }
} else {
    11 // default value
};
```

With `Iterator::find()`:
```rust
let x = (0..10).find(|v| Ok(v == target)).unwrap_or_else(11);
```

---

Proposed syntax:

```rs
let x = while (...) {
    // do work
    if (...) {
        break Some(value);
    }
} else {
    Some(42) // default value
};
```

With `std::iter::successors()` and `Iterator::try_find()`
```rust
use std::iter::successors;

let x = successors(Some(initial_value), |v| {
  Some(/* do something on v */)
}).try_find(|v| {
  if /* outer conditional */ {
    Ok(/* inner conditional */)
  } else {
    Err()
  }
}).unwrap_or_else(42);
```

You can use `std::iter::successors()` with `Iterator::last()`:
```rust
use std::iter::successors;

let x = successors(Some(initial_value), |v| {
  if /* check on v */ {
    Some(/* do something on v */)
  } else {
    None
  }
}).last().flatten();
```

clarfonthey:
I feel like using dedicated syntax for this instead of just iterator adapters does have promise, and could help in a fair few numbers of cases where such adapters aren't possible, e.g. async code.

I dunno. I like filling in this gap where `while` and `for` can't break values, but `loop` can.

SoniEx2:
Honestly would rather have

```rust
#![feature(label_break_value)]
fn main() {
    let x = 'foo: {
        for &a in [1, 2, 3].iter() {
            if a % 2 == 0 {
                break 'foo a;
            }
        }
        0
    };
}
```

In fact, you can do this today, on nightly: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=51e6caacaf3e6fae2d6dd68dfbc56c77

burdges:
Rejected previously in https://github.com/rust-lang/rfcs/issues/961#issuecomment-292746759 and https://github.com/rust-lang/rfcs/issues/1767#issuecomment-292678002

clarfonthey:
Both comments state that they're open to revisiting this.

Starwort:
That said: Here's a piece of code from one of my own projects (a Connect 4 AI). I cannot see a simple way of transforming this into any of the cases you propose; however it uses the exact pattern for/else is supposed to improve:
```rs
let mut owner = col.pieces[y];
let mut found_line = true;
for offset in 0..4 {
    let col = &self.cols[x + offset];
    if col.count < y + offset {
        found_line = false;
        break;
    }
    if col.pieces[y + offset] != owner {
        found_line = false;
        break;
    }
}
if found_line {
    return Some(match owner {
        Piece::White => GameResult::WhiteWin,
        Piece::Black => GameResult::BlackWin,
    });
}
```
I'd be interested in seeing how complex the functional approach can get with a real example (although obviously you do not have to write code for me if you do not want to)

lebensterben:
@Starwort 

```rust
let owner = col.pieces[y];
let mut found_line = true;

let found_line = &self.cols[x..x + 4]
    .iter()
    .enumerate()
    .find(|(offset, col)| col.count < y + offset || col.pieces[y + offset] != owner)
    .is_none();

if found_line {
    return Some(match owner {
        Piece::White => GameResult::WhiteWin,
        Piece::Black => GameResult::BlackWin,
    });
}
```

SoniEx2:
If you consider the label-break-value desugaring:

```rust
#![feature(label_break_value)]
fn main() {
    let x = 'foo: {
        for &a in [1, 2, 3].iter() {
            if a % 2 == 0 {
                break 'foo a;
            }
        }
        0
    };
}
```

then a good keyword for it would be "then":

```rust
fn main() {
    let x =
        for &a in [1, 2, 3].iter() {
            if a % 2 == 0 {
                break a;
            }
        } then {
            0
        };
}
```

but do we really want to add a new keyword vs [stabilizing an existing, already-implemented feature?](https://github.com/rust-lang/rust/issues/48594)

mcarton:
The “Prior art” section gives really good reasons _not_ to have this syntax.

> This feature is in both [Python (with identical semantics)](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops) and Guarded Command Language. Additionally, there has been [a proposal for this feature in Golang (closed by bot)](https://github.com/golang/go/issues/41348) and [in JuliaLang](https://github.com/JuliaLang/julia/issues/1289) (with many people proposing/preferring the Python-influenced semantics used in the JuliaLang thread, in addition to backing up the motivation).

The Go proposal mentioned in the RFC got 23 :-1: and no support vote, while the Julia proposal has been open since 2012 with little activity. This seems to indicate that people generally feel no need for such syntax.

> Unfortunately, many Python users are unfamiliar with the syntax; of [an informal survey](https://blog.glyphobet.net/blurb/2187/), only around 25% knew the meaning and 55% gave an incorrect meaning.

If only 1 out of 4 Python users know what it means, and 1 out of 2 think it means something else, this indicates to me that copying this syntax with this semantic is a very bad idea. It's too niche for most users to know, and not intuitive for users who don't know.

Starwort:
> It's too niche for most users to know, and not intuitive for users who don't know.

As Rust is a systems language, I would expect that more people would have an understanding of the desugar (at its core, a while loop just uses a guard to `goto` the loop's beginning; therefore an `else` can be attached to that guard) than Python, which is a very high-level language. Additionally, the point of that survey was that the responders couldn't look at documentation; those who already know the syntax could use it fine, those who don't can use the guard version, and those who encounter it can look at the feature's documentation and join the first group.

> then a good keyword for it would be "then":

I believe I mentioned this but the reason `then` was rejected was due to it reading as if it always occurs after a loop; with label-break-value this makes sense but it is less obvious that the entire block would be skipped (as the previous one closes first)

@lebensterben: It takes me far longer to parse that code than the equivalent for-else construct (which is really the point of the construct) - I can just about manage but it is definitely harder to read, which kinda supports my point. I'd be interested in a poll between the three options for readability though:

:one:
```rs
let owner = col.pieces[y];
for offset in 0..4 {
    let col = &self.cols[x + offset];
    if col.count < y + offset {
        break;
    }
    if col.pieces[y + offset] != owner {
        break;
    }
} else {
    return Some(match owner {
        Piece::White => GameResult::WhiteWin,
        Piece::Black => GameResult::BlackWin,
    });
}
```
:two:
```rs
let owner = col.pieces[y];

let found_line = &self.cols[x..x + 4]
    .iter()
    .enumerate()
    .find(|(offset, col)| col.count < y + offset || col.pieces[y + offset] != owner)
    .is_none();

if found_line {
    return Some(match owner {
        Piece::White => GameResult::WhiteWin,
        Piece::Black => GameResult::BlackWin,
    });
}
```
:three:
```rs
let owner = col.pieces[y];

&self.cols[x..x + 4]
    .iter()
    .enumerate()
    .find(|(offset, col)| col.count < y + offset || col.pieces[y + offset] != owner)
    .is_none()
    .then(|| match owner {
        Piece::White => GameResult::WhiteWin,
        Piece::Black => GameResult::BlackWin,
    })
```

kennytm:
TBH you should also include the `label_break_value` variant SoniEx2 has mentioned into your (small) office poll 🙂 

```rust
    'found: {
        let owner = col.pieces[y];
        for offset in 0..4 {
            let col = &self.cols[x + offset];
            if col.count < y + offset {
                break 'found;
            }
            if col.pieces[y + offset] != owner {
                break 'found;
            }
        }
        return Some(match owner {
            Piece::White => GameResult::WhiteWin,
            Piece::Black => GameResult::BlackWin,
        });
    }
```

clarfonthey:
I actually do like the idea of `then` as a keyword a lot. One of my biggest issues with the Python syntax is that the word "else" implies that the block will be run *instead of* the loop, instead of *unless* it breaks. With if-else, we mentally separate out the two branches as separate paths, whereas for-else are combined paths.

A for-then or while-then gets rid of that problem. I don't think that adding in an extra keyword should be seen as that big of a detriment, but we *technically* could use `do` instead, although I'd hesitate against it since it would be too similar to do-while loops.

To me, for-then says "do this loop, then that" and it's not too difficult to understand that a break would skip the then. Whereas for-else is "do this loop, *otherwise* do this if it *doesn't break*," which is much harder to understand.

Starwort:
What would you think about the following instead:
```rs
for i in 0..2 {
    // code
} broke (value) {
    value
} else {
    other_value
}
```
The `broke` clause could be omitted for the semantics as above, would not need the `(value)` if bare `break`s were used, and is probably clearer to those opposing for/else
The loop would still need a consistent type for each break-with-value, but it would only have to match the type of the else if the `broke` clause was removed.

Does that seem like an acceptable compromise?

Starwort:
Now that I'm thinking about it more, I actually like `while`/`broke`/`else` even more than my original RFC as common cleanup code could be moved into the `broke` clause, while still allowing everything the original proposal does

clarfonthey:
`broke` feels like it's getting to the point where you're just mimicking try blocks. At that rate, you're replacing `break` with `yeet` and `catch` with `broke`.

The "else" block avoids storing extra state whereas the "broke" block just calls a function for every broken value.

Starwort:
That's the primary reason why it would be optional; in most cases for/else would be used but in cases with complicated exit code the `broke` clause would be used for DRY purposes

PatchMixolydic:
#3152 suggests `nobreak` as a replacement for else (which was also mentioned in passing in [this comment](https://github.com/rust-lang/rfcs/pull/3163#issuecomment-902466968)).
```rust
let owner = col.pieces[y];

for offset in 0..4 {
    let col = &self.cols[x + offset];
    if col.count < y + offset {
        break;
    }

    if col.pieces[y + offset] != owner {
        break;
    }
} nobreak { // `k#nobreak` before Rust 2024 if RFC 3098 is merged
    return Some(match owner {
        Piece::White => GameResult::WhiteWin,
        Piece::Black => GameResult::BlackWin,
    });
}
```

Personally, I'd still prefer the iterator or `label_break_value` approaches. Neither of them would require reserving a unique keyword or utilizing a strange take on an existing keyword, and the `label_break_value` solution seems much more readable in my eyes, especially to those unfamiliar with for-else constructs.

durka:
In Python I find this feature both *useful* (as this RFC says at the beginning, it's a common pattern that is cumbersome to solve with a boolean) and also *very confusing*.

I can never remember that `else` means "loop did *not* `break`" because, to me, `break` is sort of an exceptional thing to do, so I first think that should jump to the `else` clause, the same way exiting early from `try` ends up in `catch`... but wait, that's not how for/else works, is it... unless... oh dear, better Google it this time just like every time.

So yeah, I agree this is a problem, but I do not support using `else` as the solution. Perhaps some macro crates can experiment with other ideas like `broke` suggested in previous comments.

joshtriplett:
@durka I agree entirely with that characterization: we need a recommended, idiomatic, Rustic solution to the underlying problem (handling loops that don't find what they're looking for), but not `for ... else`.

fstirlitz:
It’s a shame that something so apparently hastily written was submitted here. I had been drafting [a much more elaborate proposal that actually addresses the Python objection](https://github.com/fstirlitz/rust-rfcs/blob/exhaustive_ballot/text/0000-loop-exhaustion.md). But now that *this* is being rejected, it makes me doubt it’s going to have a fair hearing.

ssokolow:
@fstirlitz I don't hold prejudice against a concept based on prior proposals and I don't get the impression Rust team members do either.

Your idea is certainly better, with `exhausted` being less confusing than `else`... but, being perfectly honest, I'm still not convinced having a special `exhausted` case brings enough value to justify itself. That was always half my concern.

(eg. What about `empty` like how `else` is used in templating languages? It feels questionable to privilege this specific case but not that one. ...and, if you address both, how do we go about minimizing the sense of added complexity?)

lebensterben:
@fstirlitz 
Your draft has the same problem as this PR, that they are lack of meaningful example.

scottmcm:
Note that `ControlFlow`+`try_fold`+`try` gives some pure-library phrasings of these things, which can help.

For example, that "you can write your own Iterator::find" example becomes
```rust
let found = haystack.into_iter().try_for_each(|item| try { 
    if predicate(item) {
        return ControlFlow::Break(item);
    }
}).break_value();
```

And in general the `else { FOO }` can, using that, often become `.break_value().or_else(|| FOO )`.

So for-else feels like it's a strange intermediate, to me.  It's more complex than the basic structured control flow stuff, but also not particularly generalizable to more complex situations.  I think I'd be inclined more towards leaving it out, but having either label-break-value or the more elaborate custom-fsm construct instead.  

> it makes me doubt it’s going to have a fair hearing.

I wouldn't worry too much about that.  Take https://github.com/rust-lang/rfcs/pull/1603, for example, which was one of the more heavily 👎'd RFCs on this repo and was not accepted.  But the same idea came back a year and a bit later as https://github.com/rust-lang/rfcs/pull/2113 where it was well-received, and it's soon going to be *required*, with edition 2021.

If there's a change that's a good-enough fit, it'll happen, even if similar things were rejected many times before.

Treeniks:
I specifically remember being overwhelmed by the amount of options Rust has for loops when I first learned the language. With the existence of standard `for` and `while`, the rather unusal `loop` (compared to other languages) and the `while let` variant, Rust is already a lot more complicated when it comes to simple loops than most languages.

`for` and `while` is standard in pretty much every C-like language, and they also function the exact same across all those languages, including Rust. `while let` is a natural extension of `while` with Rust's Enum system, whereas `loop` is something I haven't seen in other languages yet (not that I know many), but also naturally integrates with Rust's "everything is an expression" philosophy. But the proposed `exhausted` syntax personally doesn't seem to be quite such a natural extension (without looking at the underlying desugaring, which is something I would rather not have to think about while programming) and also doesn't exist widely across C-like languages, thus also the confusion with the python survey I assume. I feel adding features that don't exist in most programming languages to syntax that does exist in most programming languages will add unneeded confusion for anyone either coming from another language and trying to learn Rust, or for anyone who has to switch between different language regularly.

While I personally find the proposed alternatives using Iterators to be very unreadable, the `label-break-value` alternative looks very similar to a `for ... exhausted` syntax, while using a natural extension of Rust's labels and leaving no room for confusion. I feel as though, if we were to give a programmer who has never seen or written Rust code before a snippet of code with the `for ... exhausted` syntax and a snippet with the `lable-break-value` syntax (the latter with specific `return` keywords), they would have a far easier time understanding the `label-break-value` syntax (in terms of Control Flow at least), not to mention they could be 100% sure about the Control Flow, whereas the `for ... exhausted` syntax, while the `exhausted` keyword is certainly more accurate than `else`, could still potentially leave open questions. With the `else` keyword, there would exist multiple interpretations of what it could mean, all of which exist in some other language, so they would have to pick one at random. With the `exhausted` keyword, they could *guess* what that keyword probably means, but it would be a keyword that is exclusive to Rust and thus it would still be a guess of what it means exactly. I feel as though there is something beautiful in a solution for which you don't need to know the language to be able to understand it, and the `label-break-value` syntax certainly achieves this better (although still not perfect) in my opinion.

fstirlitz:
That’s 461 words spent on an argument that boils down to ‘but it’s not very familiar to the average programmer’.  To which I reply: unfamiliarity is temporary.  This is not a difficult concept to grasp once explained, and it can be explained in less than two minutes.  Someone who has never seen Rust before is going to spend much more time on lifetime annotations than on this relative triviality.  You’d have to be pretty selective with your snippets already if you want to find something that will be *immediately* obvious to someone entirely unfamiliar with Rust.  (And even then you’re making assumptions about them being familiar with something else. To someone unfamiliar with *any* programming, all code is incomprehensible.)

The same argument could have been made about the `match` statement a while ago: until quite recently, pattern-matching was known only from relatively obscure languages like Haskell, Scala and ML.  Now languages like Java and Python have gotten one, and there are C++ and *PHP* proposals in the works, even though it doesn’t add any expressive power; you could do the same things with `switch` or `if` chains in those languages.

jhpratt:
Without reading the reference-level explanation (which shouldn't be necessary to understand your proposal), I sincerely have no idea what is even being proposed. I am _aware_ that other languages have for/else & where/else, but have **never** seen it used in the wild (probably because nobody knows what it does). The comparison to pattern matching with `match` is a stretch given that any programmer could immediately think of situations where it's useful. 

You stated five days ago that "now that this is being rejected, it makes me doubt it’s going to have a fair hearing." It has not yet been rejected, as FCP has not yet completed. There is a ten day period for a reason. If there are clear and obvious reasons to include this, then it should be so stated. But when an experienced programmer has no idea what's even being proposed, that _by itself_ is reason to close in my opinion.

erkinalp:
For-else implemented here as a macro: https://github.com/erkinalp/pythonic-for

PR #3170, opened 2021-08-31
Update 1211-mir.md

kw-fn:
looked at those references to comments which imply LLVM is the only option and added formatting for clarity around trans and borrowck

### DIFF
--- a/text/1211-mir.md
+++ b/text/1211-mir.md
@@ -12,7 +12,7 @@ well-suited to type-checking and translation.
 # Motivation
 
 The current compiler uses a single AST from the initial parse all the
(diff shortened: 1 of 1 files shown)


nikomatsakis:
I'm fine with landing these changes, they seem small and could help avoid confusion for future readers. That said, we do often leave RFCs to represent the proposal "at the time it was made", rather than generalizing them. I would not particularly want to (for example) modify the RFC to describe current MIR, I would rather than we keep the rustc-dev-guide up to date.

kw-fn:
Yes, realizing that these are legacy document, there's a similar nit here: https://github.com/rust-lang/rfcs/pull/3083

PR #3179, opened 2021-10-04
Do not include headers in search index

notriddle:
This significantly shrinks the pre-compressed search index:

    $ du -h searchindex-old.js searchindex-new.js
    26M	searchindex-old.js
    19M	searchindex-new.js

And shrinks the search index even after it's gzipped:

    $ du -h searchindex-old.js.gz searchindex-new.js.gz
    4.5M	searchindex-old.js.gz
    3.3M	searchindex-new.js.gz

This change requires a newer version of mdBook, with https://github.com/rust-lang/mdBook/pull/1637

## Screenshots

Before:

image

After:

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Install mdbook
       run: |
         mkdir mdbook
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!  This seems reasonable to me. Since there is such a large number of pages, I think having the search target individual sections probably isn't super useful.

PR #3207, opened 2021-12-12
Fix typo

sigmaSd:


nrc:
ping @rust-lang/libs seems like this could just be merged?

m-ou-se:
I don't think we should change the words of the original author to something else, if we're not sure it's just a typo. It seems to be intentional.

PR #3212, opened 2021-12-29
Change feature name, fix grammatical error

jhpratt:
The actual feature name used in the compiler is `derive_default_enum`. This changes the RFC to mirror that and additionally removes a leftover word from drafting (that is grammatically incorrect and could be confusing).


ehuss (approved):
👍

PR #3232, opened 2022-02-15
add raw-code-data

0loli:
[rendered](https://github.com/0loli/rfcs/blob/master/text/0000-raw-code-data.md)


shepmaster:
There are important sections left unfinished and the reference section doesn't appear to be thorough enough to attempt implementing such a feature. It doesn't appear that this RFC is in a state to be properly reviewed; perhaps you should retract this and start with a pre-RFC on one of the forums.

PR #3233, opened 2022-02-16
RFC: add `try_all` and `try_any` to `Iterator`

FlixCoder:
[rendered](https://github.com/FlixCoder/rfcs/blob/RFC-proposal/text/0000-iterator-try_all-try_any.md)

Hi!

As this is my first RFC / PR here, I am sorry if there is something missing or wrong.

Thank you for reviewing and suggestions :)

### DIFF
--- a/text/0000-iterator-try_all-try_any.md
+++ b/text/0000-iterator-try_all-try_any.md
@@ -1,0 +1,144 @@
+- Feature Name: iter_try_all_any
+- Start Date: 2022-02-16
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+
+[summary]: #summary
+
+Add the associated methods `try_any` and `try_all` to the `Iterator` trait, which work similar to `any` and `all`, but accept `Result` and `Option` as closure return type.
+
+# Motivation
+
+[motivation]: #motivation
+
+In some cases where one would like to use `any`/`all`, the actual condition to check is fallible. For example, let's say we have a simple key-value-store like a `HashMap`. The keys are IDs to the objects. We further have a list of IDs we want to check some condition on. We would like to use `all` on an iterator over the list to ensure all objects match our condition. However, access to elements is fallable and returns `Option`s. In other cases, the checks might involve more complex validation on the objects. The validation method could involve access to external parts and fail. Therefore, it returns a `Result<bool, Error>`.
+
+The implementor needs to return a `bool` for every element in `all`. They could use `.try_fold(true, |state, item| Ok(state && check(item)?))`, but this is less readable and does not short-circuit as `all` does. If they would do `.all(|item| check(item).unwrap_or(false))`, the error would be discarded and lost/ignored. To fully comply with expectations, there is only two possibilites (I see):
+
+- a `for` loop is required to iterate over the elements and stop early on errors or invalid items
+- collecting the elements in the iterator to a collected `Result`, then calling `all` on the inner collection
+
+Both solutions are not very elegant. The `try_all` method would work as `.try_all(check)`, returning `Result<bool, Error>` as well. It should behave similar to `all`, except working with `Try<bool>` output, just as `try_fold` behaves in comparison to `fold`.
+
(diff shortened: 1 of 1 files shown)


text/0000-iterator-try_all-try_any.md
```diff
@@ -0,0 +1,144 @@
...
+
+# Motivation
+
+[motivation]: #motivation
+
+In some cases where one would like to use `any`/`all`, the actual condition to check is fallible. For example, let's say we have a simple key-value-store like a `HashMap`. The keys are IDs to the objects. We further have a list of IDs we want to check some condition on. We would like to use `all` on an iterator over the list to ensure all objects match our condition. However, access to elements is fallable and returns `Option`s. In other cases, the checks might involve more complex validation on the objects. The validation method could involve access to external parts and fail. Therefore, it returns a `Result<bool, Error>`.
+
+The implementor needs to return a `bool` for every element in `all`. They could use `.try_fold(true, |state, item| Ok(state && check(item)?))`, but this is less readable and does not short-circuit as `all` does. If they would do `.all(|item| check(item).unwrap_or(false))`, the error would be discarded and lost/ignored. To fully comply with expectations, there is only two possibilites (I see):
```
scottmcm, line 18: Short-circuiting on both is probably easiest done with `try_for_each`, something like
```rust
    let r = it.try_for_each(|x| match check(x) {
        Ok(b) => if b { ControlFlow::Continue(()) } else { ControlFlow::Break(Ok(false)) },
        e => ControlFlow::Break(e),
    });
    let r = match r {
        ControlFlow::Continue(()) => Ok(true),
        ControlFlow::Break(r) => r,
    };
```
<https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=746bd31e2418ffacdff08855c91e6205>

But yes, that's of course much less readable than having the method provided by `Iterator`.

text/0000-iterator-try_all-try_any.md
```diff
@@ -0,0 +1,144 @@
...
+}
+```
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+Adding new methods to the trait makes it even bigger, but I would argue that is a low cost (especially in this small case) for a nice benefit.
+
+# Rationale and alternatives
```
scottmcm, line 112: ## Can there be a general adapter for these?

I did recently r+ `try_collect` for nightly (https://github.com/rust-lang/rust/pull/94041), but now I'm wondering if instead of having all of these -- especially for non-adapters -- we should expose the [shunting mechanism](https://github.com/rust-lang/rust/blob/master/library/core/src/iter/adapters/mod.rs#L130-L200) that they use somehow instead.

Notably, the `.try_collect()` could plausibly be exposed as `.try_mut(|i| i.collect())`, so I wonder if there'd be a way that `try_all` and `try_find` and friends could also be exposed via a similar mechanism.  (And thus remove `try_collect` and `try_find` and such from the trait altogether.  `try_fold` is a slightly different case as it's there to be the internal iteration primitive, and is only secondarily there as something to be used directly.)

text/0000-iterator-try_all-try_any.md
FlixCoder: Oh interesting, thanks, that is another possibility indeed. Not a good one of course, I would expect anyone to go for another route.

text/0000-iterator-try_all-try_any.md
timvermeulen: I think this is unlikely (without major language changes).

`try_find`/`try_all`/etc have to deal with a closure, which makes them a lot more complicated than `try_collect`. If you wanted to implemented `try_find` in the shape of `.try(|i| i.find(...))` then you'd have to return a `bool` from the closure, but you have a closure that returns an `impl Try<Output = bool>`. So how do you handle a failure? There are two hurdles:
- The closure needs to communicate to the iterator that a failure has happened and iteration can stop. This probably requires having a struct that wraps the given closure and implements the `Fn*` traits. This struct can then 
- After reporting the failure, the closure still needs to return _something_ (a `bool` in this case). It doesn't even matter much what it returns, as the iteration stop before the next element anyway, but it needs to return something. A `Default` bound might solve it for most methods, but what about e.g. `try_max_by_key`?

We can almost get there, but not quite all the way it seems. Implementing non-`try` methods in terms of their `try` counterparts is, of course, a lot simpler.

PR #3250, opened 2022-04-01
Extend Greenspun's tenth rule to cover the Rust Standard Library

mbartlett21:
This RFC extends Greenspun's tenth rule to cover the Rust Standard Library

### DIFF
--- a/text/0000-lisp-impl.md
+++ b/text/0000-lisp-impl.md
@@ -1,0 +1,94 @@
+- Feature Name: (`lisp_impl`)
+- Start Date: (2022-04-01)
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC extends Greenspun's tenth rule to cover the Rust Standard Library.
+
+# Motivation
+[motivation]: #motivation
+
+Greenspun's tenth rule, [in its original form,][1] states that it only applies to `C` or `Fortran` programs. However, there is an opportunity for the rule to be expanded, by including `Rust` under its umbrella. The expected outcome is that there will be a documented interface to use the interpreter (with bugs and slowness included).
+
+[1]: https://philip.greenspun.com/research/
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Contained in the `std::greenspun` module is a buggy [Lisp][2] interpreter that can be called programmatically from any Rust program that links the standard library.
+
+[2]: https://en.wikipedia.org/wiki/Lisp_(programming_language)
+
+## Examples
+
+Evaluating a Lisp program embedded in a macro:
+```rust
+use std::greenspun::{li, evaluate, LispVal};
+
+let a = 1.;
+
+let program = li!(car (cons {a} 2.5));
+
+let result = evaluate(program)?;
+
+assert_eq!(result, LispVal::from(2.5));
+```
+
+Invalid Lisp may be rejected, complete with bad errors to show to end users:
+
+```rust
+use std::greenspun::{li, evaluate};
+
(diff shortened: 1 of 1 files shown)


Nilstrieb:
While I do really like the general idea of this RFC, I have a concern regarding the usage of a macro. I don't think we should use a macro for a core feature like this, but instead add this to the Rust syntax proper, to make it easier and simpler to use.

Would there be any parsing ambiguities with this, and if yes, could we remove the ambiguous feature? 

Another problem with my proposal would be that this is in `std` and not `core`, so adding first class language support can't really work. Maybe we can add a simplified version of it to `core` as well?

Diggsey:
I think the example doesn't show the true benefits of having an embedded Lisp. It needs more brackets.

text/0000-lisp-impl.md
```diff
@@ -0,0 +1,94 @@
...
+
+Invalid Lisp may be rejected, complete with bad errors to show to end users:
+
+```rust
+use std::greenspun::{li, evaluate};
+
+let program = li!(cons 1. 2. 3.);
+
+assert_eq!(evaluate(program), String::from("too_many_PARAMS"));
+```
```
RDambrosio016, line 48: I think this should instead yield NaN, because of course an error is not a number, therefore NaN. Errors are too predictable.

chorman0773:
Hmm... I'm not sure about this. I have a WIP rust implementation in [lccc](https://github.com/LightningCreations/lccc), and I was under the impression I was going to be implementing Rust.

RDambrosio016:
> Hmm... I'm not sure about this. I have a WIP rust implementation in [lccc](https://github.com/LightningCreations/lccc), and I was under the impression I was going to be implementing Rust.

just wait until we introduce the `js!()` RFC

Lokathor:
Don't forget that Rust is only a subset of Lisp to begin with, so really this is just closing the parentheses on that)

xFrednet:
You forgot to write about the benefits of adding `NULL` as a value. Then we can represent three states with the `Option<>` type not just the two we have right now.

RDambrosio016:
> You forgot to write about the benefits of adding `NULL` as a value. Then we can represent three states with the `Option<>` type not just the two we have right now.

We will need even more once we add `std::quantum` though:
```rs
enum Option<T> {
  None,
  Some(T),
  BothSomeAndNone(T),
  Null,
  BothNullAndNone,
  BothNullAndSome(T)
}
```

text/0000-lisp-impl.md
mbartlett21: That would also mean that we don't have to use a result. We can instead always yield LispVal

mbartlett21:
> > You forgot to write about the benefits of adding `NULL` as a value. Then we can represent three states with the `Option<>` type not just the two we have right now.
> 
> We will need even more once we add `std::quantum` though:
> 
> ```rust
> enum Option<T> {
>   None,
>   Some(T),
>   BothSomeAndNone(T),
>   Null,
>   BothNullAndNone,
>   BothNullAndSome(T)
> }
> ```

You forgot `AllOfNullSomeAndNone(T)`

text/0000-lisp-impl.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Summary
+[summary]: #summary
+
+This RFC extends Greenspun's tenth rule to cover the Rust Standard Library.
+
+# Motivation
+[motivation]: #motivation
+
+Greenspun's tenth rule, [in its original form,][1] states that it only applies to `C` or `Fortran` programs. However, there is an opportunity for the rule to be expanded, by including `Rust` under its umbrella. The expected outcome is that there will be a documented interface to use the interpreter (with bugs and slowness included).
```
fstirlitz, line 14:
```suggestion
Greenspun's tenth rule, [in its original form,][1] states that it only applies to C or Fortran programs. However, there is an opportunity for the rule to be expanded, by including Rust under its umbrella. The expected outcome is that there will be a documented interface to use the interpreter (with bugs and slowness included).
```

PR #3252, opened 2022-04-04
add missing code syntax highlight to RFC 2005

SteveLauC:
First 6 code snippets in [text/2005-match-ergonomics.md](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md) don't specify programming language, I added  so that we have rust syntax highlight.


ehuss (approved):
Thanks!

PR #3253, opened 2022-04-04
Fix link for prior-art in RFC3137

slanterns:
Currently the link is interpreted as https://github.com/rust-lang/rfcs/blob/master/text/prior-art or https://rust-lang.github.io/rfcs/prior-art. This changes it to the right location.


ehuss (approved):
Thanks!

PR #3256, opened 2022-04-18
Fix broken links in RFC 1884

intgr:
These pages have gone 404, but seem worth preserving. Linked to archive.org instead.

Apparently I'm not the first one to notice this disappearance: https://www.reddit.com/r/rust/comments/senamw/what_happened_to_stjepang/


ehuss (approved):
Thanks!

PR #3294, opened 2022-07-21
Fix some issue links

ehuss:
Some of these links weren't updated when the RFCs were merged.

r? @pnkfelix  maybe? 😄


scottmcm:
Thanks!  These all look good to me.

PR #3314, opened 2022-09-09
Fix order of operations typo in RFC 2514

ssbr:
It's minor, but also a surprisingly confusing typo if you're new to this feature.


ehuss (approved):
Thanks!

PR #3317, opened 2022-09-18
RFC: Add `String::leak`

finnbear:
[Rendered](https://github.com/finnbear/rfcs/blob/string_leak/text/3317-string-leak.md)

### DIFF
--- a/text/0000-string-leak.md
+++ b/text/0000-string-leak.md
@@ -1,0 +1,71 @@
+- Feature Name: string_leak
+- Start Date: (2022-09-18)
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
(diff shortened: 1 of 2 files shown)


thomcc:
The right process for this was likely an ACP rather than an RFC, see https://std-dev-guide.rust-lang.org/feature-lifecycle/api-change-proposals.html for more info. While going through an RFC is possible, it's probably overkill for something like this. You probably should file one here: https://github.com/rust-lang/libs-team/issues/new?assignees=&labels=api-change-proposal%2C+T-libs-api&template=api-change-proposal.md&title=%28My+API+Change+Proposal%29 (unless going through the full RFC process sounds better to you...)

finnbear:
> The right process for this was likely an ACP rather than an RFC

Oh, alright, thanks for letting me know. I'll go submit an ACP :)

PR #3322, opened 2022-09-30
prominent link to RFC book

RalfJung:
I always have to search for the RFC book, usually with a search engine.

I now realize it is also set as homepage for this GitHub repository, but GH hides the homepage link in the sidebar so it is really easy to miss. The "Active RFC List" is several times more visible, but when I want to look for existing ratified RFCs it does not help...


ehuss (approved):
Seems reasonable to me, thanks!

PR #3332, opened 2022-10-20
Fix a typo in RFC 3027

albertlarsan68:
[Rendered](https://github.com/albertlarsan68/rfcs/blob/patch-1/text/3027-infallible-promotion.md)


ehuss (approved):
Thanks!

PR #3353, opened 2022-12-07
Start RFC for doc cfg

GuillaumeGomez:
[Rendered](https://github.com/GuillaumeGomez/rfcs/blob/doc-cfg/text/000-rustdoc-cfgs-handling.md)


GuillaumeGomez:
Sorry, didn't mean to open the PR upstream as it was a very very early start for a future RFC I'd like to propose.

PR #3354, opened 2022-12-07
Cross-type number comparisons

kangalioo:
[Rendered](https://github.com/kangalioo/rfcs/blob/cross-type-number-comparison/text/0000-cross-type-number-comparison.md)

### DIFF
--- a/text/0000-cross-type-number-comparison.md
+++ b/text/0000-cross-type-number-comparison.md
@@ -1,0 +1,78 @@
+- Feature Name: `cross_type_number_comparison`
+- Start Date: 2022-12-07
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Implement `PartialEq` and `PartialOrd` for all combinations of `u8`, `u16`, `u32`, `u64`, `u128`, `usize`, `i8`, `i16`, `i32`, `i64`, `i128`, `isize`, `f32`, `f64`; as well as `Eq` and `Ord` for the integer combinations.
+
+# Motivation
+[motivation]: #motivation
+
+Comparing numbers of different types, whether done deliberately or accidentally, currently results in a compilation error. This means, users either have to cascadingly change numbers to be of the same type; or, hand-roll cross-type comparisons.
+
+This is an annoyance and even a point of confusion for newer programmers. Additionally, hand-rolled comparisons are prone to subtle mistakes regarding unsigned/signed integers or float rounding. Also, achieving optimal performance is non-trivial and even more prone to mistakes.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Makes it possible to compare numbers of different types to each other, by adding `PartialOrd`/`Ord` and `PartialEq`/`Eq` trait implementations for different number types:
+- `3 < 3.5` is `true`
+- `200_u8 > -1_i8` is `true`
+- `10000000000000001_i64 == 10000000000000000_f64` is `false`
+- ...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+There will be macros to generate the trait implementations for float<->int comparison, int<->int comparison, and float<->float comparison. The actual implementation can be adapted from @orlp's [num-ord crate](https://github.com/orlp/num-ord).
+
+Comparing any integer to a float NaN returns `false` (`None` in `partial_cmp()`)
+
+In a comparison, floats are compared by the exact mathematical value they represent according to IEEE-754. For example:
+- `19999999_f32 == 20000000` because `19999999` is rounded to `20000000` in 32 bit floats
+- `20000000_f32 != 20000001` because both literals represent their value exactly, and the values are different
+- `19999999_f64 != 20000000` because f64 has enough precision to exactly represent `19999999`
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Cross-type comparisons are potentially expensive
+    - Unsigned<->signed integer comparisons are barely slower than native same-type comparisons: 5.1ns vs 5.0ns on my machine [playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=5a8704dfd953be0d0e750b2d916ccc7f)
+    - However, float<->int comparisons remain to be benchmarked and optimized
+- Cross-type comparisons may hint at a design problem in the code
+    - Quite subjective, and better solved with a lint
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Restructuring the code to use the same number type everywhere
+
+This is not always possible or desirable.
+
+## Using a third-party library like [num-ord](https://github.com/orlp/num-ord)
+
+Requires wrapping number types in the third-party newtype which is ergonomic and not possible in some situations.
+
+## Continue to hand-roll cross-type comparisons when they come up
+
+As explained above, this is easy to get wrong. Also naive implementations have suboptimal performance.
+
+# Prior art
+[prior-art]: #prior-art
+
+@orlp's [num-ord crate](https://github.com/orlp/num-ord).
+
+@lifthrasiir's [num-cmp crate](https://github.com/lifthrasiir/num-cmp) (abandoned).
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is the performance difference between float<->int and float<->float comparisons acceptable?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+- A lint against cross-type comparisons, probably in `clippy::restriction`.
--- a/text/0000-cross-type-number-comparison.md
+++ b/text/0000-cross-type-number-comparison.md
@@ -33,8 +33,12 @@ Comparing any integer to a float NaN returns `false` (`None` in `partial_cmp()`)
 
 In a comparison, floats are compared by the exact mathematical value they represent according to IEEE-754. For example:
 - `19999999_f32 == 20000000` because `19999999` is rounded to `20000000` in 32 bit floats
+    - In other words, the number `19999999` _does not exist_ in 32 bit floats
 - `20000000_f32 != 20000001` because both literals represent their value exactly, and the values are different
 - `19999999_f64 != 20000000` because f64 has enough precision to exactly represent `19999999`
+
+Type inference shouldn't be affected by this change, because the generic parameter on `PartialOrd` and `PartialEq` is defaulted to `Self`.
+Therefore, expressions like `1 == 1_u8` will continue to work by correctly inferring the first literal to be a `u8`.
 
 # Drawbacks
 [drawbacks]: #drawbacks


afetisov:
Besides the listed downsides, it will negatively affect type type inference. Currently code like
```rust
let x = 2_u16;
dbg!(x > 5);
```
compiles, because there is a single possible impl  for comparison with `u16`. If comparisons with other types are added, this code will fail to compile, since there will no longer be a default impl to fall back to.

It's also a correctness footgun. If I understand the proposition correctly, `1999999999_f32 == 2000000000` should be true, but that goes against reasonable expectations. It's an extra footgun if the result of comparison would depend on the exact integer and floating point types.

Overall, writing the comparison explicitly is simple enough and shouldn't be common enough to warrant those issues.

mcclure:
I want to note I think there's a slight difference between comparing *integer* types and comparing *between integer and floating point types*. Including float<->integer comparisons introduces some things that you could potentially view as a problem:

1. Slightly larger performance than you might expect for a comparison, with multiple branches
2. Unintuitive (non-integer-logic) behavior when the float is a NaN
3. The 1999999999_f32 == 2000000000 issue described above

I am not sure any of these are an argument *not to do the thing*, particularly because all of them apply in the *existing* case of float-to-float comparisons: float can be expensive on softfpu systems; float-to-float comparison does something that programmers often misunderstand, leading to bugs, if one float is NaN; Currently Rust evaluates "true" for the expression 1999999999_f32 == 2000000000_f32. But at least you can form the argument, whereas I don't think any of these apply in the case of integer-to-integer comparisons (unless a single branch in the case of a signed-to-unsigned comparison is a performance concern, which seems really tenuous).

thomcc:
I think the inference issues make this a non-starter. I'd expect this to break all the rust code where a variable of numeric type is compared against a number literal (for example, `a == 0`). We had a similar (but less extreme) issue in std::simd, and it broke a ton of code (even things that didn't care at all about simd).

> Type inference shouldn't be affected by this change, because the generic parameter on PartialOrd and PartialEq is defaulted to Self. Therefore, expressions like 1 == 1_u8 will continue to work by correctly inferring the first literal to be a u8.

Sadly, I don't believe this is how type inference works, at least not at the moment (and my understanding is that	extending it to handle defaults the way some folks expect requires some nontrivial design work, at the very least).

afetisov:
@mcclure How would you compare `i64::MIN` and `u64::MAX` without extra branching? You can't cast them to a larger native type.

I think the only comparisons which are more or less reasonable are comparison of same-sign different-width integers, e.g. `i16 vs isize` or `u8 vs u64`. But the inference issue remains, and it's unclear whether it helps that much in practice.

What would help imho is if `1_u8.into() == 1_u32` could compile, rather than throw a type inference error. It's a safe, unambiguous and correct way to perform such comparisons, but it can't compile since the type of the LHS can't be inferred (formally it could map into any `T` which has a `PartialEq<u32>` impl).

@scottmcm That inference error is quite similar to the one plaguing `try` blocks. Do you think there could be some change of inference which solves them both? Feels like there could be some "choose the minimal solution" logic possible.

text/0000-cross-type-number-comparison.md
```diff
@@ -0,0 +1,82 @@
...
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Restructuring the code to use the same number type everywhere
+
+This is not always possible or desirable.
+
+## Using a third-party library like [num-ord](https://github.com/orlp/num-ord)
+
+Requires wrapping number types in the third-party newtype which is ergonomic and not possible in some situations.
```
SOF3, line 61: did you mean *unergonomic* instead?

text/0000-cross-type-number-comparison.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Continue to hand-roll cross-type comparisons when they come up
+
+As explained above, this is easy to get wrong. Also naive implementations have suboptimal performance.
```
SOF3, line 65: If performance is a concern, is there any reason why the user can't just use num-ord?

text/0000-cross-type-number-comparison.md
SOF3: what are the "not possible" situations? could you elaborate on that?

mcclure:
> @mcclure How would you compare `i64::MIN` and `u64::MAX` without extra branching?

I would use branching.

petrochenkov:
>cascadingly change numbers to be of the same type

And that's the right thing to do, in most cases.

>or, hand-roll cross-type comparisons

Not, if cross-type comparisons are provided as regular non-operator functions by the standard library, for the remaining cases where it's really necessary.

burdges:
It's really enough that rustc gives helpful error messages imho, likely already the case.  

There are some nasty footguns if Rust introduces additional branches inside simple equality comparisons.  It makes auditing code for being constant time far harder for example.  

We already have `From`s that makes conversions fairly easy.  In fact, this might breaks code like `x.into() == y`.  It's be particularly awful if people convert better code like `x.into() == y` into worse ambiguous code like `x == y` just to silence bad new error.

kangalioo:
Thanks everyone for the thoughts. I compiled the arguments mentioned in the comments:

1. type interference will be broken
2. correctness footgun (`1999999999_f32 == 2000000000`)
    - arguably not a problem, because this is an intrinsic property to floats: `1999999999_f32 == `2000000000_f32` is already true in current Rust
3. unexpectedly expensive operation for a comparison / auditing code for constant time is harder
    - remains to be benchmarked (there may be clever algorithms that are very fast / constant time)
        - for int<->uint comparisons, the naive impl is already optimized to constant-time, almost zero-overhead x86 assembly
        - int<->int, uint<->uint and float<->float comparisons only require upcasting, which is constant-time and fast too
        - only int/uint<->float is tricky
            - a subset of this proposal which doesn't include int/uint<->float comparisons could be posted
    - these problems already apply to soft-fpu configurations anyways
4. cross-type comparisons can be provided as regular non-operator functions by the standard library
5. "writing the comparison explicitly is simple enough" / "We already have `From`s that makes conversions fairly easy"
    - this is a deceiving attitude because many comparison routines are unexpectedly non-trivial. I guess should have expanded on this in the RFC:
    - e.g. `iN` vs `uN`:
        - casting both to the largest signed integer doesn't work for `i128` vs `u128` and has suboptimal performance
        - casting `uN` to `iN` or `iN` to `uN` will wrap-around half the values, which is a subtle bug
    - e.g. float vs int:
        - casting float to int before comparing breaks for large floats that exceed int range: `1e30_f32 as u64 > u64::MAX` wrongly returns false
        - casting int to float unnecessarily rounds the int and causes wrong results: `(2_u64.pow(6) + 1) as f64 > 2_f64.powi(6)` wrongly returns false
        - even if you have a correct solution; making it fast is a can of worms in itself
    - the _correct_ solution requires in-depth knowledge of integer overflow and float rounding

As thomcc mentioned, the type interference issue make the RFC in its current form a non-starter. However, the issue remains that cross-type comparisons are deceptively simple, making people reach for incorrect ad-hoc implementations instead of depending on [num-ord](https://docs.rs/num-ord), thereby causing subtle correctness bugs.

Therefore, support for cross-type number comparisons should be "provided as regular non-operator functions by the standard library". Some ideas:
- `.comparable()`/`.cmp()`/`.orderable()`/`.ord()` methods on numbers which returns a struct that can be compared against any other number
- `.compare_to()` method on numbers which returns `std::cmp::Ordering`
- new comparison trait

Should that new proposal happen in a new RFC, or ACP, or pre-RFC discussion on IRLO, or somewhere else?

afetisov:
Correct integer comparisons for anything fitting in 64 bits (which is almost always the case in practice) and without specific low-level requirements are easy to implement. Just cast everything to `i128`: `i128::from(x) > i128::from(y)`. It's even likely that LLVM will output the optimal comparison anyway. `i128` and `u128` aren't that easy to compare, but their main use is precisely easy correct implementation of arithmetics on smaller integers.

For float to int comparison, it doesn't even make sense to say that one specific comparison is "correct". There are many more or less valid ways to compare, with no universal choice. If we round the float, do we round up, down, to zero or somehow else? If we cast an integer to float, how do we deal with the loss of precision? How should we represent a float if no exact representation is possible?

Various choices are possible, but the most important thing is to be explicit about the choice you make. For that reason I would be strongly against adding float-to-int comparison under any kind of generic nondescript function name. Certainly not a `PartialEq` impl, but a function name like `compare_to` is just as bad.

What is the use case where float-to-int comparisons would be common enough to warrant such ergonomic features?

kangalioo:
`x.into() == y` doesn't work when one type is not a strict subset of the other

kangalioo:
> Correct integer comparisons for anything fitting in 64 bits (which is almost always the case in practice) and without specific low-level requirements are easy to implement. Just cast everything to i128: i128::from(x) > i128::from(y). It's even likely that LLVM will output the optimal comparison anyway. i128 and u128 aren't that easy to compare, but their main use is precisely easy correct implementation of arithmetics on smaller integers.

True, this is a simple solution, but: LLVM _does not_ output the optimal comparison (5.1ns vs 5.4ns on my machine, [see benchmark from RFC](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=5a8704dfd953be0d0e750b2d916ccc7f)), and why should we not support comparing `i128` with `u128`?

> For float to int comparison, it doesn't even make sense to say that one specific comparison is "correct". There are many more or less valid ways to compare, with no universal choice. If we round the float, do we round up, down, to zero or somehow else? If we cast an integer to float, how do we deal with the loss of precision? How should we represent a float if no exact representation is possible?

We use the IEEE-754 specified exact mathematical value of the float and compare it against the mathematical value of the integer. Any rounding or casting of numbers is merely an implementation detail and does not affect semantics

SOF3:
`1999999999_f32 == 2000000000_f32` is acceptable because float comparison clear indicates that it is not a full equivalence reason. But when I write `something == int` or `int == something`, whether this is a full equivalence would depend on the type of `something`, which is arguably confusing. (Of course, you can already implement a `PartialEq<i32>`, but it is probably confusing to users anyway)

text/0000-cross-type-number-comparison.md
kangalioo: Good point, I'll address the "not possible" situations in a potential future proposal which proposes ordering functions instead of trait implementations

kangalioo:
> But when I write something == int or int == something, whether this is a full equivalence would depend on the type of something, which is arguably confusing

Fair.

In any case, this wouldn't be a problem with explicit `.comparable()`/`.cmp()`/`.orderable()`/`.ord()` methods on numbers, since any gotchas could be extensively documented on the methods. To get back to this, what would be a better place to discuss this new idea (explicit ordering methods)? A pre-RFC discussion on IRLO, or a new ACP or RFC?

afetisov:
@kangalioo 

Don't forget the most important question:

> What is the use case where float-to-int comparisons would be common enough to warrant such ergonomic features?

If you're going to propose something that controversial, point №1 is to gather the evidence that there is indeed a real significant problem with downsides outweighing the ones in your proposal. You need to show some real-world projects where it's an issue, and not because they just messed up their type design.

I don't think this proposal can fly without such real-world data.

kangalioo:
> What is the use case where float-to-int comparisons would be common enough to warrant such ergonomic features?

Plenty can be found in grep.app by searching for e.g. [`as f64 >`](https://grep.app/search?q=as%20f64%20%3E&filter[lang][0]=Rust) (try `f32` or `<`/`==` for more results).

These projects implement float<->int comparisons as utility functions and get it wrong:

Screenshot_20221209_184552

This GUI code seems to compare mouse pointer coordinates (f64) to a known point on the screen (integer) and gets it wrong (although GUI projects don't have numbers in the range where float inaccuracies begin to show):

Screenshot_20221209_185023

Next instance, although float inaccuracies won't show up here either because percentages are max 100:

Screenshot_20221209_185455

This code checks if a point lies within a circle and is subject to float inaccuracies at coordinates beyond 94906265 (square root of dense float range end):

Screenshot_20221209_191729

This code compares DNA-related usize's to a f64 threshold and is subject to float inaccuracies beyond 9007199254740992:

Screenshot_20221209_201534

PR #3356, opened 2022-12-13
Update link to blog post

Jeroendevr:
Website is not accessible on the GitHub.io domain. So change to the current domain.

[Rendered](https://github.com/Jeroendevr/rfcs/blob/patch-1/text/1857-stabilize-drop-order.md)


ehuss (approved):
Thanks!

PR #3362, opened 2023-01-03
Command improvements for ergonomics and error handling

ijackson:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3362)*

[Rendered](https://github.com/ijackson/rust-rfcs/blob/command/text/0000-command-ergonomics.md)

### DIFF
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -239,6 +239,8 @@ and the exit status as problems within the `SubprocessError`.
 ///
 ///  * If you want to capture stderr but combine it with stdout,
 ///    TODO need way to combine them!
+///    
+///  * If you wish to tolerate only nonzero exit status, call `.just_status()`.
 ///
 ///  * If you wish to tolerate other particular kind(s) of problem,
 ///    set the field for the problems you want to tolerate to `None`
@@ -342,6 +344,13 @@ impl SubprocessError {
     // and a caller can't reimplement it without making assumptons
     // about `SubprocessError`'s contents.
     fn has_problem(&self) -> bool;
+
+    /// Returns `Ok<ExitStatus>` if the only reason for the failure was a nonzero exit status.
+    /// Otherwise returns `self`.
+    ////
+    /// Use this if you want to to tolerate some exit statuses,
+    /// but still fail if there were other problems.
+    pub fn just_status(self) -> Result<ExitStatus, SubprocessError>;
 }
 impl Default for SubprocessError { ... }
 impl Clone for SubprocessError { ... } // contained io:Errors are in Arcs
@@ -468,6 +477,21 @@ Options are:
 
 Here we propose option 1: treat as `inherit`.
 
+# Examples
+
+## Conveniently running `diff`
+
+```rust
+    let result = Command::new("diff")
+        .args(["before","after"])
+        .stderr(Stdio::piped()) // optional, could just let it inherit
+        .run();
+    let status = match result {
+        Ok(()) => 0,
+        Err(err) => err.just_status()?.code(),
+    };
+```
+
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
@@ -512,65 +536,3 @@ Options might include:
    [`read2`](https://github.com/rust-lang/cargo/blob/58a961314437258065e23cb6316dfc121d96fb71/crates/cargo-util/src/read2.rs)
 
  * Expect users who want this to use pipes by hand (perhaps with threads), or async.
-
-## More convenient way to run `diff`
-
-With the proposed API,
-completely correctly running `diff(1)` would look a bit like this:
-
-```rust
-    let result = Command::new("diff")
-        .args(["before","after"])
-        .run();
-    let status = match result {
-        Ok(()) => 0,
-        Err(err) => {
-            let status = err.status();
-            err.set_status(ExitStatusExt::from_raw(0));
-            if err.has_problem() {
-                return Err(err);
-            }
-            status.code()
-        }
-    };
-```
-
-This is doable but cumbersome.
-A naive Dionysus is likely to write:
-
-```rust
-    let status = match result {
-        Ok(()) => 0,
-        Err(err) => {
-            if ! err.status().success() {
-                err.status().code()
-            } else {
-                return Err(err);
-            }
-        }
-    };
-```
-
-As it happens, this is correct in the sense that it won't malfunction,
-since actually `run()`, without piped stderr,
-cannot produce a `SubprocessError`
-containing a nonzero exit status *and* any other problem.
-But in a more complex situation it might be wrong.
-
-Perhaps:
-```rust
-impl SubprocessError {
-    /// Returns `Ok<ExitStatus>` if the only reason for the failure was a nonzero exit status.  Otherwise returns `self`.
-    ////
-    /// Use this if you want to to tolerate some exit statuses, but still fail if there were other problems.
-    pub fn just_status(self) -> Result<ExitStatus, SubprocessError>;
-}
-```
-
-Then Dionysus can write:
-```rust
-    let status = match result {
-        Ok(()) => 0,
-        Err(err) => err.just_status()?.code(),
-    };
-```
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -135,7 +135,7 @@ We aim to serve well each of the following people:
    After the child indicates EOF on its stdout,
    we will wait for it to finish and check the exit status.
 
- * `fn get_output(&mut self) -> Result<Vec<String>, SubprocessError>`:
+ * `fn get_output(&mut self) -> Result<String, SubprocessError>`:
 
    Runs the command and collects its stdout.
    Decodes the stdout as UTF-8, and fails if that's not possible.
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -88,7 +88,7 @@ impl Command {
     fn read_stdout_bytes(&mut self) -> Result<Vec<u8>, ProcessError>;
     fn read_stdout(&mut self) -> Result<String, ProcessError>;
     fn read_stdout_line(&mut self) -> Result<String, ProcessError>;
-    fn stdout_readable(&mut self) -> impl std::io::Read;
+    fn stdout_reader(&mut self) -> impl std::io::Read;
 }
 struct ProcessError { ... }
 impl From<ProcessError> for io::Error { ... }
@@ -150,13 +150,13 @@ We aim to serve well each of the following people:
    If program prints no output at all, returns an empty string
    (and this cannot be distinguished from the program printing just a newline).
 
- * `fn read_stdout_read(&mut self) -> std::process::ChildOutputStream`
-   (where `struct ChildOutputStream` implements `io::Read`
+ * `fn stdout_reader(&mut self) -> std::process::ChildOutputReader`
+   (where `struct ChildOutputReader` implements `io::Read`
    and is `Send + Sync + 'static`).
 
    Starts the command, allowing the caller to
    read the stdout in a streaming way.
-   Neither EOF nor an error will be reported by `ChildOutputStream`
+   Neither EOF nor an error will be reported by `ChildOutputReader`
    until the child has exited, *and* the stdout pipe reports EOF.
    (This includes errors due to nonempty stderr,
    if stderr was set to `piped`.)
@@ -306,7 +306,7 @@ a resulting error message unambiguous.
 ### `impl From<ProcessError> for io::Error`
 
 `ProcessError` must be convertible to `io::Error`
-so that we can use it in `ChildOutputStream`'s
+so that we can use it in `ChildOutputReader`'s
 `Read` implementation.
 This may also be convenient elsewhere.
 
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -434,6 +434,12 @@ for process invocation and output handling.
 Perhaps printing the command arguments is overly verbose,
 and we should print only the command name.
 
+Sometimes people pass sensitive information (such ass passwords) in command line arguments.
+This is not a good idea, because command line arguments are generally public on Unix.
+
+Perhaps some option could be added in the future to control this.
+For now, we propose always printing the arguments.
+
 ## `cause` in `Error` impl of `ProcessError`
 
 Because maybe several things went wrong, ever providing a `Some` `cause`
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -360,6 +360,23 @@ Alternatives and prior proposals include:
    but got bogged down due to lack of consensus on overall direction
    and some bikeshed issues.
 
+ * The `SubprocessError` type could be a
+   transparent non-exhaustive struct.
+   This might be reasonable,
+   since it's really just a bag of information
+   with getters and setters for every field.
+   But transparent structs are unfashionable in modern Rust.
+
+ * Instead of a single `SubprocessError` type used everywhere,
+   there could be a different error type for the different calls.
+   For example, `run()` could have a different error type to
+   `get_output()`,
+   since `run` doesn't need to represent UTF-8 conversion errors.
+   This would not be very in keeping with the rest of `std::process`,
+   which tends to unified types with variation selected at runtime.
+   There would still have to be *a* type as complex as `SubprocessError`,
+   since that's what `get_output_read` needs.
+
 # Prior art
 [prior-art]: #prior-art
 
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -57,17 +57,17 @@ The following incorrect program fragments are all accepted today and run to comp
 
 ```rust
     Command::new("touch")
-        .args(&["/dev/enoent/touch-1"]);
+        .args(&["/nonexistent/touch-1"]);
     // ^ programmer surely wanted to actually run the command
 
     Command::new("touch")
-        .args(&["/dev/enoent/touch-2"])
+        .args(&["/nonexistent/touch-2"])
         .spawn()?;
     // ^ accidentally failed to wait, if programmer wanted to make a daemon
     //   or zombie or something they should have to write let _ =.
 
     Command::new("touch")
-        .args(&["/dev/enoent/touch-3"])
+        .args(&["/nonexistent/touch-3"])
         .spawn()?
         .wait()?;
     // ^ accidentally failed to check exit status
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -85,8 +85,8 @@ for running the command and collecting its output:
 ```rust
 impl Command {
     fn run(&mut self) -> Result<(), ProcessError>;
-    fn read_stdout_bytes(&mut self) -> Result<Vec<u8>, ProcessError>;
-    fn read_stdout(&mut self) -> Result<String, ProcessError>;
+    fn read_stdout(&mut self) -> Result<Vec<u8>, ProcessError>;
+    fn read_stdout_string(&mut self) -> Result<String, ProcessError>;
     fn read_stdout_line(&mut self) -> Result<String, ProcessError>;
     fn stdout_reader(&mut self) -> impl std::io::Read;
 }
@@ -126,21 +126,21 @@ We aim to serve well each of the following people:
    Equivalent to `.spawn()` followed by `.status()`,
    but with better error handling.
 
- * `fn read_stdout_bytes(&mut self) -> Result<Vec<u8>, ProcessError>`:
+ * `fn read_stdout(&mut self) -> Result<Vec<u8>, ProcessError>`:
 
    Runs the command and collects its stdout.
    After the child indicates EOF on its stdout,
    we will wait for it to finish and check the exit status.
 
- * `fn read_stdout(&mut self) -> Result<String, ProcessError>`:
+ * `fn read_stdout_string(&mut self) -> Result<String, ProcessError>`:
 
-   Runs the command and collects its stdout, as for `read_stdout_bytes`.
+   Runs the command and collects its stdout, as for `read_stdout`.
    Decodes the stdout as UTF-8, and fails if that's not possible.
    Does not trim any trailing line ending.
 
  * `fn read_stdout_line(&mut self) -> Result<String, ProcessError>`:
 
-   Runs the command and collects its stdout, as for `read_stdout_bytes`.
+   Runs the command and collects its stdout, as for `read_stdout`.
    Decodes the stdout as UTF-8, and fails if that's not possible.
    Fails unless the output is a single line (with or without line ending).
    Trims the line ending (if any).
@@ -367,7 +367,7 @@ Perhaps we don't need all the `read_stdout` variants,
 and could require Bob to write out the boilerplate
 or provide his own helper function.
 
-Perhaps we don't need `read_stdout_read`.
+Perhaps we don't need `stdout_reader`.
 However,
 avoiding deadlocks when reading subprocess output,
 and also doing error checks properly,
@@ -467,17 +467,22 @@ This leaves us with the following options:
 
 We have `fs::read_to_string`.
 
-Possible names (taking `read_stdout_bytes` as the example):
+Possible names (taking `read_stdout_string` as the example):
 
- * `read_stdout_bytes` (proposed in this RFC)
- * `stdout_bytes` but `stdout` is alread taken.
- * `get_stdout_bytes`
- * `run_get_stdout_bytes`
- * `run_stdout_bytes`
+ * `read_stdout_string` (proposed in this RFC)
+ * `stdout_string` but `stdout` is alread taken.
+ * `get_stdout_string`
+ * `run_get_stdout_string`
+ * `run_stdout_string`
 
 It is difficult to convey everything that is needed in a short name.
 In particular, all of these functions spawn the program,
 and wait (at an appropriate point) for it to exit.
+
+Should `read_output` be the one that returns `Vec<u8>`
+or the one that returns `Vec<String>` ?
+Precedent in the stdlib is usually to have the bytes version undecorated
+(eg, `fs::read_to_string`).
 
 ## stderr handling default
 
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -301,11 +301,12 @@ The `io::ErrorKind` for a `ProcessError` will be:
   * Otherwise, a new kind `io::ErrorKind::ProcessFailed`,
     which means that the subprocess itself failed.
 
-### Further necessary APIs for `ProcessError`
+### Further APIs for `ProcessError`
 
-We also provide ways for this new error to be constructed,
-which will be needed by other lower level libraries besides std,
-notably async frameworks:
+A `ProcessError` is a transparent `Default` struct so it can be
+constructed outside std, for example by async frameworks.
+
+We propose the following additional methods:
 
 ```rust
 impl ProcessError {
@@ -313,6 +314,9 @@ impl ProcessError {
     ///
     /// `has_problem()` will return `false` until one of the setters
     /// is used to store an actual problem.
+    //
+    // This is a name for the `Default` impl, and not essential,
+    // although it's conventional in Rust to provide it.
     fn new_empty() -> Self { }
 
     // If we keep ExitStatusError
@@ -336,6 +340,10 @@ impl ProcessError {
     ////
     /// Use this if you want to to tolerate some exit statuses,
     /// but still fail if there were other problems.
+    //
+    // This is optional, and could be a separate feature from the rest of the RFC.
+    // But it does make running programs like `diff` considerably easier.
+    // (It is also implementable externally in terms of .has_problem()`.)
     pub fn just_status(self) -> Result<ExitStatus, ProcessError>;
 }
 impl Default for ProcessError { ... }
--- a/text/0000-command-ergonomics.md
+++ b/text/0000-command-ergonomics.md
@@ -84,14 +84,14 @@ for running the command and collecting its output:
 
 ```rust
 impl Command {
-    fn run(&mut self) -> Result<(), SubprocessError>;
-    fn read_stdout_bytes(&mut self) -> Result<Vec<u8>, SubprocessError>;
-    fn read_stdout(&mut self) -> Result<String, SubprocessError>;
-    fn read_stdout_line(&mut self) -> Result<String, SubprocessError>;
+    fn run(&mut self) -> Result<(), ProcessError>;
+    fn read_stdout_bytes(&mut self) -> Result<Vec<u8>, ProcessError>;
+    fn read_stdout(&mut self) -> Result<String, ProcessError>;
+    fn read_stdout_line(&mut self) -> Result<String, ProcessError>;
     fn stdout_readable(&mut self) -> impl std::io::Read;
 }
-struct SubprocessError { ... }
-impl From<SubprocessError> for io::Error { ... }
+struct ProcessError { ... }
+impl From<ProcessError> for io::Error { ... }
 ```
 
 The `.output()` function and `std::process::Output`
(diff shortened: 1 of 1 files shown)


text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,525 @@
...
+ * [#73126](https://github.com/rust-lang/rust/issues/73126)
+   issue: `std::process::Command` `output()` method error handling hazards
+ * [#73131](https://github.com/rust-lang/rust/issues/73131)
+   Overall tracking Issue for `std::process` error handling
+
+## Currently-accepted wrong programs
+
+The following incorrect program fragments are all accepted today and run to completion without returning any error:
+
+```
```
lebensterben, line 58:
```suggestion
```rust
```

Nilstrieb:
I agree with the motivation that `Command` is not very ergonomic, especially around the common "return error containing stderr on error" workflow.
But this RFC is a little too opinionated about `stderr` in my opinion. It strictly treats it as the "error output", which is just often not true in practice. Even `cargo` logs all its output to stderr. Therefore, considering any output to stderr an error is a bad idea.
Also, if I read this correctly, there's no way to get stdout when the command failed using these methods except with the existing `.output()`, but this plans to deprecate that.

ijackson:
> I agree with the motivation that `Command` is not very ergonomic, especially around the common "return error containing stderr on error" workflow. But this RFC is a little too opinionated about `stderr` in my opinion. It strictly treats it as the "error output",

Um?  That's not true.  In my RFC, by default, error output is sent to the Rust program's stderr (inherit).

> which is just often not true in practice. Even `cargo` logs all its output to stderr. Therefore, considering any output to stderr an error is a bad idea.

I think this depends what program you're running.  I don't think it is wrong to have the *option* to treat stderr output as an error, as a *non-default* mode.

> Also, if I read this correctly, there's no way to get stdout when the command failed using these methods except with the existing `.output()`, but this plans to deprecate that.

No, that's not true either.  If you have a `SubprocessError`, you can call the `.stdout_bytes()` method on it.  (It will only be `Some` if you called one of the `.get_output*` methods, obviously; otherwise the stdout went to wherever you configured the `Command` to send it.)

Nilstrieb:
Ah, I seem to have overlooked that, sorry.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+ * [#73131](https://github.com/rust-lang/rust/issues/73131)
+   Overall tracking Issue for `std::process` error handling
+
+## Currently-accepted wrong programs
+
+The following incorrect program fragments are all accepted today and run to completion without returning any error:
+
+```rust
+    Command::new("touch")
+        .args(&["/dev/enoent/touch-1"]);
```
joshtriplett, line 60:
```suggestion
        .args(&["/nonexistent/touch-1"]);
```

(Rather than having to explain that `/dev/enoent` is a non-existent directory, and having people wonder if because it's under `/dev` it has some nonstandard semantics attached.)

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+    // ^ programmer surely wanted to actually run the command
+
+    Command::new("touch")
+        .args(&["/dev/enoent/touch-2"])
```
joshtriplett, line 64:
```suggestion
        .args(&["/nonexistent/touch-2"])
```

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+        .spawn()?;
+    // ^ accidentally failed to wait, if programmer wanted to make a daemon
+    //   or zombie or something they should have to write let _ =.
+
+    Command::new("touch")
+        .args(&["/dev/enoent/touch-3"])
```
joshtriplett, line 70:
```suggestion
        .args(&["/nonexistent/touch-3"])
```

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+   Equivalent to `.spawn()` followed by `.status()`,
+   but with better error handling.
+
+ * `fn get_output_bytes(&mut self) -> Result<Vec<u8>, SubprocessError>`:
+
+   Runs the command and collects its stdout.
+   After the child indicates EOF on its stdout,
+   we will wait for it to finish and check the exit status.
+
+ * `fn get_output(&mut self) -> Result<Vec<String>, SubprocessError>`:
```
joshtriplett, line 138:
```suggestion
 * `fn get_output(&mut self) -> Result<String, SubprocessError>`:
```

I'm not sure if this was a typo, or if you intended this to have built-in split-by-lines. But in any case, we have methods for splitting by lines, so I think this should return the whole output as one `String`.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+
+   Runs the command and collects its stdout.
+   Decodes the stdout as UTF-8, and fails if that's not possible.
+   Does not trim any trailing line ending.
+
+ * `fn get_output_line(&mut self) -> Result<Vec<String>, SubprocessError>`:
```
joshtriplett, line 144:
```suggestion
 * `fn get_output_line(&mut self) -> Result<String, SubprocessError>`:
```

This being `Vec<String>` as well is what makes me think the above is a typo, since otherwise, this Vec would always contain one element.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+
+   Runs the command and collects its stdout.
+   Decodes the stdout as UTF-8, and fails if that's not possible.
+   Fails unless the output is a single line (with or without line ending).
```
joshtriplett, line 148: Please specify the behavior if the program has *no* output. Does this fail in that case, or is the requirement "at most one line"?

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+recommended by docs for other, lower-level functions
+which people were previously required to use.
+
+## Deprecations
+
+ * Deprecate `std::process::Command::output()`.
+   This API cannot be fixed;
+   see [#73126](https://github.com/rust-lang/rust/issues/73126).
+
+ * Apply `#[must_use]` to `Command` and `Child`.
```
joshtriplett, line 174: :tada: :tada: :tada:

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+   if a called program later starts printing warning messages.
+
+ 3. Call this a programming mistake, and panic.
+
+ 4. Somehow make this a compile error.
+   This would involve a new `CommandForWhichWeHaveSpecifiedStderr`
+   type (typestate pattern), or providing the stderr handling as a mandatory argument
+   to all the new methods.
+   These seem unpalatably unergonomic.
+
```
joshtriplett, line 434:
```suggestion

5. One of the above, and additionally provide a lint that makes a best-effort attempt to catch this at compile time.

```

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+ * `fn get_output_read(&mut self) -> std::process::ChildOutputStream`
+   (where `struct ChildOutputStream` implements `io::Read`
+   and is `Send + Sync + 'static`).
+
+   Starts the command, allowing the caller to
+   read the stdout in a streaming way.
+   Neither EOF nor an error will be reported by `ChildOutputStream`
+   until the child has exited, *and* the stdout pipe reports EOF.
+   (This includes errors due to nonempty stderr,
+   if stderr was set to `piped`.)
```
joshtriplett, line 160: This description makes me wonder about the semantics of the *other* functions in this family. Do they wait for process exit, or for process exit *and* EOF? (I'm hoping the former, or possibly "process exit and then a non-blocking read saying it would block".)

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+
+ * Python3's `subprocess.run` fails to call nonzero exit status an exception. If you don't write explicit error handling code (highly unusual in Python) you have an unchecked exit status bug.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+## Printing command arguments in `impl Display for SubprocessError`
+
+Perhaps printing the command arguments is overly verbose,
+and we should print onliy the command name.
```
joshtriplett, line 398:
```suggestion
and we should print only the command name.
```

joshtriplett:
In a broad sense, I think the approach of this RFC seems mostly reasonable, and I'd like to see something along these lines.

There are two categories of issues I'd like to raise. I think both are addressable, and I'm hoping it'll be possible to address them and merge the resulting RFC.

One family of issues are *largely* bikeshed-painting, where the functionality seems perfectly fine but I think the names are unintuitive. For those, I want to be clear that I'm not attached to any *particular* names, and primarily want to express some desired properties of a name that I think the proposed names don't have. I'd like to align on those *properties* and then I don't especially care what name we pick that meets those properties.

The other family of issues are those of functionality, where I think the proposal makes a common use case more challenging.

-----

First, the bikesheds:

- Bikeshed: We currently have programs using a method called `output` that retrieves both stdout and stderr, merged. I don't think we can switch from that to methods named `get_output*` that *only* return stdout without leading to some confusion. I would propose a name that makes it clear that these only return stdout, such as `get_stdout`. (Arguably that *may* mean it needs to be called `read_stdout` or similar, because `get_stdout` might imply "get the previously set standard output device"; in that case, we could also have `stdout_reader` for the reader method to avoid `read_stdout_reader`.)
- Bikeshed (minor): Perhaps `ProcessError` rather than `SubprocessError`? We currently use the term "process" in various places, and not "subprocess". Plus, this error type seems useful for any kind of process.
- Bikeshed: s/`ChildOutputRead`/`ChildOutputReader`/, and likewise s/`_read`/`_reader`/ for the method returning it. "reader" seems to be the ecosystem convention for "things that implement `Read`", and `Stream` is something else entirely. (Regarding the item in the "Drawbacks" section, though, I'm completely in favor of *having* this function and type.)

-----

Next, the semantics:

- Semantic: Please spell out exactly how you're expecting people to retrieve stdout-and-stderr interleaved, as they can currently do with `output`. If this isn't as simple as "call a single non-deprecated method", I think that needs improving. I *think* the proposal is that if you route stderr to the same place you route stdout, rather than capturing it in a separate pipe, then you'll get the combined output and stderr won't get treated as a failure? In any case, that needs documenting and needs to be simple. (Common use case: spawning a compiler-like tool.)

- Semantic: We should not encourage developers to fail on non-UTF-8 in the common case. I think `get_stdout` should return bytes, and we should offer a `get_stdout_string` that returns a UTF-8 `String`. Similarly, the error type should have methods `stdout` and `stderr` rather than `stdout_bytes` and `stderr_bytes`.

- Semantic (exotic): I have some use cases (involving UNIX sockets in datagram mode) for which I *cannot* keep reading stdout until EOF, and instead I *have* to stop reading stdout as soon as 1) the process exits and 2) a non-blocking read returns nothing. Would it be possible (either in-std or outside std) to implement a reader with that semantic atop these APIs?

- Semantic: Please document that we need `impl Error for ProcessError`. Also, that implementation is going to have a little trouble implementing the `cause` method, unless there's a semantic guarantee that only one of the types of error can happen. This *seems* like it'd be guaranteed for errors produced by the standard library, but the APIs for building a `ProcessError` don't statically guarantee that, nor do the accessor methods for `ProcessError`. If we could maintain that guarantee, such that there's only one chained error, that seems ideal.

- Semantic: I think we should show the command-line arguments in the default case, even though they might be verbose. However, sometimes the command-line arguments might be sensitive. For convenience for such applications, perhaps a `hide_args()` method that avoids including the args in the error. (We could get *more* exotic with a `.masked_arg()` method that supplies and masks *one* argument to the command, but that's much more complex and I don't think that should be in this RFC. `hide_args()` seems simple enough though.)

- Semantic: I can imagine the convenience value of the `_line` variant, but could you give some examples of code that would benefit from it? I'm also wondering if it'd be reasonably convenient to call `read_stdout_string` and then have a convenience method on `String` for "fail if not exactly one line", which seems useful in other contexts too.

- Semantic (minor): Please add a note to the "Further necessary APIs for `SubprocessError`" section that these APIs could be under a separate feature gate and stabilized separately; that'll make it easier to potentially do a partial stabilization.

afetisov:
> I think get_stdout should return bytes,

Shouldn't it be `OsString`? 

> However, sometimes the command-line arguments might be sensitive. For convenience for such applications, perhaps a hide_args() method that avoids including the args in the error. 

Sensitive things should not be passed as command line arguments, since those can leave traces in arbitrary places around the OS. For example, they may be included in `ps` output. We should not encourage half-baked security practices.

joshtriplett:
> Shouldn't it be OsString?

No. On Windows that would be WTF-8, not raw bytes.

tmccombs:
>> I agree with the motivation that Command is not very ergonomic, especially around the common "return error containing stderr on error" workflow. But this RFC is a little too opinionated about stderr in my opinion. It strictly treats it as the "error output",
>
> Um? That's not true. In my RFC, by default, error output is sent to the Rust program's stderr (inherit).

But what if you also want to capture stderr? 

As @joshtriplett mentioned `output` currently gives you interleaved stdout/stderr, and if that is deprecated, there isn't a way to get that.  Perhaps one way to do that would  be to add a new constructor to `Stdio` that would allow you to redirect stderr to stdout or vice versa? 

And should there be functions for getting stdout and stderr as separate `String`s or `Vec<u8>`s? Of course it is possible to implement that using `spawn` and `Child`, but that is true of all these other more ergonomic functions as well.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+
+Most callers should use these methods,
+rather than `.spawn()` or `.status()`.
+These new methods should be be
+recommended by docs for other, lower-level functions
+which people were previously required to use.
+
+## Deprecations
+
+ * Deprecate `std::process::Command::output()`.
```
thomcc, line 170: It's not clear to me that `Command::output` is that bad. It's slightly awkward and requires manually checking errors, but I'm not convinced this is worth deprecating a *extremely* widely-used function and causing a large amount of ecosystem churn (which many users will respond to with `#[allow(deprecated)]`, due to MSRV+the existing method not being broken).

I think you need to make a much clearer case for why this needs deprecation, and document the downside of ecosystem and educational churn as folks move to the new API(s).

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+    fn spawn_error() -> Option<&io::Error>;
+
+    /// If trouble included failure to talk to the child, the IO error.
+    ///
+    /// This might include problems which might be caused by child
+    /// misbehaviour.
+    fn communication_error() -> Option<&io::Error>;
+
+    /// If trouble included failed UTF-8 conversion.
+    fn utf8_error(&self) -> Option<&std::str::FromUtf8Error>;
```
thomcc, line 249: Which UTF-8 conversion does this refer to? I think this probably should be pushed down to whatever getter would return something UTF-8 validated rather than having this stored as a field on the error.

If this *is* stored (and almost certainly if it is not), we should use `std::std::Utf8Error` rather than a `&std::string::FromUtf8Error` in order to avoid needing to store the invalid UTF8 twice (and to avoid forcing us to convert eagerly).

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+
+    fn set_program(&mut self, impl Into<OsString>);
+    fn set_args(&mut self, impl IntoIterator<Item=impl Into<OsString>>);
+    fn set_stdout_bytes(output: Option<Into<Box<[u8]>>>);
+
+    fn set_status(&mut self, status: ExitStatus);
+    fn set_stderr_bytes(&mut self, stderr: impl Into<Box<u8>>);
+    fn set_spawn_error(&mut self, error: Option<io::Error>);
+    fn set_communication_error(&mut self, error: Option<io::Error>);
+    fn set_utf8_error(&mut self, error: Option<std::str::FromUtf8Error>);
```
thomcc, line 297: I think too much of this is mutable -- either we should just make these be public fields (we have `get_` and `set_` methods for them...) or we should just allow constructing one of these from parts.

Additionally it's not clear to me how these interact. Up until now, I had imagined that there was an internal enumeration 

```rs
enum SubprocessErrorCause {
    Spawn(io::Error),
    Communication(io::Error),
    // Actually, I don't think we want this one, but anyway
    Utf8(str::Utf8Error),
    // ...
}
```
but this API has individual setters for each one, which implies it either is not that way, or they have perhaps-confusing interactions.

Does it make any sense for there to be more than one error here anway? Which would gets reported in `std::error::Error::source`?

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,542 @@
...
+    /// `stderr_bytes()`, `spawn_error()` and `utf8_error()`.)
+    //
+    // We must provide this because it's needed for handling programs
+    // with unusual exit status conventions (eg `diff(1)`)
+    // and a caller can't reimplement it without making assumptons
+    // about `SubprocessError`'s contents.
+    fn has_problem(&self) -> bool;
+}
+impl Default for SubprocessError { ... }
+impl Clone for SubprocessError { ... } // contained io:Errors are in Arcs
```
thomcc, line 313: I think we should instead ask users who need a `Clone`able `SubprocessError` to use `Arc<SubprocessError>`, rather than imposing that requirement on the internals.

thomcc (commented):
I think this is well-motivated broadly, however needs some work:
1. I think deprecating `Command::output` is probably too aggressive, or at least needs a lot more justification.
2. I agree with @joshtriplett in that there's a lot bikeshed-painting to be done about the naming here -- most of these names don't follow the conventions of the ecosystem/stdlib that well.
3. I think the mutable/constructable `SubprocessError` portion of the API needs more work -- Ideally it could be a lot smaller and have fewer moving pieces. Perhaps it's trying to be too much at once?

text/0000-command-ergonomics.md
ijackson: This was a mistake, thanks.

ijackson:
Thanks for all the comments, and the encouragement.  I think in most cases I will be modifying the RFC in line with the suggestions.  And I'll add some examples (and add some more stuff to the alternatives section).

> I think the mutable/constructable `SubprocessError` portion of the API needs more work -- Ideally it could be a lot smaller and have fewer moving pieces. Perhaps it's trying to be too much at once?

I agree that it is uncomfortably large, but, sadly, I think it is inevitable.  I think what I will do is make it a transparent struct (for it is a struct, not an enum) and add a lot more commentary explaining why it is (has to be) that way.

text/0000-command-ergonomics.md
ijackson: I think it should succeed and give an empty string.  Will write that into the RFC.

text/0000-command-ergonomics.md
ijackson: This is stated explicitly in the docs for what I am now calling `read_stdout_bytes`.  All of these functions wait for both EOF and process exit.  I will add a short xref note to the other two functions.  In the real docs the spec text should probably be cut and pasted instead.

ijackson:
@joshtriplett Thanks for the review.

> First, the bikesheds:

Your suggestions mostly made sense to me, so I have simply accepted them.

I'm still unsure about `read_stdout`,
because it does more than just *read* the stdout:
it also spawns and waits for termination.
I have added some text about this to the relevant part of Unresolved Questions.

> Please spell out exactly how you're expecting people to retrieve stdout-and-stderr interleaved, as they can currently do with `output`

I agree that this is an important use case.

But think this is not in fact currently possible.
`Output` contains *two* `Vec<u8>`,
so stdout and stderr are not interleaved but separated.
And it is not possible to reconstruct the order of output.

I think this requirement should be met by a new configuration method on `Command`.
This seems conceptually independent of my RFC,
and could be done either before or after.
So I have added that to Future Possibilities.

Note also
https://github.com/rust-lang/rust/pull/88561
where I propose to provide the ability to redirect the subprocess's
stdout to one's own stderr, or vice versa.

> We should not encourage developers to fail on non-UTF-8 in the common case. I think `get_stdout` should return bytes, and we should offer a `get_stdout_string`

Isn't this just a question of naming ?
I think this should have been in your "bikeshed" list :-).

I looked around in `fs` and concluded that you're right and
we usually use the undecorated name for the bytes version.
So I've renamed
`read_stdout` to `read_stdout_string`
and
`read_stdout_bytes` to `read_stdout`.

> Semantic (exotic): I have some use cases (involving UNIX sockets in datagram mode) for which I cannot keep reading stdout until EOF, and instead I have to stop reading stdout as soon as 1) the process exits and 2) a non-blocking read returns nothing. Would it be possible (either in-std or outside std) to implement a reader with that semantic atop these APIs?

I think you can do this already with `piped()` and `Child::try_wait` and so on.
This is not the kind of thing that my new API is likely to help you very much with,
although possibly if you find that things go wrong you may find it convenient to
make a `ProcessError` out of your program's experiences.

> Semantic: Please document that we need `impl Error for ProcessError`.

Done, and discussed `cause`.

> Semantic: I think we should show the command-line arguments in the default case

I added some notes about the question of sensitive arguments.
I do still think we should show them.

> Semantic: I can imagine the convenience value of the `_line` variant, but could you give some examples of code that would benefit from it?   \[suggestion of convenience method on `String`]

Many programs on Unix are expected to print a single line.
`whoami`, `git rev-parse`, `wc`.

Its main use is that it trims the newline for you,
but you could do that (with slightly less error checking) with `.trim_end()`.

TBH I am not 100% sure this function carries its weight.
But OTOH the method function `String` approach isn't able to include the command information in its error.
On the third hard,
that is true of basically any parsing one does of the output.

In the interests of making progress, I think it's best to decouple this question from my RFC.
So I have moved it to the Future Possibilities section.

> "Further necessary APIs for `SubprocessError`" section

*Some* of these APIs are separable, but at least `.has_problem()` isn't.
I have added notes to these.

The getters and setters are gone now, which makes this much smaller.

text/0000-command-ergonomics.md
ijackson: I think you and Josh are right about this.  I have moved the deprecation to Future possibilities.

text/0000-command-ergonomics.md
ijackson: I'm using `std::str::FromUtf8Error` which doesn't contain or reference the string, but does contain the offset.  I think this doesn't suffer from the problem you describe?

text/0000-command-ergonomics.md
ijackson: I have made this a transparent struct.  Making it opaque was indeed confusing.

text/0000-command-ergonomics.md
ijackson: Fair enough, I have changed this.

text/0000-command-ergonomics.md
```diff
@@ -342,6 +344,13 @@ impl SubprocessError {
     // and a caller can't reimplement it without making assumptons
     // about `SubprocessError`'s contents.
     fn has_problem(&self) -> bool;
+
+    /// Returns `Ok<ExitStatus>` if the only reason for the failure was a nonzero exit status.
+    /// Otherwise returns `self`.
+    ////
+    /// Use this if you want to to tolerate some exit statuses,
+    /// but still fail if there were other problems.
+    pub fn just_status(self) -> Result<ExitStatus, SubprocessError>;
```
joshtriplett: This feels a little awkward to use: if you call this to filter out some exit statuses, you lose the rest of the error, which means you don't have `stdout_bytes` anymore. What code pattern would you suggest for code that wants to run `diff`, check the exit code for 0/1 to determine if there was a difference, bail on >1, and capture the patch output?

text/0000-command-ergonomics.md
joshtriplett: It seems like this helper would commonly be used for programs for which you expect to parse the output, in which case having to separately check for empty string seems to defeat part of the ergonomic win of using the helper.

But in any case, this isn't a blocker as you've split this helper out into possible future work.

joshtriplett:
Nit (emphatically not a blocker): your commit message "Rename to ChildOutputStream" seems inaccurate (specifically the "to"), as it renames `ChildOutputStream` to `ChildOutputReader`.

text/0000-command-ergonomics.md
```diff
@@ -433,6 +434,29 @@ for process invocation and output handling.
 Perhaps printing the command arguments is overly verbose,
 and we should print only the command name.
 
+## `cause` in `Error` impl of `ProcessError`
+
+Because maybe several things went wrong, ever providing a `Some` `cause`
+would involve prioritising multiple possible problems.
```
joshtriplett, line 449: Can you give an example where multiple different things could simultaneously go wrong? It *seems* like `spawn_error`, `communication_error`, and `utf8_error` should be mutually exclusive.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,634 @@
...
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+## Printing command arguments in `impl Display for ProcessError`
+
+Perhaps printing the command arguments is overly verbose,
+and we should print only the command name.
+
+Sometimes people pass sensitive information (such ass passwords) in command line arguments.
+This is not a good idea, because command line arguments are generally public on Unix.
```
joshtriplett, line 441:
```suggestion
This is not a good idea in portable software, because command line arguments are often public on Unix targets.
```

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,634 @@
...
+(which is the only way to preserve the relative ordering)
+you must do one of:
+
+ * run a command which itself does the redirection
+   (easy using the shell on Unix)
+
+ * send them each to your own stdout/stderr with `inherit`
+   and expect your caller to combine them
+
+ * sent themk to the *same* one of your stdout/stderr
```
joshtriplett, line 549:
```suggestion
 * send them to the *same* one of your stdout/stderr
```

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,634 @@
...
+We should provide something like this:
+
+```
+impl Command {
+    /// Arranges that the command's stderr will be sent to wherever its stdout is going
+    fn stderr_to_stdout();
+}
+```
+
+(It is not sensibly possible at least on Unix
```
joshtriplett, line 562:
```suggestion
(It is not easy on Unix targets
```

It is possible, though quite awkward and with some limitations. (Even more limitations if you're not on Linux.) I needed a solution to this, and ended up writing [io-mux](https://crates.io/crates/io-mux) for this.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,634 @@
...
+
+  * The `io::ErrorKind` from the spawn error, if any.
+
+  * Otherwise, a new kind `io::ErrorKind::ProcessFailed`,
+    which means that the subprocess itself failed.
+
+### Further APIs for `ProcessError`
+
+A `ProcessError` is a transparent `Default` struct so it can be
+constructed outside std, for example by async frameworks.
```
joshtriplett, line 313:
```suggestion
constructed outside std, such as by async frameworks or other
code launching processes and handling errors from them.
```

text/0000-command-ergonomics.md
joshtriplett: I wonder if this (straw proposal) might be easier to use:

```rust
/// If the process is only considered to have errored due
/// to its exit code, calls the provided filter, returning
/// `Ok(Self)` if it returns `true`, or `Err(Self)` if it
/// returns `false`.
pub fn permit_statuses(self, f: impl FnOnce(ExitStatus) -> bool) -> Result<ProcessError, ProcessError>
```

This would make it relatively straightforward to say "allow 0 or 1, reject everything else". (Still mildly annoying due to `code()` returning `Option`, but doable.)

joshtriplett:
I've nominated this for the next libs-api meeting. I anticipate proposing FCP on it, unless folks raise any other issues before then.

text/0000-command-ergonomics.md
ijackson: I quite like this idea.  I wasn't sure people would like the notion of returning `Ok(ProcessError)`, but if you think that's OK it seems OK to me.

>  mildly annoying due to `code()` returning `Option`

I wondered whether the closure should receive `ExitCode` (so would only be called if the process exited rather than was killed).  But then you can't use this to ignore (say) `SIGPIPE`.

I do kind of hate booleans.  Would `impl FnOnce(ExitStatus) -> Result<(), ()>` be too awful?

text/0000-command-ergonomics.md
ijackson: I think the following can all occur in the same execution:
 * The application calls `.stderr(piped)` and `.read_stdout()`
 * Spawn succeeds
 * The subprocess produces some invalid-UTF output
 * Then, something in the kernel goes very badly awry (or, some other thing in our Rust program wrecks our fd) and `read()` on one of the pipes fails
 * The subprocess prints some stderr (which the application has said to treat as an error)
 * The subprocess then crashes.

We must report *at least* all of: the error from `read`; the stderr output (which might well be very helpful for diagnosis); and the process's exit status.

I think we should eagerly do a unicode validity check whenever we captured the stdout, and record the unicode conversion error in the `ProcessError`.  That way a caller who calls your `.permit_status()` and gets `Ok` can know that the stdout output is UTF-8; conversely, if the command unexpectedly printed non-UTF-8, the caller can conveniently report the command line etc.

Or to put it another way, we shouldn't "short circuit" some checks and report only a subset of the errors.

ijackson:
I had been thinking about this, and in particular the implementation, and felt that `communication_error: Option<io:Error>` was suboptimal, because it makes handling of subprocess communication errors less correct (the `kind()` isn't very useful).  See the reasoning I have left in the comment there.

I'm not 100% sure whether this is better than my previous proposal.  One option woudl be to leave this one way or the other for now, and let the implementor decide.

text/0000-command-ergonomics.md
```diff
@@ -560,8 +560,8 @@ impl Command {
 }
 ```
 
-(It is not sensibly possible at least on Unix
-to get all of the stdout and stderr output
+(It [can be difficult or impossible](https://docs.rs/io-mux/latest/io_mux/) at least on Unix
```
joshtriplett, line 589: (I don't actually need this cited, it's off-topic for this RFC, I just didn't want it described as impossible. :) )

ijackson: I know you weren't wanting a plug but I like to give references :-).  If I had been the reader rather than the author, I would have appreciated that link which is why I put it in.  I can take it out if it's not appropriate...

joshtriplett: Perfectly fine, I just wanted to make sure it hadn't come across like I was asking for a plug.

text/0000-command-ergonomics.md
nagisa: Would it be possible to make this change without an RFC? Seems somewhat unfortunate to tie this together with the API additions.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,656 @@
...
+
+Perhaps we don't need all the `read_stdout` variants,
+and could require Bob to write out the boilerplate
+or provide his own helper function.
+
+Perhaps we don't need `stdout_reader`.
+However,
+avoiding deadlocks when reading subprocess output,
+and also doing error checks properly,
+is rather subtle.
```
joshtriplett, line 400: Could you clarify how `stdout_reader` makes avoiding this deadlock easier?

Is this just that you need to interleave `.write` calls to the process stdin with `.read` calls to the reader, rather than calling `.write_all` and then reading all the output at once?

ijackson: The deadlocks that `stdout_reader` avoids completely are ones ike: waiting for the command's stdout EOF, and child termination, in the wrong order; waiting for stderr EOF (which can sometimes block forever for complicated reasons).

It helps only a little with deadlocks resulting from piping both into and out of the same `Command`.  (I think it may still help in those use cases, but it *doesn't* help with "you wrote too much to the command's input while it's trying to write output to you".)

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,656 @@
...
+   `.run()` or `.read_stdout*` instead.
+   Do not deprecate `.output()` though.
+
+## stderr handling
+
+If `stderr(Stdio::piped())`,
+these new functions all collect the child's stderr.
+Then,
+if the stderr output is nonempty, this is considered an error,
+and reported in the `ProcessError`.
```
Twey, line 176: I'm a bit uncertain about this semantics.  It's popular in command-line applications to use `stderr` to collect running commentary about what the program is doing (for example `log::debug` under `env_logger`) in ways that definitely shouldn't be considered errors.

matklad: I have an experience report that erroring out on non-empty stderr is wrong! One common case here is when you run something like `rustc --version`, but that thing is `rustup` managed, so, before you get to `rustc --version` output, you have printout from `rustup` on the stderr that it downloads the appropriate version.

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,656 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+## Printing command arguments in `impl Display for ProcessError`
+
+Perhaps printing the command arguments is overly verbose,
+and we should print only the command name.
+
+Sometimes people pass sensitive information (such ass passwords) in command line arguments.
```
jieyouxu, line 466: `ass` -> `as` (sorry)

text/0000-command-ergonomics.md
```diff
@@ -0,0 +1,656 @@
...
+Perhaps we don't need `stdout_reader`.
+However,
+avoiding deadlocks when reading subprocess output,
+and also doing error checks properly,
+is rather subtle.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Alternatives and prior proposals include:
```
jieyouxu, line 405: Remark: one other possible alternative: don't introduce these APIs yet, and try to cook up some kind of wrapper on top of `std::process::Command` outside std first. It seems to me that different users may have different preferences for the `std::process::Command` APIs (especially considering how hard it is to reach any kind of consensus), which pulls the API design in many directions.

tamird, line 405: How about the alternative of just adding `#[must_use]` annotations roughly as implemented in https://github.com/rust-lang/rust/pull/81452 (plus on Output, which is missing)?

PR #3375, opened 2023-01-22
3192-dyno update plugin example

GitNiko:
fixed function `log_active_rules` parameter

[Rendered](https://github.com/GitNiko/rfcs/blob/patch-1/text/3192-dyno.md)


ehuss (approved):
Thanks!

PR #3376, opened 2023-01-22
Mark Markdown files as linguist-detectable

fwcd:
This follows the precedent of similar repos such as [`swift-evolution`](https://github.com/apple/swift-evolution/) and makes Markdown show up as the primary language in GitHub's language bar:

image


ehuss (approved):
Seems reasonable, thanks!

PR #3380, opened 2023-01-30
Permissions

DasLixou:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3380)*

Just wanna share my idea of a language with permissioned functions.
Feedback welcome, really interested in this 😊

[Rendered](https://github.com/rust-lang/rfcs/blob/741c66cbc0aea7b435354b2b6ae744888c742f6e/text/0000-permissions.md)

### DIFF
--- a/text/0000-permissions.md
+++ b/text/0000-permissions.md
@@ -39,7 +39,7 @@ pub struct Person {
 And now we want a function for `Person` which lets us age him/her by one year and another one for renaming him/her.
 ```rust
 impl Person {
-  pub fn aging(&ut self) {
+  pub fn aging(&mut self) {
     self.age += 1;
   }
   
@@ -87,7 +87,7 @@ fn main() {
 Dang it! But with permissions, we can easily fix this. So let us change some things in our `impl` of `Person`:
 ```rust
 impl Person {
-  pub fn 'Age aging(&ut self) { // this function needs the permission `'Age`
+  pub fn 'Age aging(&mut self) { // this function needs the permission `'Age`
     self.age += 1;
   }
   
--- a/text/0000-permissions.md
+++ b/text/0000-permissions.md
@@ -1,0 +1,186 @@
+- Feature Name: `permissions`
+- Start Date: 2022-01-30
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+
+# Summary
+[summary]: #summary
+
+Set permissions for functions and add permissioned borrows.
+
+# Motivation
+[motivation]: #motivation
+
+So I've been latetly writing a lot of rust code where I need a struct mutable borrowed and stored paralelly to other borrows of the same data.
+They really don't interop or block each other, they are two iterators nested and one needs the whole type to call a function on it (immutable) and get that data.
+The other iterator needs a mutable borrow of an underlying `HashMap` in that same struct to convert the key to a mutable ref to the value.
+But these two iterators nested wouldn't work, because they would both reference to the same data.
+To resolve the error: ```cannot borrow `*app` as mutable more than once at a time
+second mutable borrow occurs here```, I've just be using unsafe pointer magic..:
+```rust
+// ..with a safety comment above
+let ptr: *mut G = &mut *app;
+let inner = Self::new(&*ptr);
+```
+While I wrote this, the idea of a langauge with permissions came up to me.
+
+> The explanation of how it should work follows in the next chapter.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+So, a permission is looking like a life-time specifier.
+Imagine we have a struct for a person:
+```rust
+pub struct Person {
+  name: String,
+  age: u8,
+}
+```
+And now we want a function for `Person` which lets us age him/her by one year and another one for renaming him/her.
+```rust
+impl Person {
+  pub fn aging(&ut self) {
+    self.age += 1;
+  }
+  
+  pub fn rename(&mut self, new_name: String) -> String {
+    std::mem::replace(&mut self.name, new_name)
+  }
+}
+```
+Not too fancy, right? But we now want a `Department` for changing his name and a `Calendar` to age a person by one year.
+```rust
+pub struct Department<'p>(&'p mut Person); // ideally we want more than one person, but "simplicity"
+
+impl Department<'_> {
+  pub fn rename(&self, new_name: String) -> String {
+    self.person.rename(new_name)
+  }
+}
+
+pub struct Calendar<'p>(&'p mut Person); // ideally we want more than one person, but "simplicity"
+
+impl Calendar<'_> {
+  pub fn next_year(&self) {
+    self.person.aging();
+  }
+}
+```
+Still simple, right? So let us create some life 😊
+```rust
+fn main() {
+  let mut person = Person {
+    name: "Mike",
+    age: 0,
+  };
+  
+  let department = Department(&mut person);
+  let calendar = Calendar(&mut person); // ERROR: cannot borrow `*person` as mutable more than once at a time second mutable borrow occurs here
+  
+  calendar.next_year();
+  calendar.next_year();
+  calendar.next_year();
+  department.rename("Jonathan");
+  calendar.next_year();
+}
+```
+Dang it! But with permissions, we can easily fix this. So let us change some things in our `impl` of `Person`:
+```rust
+impl Person {
+  pub fn 'Age aging(&ut self) { // this function needs the permission `'Age`
+    self.age += 1;
+  }
+  
+  pub fn 'Name rename(&mut self, new_name: String) -> String { // this function needs the permission `'Name`
+    std::mem::replace(&mut self.name, new_name)
+  }
+}
+```
+Now let us change the borrow for `Department` and `Calendar`:
+```rust
+pub struct Department<'p>(&'p 'Name mut Person); // the borrow has the permission `'Name`
+pub struct Calendar<'p>(&'p 'Age mut Person); // the borrow has the permission `'Age`
+```
+And now let's see our main code:
+```rust
+fn main() {
+  let mut person = Person {
+    name: "Mike",
+    age: 0,
+  };
+  
+  let department = Department(&mut person); // implicit permission
+  let calendar = Calendar(&'Age mut person); // explicit permission
+  
+  calendar.next_year();
+  calendar.next_year();
+  calendar.next_year();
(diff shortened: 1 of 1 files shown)


TimNN:
There have been a number of previous discussions on similar topics, e.g. https://internals.rust-lang.org/t/view-types-based-on-pattern-matching/16879 or https://github.com/rust-lang/rfcs/issues/1215 in this repo. 

Some terms that have been applied to this kind of functionality are "partial borrows" or "view types".

text/0000-permissions.md
```diff
@@ -0,0 +1,186 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+So I've been latetly writing a lot of rust code where I need a struct mutable borrowed and stored paralelly to other borrows of the same data.
+They really don't interop or block each other, they are two iterators nested and one needs the whole type to call a function on it (immutable) and get that data.
+The other iterator needs a mutable borrow of an underlying `HashMap` in that same struct to convert the key to a mutable ref to the value.
+But these two iterators nested wouldn't work, because they would both reference to the same data.
+To resolve the error: ```cannot borrow `*app` as mutable more than once at a time
+second mutable borrow occurs here```, I've just be using unsafe pointer magic..:
```
konsumlamm, line 18: Could you write this down in code? I find it hard to understand what you mean.

text/0000-permissions.md
```diff
@@ -0,0 +1,186 @@
...
+let inner = Self::new(&*ptr);
+```
+While I wrote this, the idea of a langauge with permissions came up to me.
+
+> The explanation of how it should work follows in the next chapter.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+So, a permission is looking like a life-time specifier.
```
konsumlamm, line 31: This feels out of place. You start by saying that a permission looks like a lifetime, but then you don't continue to describe permissions, but rather first define some example types/functions.

text/0000-permissions.md
```diff
@@ -0,0 +1,186 @@
...
+  pub fn 'HackA hack_a(&mut self) {
+    self.one_data./*...*/
+  }
+  
+  pub fn 'HackB hack_b(&mut self) {
+    self.one_data./*...*/
+  }
+}
+```
+Instead, the compiler should notify that the permissions intercept with each other and they should create a sub-permission for both:
```
konsumlamm, line 140: So the compiler checks which (private) fields the functions use? This seems to be very much like #1215.

text/0000-permissions.md
```diff
@@ -0,0 +1,186 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+I think this is one of the cleanest options for these problems. Other devs may be changing their whole code structure for that or use `unsafe`, but I like the idea of permissions.
+It maybe "could" be done with macros, but then it would also generate unsafe code and wont be 100% safe.. but the simplest impl of permissions may be possible to implement via them.
+
+# Prior art
+[prior-art]: #prior-art
+
+I don't know any language who has done something like this before.
+The idea came to me by minecraft server development and there its used for giving players the permission to execute commands, so it may be similar in the smallest way possible 😅.
```
konsumlamm, line 174:
```suggestion
The idea came to me by Minecraft server development, where it is used for giving players the permission to execute commands, so it may be similar in the smallest way possible 😅.
```

kornelski:
I think this kind of functionality is definitely needed in Rust. Getters and setters are too restrictive, and it's not always feasible to expose the fields or split structs.

Compared to previous proposals, an interesting development here is having abstract names for sets of loans, instead of exposing field names in the interface.

blueglyph:
Each time we read or write code like this, we would need to check which permission allows what, to understand the implication and to make sure it's correct. It seems to go against the visibility that Rust has maintained when it comes to safety features.

```rust
  let department = Department(&mut person); // implicit permission
  let calendar = Calendar(&'Age mut person); // explicit permission
```

This also means that you create an access partition in your methods, right? Because now you have to ensure that `Age` and `Name` don't have any common field `mut` reference. This will restrict what you can do in the "base" object methods, which may prove more an annoyance than a benefit. Unless you duplicate the methods to avoid those with permissions if you need, for example, to mutate both `age` and `name`... but this implies that these other methods **cannot** be used by `Department` or `Calendar`, thus further complicating the comprehension of the code (and the job of the compiler).

Also, permissions are defined in the "base" object, not by other code that reference it or in trait implementations. So I'm not sure how the abstraction can be an advantage, unless you create your object knowing exactly how to partition the mutability to anticipate future needs. Or will you anticipate all possibilities? Perhaps in some case it's obvious, but it seems to only make sense if you already know how it will be used, so it's probably not meant to be exposed outside the scope of a crate often. Maybe it's just not the right example, because here you are directly accessing fields. Maybe a more abstract permission name makes more sense with public methods.

I would venture that the compiler should already be able, _theoretically_, to determine whether two functions requiring mutability have an overlap in their write accesses or not. But the visibility wouldn't be good, it would be somewhat improved with an explicit declaration like the one proposed here. Yet it would not be visible enough in my opinion, unless maybe it's based on fields rather than a permission abstraction, and explicit when calling the methods.

PS: _Just my nit-picking self: you start your motivation by "So", as if it were a conjunction, but there is no previous text that this could result of. I would remove it and start by "I've been writing"._

DasLixou:
Yeah.. this was just a stupid idea i wanna write out. #1215 seems way more well thought out

konsumlamm:
> Yeah.. this was just a stupid idea i wanna write out. #1215 seems way more well thought out

Note that #1215 isn't actually an RFC, it's just an issue. Someone would have to go ahead and make a PR for it.

matthieu-m:
> Yeah.. this was just a stupid idea i wanna write out. #1215 seems way more well thought out

Just because an idea is not fully fleshed out doesn't mean it's stupid. The very RFC process is about coming all together to _improve_ on an initial idea after all.

---

With regard to permissions, the compiler should be able to check them to ensure safety.

This, in turn, requires that _fields_ be annotated with the permissions, which then lets the borrow-checker:

 1. Validate that only functions declaring the right permissions can access the fields.
 2. Partition the borrows to only the set of fields tagged with the permission.
 3. Refuse concurrent mutable borrows of overlapping permissions -- ie, permissions that both borrow a common field.

Throw in an (implicit) `'All` permission which borrows all fields (thus overlaps with all other permissions) and is implicitly assumed whenever a method doesn't specify a permission -- the most common case -- and... I think this works?

DasLixou:
Yeah but there are also other thinks where we would like Permissions but which just won’t work. For example: HashMap. What If I want a permission per key. Like `'Person("Mike")` could mean that it has access to the hashmap entry with the key Mike but that  won’t be really a readable permission… I think there’s no good way to fix this.. so permissions would only work for fields which doesn’t really gets fully rid of the problem..

AaronKutch:
This sounds like a thing that could be accomplished by crates. For example, putting `RefCell`s in my arena crate https://crates.io/crates/triple_arena and using its extremely flexible immutable borrowing could probably achieve what you want. Most algorithms don't need more than two mutable borrows at once (achievable without `RefCell`s with `get2_mut`, although there are more general get-multiple-muts I could write for my crate), often you can use multiple arenas to split off orthogonal structures or use consecutive mutable borrows.

If there is something that could be improved in Rust, it is some mechanism of allowing multiple immutable and mutable borrows of elements of a resizable data structure, if the compiler can verify that there are no movement or resizing operations while the borrows are held, and keys are unique. That is the problem I run into the most when using my arena or other similar structures.

DasLixou:
Yeah but other than RefCells permissions should be a syntactic safety that it is safe without having any things in runtime

DasLixou:
I also thought about directly declaring the permissions today lol. And yeah lifetime / labeled style may be too confusing.. maybe use `*`? Like
```rust
pub fn get(&*inner self) { .. }
// or `$`?
pub fn get(&$inner self) { .. }
```
?

DasLixou:
Ok so i changed everything up. Also came across the idea of just semantic permissions. Because sometimes (mostly in traits) you just want an internal function with for example `pub(crate)` but therefor you would need a second trait and need the one trait impl the other. Permissions would make this lot easier.

clarfonthey:
Small note reading this thread: the term "permissions" made me think of C++'s "friend" syntax (which is a mess) and then when I found out what it was about, seems still like a bad name.

I would recommend calling this explicit partial lifetimes, or named partial lifetimes, since it better describes the feature. I don't care either way, but I think it would help to pitch what the feature is actually for.

I personally don't like the proposal as it is, but don't have anything better to propose than what other folks have already said, so, I'll let everyone else discuss. But figured I'd pitch in an alternative colour for the bike shed so that folks looking at this have a better idea what it's about.

burdges:
It's all partial borrows in one flavor or another..

At present, you need methods like `[T]::split_at_mut` to do this, which requires writing explicit substructs, and thus more a design decisions. 

Aside from [view types](https://internals.rust-lang.org/t/view-types-based-on-pattern-matching/16879) and https://github.com/rust-lang/rfcs/issues/1215 we also discussed this under https://github.com/rust-lang/rfcs/pull/1546 which then became https://github.com/nikomatsakis/fields-in-traits-rfc ala https://github.com/nikomatsakis/fields-in-traits-rfc/issues/16

I suspect fields-in-traits kinda blocks anything more ad hoc like this, so if you want some more ad hoc solution, then you might justify how the ad hoc solution works better somehow.  If otoh you think fields-in-traits works best, then maybe attempt to make some progress there. 

Imho, fields-in-traits works best because otherwise we'll largely loose partial borrows when programming generically.

workingjubilee:
Triaging: author seemed to address feedback [here](https://github.com/rust-lang/rfcs/pull/3380#issuecomment-1416003815). It's not clear to me that further comments warranted further direct response from the author. Seems like it's waiting on team to say something?

DasLixou:
Oh thanks for reminding me. Afaik there's nothing blocked by me, but I lately had another thought about how this could be made a bit more abstract and work for traits aswell, but I thought that this rfc was already closed lol. Should I add this option or is the rfc already too deep in the review process?

nikomatsakis:
Hi @DasLixou,

This looks like very similar to some of the work @scrabsha and I [are doing on view types](https://rust-lang.github.io/rust-project-goals/2026/view-types-experiment.html) (together with [abstract fields](https://smallcultfollowing.com/babysteps/blog/2025/02/25/view-types-redux/)). I'm going to close it but if you're still interested in exploring this feel free to reach out to me on Zulip and we can discuss.

PR #3384, opened 2023-02-07
Add RFC: uri-crates

hydroper:
[Rendered](https://github.com/hydroper/rfcs/blob/master/text/0000-uri-crates.md)

### DIFF
--- a/text/0000-uri-crates.md
+++ b/text/0000-uri-crates.md
@@ -45,13 +45,17 @@ When the dependency string contains at least one dot, it is an URI.
 
 The `package.uri` field of Cargo.toml is an optional string based in the regular expression `(www.)?[a-z_\-0-9]+(\.[a-z_\-0-9]+)+(/[a-z_\-0-9]+)*` which is similiar to an URI without a scheme, query, hash fragment and `%` octet sequences. In other words, it contains one or more identifiers delimited by dot followed by an optional slash and zero or more identifiers delimited by slash.
 
+The optional "www." prefix is ignored.
+
 #### Dependency
 
 If a dependency string contains at least one dot, the dependency string is an URI. Cargo resolves it as a crate identified by that URI in crates.io
 
+The optional "www." prefix is ignored.
+
 #### Publishing process
 
-An user cannot publish a crate with a specific URI if its domain does not exist in crates.io or does not belong to the user.
+An user cannot publish a crate with a specific URI if either its domain does not exist in crates.io, its domain does not belong to the user or the URI is duplicate.
 
 #### crates.io Domains
 
@@ -61,7 +65,7 @@ The `cargo` command allows managing domains from crates.io. For example, it allo
 - registering domains,
 - removing domains,
 - transferring domain ownership between users and
-- adding other users as owners of a domain
+- adding other users as owners of a domain.
 
 # Drawbacks
 [drawbacks]: #drawbacks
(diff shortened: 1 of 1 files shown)


slanterns:
I don't think it has a clear motivation. Why we really need an (unverified) URI(-like string) to specify the package organization, instead of some simpler ways?

17cupsofcoffee:
My main concern with this idea is that, to me at least, using URIs in this way implies a level of verification/trust?

If I were to see `package = "www.tokio.rs/foo"` in a Cargo.toml, my first assumption would be that was either where the crate was hosted (there's already precedent for this with `git` dependencies!), or that the crate was definitely owned by the same people that own the tokio.rs domain name. The fact that neither of these things are true feels really confusing and unintuitive to me.

At the very least, it seems wrong to accept an optional `www.` at the beginning, given these are explicitly *not* web addresses.

hydroper:
@17cupsofcoffee But XML, for example, similiarly assigns URIs to namespaces.

slanterns:
> @17cupsofcoffee But XML, for example, similiarly assigns URIs to namespaces.

"XXX do something" does not simply imply we can copy the same choice into Rust. XML is not a package manager so it faces different challenges with cargo, for example, it does not need to deal with package imposters.

You need to convince the users that the proposal is indeed beneficial if you want to push it further. I personally don't see enough advantage to make the extra complexity and confusion worthy (over the current situation), nor does it surpass https://github.com/rust-lang/rfcs/pull/3243.

hydroper:
I want to add that NPM also allows you to use multiple dots in a package name. You can publish anything like `com.qux.f` to NPM.

slanterns:
> I want to add that NPM also allows you to use multiple dots in a package name. You can publish anything like `com.qux.f` to NPM.

So, I'm still unsure about what is the real benefit of allowing dot in package name / forcing the package namespace to be an URL-like string. Can you add more explanation?

hydroper:
> > I want to add that NPM also allows you to use multiple dots in a package name. You can publish anything like `com.qux.f` to NPM.
> 
> So, I'm still unsure about what is the real benefit of allowing dot in package name / forcing the package namespace to be an URL-like string. Can you add more explanation?

All a readability matter.

jhpratt:
As I have stated multiple times before, what benefit does this provide over plain identifiers for namespacing? The motivation is not a motivation at all — it's a statement about namespacing in general.

You say that this is a domain name, but there is no effort to verify ownership. So how is it a domain name and not just a sequence of arbitrary characters? Does the domain even need to be registered? Are TLDs validated, or can I register `foo.example`? How do you plan on solving the issue of implied ownership of a domain? Surely you don't want some random person registering rust-lang.org and potentially doing something malicious with it. Or do you and you just haven't stated that?

With regard to the "crates.io domains" section, you mention removing domains. What happens when a domain is removed? You also mention that `cargo` can be used to transfer domains and manage the owners of a domain. How? `cargo` can't do this for crates at the moment — you have to use crates.io.

Leaving the drawbacks and unresolved questions sections blank is a non-starter. There are clear drawbacks and questions that have been posted multiple times on IRLO. Ignoring them is not an option.

hydroper:
I can't answer all questions at all because things like "trust" aren't covered by the proposal. Even though I wanted much that feature, I'll close it because sadly Rust developers don't have the same sense I've. People are forgetting that NPM does almost the same. (There are NPM packages using identifiers such as `com.qux.f`.)

jhpratt:
I will note that prior art is insufficient reason by itself. I'll leave it at that as you've closed the PR.

PR #3386, opened 2023-02-08
Remove the 'kind' field from Item

mplsgrant:
The 'kind' field is possibly a duplicate of the 'kind' field in Item Summary.

Also, the presence of the 'kind' field in Item does not match with the relevant source:

https://github.com/rust-lang/rust/blob/b082e80e20475b1ec5b0bd0dd1dac3e6759c8022/src/rustdoc-json-types/lib.rs#L65

[Rendered](https://github.com/mplsgrant/rust-lang-rfcs/blob/mpls-2963-item-kind/text/2963-rustdoc-json.md)


mplsgrant:
Misunderstood the struct. Please disregard.

PR #3387, opened 2023-02-08
fix link message typo in rfc-2789

wangkirin:
when I refer to the sparse index rfc, I found the tracking issue item has the right link ,  but the error link message, this PR fix it.

[Rendered](https://github.com/wangkirin/rfcs/blob/patch-1/text/2789-sparse-index.md)


ehuss (approved):
Thanks!

PR #3394, opened 2023-02-23
format-dynamic-fill

ModProg:
[Rendered](https://github.com/ModProg/rfcs/blob/format-dynamic-fill/text/3394-format-dynamic-fill.md)


text/0000-format-dynamic-fill.md
```diff
@@ -0,0 +1,65 @@
...
+[motivation]: #motivation
+
+To allow changing the *fill* character at run-time. Due to not being able to create or modify [`Formatter`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html), this is the only way to allow this.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Extending on [`std::fmt`#Fill/Alignment](https://doc.rust-lang.org/std/fmt/index.html#fillalignment):
+
+The value for the fill can also be provided as a [`char`](https://doc.rust-lang.org/std/primitive.char.html) in the list of parameters by adding a postfix `$` indicating that the second argument is a char specifying the fill.
```
joshtriplett, line 21:
```suggestion
The value for the fill can also be provided as a [`char`](https://doc.rust-lang.org/std/primitive.char.html) in the list of parameters by adding a postfix `$` indicating that the next argument (or the named argument if named) is a char specifying the fill.
```

text/0000-format-dynamic-fill.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It complicates the format spec, and it is unclear how much it might be used.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- If enforcing a minimum of 2 characters the `$` could be omitted, but this only makes this easier to mess up and is unnecessarily restrictive.
```
joshtriplett, line 51:
```suggestion
- If enforcing a minimum of 2 characters the `$` could be omitted, but this only makes this easier to mess up and is unnecessarily restrictive. This would also be inconsistent with other parameters that do require `$`.
```

text/0000-format-dynamic-fill.md
```diff
@@ -0,0 +1,65 @@
...
+- This can only be implemented in rust as it is impossible to drive the `Display` family traits on stable without going through the format string.
```
joshtriplett, line 52: I'm not sure what this is trying to say. Could you clarify it?

ModProg: :52? `Display::fmt` cannot be used directly on stable because `Formatter` cannot be instantiated AFAIK.

joshtriplett: The part that's confusing me is "This can only be implemented in rust"; I don't know what that's supposed to mean here.

ModProg: well in the template it asks whether this could be done as a library.

text/3394-format-dynamic-fill.md
```diff
@@ -0,0 +1,65 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It complicates the format spec, and it is unclear how much it might be used.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- If a minimum of 2 characters was enforced the `$` could be omitted, but this only makes this easier to mess up and is unnecessarily restrictive, as well as inconsistent with other parameters that do require `$`.
+- This can only be implemented in rust as it is impossible to drive the `Display` family traits in stable rust without using a format string.
```
joshtriplett, line 52: Attempting to make this more clear:

```suggestion
- This cannot be implemented in a library, as the `Formatter` used by the `Display` trait does not expose a mechanism to change the fill character, and any other mechanism would not allow using the `Display` implementations throughout the ecosystem.
```

m-ou-se:
This seems relatively straightfoward. It adds a little bit overhead to std::fmt::write to be able to support this feature, but that can probably be kept to a minimum.

---

For context, there are currently eight formatting options: width, precision, fill, alignment, sign, alternate, zero padding and debug hex. Only the first two can be picked dynamically. With the proposed change, the first three become dynamic:

| format options | type | dynamic (now) | dynamic (proposed) |
|-|-|-|-|
| width | `usize` | :heavy_check_mark: | :heavy_check_mark: |
| precision | `usize` | :heavy_check_mark: | :heavy_check_mark: |
| fill | `char` | :x: | :heavy_check_mark: |
| alignment | `enum` (none, `<`, `>`, `^`) | :x: | :x: |
| sign | `enum` (none, `+`, `-`) | :x: | :x: |
| alternate | `bool` (none, `#`) | :x: | :x: |
| zero pad | `bool` (none, `0`) | :x: | :x: |
| debug hex | `bool` (`?`, `x?`) | :x: | :x: |

There is some overhead involved in allowing an option to be dynamic, so I think it makes sense to only do this for the `usize` and `char` options, and leave the boolean/enum flags as they are, since it's easy to turn that into a simple `if` if one needs to e.g. select the sign or alignment dynamically.

m-ou-se:
> Due to not being able to create or modify Formatter, this is the only way to allow this.

Perhaps we should (also) add a way to create a Formatter.

m-ou-se:
@rfcbot merge

ModProg:
> Perhaps we should (also) add a way to create a Formatter.

I asked that on zulip as well: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Formatter.20construction.20without.20need.20for.20hardcoded.20string/near/338148530

ModProg:
> since it's easy to turn that into a simple `if` if one needs to e.g. select the sign or alignment dynamically.

Yes, but turns out, doing that for all combinations will still create a bunch of code (and actually make this otherwise very small crate, slower to compile than I had hoped) https://github.com/ModProg/interpolator/blob/577d7e4365818466203ad94a4228bfd4b1aff512/src/fmt/display.rs

dtolnay:
Are there examples of a use case in which this would be used for real?

The only example invocation in the RFC is with non-dynamic fill characters (`'+'` and `'-'`) which is presumably not how anybody would want to use this.

@rfcbot concern needs more compelling motivation

ModProg:
> Are there examples of a use case in which this would be used for real?
> 
> The only example invocation in the RFC is with non-dynamic fill characters (`'+'` and `'-'`) which is presumably not how anybody would want to use this, since that is supported without this feature.

My use case is implementing a runtime version of `format` with https://docs.rs/interpolator/. A use case I could imagine is something like having a setting that switches a program's output between aligning with a visible placeholder `.` and an invisible ` `.

For my use case this isn't the ideal solution (I'd prefer being able to create a `Formatter` directly, e.g. with a builder), but it is the only one I can currently foresee being feasibly implemented.

dtolnay:
I feel that the `interpolator` use case of "runtime version of `format`" is worthwhile for the standard library to make possible in some form. But beyond the exact use case of `interpolator`, it is not clear to me that building dynamic fill into the formatting syntax has any demand.

As I see it:

- Even if dynamic fill were supported in formatting syntax, `interpolator` use case still would require individually listing all 2112 possible format configurations (e.g. `{value:>-#0x?}`), so it would still strongly prefer custom Formatter construction over this RFC. Custom Formatter construction would dramatically simplify the `interpolator` implementation and eliminate compile time cost of compiling that many format args.

- As soon as Formatters become constructible, it seems likely that literally nobody would ever use dynamic fill ever again, other than maybe by accident while they are trying to express something else in the convoluted formatting syntax.

So I would say let's skip this feature and directly go to Formatters that can be populated with a dynamic set of configuration _without_ using format_args syntax.

@rfcbot concern expose Formatter construction instead of this

Directly adding `&mut` methods to Formatter is probably unwise:

```rust
impl<'a> Formatter<'a> {
    pub fn set_fill(&mut self, fill: char) {
        self.fill = fill;
    }
}
```

That is how C++ iostream / iomanip work (https://en.cppreference.com/w/cpp/io/manip). For example `std::cout << std::setfill('*')` changes the fill character for all subsequent writes to the same stream.

Instead maybe this kind of API:

```rust
// core::fmt

impl<'a> Formatter<'a> {
    pub fn with_fill<'b>(&'b mut self, fill: char) -> Formatter<'b>
    where
        'a: 'b,
    {
        Formatter {
            fill,
            flags: self.flags,
            ...
            buf: &mut *self.buf, // reborrow with shorter lifetime
        }
    }
}
```

```rust
// user code

struct DynamicFill<D> {
    fill: char,
    delegate: D,
}

impl<D: Display> Display for DynamicFill<D> {
    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        Display::fmt(&self.delegate, &mut formatter.with_fill(self.fill))
    }
}
```

or a more general builder:

```rust
impl Formatter<'_> {
    pub fn builder() -> FormatterBuilder;
}

impl FormatterBuilder {
    pub fn fill(&mut self, fill: char) -> &mut Self;
    pub fn build(&mut self, /*a `&'a mut dyn Write` arg goes either here or builder()*/) -> Formatter<'a>;
}
```

```rust
let mut buf = String::new();
let formatter = fmt::Formatter::builder()
    .fill(ch)
    .build(&mut buf);
thing.fmt(&mut formatter)?;
```

thomcc:
One case where I've wanted dynamic fill is implementing a C printf by mapping as much as possible to rust fmt, but that's pretty niche (I also don't have the code on hand). It would also be well-supported by @dtolnay's suggestion.

ModProg:
Yeah, this RFC arose more or less of the feeling that getting access to formatter construction/modification is something that will probably not happen in the near future, and this seamed more achievable.

I'm very much in favor of having access to the formatter (although this would mean I wasted a lot of time implementing interpolator the way I did :smile: )

m-ou-se:
In that case, it seems Formatter construction is the preferred route. So let's go in that direction instead. :)

@rfcbot cancel

ModProg:
> In that case, it seems Formatter construction is the preferred route. So let's go in that direction instead. :)

If there is a path forward for that, I'll happily agree.

Nemo157:
I also would love runtime setting of `Formatter` parameters, I currently have ~10kLoC of match converting args into `format_args!` calls, and that can't even handle the `fill` argument currently (though I guess I could have implemented support if this RFC was implemented by just doubling that match up).

dtolnay:
@rfcbot fcp close

Next step: we can equate https://github.com/rust-lang/rfcs/pull/3394#issuecomment-1563263573 to an approved ACP, so next step would be prototype it in a PR.

_Maybe_ RFC if we need to argue about whether a builder is needed. My preference is no separate builder type if possible.

PR #3400, opened 2023-03-13
Add a count parameter to the expect attribute

JarredAllen:
This RFC proposes extending the `expect` attribute` from https://github.com/rust-lang/rfcs/pull/2383

### DIFF
--- a/text/0000-expect-attribute-count-parameter.md
+++ b/text/0000-expect-attribute-count-parameter.md
@@ -1,0 +1,188 @@
+- Feature Name: Add a count parameter to the expect attribute
+- Start Date: 2023-03-13
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Add a new paramter to [the `#[expect]`
+attribute](https://github.com/rust-lang/rust/issues/54503) named `count` that asserts a
+precise count for the number of lints suppressed by the attribute. If the number
+of lints to suppress is given, then it will raise a warning unless that number
+is exactly matched.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


xFrednet:
Hey and thank you for the RFC! Currently, there is a discussion in the tracking issue https://github.com/rust-lang/rust/issues/54503 and on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Expect.20attribute.20mental.20model/near/340449104) what lint emission should actually be suppressed and fulfill the lint expectation. I would recommend putting this RFC on hold, until a decision was made in that regard.

text/0000-expect-attribute-count-parameter.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Behavior
+
+I am unaware of any languages which have a similar functionality to what I am
+describing here.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Can we allow `#[expect(..., count = 0)]` with some useful behavior? What about
```
teor2345, line 187: `expect(lint, count = 0)` seems like it should be the same as `deny(lint)`.

oskgo: Agreed. I could see `count = 0` being convenient for a tool that decreases the expected count when it is too high. Such a tool would then not have to deal with this edge case.

In my mind `expect(lint, count = 0)` should compile with a warning (or clippy lint) suggesting using `deny` instead.

JarredAllen:
> Hey and thank you for the RFC! Currently, there is a discussion in the tracking issue [rust-lang/rust#54503](https://github.com/rust-lang/rust/issues/54503) and on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Expect.20attribute.20mental.20model/near/340449104) what lint emission should actually be suppressed and fulfill the lint expectation. I would recommend putting this RFC on hold, until a decision was made in that regard.

The decision has been reached and the attribute has been stabilized, so can we un-hold this RFC? I'd still love to see this feature happen.

PR #3405, opened 2023-03-30
Add two newly created tracking issues to 2788-standard-lazy-types

slanterns:
To better reflect the current situation.
[Rendered](https://github.com/slanterns/rfcs/blob/2788-add/text/2788-standard-lazy-types.md)


ehuss (approved):
Thanks!

PR #3422, opened 2023-04-20
Blueprints

DasLixou:
I am not that experienced with RFCs, feedback welcome 😀

[Rendered](https://github.com/DasLixou/rust-rfcs/blob/DasLixou-blueprints/text/3422-blueprints.md)

### DIFF
--- a/text/0000-blueprints.md
+++ b/text/0000-blueprints.md
@@ -1,0 +1,195 @@
+- Feature Name: `blueprints`
+- Start Date: 2032-04-20
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+
+# Summary
+[summary]: #summary
+
+Adding blueprint "functions" which will be directly used instead of called.
+
+# Motivation
+[motivation]: #motivation
+
+When looking though rust code the other day, especially looking at all those `new` functions which just create the type without any other or just a few expressions with it, i wondered how this would look in assembly. So I wrote a little experiment in godbolt and saw that a new function described as below results in a function with a call:
+```rust
+struct Test(u32);
+
+impl Test {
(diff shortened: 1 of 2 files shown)


shepmaster:
> So I wrote a little experiment in godbolt and saw that a new function described as below results in a function with a call

Did you compile with optimizations? `-C opt-level=2`. When I add that, it produces

```
example::main:
        ret
```

DasLixou:
> > So I wrote a little experiment in godbolt and saw that a new function described as below results in a function with a call
> 
> Did you compile with optimizations? `-C opt-level=2`. When I add that, it produces
> 
> ```
> example::main:
>         ret
> ```

Because it is never used. But a good point... I'll look how it behaves when using optimization

DasLixou:
Yeah, you are right. It gets optimized away even without a mention of inline. But i think in some cases it's still good to have it explicitly declare that you want to blueprint it

shepmaster:
> But i think in some cases it's still good to have it explicitly declare that you want to blueprint it

That's the point of the `inline` attribute. How is `#[blueprint]` any different?

DasLixou:
> > But i think in some cases it's still good to have it explicitly declare that you want to blueprint it
> 
> That's the point of the `inline` attribute. How is `#[blueprint]` any different?

Oh. I was confused by inline and inline(always) and that inline didnt do anything so i thought it was for something else.

So should i close this?

SOF3:
As you said in the RFC, `inline(always)` already does the trick. So what is the difference from `inline(always)`?

Furthermore, inlining is not always cost-free. Increasing the size of a function could increase cache miss on (especially low-end) CPUs and result in worse performance for inlining cold paths.

DasLixou:
Ok then ill close this

PR #3432, opened 2023-05-15
Fix some mentions of `--remap-path-prefix` that should be `--remap-path-scope`

Nemo157:
cc @cbeuw 

[Rendered](https://github.com/Nemo157/rfcs/blob/fix-path-prefix-vs-scope/text/3127-trim-paths.md)


cbeuw (approved):

ehuss (approved):
Thanks!

PR #3433, opened 2023-05-21
Fix typos in RFC 0135

sekunho:
- Remove extra "is"
- Add missing closing angle brackets

[Rendered](https://github.com/sekunho/rfcs/blob/fix-0135-typos/text/0135-where.md)


ehuss (approved):
Thanks!

PR #3435, opened 2023-05-23
RFC: `#[export]` (dynamically linked crates)

m-ou-se:


text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+including all relevant type descriptions.
+
+To avoid extremely long symbol names that contain a full (encoded) version of the function signature
+and all relevant type descriptions, we use a 128-bit hash based on all this information.
+
+For example, an exported item in `foo::bar` in the crate `mycrate` would be exported with a symbol name such as:
+
+```
+_RNvNtC_7mycrate3foo3bar_f8771d213159376fafbff1d3b93bb212
+```
```
bjorn3, line 263: This doesn't seem to be able to handle multiple semver incompatible versions of the same crate. We currently include the `StableCrateId` in the symbol name for this which includes all `-Cmetadata` arguments. Cargo uses a separate `-Cmetadata` value for every crate version.

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+only the signatures of its exported items are necessary.
+Cargo will pass a flag to the Rust compiler which will stop it from generating
+code for non-exported items and function bodies.
+
+A clear separation between "public dependencies" (which used in the interface)
+and "private dependencies" (which are only used in the implementation) is required
+to avoid building unnecessary indirect dependencies.
+A system for that has been proposed in [RFC 1977](https://rust-lang.github.io/rfcs/1977-public-private-dependencies.html).
+
+### Name Mangling and Safety
```
bjorn3, line 233: What will the format of the crate metadata used to lookup symbol names and function signatures be? Will it be some stable binary format or a rust like syntax that users can directly write? The former is faster, but the later is easier to avoid breaking across rustc versions and makes it easier to prevent the user accidentally breaking the ABI by requiring the interface file to be modified for this to happen. Also will dylibs embed the crate metadata as a whole, just a hash to check compatibility or not at all.

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+In this case, using the type as part of a function signature will not result in a hash based on the full (recursive) type definition,
+but will instead be based on the user provided hash.
+
+### Standard Library
+
+Once the ["crabi"](https://github.com/rust-lang/rust/pull/105586) feature has progressed far enough,
+we should consider adding `#[export]` attributes to some standard library types, effectively committing to a stable ABI for those.
+For example, `Box`, `Vec`, `Option`, `String`, `NonZero`, `File`,
+and many others are good candidates for `#[export(unsafe_stable_abi)]`
+(if the "crabi" ABI doesn't already handle them specially).
```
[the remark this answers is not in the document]
m-ou-se: Yeah that's a separate problem that falls in the scope of "crabi" to solve. (Perhaps in a similar way as it will handle trait objects.) See also the section on that later in this RFC.

I've updated this RFC to not mention Box/Vec/String here to avoid confusion.

sdroege:
This generally looks great and useful, but there's one aspect that seems a bit strange to me. From my understanding, to link to such a shared crate you would need to have either (a version of) its source code available or some kind of types-and-functions-without-function-bodies version of it (i.e. something like the Rust version of a C header file).

Did you consider including the API of the crate in a way that rustc can understand inside the shared library, e.g. in a separate section, or as a separate interface definition file? That way the shared library build results would be all that is necessary to link to the library, and for example cargo (and other build systems) could later get a mechanism to look for system installed shared libraries in that format and directly make use of them.

text/0000-export.md
m-ou-se: The exact hashing algorithm is something to be designed later (hopefully in an incremental way so we can support e.g. `i32` before we support more complicated types).

Ideally it should be specified in a way that it can also be generated by other tools that are not rustc.

Dynamic libraries will include these hashes in the symbol names. Including more information can be useful for debugging tools, but is not necessary. (See the "future possibilities" section.)

m-ou-se:
@sdroege That's certainly a possibility and not incompatible with this RFC. The "future possibilities" section already mentions the possibility of adding more information in a separate (optional / debug) section, and what you're proposing seems to be a form of that.

> something like the Rust version of a C header file

Yes, I'm basically proposing that we can have `#[export]` attributes rather than a separate header file (like C) or a bindings crate (Rust today).

sdroege:
Sounds good to me, thanks for the fast response :) Maybe you could extend the "A tool to create a stripped, 'dynamic import only' version [...]" item in the "future possibilities" section with a sentence that this could also be as a binary format as part of a shared library section or so.

Such a mechanism would also be useful to allow non-Rust languages to link to such shared libraries without having to parse Rust code or requiring tools for converting the API definitions between any possible source and target language.

text/0000-export.md
m-ou-se: I'm not sure if we can properly support that. If e.g. `libmycrate.so` is separately compiled, we cannot (and should not) make the symbols dependent on the crate id it will have in a dependency tree of whatever binary will use this library later.

Perhaps we want to include just the major version in the symbol names, but not everyone uses versions (a cargo concept) or follows semver, nor does the version of the public *API* need to match the public of the stable *ABI*.

Maybe we need an crate-global attribute to optionally specify the ABI version (or even just a random hash) for the crate, to allow for symbols with identical paths and signatures of multiple identically named crates to exist at once, although I imagine most use cases would not use e.g. two different versions of `librustls.so` at the same time. (If that's a supported use case, they should probably be called `librustls1.so` and `librustls2.so`, using `rustls1` and `rustls2` as the name used in the symbols rather than just `rustls`.)

bjorn3: The `StableCrateId` does not depend on any dependencies. It only depends on the crate name, if it is a lib or bin and all `-Cmetadata` arguments (in sorted order). The SVH (crate hash) does depend on dependencies. In addition the current cargo impl hashes dependency versions into the `-Cmetadata` argument it passes, but that can easily be changed to only pass the major part of the semver version.

kornelski:
How about flipping the default and exporting all the same symbols as static linking does, and provide an opposite attribute to opt out? (`#[unstable_abi]` or such)

I imagine once Rust ABI situation matures, users will want to support both kinds of linkage without having to manage the public API twice, so the explicit opt in useful initially may be a liability long term. Similarly Rust was conservative with `#[must_use]` which causes clutter and is inconsistently applied, unlike Swift's `discatdableResult`.

text/0000-export.md
m-ou-se: That could work, but I'm not convinced the API version and the ABI version should be treated as one and the same. Renaming structs or non-exported pub functions/methods etc. etc. could result in an incompatible API while preserving a compatible ABI.

As I mentioned above, it seems a weird situation if a binary somehow ends up depending on two different `librustls.so`, considering they'd both have to have the same file name. (And if those are called `librustls1.so` and `librustls2.so`, then the `rustls1` and `rustls2` names are probably what should be used in the symbol names rather than just `rustls`. And that could just be a build option to cargo or a property in Cargo.toml.)

m-ou-se:
> I imagine once Rust ABI situation matures, users will want to support both kinds of linkage without having to manage the public API twice

I think that's inevitable. Basically `pub` is about an API commitment, e.g. `File::open(path)` always existing, while `#[extern]` is about an ABI commitment, e.g. `File` always being a wrapper around a file descriptor. We can't really change `pub` to suddenly also include the ABI commitment, which includes committing to not changing any invariants about the private fields.

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+For example, the private fields of a `Vec` are assumed to represent a valid "owned" pointer to an allocation together with its capacity and initialized size.
+
+If it would be possible to define a identically named type with the same fields but different (or no) invariants/assumptions,
+or just change the invariants in an existing library,
+it'd be possible to cause undefined behavior by loading the "wrong" dynamic library.
+
+Therefore, we can't allow a regular `#[export]` attribute on a type with private fields,
+since we have no way of automatically determining the invariants / unsafe assumptions about private fields.
+
+Instead, for these types, we must require the user to *unsafely* commit to
```
jamesmunns, line 320: One thing I didn't see mentioned here is whether the *size* of the type is considered. If a type ADDS a private field, this could (potentially) invalidate assumptions when passing by value.

For example if a type starts as `struct A { pub x: i32, y: u32 }`, and someone marks it as unsafe stable, can we help them at all if they make it `struct A { pub x: i32 }` or `struct A { pub x: i32, y: u32, z: u8 }`?

m-ou-se: I've clarified the RFC to mention this.

jamesmunns: Got it! I agree that having a way of forcing "by reference only" is good! I was more focused on "they've opted-in to unsafe snapshotting", but they *do* change the size, which is always going to cause UB on use. I'm unsure if it is still possible to verify the new size and the claimed size from only the hash, it just might be worth also explicitly stating the size or requiring export opaque to help avoid silent corruptions.

e.g. requiring one of:

```rust
// maybe size could also be `{ 2 * size_of::<usize>() }` or something else const for portability
#[export(unsafe_stable_abi = "ca83050b302bf0644a1417ac3fa6982a", size = 8)]
```

or

```
#[export(unsafe_stable_abi = "ca83050b302bf0644a1417ac3fa6982a", opaque)]
```

or having it possible to "decode" the size from the abi tag (e.g. part of the hash AND stored as plaintext somehow)

m-ou-se:
@sdroege Thanks! Updated!

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+### Importing Exported Items
+
+Normally, when using a crate as a dependency, any `#[export]` attributes of that crate have no effect
+and the dependency is statically linked into the resulting binary.
+
+When explicitly specifying `dynamic = true` for the dependency with `Cargo.toml`,
+or when using a `extern dyn crate …;` statement in the source code,
+only the items marked as `#[export]` will be available and the dependency will be linked dynamically.
+
+### Building Dynamic Dependencies
```
Amanieu, line 221: How does Cargo handle the case where a crate is used as a dynamic dependency in one place in the crate graph, but as a normal dependency in another place? I presume in that case it is still compiled as a normal dependency, but the crate declaring it as a dynamic dependency is still restricted to only using exported items.

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+
+```rust
+//! application crate
+
+fn main() {
+    library::hello();
+}
+```
+
+The library can then be either linked statically or dynamically, by informing cargo of the choice:
```
Amanieu, line 138: I don't think static/dynamic is the correct distinction to make here. I can see a use case for statically linked dependencies that require a stable ABI: proprietary libraries. This is quite common on Windows with C/C++ libraries.

text/0000-export.md
m-ou-se: Yeah maybe. I don't think we should do `size = 8`, but maybe we should require either `sized` or `opaque`. But I'm hoping to leave opaque types to a future proposal for now.

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+Swapping the `x` and `y` fields result in surprises and bugs and shouldn't be done,
+but it won't result in undefined behaviour, since any Rust code can swap the fields without using `unsafe`.
+
+However, for public fields, the field names are already part of the stable API, so we include them in the hash as well.
+
+It is an error to use a plain `#[export]` attribute on a type with out stable `#[repr(…)]`,
+if it has any private fields,
+or if any of the fields are not of an `#[export]`ed or builtin type.
+
+#### Types with Private Fields
```
Amanieu, line 306: I think that dealing with private fields is very complex and probably not something that many people can commit to. I would rather see a system where a type with private fields is treated as an opaque `extern type` (#1861). This effectively means that you can only interact with this type by reference and all operations on it must be done through exported functions.

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+However, making a (perhaps accidental) change to a function signature or type
+should not lead to undefined behavior at runtime.
+
+There are two ways to solve this problem:
+
+1. Make it the responsibility of the user.
+2. Make it the responsibility of the loader/linker.
+
+Option (1) simply means making everything `unsafe`, which isn't very helpful to the user.
+Option (2) means the loader (the part that loads the dynamic library at runtime) needs to perform the checks.
```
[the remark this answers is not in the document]
m-ou-se: Yes, I used "loader" as synonym for "dynamic linker" (e.g. `/lib64/ld-linux-x86-64.so.2` or w/e your system uses). The part of the operating system that loads dynamic libraries and resolves symbols when executables are loaded.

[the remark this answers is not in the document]
jtracey: Yeah they're typically the same file in practice, I just mean to ask whether libraries would be linked/loaded before or after `.init`.

text/0000-export.md
m-ou-se: I agree with that for many use cases, especially once "crabi" supports dyn trait objects. But for types like NonZero or File, we should be able to allow them to be used by value, since the invaraint / private details will likely not change. (We can still pick a new hash if File does change, but "it's just a file descriptor / handle" will be true for the forseeable future.)

I do agree this is a feature that should be less often than opaque dyn-like objects or fully public types, which is why I think it's fine that exporting a type with private fields requires an unsafe attribute, steering people away to safer alternatives for most situations.

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+With the proposal below, one only needs to add an `#[export]` attribute, and give the function a stable ABI
+(e.g. `extern "C"` or (in the future) `extern "crabi"`):
+
+```rust
+//! library crate
+
+#[export]
+pub extern "C" fn hello() {
+    println!("Hello!");
+}
```
obi1kenobi, line 127: Would going from
```rust
#[no_mangle]
pub extern "C" fn some_unique_name_for_hello() {}
```
to
```rust
#[export]
pub extern "C" fn hello() {}
```
always be a breaking change, or would there be a way to avoid it?

In other words, is it possible to make things "work out" between the `some_unique_name_for_hello` name and the mangling done by `#[export]`?

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+When explicitly specifying `dynamic = true` for the dependency with `Cargo.toml`,
+or when using a `extern dyn crate …;` statement in the source code,
+only the items marked as `#[export]` will be available and the dependency will be linked dynamically.
+
+### Building Dynamic Dependencies
+
+When using `dynamic = true` for a dependency, there is no need to build that full crate:
+only the signatures of its exported items are necessary.
+Cargo will pass a flag to the Rust compiler which will stop it from generating
+code for non-exported items and function bodies.
```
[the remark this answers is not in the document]
m-ou-se: It depends a bit on the use case what makes most sense as the default. E.g. I wouldn't expect cargo to start building `libjsonparser.so` if we expect that library file to already exist on the target system. (Just like how we don't attempt to build `libc.so` or `kernel32.dll` or similar.)

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+
+The `#[export]` attribute can be applied to any public item that is *exportable*.
+Which items are *exportable* is something that can increase over time with future proposals.
+Initially, only the following items are *exportable*:
+
+- Non-generic functions with a stable ABI (e.g. `extern "C"`)
+  for which every user defined type used in the signature is also marked as `#[export]`.
+  - This includes type associated functions ("methods").
+- Structs/enums/unions with a stable representation (e.g. `repr(i32)` or `repr(C)`).
+- Re-exports of those items (`use` statements, `type` aliases).
```
obi1kenobi, line 176: Two questions about edge cases here:
- For `#[export] pub struct Foo;` is the `Foo` _value_ (not the _type_) also exported? My read based on this is "no" and that might be a surprising difference between using a Rust crate as a static vs dynamic dependency.
- For `#[export] pub struct Foo();` is the implicit Ctor also exported? My expectation based on this is "yes" but it would be great to clarify explicitly.

These are on my mind because they were a problem recently in https://github.com/rust-lang/rust/issues/111338 and https://predr.ag/blog/breaking-semver-in-rust-by-adding-private-type-or-import/

text/0000-export.md
```diff
@@ -0,0 +1,398 @@
...
+  such that their size is not a stable ABI promise.
+- Exportable `static`s.
+- Exportable `trait`s, for e.g. dynamic trait objects. (See also https://github.com/rust-lang/rust/pull/105586.)
+- A tool to create a stripped, 'dynamic import only' version of the crate source code,
+  with only the exported items, without the function bodies.
+- Allow exporting two identically named items to create a dynamic library that is backwards compatible with an older interface,
+  including both a symbol for the old and new interface.
+- Next to the hash of the type information,
+  additionally and optionally include the full type information in an extra section,
+  to allow for (debug) tools to accurately diagnose mismatched symbol errors.
```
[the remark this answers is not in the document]
m-ou-se: I wrote a whole whiteboard full on notes on how something like that might work, but it quickly became more complicated than I was prepared for. I'll add it to the future possibilities to explore at some later point. :)

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+For types, the `#[export]` attribute represents the commitment to keep the representation of the type stable.
+(To differentiate from, for example, a `#[repr(i32)]` that only exists as an optimization rather than as a stable promise.)
+
+For functions, the `#[export]` attribute will make the function available from the dynamic library
+under a stable "mangled" symbol that uniquely represents its crate and module path *and full signature*.
+(More on that below.)
+
+For aliases of functions, an `#[export]` attribute on the `use` statement will use the
+path (and name) of the alias, not of the original function.
+(So it won't 'leak' the name of any (possibly private/unstable) module it was re-exported from.)
```
obi1kenobi, line 190: Can one apply `#[export]` to glob re-exports? This may or may not be desirable, and I might argue for "no" here.

It's been a serious problem in https://github.com/rust-lang/rust/issues/111336 https://github.com/rust-lang/rust/issues/111338 and https://predr.ag/blog/breaking-semver-in-rust-by-adding-private-type-or-import/

If it ends up being supported here, it might be a footgun and might need lints to avoid bad edge cases like in the links above.

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Privacy
+
+It is an error to export an item that is not public, or is part of a non-public module.
+The set of exported items of a crate will always be a subset of the crate's public interface.
```
obi1kenobi, line 195: Are public but `#[doc(hidden)]` items exportable?

My expectation is "no" and it might be worth calling it out explicitly here.

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+Initially, only the following items are *exportable*:
+
+- Non-generic functions with a stable ABI (e.g. `extern "C"`)
+  for which every user defined type used in the signature is also marked as `#[export]`.
+  - This includes type associated functions ("methods").
+- Structs/enums/unions with a stable representation (e.g. `repr(i32)` or `repr(C)`).
+- Re-exports of those items (`use` statements, `type` aliases).
+
+An `#[export]` attribute can also be applied to a crate, module, and non-generic type `impl` block,
+which is simply equivalent to applying the attribute to every public item within it.
```
shepmaster, line 179: I asked for clarification offline so copying it here...

If a public item is not exportable for whatever reason, this would generate an error. For example, this code would fail to compile:

```rust
#[export]
mod foo {
    pub fn can_be_exported_in_original_implementation() {}

    pub fn can_not_be_exported_in_original_implementation() {}
}
```

This prevents an author from accidentally attempting to mark something as exported and then a future revision to the export feature actually makes it exported.

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+and all relevant type descriptions, we use a 128-bit hash based on all this information.
+
+For example, an exported item in `foo::bar` in the crate `mycrate` would be exported with a symbol name such as:
+
+```
+_RNvNtC_7mycrate3foo3bar_f8771d213159376fafbff1d3b93bb212
+```
+
+Where the first part is the (mangled) path and name of the item,
+and the second part is the hexadecimal representation of a 128 hash of all relevant signature and type information.
```
obi1kenobi, line 266:
```suggestion
and the second part is the hexadecimal representation of a 128-bit hash of all relevant signature and type information.
```

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+(To differentiate from, for example, a `#[repr(i32)]` that only exists as an optimization rather than as a stable promise.)
+
+For functions, the `#[export]` attribute will make the function available from the dynamic library
+under a stable "mangled" symbol that uniquely represents its crate and module path *and full signature*.
+(More on that below.)
+
+For aliases of functions, an `#[export]` attribute on the `use` statement will use the
+path (and name) of the alias, not of the original function.
+(So it won't 'leak' the name of any (possibly private/unstable) module it was re-exported from.)
+
```
obi1kenobi, line 191: Consider throwing in a quick mention that removing `#[export]` is considered a major breaking change since it affects the ABI. I think this is pretty clear, but will let me have an official document I can cite in `cargo-semver-checks` when we implement a semver-check for this.
```suggestion

Removing `#[export]` from an item that was previously exported is a major breaking change,
since it removes that item from the stable ABI.

```

obi1kenobi (commented):
This is great and I'm excited to see it happen! Thanks for putting in the work on it.

I added a few questions and suggestions to clarify edge cases relevant to semver and the name resolution edge cases I recently ran into. Nothing critical, should be straightforward to resolve.

text/0000-export.md
```diff
@@ -365,6 +365,8 @@ and many others are good candidates for `#[export(unsafe_stable_abi)]`
 - Next to the hash of the type information,
   additionally and optionally include the full type information in an extra section,
   to allow for (debug) tools to accurately diagnose mismatched symbol errors.
+- Some kind of `#[export_inline]` feature to for functions that will be inlined into the crate that calls it,
```
obi1kenobi, line 368:
```suggestion
- Some kind of `#[export_inline]` feature to allow for functions that will be inlined into the calling crate,
```
Just rephrasing and adding a missing word.

text/0000-export.md
m-ou-se: I think we should allow exporting the same item under multiple symbols. Maybe something like `#[export(by_path, by_path_and_hash, symbol = "a", symbol = "b")]` or something like that. I'd like to leave that as a future possibility in this RFC to keep it focussed, though. (But definitely something to consider before stabilizing `#[export]`.)

text/0000-export.md
m-ou-se: Great question. I don't have an answer to that now. I'll think about it a bit. :)

text/0000-export.md
m-ou-se: How ABI stability relates to *API* version is an interesting question. I'll have to think about that a bit.

text/0000-export.md
m-ou-se: I lean towards either 'no' or at least 'should raise a lint', but that seems to be a question to be answered before stabilization, not one that needs to be answered in detail now.

text/0000-export.md
obi1kenobi: I think there's a bad semver hazard there, because breaking changes in `#[doc(hidden)]` items are not considered semver-major for API purposes.

We might end up in one of two uncomfortable situations here:
- API and ABI adhere to different semver rules: breaking changes on `#[doc(hidden)] #[export]` items are semver-major.
- My crate that needs `libfoo-v1.so` and your crate that needs `libfoo-v1.so` actually need _mutually incompatible versions_, a situation that cannot be resolved by just using the newer of the libfoo v1 requirements and instead requires having both versions present. The analogous case for static dependencies is prohibited by cargo's resolver.

IMHO both of these violate the principle of least astonishment for the user, and I'd advocate for sacrificing a touch of flexibility to avoid an accidental footgun.

text/0000-export.md
obi1kenobi: If you'd like a sounding board to bounce ideas off of, give me a ping!

I started the `cargo-semver-checks` project so I've been trying to answer the same question coming from the opposite direction — e.g. we already treat removing `#[repr(C)]` as a major breaking change.

text/0000-export.md
m-ou-se: I think neither of those examples should result in any symbols being exported from the dynamic library, but I think in both cases the value/function should be usable in a crate that imports the library.

Dynamically linking the `Foo()` function or `Foo` value does not make sense, since they only have one possible trivial definition based on the type definition. (It's not possible for the implementation of the `Foo()` function to change as long as the type is just `struct Foo();`.)

(Perhaps we can simply think of this as a case of the 'inline export' concept described in https://github.com/rust-lang/rfcs/pull/3435#discussion_r1202571220)

obi1kenobi: Would the answer change for `#[export] pub struct Foo(pub Bar, pub Baz);` at all, assuming `Bar` and `Baz` are exported types?

I don't have an intuitive understanding of inline exports yet (specifically how they are different than normal static linking _for those items specifically_), so I wanted to clarify the boundaries on "trivial definition" — whether it's only for empty tuples or also for more complex tuple structs where things like alignment and padding might also be in play.

m-ou-se: Same answer for that case. The constructor of `Foo` does not need to be dynamically linked, if the only possible implemementation is just to fill in the fields which are already public and promised to be ABI stable.

`Foo { 0: .., 1: .. }` and `Foo(.., ..)` should just do the same. Neither should result in a function being called in a dynamic library.

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+
+Therefore, we can't allow a regular `#[export]` attribute on a type with private fields,
+since we have no way of automatically determining the invariants / unsafe assumptions about private fields.
+
+Instead, for these types, we must require the user to *unsafely* commit to
+ABI stability if they want to make the type available to exported functions.
+
+Using `#[export(unsafe_stable_abi = «hash»)]`, one can make the (unsafe) promise
+that the type will remain ABI compatible as long as the provided hash remains the same.
+The hash must have been randomly generated to ensure uniqueness (which is part of the unsafe promise).
```
str4d, line 325: As a cryptographic engineer, calling this a `hash` is confusing to me; as specified, it isn't a hash of anything, but instead is sampled randomness. I'm guessing from the naming that the intention is for this user-provided random value to take the place of the hash that a fully-public type would have in its mangled name?

Perhaps `commitment` would be a more clarifying name: for public types, the API's commitment is made via a hash of the public API, and for private-containing types, the commitment is made directly and unsafely by the user.

text/0000-export.md
```diff
@@ -0,0 +1,401 @@
...
+#[export(unsafe_stable_abi = "ca83050b302bf0644a1417ac3fa6982a")]
+#[repr(C)]
+pub struct ListNode {
+    next: *const ListNode,
+    value: i32,
+}
+```
+
+In this case, using the type as part of a function signature will not result in a hash based on the full (recursive) type definition,
+but will instead be based on the user provided hash (and the size and alignment of the type).
```
str4d, line 337: Is the intention that the regular function signature's hash (commitment) always be derived by fully recursing through the type definition? An alternative would be to structure the commitments as a tree, so the hash (commitment) of an outer type is derived from the hashes (commitments) of its inner types (plus size and alignment). Then public and private types are treated identically. This might mean more hashing overall, but the CPU work done to hash common inner types could be cached and reused across the exported API (which for a large API might actually result in less hashing overall).

programmerjake: basically a [Merkle Tree](https://en.wikipedia.org/wiki/Merkle_tree) with a tree node for each type.

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+pub extern "C" fn hello() {
+    println!("Hello!");
+}
+```
+
+```rust
+//! application crate
+
+fn main() {
+    library::hello();
```
petrochenkov, line 134: I guess my main question after seeing this example is whether this proposal requires a stable Rust metadata format, or it just works on symbol names.

How is `library::hello` resolved here?

Does `library` contain some metadata saying that the crate root module contains a name `hello` and it's a function?
That means reading Rust metadata.
Keeping metadata format compatible across rustc versions (by making it stable, or by keeping logic for reading all previous metadata versions) would put *a lot* of maintenance burden on rustc development, maybe a prohibitive amount of maintenance burden.

Or we just hash the whole `library::hello` path to obtain some string like `hello_HASH` and dlsym that string from the dynamic library?
In that case hashing the path only seems not to be enough, because it means the argument types and return type are not included into the hash.
Do we need to hash the whole call expression `library::hello()` instead? Even then we don't know the return type to hash.
Also, if the hash requires knowing types then resolution of such paths needs to be delayed until type checking, is that right?

I don't like any of these alternatives, TBH.

m-ou-se: Have you fully read the RFC? The `library::hello` path in that example is resolved no different than it is today. There is still a crate named `library` with a `hello` item. I'm not proposing to somehow enable generate symbol names out of thin air based on just the call expression. The symbol name comes from the definition of the item.

> whether this proposal requires a stable Rust metadata format, or it just works on symbol names

A metadata format is perhaps a future possibility, but definitely not part of this RFC. This RFC proposes symbols with a hash based on all information relevant for type safety. That hash does need to be stable, but can start out with strict limitations for only certain kinds of signatures and then slowly extended over time.

text/0000-export.md
m-ou-se: > I'm guessing from the naming that the intention is for this user-provided random value to take the place of the hash that a fully-public type would have in its mangled name?

Yes.

It represents the hash of all relevant memory safety information, which for types with private fields includes all safety invariants. Because those are not expressed directly in the type system and we cannot hash someone's mental model, this just uses a random number that the user promises to change when the inveriants change. :)

Happy to call it something else, but the name of this placeholder isn't really an important part of the proposal.

text/0000-export.md
m-ou-se: > Then public and private types are treated identically.

It's critical that they are not treated identically. The safety related invariants of a type with private fields are not fully described by its fields.

E.g. a `struct NonZero { value: i32 }` and `struct NonSeven { value: i32 }` should definitely not be seen as equal.

Similarly, a struct with a private field might be changed over time to have additional safety invariants without changing any of its fields.

And there are also subtle situations where a someone might change the fields of a type without changing its ABI (e.g. changing a `Option<NonNull>` to a raw pointer, changing a `T` to `Wrapping<T>`, moving all fields into an "Inner" struct, etc.).

text/0000-export.md
digama0: I don't see why this needs to be a hash or big random number at all, and it is harmful for readability and will cause all sorts of auditability problems (is this truly random?) Why can't this just be an extra salt on top of all the usual information that goes into the ABI hash for the type. Then it is effectively just "ABI_version = 4" or what have you, you should bump it when something changes and this is easy for users to understand.

text/0000-export.md
petrochenkov: >The `library::hello` path in that example is resolved no different than it is today.

Then I don't understand how can the dynamic library and the application be built by different versions of rustc ("2. Cases where dynamic library and application can be compiled and shipped separately from each other.").

To resolve paths like we do it today we need to read a lot of various Rust metadata from the dynamic library, but the metadata details typically change between every rustc release and rustc 1.N.0 won't be able to correctly parse metadata generated by rustc 1.M.0.

petrochenkov: I read the whole RFC, and I have an impression that either I'm missing something big, or it focuses in great detail on secondary issues while skipping on the primary problem with the metadata format incompatibility.

digama0: Let's say that we want to dynamically link `library` compiled by rust 1.M.0 with `main` compiled by 1.N.0 (where N > M for concreteness). My understanding is that you would build both `main` and `library` with cargo + rustc 1.N.0 while signaling that the `library` dependency is `dynamic = true`. This is basically exactly like a statically linked build process except that the code of `library` is not emitted and instead symbols using the stable ABI mangling scheme are called instead, i.e. `main` would call `_Rlibrary5hello_ABCD` or what have you.

The previous compile of the library crate by 1.M.0 also produced a function of the name `_Rlibrary5hello_ABCD` because it had the same stable ABI hash, so when these two are linked together (as one would do for an FFI binding) everything works out. The metadata itself is not shared.

There are some future work discussions about being able to skip parts of `library` while building `main` on 1.N.0 but it doesn't seem to be part of the RFC as written.

p-avital:
While I'm really happy to see some progress on the ABI front, I feel like this still doesn't address the most painful point of building shared libraries in Rust: "What if some of my dependencies didn't brand their types with `#[export]`?". We already have ways to get a stable ABI for types we fully own (C-repr, abi_stable, stabby...), but as soon as dependencies are introduced, things get real ugly real fast.

I really wish we'd put the option on the table of having the compiler "contaminate" types that are used by `#[export]`ed symbols. While this doesn't solve the "mismatching library version" part of the problem, there are existing ways to check for this problem. There are no ways to force a dependency to use a stable layout for some of its types.

text/0000-export.md
petrochenkov: Ah, I see, the "Building Dynamic Dependencies" paragraph talks about this, although in a way that is obscured by potential future optimizations.

So, we
- still need the source code for the dynamic library and all its dependencies;
- instead of doing `cargo build` on that dynamic library we are now basically doing `cargo check` on it (with rustc 1.N.0) to get its rmeta.

And the benefits we get are
- saving some time on not doing codegen;
- having opportunity to make changes to the pre-build dynamic library as long as they don't *sufficiently* diverge from the previously published source code used for generating rmeta.

All this sounds more or less reasonable.

Diggsey:
One of the main use-cases for dynamic loading of libraries is for plugin-like systems, where you might have multiple libraries exposing the same API.

This RFC does not appear to be designed to solve this use-case, but it solves several closely related problems, and it would be a shame if the plugin use-case required a totally orthogonal approach.

Is there a way this RFC could be naturally built upon to support such a feature?

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+
+Note that, technically, the names of the type and the fields are not relevant, at least _not for memory safety_.
+Swapping the `x` and `y` fields result in surprises and bugs and shouldn't be done,
+but it won't result in undefined behaviour, since any Rust code can swap the fields without using `unsafe`.
+
+However, for public fields, the field names are already part of the stable API, so we include them in the hash as well.
+
+It is an error to use a plain `#[export]` attribute on a type with out stable `#[repr(…)]`,
+if it has any private fields,
+or if any of the fields are not of an `#[export]`ed or builtin type.
```
Mark-Simulacrum, line 304: Is builtin type right? For example, `*const Foo` is stable ABI but is opaque-ish if Foo isn't export too. &str is also builtin but doesn't feel like we should commit to a stable ABI for (at least yet). Maybe worth listing the specific types here rather than naming a category?

digama0: I think this work is largely being offloaded to the "crabi" ABI RFC. Probably the best thing to say here would be "FFI-safe type" in the sense of the `improper_ctypes` lint.

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+  including both a symbol for the old and new interface.
+- Next to the hash of the type information,
+  additionally and optionally include the full type information in an extra section,
+  to allow for (debug) tools to accurately diagnose mismatched symbol errors.
+- Some kind of `#[export_inline]` feature to allow for functions that will be inlined into the calling crate,
+  rather than being part of the dynamic library, which will only be able to call exported items.
+
+## Alternatives
+
+- Alternatives for using a hash of all relevant type information:
```
Mark-Simulacrum, line 373: I think this RFC might want to talk about why this is the right choice where we previously made a different one in the move from legacy to v0 (which moved away from a hash, at least mostly). I think given the use case here this may be better, but the discussion seems worthwhile.

programmerjake: it seems like a good idea to me to combine both of them -- `v0` mangling so debuggers and similar can get the full source symbol name including generic arguments, and the hash to detect ABI breaks (ABI breaks won't necessarily change the `v0` mangling, such as reordering struct fields)

text/0000-export.md
Mark-Simulacrum: I don't think that lint is appropriate here, it has the same questions/problems with it.

digama0: I don't know why this RFC in particular would need to solve that problem though, the problem is pre-existing and this RFC isn't trying to address it.

Mark-Simulacrum: This RFC is prescribing an error, rather than a lint, and seems intended to start with a very small set and expand over time. So it seems to me that it *is* trying to do something different; it's also a *safe* interface AFAICT - so that also matters.

text/0000-export.md
m-ou-se: > still need the source code for the dynamic library and all its dependencies

Since we only need the metadata and not the function bodies, we could skip some things. Once cargo has a separation between public and private dependencies, we can skip private dependencies entirely.

text/0000-export.md
m-ou-se: Ah yes, I should clarify this section a bit. The rules I had in mind are a bit stricter than what I wrote down in the RFC. (Matching your expectations.)

text/0000-export.md
bjorn3: You need private dependencies if they are proc macros. Proc macros may expand to `#[export]` attributes or items with `#[export]` attributes.

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+
+Therefore, we can't allow a regular `#[export]` attribute on a type with private fields,
+since we have no way of automatically determining the invariants / unsafe assumptions about private fields.
+
+Instead, for these types, we must require the user to *unsafely* commit to
+ABI stability if they want to make the type available to exported functions.
+
+Using `#[export(unsafe_stable_abi = «hash»)]`, one can make the (unsafe) promise
+that the type will remain ABI compatible as long as the provided hash remains the same.
+The hash must have been randomly generated to ensure uniqueness (which is part of the unsafe promise).
```
riking, line 325: I'd like to see a brief mention of tool support here. People aren't typically familiar with the facilities to generate random numbers locally.

> You can generate an unsafe_stable_abi hash by setting it to an empty string and running `cargo fix`.

m-ou-se: I wouldn't be surprised if rust-analyzer would get some kind of "generate new random number" action for these attributes.

Producing a random number as a suggestion in the diagnostic when leaving the attribute empty (which could be used by cargo fix) seems helpful.

digama0: But why does it have to be a random number in the first place? As I mentioned before, this is terrible for code review - if the point is to avoid some kind of collision then it is susceptible to spoofing. A simple incrementing number should be sufficient for ABI versioning purposes.

m-ou-se: > A simple incrementing number or string should be sufficient for ABI versioning purposes.

That is not sufficient. It needs to be globally unique. Otherwise you'd need to (unsafely!) assume that the same (crate) name always refers to the exact same codebase with linear versioning. If a crate is forked and both forks introduce a different new safety invariant on private fields, it would go wrong if both just increment their number, resulting in the same number being used for two ABI-incompatible but identically named types.

In the Rust ecosystem there is no expectation that crate/package names are globally unique. (There exists a lot of Rust code outside of crates.io.) There is no expectation to put e.g. a domain name in the package/crate name to make them unique, like in some other languages.

digama0: This is also why I suggested a string instead of a number, it's a lot easier to put necessary disambiguating information in a string compared to a number. But I don't see how this changes the equation really, since it's unsafe either way. The rule is not "it must be globally unique", it is "this safety invariant must exactly match the safety invariant on any other ABI compatible types with the same `unsafe_stable_abi` which are linked with this crate" which is a much more manageable auditing task.

stlankes:
Maybe I misunderstood something. Why do you limit the RFC to shared libraries? Static libraries are sometimes also useful. Ok, it is perhaps only a special case. But we develop a library operating systems and link a static library as kernel to the application. Currently, the interface between application and libos is a C interface....

text/0000-export.md
clarfonthey: These arguments actually bring into question whether there should be a supported ABI version in Cargo as well. Since Cargo *already* reconciles latest-compatible updates for APIs, why shouldn't it do the same with ABIs too?

clarfonthey:
So, I have a feeling this was explicitly left out of the RFC, but it feels worth mentioning *something* about the intended loading mechanism for these dynamic crates.

I think it is (mostly implied) that `cargo run` will work out-of-the-box even with these settings, although it would be good to mention in the RFC explicitly as well. It would also be useful to clarify if this enables shared storage of dynamically-compiled crates too, although I would understand initially tightening this down and deferring that as well.

Obviously, loading dynamic code is complicated and each operating system has its own way of dealing with things. There are "system-wide" locations as well as "app-specific" locations that will be most suited further down the road. However, I'm not 100% sure how exactly those would fit into this current proposal. The most obvious mechanism currently is that used by `cargo install`, but the RFC also doesn't say anything about whether that would work with this mechanism. I don't think this should be required for this proposal, but I think it should at least have some sort of opinion on the path there, and what kinds of steps would need to be taken to design that. I don't think end users should have to operate differently to `cargo install` a crate that's dynamically linked, but perhaps this would also lead to some notion of `cargo install`ing library crates too? It all seems a bit fuzzy.

digama0:
Why would `cargo run` do anything like an install / put things in a global location? I would assume that `cargo run` on an executable with a dynamic dependency would just put the executable in `target/debug/myprog` and the `.dll` or `.so` for the library in `target/debug/deps/mylib.so`.

clarfonthey:
> Why would `cargo run` do anything like an install / put things in a global location? I would assume that `cargo run` on an executable with a dynamic dependency would just put the executable in `target/debug/myprog` and the `.dll` or `.so` for the library in `target/debug/deps/mylib.so`.

This isn't very out-there to consider when you remember that cargo will globally cache crate metadata and information from registries. What's to assume that dynamically built crates wouldn't be put there as well, since they by design could be used across different builds?

For example, I have eight different versions of the `serde` crate source in my `.cargo/registry` folder when they could have been copied into individual `target` folders, but aren't because they wouldn't change across workspaces that use them. The same could happen for dynamic library builds too.

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+### Importing Exported Items
+
+Normally, when using a crate as a dependency, any `#[export]` attributes of that crate have no effect
+and the dependency is statically linked into the resulting binary.
+
+When explicitly specifying `dynamic = true` for the dependency with `Cargo.toml`,
+or when using a `extern dyn crate …;` statement in the source code,
+only the items marked as `#[export]` will be available and the dependency will be linked dynamically.
+
+### Building Dynamic Dependencies
```
programmerjake, line 221: one issue we'll have with compiling a crate once for making a `.so` file and again for making a program that links against the `.so` is that compiling the same code multiple times isn't guaranteed to produce the same output, an extreme example is a proc macro designed to produce random numbers, which then end up as constants in the generated code and/or symbol names.

text/0000-export.md
RalfJung: > This effectively means that you can only interact with this type by reference and all operations on it must be done through exported functions.

I don't see how that helps about the private invariant issue -- if the invariants change, it is quite important that this is considered a separate type, even if everything is by-reference.

text/0000-export.md
```diff
@@ -0,0 +1,403 @@
...
+#[export(unsafe_stable_abi = "ca83050b302bf0644a1417ac3fa6982a")]
+#[repr(C)]
+pub struct ListNode {
+    next: *const ListNode,
+    value: i32,
+}
+```
+
+In this case, using the type as part of a function signature will not result in a hash based on the full (recursive) type definition,
+but will instead be based on the user provided hash (and the size and alignment of the type).
```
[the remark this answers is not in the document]
m-ou-se: That's a good point! Thanks!

Then I suppose it should still complete recursive traversal, and hash the provided hash *in addition* to that to account for the safety invariants.

I think it can still be useful to have a way to fully override the hash, in case of a change in fields/grouping/types that remains ABI compatible. (Changing some u32 to i32, regrouping fields in different inner wrapper structs, etc.) But that should be used very rarely. (I imagine we might want to use that for a few things in the standard library, but not much else.)

text/0000-export.md
m-ou-se: > which are linked with this crate

The problem is that "this crate" is meaningless. Outside crates.io, crate names aren't unique.

We could easily allow a string literal instead of a hash, such that you can write "the first field must be a valid file descriptor and the second field must be a prime number" (and hash that) rather than specifying a random token like b58d7aabb705df024fa578d9a0e20a7c, but I'm not sure if that's much better.

Using GUIDs to uniquely identify something seems pretty standard for many forms of identification related to dynamic linking/FFI in various languages.

Another option would be to specify one random hash/guid for the whole crate at the top level, and then use only simple incremental numbers for the types in the crate. But that has a higher chance of resulting in issues when working on multiple versions/branches, if two branches both bump a number up to the same new number with incompatible safety invariants.

text/0000-export.md
m-ou-se: The point of this RFC is that you do *not* need the exact same crate for creating the .so and for linking to it. (So you can swap compatible versions, compile it with a different compiler or with different settings, etc.) You only need something that is ABI compatible, and having a mismatch does not result in memory unsafety thanks to all relevant safety information being hashed into the symbols.

m-ou-se:
> Maybe I misunderstood something. Why do you limit the RFC to shared libraries? Static libraries are sometimes also useful. Ok, it is perhaps only a special case. But we develop a library operating systems and link a static library as kernel to the application. Currently, the interface between application and libos is a C interface....

That was also brought up here: https://github.com/rust-lang/rfcs/pull/3435#discussion_r1202505606:

> That's a good point. Perhaps that means we should use slightly different names, but I don't think that changes anything else substantially about the RFC.

text/0000-export.md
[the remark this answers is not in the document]
RalfJung: Yeah there might be niche uses but it is very dangerous (unless all fields have primitive types whose invariant is never going to change)

RalfJung:
Is it possible to `export` an `unsafe fn`?

I think that has some similar issues to structs with invariants: even if the signature of the function stays the same, its safety requirements might change, therefore somehow those requirements have to be encoded into the symbol name.

text/0000-export.md
RalfJung: > having a mismatch does not result in memory unsafety thanks to all relevant safety information being hashed into the symbols.

I think there's one point where this does not work: when safety depends on functional properties of upstream code. Imagine crate A exporting a `fn make_int() -> i32`. Create B re-exports that function. Crate C does something like
```rust
if A::make_int() != B::make_int() { unsafe { unreachable_unchecked(); } }
```
This is perfectly sound as long as `B` promises that its function always behaves like that of `A`. Even if an updated `A` changes the integer returned, things will keep working.

But if we are in a dynamic linking situation, can't it happen that we get one version of the function via `B` (since maybe it got  inlined or whatever) but a different version via `A` (since maybe we are dynamically linking that one)? Basically exactly the reason we need hashes for types with private fields?

OTOH it seems like there are other issues in that situation (e.g. a `static mut` in `A` that would suddenly behave differently because now there are 2 copies of it), so I might also be misunderstanding what exactly is going on during dynamic linking here.

digama0: @RalfJung Did you mean to call `make_int()` there, or compare the function pointers? Obviously unless `make_int()` is a pure function we would not necessarily be allowed to assume it always returns the same value.

RalfJung: Right I was assuming a promise of `A` that the function is pure.

(It's common for unsafe code to depend on functional properties of safe upstream code like that.)

text/0000-export.md
str4d: > > Then public and private types are treated identically.
> 
> It's critical that they are not treated identically.

I think you misunderstood my comment; I apologise for not being clearer.

I was not saying that they are treated identically in terms of _how they are hashed_; indeed as you say, the types with private fields have additional considerations, which are represented by (what you call) the hash that the user sets on the type.

I was saying that they are treated identically in terms of _how they are used as children of public parent types_. This would make the logic for calculating the commitment to a public type simpler; there would not be two separate pathways for handling public types (recursing down) and private-containing types (grab the user-pinned hash value). Instead, the logic would always take the hash of the child type, and that hash would either be computed (for public types) or pinned (for private-containing types).

text/0000-export.md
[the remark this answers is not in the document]
tgross35: This is something that I could see being different for different profiles too. E.g. you may prefer to build from source for `cargo test` so you don't need the correct version installed, but any release builds would want to link the system. It would be nice if there were some way to configure this.

I assume the rustc flag will be something like `--extern=dynamic=...` that treats the extern crate like a header file?

tgross35:
A few minor questions:

1. Will it be possible to do `#[export, no_mangle]` (or separate attributes)? To create one interface and link it from both C and Rust, this would only do the "treat dependencies as header files" part. Maybe this could be a good minimum viable product since it punts the mangling question that is still under heavy discussion?
2. How heavily will real-world usage depend on Cargo supporting binary dependencies https://github.com/rust-lang/cargo/issues/9096? Right now that is kind of a blocker to ergonomic rust shared libraries (at least for my use case)
3. How does this interact with `dylib` vs. `cdylib`? I assume that any prepackaged artifacts generally need to be `cdylib` in order to ensure stability across different compiler versions, but the interaction with `libstd` is a bit weird here (the differences between these two interfaces isn't very well documented)
4. Is this RFC more or less looking for team review at this point? Or are there heavy changes / blockers expected

m-ou-se:
Update: We're working on launching a research project around the problem this RFC is trying to solve. See https://github.com/rust-lang/rust-project-goals/pull/155

m-ou-se:
> Will it be possible to do `#[export, no_mangle]` (or separate attributes)? To create one interface and link it from both C and Rust, this would only do the "treat dependencies as header files" part (i.e. layout & symbols only, no codegen). Maybe this could be a good minimum viable product since it punts the mangling question that is still under heavy discussion?

If we allow both attributes, I'd expect the result would be that the item will be available under two symbol names. (The `export` mangled one, and the `no_mangle` unmangled one.)

It sounds like you're asking if we could do `#[extern]` without mangling. I think the answer is a no, as the main point of this RFC is to achieve safety through symbol mangling.

Nonetheless it's an interesting question what would remain of this RFC if we postponed that part for an initial implementation.

> How heavily will real-world usage depend on Cargo supporting binary dependencies [Tracking Issue for RFC 3028: Allow "artifact dependencies" on bin, cdylib, and staticlib crates cargo#9096](https://github.com/rust-lang/cargo/issues/9096)? Right now that is kind of a blocker to ergonomic rust shared libraries (at least for my use case)

We'd ideally need cargo to understand the concept of a dynamically linked dependency to make this an ergonomic feature. However, using (generated?) 'header-only' crates, we can make this ergonomic enough for a first usable version without any special cargo support.

> How does this interact with `dylib` vs. `cdylib`? I assume that any prepackaged artifacts generally need to be `cdylib` in order to ensure stability across different compiler versions, but the interaction with `libstd` is a bit weird here (the differences between these two interfaces isn't very well documented)

Yup, the interaction of different libstds is an unsolved problem. The easiest solution is to use 'cdylib', but we will need to address this problem if we ever expect to seemlessly transfer types like `String` through extern boundaries, which could be allocated on one side and dropped on another.

> Is this RFC more or less looking for team review at this point? Or are there large changes / blockers expected

I still need to process all the feedback and update the RFC. But those are mostly relatively minor changes.

Regardless of those changes, I think team review and approval would be easier if we have more information on alternative designs so the tradeoffs can be judged better. Hence the proposal for a research project: https://github.com/hexcatnl/rust-project-goals/blob/safe-linking/src/2025h1/safe-linking.md

text/0000-export.md
kornelski: I'd prefer to have separate versions for source code compatibility and for ABI compatibility.

For example, changing a function argument from `u32` to `#[repr(transparent)] NewType(u32)` is a source-code semver-major, but it shouldn't be a change for the ABI.

Conversely, a change from `#[export]` to `#[inline(always)]` is an ABI break, but does not affect compatibility when building from source.

I suppose separate versioning could be done by publishing a separate crate with just re-exports for the ABI. However, I don't know how hard would it be to maintain two crates, and not get into issues with orphan rules, or crate names in symbol names making it impossible to seamlessly substitute items.

Skgland:
> > How heavily will real-world usage depend on Cargo supporting binary dependencies [Tracking Issue for RFC 3028: Allow "artifact dependencies" on bin, cdylib, and staticlib crates cargo#9096](https://github.com/rust-lang/cargo/issues/9096)? Right now that is kind of a blocker to ergonomic rust shared libraries (at least for my use case)
> 
> We'd ideally need cargo to understand the concept of a dynamically linked dependency to make this an ergonomic feature. However, using (generated?) 'header-only' crates, we can make this ergonomic enough for a first usable version without any special cargo support.

How would this work with `cargo install` wouldn't cargo need to know to install both the binary and all its dynamically linked dependencies. Where would those dynamic libraries live and how would it work if one binary depends on `foobar@1.*` while another depends on `foobar@2.*` as I would expect both to result in a `foobar.dll` that are likely ABI incompatible.


Edit:

Also how should a dependency be handled that appears multiple times in a crates dependency graph as both dynamically and statically linked?

tmvkrpxl0:
I think this should also support conditionally using stable ABIs of crate author's choice only when it's being compiled as dynamic library.

Say my project is statically linking to another crate that uses `#[export]` attributes a lot. Then many of its types would use different ABI which may not be optimal when I'm statically linking to it.
Compiler should compile those types using unstable default ABI for optimal performance.
 
This could be achieved by providing options like `#[export(abi(C))]` so that it can specify ABI itself instead of relying on `extern` or `#[repr]` which would conditionally apply other ABI when it's needed.
Or it could provide some cfg variable for crate's authors to use with `#[cfg...]` but I don't think that'd be compatible with `extern`

tmvkrpxl0:
> I think this should also support conditionally using stable ABIs of crate author's choice only when it's being compiled as dynamic library.
> 
> Say my project is statically linking to another crate that uses `#[export]` attributes a lot. Then many of its types would use different ABI which may not be optimal when I'm statically linking to it. Compiler should compile those types using unstable default ABI for optimal performance.
> 
> This could be achieved by providing options like `#[export(abi(C))]` so that it can specify ABI itself instead of relying on `extern` or `#[repr]` which would conditionally apply other ABI when it's needed. Or it could provide some cfg variable for crate's authors to use with `#[cfg...]` but I don't think that'd be compatible with `extern`

nvm it's already handled in the rfc:
```md
# Importing Exported Items
Normally, when using a crate as a dependency, any #[export] attributes of that crate have no effect and the dependency is statically linked into the resulting binary.

When explicitly specifying dynamic = true for the dependency with Cargo.toml, or when using a extern dyn crate …; statement in the source code, only the items marked as #[export] will be available and the dependency will be linked dynamically.
```

text/0000-export.md
tmvkrpxl0: @kornelski Shouldn't the `NewType` be incompatible? Even if the bit representation is same, in your example it doesn't have `pub`, which could mean it hold some invariant information. It is possible for the change to occur to enforce certain invariant, which can coincidentally have same bit representation. In that case it should be regarded as incompatible in case previous use cases are not compatible with new approach.
In this sense, I think `API` incompatibility should also imply `ABI` incompatibility.
If library author want to keep the ABI compatibility, they should provide wrapper function that has same API signature as previous version.

PR #3441, opened 2023-05-30
Fix autolink syntax on 3216

notriddle:
[Rendered](https://github.com/rust-lang/rfcs/blob/notriddle/3216-autolink/text/3216-closure-lifetime-binder.md)


ehuss (approved):
Thanks!

PR #3442, opened 2023-06-01
Fragment Specifiers for Generic Arguments

JarredAllen:
I wrote up my thoughts from [this rust-internals thread](https://internals.rust-lang.org/t/macro-rules-metavariable-for-generic/18889) into an RFC.

### DIFF
--- a/text/0000-fragment-specifiers-for-generic-arguments.md
+++ b/text/0000-fragment-specifiers-for-generic-arguments.md
@@ -1,0 +1,155 @@
+- Feature Name: fragment-specifiers-for-generic-arguments
+- Start Date: 2023-05-31
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Right now, there is no support for parsing the syntax of generic
+parameters/arguments in declarative macros. This makes it difficult to
+impossible to write a declarative macro that handles arbitrary generics easily.
+I propose adding fragment specifiers to parse a generic parameter definition as
+a whole and also parts of the definition.
+
+# Motivation
+[motivation]: #motivation
+
+I personally encountered this issue when attempting to write a declarative macro
+to implement a trait on a type. I wanted to write something like this minimal
+toy example:
+
+```rust
+macro_rules! implement_debug {
+    { $params:generic $type:ty } => {
+        impl $params Debug for $ty {
+            /* .. implementation goes here .. */
+        }
+    };
+}
+
+struct Container<'a, T>(&'a T);
+
+implement_debug!(<'a, T: Debug + 'a> Container<'a, T>);
+```
+
+However, with the current state of declarative macros, there's no way to parse
+arbitrary generic parameters in the body of the macro, forcing this macro of
+mine to be a procedural macro. However, declarative macros are easier to read
+and write, and this could be a declarative macro if there was only a way to
+parse generic parameters.
+
+Additionally, more complicated macros want to be able to parse each parameters
+and its bounds for use in various places, so I'd like this as well:
+
+```rust
+macro_rules! implement_debug {
+    { < $( $param:generic_param $( : $bound:generic_bound )? ),+ > $type:ty } => {
+        impl < $( $param $( : $bound )? ),+ > Debug for $ty {
+            /* .. implementation goes here .. */
+        }
+    };
+}
+
+struct Container<'a, T>(&'a T);
+
+implement_debug!(<'a, T: Debug + 'a> Container<'a, T>);
+```
+
+Any toy example will be obviously redundant with `:generic`, but more involved
+macros sometimes want it.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When explaining fragment specifiers, this can be explained by adding the new
+fragment specifiers and their descriptions:
+
+* `:generic`: A full set of generic parameters and their bounds (e.g. `<'a, T:
+  'a + SomeTrait, const N: usize>`)
+* `:generic_param`: A generic parameter (e.g. `'a`, `T`, or `const N`)
+* `:generic_bound`: Bounds on a generic parameter (e.g. `'lifetime + SomeTrait`
+  on a type or `usize` on a const parameter).
+* `:generic_default`: A default value for a generic type or lifetime
+
+These four parameters are designed to make it easier to write declarative macros
+that take in generic arguments (e.g. to use with a type or function), and then
+use them to be generic on e.g. type definitions or `impl` blocks.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Exact parsing behavior:
+* `:generic` matches the
+  [`GenericParams`](https://doc.rust-lang.org/reference/items/generics.html)
+  grammar item.
+* `:generic_param` matches any of a lifetime, an identifier, or `const` followed
+  by an identifier.
+* `:generic_bound` matches the
+  [`TypeParamBounds`](https://doc.rust-lang.org/reference/trait-bounds.html)
+  (can be the bounds on a type parameter) or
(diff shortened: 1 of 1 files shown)


text/0000-fragment-specifiers-for-generic-arguments.md
```diff
@@ -0,0 +1,155 @@
...
+
+Are these the best fragment specifiers to use for parsing macros? I'd like
+opinions from other people who write macros that would want something like this
+about if breaking up generics into some other form might be more useful. I think
+this set of fragment specifiers is the best for my use cases, but other people
+might be interested in macros that parse differently and they might want other
+things instead.
+
+Also, should `:generic_bound` include the preceding `:` in the match (e.g. `: 'a
++ SomeTrait` in the example above)? And likewise with the `=` before
```
programmerjake, line 146: don't word-wrap in the middle of a \` code span, it doesn't get rendered properly

JarredAllen: Oops, I guess I need a better editor config. Fixed in d77ac900

text/0000-fragment-specifiers-for-generic-arguments.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When explaining fragment specifiers, this can be explained by adding the new
+fragment specifiers and their descriptions:
+
+* `:generic`: A full set of generic parameters and their bounds (e.g. `<'a, T:
+  'a + SomeTrait, const N: usize>`)
+* `:generic_param`: A generic parameter (e.g. `'a`, `T`, or `const N`)
```
oxalica, line 70: `const N` does not work for the macro example above, `$param:generic_param $( : $bound:generic_bound )?`. Const generics have the syntax `const $param:ident : $ty:ty $(= $default:expr)?`. Note that it is a type instead of generic bounds after the semicolon, so syntactically it should also accept non-trait-like types like `const N: (i32, i32)`, though it's forbidden currently.

text/0000-fragment-specifiers-for-generic-arguments.md
```diff
@@ -0,0 +1,155 @@
...
+  [`GenericParams`](https://doc.rust-lang.org/reference/items/generics.html)
+  grammar item.
+* `:generic_param` matches any of a lifetime, an identifier, or `const` followed
+  by an identifier.
+* `:generic_bound` matches the
+  [`TypeParamBounds`](https://doc.rust-lang.org/reference/trait-bounds.html)
+  (can be the bounds on a type parameter) or
+  [`LifetimeBounds`](https://doc.rust-lang.org/reference/trait-bounds.html) (can
+  be the bounds on a lifetime parameter) grammar items, or a type (can be the
+  bounds on a const parameter).
```
oxalica, line 93: Should it be named "generic_bound" or "generic_bound**s**"?
Also that `TypeParamBounds` does not allow empty, while the generic syntax do allow `<T: /*nothing*/>` (except for const generics). If we want it to match multiple bounds, it should also match no bounds.

oxalica (changes requested):
This RFC does not include specifiers for where-clause, which is highly related to generics. Is this intentional? Maybe we should mention this in text.

clarfonthey:
> This RFC does not include specifiers for where-clause, which is highly related to generics. Is this intentional? Maybe we should mention this in text.

I think this would be nice longer-term, but at least for now, the extremely simple case of being able to check for pieces like `<T, const N>` without resorting to using token-trees is pretty good on its own. That said, the inclusion of bounds in this RFC does beg the inclusion of where bounds too.

Aloso:
I like the general idea of this RFC, but I don't like that a macro accepting `$( $p:generic_param $( : $b:generic_bound )? ),*` can accept illegal syntax (like a lifetime followed by a trait bound, or a const generic followed by a lifetime bound).

I also don't like that matching all generics syntax is pretty verbose:

```rust
< $(
    $p:generic_param
    $( : $( $b:generic_bound )? )?
    $( = $d:generic_default )?
),* $(,)? >
```

a `where` clause could be written like this, if we require trailing commas:

```rust
where $(
    $( $p1:lifetime $( : $( $b1:generic_bound )? )? , )?
    $( $p2:path $( : $( $b2:generic_bound )? )? , )?
)*
```

But this isn't forward compatible with

- [return type notation](https://github.com/rust-lang/rust/pull/109010) (`where Self::method(): OtherTrait`)
- [equality constraints](https://github.com/rust-lang/rust/issues/92827) (`where Self::ID = {N + 1}`)
- possibly more

Aloso:
Maybe it would be best to remove `:generic_param`, `:generic_bound` and `:generic_default` from the RFC (unless you can demonstrate with an example that they're useful), and instead add `:where_clause`.

text/0000-fragment-specifiers-for-generic-arguments.md
JarredAllen: This was a deliberate deviation from the default syntax to make it easier to parse an arbitrary set of generic parameters. If we use the fragment specifiers as I defined them in the example macro defined in the RFC (`< $( $param:generic_param $( : $bound:generic_bound )? ),+ > `, then this successfully matches `<const N: usize>` (`$param` being `const N` and `$bound` being `usize`.

If it matches as you're requesting that it match, then the same pattern will match this trait, but it will also allow in nonsense like `<const N: usize: 'a + SomeTrait>` which my approach avoids. I'm not sure how one could change the pattern to avoid allowing for nonsense like that.

Though also, my approach would match `<const N>`, which is also nonsense, so idk which one would be better to reject while matching the macro. I'm open to changing it if you have arguments for your way of splitting the definition between the two fragment specifiers being better.

To your last point, the definition that I give for `:generic_bounds` does explicitly allow for a type, so that syntax would be accepted.

Aloso: Your approach also matches `const N: 'a` and `'a: usize`

text/0000-fragment-specifiers-for-generic-arguments.md
JarredAllen: Neat, I never knew that Rust allowed `<T:>` as a generic parameter list. That doesn't seem very useful to me, but I'll change it to allow that since I agree that it should be matched if Rust allows it.

JarredAllen:
I added a `:where_clause`, as per multiple requests. They were excluded initially because I forgot that they existed, not for any real reason.

JarredAllen:
> I like the general idea of this RFC, but I don't like that a macro accepting `$( $p:generic_param $( : $b:generic_bound )? ),*` can accept illegal syntax (like a lifetime followed by a trait bound, or a const generic followed by a lifetime bound).
> 
> I also don't like that matching all generics syntax with the more granular fragments is pretty verbose:

Yeah, sadly the only way to prevent it from accepting illegal syntaxes like `const N: 'a` or `'a: usize` would make it even more verbose and complicated to match the entire generics syntax, which makes your other point worse. Though, you don't need the `$()?` around `:generic_bound` anymore, I didn't know that nothing could be put there and have since updated it to allow that, so it's slightly less verbose.

I tried to strike a happy medium where relatively few illegal syntaxes are allowed (I think they've all been said on this page? Unless I'm missing some) but also it's reasonably non-verbose to parse the full generic parameter definition syntax and we're not adding too many new fragment specifiers into the language, and the split is done in such a way that it could be useful for implementing macros.

I'm open to any suggestions you have for making that trade-off better, but this was the best balance between the two goals I could think of.

text/0000-fragment-specifiers-for-generic-arguments.md
```diff
@@ -0,0 +1,166 @@
...
+Any toy example will be obviously redundant with `:generic`, but more involved
+macros sometimes want it.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When explaining fragment specifiers, this can be explained by adding the new
+fragment specifiers and their descriptions:
+
+* `:generic`: A full set of generic parameters and their bounds (e.g. `<'a, T:
```
CohenArthur, line 70: maybe a nit but `generic` being singular confuses me. I think it would make more sense for it to be named `generics` or `generic_set`, or anything to add plurality to the concept, especially since it would be responsible for parsing the opening and closing angle brackets

PR #3449, opened 2023-06-18
Contextual target feature detection

calebzulawski:
Add new `is_{arch}_feature_enabled` macros for detecting at compile time if a target feature is enabled in a context.

### DIFF
--- a/text/0000-target-feature-enabled.md
+++ b/text/0000-target-feature-enabled.md
@@ -1,0 +1,194 @@
+- Feature Name: `target_feature_enabled`
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Add new `is_{arch}_feature_enabled` macros for detecting at compile time if a target feature is enabled in a context.
+
+# Motivation
+[motivation]: #motivation
+
+[RFC #2045](https://github.com/rust-lang/rfcs/pull/2045) added two methods of querying target features:
+* conditional compilation, e.g. `#[cfg(target_feature = "avx")]`
+* runtime detection, e.g. `is_x86_feature_detected!("avx")`
+
+Conditional compilation only allows querying the base target features, and does not interact with `#[target_feature]`.
+Runtime detection is necessary for safely calling functions tagged with `#[target_feature]`, but incurs a runtime overhead.
+In some cases, it is necessary to determine which target features are enabled at code generation (particularly after inlining) to allow various optimizations:
+
+```rust
+#[inline(always)]
+fn conditional_compilation(...) {
+    // This branch is always optimized out, and depends on the target features enabled at code generation.
+    // If this were `cfg!(target_feature = "avx")`, this branch would never select the AVX version
+    // (without enabling AVX for the entire binary).
+    if is_x86_feature_enabled!("avx") {
+        unsafe { avx_implementation(...) }
+    } else {
+        generic_implementation(...)
+    }
+}
+
+#[inline(always)]
+fn runtime_detection(...) {
+    // If AVX is enabled at code generation, this branch is optimized out and runtime detection is skipped.
+    if is_x86_feature_enabled!("avx") || is_x86_feature_detected!("avx") {
+        unsafe { avx_implementation(...) }
+    } else {
+        generic_implementation(...)
+    }
+}
+
+#[target_feature(enable = "avx")]
+unsafe fn with_avx_enabled(...) {
+    // This call selects the AVX implementation, because of this function's target features.
+    conditional_compilation(...);
+
+    // This call selects the AVX implementation, skipping runtime detection!
+    runtime_detection(...);
+}
+```
+
+A particularly useful case is nested runtime detection:
+
+```rust
+#[inline(always)]
+fn first(...) {
+    #[target_feature(enable = "avx")]
+    #[inline]
+    unsafe fn first_avx(...) {
+        second(...)
+    }
+    
+    #[inline(always)]
+    fn first_generic(...) {
+        second(...)
+    }
+
+    if is_x86_feature_detected!("avx") {
+        unsafe { first_avx(...) }
+    } else {
+        first_generic(...)
+    }
+}
+
+#[inline(always)]
+fn second(...) {
+    #[target_feature(enable = "avx")]
+    #[inline]
+    unsafe fn second_avx(...) {
+        ...
+    }
+    
+    #[inline(always)]
+    fn second_generic(...) {
+        ...
+    }
+
+    if is_x86_feature_enabled!("avx") || is_x86_feature_detected!("avx") {
+        unsafe { second_avx(...) }
+    } else {
+        second_generic(...)
+    }
+}
+```
+
+After inlining, calling `first` only runs target feature detection once!  The nested runtime detection is elided.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The syntax for `is_{arch}_feature_enabled` macro is like `is_{arch}_feature_detected`.
+
+Instead of detecting if the target feature is supported at runtime, the macro returns whether the target feature is supported in the current context at compile time.
+
+A target feature is supported in a particular context if at least one of the following is true:
+* The target supports the feature by default, or it is enabled with the `-Ctarget-feature` flag (i.e. `cfg!(target_feature = "feature")` is true)
+* The function containing the macro invocation is annotated with `#[target_feature(enable = "feature")]`
+* The function containing the macro invocation is inlined into a function annotated with `#[target_feature(enable = "feature")]`
+
+## Example
+```rust
+/// Computes `(a * b) + c` with unspecified rounding.
+#[inline]
+fn fast_mul_add(a: f32, b: f32, c: f32) -> f32 {
+    if is_x86_feature_enabled!("fma") {
+        a.mul_add(b, c) // a single instruction faster than separate mul and add
+    } else {
+        a * b + c // separate mul and add is faster than calling the library `fmaf` function
+    }
+}
+```
+
+## Keep in mind
+The accuracy of this macro is "best-effort": depending on your particular code and Rust configuration, it may still return `false` even if the feature is enabled.
+For example:
+* Using `#[inline(always)]` may be more accurate than `#[inline]`
+* This feature depends on MIR inlining being enabled
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The macros generate calls to new compiler intrinsics.
+These intrinsics are lowered by each codegen backend to the appropriate `true` or `false` value depending on the target features of the parent function.
+
+Codegen operates on optimized MIR, so the intrinsic is lowered after MIR inlining has occurred.
+When lowering the intrinsic, the backend can optionally account for additional inlining passes.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Target features are already complicated, and adding this macro increases complexity.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Compared to runtime detection
+In small functions, the runtime cost of feature detection and branching is unacceptable.
+Runtime detection is also inefficient and redundant when used in many functions that may call each other.
+
+## Compared to conditional compilation
+[RFC #2045](https://github.com/rust-lang/rfcs/pull/2045) proprosed making `#[cfg(target_feature = "feature")]` context-dependent, but this was never implemented.
+`cfg` is always consistent throughout a program, so making it context dependent might be confusing and lead to mistakes.
+Additionally, it's not clear how context-dependent `#[cfg(target_feature = "feature")]` on items (rather than blocks) should or could work.
+Adding a new `is_{arch}_feature_enabled` macro avoids this complexity.
+
+## Compared to a library
+This kind of feature necessarily requires compiler support.
+
+Some crates, such as [`glam`](https://github.com/bitshifter/glam-rs/blob/47630c35adb52c08def02cce4ddfbddaa45d1b9d/src/f32/sse2/vec4.rs#L648-L658) or [`simd-json`](https://github.com/simd-lite/simd-json/blob/21f7878505a472ab93077b18294888d73ce330b9/src/lib.rs#L152-L159), opt to use `cfg` which simplifies the library API, but is pessimistic since many applications compile with default target features to maximize compatibility.
+Other libraries, such as [`rustfft`](https://docs.rs/rustfft/6.1.0/rustfft/struct.FftPlannerAvx.html) or [`highway`](https://docs.rs/highway/1.0.0/highway/index.html), leave the decision up to the user at the cost of a more verbose API, especially for users that are not aware of or interested in SIMD or target features.
+
+The [`multiversion`](https://github.com/calebzulawski/multiversion) crate provides a `target_cfg` macro which is identical to `cfg`, but accounts for the `#[target_feature]` attribute.
+This is similar functionality to this RFC, but is limited to functions tagged with special attributes and does not account for inlining.
+Therefore, `multiversion` is unsuitable for creating small reusable functions that are intended to inline into another function with target features enabled.
+
+# Prior art
+[prior-art]: #prior-art
+
+I've already mentioned [RFC #2045](https://github.com/rust-lang/rfcs/pull/2045) a couple times, which briefly discusses this problem, but the proposed solution was never implemented.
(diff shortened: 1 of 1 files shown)


calebzulawski:
Previous discussion: [#project-portable-simd zulip](https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/MIR.20inlining.20.2B.20target.20features), [#t-libs/stdarch zulip](https://rust-lang.zulipchat.com/#narrow/stream/208962-t-libs.2Fstdarch/topic/Determining.20if.20a.20target.20feature.20is.20enabled.20after.20inlining), [Rust Internals](https://internals.rust-lang.org/t/pre-rfc-contextual-target-feature-detection/18953)

clarfonthey:
I'm a bit confused how this would be different from doing `cfg!(target_feature = "...")`, considering how the RFC doesn't even mention that syntax. Isn't that effectively what this does, except with the guarantee that it will allow context-sensitive optimisations? Is there a reason to not just fix the cfg to do that?

Lokathor:
`cfg` is globally consistent.

The specific intent of this new macro is that when a function gets inlined into a caller with more target features enabled the inlined function will (as often as possible, best effort, etc) see the additional target features available within the caller's scope now that it's been inlined (and then pick a faster way to do the operation).

calebzulawski:
A simple example of why I don't think `cfg` is appropriate:
```rust
#[cfg(target_feature = "avx")]
fn inner() {}

#[target_feature(enable = "avx")]
fn outer() {
    #[cfg(target_feature = "avx")]
    {
        inner()
    }

    if cfg!(target_feature = "avx") {
        // something that depends on having called `inner`
    }
}
```

If this is compiled without `-Ctarget-feature=+avx`, `inner` is not enabled.  If `#[cfg]` in `outer` becomes context dependent, this pattern is now impossible, since there is no way to query the base target features.  An argument could be made that only `cfg!(target_feature)` should be context-dependent, but this inconsistency would be confusing, in my opinion.  In the example above, it could even result in unsoundness.  Adding a new macro separates contextual detection entirely from conditional compilation.

Diggsey:
Could the `is_x86_feature_detected!("avx")` functions not just be made more optimizable?

Either it's inlined into a scope where the answer is already known (because of the target features) - in which case LLVM can optimize away the unused branch anyway.

Or, it's inlined into a scope where the answer is not clear: in that case a branch is needed, but it should be possible to push the expensive detection part out of any hot code paths, and the branch prediction should be very accurate?

Basically, I'm asking if the "slow path without a branch" is so much faster than the "fast path with a branch" that it makes sense to introduce an entirely new set of macros?

And, assuming the difference *is* big enough to matter, could a more general macro like this work instead?

```rust
if const_fallback!(false, is_x86_feature_detected!("avx")) {
    ...
} else {
   ...
}
```

Where `const_fallback!` has the "magic" that it tries to evaluate the second parameter at compile time, and if it cannot, it evaluates to the first parameter.

programmerjake:
> Basically, I'm asking if the "slow path without a branch" is so much faster than the "fast path with a branch" that it makes sense to introduce an entirely new set of macros?

in some cases, yes. an example is the `approx_mul_add` function that compiles to just 1 (fma) or 2 instructions (mul + add) and is used in hot loops inlined across crates:
https://internals.rust-lang.org/t/pre-rfc-contextual-target-feature-detection/18953/10?u=programmerjake
https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=47639908d8de532d409d383bfa5b3f6a
> 
> And, assuming the difference _is_ big enough to matter, could a more general macro like this work instead?

we would still want a separate macro because we likely want to eventually have a llvm intrinsic that is lowered to a compile-time constant after all inlining (mir and llvm (both before and during link-time)) is complete. if we used your `const_fallback!` macro we would still need to special-case the whole combination of both macros, which is more complex, and could get much harder since additional expressions might be inserted between the macros:
```rust
const fn blah(v: bool) -> bool { ... }
const_fallback!(false, blah(is_x86_feature_detected!("fma")))
```

calebzulawski:
For a concrete example, here's the asm: https://rust.godbolt.org/z/Te4nsjaK3

Both options, with and without `fma`, are preferable to the runtime detection option which adds considerable overhead.  For such a small function, runtime detection is infeasible.  The proposed macros are intended to work hand in hand with runtime detection--small functions like this example can inline into larger functions that are selected with runtime detection.  Since `#[target_feature]` functions can only be safely called after runtime feature detection, using `is_x86_feature_enabled` can be thought of as deferring to an earlier `is_x86_feature_detected`.

programmerjake:
> For a concrete example, here's the asm: https://rust.godbolt.org/z/Te4nsjaK3

corrected example with `detect_mul_add` actually calling a function with enabled target features:
https://rust.godbolt.org/z/GKaaWG73M

CAD97:
For this example

<details><summary>this example</summary>

```rust
#[inline]
fn algorithm(v: &mut [f32]) {
    for v in v {
        *v = fast_sin_2pi(*v);
    }
}

#[inline(never)]
#[target_feature(enable = "fma")]
unsafe fn with_fma(v: &mut [f32]) {
    algorithm(v)
}

#[inline(never)]
fn without_fma(v: &mut [f32]) {
    algorithm(v)
}

pub fn dispatch(v: &mut [f32]) {
    if is_x86_feature_detected!("fma") {
        unsafe { with_fma(v) };
    } else {
        without_fma(v);
    }
}
```

-----
</details>

it should in theory be possible for LLVM to compile both

```rust
pub fn dispatch(v: &mut [f32]) {
    if is_x86_feature_detected!("fma") {
        unsafe { with_fma(v) };
    } else {
        without_fma(v);
    }
}

// and

pub fn dispatch(v: &mut [f32]) {
    if is_x86_feature_detected!("fma") {
        algorithm(v);
    } else {
        algorithm(v);
    }
}

// or even

pub fn dispatch(v: &mut [f32]) {
    algorithm(v);
}
```

identically. This is what is meant by hoisting; the backend should be able to take `is_x86_feature_detected!("fma")` out of where it is in a hot inlined loop and only check it a single time at the start of whatever function the inlining stops at.

Of course, whether the optimizer actually does so, and is willing to do the duplicated outlining considered desirable here is a different question. What makes this interesting is that it's not just a question of what can a good optimizer do; it's a question of (pseudo) guarantees.

For `is_x86_feature_enabled!` to provide the desired properties, I think it needs to be restricted to being used in `#[inline(always)]` functions, and for said functions to be inlined during the MIR stage, before being lowered to the codegen backend; when it gets to the codegen backend, it just sees `if true` or `if false` (or not even that if MIR opts collapsed that).

The reason/problem is that optimization is bottom-up. If you're asking the backend optimizer to answer `is_x86_feature_enabled!` post inlining, you're asking for a major change to how inlining/optimization works to support that. (Namely, the backend needs to multiversion any functions which include `is_x86_feature_enabled!` branches, with the parent caller choosing which to call based on its feature enabled state (or multiversion and inline, based on some sort of multiversioned inlining heuristic).)

As I understand it, `is_x86_feature_detected!` is just a function call, as far as LLVM is concerned. Any meaningful improvement probably requires cooperation from LLVM, and I have a minor expectation that it will be easier to get "`std_detect::detect::cache::detect_and_initialize` is known to only ever be called at most once; treat calling it a second time as UB; and `std_detect::detect::cache::CACHE` is only modified from `detect_and_initialize`, so it won't change after that first call and loads can be coalesced" (optimize `is_x86_feature_detected!` better) than "an intrinsic which tells me if a target feature is enabled after inlining" (`is_x86_feature_enabled!` works).

On the other hand, implementing target feature multiversioning in MIR is entirely contained in the domain of rustc and doesn't rely on codegen backends to cooperate. I don't know the status of MIR optimization and inlining, and if an MIR-only `is_target_feature_enabled!` would be usefully functional.

I'm not an expert in inlining/optimizations, so I may be incorrect in believing doing both meaningful inlining/optimization and `is_target_feature_enabled!` basically requires function multiversioning[^1], so I'll defer to @rust-lang/wg-mir-opt there. But I'm fairly confident.

[^1]: Rationale: especially as is the case here, where you have `for _ in _ { if enabled!() { a() } else { b() } }`, optimization *very much* relies on the ability to assume one arm or the other is taken. Even if we know that `test!()` is constant, it's not always the case that hoisting the test out of the loop and duplicating the loop is optimal. And even if `test!()` has special "always hoist to function entry and deduplicate" semantics (hey, that's what's desired for `detected!()`!), you're additionally still a) asking for noninlined calls to skip the test, and b) inlining heuristics to inline as if the function were just one of the tested arms, not both. That's essentially exactly what multiversioning is.

    Interestingly, even with a inlining-unaware `is_target_feature_enabled!`, I believe it should in theory be possible to implement "perfect" full multiversioning in userland, though still requiring listing desired multiversioned target feature sets in `#[multiversion]` and calling functions as `multiversioned!(approx_mul_add)(base, mul, add)`[^2] instead of `approx_mul_add(base, mul, add)`.

[^2]: <details><summary>Desugar: untested, very roughly,</summary>

    ```rust
    #[multiversioned(targets("a", "b"))]
    fn approx_mul_add(base: f32, mul: f32, add: f32) -> f32 { .. }
    // becomes
    enum approx_mul_add {}
    impl Multiversioned<approx_mul_add, target_feature_bitflags!("a")> {
        #[target_feature(enable("a"))]
        unsafe fn call(base: f32, mul: f32, add: f32) -> f32 { .. }
    }
    impl Multiversioned<approx_mul_add, target_feature_bitflags!("b")> {
        #[target_feature(enable("b"))]
        unsafe fn call(base: f32, mul: f32, add: f32) -> f32 { .. }
    }

    // and
    multiversioned!(approx_mul_add)
    // becomes
    (Multiversioned::<approx_mul_add, enabled_target_feature_bitflags!()>::call)
    ```

    although tbf I'm not 100% on everything working out perfectly here; it probably needs some autoref specialization and might not like the use of a function pointer needing devirtualization, definitely not the fact that I've not properly covered the `unsafe` yet. Also probably not quite happy w.r.t. orphan rules yet, either; a trait impl with associated type and const to attach a function pointer to was the original concept but simplified for demo here. (Further modifications to make the scheme work are left as an exercise to the reader.)

    </details>

riking:
> `is_{arch}_feature_detected` could make use of this macro for additional optimization opportunity.

Could the detection macro simply be replaced with a constant true when the target feature is enabled? Reduce the added API.

nagisa:
If my memory serves me right it was always the intent for the `is_*_feature_detected!` to take `#[target_feature(enable="")]` into account, but only for the function the macro is invoked inside.

Unfortunately I don’t think you can rely on inlining in any way or form to make this work – it is strictly an optimization and there are plenty of examples of code where inlining cannot happen (such as if the two functions are a part of the same strongly connected callee graph component.)

And if you’re okay with it being “just” an optimization there are a good deal of implementation approaches how this could be implemented without adding additional API: it could be handled entirely in Rust, only being limited by its own inliner; it could be implemented as a LLVM optimization, by adding some more knowledge of cpuid et al. to it. In either case this all can easily collapse as an inlining decision is made, or optimization is disabled in some way.

programmerjake:
> And if you’re okay with it being “just” an optimization there are a good deal of implementation approaches how this could be implemented without adding additional API: it could be handled entirely in Rust, only being limited by its own inliner; it could be implemented as a LLVM optimization, by adding some more knowledge of cpuid et al. to it. In either case this all can easily collapse as an inlining decision is made, or optimization is disabled in some way.

please elaborate on how you think it can be implemented without additional API, because afaict `is_*_feature_detected` is totally unsuitable in some cases (such as the mul&add example) due to including dynamic cpu feature detection -- in those cases it's better to use the version of code without the features than to ever include dynamic detection.

CAD97:
Even if `is_target_feature_detected!` is handled by the backend (thus can be perfectly reliably hoisted after inlining), there are still two distinct desires between "dispatch based on the runtime availability of a target feature" (`_detected!`) and "choose based on the codegen availability of a target feature" (`_enabled!`) that motivates two different ways of asking the question. (Static availability is also a legitimate query, but given querying codegen availability as an option, mostly just to statically cull out unnecessary code paths.)

The point of using `_enabled!` is that the code using it would prefer using the disabled path instead of spending any real effort to use the enabled path. It's providing the enabled path as preferable when the target feature is available, but explicitly saying that taking the disabled path "spuriously" (i.e. even though the feature is available) is permissible.

It doesn't really matter how the feature is implemented (e.g. a post-inlining decision, transparent function multiversioned dispatch, etc), it's deliberately *supposed to be* best-effort, and perhaps *not even that*; code using `_enabled!` is *specifically* saying that it's *preferable* to take the more general option than to spend too much effort trying to take the specific option when that could block other optimizations. It's basically the dev asking to non-deterministically take one of the arms while additionally providing a heuristic to use to determine which is likely more performant.

You can fairly accurately model `if is_target_feature_enabled!(abc)` as `if nondet_with_preference(true) && is_target_feature_detected!(abc)` (assuming backend knowledge of `_detected!`). You'd lose some developer intent and a desired[^1] performance property (guaranteed absence of introducing a dynamic check), but it (IIUC) models the intended semantics of "just choose the best codegen, either option is acceptable."

[^1]: In full truth, especially with the optimizer knowing the memory location to read for `_detected!` (or conservative) instead of needing to run `cpuid`, it could potentially determine that introducing the `_detected!` dispatch (as is done manually in the main example) is a desirable optimization without penalizing the unavailable case too much. It seems rare for the optimizer to make that choice, but I think retaining that permission is reasonable.

If we want to get *even further* into micro-opt, the ability to query "`is_target_feature_already_detected!`" (`_detected!` with `false` instead of cache initialization as a fallback) could also be useful. (However, convincing LLVM that cache initialization happens max once per thread achieves most of the benefit.)

nagisa:
If we relied on rustc’s own inlining for this: making these detection macros a compiler built-in would be an option. It can be then lowered at some point past the inlining+whatever other prerequisite optimisations phase. That’s basically 

If we implemented something in LLVM for this, we’d emit a cpuid-like intrinsic which we’d then rely on LLVM to remove at some point if it noticed it being called in a function with a corresponding `target-feature` attribute. No other changes to the implementation of the macro would be necessary in this case, I believe.

As the RFC acknowledges, and the comment above reiterates, neither of these can guarantee the behaviour you desire. As soon as any reliance on optimization comes into a play its just that – an optimization.

Now that I have written that, incorporating this sort of an optimization into the existing `is_*_feature_detected` is probably the only valid option, I suspect. Otherwise we’re looking at having non-idempotent optimizations in the compiler which sounds like a recipe for disaster (we’ve had soundness bugs related to that on multiple occasions so far.)

programmerjake:
imho it's perfectly fine for `is_*_feature_enabled` to be optimization-dependent, but critically it must be a compile-time constant -- even a load from already detected cpu features and branch can be too expensive. this is exactly how `1.0f32.copysign(v / 0.0)` for `v == 0.0` gives different answers depending on optimization on x86_64, except more useful.

https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=7e146c90c1327ba8deef84b29e9c40bf
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7e146c90c1327ba8deef84b29e9c40bf

programmerjake:
to avoid optimizations causing unsoundness, you just have to mark `is_*_feature_enabled` as behaving like a non-pure function, aka. every time it's called, it can conceptually return a different result. this deals with what happens when a function is optimized, then inlined into functions with a mix of target features (some with, some without the feature of interest), then `is_*_feature_enabled` is lowered to constants. basically, optimization can't duplicate `is_*_feature_enabled` and assume all copies always give the same answer.

calebzulawski:
> Now that I have written that, incorporating this sort of an behaviour into the existing `is_*_feature_detected` is probably the only valid option, I suspect. Otherwise we’re looking at having non-idempotent optimizations in the compiler which sounds like a recipe for disaster (we’ve had soundness bugs related to that on multiple occasions so far.)

In my opinion, this isn't especially useful (though worth implementing if this RFC is adopted).  In many ways the intention of this RFC is "optimization", e.g. using a faster instruction to do an operation, but it's also more than that.  In a "hot" path, such as a loop over SIMD, it's dramatically faster to simply use a worse instruction rather than include even a single branch.  The `is_{arch}_feature_detected` macros will always be inappropriate for such use cases even with this optimization, because the fallback branch is never worth it.  This RFC opens the door to a new pattern currently not possible in Rust.

I agree that this is _trickier_ to use correctly than `is_{arch}_feature_detected`, which is why I think it deserves its own macro with well-documented semantics.  I don't think it's necessarily dangerous, however, especially since this new macro pairs nicely with `std::simd` and may involve no `unsafe` whatsoever.  This is perhaps a stretch, but I think [`compare_exchange_weak`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.compare_exchange_weak) is a decent comparison: the result can fail spuriously to enable additional optimization, and it requires a little extra care to use correctly, but is also well documented.  I suppose the `is_{arch}_feature_enabled` macro could return something like `Some<bool>` or `Result<bool, _>` rather than spuriously returning `false` if not completely inlined, but I don't immediately see a practical way of implementing that.

PR #3451, opened 2023-06-28
Additional float types

ecnelises:
[Rendered](https://github.com/ecnelises/rust-rfcs/blob/additional-floats/text/3451-additional-float-types.md)

### DIFF
--- a/text/3451-additional-float-types.md
+++ b/text/3451-additional-float-types.md
@@ -6,19 +6,19 @@
 # Summary
 [summary]: #summary
 
-This RFC proposes new floating point types `f16` and `f128` into core language and standard library. Also this RFC introduces `f80`, `doubledouble`, `bf16` into `core::arch` for inter-op with existing native code.
+This RFC proposes new floating point types `f16` and `f128` into core language and standard library. Also, this RFC introduces `f80`, `f64f64`, and `bf16` into `core::arch` for target-specific support, and `core::ffi::c_longdouble` for FFI interop.
 
 # Motivation
 [motivation]: #motivation
 
-IEEE-754 standard defines binary floating point formats, including binary16, binary32, binary64 and binary128. The binary32 and binary64 correspond to `f32` and `f64` types in Rust, while binary16 and binary128 are used in multiple scenarios (machine learning, scientific computing, etc.) and accepted by some modern architectures (by software or hardware).
+[IEEE-754] standard defines binary floating point formats, including `binary16`, `binary32`, `binary64` and `binary128`. `binary32` and `binary64` correspond to `f32` and `f64` types in Rust, but there is currently no representation for `binary16` or `binary128`; these have uses in multiple scenarios (machine learning, scientific computing, etc.) and accepted by some modern architectures (by software or hardware), so this RFC proposes to add representations for them to the language.
 
-In C/C++ world, there're already types representing these formats, along with more legacy non-standard types specific to some platform. Introduce them in a limited way would help improve FFI against such code.
+In C/C++ world, there are already types representing these formats, along with more legacy non-standard types specific to some platform. Introduce them in a limited way would help improve FFI against such code.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-`f16` and `f128` are primitive floating types, they can be used just like `f32` or `f64`. They always conform to binary16 and binary128 format defined in IEEE-754, which means size of `f16` is always 16-bit, and size of `f128` is always 128-bit.
+`f16` and `f128` are primitive floating types, they can be used just like `f32` or `f64`. They always conform to the `binary16` and `binary128` formats defined in [IEEE-754], which means size of `f16` is always 16-bit, and size of `f128` is always 128-bit.
 
 ```rust
 let val1 = 1.0; // Default type is still f64
@@ -31,7 +31,9 @@ println!("Size of f128 in bytes: {}", std::mem::size_of_val(&val2)); // 16
 println!("Size of f16 in bytes: {}", std::mem::size_of_val(&val3)); // 2
 ```
 
-Because not every target supports `f16` and `f128`, compiler provides conditional guards for them:
+`f16` and `f128` will only be available on hardware that supports or natively emulates these type via LLVM's `half` and `fp128`, as mentioned in the [LLVM reference for floating types]. This means that the semantics of `f16` and `f128` are fixed as IEEE compliant in every supported platform, different from `long double` in C.
+
+Because not every target supports `f16` and `f128`, compiler provides conditional guards.
 
 ```rust
 #[cfg(target_has_f128)]
@@ -43,14 +45,16 @@ fn get_f16() -> f16 { 1.0f16 }
 
 All operators, constants and math functions defined for `f32` and `f64` in core, are also defined for `f16` and `f128`, and guarded by respective conditional guards.
 
-`f80` type is defined in `core::arch::{x86, x86_64}`. `doubledouble` type is defined in `core::arch::{powerpc, powerpc64}`. `bf16` type is defined in `core::arch::{arm, aarch64, x86, x86_64}`. They do not have literal representation.
+- The `f80` type is defined in `core::arch::{x86, x86_64}` as 80-bit extended precision.
+- The `f64f64` type is defined in `core::arch::{powerpc, powerpc64}` and represent's PowerPC's non-IEEE double-double format (two `f64`s used to aproximate `f128`).
+- `bf16` type is defined in `core::arch::{arm, aarch64, x86, x86_64}` and  represents the "brain" float, a truncated `f32` with SIMD support on some hardware. These types do not have literal representation.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
 ## `f16` type
 
-`f16` consists of 1 bit of sign, 5 bits of exponent, 10 bits of mantissa.
+`f16` consists of 1 bit of sign, 5 bits of exponent, 10 bits of mantissa. It is always in accordance with [IEEE-754].
 
 The following `From` and `TryFrom` traits are implemented for conversion between `f16` and other types:
 
@@ -70,7 +74,7 @@ impl From<i8> for f16 { /* ... */ }
 
 `f128` is available for on targets having (1) hardware instructions or software emulation for 128-bit float type; (2) backend support for `f128` type on the target; (3) essential target features enabled (if any).
 
-The list of targets supporting `f128` type may change over time. Initially, it includes `powerpc64le-*`.
+The list of targets supporting `f128` type may change over time. Initially, it includes `powerpc64le-*`, `x86_64-*` and `aarch64-*`
 
 The following traits are also implemented for conversion between `f128` and other types:
 
@@ -91,14 +95,24 @@ impl From<i64> for f128 { /* ... */ }
 
 `f128` will generate `fp128` type in LLVM IR.
 
-
-`std::simd` defines new vector types with `f16` or `f128` element: `f16x2` `f16x4` `f16x8` `f16x16` `f16x32` `f128x2` `f128x4`.
-
-For `doubledouble` type, conversion intrinsics are available under `core::arch::{powerpc, powerpc64}`. For `f80` type, conversion intrinsics are available under `core::arch::{x86, x86_64}`.
+For `f64f64` type, conversion intrinsics are available under `core::arch::{powerpc, powerpc64}`. For `f80` type, conversion intrinsics are available under `core::arch::{x86, x86_64}`.
 
 ## Architectures specific types
 
-As for non-standard types, `f80` generates `x86_fp80`, `doubledouble` generates `ppc_fp128`, `bf16` generates `bfloat`.
+- `core::arch::{x86, x86_64}::f80` generates LLVM's `x86_fp80`, 80-bit extended precision
+- `core::arch::{powerpc, powerpc64}::f64f64` generates LLVM's `ppc_fp128`, a `f128` emulated type via dual `f64`s
+- `core::arch::{arm, aarch64, x86, x86_64}::bf16` generates LLVM's `bfloat`, 16-bit "brain" floats used in AVX and ARMv8.6-A
+
+Where possible, `From` will be implemented to convert `f80` and `f64f64` to `f128`.
+
+## FFI types
+
+`core::ffi::c_longdouble` will always represent whatever `long double` does in C. Rust will defer to the compiler backend (LLVM) for what exactly this represents, but it will approximately be:
+
+- 80-bit extended precision (f80) on `x86` and `x86_64`:
+- `f64` double precision with MSVC
+- `f128` quadruple precision on AArch64
+- `f64f64` on PowerPC
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -129,11 +143,14 @@ We have a previous proposal on `f16b` type to represent `bfloat16`: https://gith
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-This proposal does not introduce `c_longdouble` type for FFI, because it means one of `f128`, `doubledouble`, `f64` or `f80` on different cases. Also for `c_float128`.
+This proposal does not introduce `c_longdouble` type for FFI, because it means one of `f128`, `f64f64`, `f64` or `f80` on different cases. Also for `c_float128`.
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
-More functions will be added to those platform dependent float types, like casting between `f128` and `doubledouble`.
+More functions will be added to those platform dependent float types, like casting between `f128` and `f64f64`.
 
 For targets not supporting `f16` or `f128`, we may be able to introduce a 'limited mode', where the types are not fully functional, but user can load, store and call functions with such arguments.
+
+[LLVM reference for floating types]: https://llvm.org/docs/LangRef.html#floating-point-types
+[IEEE-754]: https://en.wikipedia.org/wiki/IEEE_754
--- a/text/3451-additional-float-types.md
+++ b/text/3451-additional-float-types.md
@@ -1,6 +1,6 @@
 - Feature Name: `additional-float-types`
 - Start Date: 2023-6-28
-- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- RFC PR: [rust-lang/rfcs#3451](https://github.com/rust-lang/rfcs/pull/3451)
 - Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
 
 # Summary
--- a/text/0000-additional-float-types.md
+++ b/text/0000-additional-float-types.md
@@ -1,0 +1,139 @@
+- Feature Name: `additional-float-types`
+- Start Date: 2023-6-28
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes new floating point types `f16` and `f128` into core language and standard library. Also this RFC introduces `f80`, `doubledouble`, `bf16` into `core::arch` for inter-op with existing native code.
+
+# Motivation
+[motivation]: #motivation
+
+IEEE-754 standard defines binary floating point formats, including binary16, binary32, binary64 and binary128. The binary32 and binary64 correspond to `f32` and `f64` types in Rust, while binary16 and binary128 are used in multiple scenarios (machine learning, scientific computing, etc.) and accepted by some modern architectures (by software or hardware).
+
+In C/C++ world, there're already types representing these formats, along with more legacy non-standard types specific to some platform. Introduce them in a limited way would help improve FFI against such code.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+`f16` and `f128` are primitive floating types, they can be used just like `f32` or `f64`. They always conform to binary16 and binary128 format defined in IEEE-754, which means size of `f16` is always 16-bit, and size of `f128` is always 128-bit.
+
+```rust
+let val1 = 1.0; // Default type is still f64
+let val2: f128 = 1.0;
+let val3: f16 = 1.0;
+let val4 = 1.0f128; // Suffix of f128 literal
+let val5 = 1.0f16; // Suffix of f16 literal
+
+println!("Size of f128 in bytes: {}", std::mem::size_of_val(&val2)); // 16
+println!("Size of f16 in bytes: {}", std::mem::size_of_val(&val3)); // 2
+```
+
+Because not every target supports `f16` and `f128`, compiler provides conditional guards for them:
+
+```rust
+#[cfg(target_has_f128)]
+fn get_f128() -> f128 { 1.0f128 }
+
+#[cfg(target_has_f16)]
+fn get_f16() -> f16 { 1.0f16 }
+```
+
+All operators, constants and math functions defined for `f32` and `f64` in core, are also defined for `f16` and `f128`, and guarded by respective conditional guards.
+
+`f80` type is defined in `core::arch::{x86, x86_64}`. `doubledouble` type is defined in `core::arch::{powerpc, powerpc64}`. `bf16` type is defined in `core::arch::{arm, aarch64, x86, x86_64}`. They do not have literal representation.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## `f16` type
+
+`f16` consists of 1 bit of sign, 5 bits of exponent, 10 bits of mantissa.
+
+The following `From` and `TryFrom` traits are implemented for conversion between `f16` and other types:
+
+```rust
+impl From<f16> for f32 { /* ... */ }
+impl From<f16> for f64 { /* ... */ }
+impl From<bool> for f16 { /* ... */ }
+impl From<u8> for f16 { /* ... */ }
+impl From<i8> for f16 { /* ... */ }
+```
+
+`f16` will generate `half` type in LLVM IR.
+
+## `f128` type
+
+`f128` consists of 1 bit of sign, 15 bits of exponent, 112 bits of mantissa.
+
+`f128` is available for on targets having (1) hardware instructions or software emulation for 128-bit float type; (2) backend support for `f128` type on the target; (3) essential target features enabled (if any).
+
+The list of targets supporting `f128` type may change over time. Initially, it includes `powerpc64le-*`.
+
+The following traits are also implemented for conversion between `f128` and other types:
+
+```rust
+impl From<f16> for f128 { /* ... */ }
+impl From<f32> for f128 { /* ... */ }
+impl From<f64> for f128 { /* ... */ }
+impl From<bool> for f128 { /* ... */ }
+impl From<u8> for f128 { /* ... */ }
+impl From<i8> for f128 { /* ... */ }
+impl From<u16> for f128 { /* ... */ }
+impl From<i16> for f128 { /* ... */ }
+impl From<u32> for f128 { /* ... */ }
+impl From<i32> for f128 { /* ... */ }
+impl From<u64> for f128 { /* ... */ }
+impl From<i64> for f128 { /* ... */ }
+```
+
+`f128` will generate `fp128` type in LLVM IR.
+
+
+`std::simd` defines new vector types with `f16` or `f128` element: `f16x2` `f16x4` `f16x8` `f16x16` `f16x32` `f128x2` `f128x4`.
+
+For `doubledouble` type, conversion intrinsics are available under `core::arch::{powerpc, powerpc64}`. For `f80` type, conversion intrinsics are available under `core::arch::{x86, x86_64}`.
+
+## Architectures specific types
+
+As for non-standard types, `f80` generates `x86_fp80`, `doubledouble` generates `ppc_fp128`, `bf16` generates `bfloat`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Unlike f32 and f64, although there are platform independent implementation of supplementary intrinsics on these types, not every target support the two types natively, with regards to the ABI. Adding them will be a challenge for handling different cases.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+There are some crates aiming for similar functionality:
+
+- [f128](https://github.com/jkarns275/f128) provides binding to `__float128` type in GCC.
+- [half](https://github.com/starkat99/half-rs) provides implementation of binary16 and bfloat16 types.
+
+However, besides the disadvantage of usage inconsistency between primitive type and type from crate, there are still issues around those bindings.
+
+The availablity of additional float types depends on CPU/OS/ABI/features of different targets heavily. Evolution of LLVM may also unlock possibility of the types on new targets. Implementing them in compiler handles the stuff at the best location.
+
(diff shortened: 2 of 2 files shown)


lygstate:
> [Rendered](https://github.com/ecnelises/rust-rfcs/blob/additional-floats/text/0000-additional-float-types.md)

This files should be exist

lygstate:
How to define literal for 
`bf16` and `doubledouble`

Suggest use `f64x2` to replace doubledouble
and `f64x2` can be used as literal suffix.

clarfonthey:
I'm okay with this for the most part, except that I disagree that `f16` and `f128` should only be limited to targets that support them natively. I think that having target features for float support is a good idea, especially considering how it would help the current situation with floating-point operations in libcore, but I don't think that actually using these types should be gated behind hardware support.

A good example of this is `u128`, which is not natively supported on many platforms, but still available regardless of target. We can still emulate it, and we can still use it, therefore we can define methods for it. IMHO, gating these types behind target features just adds unnecessary boilerplate for libraries wanting to implement methods for them, and people who are writing performance-sensitive code will *already* take into account whether these types are fast on the relevant targets before using them.

lygstate:
> I'm okay with this for the most part, except that I disagree that `f16` and `f128` should only be limited to targets that support them natively. I think that having target features for float support is a good idea, especially considering how it would help the current situation with floating-point operations in libcore, but I don't think that actually using these types should be gated behind hardware support.
> 
> A good example of this is `u128`, which is not natively supported on many platforms, but still available regardless of target. We can still emulate it, and we can still use it, therefore we can define methods for it. IMHO, gating these types behind target features just adds unnecessary boilerplate for libraries wanting to implement methods for them, and people who are writing performance-sensitive code will _already_ take into account whether these types are fast on the relevant targets before using them.
> 
> Additionally, while I don't feel _too_ strongly about it, I don't like the idea of naming the x86 "long double" type `f80`, since it implies that it's also part of the IEEE 754 standard when it's not. I'm not too against it, though, since it is specific to `std::arch`.

`f80` is a lot shorter, easy for literal

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+let val2: f128 = 1.0;
+let val3: f16 = 1.0;
+let val4 = 1.0f128; // Suffix of f128 literal
+let val5 = 1.0f16; // Suffix of f16 literal
+
+println!("Size of f128 in bytes: {}", std::mem::size_of_val(&val2)); // 16
+println!("Size of f16 in bytes: {}", std::mem::size_of_val(&val3)); // 2
+```
+
+Because not every target supports `f16` and `f128`, compiler provides conditional guards for them:
```
tgross35, line 34:
```suggestion
`f16` and `f128` will only be available on hardware that supports or natively emulates these type via
LLVM's `half` and `fp128` types as mentioned in the [LLVM floating point] documentation. This. means
that `f128` will never be represented as an 80-bit type, which is the case for C's `long double`.

Because not every target supports `f16` and `f128`, compiler provides conditional guards for them:
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
+- Feature Name: `additional-float-types`
+- Start Date: 2023-6-28
+- RFC PR: [rust-lang/rfcs#3451](https://github.com/rust-lang/rfcs/pull/3451)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes new floating point types `f16` and `f128` into core language and standard library. Also this RFC introduces `f80`, `doubledouble`, `bf16` into `core::arch` for inter-op with existing native code.
```
tgross35, line 9:
```suggestion
This RFC proposes new floating point types `f16` and `f128` into core language and standard
library. Also, this RFC introduces `f80`, `doubledouble`, and `bf16` into `core::arch` for
target-specific support, and `core::ffi::c_longdouble` for FFI interop.
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+#[cfg(target_has_f128)]
+fn get_f128() -> f128 { 1.0f128 }
+
+#[cfg(target_has_f16)]
+fn get_f16() -> f16 { 1.0f16 }
+```
+
+All operators, constants and math functions defined for `f32` and `f64` in core, are also defined for `f16` and `f128`, and guarded by respective conditional guards.
+
+`f80` type is defined in `core::arch::{x86, x86_64}`. `doubledouble` type is defined in `core::arch::{powerpc, powerpc64}`. `bf16` type is defined in `core::arch::{arm, aarch64, x86, x86_64}`. They do not have literal representation.
```
tgross35, line 46:
```suggestion
The `f80` type is defined in `core::arch::{x86, x86_64}` as 80-bit extended precision. The `doubledouble`
type is defined in `core::arch::{powerpc, powerpc64}` and represent's PowerPC's non-IEEE double-double
format (two `f64`s used to aproximate `f128`). `bf16` type is defined in `core::arch::{arm, aarch64, x86, x86_64}` and  represents the "brain" float, a truncated `f32` with SIMD support on some hardware. These
types do not have literal representation.

When working with FFI, the `core::ffi::c_longdouble` type can be used to match whatever type
`long double` represents in C.
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## `f16` type
+
+`f16` consists of 1 bit of sign, 5 bits of exponent, 10 bits of mantissa.
```
tgross35, line 53:
```suggestion
`f16` consists of 1 bit of sign, 5 bits of exponent, 10 bits of mantissa. It is always
in accordance with [IEEE 754].
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+`f128` will generate `fp128` type in LLVM IR.
+
+
+`std::simd` defines new vector types with `f16` or `f128` element: `f16x2` `f16x4` `f16x8` `f16x16` `f16x32` `f128x2` `f128x4`.
+
+For `doubledouble` type, conversion intrinsics are available under `core::arch::{powerpc, powerpc64}`. For `f80` type, conversion intrinsics are available under `core::arch::{x86, x86_64}`.
+
+## Architectures specific types
+
+As for non-standard types, `f80` generates `x86_fp80`, `doubledouble` generates `ppc_fp128`, `bf16` generates `bfloat`.
```
tgross35, line 101:
```suggestion
## Architectures specific types

- `core::arch::{x86, x86_64}::f80` generates LLVM's `x86_fp80`, 80-bit extended precision
- `core::arch::{powerpc, powerpc64}::doubledouble` generates LLVM's `ppc_fp128`, a `f128` emulated
    type via dual `f64`s
- `core::arch::{arm, aarch64, x86, x86_64}bf16` generates LLVM's `bfloat`, 16-bit "brain" floats
    used in AVX and ARMv8.6-A
    
Where possible, `From` will be implemented to convert `f80` and `doubledouble` to `f128`.

## FFI types

`core::ffi::c_longdouble` will always represent whatever `long double` does in C. Rust will defer to the
compiler backend (LLVM) for what exactly this represents, but it will approximately be:

- 80-bit extended precision (f80) on `x86` and `x86_64`: 
- `f64` double precision with MSVC
- `f128` quadruple precision on AArch64
- `doubledouble` on PowerPC
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+[unresolved-questions]: #unresolved-questions
+
+This proposal does not introduce `c_longdouble` type for FFI, because it means one of `f128`, `doubledouble`, `f64` or `f80` on different cases. Also for `c_float128`.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+More functions will be added to those platform dependent float types, like casting between `f128` and `doubledouble`.
+
+For targets not supporting `f16` or `f128`, we may be able to introduce a 'limited mode', where the types are not fully functional, but user can load, store and call functions with such arguments.
```
tgross35, line 139:
```suggestion
For targets not supporting `f16` or `f128`, we may be able to introduce a 'limited mode', where the types are not fully functional, but user can load, store and call functions with such arguments.

[LLVM floating point]: https://llvm.org/docs/LangRef.html#floating-point-types
[IEEE 754]: https://en.wikipedia.org/wiki/IEEE_754
```

Add links used elsewhere

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Summary
+[summary]: #summary
+
+This RFC proposes new floating point types `f16` and `f128` into core language and standard library. Also this RFC introduces `f80`, `doubledouble`, `bf16` into `core::arch` for inter-op with existing native code.
+
+# Motivation
+[motivation]: #motivation
+
+IEEE-754 standard defines binary floating point formats, including binary16, binary32, binary64 and binary128. The binary32 and binary64 correspond to `f32` and `f64` types in Rust, while binary16 and binary128 are used in multiple scenarios (machine learning, scientific computing, etc.) and accepted by some modern architectures (by software or hardware).
```
tgross35, line 14:
```suggestion
IEEE-754 standard defines binary floating point formats, including `binary16`, `binary32`, `binary64` and
`binary128`. `binary32` and `binary64 `correspond to `f32` and `f64` types in Rust, but there is currently
no representation for `binary16` or `binary128`; these have uses in multiple scenarios (machine learning,
scientific computing, etc.) and accepted by some modern architectures (by software or hardware), so this
RFC proposes to add representations for them to the language.
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+In C/C++ world, there're already types representing these formats, along with more legacy non-standard types specific to some platform. Introduce them in a limited way would help improve FFI against such code.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+`f16` and `f128` are primitive floating types, they can be used just like `f32` or `f64`. They always conform to binary16 and binary128 format defined in IEEE-754, which means size of `f16` is always 16-bit, and size of `f128` is always 128-bit.
```
tgross35, line 21:
```suggestion
`f16` and `f128` are primitive floating types, they can be used just like `f32` or `f64`. They always
conform to the `binary16` and `binary128` formats defined in [IEEE 754], which means size of `f16`
is always 16-bit, and size of `f128` is always 128-bit.
```

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+
+`f16` will generate `half` type in LLVM IR.
+
+## `f128` type
+
+`f128` consists of 1 bit of sign, 15 bits of exponent, 112 bits of mantissa.
+
+`f128` is available for on targets having (1) hardware instructions or software emulation for 128-bit float type; (2) backend support for `f128` type on the target; (3) essential target features enabled (if any).
+
+The list of targets supporting `f128` type may change over time. Initially, it includes `powerpc64le-*`.
```
tgross35, line 73:
```suggestion
The list of targets supporting `f128` type may change over time. Initially, it includes `powerpc64le-*`.
`x86_64-*` and `aarch64-*`
```

I don't know for sure what targets support it, but we should aim to at least support the major 64-bit CPUs
here at first

tgross35 (commented):
I added some more detail in various places, and added `core::ffi::c_longdouble` (which I'm fairly certain we would want)

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+impl From<u32> for f128 { /* ... */ }
+impl From<i32> for f128 { /* ... */ }
+impl From<u64> for f128 { /* ... */ }
+impl From<i64> for f128 { /* ... */ }
+```
+
+`f128` will generate `fp128` type in LLVM IR.
+
+
+`std::simd` defines new vector types with `f16` or `f128` element: `f16x2` `f16x4` `f16x8` `f16x16` `f16x32` `f128x2` `f128x4`.
```
programmerjake, line 95: `std::simd` should probably be removed from the RFC both because `std::simd` is still changing and because we explicitly don't support any 128-bit element types for now, iirc due to bugs in the powerpc backend. We plan to add them back in in the future. We can also add additional element types as needed. @rust-lang/project-portable-simd

tgross35:
I agree with the comments about `f80`; I think I would rather see it named `x86_fp80` (to match the LLVM intrinsic) to make it clear that it is not a standard type and to discourage use.

text/3451-additional-float-types.md
programmerjake, line 46: `bf16` is supported on a wide range of newer architectures, such as powerpc, x86, arm, and (WIP) risc-v. imho it should not be classified as architecture-specific but instead more like `f16`

lygstate:
> I agree with the comments about `f80`; I think I would rather see it named `x86_fp80` (to match the LLVM intrinsic) to make it clear that it is not a standard type and to discourage use.
> 
> A good comment from the issue thread that summed this up: [#2629 (comment)](https://github.com/rust-lang/rfcs/issues/2629#issuecomment-980443628)

A bit too long, how about f80x  x means extension
And f128x for double double

text/3451-additional-float-types.md
lygstate: Yeap bf16 should be simulated when target arch is not supported

thomcc:
Note that supporting f80 (at all) might complicate the situation being discussed in https://github.com/rust-lang/rust/pull/113053, which suggests that we might be able to manipulate the control word (to set a 53 bit mantissa) in order to get more consistent behavior for f64.

the8472:
Would it make sense to consider a subset? Especially the 16bit floats are a growing field compared to the legacy types.

text/3451-additional-float-types.md
```diff
@@ -0,0 +1,139 @@
...
+For `doubledouble` type, conversion intrinsics are available under `core::arch::{powerpc, powerpc64}`. For `f80` type, conversion intrinsics are available under `core::arch::{x86, x86_64}`.
+
+## Architectures specific types
+
+As for non-standard types, `f80` generates `x86_fp80`, `doubledouble` generates `ppc_fp128`, `bf16` generates `bfloat`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Unlike f32 and f64, although there are platform independent implementation of supplementary intrinsics on these types, not every target support the two types natively, with regards to the ABI. Adding them will be a challenge for handling different cases.
```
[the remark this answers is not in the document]
bjorn3: For riscv without hardware float support there is a defined soft-float ABI. There is not for f16/bf16. Same for x86_64. Many other architectures likely don't have a defined soft-float abi for f128 either. And as I understand it AArch64 doesn't have a soft-float abi at all as Neon support is mandatory and floats are even allowed inside the kernel unlike eg x86_64 where floats are disabled in the kernel to avoid having to save them on syscalls.

tgross35:
I started trying to get together a better collection of what platforms & features support the different types to serve as a reference when implementing and help steer choices for this RFC (we can pull the contents in at some point). The doc is editable (if logged in), please add info & links if you know of any: https://hackmd.io/@8W5l8q6-Qyyn_vKh2-L0RQ/rJpZsmcdh/edit

Also, zulip thread for this RFC: https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/f16.20.26.20f128.20RFC

programmerjake:
note that PowerISA's F128 extension has a specific instruction to aid emulation of `f80`: `xsrqpxp`

text/3451-additional-float-types.md
lygstate: Suggest use these symbols, all start with `f` prefix that consistence
- `f128`: C `_Float128`, LLVM `fp128`, GCC `__float128`
- `f16`: C `_Float16`, LLVM `half`, GCC `__fp16`
- `f16b`: C++ `std::bfloat16_t`, LLVM `bfloat`, GCC `__bf16`
- `f80x`: LLVM `x86_fp80`, GCC `__float80`
- `f128x`: LLVM `ppc_fp128`, GCC `__ibm128`

tgross35: +1 for `f16b` in favor over `bf16` for consistency, I liked that about @joshtriplett's original proposal.

I don't think we should introduce something like `f128x` - `doubledouble` or something like the GCC or LLVM types would be better IMO. Reason being, it's kind of ambiguous and specific to one architecture - PowerPC is [even moving away from it](https://fedoraproject.org/wiki/Changes/PPC64LE_Float128_Transition). Better to give it an unambigous name since it will be used relatively rarely.

programmerjake: i think we should use `bf16` rather than `f16b` since that is widely recognized whereas `f16b` isn't, and `f64_f64` instead of `f128x` since it really is 2 `f64` values and could be easily emulated on any other architecture (do not use `f64x2` since that's already used by `Simd<f64, 2>`). also `f<N>x` names are more or less defined by IEEE 754 to be wider than `N` bits, so e.g. `f64x` would be approximately any type wider than `f64` but less than `f128` such as `f80`, `f16x` could be the `f24` type used by some GPUs for depth buffers. so logically `f80x` would need to be more than 80 bits and `f128x` would need to be more than 128 bits.

lygstate: > +1 for `f16b` in favor over `bf16` for consistency, I liked that about @joshtriplett's original proposal.
> 
> I don't think we should introduce something like `f128x` - `doubledouble` or something like the GCC or LLVM types would be better IMO. Reason being, it's kind of ambiguous and specific to one architecture - PowerPC is [even moving away from it](https://fedoraproject.org/wiki/Changes/PPC64LE_Float128_Transition). Better to give it an unambigous name since it will be used relatively rarely.

`f64f64` is OK, no need the underscore looks like `doubledouble`, `f80e` instead `f80x` if `f80x` is not acceptable

aaronfranke: What about `f64x2` to indicate it's two f64 glued together?

lygstate: (do not use f64x2 since that's already used by Simd<f64, 2>)

Muon:
> Note that supporting 80-bit floats (at all) might complicate the situation being discussed in [rust-lang/rust#113053](https://github.com/rust-lang/rust/pull/113053), which suggests that we might be able to manipulate the control word (to set a 53 bit mantissa) in order to get more consistent behavior for f64.

I would not say that it complicates anything. It's just another setting of the control word. In any case, such concerns are to be addressed in LLVM, not rustc. If my suggested changes are implemented and accepted, rustc doesn't need to change anything.

aaronfranke:
@the8472 If it helped with acceptance, I think splitting `f16` and `f128` to get those merged in faster would be very welcome. There is less to be debated with those, their behavior is highly standardized by IEEE, they are desired on all platforms, and their names are obvious, so the main challenge is implementation. The others have complexity that needs debate including interop with other languages and if they should be available on all platforms.

joshtriplett:
> Also I wholeheartedly support the 80 bit support. It makes life a lot nicer working with the x87. Though I agree that the name should make clear that it's not an IEEE format.

Agreed, though I definitely think it should stay in `arch` as a target-specific type.

By contrast, I think `f16b` (or whatever we end up calling it) should be available on all targets, like `f16` and `f128`.

workingjubilee:
I agree that we should basically consider `f128` and `f16` separately from the rest, especially as "long double" has a bunch of thorny "gotchas" around its repr, e.g. on systems with the x87 FPU it is **not** guaranteed that C's "long double" is indeed the fully extended 80-bit float.

text/3451-additional-float-types.md
RalfJung, line 46: For people that have never heard of `bf16` or `doubledouble` (which I assume at 16 and 128 bits in size, respectively), it would be good to link to some sort of document explaining them, and how they differ from `f16` and `f128`, respectively.

programmerjake: nobody seems to agree on `bf16` semantics:
arm has both round as normal with subnormals supported and round to odd with subnormals not supported.
x86 has round to nearest with subnormals not supported.
powerpc has round as normal with subnormals supported.
all isas have round towards zero with subnormals supported (just `f32::to_bits(v) >> 16`).

clarfonthey:
Since a lot of people are sort of dancing around it, here's my take on a precise proposal for how to modify the RFC:

1. `f16` and `f128` should be unconditionally added to the language with the same semantics as existing `f32` and `f64`.
2. There should be a minimal alias for `c_longdouble` in the `core::ffi` module, but its semantics should be explicitly dialed back for the time being.

For `f16` and `f128` semantics, it should support the generic `{float}` literals as well as explicit suffixes. They should support basic arithmetic operations via `core`, and advanced arithmetic operations via `std`. This can be tweaked as the semantics for `f32` and `f64` are further refined.

For `c_longdouble`, we should ensure that `From` conversions always exist from `f16`, `f32`, and `f64`, which will always happen for `f128` but might additionally be required for things like the non-standard 80-bit floats. However, beyond that and ABI compatibility, I don't think anything else should be offered initially. We can eventually export all possible versions of this in `core::arch`, but for now, the only cases where it should work perfectly should be when it corresponds to `f64` or `f128`; all other cases should be effectively an opaque type as far as most users are concerned.

It should be relatively easy to add in other non-standard float types as experimental in the meantime. My proposal is that these should work very similarly to how I present `c_longdouble` here: they should have some sense of ABI compatibility with C, and `From` conversions should be added whenever possible, but no arithmetic until we nail down the semantics. People can implement their own versions with inline assembly and other unsafe code in the meantime, but at least crates can all agree on what the types are.

My suspicion for something like `bf16` is that it would be already quite useful if people just had methods to truncate `f32` values into it, since this is very fast and it's what `bf16` was designed for. This would require some sense of how we want to do "truncating conversion" in general (right now `as` is the only reasonable method, but we want to generalise that to a trait), but deciding on how those conversions work isn't necessary to simply add the type in an unstable way.

Larsouille25:
is there a real demand for f16 ? Because, yes there just 2 bytes long but f16 are not very precise

aaronfranke:
@Larsouille25 `f16` is useful when you want to store a lot of float data with minimal size, and don't need much precision. For example, floating-point HDR images can be stored in `f16`, the 10-bit mantissa being more precision bits than the standard 8-bit 0-255 range, and the 5-bit exponent being more than enough. See [EXR](https://openexr.com/en/latest/TechnicalIntroduction.html#the-half-data-type) and formats [RGBAh](https://developer.apple.com/documentation/coreimage/ciformat/1437998-rgbah), [RGh](https://developer.apple.com/documentation/coreimage/ciformat/1437729-rgh),  [Rh](https://developer.apple.com/documentation/coreimage/ciformat/1438219-rh).

Another example, quantized mesh data, for some use cases of meshes you really don't care if a vertex is a fraction of a millimeter off, and meshes have a lot of vertices, so using `f16` can half the required memory size.

clarfonthey:
Also adding onto what @aaronfranke said, in general, values between 0 and 1 are quite precise for f16s, since they have a 10-bit significand. These values not only are useful for colour, but loads of different things only need to store a float in this range and using up half the memory is quite a benefit.

aaronfranke:
I opened #3453 which is mostly a subset of this RFC, a lot of text was copied from here. It only includes `f16` and `f128`, and changes them to be available on all targets instead of conditionally available.

Nemo157:
Prior art should also mention <https://crates.io/crates/half>, which is at ~2million downloads/month, showing there is a lot of demand for `f16` (I'm not sure if there's a similar crate for `f128`).

aaronfranke:
@Nemo157 There is already a link to https://github.com/starkat99/half-rs but a crate link may be better.

Nemo157:
Ah, under alternatives. I only checked the prior art section (and I think highlighting the wide usage in the RFC would be useful).

tgross35:
> I opened #3453 which is mostly a subset of this RFC, a lot of text was copied from here. It only includes `f16` and `f128`, and changes them to be available on all targets instead of conditionally available.

With this in mind, it would probably be good to shave `f16` and `f128` out of this RFC and make this only about the compatibility types in `core::arch` / `core::ffi`. And `bf16`, wherever that may end up (side note: the link to Josh's bf16 proposal in this RFC should go to the PR https://github.com/rust-lang/rfcs/pull/2690 and not to the file on the branch).

lygstate:
@tgross35 Your link is broken, I updated a new one https://hackmd.io/8Ptau058RmGqV0ZYJ49S9A

tgross35:
Thank you, formatting fixed: https://hackmd.io/@8W5l8q6-Qyyn_vKh2-L0RQ/rJpZsmcdh

ecnelises:
> I opened #3453 which is mostly a subset of this RFC, a lot of text was copied from here. It only includes `f16` and `f128`, and changes them to be available on all targets instead of conditionally available.

I'm still doubting if it's feasible to support it on all platforms. From the view of backend (LLVM, more especially), `half` and `fp128` are not always 'implemented' or even 'defined'.

(let F be either `f16` and `f128`)

1. If target supports instructions for F, everything is okay.
2. If target ABI supports software emulation for F, everything is okay.
3. If target ABI does NOT support software emulation for F, but there's another type with same storage (size/alignment), could it be used as storage type of F (and sure, F is storage-only in such case)? This may cause ABI compatibility issue.
4. If target ABI does NOT support emulation for F, and no such type exists as storage-only equivalent. We can do nothing here.
5. If target ABI supports F, but backend (LLVM, etc.) does not implemented the feature, should compiler also have some guard about it?

lygstate:
> > I opened #3453 which is mostly a subset of this RFC, a lot of text was copied from here. It only includes `f16` and `f128`, and changes them to be available on all targets instead of conditionally available.
> 
> I'm still doubting if it's feasible to support it on all platforms. From the view of backend (LLVM, more especially), `half` and `fp128` are not always 'implemented' or even 'defined'.
> 
> (let F be either `f16` and `f128`)
> 
> 1. If target supports instructions for F, everything is okay.
> 2. If target ABI supports software emulation for F, everything is okay.
> 3. If target ABI does NOT support software emulation for F, but there's another type with same storage (size/alignment), could it be used as storage type of F (and sure, F is storage-only in such case)? This may cause ABI compatibility issue.
> 4. If target ABI does NOT support emulation for F, and no such type exists as storage-only equivalent. We can do nothing here.
> 5. If target ABI supports F, but backend (LLVM, etc.) does not implemented the feature, should compiler also have some guard about it?

don't worry, C++ already have both `f16` and `f128`, that is std::float16_t, std::float32_t, std::float64_t, std::float128_t, [std::bfloat16_t](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format)

aaronfranke:
@ecnelises This is a good question, but I disagree about this point:

> 4. If target ABI does NOT support emulation for F, and no such type exists as storage-only equivalent. We can do nothing here.

Yes, we can do something here. There is no reason we can't support our own emulation. Any calculation can be performed on a turing-complete computer. The specific details may be a challenge to implementers in some cases, but it is not impossible to support these types everywhere, therefore they should be supported everywhere.

text/3451-additional-float-types.md
ecnelises: Thanks, I did not add the mention of `longdouble` yet. More things need to be clarified:

- Is there always only 1 long double for each `(arch, abi, os)` tuple? For example, `powerpc64le-unknown-linux-gnu` can use either double or doubledouble or IEEE binary128 as `long double` by `-mabi=(ieee|ibm)longdouble` and `-mlong-double-(64|128)`.
- Is mangling of `long double` the same regardless of its underlying semantics?
- Some targets (also `powerpc64le` for example) support `.gnu_attribute`, so that linker can differentiate objects compiled by different long double ABI. Should Rust programs using `c_longdouble` emit such attribute?

ecnelises:
> @ecnelises This is a good question, but I disagree about this point:
> 
> > 4. If target ABI does NOT support emulation for F, and no such type exists as storage-only equivalent. We can do nothing here.
> 
> Yes, we can do something here. There is no reason we can't support our own emulation. Any calculation can be performed on a turing-complete computer. The specific details may be a challenge to implementers in some cases, but it is not impossible to support these types everywhere, therefore they should be supported everywhere.

If we do plan to implement the types everywhere, the 4 becomes 5. Even no `has_f128` guards need to exist, compiler should still provide useful diagnostics if the types are used but not implemented by backend yet.

> don't worry, C++ already have both `f16` and `f128`, that is std::float16_t, std::float32_t, std::float64_t, std::float128_t, [std::bfloat16_t](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format)

I think they're optional? https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1467r4.html

scottmcm:
Overall, I like the idea of separating out the real IEEE types from the random stuff, and making `f16` and `f128` not arch-specific.

> `f80` is a lot shorter, easy for literal, we can clarify that only f16 f32 f64 f128 are IEEE 754 standard.

I don't think that's important for a niche type, though.  It doesn't need a name any nicer than [`__m512i`](https://doc.rust-lang.org/nightly/core/arch/x86_64/struct.__m512i.html) and friends get.

For that matter, what's the goal of having a type for x87's weird things?  It's mentioned in neither the Motivation nor the Rationale sections.  How much new Rust code is going to be written targeting those types specifically instead of normal `double`s?  What's the size and alignment of such a type in memory?  Are modern chips putting any effort into making it work efficiently, when x87 (and MMX) tend to be avoided by modern compilers?

> `bf16` [...] These types do not have literal representation.

Why not?  It seems odd to me that `f16` would support literals but `bf16` wouldn't.

Jules-Bertholet:
> For that matter, what's the goal of having a type for x87's weird things?

One possibility I could see, is that if Rust some day decides to accept the performance hit and make `f64` fully IEEE-compliant on no-SSE x86-32, `f80` would allow code that doesn't need full compliance to recover some performance on these platforms.

tgross35:
@ecnelises would you be able to update this RFC to be about `bf16`, `ppc_fp128` and `x86_fp80`, and reference #3453 as a subset that should be developed first? Specifically, update the PR title to specifically mention those types, and add a link to the other RFC.

We can work on both RFCs in parallel but readers need to be clear about what the relationship between the two is - it's currently ambiguous.

ecnelises:
> @ecnelises would you be able to update this RFC to be about `bf16`, `ppc_fp128` and `x86_fp80`, and reference #3453 as a subset that should be developed first? Specifically, update the PR title to specifically mention those types, and add a link to the other RFC.
> 
> We can work on both RFCs in parallel but readers need to be clear about what the relationship between the two is - it's currently ambiguous.

I’d like to update this RFC if separating IEEE compliant types and target related types is considered necessary, and my original changes with updates are correctly applied in the related proposal. 😶

ecnelises:
> @ecnelises would you be able to update this RFC to be about `bf16`, `ppc_fp128` and `x86_fp80`, and reference #3453 as a subset that should be developed first? Specifically, update the PR title to specifically mention those types, and add a link to the other RFC.
> 
> We can work on both RFCs in parallel but readers need to be clear about what the relationship between the two is - it's currently ambiguous.

Thanks for your and other's comments. I generally agree that a proposal for IEEE-754 compliant `f16`/`f128` and for target specific types are quite different. So I created another one #3456, focusing on `bf16`, `f64f64`, `f80` and `c_longdouble` types.

I'm not familiar with Rust community's RFC policy. So I'm in doubt whether kicking off new post directly from another people without previous commit/author information shows enough respect to the original author.

tgross35:
> Thanks for your and other's comments. I generally agree that a proposal for IEEE-754 compliant `f16`/`f128` and for target specific types are quite different. So I created another one #3456, focusing on `bf16`, `f64f64`, `f80` and `c_longdouble` types.
> 
> I'm not familiar with Rust community's RFC policy. So I'm in doubt whether kicking off new post directly from another people without previous commit/author information shows enough respect to the original author.

There is no policy on this, but consulting before splitting off your PR would have been, well, nice. It would probably be good for the two RFC authors to come to consensus on Zulip about what the relationship among these RFCs should be, and how to harmonize them a bit (or even figure out coauthoring) to help give the community a better picture of what exactly is being proposed. There is an existing floats rfc topic that can be used for discussion https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/f16.20.26.20f128.20RFC

ecnelises:
I think we can close this since #3456 and #3453 were splitted.

PR #3454, opened 2023-07-02
Fix indentation of sub-points in the RFC process list in the README

aaronfranke:
[Rendered](https://github.com/aaronfranke/rfcs/tree/fix-readme-indent#what-the-process-is)

Based on the positions of the `-` symbols, and [how it is currently rendered](https://github.com/rust-lang/rfcs#what-the-process-is), these are sub-points. However, the text was not indented properly in 2 out of 3 of the sub-points. This PR fixes the indentation of the text in the sub-points.

Also, a stray trailing space character was removed on line 6, my text editor did this automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [Rust RFCs]: #rust-rfcs
 
 The "RFC" (request for comments) process is intended to provide a consistent
-and controlled path for changes to Rust (such as new features) so that all 
+and controlled path for changes to Rust (such as new features) so that all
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!  I believe it renders identically, as the indentation of subsequent lines isn't important if they are part of the same contiguous block. However, I can see that if you are reading the raw source it looks a little odd to be inconsistent.

PR #3459, opened 2023-07-19
Error Display (std::error::Error::fmt_error)

ijackson:
[Rendered](https://github.com/ijackson/rust-rfcs/blob/error/text/0000-error-display.md)

### DIFF
--- a/text/0000-error-display.md
+++ b/text/0000-error-display.md
@@ -1,0 +1,329 @@
+- Feature Name: separate_error_fmt
+- Start Date: 2027-07-19
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+A new `error_fmt` method on `std::error::Error`,
+so that we can distinguish:
+
+ * Requests to just display an error for human consumption
+   (`Display`)
+ * The internal implementation of printing a particular error,
+   excluding its `source`s (`error_fmt`).
+
+Transitional and compatibility arrangements to make this workable.
+
+# Motivation
+[motivation]:
+
+Correctly printing errors in Rust
+(and defining errors that print correctly)
+is too hard.
+
+We want to be able to get from where we are now
+to a situation with the following properties:
+
+ * Just printing an error with `eprintln!("{error}")` will reliably
+   do something useful.
+ * Errors can be printed in a fancy report-like style with inspection
+   of source errors, if desired.
+ * Messages, and parts of them, are not duplicated.
+ * Implementing an error type isn't significantly harder than today.
+ * Warts (induced by backward compatibility requirements) are avoided
+   as much as possible.
+
+# Guide-level explanation (synchronic - where we want to end up)
+[guide-level-explanation]: #guide-level-explanation
+
+### Background (existing situation, will not be changed by this RFC)
+
+Most errors should implement `std::error::Error`.
+
+Errors can have a "source": an underlying error which caused this one.
+That underlying error can in turn have a source,
+forming a causal chain.
+
+### Printing errors (new doctrine)
+
+Errors can be printed in two main ways:
+Every error implements `Display`
+and provides an `error_fmt` method.
+
+The `Display` implementation *does* print the source.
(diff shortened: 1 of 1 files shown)


shepmaster:
I haven't fully read the RFC, but wanted to drop some additional prior art. [`snafu::report`](https://docs.rs/snafu/latest/snafu/attr.report.html) recognizes that many errors in the ecosystem basically implement `Display` as `write!(f, "My unique error text: {source}")`, which looks terrible when trying to format a nested error as you end up with:

```
1: C: B: A
2: B: A
3: A
```

`Report` post-processes those error messages to remove the duplicated content, producing:

```
1: C
2: B
3: A
```

joshtriplett:
cc @yaahc, whose `eyre` crate has an extensive reporting mechanism.

Stargateur:
I like the concept but I don't like much the limitation of current Display and Debug trait, you can't really do any sort of user configuration, like for example, carry information like if NO_COLOR is set in the environnement (you need to use static) thus I got no idea how we could solve this.

The fmt argument I mostly ignore all its features and just use it cause it has the write item.

I asked something very similar to tracing crate https://github.com/tokio-rs/tracing/issues/2226

text/0000-error-display.md
```diff
@@ -0,0 +1,330 @@
+- Feature Name: separate_error_fmt
+- Start Date: 2027-07-19
```
Stargateur, line 2: wrong date

text/0000-error-display.md
```diff
@@ -0,0 +1,330 @@
...
+
+/// Displays `E` and all its sources; output is similar to "error: source".
+///
+/// When used with `{:#}`, prints a multi-line "caused by" chain.
+///
+/// Does ad-hoc deduplication, as follows: Records the string of the
+/// each error displayed, and suppresses printing of the source if the
+/// source error text is textually contained within the previous error
+/// text.
+default impl<E> Display for E where E: Error + ErrorDisplay { ... }
```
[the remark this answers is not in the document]
ijackson: This ought to have been deleted from here during editing, yes.  I've removed it.

text/0000-error-display.md
ijackson: Oops, thanks :-).  Fixed.

ijackson:
Thanks to everyone for the review and comments.  I've updated the RFC.  Some more detailed reponses:

@shepmaster
> I haven't fully read the RFC, but wanted to drop some additional prior art. [`snafu::report`](https://docs.rs/snafu/latest/snafu/attr.report.html) [addresses this problem]

Thanks for the reference, indeed.  I wasn't aware of that.  I think that this textual deduplication approach has been independently invented (at least) twice shows it's a viable strategy.

@Stargateur
> I like the concept but I don't like much the limitation of current Display and Debug trait, you can't really do any sort of user configuration, like for example, carry information like if NO_COLOR is set in the environnement (you need to use static) thus I got no idea how we could solve this.

Indeed, this RFC won't eliminate the desire to do that, and it doesn't help with it very much.  For example, if you want coloured error messages, you'll still need to write a reporter that prints errors in colour, and arrange to call it everywhere you print an error.

But it does make it marginally easier: at least, with this RFC, you will (eventually) be able to reliably format the individual errors without their sources being mixed in.

@oxalica
> I doubt that if this can be achieved by the alternative flag `{:#}`

I have added a new section about this to Alternatives.  As you will see from my text, I think there are serious downsides to that approach.

@joshtriplett
> cc @yaahc, whose `eyre` crate has an extensive reporting mechanism.

Yes.  One way of looking at `eyre` is provides errors with a better  `Display`,  by wrapping them up.  I have added a reference.  `eyre` makes printing sources depend on `{}` vs `{:?}`, so I mention that too.

tgross35:
GH isn't highlighting the code, could you change the fences to ` ```rust`?

kpreid:
I am glad to hear of a plan to make it easy and automatic to show `source` chains, thus relieving the tension between showing sources in `Display` (DWIM for printing and logging; makes it harder for error handling code to neglect to show the sources) and not showing sources in `Display` (enables clean structured chain printing). I don't like the result that nearly every error will automatically have the same default `Display` implementation, and that that implementation will have to _choose_ some chain formatting (the standard library mostly does not put _choices of syntax_ into `Display`, except for things like integers being base 10). But, an application can always introduce its own top-level error type with choice of `Display`, which is already common practice with e.g. `anyhow`.

PR #3464, opened 2023-07-25
Crate quarantine

LawnGnome:
/cc @rust-lang/crates-io, @rust-lang/cargo, @rust-lang/mods

### DIFF
--- a/text/0000-crate-quarantine.md
+++ b/text/0000-crate-quarantine.md
@@ -1,0 +1,157 @@
+- Feature Name: `crate-quarantine`
+- Start Date: 2023-07-25
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Add an off-by-default quarantine system to crates.io that can be used to prevent crate versions from being immediately published to the crate index based on rules defined by the crates.io team.
+
+# Motivation
+[motivation]: #motivation
+
+Packaging ecosystems aren't warm, friendly places any more. PyPI has repeatedly been forced to deal with attacks of spam[^pypi-2021-03][^pypi-2021-06] and malware[^pypi-2023-05]. In March 2023, an analysis indicated that more than half the packages uploaded to NPM were SEO spam[^npm-2023-03], and vendors such as jFrog have previously found significant numbers of malware packages on NPM[^npm-malware].
+
+So far, crates.io has been relatively fortunate to only have isolated incidents[^rustdecimal], but the increasing popularity of Rust only makes it more and more likely that we will also be the target of similar spam and malware attacks.
+
+There are mitigating factors at play: in particular, our reliance on GitHub for authentication, while problematic from purist FOSS and ecosystem diversity perspectives, means that we benefit from GitHub's machinery in terms of detecting bad actors.
+
+Regardless, we need to be able to prevent malicious packages from being uploaded, both in emergency situations (for example, mitigating spam attacks) and in more mundane cases where we can confidently identify a potentially malicious user. This RFC proposes an off-by-default quarantine approach that allows us to examine packages before they are published when it is considered warranted.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This project will add a new admin feature to crates.io that allows crate publishes matching specific properties to be placed into a quarantine state instead of being immediately published.
+
+A version in the quarantined state will have its `.crate` file published to S3 to simplify review, and will appear in crates.io API responses an an unpublished crate, but will _not_ be added to the index — and will therefore _not_ be installable via `cargo` — until it has been reviewed and approved.
+
+`cargo publish` will also be modified to report when an uploaded crate version has been placed into quarantine. 
+
+Once a crate version has been quarantined, the a member of the crates.io team will review the quarantined crate within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail, assuming crates.io has a verified e-mail available for their account.
+
+Should the crate owner(s) disagree with the quarantine action, they may appeal to the moderation team.
+
+## When will a quarantine rule be added?
+
+Ultimately, this will be decided by the crates.io team on a case by case basis. There are two primary scenarios that are currently envisioned where this functionality would be used:
+
+### Large scale spam attack
+
+In the event of a large scale spam attack on the entire site, we may want to either restrict individual users (if the spam is coming from a finite set of users) or quarantine all published versions temporarily (using a `.*` crate name rule) until it subsides, at which point we can analyse which crate versions are actually valid.
+
+### Malicious users
+
+When we have high confidence (due to user reports or automated scanning) that a user is uploading malicious crates: in this case, a rule can be added for that specific user until the crates.io team is able to communicate with them.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The initial implementation of this RFC will only allow for rules on the crate name or a specific user. The fields and metadata that a rule can match against may be expanded in the future if required for operational reasons.
+
+## crates.io
+
+Most changes will need to occur on crates.io. They are described below.
+
+### Database
+
+New tables will need to be added to capture quarantine rules on a per-crate and per-user basis. Crates will be matched by matching their name against a regular expression in each rule; users will be matched by their external user ID[^uid].
+
+The `versions` table will need a new `quarantined` boolean flag, much like the existing `yanked` flag.[^state]
+
+### Publishing
+
+The crates.io database will be expanded to store and capture rules that, if matched, will cause a crate version that is being published to enter the `quarantined` state, rather than being immediately published. These rules will be defined and maintained by crates.io admins.
+
+The publish endpoint will notify the publisher (generally `cargo publish`) of the quarantined version by extending the response[^publish-response] in two ways:
+
+1. A new field will be added to the response called `quarantined`.
+2. A human readable warning will be added to the `warnings` field within the response.
+
+When a crate version is quarantined, a support ticket will also be opened within the crates.io help desk, thereby notifying the crates.io team of the newly quarantined version.
+
+### API
+
+A new field will be added to API endpoints that return crate versions called `published`. This will only be `false` for crate versions that are in quarantine.
+
+### Admin interface
+
+The crates.io admin console will be extended to provide the following new functionality:
+
+#### Rule definition
+
+A new CRUD area will be provided for rules to be defined and managed.
+
+#### Review
+
+A new review area will be added that lists crate versions currently in quarantine, along with options to download the `.crate` file and manage the version (ie by approving or rejecting the crate version).
+
+## `cargo publish`
+
+The wait loop in `cargo publish`[^wait-loop] currently refreshes the crate index and waits for the new crate version to appear.
+
+This will be extended to check the `quarantined` flag in the response; if present and `true`, `cargo publish` will not wait and will immediately report the warning(s) in the response to the user.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Project team member time
+
+The major drawback here is time: an actual human is going to need to be able to review any quarantined crate versions, which will require both expertise and availability. A crate being quarantined will almost certainly frustrate its owner(s), no matter what communication is provided, and the more time that is required to review the crate, the worse that experience will be.
+
+In the near term, it is unlikely that quarantine will be used in any day-to-day manner. The intention is to provide a break-glass-in-case-of-emergency option more than something that is in continuous use, but if this is used more frequently in normal operation, then workload will quickly become a concern.
+
+## Denial of service vectors
+
+This proposal addresses the needs of a Rust user — who doesn't want to inadvertently download and use a compromised or malicious crate — but does little to prevent denial of service on crates.io: both technically (since most publication steps will still take place for a quarantined crate version), and socially (since a human has to review the crate version).
+
+Overall, this is still an improvement in ecosystem security (since it can be used to prevent the index including malware), but it's not an all-encompassing magic solution to all our ecosystem security woes.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The main alternative approach that would be available would be to, in essence, do nothing. Our current process is to handle malware and spam after crates are published, and if we needed to cut off a deluge of spam, we would be able to do so at the webserver level. It can be argued that this may be sufficient, given the relatively low level of incidents so far.
+
+Were we to keep an approach of dealing with issues ex post facto, we could adopt some of the work PyPI is doing around allowing trusted reporters to flag problematic crates.[^pypi-malware]
+
+Another possible approach would be to require review of all new crates before they are published, but it's unlikely we have the bandwidth to handle that within the Rust Project.
+
+If we decide to accept this RFC and provide a mechanism for attempting to quell malware and spam before crates are published, then there aren't a lot of other ways to implement this — ultimately, that decision has to be made as `cargo publish` is run, and there's no real scenario where this can be done in a purely client-side manner. (Indeed, we could technically implement this purely in crates.io, but the user experience of having `cargo publish` simply time out while waiting for the crate version to appear in the index would be suboptimal.)
+
+# Prior art
+
+Python and Node have both been notably subject to attacks in recent years.
+
+## Python
+
+PyPI has a list of prohibited words that cannot be used in package names[^pypi-prohibited], but no other quarantine or moderation functionality other than ex post facto deletion of packages in response to reports.
+
+## Node
+
+Unlike PyPI with its Warehouse project, the NPM registry isn't open source. There is, however, no indication of a pre-moderation or quarantine sy There was something of a space theme that pervaded the Embedded Linux Conference (ELC) portion of the 2023 Embedded Open Source Summit (EOSS), which is an umbrella event for various sub-conferences related to embedded open-source development. That may partly be because one of the organizers of EOSS (and ELC), Tim Bird, described himself as "a bit of a space junkie"; he made that observation during a panel session that he led on embedded Linux in space. Bird and four panelists discussed various aspects of the use of Linux in space-related systems, including where it has been used, the characteristics and challenges of aerospace deployments, certification of Linux for aerospace use, and more. stem in their public documentation.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+## Allowing trusted reporters to flag problematic users and/or crates
+
+The Python Software Foundation is starting a project to improve malware detection[^pypi-malware], with one of its key pillars being a system where trusted reporters are allowed to make reports that will feed directly into PyPI's triage system.
+
+Depending on the outcome of that project, we may want to consider a similar approach that can feed into quarantine rules.
+
+## Identifying malicious packages automatically
+
+There is considerable existing tooling available — both in the open source and commercial worlds — to identify malware in an automated manner. In the future, we may want to consider adding low latency, low cost checks to the publishing pipeline in the future that could be used to divert potentially malicious crates into quarantine without an existing user or crate rule.
+
+[^npm-2023-03]: https://blog.sandworm.dev/one-in-two-new-npm-packages-is-seo-spam-right-now
+[^npm-malware]: https://jfrog.com/blog/malware-civil-war-malicious-npm-packages-targeting-malware-authors/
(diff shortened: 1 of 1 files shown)


text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,157 @@
...
+
+Regardless, we need to be able to prevent malicious packages from being uploaded, both in emergency situations (for example, mitigating spam attacks) and in more mundane cases where we can confidently identify a potentially malicious user. This RFC proposes an off-by-default quarantine approach that allows us to examine packages before they are published when it is considered warranted.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This project will add a new admin feature to crates.io that allows crate publishes matching specific properties to be placed into a quarantine state instead of being immediately published.
+
+A version in the quarantined state will have its `.crate` file published to S3 to simplify review, and will appear in crates.io API responses an an unpublished crate, but will _not_ be added to the index — and will therefore _not_ be installable via `cargo` — until it has been reviewed and approved.
```
calebcartwright, line 27: typo:

```suggestion
A version in the quarantined state will have its `.crate` file published to S3 to simplify review, and will appear in crates.io API responses as an unpublished crate, but will _not_ be added to the index — and will therefore _not_ be installable via `cargo` — until it has been reviewed and approved.
```

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,157 @@
...
+
+`cargo publish` will also be modified to report when an uploaded crate version has been placed into quarantine. 
+
+Once a crate version has been quarantined, the a member of the crates.io team will review the quarantined crate within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail, assuming crates.io has a verified e-mail available for their account.
```
calebcartwright, line 31:
```suggestion
Once a crate version has been quarantined, then a member of the crates.io team will review the quarantined crate within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail, assuming crates.io has a verified e-mail available for their account.
```

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,157 @@
...
+
+Ultimately, this will be decided by the crates.io team on a case by case basis. There are two primary scenarios that are currently envisioned where this functionality would be used:
+
+### Large scale spam attack
+
+In the event of a large scale spam attack on the entire site, we may want to either restrict individual users (if the spam is coming from a finite set of users) or quarantine all published versions temporarily (using a `.*` crate name rule) until it subsides, at which point we can analyse which crate versions are actually valid.
+
+### Malicious users
+
+When we have high confidence (due to user reports or automated scanning) that a user is uploading malicious crates: in this case, a rule can be added for that specific user until the crates.io team is able to communicate with them.
```
calebcartwright, line 45: Mildly confusing structure, but perhaps I'm just misunderstanding, perhaps something like this?

```suggestion
When we have high confidence (due to user reports or automated scanning) that a user is uploading malicious crates then a rule can be added for that specific user until the crates.io team is able to communicate with them.
```

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,157 @@
...
+## Denial of service vectors
+
+This proposal addresses the needs of a Rust user — who doesn't want to inadvertently download and use a compromised or malicious crate — but does little to prevent denial of service on crates.io: both technically (since most publication steps will still take place for a quarantined crate version), and socially (since a human has to review the crate version).
+
+Overall, this is still an improvement in ecosystem security (since it can be used to prevent the index including malware), but it's not an all-encompassing magic solution to all our ecosystem security woes.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The main alternative approach that would be available would be to, in essence, do nothing. Our current process is to handle malware and spam after crates are published, and if we needed to cut off a deluge of spam, we would be able to do so at the webserver level. It can be argued that this may be sufficient, given the relatively low level of incidents so far.
```
calebcartwright, line 113:
```suggestion
The main alternative approach that would be available would be to, in essence, do nothing. Our current process is to handle malware and spam after crates are published, and if needed to cut off a deluge of spam, we have the ability to do so at the webserver level. It can be argued that this may be sufficient, given the relatively low level of incidents so far.
```

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Python
+
+PyPI has a list of prohibited words that cannot be used in package names[^pypi-prohibited], but no other quarantine or moderation functionality other than ex post facto deletion of packages in response to reports.
+
+## Node
+
+Unlike PyPI with its Warehouse project, the NPM registry isn't open source. There is, however, no indication of a pre-moderation or quarantine sy There was something of a space theme that pervaded the Embedded Linux Conference (ELC) portion of the 2023 Embedded Open Source Summit (EOSS), which is an umbrella event for various sub-conferences related to embedded open-source development. That may partly be because one of the organizers of EOSS (and ELC), Tim Bird, described himself as "a bit of a space junkie"; he made that observation during a panel session that he led on embedded Linux in space. Bird and four panelists discussed various aspects of the use of Linux in space-related systems, including where it has been used, the characteristics and challenges of aerospace deployments, certification of Linux for aerospace use, and more. stem in their public documentation.
```
calebcartwright, line 131: Possible incomplete sentence?

```suggestion
Unlike PyPI with its Warehouse project, the NPM registry isn't open source. There is, however, no indication of a pre-moderation or quarantine system. There was something of a space theme that pervaded the Embedded Linux Conference (ELC) portion of the 2023 Embedded Open Source Summit (EOSS), which is an umbrella event for various sub-conferences related to embedded open-source development. That may partly be because one of the organizers of EOSS (and ELC), Tim Bird, described himself as "a bit of a space junkie"; he made that observation during a panel session that he led on embedded Linux in space. Bird and four panelists discussed various aspects of the use of Linux in space-related systems, including where it has been used, the characteristics and challenges of aerospace deployments, certification of Linux for aerospace use, and more. stem in their public documentation.
```

LawnGnome: Looks like tap to click has completely messed me up there. Thanks, will fix.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+In the event of a large scale spam attack on the entire site, we may want to either restrict individual users (if the spam is coming from a finite set of users) or quarantine all published versions temporarily (using a `.*` crate name rule) until it subsides, at which point we can analyse which crate versions are actually valid.
+
+### Malicious users
+
+When we have high confidence (due to user reports or automated scanning) that a user is uploading malicious crates then a rule can be added for that specific user until the crates.io team is able to communicate with them.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The initial implementation of this RFC will only allow for rules on the crate name or a specific user. The fields and metadata that a rule can match against may be expanded in the future if required for operational reasons.
```
kpreid, line 49: A small concern: This paragraph, together with the fact that this RFC discusses two rule types specifically, could be taken to imply that future quarantine rule types will also go through the RFC process. I would suggest stating that further rule types may be added by the crates.io team as they see fit. This ensures that there is no question that they have permission to build new tools to respond to new attacks.

(The part that matters as a public decision, that non-team-members have to worry about, is how quarantining is presented in the APIs/UIs. That won't change in the presence of new rule types.)

tarcieri:
I've been interested in a similar feature for internal use within a team of crates.io end users that releases crates: they could be configured to be immediately quarantined upon publication until approved by one (or potentially `k` of `n`) additional publisher beyond the person who published it before it goes live.

Such a 2-person (or more) system would prevent the compromise of a single publisher's credentials from translating directly to publication of a malicious crate. This would be useful for high-value crates. It would also help secure systems where crates are published by a release automation, then approved by a human/`k` human(s).

I'm curious if the design of the quarantine system could have an associated policy about how a quarantined crate can be published, i.e. approval of other crate publishers versus approval by crates.io administrators.

tarcieri:
@epage if a quarantine system were available to end-users, it could be used to implement https://github.com/rust-lang/cargo/issues/1169.

You could stage all of the releases in quarantine, then set them live at the same time.

epage:
Yes, we could have sandboxed indexes that these quarantined packages go to.

That requires a lot of more design and implementation work than this RFC proposes.

tarcieri:
I think all it requires out of the gate is attaching an AuthZ policy about how a crate can be un-quarantined/published to each quarantine state.

Arguably if you only have one AuthZ policy you don’t need that. It can be backfilled later.

I just hope you can keep the design flexible enough that it could potentially accommodate these features in the future.

epage:
As a future possibility, I think its worth it.  We just need to recognize that is likely not a simple addition on top of this.   The design work alone would need to examine the whole publish pipeline and address parts at each stage (`cargo publish`, crates.io API, crates.io index and `.crate` storage, the local index and crate storage, etc).

text/0000-crate-quarantine.md
LawnGnome: Thanks — that was basically the intent anyway, but you're right that it wasn't clear enough.

LawnGnome:
> I've been interested in a similar feature for internal use within a team of crates.io end users that releases crates: they could be configured to be immediately quarantined upon publication until approved by one (or potentially `k` of `n`) additional publisher beyond the person who published it before it goes live.

That definitely sounds like an interesting area to explore! Consensus based peer approval is something that might go very well with distributed reviews in `cargo vet`, or other similar tooling. And, of course, one of my concerns in this RFC is workload on the crates.io team, so being able to distribute that more widely would be beneficial in the long run if this mechanism is used for more than just dealing with immediate spam attacks and bad actors.

That said, as @epage said, I'd prefer to keep this RFC focused on those more prosaic concerns right now — I don't believe anything in this RFC would prevent future work (and, indeed, simply decoupling the concept of a crate version being available in the index from running `cargo publish` will help significantly there), but this is largely around getting the crates.io team the tooling we need today, rather than trying to design a full review/quarantine system that is all things to all people for the future.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+
+Ultimately, this will be decided by the crates.io team on a case by case basis. There are two primary scenarios that are currently envisioned where this functionality would be used:
+
+### Large scale spam attack
+
+In the event of a large scale spam attack on the entire site, we may want to either restrict individual users (if the spam is coming from a finite set of users) or quarantine all published versions temporarily (using a `.*` crate name rule) until it subsides, at which point we can analyse which crate versions are actually valid.
+
+### Malicious users
+
+When we have high confidence (due to user reports or automated scanning) that a user is uploading malicious crates then a rule can be added for that specific user until the crates.io team is able to communicate with them.
```
joshtriplett, line 44: This phrasing implies that you will always use this mechanism, rather than (for instance) banning a malicious user. You may not want to pre-commit to that as the required action; doing so would remove flexibility.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This project will add a new admin feature to crates.io that allows crate publishes matching specific properties to be placed into a quarantine state instead of being immediately published.
+
+A version in the quarantined state will have its `.crate` file published to S3 to simplify review, and will appear in crates.io API responses as an unpublished crate, but will _not_ be added to the index — and will therefore _not_ be installable via `cargo` — until it has been reviewed and approved.
+
+`cargo publish` will also be modified to report when an uploaded crate version has been placed into quarantine. 
+
+Once a crate version has been quarantined, then a member of the crates.io team will review the quarantined crate within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail, assuming crates.io has a verified e-mail available for their account.
```
joshtriplett, line 30:
```suggestion
Once a crate version has been quarantined, then a member of the crates.io team will review the quarantined crate, ideally within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail, assuming crates.io has a verified e-mail available for their account.
```

I don't think this should be a hard commit.

text/0000-crate-quarantine.md
LawnGnome: Definitely not the intent, thanks! Slight clarification applied.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+### Publishing
+
+The crates.io database will be expanded to store and capture rules that, if matched, will cause a crate version that is being published to enter the `quarantined` state, rather than being immediately published. These rules will be defined and maintained by crates.io admins.
+
+The publish endpoint will notify the publisher (generally `cargo publish`) of the quarantined version by extending the response[^publish-response] in two ways:
+
+1. A new field will be added to the response called `quarantined`.
+2. A human readable warning will be added to the `warnings` field within the response.
+
+When a crate version is quarantined, a support ticket will also be opened within the crates.io help desk, thereby notifying the crates.io team of the newly quarantined version.
```
teor2345, line 70: If there are a large number of crates, the number of tickets could also represent a spam or denial of service attack on the help desk system. So I'd suggest:

```suggestion
When a crate version is quarantined, a support ticket will also be opened within the crates.io help desk, thereby notifying the crates.io team of the newly quarantined version. if additional crates are quarantined for that user or regex, a comment will be added to the existing open ticket. (This prevents the help desk being spammed.)
```

LawnGnome: I like that, thanks!

teor2345: The other alternative is updating the existing ticket description. But I don't think the difference matters much at an RFC level.

LawnGnome: Agreed.

ijackson:
Under this proposal, quarantining a crate could very seriously complicate publication of a multi-crate workspace (or other kind of "deep stack").  Rust wants us to make smaller crates for separate compilation reasons, but cargo then insists that these crates which can be conceptually part of a single program must be published separately.  For release management to be practical, there must be a way to publish a workspace even if some of the involved crates end up quarantined.

This does not seem to be dicussed at all in the RFC.  IMO it should be.

Solutions to this problem that occurred to me:

 * Allow publication to refer to quarantined dependencies, linking the new crate to the quarantined of those dependencies.  (This seems like it would be a lot of work.)
 * Allow overriding of the "dependencies actually exist" check.  The user would be expected to retry the publication of the failed crate(s) with the new flag.  (This seems easy, but the resulting UX is poor.)
 * Allow overriding of the "dependencies actually exist" check.  Automatically apply this when a dependency doesn't exist but might be be resolvable using quarantined uploads.  (This needs a way to check for the existence of a quarantined crate.)
 * Do the server side "dependencies actually exist" check *after* quarantine review,; automatically quarantine any crate that has a quarantined dependency.  (This may be doable; the UX will be poor, since the user would  need to redo with `--no-verify`)
 * Define circumastances where we promise not to apply quarantine.  (But it seems like it would be difficult to define rules that wouldn't be exploitable.)

Nemo157:
> When a _new_ crate is quarantined, the _name_ must _not_ be visible via any public API. Otherwise it would be possible for a malicious actor to see these, and engage in "front-running", grabbing the crate name with a crate tailored to not be quarantined.

This is not possible:

> A version in the quarantined state [..] will appear in crates.io API responses as an unpublished crate, but will _not_ be added to the index [..]

By being part of the API the crate has already been registered in the database as owned by the publishing user, and the published version registered on that crate. It's simply the index metadata telling cargo about the crate that hasn't been updated yet.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This project will add a new admin feature to crates.io that allows crate publishes matching specific properties to be placed into a quarantine state instead of being immediately published.
+
+A version in the quarantined state will have its `.crate` file published to S3 to simplify review, and will appear in crates.io API responses as an unpublished crate, but will _not_ be added to the index — and will therefore _not_ be installable via `cargo` — until it has been reviewed and approved.
+
+`cargo publish` will also be modified to report when an uploaded crate version has been placed into quarantine. 
+
+Once a crate version has been quarantined, then a member of the crates.io team will review the quarantined crate, ideally within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail, assuming crates.io has a verified e-mail available for their account.
```
Turbo87, line 30:
```suggestion
Once a crate version has been quarantined, then a member of the crates.io team will review the quarantined crate, ideally within 1 week[^sla]. Once reviewed, the crate may be either approved and published to the index, or rejected and deleted. Either way, the crate owner(s) will be notified by e-mail.
```

We only allow publishes from users with verified email addresses these days :)

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+
+Should the crate owner(s) disagree with the quarantine action, they may appeal to the moderation team.
```
Turbo87, line 32: Do we want to specify how long we will wait for an answer of the crate owner before potentially deleting a quarantined crate?

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+The publish endpoint will notify the publisher (generally `cargo publish`) of the quarantined version by extending the response[^publish-response] in two ways:
+
+1. A new field will be added to the response called `quarantined`.
+2. A human readable warning will be added to the `warnings` field within the response.
+
+When a crate version is quarantined, a support ticket will also be opened within the crates.io help desk, thereby notifying the crates.io team of the newly quarantined version. If additional crates are quarantined for that user or regex, a comment will be added to that ticket, if it is still open. (This prevents the help desk being spammed.)
+
+### API
+
+A new field will be added to API endpoints that return crate versions called `published`. This will only be `false` for crate versions that are in quarantine.
```
Turbo87, line 74: I'm wondering if we should use something like `state: "published"` and `state: "quarantined"` instead. this would allow us to add other states in the future without adding more fields to it. one example could be temporary suspension of a crate/version due to policy violations or retroactive quarantine due to potentially malicious code being discovered.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+
+The `versions` table will need a new `quarantined` boolean flag, much like the existing `yanked` flag.[^state]
+
+### Publishing
+
+The crates.io database will be expanded to store and capture rules that, if matched, will cause a crate version that is being published to enter the `quarantined` state, rather than being immediately published. These rules will be defined and maintained by crates.io admins.
+
+The publish endpoint will notify the publisher (generally `cargo publish`) of the quarantined version by extending the response[^publish-response] in two ways:
+
+1. A new field will be added to the response called `quarantined`.
```
Turbo87, line 67: What do you think about adding a `version` field instead, containing the version that was just created? That would automatically contain the quarantine status (see "API" section below) and might be useful in other ways too.

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+
+A new CRUD area will be provided for rules to be defined and managed.
+
+#### Review
+
+A new review area will be added that lists crate versions currently in quarantine, along with options to download the `.crate` file and manage the version (ie by approving or rejecting the crate version).
+
+## `cargo publish`
+
+The wait loop in `cargo publish`[^wait-loop] currently refreshes the crate index and waits for the new crate version to appear.
```
Turbo87, line 90: looks like the `^wait-loop` reference is missing below 😉

text/0000-crate-quarantine.md
Turbo87: thinking about this some more, I guess we could also encode whether the version has been published to the indexes yet. OTOH we have two of them now, so a single field might not be enough for that then 🤔

text/0000-crate-quarantine.md
```diff
@@ -0,0 +1,156 @@
...
+
+Should the crate owner(s) disagree with the quarantine action, they may appeal to the moderation team.
+
+## When will a quarantine rule be added?
+
+Ultimately, this will be decided by the crates.io team on a case by case basis. There are two primary scenarios that are currently envisioned where this functionality would be used:
+
+### Large scale spam attack
+
+In the event of a large scale spam attack on the entire site, we may want to either restrict individual users (if the spam is coming from a finite set of users) or quarantine all published versions temporarily (using a `.*` crate name rule) until it subsides, at which point we can analyse which crate versions are actually valid.
```
Turbo87, line 40: another rule that might be useful: putting **new** crates into quarantine but allowing new versions of existing crates. not necessarily right now for this RFC, but we should keep it in mind when we implement this.

PR #3466, opened 2023-07-30
Move out of deref for `ManuallyDrop`

Jules-Bertholet:


text/3466-manuallydrop-deref-move.md
```diff
@@ -0,0 +1,159 @@
...
+}
+```
+
+This RFC extends this capability to `ManuallyDrop`.
+
+```rust
+use core::mem::ManuallyDrop;
+
+let m: ManuallyDrop<(String, String)> = ManuallyDrop::new(("hello".to_owned(), "world".to_owned()));
+let _: String = m.1; // `m.1` dropped here
```
programmerjake, line 59:
```suggestion
drop(m.1); // `m.1` moved out of here
```
just `let _ =` never moves in the first place

text/3466-manuallydrop-deref-move.md
```diff
@@ -0,0 +1,159 @@
...
+struct Foo(String, String);
+
+impl Drop for Foo {
+    fn drop(&mut self) {
+        println!("down to the ground");
+    }
+}
+
+let m: ManuallyDrop<Foo> = ManuallyDrop::new(Foo("hello".to_owned(), "world".to_owned()));
+let _: String = m.1; // `m.1` dropped here
```
programmerjake, line 75:
```suggestion
drop(m.1); // `m.1` moved out of here
```

text/3466-manuallydrop-deref-move.md
Jules-Bertholet: Thanks, fixed

text/3466-manuallydrop-deref-move.md
```diff
@@ -0,0 +1,159 @@
...
+        DefuseWrapper {
+            bomb: Bomb {
+                is_armed: true,
+            }
+        }
+    }
+}
+```
+
+If code like the above example actually exists in the ecosystem, this proposal would make it unsound.
```
fee1-dead, line 142: I actually do have an unpublished crate that relies on fields not being able to be moved out if I implement `Drop`. Basically `SomeStruct` has a field with type `FieldReader`. The latter type is a ZST that implements `Deref` based on the address of `&FieldReader` (in this case, it reads other fields from `SomeStruct`), that relies on the fact that if I make `SomeStruct` implement `Drop`, and construction of `FieldReader` unsafe, then all pointers to `FieldReader` are guaranteed to exist only within `SomeStruct`, since moving out of `SomeStruct` isn't allowed.

It would be nice if there is a way to spell "can't move out of any fields of this struct" that does not involve a no-op drop implementation, or maybe there is a way already and I just don't know about it.

Jules-Bertholet: What I would expect/was expecting is that in  practice, this guarantee would be provided via privacy. Your `FieldReader` would be a private field with a public `get()`/`get_mut()` accessor.

(Also, your use-case might be impacted by subobject provenance discussions, you might want to share details at https://github.com/rust-lang/unsafe-code-guidelines/issues/256)

fee1-dead: What I'm hoping to achieve with `FieldReader` is that I want it to be a public field, while emulating getters through `Deref`, so using accessor methods defeats its point entirely.

ijackson: This conversation reminds me of the debate over the [soundness of `partial-borrow` vs `replace_with`](https://lib.rs/crates/partial-borrow#readme-compatibility---language-assumption-about-owned-values-from-references).  In `partial-borrow` I also (ab)use a language inflexibility to allow the use of fields rather than accessors.

(Providing fields rather than accessors is more than just a simple ergonomic improvement to elide some `()`.  There are important borrowck features that are available only via borrowck's special handling of fields.)

So it seems both @fee1-dead's crate and `partial-borrow` would benefit from some way to get access to those "field-specific" borrowck behaviours.  In the meantime this is another case where not only (i) adding a new language feature can make existing code unsound (ii) that existing code is not theoretical and cannot be readily expressed a different way.

fee1-dead: > (Providing fields rather than accessors is more than just a simple ergonomic improvement to elide some `()`. There are important borrowck features that are available only via borrowck's special handling of fields.)

Yes. I forgot to mention this in my comment, but this is one big reason I'm using a `FieldReader` design, to allow having mutable references for two fields.

Jules-Bertholet:
I've expanded the "Rationale and Alternatives" section in light of new discussion [on Internals](https://internals.rust-lang.org/t/destructuring-droppable-structs/20993/18).

PR #3468, opened 2023-08-02
markup_macro

hydroper:
A simple feature for generic markup types.

[Rendered](https://github.com/hydroper/rust-rfcs/blob/master/text/0000-markup-macro.md)

### DIFF
--- a/text/0000-markup-macro.md
+++ b/text/0000-markup-macro.md
@@ -13,6 +13,8 @@ The `markup!` and `markup_type!` macros allow constructing arbitrary node types,
 
 Applications may use node hierarchies as a fundamental way of describing graphics. Without the `markup!` macro, the developer must manually construct a node by chaining methods that are usually prefixed by `set_` and call a chaining `append_children` method.
 
+A procedure macro is limited for cases where the node type provide common `set_` prefixed methods for all node kinds, but where the node type does not provide methods that are very specific to a node kind.
+
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
--- a/text/0000-markup-macro.md
+++ b/text/0000-markup-macro.md
@@ -1,0 +1,97 @@
+- Feature Name: `markup_macro`
+- Start Date: 2023-08-01
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+The `markup!` and `markup_type!` macros allow constructing arbitrary node types, where each node contains arbitrary attributes and zero or more child nodes.
+
+# Motivation
+[motivation]: #motivation
+
+Applications may use node hierarchies as a fundamental way of describing graphics. Without the `markup!` macro, the developer must manually construct a node by chaining methods that are usually prefixed by `set_` and call a chaining `append_children` method.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `markup!` macro takes elements of the form:
+
+```plain
+<Node/>
+<Node></Node>
+{node}
+{iterable}
+```
+
+The angle brackets forms are similiar to XML, accepting attributes with interpolations, where `Node` must be a type defined by the macro `markup_type!`.
+
+The `markup_type!` macro is used to define a node type can be constructed by the `markup!` macro. Besides that, this macro describes how children are added, which attributes are valid, which is the type of each attribute and to which expression the attribute assignment translates.
+
+The following program defines a `Node` type that can be constructed via `markup!`:
+
+```rust
+markup_type! {
+    pub struct Node;
+
+    // constructs the node in an initial state.
+    fn new() -> Node {
+        Node
+    }
+
+    // the `append_child` function describes
+    // how the contained nodes are added into the collection.
+    fn append_child(parent: Node, child: Node) {
+        // append child here.
+    }
+
+    // define the attribute `some_attr`
+    fn some_attr(node: Node, value: u64) {
+        // set attribute here
+    }
+}
+
+let node = markup!(
+    <Node>
+        <Node some_attr={15}/>
+    </Node>
+);
+```
+
+### Return
+
+The `markup!` macro returns either a single node or a `Vec` of nodes if it contains more than one node at the top.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The `markup!` macro takes a sequence of items: XML-based tags and interpolations:
+
+- An interpolation accepts a node or an iterable. If it evaluates to an iterable, then that iterable contributes all of its items to the enclosing tag or to the top `markup!`.
+- A XML-based tag consists of a type, optional attributes (supporting interpolation) and optional children. It is an error if the type is not defined by `markup_type!`. For each attribute, based in `markup_type!`, validate if it exists and evaluate its assignment. For each child, evaluate it and call `append_child` from `markup_type!`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This adds two macros to the standard library which must be implemented at the language-level due to how attributes are translated into expressions.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- The `markup_type!` macro allows describing attributes of arbitrary name whose type errors are caught at compile-time. A trait could not be used instead.
+
+# Prior art
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Generic nodes.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Nothing yet.
--- a/text/0000-markup-macro.md
+++ b/text/0000-markup-macro.md
@@ -53,10 +53,10 @@ markup_type! {
         Node
     }
 
-    // the `append_child` function describes
+    // the `append_children` function describes
     // how the contained nodes are added into the collection.
-    fn append_child(parent: Node, child: Node) {
-        // append child here.
+    fn append_children(parent: Node, child: Vec<Node>) {
+        // append children here.
     }
 
     // define the attribute `some_attr`
@@ -82,7 +82,7 @@ The `markup!` macro returns either a single node or a `Vec` of nodes if it conta
 The `markup!` macro takes a sequence of items: XML-based tags and interpolations:
 
 - An interpolation accepts a node or an iterable. If it evaluates to an iterable, then that iterable contributes all of its items to the enclosing tag or to the top `markup!`.
-- A XML-based tag consists of a type, optional attributes (supporting interpolation) and optional children. It is an error if the type is not defined by `markup_type!`. For each attribute, based in `markup_type!`, validate if it exists and evaluate its assignment. For each child, evaluate it and call `append_child` from `markup_type!`.
+- A XML-based tag consists of a type, optional attributes (supporting interpolation) and optional children. It is an error if the type is not defined by `markup_type!`. For each attribute, based in `markup_type!`, validate if it exists and evaluate its assignment. Call `append_children` from `markup_type!` to append children.
 
 The `markup_type!` macro puts no retriction about which type is returned by `new`; that is, `new` may return a different type from the enclosing `markup_type!`'s type. For example, `Button` may be construct via `new` resulting into a `Node`.
 
--- a/text/0000-markup-macro.md
+++ b/text/0000-markup-macro.md
@@ -99,7 +99,7 @@ This adds two macros to the standard library which must be implemented at the la
 # Prior art
 [prior-art]: #prior-art
 
-N/A
+- JSX (and TypeScript) supports similiar markups without using a macro. JSX and TypeScript use a transform that occurs at the compiler API.
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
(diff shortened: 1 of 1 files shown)


programmerjake:
this seems like it could easily be a library using proc-macros

SOF3:
A lot of libraries here <https://www.arewewebyet.org/topics/templating/> already use this kind of syntax.

Also my two cents on this syntax — why does everyone want to use `</markup>` syntax instead of the more Rust-like braces like [horrorshow](https://docs.rs/horrorshow) did?

hydroper:
@programmerjake No one told me `proc_macro` could be used for this...

@SOF3 The Rust syntax isn't very suitable for this IMO, as the markup includes attributes and children and XML is a lingua franca. This `horrowshow` you've mentioned uses a different syntax

hydroper:
There's still one issue with `proc_macro`: it'll require me to determine a type by its identifier rather than by its actual lexical reference. For example, `"Button"` instead of `Button`. This is only possible if there a way to evaluate a type's identifier in the markup, resulting into a `TypeId`.

SOF3:
Why do you need to determine the contents of a type instead of just generating code using it? All the existing XML-like proc macros are just working fine.

> The Rust syntax isn't very suitable for this IMO, as the markup includes attributes and children and XML is a lingua franca. This horrowshow you've mentioned uses a different syntax

Crates like [horrorshow](https://docs.rs/horrorshow) and [defy](https://docs.rs/defy) are able to generate all valid XML structures including attributes and children without any problems whilst keeping the syntax Rust-like so that it is more editor-friendly and formatter-friendly. What's the problem with them?

And most importantly, why would we have a markup macro in the standard library when the stnadard library did not even include any XML parsers to begin with?

hydroper:
> Why do you need to determine the contents of a type instead of just generating code using it?

Because `Node` isn't a `Button` and `Button` isn't a `Node`. They don't share the same `set_` methods for instance.

> Crates like [horrorshow](https://docs.rs/horrorshow) and [defy](https://docs.rs/defy) are able to generate all valid XML structures including attributes and children without any problems whilst keeping the syntax Rust-like so that it is more editor-friendly and formatter-friendly. What's the problem with them?

I've thought that a Rust-like syntax would go well before, but after thinking about attributes and children together, it looks less convenient than a XML-like one. And I'm not going to generate HTML for building websites (if I were to do this, I'd rather not use Rust).

> And most importantly, why would we have a markup macro in the standard library when the stnadard library did not even include any XML parsers to begin with?

- 1. This isn't a complex addition. The syntax is pretty easy.
- 2. XML is a lingua-franca and looks good for attributes. The interpolation syntax also looks readable.

SkiFire13:
> Because `Node` isn't a `Button` and `Button` isn't a `Node`. They don't share the same `set_` methods for instance.

What's wrong with letting the compiler raise an error if such method is not available?

> This isn't a complex addition. The syntax is pretty easy.

This is complex in that it adds a bunch of stuff that don't fit with the rest of the language, having its own syntax and its own semantics, mostly different from what we have today.

> XML is a lingua-franca and looks good for attributes.

Why do we need a lingua franca for Rust syntax though?

hydroper:
> What's wrong with letting the compiler raise an error if such method is not available?

Well, it happens that I don't want to expose all methods to `Node` for example. For example, `set_warning` exists only in `Button`. How will the `markup!` macro (using just `proc_macro`) know that it has to do `node.to::<Button>().set_warning(v)` before chaining `set_` methods?

hydroper:
> This RFC has no mention of any of these. There's no mention that somehow node.to::<Button>().set_warning(v) should be generated in some cases. And this is not missing detail: this RFC contains no mention of what the compiler should generate at all.

Now on motivation

> This seems like a domain problem of your program/library, you should try not to mix it with the features that the language should offer.

I see, but it seems like there's no elegant solution using a procedural macro. With this proposal, there'll be proper compile-time verification, which I think is good

hydroper:
Also, I just remembered: JSX has similiar markup, but using a transform at the compiler side.

SOF3:
JSX is not part of the standard JavaScript/ECMAScript, and JSX ends up getting preprocessed into regular JavaScript anyway, which is effectively what a proc macro is about.

ahicks92:
Nothing stops a procedural macro from allowing syntax like:

```
<full::path::to::type ...>
</full::path::to::type>
```

That is, it is already possible to implement macros that don't take into account the lexical scope in the same way that the rest of the language doesn't take into account the lexical scope if someone wanted 

I think this is missing an explanation of why you can't just write a library which will use a trait or set of traits.  It is possible to do things like name all attributes via types (e.g. `struct Position`), transform ids in the procedural macro, then implement traits which prevent setting invalid attributes on the wrong types at compile time (I have done this but I don't remember how and it's involved; but that doesn't change that it's doable).

I can't think of a use case for this which would want to combine multiple libraries in a way where the downstream code isn't abstracting over it in some way that's consistent.  I can't think of a reason you'd want this for a single library.  Additionally everyone would likely have to release new versions of everything to migrate to it.  It's not really a small feature as soon as you consider the implications on the ecosystem when it comes to code churn to get everything over.  We are effectively stuck with the old solutions forever in any case.

It looks like part of the reasoning here is that, in some cases, the macro is going to insert conversions.  This is against every other place in the language save a very few exceptions involving primitive types, and can easily lead to invisible and expensive `.into()` calls which box for example.  It also violates the implicit idea that looking at a macro makes it clear what will be generated, since there would be spooky action at a distance here based on whether or not other types happen to implement conversions today.

I also don't think that the implied universality of XML is really apt.  Outside jsx and things like it I haven't seen XML in a long time and it is definitely the case that a lot of people are against it in favor of JSON.  This certainly has uses and I agree that jsx-type expression of node trees is useful for the stated case, but the standard library tries to avoid assuming that allocators are present where possible and putting something in it whose core justification is graphics seems very much against the ethos of a standard library that tries to work on the lowest common denominator whenever possible as a core philosophy (at least, a core philosophy when looking in from the outside).

If nodes have "base" nodes and you're trying to write a DSL then one effective way to expose a good interface to the user is to implement DerefMut.  If there is no base, the only thing a conversion can do is "downgrade" the type to something else.  While base class = DerefMut is ugly in the common case, a DSL is allowed to use weird tricks in my book.  This avoids allocation concerns and also doesn't need help from the compiler beyond the implicit dereferencing that already exists.

hydroper:
@ahicks92 I already knew about the fully-qualified case, but I was referring to lexical names where you don't want verbosity.

Anyway, you got me the idea of using a trait specifically for this markup, which mustn't be included lexically, so that the markup uses that trait for setting the attributes (`<node as rialight::graphics::markup::Markup>.set_attribute(v)`). That could work indeed...

hydroper:
@ahicks92

> If nodes have "base" nodes and you're trying to write a DSL then one effective way to expose a good interface to the user is to implement DerefMut. If there is no base, the only thing a conversion can do is "downgrade" the type to something else. While base class = DerefMut is ugly in the common case, a DSL is allowed to use weird tricks in my book. This avoids allocation concerns and also doesn't need help from the compiler beyond the implicit dereferencing that already exists.

That's a possibility I could consider, convert to something else. It happens that I'll use the composition-approach, so `Node` holds an union of boxed variants like `Button` for example.

To clarify, `Button` can't refer back to its corresponding `Node` because `Button::new()` returns `Node` and if it were to return `Arc<Button>` instead, the weak reference to the `Node` has been discarded already. This has something to do with one of my reasonings above, but anyway...

I agree that I can throw a node when building it in the markup to a wrapper type that includes (_base_, _kind_) and all methods specific to that kind, and inheriting all methods from another trait which acts as a delegate to the base. That involves use of a hidden method that is defined in the kind itself.

hydroper:
After @ahicks92 comment, I realized this proposal isn't necessary and I can specialize that macro more for building graphical nodes (atomically reference-counted) and custom UI components that the developer can build over them.

hydroper:
[Here's what I annotated](https://github.com/rialight/api#graphics-markup-and-custom-nodes)

hydroper:
Even though, I'm still a little confused whether I use traits or composition................

Not sure if I should ask again in the community if traits could impact performance.

programmerjake:
if you're not using `dyn Trait`, traits do not impact performance, they're just a fancy way to call functions (and access consts and types). (they may affect compile-time performance to a small extent)

hydroper:
@programmerjake Yeah, my worry is that `Arc<dyn Node>` could be operated slowly when performing node tree operations such as resolving node path... Inheriting parent properties such as `skin`, and so on.

hydroper:
Okay, I've an idea of renaming the thing returned by `Markup::build()` into the node kind itself... And the node kind itself isn't exactly what's stored in its `Node` equivalent; there'll be a _hidden_ separate structure automatically generated for it within the `define_node!` macro, maybe? Also an `Into<Node>` implementation for covariance. And with this, I'll be able to type `let button: Button = markup!{...};` instead when desired instead of `let button: Node = markup! {...}`. Will that be useful though? Not generally, but allows for some flexibility

ahicks92:
On modern microprocessors even `dyn trait` is often executed fast enough that you can't notice the difference due to speculative execution.  The real cost of that approach is the Arcs, which are one heap allocation each.  The cost of looking up a function pointer in a vtable is probably significantly less (and dominated by) the cost of pointer chasing.  But none of us can know without benchmarking your library.  You can't reason this out like that.

You can possibly avoid this overhead if the dsl built by the macro is different types than your runtime structure such that getting to the runtime structure has a conversion step.  In the case of "static" trees (e.g. no splat operator that lets code insert more nodes) this can compile directly to struct literals and possibly slices `[(&'static str, Value)]` where Value is some sort of multi-value enum (e.g. like serde_json's Value type).  The macro could even just output `Arc::new(NodeHere::new(NodeDescriptor { ... }))` which even gets around the problems with mutability through Arcs, and it could deal with dynamic node collections being hooked in via Cow and something like Smallvec to avoid allocations etc. where possible.  If the performance cost of the slice bothers you, it is possible to do clever things with implementing traits on tuples so that the compiler can fully inline; this would allow the macro to generate `(("a", 5), ("b", "foo"))` thus getting around the need for loops etc.

I'm not interested in doing it but I am pretty certain I could get something as efficient as hand-written code if I spent a couple days on this problem, and that such a thing would fulfill the use case as laid out in the RFC.  I'm sure one of the mentioned crates has either put in the work, or discovered that it's not worth it because other costs dominate.  I'm reasonably sure that if I was feeling sufficiently clever and had a  use case, I could make this work on top of something like ArrayVec so that it lives on the stack, though that would come with an upper limit on the number of nodes as well as a few others.

I'm not sure what you mean about scoping.  Rust as a language has lexical scopes and relative/fully-qualified paths.  There's no third kind of scoping for the macro to use, unless you're hoping that the nodes are available without imports.  You can do everything else you want to do without help from the compiler, it just requires really being clever internally.  Macros can even "use" "specialization" via [Dtolnay Specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md) which can even allow for all the extra super weird special cases (that repository also has other useful macro tricks in it).

If you still think this should be in the standard library, then IMO there needs to be a much better motivation section.

hydroper:
Thanks for the arguments. Hmm... I think procedural macros will work! And I'll also end up supporting covariance even without use of traits.

> #### Graphics Markup and Custom Nodes
> 
> Define two [procedural macros](https://doc.rust-lang.org/reference/procedural-macros.html) that facilitate defining nodes and custom UI components. For example, `define_node!` generates a separate `KKindData` structure, which is contained by `K` (node kind) itself. `K` contains (_base_, _data_). `K::new()` constructs an empty `K`. `K` inherits `NodeKind`, inheriting everything from _base_ (such as `set_skin` and `parent()`). _data_ is an `Arc<KKindData>`.
> 
> Syntax:
> 
> - `define_node!` is given field-like attributes somewhere, aggregating to a `struct` and automatically aggregating `set_` prefixed methods to `impl K`. There might be support for specifying `set_` methods explicitly too if special processing of the attribute value is desired.
> 
> It makes sense for UI components to be nodes, therefore `UiComponent` inherits `NodeKind`. They are defined with a similiar macro `define_ui_component!`.
> 
> Ideally there'll be three macros: `markup!`, `define_node!` and `define_ui_component!`.
> 
> #### How Nodes Are Implemented
> 
> - `Node` contains an `Arc<NonRefNode>`
> - `NonRefNode` contains common fields and the stores the actual node kind data as `Arc<dyn Any>`.
> - `K::new()` takes no arguments and returns _K_
> - _K_ contains (_base_, _kind data_).
> - _kind data_ is of type `Arc<KKindData>`
> - The macros `define_node!` and `define_ui_component!` generate a `KKindData` structure
>   - `KKindData` must have a `#[doc(hidden)]` attribute
> - `NodeKind` will implement `Into<Node>`, evaluating to _base_ (the kind as the `Node` type).
> - `NodeKind` has a static function `reference_cast` that takes a `node: Node` and returns `Option<K>`. This is used by `Node` methods such as `.to`, which unfortunately have no access to the type from the node kind's data structure (`KKindData`).
> - The `markup!` macro will build nodes using something like `let node: Node = K::new().into(); node`, chaining `set_` methods after `::new()`

ahicks92:
At this point though, I think it's important to specify whether or not you consider this RFC to still be worth considering?

hydroper:
@ahicks92 I was already finding my proposal a bit limited. The proposed `markup_type!` is so generic that it can't inherit attributes from the base type (`Node`). It'd only be possible to do that if the node kind could contain the `Node` itself, but that would create a never-released `Arc`. If I use `Weak`, I get an issue too: the `Weak` gets empty right after `K::new()`.

ahicks92:
I'm not sure what the concrete problem is, because the RFC is too abstract and is saying that it does things libraries that already exist do, but you're saying you can't do them without help from the compiler.

Maybe this is yet another "How do I do inheritance in rust" RFC, where the core problem is that in many ways the answer is "you don't"?

The only thing I can think of that would cause me to need help from the compiler if I were trying to do this is if I was trying to make the system fully pluggable, since in that case it'd have to be a bit magic to work across all the involved crates.  But even then the overhead of having dynamic attributes in a list is still pretty small compared to everything else you'd be doing with the tree anyway, I think, and I could just do that and it's fine even if the nodes are in 20 different crates.

But honestly I will leave the rest of this to others because mostly I popped in because this was interesting, and I'm not sure I'm really adding anything to the discussion at this point.

PR #3472, opened 2023-08-13
Fix start date format in RFC 0201

xzmeng:
Fix start date accoding to the format of `YYYY-MM-DD` in `0000-template.md`.

[Rendered](https://github.com/xzmeng/rfcs/blob/fix-start-date-format-0201/text/0201-error-chaining.md)


ehuss (approved):
Thanks!

PR #3476, opened 2023-08-20
RFC: `impl-trait-for-fn`

DasLixou:
Adds the possibility to implement traits directly for a specific function next to generic implementation based on function signatures.

### DIFF
--- a/text/0000-impl-trait-for-fn.md
+++ b/text/0000-impl-trait-for-fn.md
@@ -1,0 +1,90 @@
+- Feature Name: `impl-trait-for-fn`
+- Start Date: 2023-08-20
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: N/A
+
+# Summary
+[summary]: #summary
+
+Support for implementing traits on functions
+
+# Motivation
+[motivation]: #motivation
+
+I was trying to make something similar to bevy's system functions. And for safety reasons, they check for conflicts between SystemParams, so that a function requiring `Res<A>` and `ResMut<A>` [panic](https://github.com/bevyengine/bevy/blob/main/crates/bevy_ecs/src/system/system_param.rs#L421).
+
+Then after I heard about axum's [`#[debug_handler]`](https://docs.rs/axum/latest/axum/attr.debug_handler.html) I wanted to do something similar to my copy of bevy systems, so that I get compile time errors when there is a conflict. I wanted even more, I wanted to force the user to mark the function with a specific proc attribute macro in order to make it possible to pass it into my code and call itself a system.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+As we all know, you can implement a trait for a struct with the following syntax
+```rust
+struct Timmy;
+impl Person for Timmy {
+    fn greet() {
+        println!("Hey it's me, Timmy!");
+    }
+}
+```
+And we can also implement a trait for all functions with a specific signature like this
+```rust
+impl<F> ValidSignature for F
+    where F: Fn(i32) -> bool
+{
+    /* ... */
+}
+```
+Now we can also implement traits for specific functions only
+```rust
+fn valid() {}
+impl ValidFunction for fn valid {
+    /* ... */
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+It gives the possibility to also implement a trait directly scoped to a function instead of generic implementation of multiple. Other than that, it basically behaves the same. It should also be possible to implement them via proc attribute macros:
+```rust
+#[impl_debug_name = "Greeting"]
+fn greet() {
+    /* ... */
+}
+// should expand to something like
+fn greet() {
+    /* ... */
+}
+impl FnDebugName for fn greet() {
+    fn debug_name() -> &'static str {
+        "Greeting"
+    }
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+i dont know any
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+I think it's a easy task because we can already implement traits for a group of functions with the same signature, so why shouldn't we also implement single scoped impls?
+
+# Prior art
+[prior-art]: #prior-art
+
+i dont know any
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is the syntax good? I feel like we could drop the `fn` to from `impl Trait for fn function` to `impl Trait for function`.
+- What about closures? They don't even have names so targetting them would be quite difficult. I wouldn't want to use the compiler generated mess of a name like `[closure@src/main.rs:13:18: 13:20]`. It would also contain line numbers which would be changing quite often so thats not ideal.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+- also make it possible to implement traits for closures directly.


text/3476-impl-trait-for-fn.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+i dont know any
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is the syntax good? I feel like we could drop the `fn` to from `impl Trait for fn function` to `impl Trait for function`.
```
orowith2os, line 84: This makes sense, I think - lints will make sure types are styled properly, and the `impl Trait for function` syntax already works to differentiate between a function and type due to it obviously being a function, at least in the reference you gave.

One question - what about input and output types for the function? You gave an example of a function that has no input and no output types. Would it be like how other types work now, with it just being `impl Trait for any_function()` and `fn any_function(x: i32) -> 32`?

DasLixou: I've on purpose left out the params parens and return type arrows in the impl so that you don't have to write them.
So
```rust
fn itoa(i32) -> String { .. }
```
would still be implemented with
```rust
impl MyTrait for fn itoa { .. }
```
This is possible because you cannot have two functions with different params but the same name. Im also quite sure that we also have no complications with targeting what fn/type to implement, but if so, there's the `fn` keyword in the impl. Otherwise we could think about dropping it.

But yeah this is was not well written by me. I'll change a fn signature in an example so it's easier to understand.

orowith2os: I'd say it's fine to get rid of the `fn` keyword, but only so long as the function name is *clearly* a function in the `impl` bit, like by having any function, no matter its input types or outputs, get traits with `impl Trait for function()`

DasLixou:
Im unsure about fns of impl blocks. Should this be allowed?:
```rust
struct MyStruct;
impl MyStruct {
    fn new() -> Self { Self }
}
impl Creatable for fn MyStruct::new {
    /* ... */
}
```
I have to check whether this applies to the function traits. When the `fn new` implements `Fn()` (I think it is, I just have to make sure), then we probably also want the possibility to implement directly on them, right?

text/3476-impl-trait-for-fn.md
DasLixou: @orowith2os can this be marked as resolved?

DasLixou: 😮 now i see where you got that name() idea from. I mistyped in one example 😅

orowith2os: So long as there's either the `fn` keyword in the `impl` or some identifier (`()`) to make it clear that it's a function, lgtm :)

burdges:
`Fn`, `FnMut`, or `FnOnce` are closures, not functions.  Ain't clear how you'd name the type of a closure, but in theory there is a type there which could satisfy extra traits.  And the compiler does them sometimes.

A type like `fn(..) -> ..` is a [function pointer type](https://doc.rust-lang.org/std/primitive.fn.html).  I doubt `impl Trait for fn foo` could make sense, because `fn bar(..) -> .. { ... }` is a value of this function pointer type, not a type itself. 

Just fyi [`#![feature(fn_traits)]`](https://doc.rust-lang.org/stable/unstable-book/library-features/fn-traits.html) permit creating closures manually, but one should typically forgo the function call syntax, and just define the trait you desire.

DasLixou:
> `Fn`, `FnMut`, or `FnOnce` are closures, not functions.  Ain't clear how you'd name the type of a closure, but in theory there is a type there which could satisfy extra traits.  And the compiler does them sometimes.
> 
> 
> 
> A type like `fn(..) -> ..` is a [function pointer type](https://doc.rust-lang.org/std/primitive.fn.html).  I think `impl Trait for fn(..) -> ..` covers all function with that signature.  I doubt `impl Trait for fn foo` could make sense, because `fn bar(..) -> .. { ... }` is a value of this function pointer type, not a type itself. 
> 
> 
> 
> Just fyi [`#![feature(fn_traits)]`](https://doc.rust-lang.org/stable/unstable-book/library-features/fn-traits.html) permit creating closures manually, but one should typically forgo the function call syntax, and just define the trait you desire.

I'm aware of that. Fn traits can focus on closures, but also target normal functions. Also this RFC is for implementing a trait on a function directly because something like `impl<F: Fn(i32) -> bool> MyTrait for F` is already possible but not for a single function only.
Pointing at closures is also a thing I mentioned in the "Unresolved Questions" chapter.

text/3476-impl-trait-for-fn.md
```diff
@@ -0,0 +1,120 @@
...
+impl<F> ValidSignature for F
+    where F: Fn(i32) -> bool
+{
+    /* ... */
+}
+```
+Now we can also implement traits for specific functions only
+```rust
+fn valid() {}
+impl ValidFunction for fn valid {
```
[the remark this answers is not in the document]
DasLixou: > Every function already has an implicit type. So I assume that what you mean is that `fn x` is syntax to refer to the type of `x`. The RFC should elaborate that, instead of just being around trait impls.

Oh sorry 😅 yes thats exactly what I want. Also: in the unresolved questions tab i wasnt sure if we could drop the `fn` in the impl block. But I think it is unique, right? So what do you think about dropping the `fn`?
Also: when you say elaborating to the type and not just about trait impls, another question came to my mind: Do we maybe also want to impl directly on a function (I currently work on smth that would really benefit from such a thing)?
Then together with a dropped fn this would look smth like
```rust
fn itoa(val: i32) -> String { .. }
impl itoa { .. }
// or without dropping fn
impl fn itoa { .. }
```

burdges:
You cannot `impl Trait for value` anywhere in Rust, so doing exactly this sounds impossible.

You should ask if const generics could achieve similar, roughly like:
```
type F = fn(..) -> ..;

const GOOD_FN: [F] = [ foo, bar, baz ];

impl Trait for F
where GOOD_FN.contains(self)
{
}
```
I've no idea if this makes sense either, but at least it resembles topics being discussed elsewhere.

There is only one `impl Trait for F` of course, and `F` is defined elsewhere, so the orphan rules require `Trait` be local here, but you could still seal `fn` parameters like this.

There are many reasons why one would choose an explicit type over an fn or closure, so then invocations look like `MyFn::go(..)`, so you can do whatever you like.
```
pub trait MyFn {
    fn go(..) -> ..;
}
```

DasLixou:
@burdges I think you got confused by the syntax. When writing `impl MyTrait for fn x` it doesn't implement `MyTrait` for the function pointer, rather for the ghosttype behind the function `x`. Hope that clears things up 😅

burdges:
Afaik there is no ghost type behind the function `x`, only one behind closures.

burdges:
If some ghost type already exist then whatever, not sure why it'd exist but maybe so.  

Anyways, that playground link seemingly uses the function pointer type, not some ghost type:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=94e692e9d42ed1ad05f068f4777da009

DasLixou:
> If some ghost type already exist then whatever, not sure why it'd exist but maybe so.
> 
> Anyways, that playground link seemingly uses the function pointer type, not some ghost type:
> 
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=94e692e9d42ed1ad05f068f4777da009
> 
> You could try printing [`type_name`](https://doc.rust-lang.org/std/any/fn.type_name.html) to get a final answer there.

interesting approach...

DasLixou:
@burdges 
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=933d56be52fa879eabe2816c1f7da7ce
this looks like a ghost type

text/3476-impl-trait-for-fn.md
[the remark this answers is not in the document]
DasLixou: @Nilstrieb can this be marked as resolved?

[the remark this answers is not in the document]
Nilstrieb: Basically what I'm saying is that "allow impls for function definition types" is the wrong way to approach this. The real feature here is being able to refer to function definition types, so the RFC should focus on that. Mentioning that function definition types are considered local for coherence and can therefore be used in trait impls is then just a small sidenote.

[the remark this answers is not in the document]
DasLixou: @Nilstrieb ohhh... thats a great point but I'm not sure if i can rewrite it for it that much because the motivation / goal is more on trait impls directly. But ill try to make it focus more on this point. 
Thanks for the feedback 😄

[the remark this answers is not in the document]
DasLixou: @Nilstrieb may you recheck the RFC? I rewrote it but i'm unsure if it's still too focused on the impl thing 😅

clarfonthey:
So, historically, I've desired the ability to directly name the types of individual functions before. However, I eventually conceded that type-alias-impl-trait was the solution for this feature, since it seems unclear what the benefit would be.

Here, the main use case presented is implementing traits for functions, which… I dunno, I don't really like that. To me, it feels like we should have the inverse of that: allow implementing *functions* (i.e., `FnMut`, `Fn`, and `FnOnce`) for other kinds of objects. To me, having to implement a trait for a function feels like a hack because other kinds of objects can't be functions.

DasLixou:
@clarfonthey so I kind of understand your arguments and I totally agree that we also want the user to be able to implement Fn traits on his own objects, but I think it would be the best to have both. 
I am also currently writing a library where I would need this and having the user to make a zero-typed struct and implementing the Fn trait + my trait vs. Just using a function makes really a difference.

DasLixou:
@clarfonthey wait... how does type-alias-impl-trat have anything to do with this RFC?

clarfonthey:
> @clarfonthey wait... how does type-alias-impl-trat have anything to do with this RFC?

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=f704eaba39577e9f9321782fa84afdf5

```rust
#![feature(type_alias_impl_trait)]
use std::iter::Map;

type MyFn = impl FnMut(u32) -> u32;

fn square(val: u32) -> u32 {
    val * val
}

fn squaring<I: Iterator<Item = u32>>(iter: I) -> Map<I, MyFn> {
    iter.map(square)
}
```

clarfonthey:
I mean, that's not really how the RFC process works. These features take time and energy to implement, and it's extremely unclear why that energy would be spent toward implementing this feature instead of:

1. Using type-alias-impl-trait in case you need to directly reference the type for a function (since that feature is actively being developed and would certainly be finished before this feature were even started)
2. Adding extra methods to traits instead of implementing those traits with fewer methods for functions

Like, sure, this feature as-is wouldn't actually conflict with a lot. But that's only a small bit of the RFC process: the rest is making sure that it fits in with the rest of the language, and is worth implementing. I don't really see many arguments in favour of that.

joshtriplett:
Semantics aside (and I do like the idea of being able to name function types), rather than having a function-specific syntax like `fn`, this *seems* like a good argument for having a general `type_of!`, that works on functions and on anything else.

orowith2os:
A `type_of!` macro would also help in turning generics into a variadic function, for size optimizations. I have an open issue (https://github.com/rust-lang/rust/issues/114598) about this, but it's lacking exactly that.

DasLixou:
@joshtriplett I also thought of something like type_of keyword instead of fn but didn't know that there is already planning (I guess) for such a macro.

burdges:
Is this useful?  It maybe useful to have traits that declare invariants, like 
```
let my_closure = |..| ..;
unsafe impl SafeforFoo for type_of!(my_closure) { }
something_that_requires_foo(my_closure)
```
It'd work if `my_closure` was a bare function too, but would it still be useful that way?  I'd think _way_ less useful than for closures..

DasLixou:
@burdges both fn_traits and your example seem quite promising! I totally forgot that you can write an impl everywhere, so that would also eliminate the closure problem. Also: rust has a `type` keyword reserved, does it have a usecase? Could we write `impl SafeForFoo for type my_closure {}`?

DasLixou:
@orowith2os does there already exist a RFC for type_of?

programmerjake:
> Could we write `impl SafeForFoo for type my_closure {}`?

afaict that should work, we can borrow C's `sizeof` syntax and make `type <identifier>` and `type(<expression>)` work as type-of.

programmerjake:
> > > Could we write `impl SafeForFoo for type my_closure {}`?
> > 
> > 
> > afaict that should work, we can borrow C's `sizeof` syntax and make `type <identifier>` and `type(<expression>)` work as type-of.
> 
> Now the question is what we understand as an . A variable name would be an identifier as well as an expression

yes, that simply means both `type my_var` and `type(my_var)` (and `type((my_var))` and so on) are all valid. but `type a * b` would not be since it's missing parenthesis.  this is similar to const generic arguments `f::<A>` or `f::<{ A * B }>`

kpreid:
> Using type-alias-impl-trait in case you need to directly reference the type for a function

It seems that you cannot (yet) use TAIT to implement a trait:

```rust
#![feature(type_alias_impl_trait)]
use std::fmt;

fn foo() {}

type Foo = impl Sized;
fn _bind_foo() -> Foo { foo }

impl fmt::Debug for Foo {
    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(fmt, "hello world")
    }
}
```

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
 --> src/lib.rs:9:1
  |
9 | impl fmt::Debug for Foo {
  | ^^^^^^^^^^^^^^^^^^^^---
  | |                   |
  | |                   `Foo` is not defined in the current crate
  | impl doesn't use only types from inside the current crate
```

So, TAIT doesn't solve this particular naming-types problem (implementing a trait for a function item type), or at least the current implementation does not.

orowith2os:
> @orowith2os does there already exist a RFC for type_of?

There does not. I would make one myself, but I'm not deep enough into the Rust ecosystem to put together an RFC that's worth anything.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The type behind functions already exists, we just need to expose it to the user.
+The hard part would be allowing derives, because that may break some things.
+
+# Prior art
+[prior-art]: #prior-art
+
+i dont know any
```
SOF3, line 152: Are there any existing workarounds for this, e.g. through macros in specific scenarios etc?

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+[unresolved-questions]: #unresolved-questions
+
+- Is the syntax good? It could create confusion between a function pointer.
+- What about closures? They don't even have names so targetting them would be quite difficult. I wouldn't want to use the compiler generated mess of a name like `[closure@src/main.rs:13:18: 13:20]`. It would also contain line numbers which would be changing quite often so thats not ideal.
+- I provided a possible solution for a `fn implicit_generic(val: impl Clone) -> impl ToString` function, but because we currently don't have a defined syntax for those generics in types, thus we can't use `impl Trait` as types for fields in structs, we should think about this more, maybe don't implement exposed types of function for such `fn`s and wait for another RFC?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+- Also expose the type of closures
```
SOF3, line 164: Consider including consistency of display names as well. Currently:

- `std::any::type_name_of_val(&drop::<()>)` evaluates to `"core::mem::drop<()>"`
- When a specific function pointer appears in a compiler error, it looks like this:

```
5 |     let () = drop::<()>;
  |         ^^   ---------- this expression has type `fn(()) {std::mem::drop::<()>}`
  |         |
  |         expected fn item, found `()`
```

Choices of display include "fn item" (as opposed to "fn pointer" if it is cast to `fn(())` first), the function path and a mix of function pointer + `{path}`. It might be more consistent if the syntax eventually adopted in this RFC is consistent with the syntax in compiler diagnostics.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+As we all know, you can refer to a struct by its name and for example implement a trait
+```rust
+struct Timmy;
+impl Person for Timmy {
+    fn greet() {
+        println!("Hey it's me, Timmy!");
+    }
+}
+```
+When we want to target a specific function for a trait implementation, we somehow need to get to the type behind it. That is being done with the `fn` keyword as follows
```
SOF3, line 32: Consider adding a section somewhere to define or clarify terminology, such as "fn item", "fn pointer", "specific fn pointer" or whatever you use.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+As described in the **Guide-level explanation**, with the syntax `fn <fn_path>`, we can reference the type behind the named function.
+
+When the function is for example in a different mod, it should be referenced by its path
+```rust
+mod sub {
+    fn sub_mod_fn() { .. }
+}
+impl fn sub::sub_mod_fn {
+    /* ... */
+}
```
SOF3, line 85: Are specific function pointers currently treated as a local type (terminology from [E0210](https://github.com/rust-lang/rust/blob/master/compiler/rustc_error_codes/src/error_codes/E0210.md))? While this is definitely cool, the ability to add custom inherent impls on a specific function pointer brings a superset of possible issues that could be brought by normal trait impls. Should we analyze the two issues separately?

It is definitely [currently possible to implement a trait on function pointers in general](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=0dc8c5950decf2ea9b8f46ca672f440b), no matter specific or not, but I don't know if the inability to inherent-impl a specific fn pointer to a local function is merely due to its Voldemort properties or if it is intentionally so (or is it already possible through some trick I am not aware of?) or if it causes any incompatibility with the current dot-deref checks (since now `fn_name.foo()` could be `<fn fn_name>::foo(fn_name)`, `<&(fn fn_name>>::foo(fn_name)`, `<fn()>::foo(fn_name)`, `<&fn()>::foo(fn_name)`, etc).

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+It should be also possible to get the type of functions inside impl blocks:
```
SOF3, line 88:
```suggestion
It should be also possible to get the type of associated functions:
```

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+
+```rust
+struct MyStruct;
+impl MyStruct {
+    fn new() -> Self { Self }
+}
+impl fn MyStruct::new {
+    /* ... */
+}
+```
```
SOF3, line 98: What about associated types in trait-impl blocks? Do we write

```
fn <MyStruct as Debug>::fmt
```

If `Debug` is already in scope, is `fn MyStruct::fmt` supposed to work regardless, using the same inference rules as the normal function reference?

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+
+It should be also possible to get the type of functions inside impl blocks:
+
+```rust
+struct MyStruct;
+impl MyStruct {
+    fn new() -> Self { Self }
+}
+impl fn MyStruct::new {
```
SOF3, line 95: Do we use turbofish syntax if `MyStruct` takes a generic parameter? e.g.

```rust
struct MyStruct<T>(T);
```

Do we write `fn MyStruct<T>::new` or `fn MyStruct::<T>::new`? If we choose the latter, would it cause any inconsistency from the `fn send<T>` syntax below (as opposed to `fn send::<T>`)?

SOF3, line 95: Any potential parsing ambiguity if we are taking a fn pointer on an associated function of a fn pointer? e.g.

```rust
fn foo() {}
impl fn foo {
    fn bar() {}
}

// how do we parse this?
type FooBar = fn fn foo::bar;
```

Definitely a very bad syntax that must be disallowed, but better specify it in the reference-level explanation.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+fn send<T: Send>(val: T) {}
+impl<T: Send> ParcelStation for fn send<T> {
+    /* ... */
+}
+```
+
+When we have an implicit generic, they will be appended in order to the generic list:
+```rust
+fn implicit_generic(val: impl Clone) -> impl ToString {}
+impl<T: Send, U: ToString> for fn implicit_generic<T, U> {
```
SOF3, line 111: This is inconsistent with the function call syntax:

```rust
fn foo<T>(val: impl Clone) {}

// normal call
foo::<i32>("");

// now the specific fn pointer syntax suddenly becomes
type Alias = fn foo<i32, &'static str>;
```

I would prefer just leaving them Voldemort for now.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+    /* ... */
+}
+```
+
+Just as structs and enums have the possibility to derive traits to automatically generate code, function type do too
```
SOF3, line 116: This may be a major problem for macro developers. For one thing, `syn::DeriveInput` currently has no `Verbatim` variant, so an extra derive input type will most likely be impossible to implement without introducing syn v3 (oh no, we just got (not yet) everyone upgraded to syn v2).

Plus, I don't see any practical reason to need to derive traits for specific fn pointers (which are always ZST) other than for `Debug`, which we can already implement in libcore instead. If traits defined by libraries have a specific requirement to implement on specific function pointers that cannot be done generically, they can use an attribute macro instead of a derive macro for this specific case.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+i dont know any
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is the syntax good? It could create confusion between a function pointer.
```
SOF3, line 157: Not familiar with compiler parsing internals, but could also consider `fn {path}`, which is slightly similar to the current compiler diagnostics.

text/3476-expose-fn-type.md
DasLixou: I don't think so. I couldn't imagine how you could get behind a specific type of a function at all.

text/3476-expose-fn-type.md
DasLixou: this is a fair point, but i feel like that this will just make more boilerplate. we already specified the function and i dont want to make my impl even longer... on the other side: WAIT A MINUTE! this could solve our "what to do with `impl Trait` type param" problem...
So that we write
```rust
fn my_fn(x: impl ToString) -> String { x.to_string() }
impl<X: ToString> fn(X) my_fn {
   // ...
}
```
Ok that would be really cool.
So would you agree that boilerplate is okay here? @SOF3 

"Problem" is that this RFC tends to turn into a more generic `type_of` thing where this syntax wouldn't be possible anymore.. but type_of is another story so what you requested might be the solution i needed

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,164 @@
...
+But what about explicitly designing the `FnContainer` for a specific function, just like the compiler does when resolving the generics. This will work the same as with the trait impl from above:
+```rust
+struct GoodsContainer {
+    inner: fn goods,
+}
+fn goods() {}
+
+let contained_goods = GoodsContainer {
+    inner: goods,
+}
```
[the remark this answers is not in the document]
DasLixou: yeah its a ZST which holds the function (also ZST) but is able to call it. I just scrambled up something to try to explain it. But i wouldn't know a better example. Maybe ill add an impl block to it

text/3476-expose-fn-type.md
DasLixou: Yeah `<fn fn_name>::foo(fn_name)` should be possible... the other ones i kind of don't understand how they would be possible.

DasLixou: Also ill add a note for similar behavior as E0210

text/3476-expose-fn-type.md
DasLixou: That is acutally a good point! I kind of start to like it when so many people think about niche things i never thought of 😅.
I'll look at how it is with associated types (e.g. if `<MyStruct as Iterator>::Item` is needed or if i can just use `MyStruct::Item` and refer to that).

text/3476-expose-fn-type.md
DasLixou: I'll look into that. I also see another problem here: Imagine this
```rust
struct MyStruct<T> {
    pub fn new(x: T) -> Self {
        // ...
    }
}
```
is it
`impl<T> fn MyStruct[::]<T>::new`
or is it
`impl<T> fn MyStruct::new[::]<T>`
?
what do you think? is there a prefered situation for this in rust or should we maybe allow both? @SOF3

text/3476-expose-fn-type.md
DasLixou: @SOF3 [this](https://github.com/rust-lang/rfcs/pull/3476#discussion_r1305504716) may be a solution, but what do you mean by "leaving them Voldemort"?

text/3476-expose-fn-type.md
DasLixou: That's the feedback i was waiting for 😅. I wasn't sure how easily it will be so i threw it into the RFC and waited to see if anyone would complain. So back to attribute macros

text/3476-expose-fn-type.md
DasLixou: isn't that what i already do? or do you mean with explicit `{}` wrapped around?

text/3476-expose-fn-type.md
DasLixou: So i am currently rewriting and for now I will do this partially but use `fn_name` rather than `{fn_name}` because i think this could collide when we want to do something like this
```rust
impl MyTrait for fn() -> bool {

}
```
I don't know if something like this is planned or if this is already marked as "no" (if so please say it to me so ill change it in the RFC), but if it isn't, then the compiler may not be able to differentiate between `fn() {name} {block}` and `fn() {block}`

text/3476-expose-fn-type.md
DasLixou: Yeah this should definitly be forbidden. If we do the `{}` syntax this would be `fn {fn {foo}::bar}`. Otherwise I'm currently not sure if we would want `fn (fn foo)::bar` or `fn <fn foo>::bar`. I'll leave this open until we found a solution (maybe also for the `{}` syntax)

text/3476-expose-fn-type.md
SOF3: @DasLixou the `&` case is because of dot operator auto deref, the last two are because of type coercion

text/3476-expose-fn-type.md
SOF3: > this is a fair point, but i feel like that this will just make more boilerplate. we already specified the function and i dont want to make my impl even longer... on the other side: WAIT A MINUTE! this could solve our "what to do with `impl Trait` type param" problem... So that we write
> 
> ```rust
> fn my_fn(x: impl ToString) -> String { x.to_string() }
> impl<X: ToString> fn(X) my_fn {
>    // ...
> }
> ```
> 
> Ok that would be really cool. So would you agree that boilerplate is okay here? @SOF3
> 
> "Problem" is that this RFC tends to turn into a more generic `type_of` thing where this syntax wouldn't be possible anymore.. but type_of is another story so what you requested might be the solution i needed

not a fan of this. two type expressions joined together without a delimiter is most likely not acceptable to the compiler team, considering we can't do the same with decl macro inputs either.

SOF3: > 

what about just `fn {fn_name}`?

text/3476-expose-fn-type.md
SOF3: definitely the former. `MyStruct::new<T>` doesn't make sense because the type parameter is on MyStruct not the associated function. This would go wrong if `new` itself also accepts type parameters.

As for whether to use turbofish, I guess this depends on how the compiler parses this expression. I'm not a parser expert, so this part needs some input from the compiler team.

text/3476-expose-fn-type.md
SOF3: just don't accept this type. make it unable to name just like closure types.

text/3476-expose-fn-type.md
SOF3: yes

text/3476-expose-fn-type.md
DasLixou: What do you mean by joined together?

SOF3: if it has a return type, the return expression will immediately be followed by the function path, which might lead to parsing ambiguities

DasLixou: I don't think that's a problem.

text/3476-expose-fn-type.md
DasLixou: But what does <fn()> here even mean? For me it looks like a Fn pointer

text/3476-expose-fn-type.md
DasLixou: So what about moving the name front? `fn my_func` and `fn my_func(prms) -> ret`. But that would not match with the error syntax u sent

clarfonthey:
> It seems that you cannot (yet) use TAIT to implement a trait:

Right, that's why I separated referencing the type of a function from:

> Adding extra methods to traits instead of implementing those traits with fewer methods for functions

Spelled out, I'm proposing that you convert something like this:

```rust
fn do_something(x: u32) -> u32 { todo!() }

impl MyTrait for fn do_something {
    fn my_method(self) -> u32 { todo!() }
}
```

and replace it with:

```
struct MyObject;

impl MyTrait for do_something {
    fn my_method(self) -> u32 { todo!() }
    fn do_it(i: u32) -> u32 { todo!() }
}
```

If you control `MyTrait`, rather than asking for someone to implement `(FnMut(u32) -> u32) + MyTrait`, ask them to just implement `MyTrait` and add a method to that trait with the appropriate signature.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,199 @@
...
+fn is_positive(a: i32) -> bool { /* ... */ }
+impl MyTrait for fn is_positive {
+    /* ... */
+}
+```
+
+> 💡 NOTE: Even when we need to describe the function type but the return type is `()`, we can (just as for function pointers and function traits) drop the `-> ()` from the type. (This should also be added as a lint).
+
+---
+A function with a more complex signature, like a const function, we just ignore that when naming the type:
```
madsmtm, line 63:
```suggestion
A function with a more complex signature, like a function that specifies `const`, `unsafe` or `extern "ABI"`, we just ignore that when naming the type:
```

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,199 @@
...
+- RFC PR: [rust-lang/rfcs#3476](https://github.com/rust-lang/rfcs/pull/3476)
+- Rust Issue: N/A
+
+# Summary
+[summary]: #summary
+
+This exposes the function type of a function item to the user.
+
+# Motivation
+[motivation]: #motivation
```
madsmtm, line 12: I'll add more motivation: In Swift, some functions have an associated selector that you can access with [`#selector`](https://developer.apple.com/documentation/swift/using-objective-c-runtime-features-in-swift).

In my crate `objc2`, it would be immensely beautiful (and useful) to be able to do something similar, e.g. access a function's selector using something like `MyClass::my_function::Selector` or `selector(MyClass::my_function)`, instead of having to know the selector name (which might be something completely different than the function name).

madsmtm (approved):

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,199 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The type behind functions already exists, we just need to expose it to the user.
+
+# Prior art
+[prior-art]: #prior-art
+
+i dont know any
+
+# Unresolved questions
```
madsmtm, line 190: I think this may need some work regarding lifetimes?

https://github.com/rust-lang/rust/issues/56105 is still unresolved, and I think it applies here as well?

text/3476-expose-fn-type.md
SOF3: actually my point was that we should make them consistent, but we could change the diagnostic display instead of the syntax

text/3476-expose-fn-type.md
DasLixou: May I just 1 to 1 copy that sentence? 😅

text/3476-expose-fn-type.md
DasLixou: Ah yeah I'll add a lifetime example as well. Also I'll look if I need to link that tracking issue

text/3476-expose-fn-type.md
DasLixou: So is the syntax that I currently have in the RFC ok?

text/3476-expose-fn-type.md
DasLixou: @SOF3 So I made some tests and would say that it's more "rusty" when we do `MyStruct::<T>::new` instead of `MyStruct<T>::new` (https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=316deb57a2d9552d8284a35fb56db2a0)

What do you think?

text/3476-expose-fn-type.md
DasLixou: I added a `ToDo's` section at the end where i proposed a syntax change but I'm not sure if it is that well with the generic in the for "structure"? 😅

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,269 @@
...
+  |
+  = note: expected unit type `()`
+               found fn item `for<'a> fn(&'a _) -> (i32, bool) {cool::<_>}`
+
+For more information about this error, try `rustc --explain E0308`.
+```
+For consistency, we should change the syntax to `for<'a, T: Clone> fn(&'a T) -> (i32, bool) cool` (I'm not sure if we should put generics in the for)
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
eopb, line 248: It think it's worth mentioning that these named fn types could be used to name private return types and private arguments on public functions.

```rust
/// Workaround for `fn_traits` and/or `unboxed_closures` copied from
/// https://docs.rs/c/latest/src/never_say_never/lib.rs.html#1-77
mod fn_traits {
    pub trait FnOnce<Args> {
        type Output;
    }

    impl<F, R> FnOnce<()> for F
    where
        F: ::core::ops::FnOnce() -> R,
    {
        type Output = R;
    }
}

mod public {
    mod private {
        pub struct Private;
    }
    fn returns_private() -> private::Private {
        private::Private
    }
}

// We get to name `Private` by going through `returns_private`.
// This is only possible if we can name the functions type.
pub type Private = <fn public::returns_private as fn_traits::FnOnce<()>>::Output;
```

Accessing `Private` from the outer scope is not something that's possible today. This isn't something I'd mind having as private return types on public functions can be quite annoying :sweat_smile:. Fortunately type aliases can't be used as unit struct constructors. That would have made this type leak a much bigger issue.

eopb:
It think it's worth mentioning in the RFC that these named fn types could be used to name private return types and private arguments on public functions.

```rust
/// Workaround for `fn_traits` and/or `unboxed_closures` copied from
/// https://docs.rs/c/latest/src/never_say_never/lib.rs.html#1-77
mod fn_traits {
    pub trait FnOnce<Args> {
        type Output;
    }

    impl<F, R> FnOnce<()> for F
    where
        F: ::core::ops::FnOnce() -> R,
    {
        type Output = R;
    }
}

mod public {
    mod private {
        pub struct Private;
    }
    fn returns_private() -> private::Private {
        private::Private
    }
}

// We get to name `Private` by going through `returns_private`.
// This is only possible if we can name the functions type.
pub type Private = <fn public::returns_private as fn_traits::FnOnce<()>>::Output;
```

Accessing `Private` from the outer scope is not something that's possible today. This isn't something I'd mind having as private return types on public functions can be quite annoying :sweat_smile:. Fortunately type aliases can't be used as unit struct constructors. That would have made this type leak a much bigger issue.

Aloso:
> Also: rust has a `type` keyword reserved

Even better: Rust also has a [`typeof` keyword](https://doc.rust-lang.org/reference/keywords.html#reserved-keywords) reserved 😄

DasLixou:
@eopb Can public functions even have private types?

DasLixou:
@Aloso Im currently torn back and forth between making this a function-specific thing and a general purpose typeof thing.
But for typeof we would have the problem with "what do we do with (ill call them) implicit generics (e.g. `impl Trait`)."
In this RFC we have a solution but when making this general typeof thing this won't work anymore.

eopb:
> @eopb Can public functions even have private types?

[Yes, they can](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=538b15e2c4334dfc04cb90c312bba362), but I may have been wrong to think that my "trick" would be able to leak types since I think the compiler checks that type aliases don't point to private types. That should prevent it from being a problem. You can probably safely ignore my comment.

madsmtm:
Also, i think the compiler team is currently [working on changing how this works](https://github.com/rust-lang/rust/pull/113126#issuecomment-1647895643), hopefully making it even more robust.

DasLixou:
@eopb nope. Your problem was that fn was private. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7880db8f7f0e5b217e1bd1ac76fda15b
As you can see we error

eopb:
@DasLixou no, my function was public to the outer scope while the type was private to that same scope, see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9b1f39a040765a5ac5285686e3435cd

The compiler is only strict about private types on public functions when the type is very restricted, such as when it's `pub(super)`

text/3476-expose-fn-type.md
SOF3: Yep, but since we are in type context not expression context here, turbofish is probably unnecessary. The turbofish syntax is required for expressions only because of ambiguity with the less-than operator (`<`), but we don't have such ambiguity if we specify that `fn` is always followed by a path instead of an expression.

Of course, you might also want an expression there if it is a generic `typeof` operator, but this is not the case for the scope of this RFC.

SOF3:
the precedence is pretty unclear with the `fn xxx` syntax. if we don't switch to braces, some specification on the precedence might be necessary.

DasLixou:
> the precedence is pretty unclear with the `fn xxx` syntax. if we don't switch to braces, some specification on the precedence might be necessary.

But there aren't any operators involved or do I get something wrong?

SOF3:
found a counterexample if we don't do turbofish:

```rs
let x = foo as fn foo<bar>::qux;
```

could also be interpreted as

```rs
let x = (foo as fn foo) < bar > (::qux);
```
although chained operators are not allowed, and top-level identifiers cannot be values (they can only be crate names), it is a parsing challenge nonetheless.

DasLixou:
any updates on this?

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **function pointer**: pointer type with the type syntax `fn(?) -> ?` directly pointing at a function, not the type implementing the `Fn[Once/Mut](?) -> ?` traits.
+- **function item** (or just function): a declared function in code. free-standing or associated to a type.
+- **function group**: many non-specific functions with the same signature (params, return type, etc.)
+- **function trait(s)**: the `Fn[Once/Mut](?) -> ?` traits
+- **function type**: the type behind a function, which also implements the function traits.
+- **fixed type**: directly named type, no generic / `impl Trait`.
+- **describe the function type**: write `fn(..) -> ? name` instead of just `fn name`.
+
+# Guide-level explanation
```
madsmtm, line 40: Although I agree that this is where we want to end up, I think that this RFC is focusing too much on the `impl Trait for [FunctionType]` use-case.

Maybe it would make sense to restrict the RFC to just figuring out the syntax for `fn` items? Or make a new RFC for just the syntax? I would be willing to drive that effort, if you think so?

Implementing traits for functions have many other nuances, I'll just name a few that I don't think have been explored enough in this RFC:
- Coherence: what traits should you be allowed to implement for your function? Can I implement `IntoIterator` for my function? What if 10 years down the road the language wants to start implementing `IntoIterator` for function items of the form `fn() -> impl Iterator`?
- How might it affect inference rules if you implement a trait for your function, where the trait is also implemented for `fn` pointers.

madsmtm: I tried to find some other examples of when you'd want to talk about the function item type instead of just using the function pointer, and I honestly couldn't really find a compelling example, which is supposedly why it doesn't exist yet.

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,269 @@
...
+```rust
+async fn request_name(id: PersonID) -> String { .. }
+
+impl Requestable for async fn request_name {
+    /* ... */
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
```
madsmtm, line 97: I think it might make sense to explain this relating to [the current reference docs on function items](https://doc.rust-lang.org/reference/types/function-item.html), and how we'd update that given that we now _have_ syntax for talking about the function type.

madsmtm (changes requested):

madsmtm:
Okay so I just found [this draft RFC on named function types](https://rust-lang.github.io/impl-trait-initiative/RFCs/named-function-types.html), which contains a lot of thoughts that are very similar to what this RFC is proposing.

[I may open PRs to that repo](https://github.com/rust-lang/impl-trait-initiative/issues/17) to merge relevant parts from this RFC text into that.

text/3476-expose-fn-type.md
DasLixou: @madsmtm good point with the fn thing! I honestly wonder how that would integrate with generator functions...

text/3476-expose-fn-type.md
```diff
@@ -0,0 +1,269 @@
...
+We can take a shortcut and use the `async` keyword, as long as the `Output` assoc type in the Future is still fixed
+```rust
+async fn request_name(id: PersonID) -> String { .. }
+
+impl Requestable for async fn request_name {
+    /* ... */
+}
+```
+
+# Reference-level explanation
```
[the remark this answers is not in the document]
DasLixou: wow... that is a lot of thinking going in there... I suppose that you are going to take over the whole scenario with the more general syntax approach?

text/3476-expose-fn-type.md
madsmtm: Note to self: The standard library uses the macro `impl_fn_for_zst!` to support exactly the use-case of naming a function item, because they store it in generic helper structs, and would like that storage to be zero-cost.

jedbrown:
I found this while exploring type-handling for automatic differentiation and I'd like to share that use case. Suppose we have a function
```rs
fn f(x: &[f32], y: &[f32], z: &mut [f32]) {
    // ...
}
```
With automatic differentiation, one chooses `Reverse` or `Forward` mode (analogous to left- and right-associativity) and specifies *activity* for each argument, which leads to a combinatorial number of possible variants. An attractive option would be to be able to do something like the following. (Note that this split mode is mostly used for reverse-mode AD, but it's also relevant to forward-mode, which may be more intuitive for the reader.)
```rs
impl ForwardDiff<(Dual, Const, Dual)> for fn f {
    type Tape = Box<[f32]>;
    fn augment_forward(x: &[f32], y: &[f32], z: &mut [f32] -> Self::Tape {
        // evaluate f(x,y,z) while storing intermediate data to make subsequent differentiation faster
    }
    fn eval_forward(x: &[f32], dx: &[f32], y: &[f32], z: &mut [f32], dz: &mut [f32], tape: &Self::Tape) {
        // note that there is no dy because y's activity is Const
    }
}
```
The above could be called like
```rs
let tape = f.augment_forward(&x, &y, &mut z);
let dx = [1.0, 0.0, 0.0]; // direction in which to differentiate
let mut dz = [0.0; 3];
f.eval_forward(&x, &dx, &y, &mut z, &mut dz, &tape);
// use z, dz
```
Notes:
* `ForwardDiff</* activity */>` takes activity and it's useful to implement for different combinations (but usually not all combinations). This makes it hard to name regular methods. I'm not concerned here about precise calling syntax and whether we pass something like `Dual(&x, &dx)` instead of `&x, &dx` as separate arguments.
* It's useful to associate derivatives with trait methods (even if not object-safely).
* In a lot of materials and physics contexts, traits or structs will have many base functions (like `f`) that need associated derivatives.

jedbrown:
Also, whatever syntax is chosen, I think it's desirable to specify interaction with [`std::any::type_name_of_val`](https://doc.rust-lang.org/std/any/fn.type_name_of_val.html) so that `std::any::type_name_of_val(&f)` will return something interpretable as a type (it currently just returns the function name, which is a value).

PR #3489, opened 2023-09-10
Update 2094-nll.md

davidyangss:
the type of *p is Foo, so here use p or *q.

[Rendered](https://github.com/davidyangss/rfcs/blob/patch-1/text/2094-nll.md)

### DIFF
--- a/text/2094-nll.md
+++ b/text/2094-nll.md
@@ -919,7 +919,7 @@ let foo = Foo { ... };
 let p: &'p mut Foo = &mut foo;
 let q: &'q mut &'p mut Foo = &mut p;
 let r: &'r mut Foo = &mut **q;
-use(*p); // <-- This line should result in an ERROR
+use(p); // or use(*q); // <-- This line should result in an ERROR
 use(r);
 ```
 


davidyangss:
reborrow # Example3. 
the type of *p is Foo, so here use p or *q.

ehuss:
Is it an error to use `*q`?  The following seems to work for me:

```rust
fn r#use(_s: &mut String) {}

fn main() {
    let mut foo = String::new();
    let mut p: &mut String = &mut foo;
    let q: &mut &mut String = &mut p;
    let r: &mut String = &mut **q;
    r#use(*q); // <-- This line should result in an ERROR
    // r#use(r);
}
```

I'm not sure what the original was trying to illustrate. `*p` won't work assuming the type is not Copy, independent of NLL.

davidyangss:
I'm learning Rust recently. When I read the RFC, I thought "Example 3" might have a bug. 
Example 3，code
image
image
An "error[E0308]: mismatched types". I don't think this case is for "mismatched types".
Through my understanding of "reborrow", I think that the "reborrow" description can be reflected by changing it to use(p) or use(*q).

digama0:
I believe the way one is supposed to read that code is not to treat `use` literally as a function, but rather as "some code that uses `p`". To turn it literally into a function you would need to have `use1` and `use2`, but also (from context) I think these expressions are only meant to be used as places, not as values, because if you just did `use1(*p)` this would be an error for moving a non-copy value out of a &mut ref, which doesn't seem to be the point. So it would need to be more like `use1(&mut *p)` and `use2(&mut r)` where `use1` and `use2` have the appropriate types to make that typecheck.

davidyangss:
Thanks @digama0  for the answer. I accept this answer ( use1(&mut *p) and use2(&mut r) )

PR #3490, opened 2023-09-18
RFC: `const` functions in traits

tgross35:


tgross35:
cc @rust-lang/wg-const-eval

clarfonthey:
Going to be rather blunt: have you been following the development of const-fn-in-trait at all?

Because, this is a planned goal, but one of the reasons why keyword generics exist in the first place is to solve this problem. This RFC strikes me as being written by someone who hasn't followed that development, and just wants const trait support sooner. The problem is that we really can't offer it much sooner than finishing up the existing proposals.

Sure, we could tag some functions as `const` in traits, require that said functions are always `const`, then let you use those in const functions since the "constness" of the methods is never in question. But that's an incredibly restrictive constraint, and it doesn't really gain us much beyond our current system. A lot of the actual functionality folks are reaching for can be achieved with the current const fn system via hacks, and the stuff that *can't* won't really be helped by this change. All it seems like is a stopgap solution that will not really help much, and just delay the existing const trait effort further.

tgross35:
> Going to be rather blunt: have you been following the development of const-fn-in-trait at all?

I am not familiar with anything specifically called `const-fn-in-trait`, nor can I find any specific information on it. If this isn't the same as keyword-generics-initiative/effects/const_generic_const_fn_bounds, do you have a link? Otherwise, I believe I am fairly up to date with what is publicly available.

> Sure, we could tag some functions as const in traits, require that said functions are always const, then let you use those in const functions since the "constness" of the methods is never in question. But that's an incredibly restrictive constraint, and it doesn't really gain us much beyond our current system. A lot of the actual functionality folks are reaching for can be achieved with the current const fn system via hacks, and the stuff that can't won't really be helped by this change. All it seems like is a stopgap solution that will not really help much, and just delay the existing const trait effort further.

It is true that this RFC is encompassed by the effects goals. But effects are still very much in the experimentation stage and have a much larger and more complex scope (I have to assume we are still many months away from RFCs), so it did not seem there would be any harm in proposing a minor and fairly unobjectionable subset at this time. I also did some due diligence to ensure this wouldn't conflict [1](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Compile.20time.20reflection/near/390057245) [2](https://internals.rust-lang.org/t/blockers-to-const-in-traits-pre-effects-system/19514).

I also have to disagree that this isn't useful on its own, I come across uses quite frequently in projects that use a lot of statics (e.g. modular C+Rust projects).

programmerjake:
> I also have to disagree that this isn't useful on its own, I come across uses quite frequently in projects that use a lot of statics (e.g. modular C+Rust projects).

I know I definitely will use this, currently I have to make-do with a bunch of `const`s in traits and other annoying workarounds. I even considered building an interpreter to take a `const` from a `trait` and treat it as code that can be run in CTFE.

Matthew-Chidlow:
While some form of optional const-ness is needed in traits, regardless of how that is done, always const functions are something that you _should_ be able to do. Since a trait is a contract and so its natural to be able to add const-ness to that contract, and this is exactly what has been done for async, you can add async to the contract to force a function to always be async so extending that to const is only natural.

Plus I would definitely use this if it existed.

bluebear94:
Would [refinining](https://rust-lang.github.io/rfcs/3245-refined-impls.html) a trait method by adding `const` be a further possibility for this RFC?

tgross35:
> Would [refinining](https://rust-lang.github.io/rfcs/3245-refined-impls.html) a trait method by adding `const` be a further possibility for this RFC?

I don't intend to cover allowing `const` implementations where the trait does not require `const`, just to keep this RFC scoped to the most basic need. However, that is actually called out in the refine RFC as a possibility and I think it would interact well with this.

Thanks for bringing it up, I added that under future possibilities.

matthieu-m:
I'd like to share a concrete usecase for this feature.

In the [RFC: Introduce the Store API for great good](https://github.com/rust-lang/rfcs/pull/3446) the `Store` trait has been artificially split into `StoreDangling` which exposes a single `dangling` method and `Store` itself which exposes the rest of the methods _just_ so that dangling handles can be produced in `const` contexts -- and thus `Vec::new` and al. can be `const` -- without requiring that the entire `Store` implementation be `const` (hint: most won't be).

If this RFC were to be accepted, the Store API could be simplified to a single `Store` trait, with a required `const fn dangling(...)` method, and everyone would rejoice.

text/0000-const-fn-in-trait.md
```diff
@@ -0,0 +1,148 @@
...
+This RFC proposes allowing functions in a trait to be marked `const`, meaning
+they can be called from a const context like other const functions. Use cases
+include:
+
+- Needing a constructor that can create a static object for C-style plugins
+- Wanting to reduce code duplication in const contexts
+- Subtraits that want to provide defaults based on supertraits
+
+Workarounds typically involve a combination of wrapper functions, macros, and
+associated consts. This RFC will eliminate the need for such workarounds.
```
madsmtm, line 31: I feel it may be useful to note some of these workarounds?

Essentially, I'm searching for an illustration of what is possible (although maybe hugely inconvenient and ugly) in current Rust, and what new possibilities would be enabled by this proposal, if any?

text/0000-const-fn-in-trait.md
```diff
@@ -0,0 +1,148 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Functions within a trait can be marked const:
+
+```rust
+trait GlobalState {
+    /// Create a state that will be held in a global static
+    const fn build(u32) -> State;
```
madsmtm, line 42: Nit:
```suggestion
    const fn build(base_value: u32) -> State;
```

text/0000-const-fn-in-trait.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+And then the function can be called in const contexts, including as generic
+calls within other const functions:
+
+```rust
+/// Add a named item to our global state register
+const fn register_state<T: GlobalState>(name: &'static str, item: T) {
+    // ...
+    STATES[0] = (name, item.build())
```
madsmtm, line 63: Nit: Should provide some `base_value` to the `build` method

text/0000-const-fn-in-trait.md
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/// Or, use with a single item
+const DEFAULT_STATE: State = MyFavoriteStruct.build();
```
madsmtm, line 67: Nit: This would be called as:
```suggestion
const DEFAULT_STATE: State = MyFavoriteStruct::build(42);
```

madsmtm (approved):

BD103:
Just a small question: how would this work with trait objects? Will `const` functions still be turned into runtime functions when included in a VTable?

For instance:

```rust
trait Foo {
    const fn foo();
}

fn run_any_foo(trait_object: &dyn Foo) {
    trait_object.foo();
}

const fn run_any_foo(trait_object: &dyn Foo) {
    trait_object.foo();
}
```

If a `const` trait method is called from a `const` function, will it all be evaluated at compile time or raise an error? Will a `const` trait methods have to be included in the VTable, even if it is not necessarily called at runtime?

I don't have any specific ideas for changes to the RFC, I'm just making sure trait objects aren't forgotten. :)

text/0000-const-fn-in-trait.md
tgross35: Do you mean that you are looking for a sample of what a solution might look like today? Or just a more in depth explanation from the list?

tgross35: Maybe just a better explanation of each workaround or the use cases would be best, current examples are usually pretty large and vary a lot

tgross35:
> Just a small question: how would this work with trait objects? Will `const` functions still be turned into runtime functions when included in a VTable?
> 
> [...]
> 
> If a `const` trait method is called from a `const` function, will it all be evaluated at compile time or raise an error? Will a `const` trait methods have to be included in the VTable, even if it is not necessarily called at runtime?
> 
> I don't have any specific ideas for changes to the RFC, I'm just making sure trait objects aren't forgotten. :)

That is a great question!

At runtime, `const` functions act idential to normal functions so I think it should be the same in traits - that is, the functions will be in the VTable. I am not sure what is best with respect to calling `const-fn-in-trait` functions from a `&dyn Foo` - it seems doable at first glance but because that may come with hidden trickiness.

I think maybe it is best to leave thta under unresolved questions or future possibilities for now, since more information will probably become apparent at implementation time. I'll update the RFC soon

text/0000-const-fn-in-trait.md
madsmtm: Hmm, mostly I fear that there may be a lot of edge-cases that have not yet been considered in the proposal?

And then I think that if I saw a few instances of "this thing can already be done today, just not as cleanly", then it'd assuage my worries on that front.

clarfonthey:
Responding after re-reading the RFC, understanding for real what's actually going on. I've marked my previous comment to clarify that I really wasn't responding properly to this RFC, although I don't like deleting things since it makes discussions confusing to read.

I agree that having `const fn`s in traits which are always `const`, regardless of implementer, is reasonable. The main issue is that semantically, it's weird to distinguish between a trait method which can opt into always being meaningful in a const context, versus one which *must* be const.

The reason why I had such a knee-jerk reaction to the RFC is because, well… there's the issue of potentially abusing this feature until keyword generics are released, where people start offering const and non-const versions of traits. This will just make migrating back into keyword generics a headache, although I agree that simply not having functionality in the meantime is a bad solution.

Overall it's just really confusing to reason about `const` because it's effectively backwards what it should be. Since something applicable at `const` time is also applicable at runtime, we should instead have a `nonconst` keyword that indicates when something it *not* available at const time, which would make the logic around other keywords (like `mut` and `unsafe`) more consistent with `const`. Of course, that's not what we have now, and we're stuck with it.

fee1-dead (commented):
I think the main issue with this is that of technical debt. If we ever wanted traits like `Iterator`s to be const and usable in const contexts, then we need to introduce constness to those traits in a backwards-compatible way. This RFC cannot help with eventually making `Iterator` const, since adding `const` to a trait function would be a breaking change. This means that a future RFC that proposes "const traits", like the experimental feature we currently have under `const_trait_impl` and `effects` will eventually need to be proposed.

But, if we allow the ability to have const functions in traits, we are suggesting traits like `ConstAdd` or `ConstIterator` so that someone can use those traits in const functions with generic parameters. Perhaps this works fine when someone is implementing something like a generic factorial function and all the types they are handling can have subtraction and multiplication done in compile time, but this potentially creates fragmentation where non-const functions are unable to call generic `const` functions due to the strict trait impl requirements. If we are pushing to stabilize this faster than const traits, then this would be concerning.

Also, cc @rust-lang/wg-const-eval since I didn't get notified for the first ping it seems

oli-obk:
I would prefer to postpone this RFC until we have a good implementation of const traits. Landing this RFC would suggest that we'd also work on it, and I don't think we can do that until const the const traits work settles down, as it's already hard enough to keep one major `const` feature implementation in mind.

matthieu-m:
> This RFC cannot help with eventually making `Iterator` const, since adding `const` to a trait function would be a breaking change. 

Indeed, it cannot, but I do not see this as a problem.

We can today have inherent `impl` blocks which mix-and-match `const`, `async`, and neither-const-nor-async associated functions, and arguably we need the same for traits.

I already mentioned the usecase of the `Store` API. Today, in order to produce a dangling handle in `const` contexts, which is necessary for a `const` `Vec::new`, a separate trait had to be introduced for the one function that requires to be `const`. It's workable, but not ergonomic.

The clearly better API is for the `Store` trait to be implemented as both const and non-const -- depending on what's possible -- and for its `dangling` method to _always_ be `const`.

In the end, we need the flexibility of both `const` functions in traits _and_ `const` traits, neither one subsumes the other, they are just appropriate in different circumstances.

> I would prefer to postpone this RFC until we have a good implementation of const traits. Landing this RFC would suggest that we'd also work on it, and I don't think we can do that until const the const traits work settles down, as it's already hard enough to keep one major `const` feature implementation in mind.

I'm ambivalent on this.

I understand the motivation, however it may also be worth accepting such an RFC to signal that `const` traits will _not_ be the end of the story, and developers for which `const` methods in traits would be a better fit may want to wait for it, rather than stabilize work-arounds based on `const` traits.

oli-obk:
> I understand the motivation, however it may also be worth accepting such an RFC to signal that `const` traits will _not_ be the end of the story,

We don't even have an accepted const trait RFC 😅

Thomas-Mewily:
If forcing certain functions of a trait to be const is too restrictive, perhaps adding some generic constraint on what must be const in the trait can solve it ?

```rust
const fn get_const_default<T>() 
   where T : Default, <T as Default>::default : Const 
{
    T::default()
}
// `Const` can be a marker trait generated by the compiler I guess
```

Thomas-Mewily:
> > If forcing certain functions of a trait to be const is too restrictive, perhaps adding some generic constraint on what must be const in the trait can solve it ?
> 
> You rediscovered the WIP [effects/keyword generics](https://github.com/rust-lang/keyword-generics-initiative) system :) (of which this was intended to be a compatible subset)

Yep indeed it look like ^^ Thank for the link :)

PR #3492, opened 2023-09-19
Update 2495-min-rust-version.md links

chris-ha458:
the PR and issue have been updated with different semantics

### DIFF
--- a/text/2495-min-rust-version.md
+++ b/text/2495-min-rust-version.md
@@ -1,7 +1,7 @@
 - Feature Name: `min_rust_version`
 - Start Date: 2018-06-28
-- RFC PR: [rust-lang/rfcs#2495](https://github.com/rust-lang/rfcs/pull/2495)
-- Rust Issue: [rust-lang/rust#65262](https://github.com/rust-lang/rust/issues/65262)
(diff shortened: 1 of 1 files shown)


ehuss:
I would lean towards not approving this since we generally have a policy of not updating RFCs after they are merged. Almost all RFCs go through at least some minor changes during the implementation and stabilization phases, and those changes are almost never reflected back into the RFC. The updated information exists in the documentation, and we don't expect RFCs to be a reference source or living documents. They are a snapshot of the decision made at the time it was approved. I would prefer not to give the illusion that the RFC documents are up-to-date, because then we would need to review/change the entire RFC.

Urgau:
We discussed this pull request during @rust-lang/internal-sites triage of non-RFC PRs in [#t-internal-sites > Stale non-RFC PRs @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/538071-t-internal-sites/topic/Stale.20non-RFC.20PRs/near/546011978).

As mentioned in [The RFC life-cycle](https://github.com/rust-lang/rfcs?tab=readme-ov-file#the-rfc-life-cycle), RFCs are historical snapshots of a decision made in the past and "only very minor changes should be submitted as amendments".

Therefore, we are closing this PR in accordance with the policy.

PR #3494, opened 2023-09-20
unauthenticated @rustbot label T-*

eopb:
This is allowed in [`rust-lang/rust`](https://github.com/rust-lang/rust/blob/62d9034a0d571b78e518727d6cb4b090569e5238/triagebot.toml#L12).

Others have also expected it to work in this repo:
- https://github.com/rust-lang/rfcs/pull/3490#issuecomment-1722720545
- https://github.com/rust-lang/rfcs/pull/3485#issuecomment-1712657835

@rustbot label not-rfc

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -1,5 +1,6 @@
 [relabel]
 allow-unauthenticated = [
     "A-*",
+    "T-*",
     "not-rfc",
 ]


ehuss (approved):
Seems reasonable to me, thanks! The team assigned to an RFC is an important property, but I don't think this is likely going to be abused, and it is easy to repair if a mistake is made, and this can always be reverted if it is a problem.

PR #3512, opened 2023-10-10
Add closure-move-bindings RFC

SOF3:


text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,285 @@
...
+dbg!(foo); // foo is now 2
+```
+
+You can add a `move` keyword in front of the closure
+to indicate that all captured bindings are moved into the closure
+instead of referenced:
+
+```rs
+let mut foo = 1;
+let mut closure = || { foo = 2; };
```
programmerjake, line 62:
```suggestion
let mut closure = move || { foo = 2; };
```

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,285 @@
...
+> &nbsp;&nbsp; (_Expression_ | `->` _TypeNoBounds_&nbsp;_BlockExpression_)>\
+> _MoveBindings_ :\
+> &nbsp;&nbsp; `(` ( _MoveBinding_ (`,` _MoveBinding_)<sup>\*</sup> `,`<sup>?</sup> )<sup>?</sup> `)`\
+> _MoveBinding_ :\
+> &nbsp;&nbsp; _NamedMoveBinding_ | _UnnamedMoveBinding_\
+> _NamedMoveBinding_ :\
+> &nbsp;&nbsp; _PatternNoTopAlt_ `=` _Expression_\
+> _UnnamedMoveBinding_ :\
+> &nbsp;&nbsp; `mut`<sup>?</sup> ( _IdentifierExpression_ | _MethodCallExpression_ )\
+> &nbsp;&nbsp; ( _Identifier_ `=` | `mut`)<sup>?</sup> ( _IdentifierExpression_ | _MethodCallExpression_ )\
```
programmerjake, line 134: this looks like an extraneous line

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+let mut closure = move || { foo = 2; };
+closure();
+dbg!(foo); // foo is still 1, but the copy of `foo` in `closure` is 2
+```
+
+Note that `foo` is _copied_ during move in this example
+as `i32` implements `Copy`.
+
+If a closure captures multiple bindings,
+all the `move` keywoed makes them all captured by moving.
```
Veykril, line 71:
```suggestion
all the `move` keyword makes them all captured by moving.
```

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+> &nbsp;&nbsp; ( `||` | `|` _ClosureParameters_<sup>?</sup> `|`  )\
+> &nbsp;&nbsp; (_Expression_ | `->` _TypeNoBounds_&nbsp;_BlockExpression_)>\
+> _MoveBindings_ :\
+> &nbsp;&nbsp; `(` ( _MoveBinding_ (`,` _MoveBinding_)<sup>\*</sup> `,`<sup>?</sup> )<sup>?</sup> `)`\
+> _MoveBinding_ :\
+> &nbsp;&nbsp; _NamedMoveBinding_ | _UnnamedMoveBinding_\
+> _NamedMoveBinding_ :\
+> &nbsp;&nbsp; _PatternNoTopAlt_ `=` _Expression_\
+> _UnnamedMoveBinding_ :\
+> &nbsp;&nbsp; `mut`<sup>?</sup> ( _IdentifierExpression_ | _MethodCallExpression_ )\
```
Veykril, line 133: This requires infinite lookahead for parsing if I am not mistaken, a method call expression can begin with any expression which we cannot differentiate from a pattern without attempting to parse the whole construct first. This is I believe the reason as to why assignment destructuring (`pat = expr`) doesn't parse full patterns either, it accepts a subset of expressions on the lhs and interprets those are patterns.

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+> _ClosureParameters_ :\
+> &nbsp;&nbsp; _ClosureParam_ (`,` _ClosureParam_)<sup>\*</sup> `,`<sup>?</sup>\
+> _ClosureParam_ :\
+> &nbsp;&nbsp; _OuterAttribute_<sup>\*</sup> _PatternNoTopAlt_&nbsp;( `:` _Type_  )<sup>?</sup>
+
+Closure expressions are clsasified into two main types,
```
Veykril, line 139:
```suggestion
Closure expressions are classified into two main types,
```

text/3512-closure-move-bindings.md
SOF3: As specified below, the MethodCallExpression always starts with an identifier, so this is not a parsing issue right now. But it is indeed problematic if we want to relieve this restriction in the future.

SOF3: My mental scratch would be to parse it like this:

```
if peek(Identifier) && !peek("=") {
    binding = parse(UnnamedMoveBinding)
} else {
    binding = NamedMoveBinding { parse(Pattern), parse("="), parse(Expr) }
}
```

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+}
+```
+
+This RFC proposes a more concise syntax to express these moving semantics.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A closure may capture bindings in its defining scope.
+Bindings are captured by reference by default:
```
veber-alex, line 47: This is not accurate.
Bindings are "captured by example", so if the closure needs a reference to the value it's captured by reference and if needs an owned value it's captured by move...except for `Copy` types which always capture by reference unless you `move`.

Issue with more details:
https://github.com/rust-lang/rust/issues/100905

I think we should consider changing this behavior in a new edition and seeing if this helps without needing new syntax.
Personally for me every time I need the workarounds described in this RFC it was because of this.

SOF3: Does it sound reasonable to detect this as a breaking change and fix/warn in automated edition migration tools?

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+```rs
+{
+    let foo = &foo;
+    move || run(foo, bar)
+}
+```
+
+- It is a very frequent scenario to clone a value into a closure
+(especially common with `Rc`/`Arc`-based values),
+but even the simplest scenario requires three lines of boilerplate:
```
teor2345, line 32: I might be missing some implied boilerplate here, but is this just two lines?
```suggestion
but even the simplest scenario requires two lines of boilerplate:
```

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+let mut closure = move || { foo = 2; };
+closure();
+dbg!(foo); // foo is still 1, but the copy of `foo` in `closure` is 2
+```
+
+Note that `foo` is _copied_ during move in this example
+as `i32` implements `Copy`.
+
+If a closure captures multiple bindings,
+all the `move` keyword makes them all captured by moving.
```
teor2345, line 71: Extra word?

```suggestion
the `move` keyword makes them all captured by moving.
```

text/3512-closure-move-bindings.md
SOF3: It could be just one line (and maybe even inlined in another expression). So that's 3.5 lines of boilerplate.

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+it starts with a `move` token immediately followed by a `|` token,
+without any parentheses in between.
+
+## _ImplicitReference_ closures
+
+When the parentheses for _MoveBindings_ is present, or when the `move` keyword is absent,
+the closure expression is of the _ImplicitReference_ type, where
+all local variables in the closure construction scope not shadowed by any _MoveBinding_
+are implicitly captured into the closure by shared or mutable reference on demand,
+preferring shared reference if possible.
```
danielhenrymantilla, line 151: I'd like to ask for a special sigil for this "and other implicit captures" behavior:

```rs
move(x, y, ..) || { … }
```

Indeed, by requiring a trailing `..` in order to allow implicit captures:
  - such implicit capture is now clearer: the `x, y` enumeration is now visibly non-exhaustive;
  - it gives the option not to provide `, ..`, thereby requiring exhaustive enumeration of captures: _this gives the power to the person defining the closure to **restrict** the kind of captures a closure may use_.
      - Given that the motivation for this syntax is finer-grained control, it would be a pity not to be featuring such a capability, and have to wait for a later edition-requiring amendment to get into this behavior.
      - reasons for restricting external captures can be:
          - soundness of more advanced designs (_e.g._, macros and advanced lifetime shenanigans);
          - especially in FFI or other type-erased scenarios where certain lifetimes may have been lost;
          - (or when worried about things like `{A,}Rc` cycles in case of implicit by-move captures)
          - better control of a closure size;
          - _helps teachability of how closures work_: `move(x) |…| { … }` would always be an `x`-capturing closure, no matter the body of `...`, thereby properly decoupling implementation details from API surface (of the closure type).
  - it mirror struct pattern destructuring, wherein a `Foo { x, .. }` pattern is not expected to be exhaustive over the fields it has, whereas `Foo { x }` is.

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+
+Due to backwards compatibility, this RFC proposes a new syntax
+that is an extension of capture-by-move
+but actually looks more similar to capture-by-reference,
+thus confusing new users.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Capture-by-reference is the default behavior for implicit captures for two reasons:
```
danielhenrymantilla, line 206: I'd like to suggest the opposite, here, actually: for `move(..)` (or `move()` if my suggestion for `..` is disregrejected) to mean the same as `move`.

  - big consistency gain: `move(..) ||` being closer to `move ||` than to `||`;
  - make by-reference captures be the odd one that stands out. I agree this may be subjective, but bear with me:

    There are two kind of closures:
      - immediately-called closures, such as with `Option::map()` or `panic::catch_unwind()`;
      - long-lived closures (`: 'long`), such as the `spawn()` family.
    The former, 99% of the time, accomodates to fully implicit `|| {}` capture rules, and the latter, kind of accomodate to `move ||` closures, except for when specific "right-before `move` adjustments" are needed, which is _very often_ just a `.clone()` (and sometimes an `Arc::downgrade()`).

    This very RFC, by providing explicit captures, is thus mostly helping the latter case: the one with `: 'long`-lived requirements. In such a scenario, capturing by reference is rather non-desired, or something niche stemming from a long-lived borrowable (_e.g._, a variable with outlives the `'env` of a `scope::spawn()`)). It thus does seem rather legitimate to favor by-`move` captures for the `..`-implicit ones, much like `move ||` does.

  - lastly, although I agree this may not be that important to others, I feel like it would be the most helpful syntax w.r.t. teachability: having the word `move` not be neutered by some explicit `&` sigil (_e.g._, in `move(&x) || x + 42)`) mean a by-ref capture kind of denaturates/dilutes the meaning of the word `move`

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+> &nbsp;&nbsp; ( `||` | `|` _ClosureParameters_<sup>?</sup> `|`  )\
+> &nbsp;&nbsp; (_Expression_ | `->` _TypeNoBounds_&nbsp;_BlockExpression_)>\
+> _MoveBindings_ :\
+> &nbsp;&nbsp; `(` ( _MoveBinding_ (`,` _MoveBinding_)<sup>\*</sup> `,`<sup>?</sup> )<sup>?</sup> `)`\
+> _MoveBinding_ :\
+> &nbsp;&nbsp; _NamedMoveBinding_ | _UnnamedMoveBinding_\
+> _NamedMoveBinding_ :\
+> &nbsp;&nbsp; _PatternNoTopAlt_ `=` _Expression_\
+> _UnnamedMoveBinding_ :\
+> &nbsp;&nbsp; `mut`<sup>?</sup> ( _IdentifierExpression_ | _MethodCallExpression_ )\
```
danielhenrymantilla, line 133: I think we should also have `move(&x, &mut y)` syntax, for the specific cases of explicit by-reference captures:

  - not only for my suggested implicit-by-`move` change, should it be accepted;
  - but even in the case of the current implict-by-ref-captures semantics of the RFC, when wanting to be explicit, it would be nicer to be using `move(&x, &mut y)` than `move(x = &x, y = &mut y)`, especially when `x` and `y` are `longer_variable_names`.
  - there would otherwise be no way to mimic, explicitly, the "magic binding" semantics which implicit by-ref closures captures currently have (_e.g._, when having `x: i32`, an implicit by-ref capture of `x` makes it so the closure has captured `&x`, but the `x` name is still fully usable to refer to the `*&x` place).

If we didn't have this sugar, I could very well see myself having a `.by_ref()` and `.by_mut()` couple of blanket-implemented extension methods.
  - whilst these methods could be a bearable alternative, I don't think it's worth the hassle w.r.t. adding another 1-`lookahead` case for these parsing rules:

```rs
    $binding:ident $(. $method…)?
OR
    &$(mut)? $binding:ident
OR
    $binding:pat = $e:expr
```

text/3512-closure-move-bindings.md
hamza1311: > Indeed, by requiring a trailing `..` in order to allow implicit captures:

I assume you mean implicit captures _by-value_? or keeping the current default capture behavior (without `move`)? `Foo { x, .. }` means all fields except `x` are ignored so it would make sense if all fields other than `x` were ignored in `move(x, ..)` when doing move captures

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+```rs
+let foo = vec![1];
+let mut closure = move(mut foo = foo.clone()) || {
+    foo.push(2);
+};
+closure();
+dbg!(foo); // the outer `foo` is still [1] because only the cloned copy was mutated
+```
+
+The above may be simplified to `move(mut foo.clone())` as well.
```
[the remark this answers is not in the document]
hamza1311: I agree this should be left for future, if added at all.

By looking at `move(mut foo.clone())`, it is unclear what it should bind to. In this situation, `foo` is the only ident but in `move(mut foo.bar.clone())`, is it supposed to be bound to `bar`  or `foo`? That sort of ambiguity isn't a good thing to add.

hamza1311 (approved):
Finally! After years of discussion, thanks @SOF3 for finally pushing forward and creating the RFC

text/3512-closure-move-bindings.md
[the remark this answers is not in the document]
SOF3: that's why it is strictly limited to a single identify in the method call receiver.

camsteffen:
Syntax bikeshed. My brain really wants to parse `move(x)` as a function call. What if it looked like this?

```rust
move { x, y: y.clone() } || foo(x, y);
```

This syntax is analogous to a struct literal. In that way I think it might be more intuitive, while also less likely to be confused as something it's not (function call) because the lowercase `move` is an upfront hint that it is not a struct.

danielhenrymantilla:
[my .02c] @camsteffen I agree the `move()` may look a bit too much like a call, making the braced syntax appealing; but the issue that could be pointed out, then, is that it doesn't map that well to the shorthand syntax of `x.clone()` (or my suggested `&var`):

```rs
move { x, &y, w.clone(), z: foo } |…| { … }
```

  - (it also has the issue of introducing yet another whitespace in this syntax, contrary to `()` and `[]` which "stick" to `move`)

Maybe an in-between, which, for better or for worse, would resemble C++, would be to be using square brackets? 🤷 

```rs
move[x, &y, w.clone(), z = foo] |…| { … }
```

___

Be it as it may, I'd be fine with any choice among these three syntaxes, on condition that it not preclude the shorthand capture syntax.

CryZe:
`async move { x, y, z } { x }` would probably be ambiguous, otherwise I really like that syntax.

nhuurre:
If I may add my two bikeshedding cents, I think the syntax is cleanest when name bindings look like `let`-statements. That is, instead of a `move`-clause before the closure I propose a `move let` as the first statement in the closure body.
```rust
|| {
    move let foo = foo.clone();
    foo.run()
}
```
It is of course potentially confusing that the `move let` statement appears inside the closure body but is executed immediately on closure construction, not invocation. I consider this analogous to `static` items that may appear inside the body but are executed at compile time:
```rust
|| {
    static X : T = f1(); // f1 called once at compile time
    let foo : T = f2();  // f2 called every time the closure is called
    foo.bar(&X)
}
```
Hopefully the requirement that `move let` must be the first statement in the closure makes it sufficiently distinct from the rest of the body even without delimiters.

Also, I don't think omitting the `let` causes any ambiguity, i.e. the syntax could be just `move $PATTERN = $EXPR;` instead of `move let $PATTERN = $EXPR;`. You could write `move x;` as shorthand for `move x = x;`.

The existing syntax of `move` before the closure
```rust
move || {
    x + y
}
```
could be interpreted as shorthand for listing every implicitly captured name:
```rust
|| {
    move (x,y);
    x + y
}
```

SOF3:
@nhuurre I think the case for `static` is a bit different, in that `static` initializers are always const expressions so it does not matter when they are executed, but `move` initializers may have side effects and the execution time matters.

SOF3:
@camsteffen maybe it's just me, but `move { foo } || foo()` looks like an expression joining a struct literal and a function call with a logical or as well.

That said, it has never occurred to me (maybe it does to those unfamiliar with Rust) that the following looks like a logical or expression:

```rs
move || foo()
```

As with @danielhenrymantilla's comment, it is equally likely to read `move[foo] || foo()` as a logical or expression with an array index.

Maybe it's because almost every syntax highlighter (no matter semantic or lexical) always highlights the `move` keyword? In that case, it would be obvious that `move()` is not a function call because it involves a keyword, as much as `break (a, b)` wouldn't cause ambiguity that we are calling a function called `break` with arguments `a` and `b`.  From this perspective, perhaps it's better to leave a space between `move` and the `(` in the style guide?

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,284 @@
...
+
+If a closure captures multiple bindings,
+the `move` keyword makes them all captured by moving.
+To move only specific bindings,
+list them in parentheses after `move`:
+
+```rs
+let foo = 1;
+let mut bar = 2;
+let mut closure = move(mut foo) || {
```
SOF3, line 78: I just noticed that GitHub renders `move()` with the color of a function call instead of as a keyword. This adds to the previously discussed problem of ambiguity as a function call.

nikomatsakis:
I have previously written a [draft RFC](https://hackmd.io/@nikomatsakis/SyI0eMFXO) around adding capture clauses -- I still like that design. I'd love to see it compared/contrasted with this one (haven't had time to fully read the RFC yet), as well as to know whether this one achieves the various bits of motivation.

One obvious difference is that I did not support `move(x = y+1)`, though that would be an easy extension. I did support `move(x.clone())` which would effectively be shorthand for `move(x = x.clone())`.

SOF3:
Based on the comments above, I am planning to separate the RFC into multiple features:

- `closure_move_bindings`: Introduce the syntax `move(foo = bar) || expr` and `move(mut foo = bar) || expr`, where `foo` is an ident and `bar` and `expr` are expressions. This is just an alternative syntax for `{ let mut foo = bar; || expr}` with the appropriate move semantics.
- `closure_move_binding_ident_shorthand`: Introduce the syntax `move(foo) || expr` and `move(mut foo) || expr`, where `foo` is an ident. This is an alternative syntax for `move(foo = foo) || expr`.
- `closure_move_binding_method_call`: Introduce the `move(foo.bar()) || expr` syntax, where `foo` and `bar` are idents. This is an alternative syntax for `move(foo = foo.bar()) || expr`.

SOF3:
Another issue: Is it better to use `:` instead of `=` to separate the pattern and the expression? That would be more consistent with the `field: expr` syntax in struct literals, which would make more sense if `field: field` can be shortened into `field`.

danielhenrymantilla:
@SOF3, regarding `name: value` _vs._ `name = value`:
  - it does have a nice "symmetry" with braced struct syntax, _assuming braces being used as delimiters_;
  - this, in turn, "curses" the syntax not to have too much feature creep, lest there be inconsistencies between braced struct syntax and this closure capture one:
      - `move(&x, &mut y)` would have to become `move { ref x, ref mut y }` (and there seems to be kind of an implicit but general consensus that `ref [mut]` is confusing for newcomers and something which new syntax should try to avoid).
      - what about `move(var.method())`? No such braced syntax.

So even if the different shorthands end up split as follow-up features / RFCs, if keeping them in mind, we may actually want to avoid `var: value` syntax for the captures of this RFC

text/3512-closure-move-bindings.md
joshtriplett: I think, rather than adding `move(..)` (which could just be written `move`) or `move()` (which could just be omitted), we should have a warn-by-default lint for an empty capture list (with a suggestion that removes it and a note that mentions using `move` if you want to capture everything), and add it to the list of lints suppressed in macros (since macros might want to generate a capture list).

text/3512-closure-move-bindings.md
```diff
@@ -0,0 +1,281 @@
...
+This RFC intends to solve the ergonomic issues for these common scenarios first
+and leave more room for future enhancement when other frequent patterns are identified.
+
+Alternative approaches previously proposed
+include explicitly adding support for the `clone` keyword.
+This RFC does not favor such suggestions
+as they make the fundamental closure expression syntax
+unnecessarily dependent on the `clone` language item,
+and does not offer possibilities for alternative transformers.
+
```
joshtriplett, line 227:
```suggestion

## `move` inside parameter list

As an alternative to having `move(binding)` ahead of a closure, we could put `move binding` in the parameter list: `|x: T, move y, move z = z.clone()| { ... }`.

This would have the advantage of keeping the list of declared names in one place, and giving a view of closures as having some bindings passed in as arguments and other bindings captured from the containing scope.

However, this would have the disadvantage of visually looking like the caller could pass the value in as a parameter, which it cannot.
```

text/3512-closure-move-bindings.md
SkiFire13: `move` is already a keyword and cannot be the identifier of a function, so `move(` unambiguously starts a captures list.

text/3512-closure-move-bindings.md
SkiFire13: +1 for this

If `..` also allowed to specify a default capture mode it would be a killer feature for UI development and other similar cases where a lot of variables are captured by cloning.

> `Foo { x, .. }` means all fields except x are ignored

IMO `..` just means "the rest" when used as a pattern, as it always matches all the elements that weren't already in some other way. It just so happens that with structs/tuples it doesn't allow to create bindings for the matches elements, though with slices this is possible (e.g. `[head, tail @ ..]`).

rben01:
To match `use<'captures>` syntax, I think this should be spelled `move<x, y, z>`. Then we'd have a general pattern of a keyword followed by angle brackets signifying that we're parameterizing a bit of syntax. `move(x, y, z)` looks too much like a function call.

This could also open the door to other closure-modifying keywords, like `clone<x, y, z>`. `clone(x, y, z)` cannot be permitted due to `clone(...) || {...}` parsing ambiguously as either `function-call OR expr` or `closure-modifier closure-declaration`, but I believe `clone<x, y, z> || { }` parses unambiguously. (This is not an issue for `move` because it's already a keyword. But angle brackets would, I think, let us add new closure modifiers in a backwards compatible way.)

programmerjake:
> I believe `clone<x, y, z> || { }` parses unambiguously.

there's ambiguity when used where a comma-separated list of expressions is accepted. e.g.: `f(clone<x, y, z>||{})` could be parsed as `f((clone < x), y, (z > (|| {})))`

SkiFire13:
> To match `use<'captures>` syntax, I think this should be spelled `move<x, y, z>`. Then we'd have a general pattern of a keyword followed by angle brackets signifying that we're parameterizing a bit of syntax. `move(x, y, z)` looks too much like a function call.

Note that the use of angle brackets is already generally a pattern for parameterizing some **type level** value. Extending this to normal expressions would make it lose this meaning.

SOF3:
The items in `use<>` are all generics. Bindings are not generics.

CheaterCodes:
Because of #3680, Ive also had some thoughts about this that I'd like to share:
- From my understanding, captures operate on places, not on bindings. Even if we don't do this in the first round, it would be good to keep open the door to allow capturing specific paths explicitly in the future.
- I believe we should enable the possibility of making an exhaustive list of captures. My first approach would be that `move() || {}` actually prevents all captures, `move(..) || {}` moves every implicit captures and `move(auto ..) || {}` uses the default (inferred) capture mode for every non-specified captures.
- There are currently four capture modes for closures:
  - immutable borrow `ref x` or `&x`
  - mutable borrow `mut ref x` or `&mut x`
  - move `x`
  - unique immutable borrow, which is... uh... yeah idk, maybe not important? But worth thinking about.
- Having an explicit `clone x` would be nice for ergonomics, lots of people want that. But I think it's tricky because
  - it's unclear what exactly gets cloned for e.g. an `&Rc<RefCell<MyStruct>>>`
  - It doesn't work well with path based captures
- Defining inline bindings `x = Arc::clone(x)` fixes the cloning problem, though I'd argue it's syntactically not *much* better than capturing by reference and cloning inside the closure. However, I suppose it's actually more similar to cloning *outside* the closure (closure doesn't borrow a local), so maybe it is significantly better, but that should probably be explained in the RFC

PR #3523, opened 2023-11-10
Create 0000-cargo-dns.md

hydroper:
[Rendered](https://github.com/hydroper/rust-rfcs/blob/master/text/0000-cargo-dns.md)

### DIFF
--- a/text/0000-cargo-dns.md
+++ b/text/0000-cargo-dns.md
@@ -20,7 +20,7 @@ It is predictable that the authors of `com` and `org` are unlikely to introduce 
 The top-level `org` and `com` package namespaces at Cargo are extended to allow organizations to introduce their own crates to them,
 and Cargo packages are allowed to contain multiple `::` delimiters in their names. Packages originating from organization namespaces are referred to as *domain-tied* packages.
 
-Names that conflict with existing items of the existing crates `org` and `com` at crates.io result in a compile time error.
+Names that conflict with existing items of the existing crates `org` and `com` at crates.io result in a compile time error if they are in scope aand these items are modules.
 
 Publishing domain-tied packages currently requires that the GitHub user is a member of the domain's organization. The domain's organization
 is taken from the manifest's repository URL. Building domain-tied packages locally does not require authentication.
@@ -31,7 +31,7 @@ is taken from the manifest's repository URL. Building domain-tied packages local
 This proposal adds the following functionality to Cargo:
 
 * A package's name is allowed to contain multiple `::` delimiters. The subsequent segment in the name is a subcrate of the previous segment's crate.
-* Any user is allowed to publish domain-tied packages; that is, packages belonging the namespace of the existing packages `org` and `com`.
+* Any user is allowed to publish domain-tied packages; that is, packages belonging to the namespace of the existing packages `org` and `com`.
   * The publishing process deduces the domain's organization from the manifest's repository URL.
   * It is a publishing error if the manifest's repository URL does not specify an organization-authored repository but rather a personal-authored repository.
   * It is a publishing error if the authenticated GitHub user is not a member of the organization.
--- a/text/0000-cargo-dns.md
+++ b/text/0000-cargo-dns.md
@@ -1,0 +1,72 @@
+- Feature Name: `cargo_dns`
+- Start Date: 2023-11-10
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC adds support for a organization Domain Name System (DNS) to Cargo.
+
+# Motivation
+[motivation]: #motivation
+
+Packages published at a certain domain, either in the form `org::q::p` or in the form `com::q::p`, clarify that `q` is an organization and not a project.
+It is predictable that the authors of `com` and `org` are unlikely to introduce their own subpackages to these packages.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The top-level `org` and `com` package namespaces at Cargo are extended to allow organizations to introduce their own crates to them,
+and Cargo packages are allowed to contain multiple `::` delimiters in their names. Packages originating from organization namespaces are referred to as *domain-tied* packages.
+
+Names that conflict with existing items of the existing crates `org` and `com` at crates.io result in a compile time error.
+
+Publishing domain-tied packages currently requires that the GitHub user is a member of the domain's organization. The domain's organization
+is taken from the manifest's repository URL. Building domain-tied packages locally does not require authentication.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This proposal adds the following functionality to Cargo:
+
+* A package's name is allowed to contain multiple `::` delimiters. The subsequent segment in the name is a subcrate of the previous segment's crate.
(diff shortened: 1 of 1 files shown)


programmerjake:
maybe it would be better to put all dns names under `net::` (since `net` isn't used currently), e.g. `example.com` would be `net::com::example` and `sub.example.test` would be `net::test::example::sub`

programmerjake:
and if the repo's ip address was `1.2.3.4` it'd be under `net::arpa::in_addr::_1::_2::_3::_4` (based off [reverse DNS lookups](https://en.wikipedia.org/wiki/Reverse_DNS_lookup#IPv4_reverse_resolution))

clarfonthey:
I hate this change for the same reason that I hate it in Java and co., and I especially hate it in Go!

This introduces all sorts of nonsense issues like:

* Our organization has a dot-com and a dot-org domain. Maybe some packages should go under one, and some under the other? In that case, there's no top-level link to the same organization any more.
* At what part does the host end and the resource begin? (e.g. for io.github.rust-lang, does this mean rust-lang.github.io or github.io/rust-lang? this is why GitHub uses .io for pages and .com for repos, but it's not a good reason to use this system, just a hack to get around it)
* What benefit do we get from having to type five extra letters before every package name? (`com::`, `org::`, etc.)
* What if a project's URL changes? Should the package name change too? (this is why I especially hate it in Go; the project URL *is* the package name, meaning that changing that *is* a breaking change)

And all of this for… what, exactly? Don't get me wrong, I think that DNS *can* be a nice system for things, but definitely not here.

hydroper:
Like I said, this is not a required change so far. If it proves to be useful futurely, I leaved my RFC on. The publishing process could be more complex to require registered domains.

Nilstrieb:
This RFC does not describe a problem. Solutions without problems are not useful

hydroper:
@Nilstrieb I am well aware, the thing is that the future is unknown and it might be good to worry of the `net`, `com` and `org` crates.

carols10cents:
> This is not an urgent or necessary feature currently.

The crates.io team has very little bandwidth, so any non-urgent or unnecessary features are extremely unlikely to happen. 

For this proposal in particular, it seems like a separate package registry website could implement this sort of scheme to prove it out first, without needing support from the crates.io or cargo teams. That would be my preference at this time.

epage:
I'd recommend starting with a PreRFC on [internals](http://internals.rust-lang.org/) to help vet an idea and collect input. For example, this isn't acknowledging the existing packages-as-namespaces RFC that is going through FCP for some of the relevant teams. This also should go into more discussion about how this ties into any existing packages this conflicts with.

Eh2406:
Why are `com` and `org` special? What about other top level domains like `rs`, `io` and `blog`? What happens if new top-level domains are added that conflict with existing packages? For that matter, how to handle the existing https://crates.io/crates/com and https://crates.io/crates/org ?

ahicks92:
Also people are working on namespacing already, so is this mechanism supposed to be in addition to that or replacing it?

We only really need one namespacing solution.

This is also tied to GitHub, and what stops me from just putting a random url in my package's repository field in order to publish under whatever top-level domain I want?  There has to be some kind of validation to make this work.

I think that the better solution is to allow users to register names via some UI in crates.io or something, if we did want to go here.  If those have to be dns, then maybe also force them to validate via txt records as GSuite and similar do.

In addition to this being unnecessary/low priority, it is very much underspecified.  What about cname?  Which DNS resolver will crates.io use?  What about mirrors and other registries?  What's the plan for phasing this in so that it's not a giant slap in the backward compatibility face when a project wants to bump MSRV to migrate to it?

What happens if a domain or GitHub org changes owners?  Is that manual?  If it's manual, handled by who?  How does one authenticate such a claim vs hacking/social engineering?

I could probably think of more but I agree this isn't ready to be an RFC.

AaronKutch:
Go exemplifies a particular kind of surface level bikeshedding that looks good on the surface and for first pass toy examples, but fails to see the insane non-local turmoil that is forced upon you. For example, public/private functions are differentiated by capitalizing the first letter of a function. This has elegant reasons, but assumes that you get your interfaces right the first time, and eliminates the future possibility of things like `pub(in mod...)` in Rust. There are other things, but the absolute worst of them is Go's import system. `go The feature I always

AaronKutch:
Go exemplifies a particular kind of surface level bikeshedding that looks good on the surface and for first pass toy examples, but fails to see the insane non-local turmoil that is forced upon you in the inevitable need to eventually refactor a large project or try to incorporate something modified for your own uses. For example, public/private functions are differentiated by capitalizing the first letter of a function. This has elegant reasons, but assumes that you get your interfaces right the first time, and eliminates the future possibility of things like `pub(in mod...)` in Rust. Ctrl+F and replace won't save you if it is some commonly named getter or setter. There are other things, but the absolute worst of them is Go's import system. DNS paths are much longer and inherently more difficult to change. When changing just the middle of the paths to change organizations or using different variations of the same analogous suite, it is easy to miss it in one place and have the old thing still being used without you realizing. I have personally encountered the intense displeasure of renaming repos and trying to find the magical combination of`go mod vendor` `replace` to get the specific commit I want. The Rust `Cargo.toml` feature I always want to reach for is either the `git`+`rev` or `path`+ a workspace, which just works™ on the first try, >95 percent of the time. Changing it requires merely one point.

ahicks92:
Well, even with this RFC I believe that the config goes in Cargo.toml anyway so that's not a concern.  Worst case you just don't compile anymore, there can't be hidden internal dependencies to a module.

Now that said looking at this again, to add one more thing to the problem pile, you can't rely on the org and com packages not to publish a new release.  It's not sufficient to say that it's illegal to publish a package which conflicts with an item exported by those crates because those crates can change their mind tomorrow.  It's also almost certainly a lot of work to make anything remote understand the current state of those crates without building them, and of course there's also all the old versions people might depend on as well.  I'm not a fan of this RFC at all, to be honest, but at least as regards this part of it the only sensible thing you can do is say that it is not valid for a user to depend on both some package `a` and also some package `a::b`.  Everything else (e.g. validation, allowing it to work if there's not overlaps) relies on `a` being a well-behaved member of the ecosystem and taking on additional constraints such as never adding a new module with a short, generic name.

It's not necessary for anyone to own their own domain name either. You just need a way for a user to register some sort of namespace--if Google really is attached to this whole com.google.bla thing they can go register com_google or something.  We don't currently have any namespacing at all and I don't know which of many things is the latest iteration of that, but "you own this top-level name congrats" is a much simpler solution, no need to tie it to dns.

carols10cents:
This needs more iteration in a pre-RFC on internals before it's ready for discussion as an RFC here.

@rfcbot fcp close

SOF3:
It is confusing whether we are talking about a Cargo policy or a crates.io policy. If it is just for crates.io, all users are already authenticated through GitHub anyway, and I don't see why additionally using DNS helps.

ahicks92:
That's not particularly relevant and is one reason this RFC is underspecified. While it is the case that crates.io auths via GitHub, you may put whatever you want in the repository field including links to non-GitHub repositories and repositories you don't personally own.

hydroper:
I like domains due to readability and closeness to application IDs:

- `com::ea::n4s` versus `ea::n4s`
- `org::generalrelativity::foam` versus `generalrelativity::foam`
  - The latter looks like a category rather than an organization name

Relying on GitHub organizations may be removed as part of the proposal in the future. I'm not entirely familiar with domain setup; although I had purchased one temporarily around 2015, I didn't deal with records like you said.

This is an extension to the namespacing proposal, not a replacement, and I don't need to clarify anything in the PR I think.

hydroper:
My PRs have a tendency of being emotionally reacted, but I don't really care!

hydroper:
I am not planning to spend time on this now though, so closing...

PR #3539, opened 2023-12-11
chore: fix typo in Pin README

1eokim:
fixed a typo in the Pin RFC

[Rendered](https://github.com/1eokim/rfcs/blob/patch-1/text/2349-pin.md)


ehuss (approved):
Thanks!

PR #3544, opened 2023-12-16
t-opsem RFC: link to t-types RFC

RalfJung:
Editorial change:
The t-opsem RFC references the t-types RFC, but doesn't actually link to it. The RFC is not called "t-types" so searching for that won't even bring it up. Let's do the internet thing and put in a hyperlink. :)

[Rendered](https://github.com/RalfJung/rfcs/blob/t-opsem-link/text/3346-t-opsem.md)

### DIFF
--- a/text/3346-t-opsem.md
+++ b/text/3346-t-opsem.md
@@ -5,7 +5,7 @@
 
 # Summary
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #3568, opened 2024-02-12
Add renovatebot config file

Turbo87:
see https://github.com/rust-lang/rfcs/pull/3566#issuecomment-1939074653

This adds a basic config file with updates throttled to once-per-month. Once this is merged the corresponding GitHub App needs to be enabled by the infra team for this repo.


ehuss (approved):
Thanks! I'm not very familiar with renovatebot, but seems like a good start.

PR #3571, opened 2024-02-12
Fix typo: Replace `unsafe` with `unsafe_code`, because `unsafe` is not a lint.

Feriixu:
RFC 3389 refers to a nonexistant lint called `unsafe`. However `unsafe_code` does exist.
This is slightly confusing for readers. This PR does not change the meaning of the RFC, therefore I think this is an improvement.

Edit: Apologies, GitHub had automatically put a link to the RFC as this comment. I have since added an explanation.


ehuss (approved):
Thanks!

PR #3576, opened 2024-02-20
RFC: RArrow for Pointer Ergonomics

EEliisaa:
This RFC improves ergonomics for pointers in unsafe Rust. It adds the RArrow token as a single-dereference member access operator. `x->field` desugars to `(*x).field`, and `x->method()` desugars to `(*x).method()`.


clarfonthey:
For an RFC, you're not supposed to edit the template; you're supposed to copy it into a new file. As-is, this couldn't be merged.

PR #3587, opened 2024-03-12
Update 2294-if-let-guard.md

chdalski:
Fix typo

[Rendered](https://github.com/chdalski/rfcs/blob/patch-1/text/2294-if-let-guard.md)

### DIFF
--- a/text/2294-if-let-guard.md
+++ b/text/2294-if-let-guard.md
@@ -11,7 +11,7 @@ Allow `if let` guards in `match` expressions.
 # Motivation
 [motivation]: #motivation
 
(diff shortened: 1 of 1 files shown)


foresterre:
It's very well possible that this is [not a typo](https://en.m.wikipedia.org/wiki/If_and_only_if), considering the context. _iff_ stands for _if and only if_ (in logic often represented by a double arrow: _↔_).

chdalski:
@foresterre I didn't know that one, thank you.
And you're right that would actually make sense in the given context.

chdalski:
Seems to be [no typo](https://github.com/rust-lang/rfcs/pull/3587#issuecomment-1991901260)

PR #3590, opened 2024-03-19
Fix "higher kinded" -> "higher ranked" in RFC 3498

traviscross:
In RFC 3498 "Lifetime Capture Rules 2024" we specify that lifetime parameters from `for<..>` binders are captured under the rules.

Currently opaque types in Rust do not support capturing these.  In early drafts of the document, we had called the missing feature "higher ranked lifetime bounds on nested opaque types".  However, we had then noticed that the relevant error message in `rustc` called these "higher kinded" instead:

> error: higher kinded lifetime bounds on nested opaque types are not supported yet

We changed later drafts to follow that language.  But that language was wrong.  These are definitely "higher ranked" lifetime bounds, not "higher kinded" ones, whatever that might mean.  We've now fixed the error message emitted by `rustc` in:

- https://github.com/rust-lang/rust/pull/122100

Correspondingly, let's now fix this error in the RFC.

[Rendered](https://github.com/rust-lang/rfcs/blob/TC/fix-higher-ranked-not-higher-kinded/text/3498-lifetime-capture-rules-2024.md)

cc @oli-obk @compiler-errors

### DIFF
--- a/text/3498-lifetime-capture-rules-2024.md
+++ b/text/3498-lifetime-capture-rules-2024.md
@@ -288,7 +288,7 @@ impl<T> Foo<T> {
 
 ### Capturing lifetimes from `for<..>` binders
 
-Once higher kinded lifetime bounds on nested opaque types are supported in Rust (see [#104288][]), the following code will become legal:
+Once higher ranked lifetime bounds on nested opaque types are supported in Rust (see [#104288][]), the following code will become legal:
 
 ```rust
 trait Trait<'a> {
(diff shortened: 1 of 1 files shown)


fmease (approved):
Thanks for cleaning this up! 🙏 Higher *ranked* is definitely the correct term here.

A higher *kinded* lifetime bound would be something of kind  `K -> Constraint` with `K` in `Type`, `Region` or `Const`, haha, if we assume an outlives-bound. E.g., a higher-kinded `B` where `B<'a>` normalizes to outlives-bound `'a: 'b` with lifetime `'a` variable, `'b` fixed or a higher-kinded `P` where `P<T>`  normalizes to outlives-bound `T: 'a` with `T` variable, `'a` fixed.

PR #3612, opened 2024-04-12
Update 2289-associated-type-bounds.md

slanterns:
1. `2071-impl-trait-type-alias.md` has been moved to another file;
2. Use relative path to make it works better on https://rust-lang.github.io/rfcs/.

[Rendered](https://github.com/slanterns/rfcs/blob/patch-1/text/2289-associated-type-bounds.md)


ehuss (approved):
Thanks!

PR #3613, opened 2024-04-18
chore: remove repetitive words

growfrow:
[Rendered](https://github.com/growfrow/rfcs/blob/master/text/1298-incremental-compilation.md)


ehuss (approved):
Thanks!

PR #3622, opened 2024-05-02
both the RFC file name and link in the file should be updated

RalfJung:
At least I assume this is the intention? It seems strange to update the file but not the link in said file.


RalfJung:
One annoying aspect of this renaming business is that the "Rendered" link that rfcbot adds will then be dangling. Should the README say to update the link?

ehuss (approved):
Yep, thanks!

PR #3632, opened 2024-05-10
[RFC] externally implementable functions

m-ou-se:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3632)*

An alternative to this is https://github.com/rust-lang/rfcs/pull/3635

[Rendered](https://github.com/m-ou-se/rfcs/blob/extern-impl-fn/text/0000-externally-implementable-functions.md)

Tracking:
- https://github.com/rust-lang/rust/issues/125418

### DIFF
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -160,7 +160,7 @@ has been proposed before, which basically does this for *types*. Doing this for 
 - What should the syntax be once we stabilize this?
 - How should this work in dynamic libraries?
 - Should there be a way to specify that implementing the function is unsafe, separately from whether the function itself is unsafe?
-- Should not having an implementation be an error when the function is never called (after dead code elimination)?
+- Should not having an implementation be an error when the function is never called?
 - If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
 - An `extern impl fn` that's marked as `pub(crate)` but is nonetheless pub to *implement* could surprise people. Is there some way we can make this less surprising? Should we require that all `extern impl fn` have `pub` visibility?
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -124,6 +124,8 @@ has been proposed before, which basically does this for *types*. Doing this for 
 
 # Unresolved questions
 
+- Should we provide a mechanism to set an `extern impl fn` using `=` from an existing `fn` value, rather than writing a body? For instance, `impl fn x::y = a::b;`
+- Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
 - What should the syntax be once we stabilize this?
 - How should this work in dynamic libraries?
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -101,7 +101,7 @@ The implementation will be based on the same mechanisms as used today for the `p
 
 The compiler of the root crate will find the implementation of all externally implementable functions and give an error
 if more than one implementation is found for any of them.
-If none are found, and there are any calls to the `extern impl fn` that have not been dead-code eliminated, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
+If none are found, the result is either an error, or—if the `extern impl fn` has a default body—an implementation
 is generated that calls that default body.
 
 # Drawbacks
@@ -130,6 +130,7 @@ has been proposed before, which basically does this for *types*. Doing this for 
 - Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
 - What should the syntax be once we stabilize this?
 - How should this work in dynamic libraries?
+- Should not having an implementation be an error when the function is never called (after dead code elimination)?
 - If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
 
 # Future possibilities
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -79,7 +79,9 @@ impl fn x::b() {
 
 ## Signature
 
-It is an error to have a different signature for the `impl fn` item
+It is an error to have a different signature for the `impl fn` item.
+
+(Whether `#[track_caller]` is used or not is considered part of the signature here.)
 
 ## No impl
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -130,6 +130,7 @@ has been proposed before, which basically does this for *types*. Doing this for 
 # Future possibilities
 
 - Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
+- Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
 - Using this mechanism in the standard library to make more parts overridable. For example:
   - Allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
     (The Rust for Linux project would make use of this.)
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -109,7 +109,8 @@ is generated that calls that default body.
 - It encourages globally defined behaviour.
   - Counterargument: We are already doing this anyway, both inside the standard library (e.g. panic_handler, allocator)
     and outside (e.g. global logger). This just makes it much easier (and safer) to get right.
-- This will invite the addition of many hooks to the standard library to modify existing behavior. While we should consider such possibilities, this RFC does not propose that every piece of standard library behavior should be replaceable; std is not purely a dispatch layer for a set of hooks.
+- This will invite the addition of many hooks to the standard library to modify existing behavior.
+  While we should consider such possibilities, this RFC does not propose that every piece of standard library behavior should be replaceable.
 
 # Rationale and alternatives
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -144,3 +144,4 @@ has been proposed before, which basically does this for *types*. Doing this for 
     (This enables custom testing frameworks to capture output. It is also extremely useful on targets like wasm.)
 - This could possibly be extended to groups of functions in the form of a `trait` that can be globally implemented.
   (E.g. `extern impl AsyncRuntime`, to say that there must be a global implementation of that trait.)
+  - Given an `extern impl Trait` feature, could we provide a compatibility mechanism so that a crate providing an `extern impl fn` can migrate to an `extern impl Trait` in a compatible way, such that crates doing `impl fn` will still be compatible with the new `extern impl Trait`?
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -129,9 +129,11 @@ has been proposed before, which basically does this for *types*. Doing this for 
 
 # Future possibilities
 
-- Doing this for `static` items too.
+- Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
 - Using this mechanism in the standard library to make more parts overridable. For example:
   - Allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
     (The Rust for Linux project would make use of this.)
   - Allowing overriding `write_to_stdout` and `write_to_stderr`.
     (This enables custom testing frameworks to capture output. It is also extremely useful on targets like wasm.)
+- This could possibly be extended to groups of functions in the form of a `trait` that can be globally implemented.
+  (E.g. `extern impl AsyncRuntime`, to say that there must be a global implementation of that trait.)
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -131,6 +131,7 @@ has been proposed before, which basically does this for *types*. Doing this for 
 
 - Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
 - Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
+- We could add a mechanism for arbitrating between multiple provided implementations. For instance, if a crate A depended on B and C, and both B and C provide implementations of an `extern impl fn`, rather than an error, A could provide its own implementation overriding both.
 - Using this mechanism in the standard library to make more parts overridable. For example:
   - Allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
     (The Rust for Linux project would make use of this.)
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -101,7 +101,7 @@ The implementation will be based on the same mechanisms as used today for the `p
 
 The compiler of the root crate will find the implementation of all externally implementable functions and give an error
 if more than one implementation is found for any of them.
-If none are found, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
+If none are found, and there are any calls to the `extern impl fn` that have not been dead-code eliminated, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
 is generated that calls that default body.
 
 # Drawbacks
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -148,6 +148,10 @@ A downside is that it is not possible to allow this "only implementable but not 
 
 An alternative could be to use the same visibility for both implementing an calling, which would simply mean that the function (or an alias to it) will always have to be `pub`.
 
+## Configuration
+
+An `extern impl fn` may have `#[cfg(...)]` attributes applied to it as usual. For instance, a crate may only provide an `extern impl fn` with a given feature flag enabled, and might then use the same feature flag to conditionally provide make other functions depending on that `extern impl fn`. This is a useful pattern for crates that don't want to provide a default implementation but want to avoid producing a compilation error unless the function is needed.
+
 # Prior art
 
 [RFC 2494 "Existential types with external definition"](https://github.com/rust-lang/rfcs/pull/2492)
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -109,6 +109,7 @@ is generated that calls that default body.
 - It encourages globally defined behaviour.
   - Counterargument: We are already doing this anyway, both inside the standard library (e.g. panic_handler, allocator)
     and outside (e.g. global logger). This just makes it much easier (and safer) to get right.
+- This will invite the addition of many hooks to the standard library to modify existing behavior. While we should consider such possibilities, this RFC does not propose that every piece of standard library behavior should be replaceable; std is not purely a dispatch layer for a set of hooks.
 
 # Rationale and alternatives
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -130,6 +130,7 @@ has been proposed before, which basically does this for *types*. Doing this for 
 - Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
 - What should the syntax be once we stabilize this?
 - How should this work in dynamic libraries?
+- If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
 
 # Future possibilities
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -16,7 +16,8 @@ extern impl fn panic_handler(_: &PanicInfo) -> !;
 
 // user:
 
-impl fn core::panic::panic_handler(_: &PanicInfo) -> ! {
+impl fn core::panic::panic_handler(panic_info: &PanicInfo) -> ! {
+    eprintln!("panic: {panic_info:?}");
     loop {}
 }
 ```
@@ -83,13 +84,13 @@ It is an error to have a different signature for the `impl fn` item.
 
 (Whether `#[track_caller]` is used or not is considered part of the signature here.)
 
-## No impl
+## One impl
 
 It is an error to have no `impl fn` item (in any crate) for an `extern impl fn` item without a body.
 
-## Duplicates
-
 It is an error to have multiple `impl fn` items (across all crates) for the same `extern impl fn` item.
+
+Note: This means that adding or removing an `impl fn` item is a semver incompatible change.
 
 ## Visibility
 
@@ -117,10 +118,35 @@ is generated that calls that default body.
 
 # Rationale and alternatives
 
-- The syntax re-uses existing keywords. Alternatively, we could:
+## Syntax
+
+The syntax re-uses existing keywords. Alternatively, we could:
   - Use the `override` reserved keyword.
   - Add a new (contextual) keyword (e.g. `existential fn`).
   - Use an attribute (e.g. `#[extern_impl]`) instead.
+
+## Functions or statics
+
+This RFC only proposes externally implementable *functions*.
+
+An alternative is to only provide externally definable *statics* instead.
+
+That would be equivalent in power: one can store a function pointer in a static, and one can return a reference to a static from a function ([RFC 3635](https://github.com/rust-lang/rfcs/pull/3635)).
+
+(Another alternative, of course is to provide both. See future possibilities.)
+
+## Visibility
+
+There are two kinds of visibilities to be considered for externally implementable functions:
+who can *implement* the function, and who can *call* the function.
+
+Not allowing the function to be implemented by other crates nullifies the functionality, as the entire point of externally implementable functions is that they can be implemented in another crate. This visibility is therefore always (implicitly) "pub".
+
+Allowing a more restricted (that is, not `pub`) visibility for *calling* the function can be useful. For example, today's `#[panic_handler]` can be defined by any crate, but can not be called directly. (Only indirectly through `panic!()` and friends.)
+
+A downside is that it is not possible to allow this "only implementable but not publicly callable" visibility through an alias.
+
+An alternative could be to use the same visibility for both implementing an calling, which would simply mean that the function (or an alias to it) will always have to be `pub`.
 
 # Prior art
 
@@ -133,11 +159,13 @@ has been proposed before, which basically does this for *types*. Doing this for 
 - Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
 - What should the syntax be once we stabilize this?
 - How should this work in dynamic libraries?
+- Should there be a way to specify that implementing the function is unsafe, separately from whether the function itself is unsafe?
 - Should not having an implementation be an error when the function is never called (after dead code elimination)?
 - If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
 
 # Future possibilities
 
+- Adding a syntax to specify an existing function as the impl. E.g. `impl core::panic_handler = my_panic_handler;`
 - Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
 - Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
 - We could add a mechanism for arbitrating between multiple provided implementations. For instance, if a crate A depended on B and C, and both B and C provide implementations of an `extern impl fn`, rather than an error, A could provide its own implementation overriding both.
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -96,6 +96,7 @@ It is an error to have multiple `impl fn` items (across all crates) for the same
 `extern impl fn` items can have a visibility specifier (like `pub`), which determines who can *call* the function (or create pointers to it, etc.).
 
 *Implementing* the function can be done by any crate that can name the item.
+(The `impl fn` item will need to name the item to implement, which could be directly or through an alias/re-export.)
 
 # Implementation
 
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -1,0 +1,137 @@
+- Feature Name: `extern_impl_fn`
+- Start Date: 2024-05-10
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+
+A mechanism for defining a function whose implementation can be defined (or overridden) in another crate.
+
+Example 1:
+
+```rust
+// core::panic:
+
+extern impl fn panic_handler(_: &PanicInfo) -> !;
+
+// user:
+
+impl fn core::panic::panic_handler(_: &PanicInfo) -> ! {
+    loop {}
+}
+```
+
+Example 2:
+
+```rust
+// log crate:
+
+extern impl fn logger() -> Logger {
+    Logger::default()
+}
+
+// user:
+
+impl fn log::logger() -> Logger {
+    Logger::to_stdout().with_colors()
+}
+```
+
+# Motivation
+
+We have several items in the standard library that are overridable/definable by the user crate.
+For example, the (no_std) `panic_handler`, the global allocator for `alloc`, and so on.
+
+Each of those is a special lang item with its own special handling.
+Having a general mechanism simplifies the language and makes this functionality available for other crates, and potentially for more use cases in core/alloc/std.
+
+# Explanation
+
+A function can be defined as "externally implementable" using `extern impl` as follows:
+
+```rust
+// In crate `x`:
+
+// Without a body:
+extern impl fn a();
+
+// With a body:
+extern impl fn b() {
+    println!("default impl");
+}
+```
+
+Another crate can then provide (or override) the implementation of these functions using `impl fn` syntax (using their path) as follows:
+
+```rust
+// In another crate:
+
+impl fn x::a() {
+    println!("my implementation of a");
+}
+
+impl fn x::b() {
+    println!("my implementation of b");
+}
+```
+
+# Details
+
+## Signature
+
+It is an error to have a different signature for the `impl fn` item
+
+## No impl
+
+It is an error to have no `impl fn` item (in any crate) for an `extern impl fn` item without a body.
+
+## Duplicates
+
+It is an error to have multiple `impl fn` items (across all crates) for the same `extern impl fn` item.
+
+## Visibility
+
+`extern impl fn` items can have a visibility specifier (like `pub`), which determines who can *call* the function (or create pointers to it, etc.).
+
+*Implementing* the function can be done by any crate that can name the item.
+
+# Implementation
+
+The implementation will be based on the same mechanisms as used today for the `panic_handler` and `#[global_allocator]` features.
+
+The compiler of the root crate will find the implementation of all externally implementable functions and give an error
+if more than one implementation is found for any of them.
+If none are found, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
+is generated that calls that default body.
+
+# Drawbacks
+
+- It encourages globally defined behaviour.
+  - Counterargument: We are already doing this anyway, both inside the standard library (e.g. panic_handler, allocator)
+    and outside (e.g. global logger). This just makes it much easier (and safer) to get right.
+
+# Rationale and alternatives
+
+- The syntax re-uses existing keywords. Alternatively, we could:
+  - Use the `override` reserved keyword.
+  - Add a new (contextual) keyword (e.g. `existential fn`).
+  - Use an attribute (e.g. `#[extern_impl]`) instead.
+
+# Prior art
+
+[RFC 2494 "Existential types with external definition"](https://github.com/rust-lang/rfcs/pull/2492)
+has been proposed before, which basically does this for *types*. Doing this for functions (as a start) saves a lot of complexity.
+
+# Unresolved questions
+
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+
+# Future possibilities
+
+- Doing this for `static` items too.
+- Using this mechanism in the standard library to make more parts overridable. For example:
+  - Allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
+    (The Rust for Linux project would make use of this.)
+  - Allowing overriding `write_to_stdout` and `write_to_stderr`.
+    (This enables custom testing frameworks to capture output. It is also extremely useful on targets like wasm.)
--- a/text/0000-externally-implementable-functions.md
+++ b/text/0000-externally-implementable-functions.md
@@ -162,6 +162,7 @@ has been proposed before, which basically does this for *types*. Doing this for 
 - Should there be a way to specify that implementing the function is unsafe, separately from whether the function itself is unsafe?
 - Should not having an implementation be an error when the function is never called (after dead code elimination)?
 - If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
+- An `extern impl fn` that's marked as `pub(crate)` but is nonetheless pub to *implement* could surprise people. Is there some way we can make this less surprising? Should we require that all `extern impl fn` have `pub` visibility?
 
 # Future possibilities
 


text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Visibility
+
+`extern impl fn` items can have a visibility specifier (like `pub`), which determines who can *call* the function (or create pointers to it, etc.).
+
+*Implementing* the function can be done by any crate that name the item.
+
+# Implementation
+
+The implementation will be based on the same mechanism as used today for the `panic_handler` and `#[global_allocator]` features.
```
bjorn3, line 100: These are using different mechanisms. panic_handler doesn't allow defaults. global_allocator does, which complicates things a lot as we aren't relying on weak symbols but instead generate a shim right before linking.

jdonszelmann:
How do you deal with the following case:

crate A defines an extern function f.

crate B imports A and implements f.
crate C imports A and implements f.

crate D imports B and C but cannot because both of them implement f and they conflict.

text/0000-externally-implementable-functions.md
m-ou-se: Yes, that's why I'm mentioning both. Which mechanism is used depends on whether you provide a default body or not.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,135 @@
...
+is generated that calls that default body.
+
+# Drawbacks
+
+- It encourages globally defined behaviour.
+  - Counterargument: We are already doing this anyway, both inside the standard library (e.g. panic_handler, allocator)
+    and outside (e.g. global logger). This just makes it much easier (and safer) to get right.
+
+# Rationale and alternatives
+
```
bjorn3, line 114: Should we allow grouping multiple functions together like global_allocator in this RFC? Or should that be left as future potential improvement?

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,135 @@
...
+It is an error to have a different signature for the `impl fn` item
+
+## No impl
+
+It is an error to have no `impl fn` item (in any crate) for an `extern impl fn` item without a body.
+
+## Duplicates
+
+It is an error to have multiple `impl fn` items (across all crates) for the same `extern impl fn` item.
```
epage, line 90: I assume this is excluding the default body

epage, line 90: We should probably call out that this means that adding a `impl fn` function body is a breaking change.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Visibility
+
+`extern impl fn` items can have a visibility specifier (like `pub`), which determines who can *call* the function (or create pointers to it, etc.).
+
+*Implementing* the function can be done by any crate that name the item.
```
programmerjake, line 96:
```suggestion
*Implementing* the function can be done by any crate that can name the item.
```

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,135 @@
...
+
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+
+# Future possibilities
+
+- Doing this for `static` items too.
+- Using this mechanism in the standard library to make more parts overridable.
+  - E.g. allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
+    (The Rust for Linux project would make use of this.)
```
epage, line 135: I'd like to have `stdout` and `stderr` but the return type would severely limit the use.

m-ou-se: Would something like `extern impl fn write_to_stdout(_: &str)` suffice?

m-ou-se:
> I agree conflicting implementations are a compiler error, but if libraries do it then you make some libraries mutually exclusive.

Yes, those would be mutually exclusive. Exactly how today panic-halt, panic-semihosting, and panic-reset are all mutually exclusive.

text/0000-externally-implementable-functions.md
programmerjake: you could work around that with a TAIT:
```rust
pub trait MyFunctions {
    fn fn1() -> String;
    fn fn2(a: String, b: u32);
}

pub type MyFunctionsImpl = impl MyFunctions;

fn f(v: Infallible) -> MyFunctionsImpl {
    my_functions(v)
}

pub extern impl fn my_functions(v: Infallible) -> impl MyFunctions;

pub fn fn3() -> String {
    MyFunctionsImpl::fn1()
}
```

text/0000-externally-implementable-functions.md
epage: Good point!  Whatever signature would be appropriate, yes!

This could help
- `anstream` which provides an alternative `stdout` / `stderr`
- custom test harnesses capturing output

jdonszelmann:
Can libraries define a new default implementation of an extern function? Something like
```rust
// crate A:

extern impl fn log::logger() -> Logger {
    Logger::to_stdout().with_colors()
}

// crate B imports A:
extern impl fn log::logger() -> Logger {
    Logger::to_stderr().with_colors()
}

// crate C imports B:
impl fn log::logger() -> Logger {
    Logger::to_file("log.txt")
}

File: text/0000-externally-implementable-functions.md (hunk 5 above)
Comment (m-ou-se) [reply text/0000-externally-implementable-functions.md]: Added that to the future possibilities.

**Comment by m-ou-se** [comment]
> Can libraries define a new default implementation of an extern function?

No, I don't think we should do that. It'd be hard to define the priority of the multiple defaults.

File: text/0000-externally-implementable-functions.md (hunk 3 above)
Comment (m-ou-se) [reply text/0000-externally-implementable-functions.md:90]: Yup. You can have zero or one `impl fn` item for the `extern impl fn` items that have a body.

File: text/0000-externally-implementable-functions.md (hunk 2 above)
Comment (m-ou-se) [reply text/0000-externally-implementable-functions.md:114]: I think that'd be part of a potential future (more compplicated) RFC, such as https://github.com/rust-lang/rfcs/pull/2492

**Comment by juntyr** [comment]
> How do you deal with the following case?
> 
> crate A defines an extern function f.
> 
> crate B imports A and implements f. crate C imports A and implements f.
> 
> crate D imports B and C but cannot because both of them implement f and they conflict.
> 
> I agree conflicting implementations are a compiler error, but if libraries do it then you make some libraries mutually exclusive.
> 
> Edit: as I noted below, you might make it so B and C can both be imported as long as D also implements f

On the one hand, I would hope that crates that implement externally implementable functions would be minimal, i.e. only have this implementation and nothing else, so that including both in the crate graph would always be a clear and understandable error (and not a mistake because you'd want functionality from both).

On the other hand, the example of logging shows that there are clear cases where you might want to combine different loggers into some super-logger that is specific for your use case. Logging implementation crates could then either ship a companion crate that only implements the function, or could have a non-default feature to enable the implementation.

**Comment by jdonszelmann** [comment]
Simply feature gating these implementations solves a lot of the problems I think.

File: text/0000-externally-implementable-functions.md [hunk 6]
```diff
@@ -0,0 +1,137 @@
...
+has been proposed before, which basically does this for *types*. Doing this for functions (as a start) saves a lot of complexity.
+
+# Unresolved questions
+
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+
+# Future possibilities
+
+- Doing this for `static` items too.
```
kennytm, line 132: i think that any item that can be placed into an `extern "Rust" {}` block should be allowed :thinking: 

supporting `static` is also closer to the current `#[global_allocator]` which is applied to a `static`.

Diggsey:
> If none are found, the result is either an error, or, if the extern impl fn has a default body, an implementation is generated that calls that default body.

I think the RFC needs to clarify how this works with different crate types. Presumably this check is not required when building a library crate (or else the feature would be useless). What about when building a `dylib` or `cdylib`?

Also, I think it would be good if the root (binary) crate could resolve the issue with multiple implementations by providing its own implementation.

m-ou-se:
> I think it would be good if the root (binary) crate could resolve the issue with multiple implementations by providing its own implementation.

That's a possible extension. But that's not what we currently have for the panic handler or global allocator, right?

text/0000-externally-implementable-functions.md
m-ou-se: Added a note on that.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+- It encourages globally defined behaviour.
+  - Counterargument: We are already doing this anyway, both inside the standard library (e.g. panic_handler, allocator)
+    and outside (e.g. global logger). This just makes it much easier (and safer) to get right.
+
+# Rationale and alternatives
+
+- The syntax re-uses existing keywords. Alternatively, we could:
+  - Use the `override` reserved keyword.
+  - Add a new (contextual) keyword (e.g. `existential fn`).
+  - Use an attribute (e.g. `#[extern_impl]`) instead.
```
fbstj, line 118: should the following alternative be mentioned / discussed:

- multiple impl's are allowed
- the root crate must import the impl they want
- the normal-default impl is imported via the prelude

also just had the thought: does `use crateA::different_name as panic_handler` work similar to how (I believe) it works for `main`?

burdges:
It's unlikely the root binary crate should provide these or micro-manage them too much.

It's most likely library crate features control these, which enables many options.  If you do want micro-managment by the root binary crate then these could themselves by micro-crates too. 

I'd think `pub const` and `pub static` would be the most likely future extensions here, but `pub fn` groups, or `pub mod`, could probably wait for more existential types discussions ala https://github.com/rust-lang/rfcs/pull/2492

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+impl fn x::b() {
+    println!("my implementation of b");
+}
+```
+
+# Details
+
+## Signature
+
+It is an error to have a different signature for the `impl fn` item
```
BoxyUwU, line 82: Should this use the same rules as trait Def Vs impls where the implementor can write a more general signature?

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+if more than one implementation is found for any of them.
+If none are found, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
+is generated that calls that default body.
+
+# Drawbacks
+
+- It encourages globally defined behaviour.
+  - Counterargument: We are already doing this anyway, both inside the standard library (e.g. panic_handler, allocator)
+    and outside (e.g. global logger). This just makes it much easier (and safer) to get right.
+
```
joshtriplett, line 112:
```suggestion
- This will invite the addition of many hooks to the standard library to modify existing behavior. While we should consider such possibilities, this RFC does not propose that every piece of standard library behavior should be replaceable; std is not purely a dispatch layer for a set of hooks.

```

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+*Implementing* the function can be done by any crate that can name the item.
+
+# Implementation
+
+The implementation will be based on the same mechanisms as used today for the `panic_handler` and `#[global_allocator]` features.
+
+The compiler of the root crate will find the implementation of all externally implementable functions and give an error
+if more than one implementation is found for any of them.
+If none are found, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
+is generated that calls that default body.
```
joshtriplett, line 105:
```suggestion
If none are found, and there are any calls to the `extern impl fn` that have not been dead-code eliminated, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
is generated that calls that default body.
```

It should be possible to leave an `extern impl fn` unimplemented iff you never call it.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Duplicates
+
+It is an error to have multiple `impl fn` items (across all crates) for the same `extern impl fn` item.
+
+## Visibility
+
+`extern impl fn` items can have a visibility specifier (like `pub`), which determines who can *call* the function (or create pointers to it, etc.).
+
+*Implementing* the function can be done by any crate that can name the item.
```
joshtriplett, line 96: Could you clarify whether this is intended to also tie into visibility? For instance, a `pub(crate) extern impl fn` can only have an implementation provided by the crate, right?

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Unresolved questions
+
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+
+# Future possibilities
+
+- Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
+- Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
```
joshtriplett, line 133:
```suggestion
- Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
- We could add a mechanism for arbitrating between multiple provided implementations. For instance, if a crate A depended on B and C, and both B and C provide implementations of an `extern impl fn`, rather than an error, A could provide its own implementation overriding both.
```

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,140 @@
...
+  - Use an attribute (e.g. `#[extern_impl]`) instead.
+
+# Prior art
+
+[RFC 2494 "Existential types with external definition"](https://github.com/rust-lang/rfcs/pull/2492)
+has been proposed before, which basically does this for *types*. Doing this for functions (as a start) saves a lot of complexity.
+
+# Unresolved questions
+
+- What should the syntax be once we stabilize this?
```
joshtriplett, line 127:
```suggestion
- Should we provide a mechanism to set an `extern impl fn` using `=` from an existing `fn` value, rather than writing a body? For instance, `impl fn x::y = a::b;`
- What should the syntax be once we stabilize this?
```

joshtriplett:
I think this is a great idea, and I think we should try it as an experiment.

@rfcbot merge

@rfcbot concern github-suggestions

text/0000-externally-implementable-functions.md
m-ou-se: The idea is that if you do:

```rust
pub mod a {
    pub(crate) extern impl fn x();
}
```

The, other crates can provide an implementation (because `a` is `pub`, allowing them to name `a::x`), but they cannot call it (because the function is not public itself).

text/0000-externally-implementable-functions.md
m-ou-se: Probably, yes.

text/0000-externally-implementable-functions.md
joshtriplett: @m-ou-se That seems potentially confusing. Is there some way to set the visibility of being able to *implement* it? Is there value in being able to do so?

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,145 @@
...
+
+[RFC 2494 "Existential types with external definition"](https://github.com/rust-lang/rfcs/pull/2492)
+has been proposed before, which basically does this for *types*. Doing this for functions (as a start) saves a lot of complexity.
+
+# Unresolved questions
+
+- Should we provide a mechanism to set an `extern impl fn` using `=` from an existing `fn` value, rather than writing a body? For instance, `impl fn x::y = a::b;`
+- Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
```
joshtriplett, line 132:
```suggestion
- How should this work in dynamic libraries?
- If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
```

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,145 @@
...
+- Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
+- Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
+- We could add a mechanism for arbitrating between multiple provided implementations. For instance, if a crate A depended on B and C, and both B and C provide implementations of an `extern impl fn`, rather than an error, A could provide its own implementation overriding both.
+- Using this mechanism in the standard library to make more parts overridable. For example:
+  - Allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
+    (The Rust for Linux project would make use of this.)
+  - Allowing overriding `write_to_stdout` and `write_to_stderr`.
+    (This enables custom testing frameworks to capture output. It is also extremely useful on targets like wasm.)
+- This could possibly be extended to groups of functions in the form of a `trait` that can be globally implemented.
+  (E.g. `extern impl AsyncRuntime`, to say that there must be a global implementation of that trait.)
```
joshtriplett, line 145:
```suggestion
  (E.g. `extern impl AsyncRuntime`, to say that there must be a global implementation of that trait.)
  - Given an `extern impl Trait` feature, could we provide a compatibility mechanism so that a crate providing an `extern impl fn` can migrate to an `extern impl Trait` in a compatible way, such that crates doing `impl fn` will still be compatible with the new `extern impl Trait`?
```

text/0000-externally-implementable-functions.md
m-ou-se: No, because the point of this feature is allowing other crates to implement it. So unless you want to propose a kind of visibility that includes some crates but not others, that just implies full public.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,147 @@
...
+}
+
+impl fn x::b() {
+    println!("my implementation of b");
+}
+```
+
+# Details
+
+## Signature
```
RalfJung, line 80: I think you need to disallow `#[track_caller]` (or at least enforce that it is coherent on both the declaration and definition side), since track_caller is effectively adding another function argument.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,147 @@
...
+// Without a body:
+extern impl fn a();
+
+// With a body:
+extern impl fn b() {
+    println!("default impl");
+}
+```
+
+Another crate can then provide (or override) the implementation of these functions using `impl fn` syntax (using their path) as follows:
```
RalfJung, line 64: If the `extern impl fn` is re-exported, can `impl fn` refer to the re-export?

text/0000-externally-implementable-functions.md
RalfJung: So if the `extern impl fn` is in a private module, it is impossible to implement?

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,147 @@
...
+impl fn x::a() {
+    println!("my implementation of a");
+}
+
+impl fn x::b() {
+    println!("my implementation of b");
+}
+```
+
+# Details
```
RalfJung, line 78: Is there a particular reason you diverged from the standard RFC template?

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Should we provide a mechanism to set an `extern impl fn` using `=` from an existing `fn` value, rather than writing a body? For instance, `impl fn x::y = a::b;`
+- Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+- If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
+
+# Future possibilities
+
+- Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
```
RalfJung, line 137: FWIW, such a static of function pointer type would basically subsume the RFC.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,147 @@
...
+*Implementing* the function can be done by any crate that can name the item.
+
+# Implementation
+
+The implementation will be based on the same mechanisms as used today for the `panic_handler` and `#[global_allocator]` features.
+
+The compiler of the root crate will find the implementation of all externally implementable functions and give an error
+if more than one implementation is found for any of them.
+If none are found, and there are any calls to the `extern impl fn` that have not been dead-code eliminated, the result is either an error, or, if the `extern impl fn` has a default body, an implementation
+is generated that calls that default body.
```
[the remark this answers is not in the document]
bjorn3: I would expect rustc to unconditionally check that it exist when linking, just like for panic_handler.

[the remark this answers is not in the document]
m-ou-se: :+1: Moved the part about dead code elimination to the unresolved questions.

text/0000-externally-implementable-functions.md
m-ou-se: This worked better. Otherwise I'l' just end up repeating the same thing twice. 🤷‍♀️

(I use the template as a useful guide, not as a fixed form to match. Most of my RFCs don't use the exact template structure.)

text/0000-externally-implementable-functions.md
m-ou-se: Yeah, `#[track_caller]` should be considered part of the signature for this purpose.

text/0000-externally-implementable-functions.md
Mark-Simulacrum: I would strongly prefer avoiding paths being allowed in the actual function declaration, because it breaks "fn $name" search from being essentially 100% reliable... that is a very useful property in Rust today.

It doesn't seem like we lose much by doing that since you can always use it first?

kennytm:
@burdges 

> I'd think pub const and pub static would be the most likely future extensions here.

no no no please don't extend this RFC to `const`. For `static` and `fn` one could implement the feature via the linker. But for `const` it can be used in compile-time constructs like `[T; N]` and `Array<T, N>` meaning the compiler has to consider multiple crates at the same time to fill in the `extern const` value. It makes the compilation model extremely complex without much conceivable benefits compared to `extern static`.

text/0000-externally-implementable-functions.md
kennytm: @m-ou-se 

> The idea is that if you do:
> 
> ```rust
> pub mod a {
>     pub(crate) extern impl fn x();
> }
> ```
>
> The, other crates can provide an implementation (because `a` is `pub`, allowing them to name `a::x`), but they cannot call it (because the function is not public itself).

Yeah this is very confusing. Could you make it respect the normal privacy rule and reuse e.g. #3323 to explicitly deny call/ref permission from dependencies?

```rust
pub mod a {
    pub restrict_use(crate) extern impl fn x();
}
```

text/0000-externally-implementable-functions.md
RalfJung: We should have common guidance on how to use the template. I've (sometimes painfully) shoved my RFCs into the structure as I thought it was mandatory (and I was given that impression when I used an outdated version of the template) and now I see other people ignore the structure -- that's frustrating.

text/0000-externally-implementable-functions.md
RalfJung: > So if the extern impl fn is in a private module, it is impossible to implement?

It could be re-exported, but that reexport then presumably allows *both* calling and implementing the item. The magic behavior where these two permissions are different can only be obtain via the original definition of the item, IIUC.

deltragon:
I think this may just be a future possibility, but I would assume that some uses would need the implementation to uphold a special contract - so it might make sense to have unsafe-to-impl externally implemented functions?
(Being a future possibility, this could also be added later, meaning that for now the implementation will always have no restrictions)

Lokathor:
Adding restrictions is generally seen as a breaking change, so we wouldn't want to stabilize specific extern impl fns without those restrictions, so having the ability to make restrictions would effectively need to be part of the base RFC concept.

RalfJung:
The existing panic_handler has no such restrictions. Same for the unstable alloc_error_handler.

For global_alloc it's a bit more tricky as it's a bunch of functions and they're unsafe to call and on a trait that's unsafe to implement...

text/0000-externally-implementable-functions.md
RalfJung: Possibly global_alloc could at least use the same internal mechanism, even if it's not visible to the user?
```rust
#[global_alloc]
static ALLOC: MyAlloc = ...;
```
could expand do something like
```rust
static ALLOC: MyAlloc = ...;

impl fn alloc::alloc::alloc(layout: Layout) -> *mut u8 {
  ALLOC.alloc(layout)
}
impl fn alloc::alloc::dealloc(layout: Layout) -> *mut u8 {
  ALLOC.dealloc(layout)
}
// ...
```

Then codegen and Miri would only have to support one such mechanism. :)

comex:
In the alternatives section, you might want to mention the existing ability to use `#[no_mangle]` in the defining crate, and `extern {}` blocks in the calling crate.

The downsides of that approach I can think of are:
- Global namespace of function names
- Forced to be `unsafe`
- Doesn't work with dylibs (without platform-specific hacks)
  - But note that global_allocator [currently doesn't work with dylibs either](https://github.com/rust-lang/rust/issues/114518#issuecomment-1666587664).
- Cargo doesn't know about it
  - But would this feature get any special Cargo integration?  It would be nice to be able to say "this crate is a dependency but only at link time", improving build parallelism.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,151 @@
...
+- Using this mechanism in the standard library to make more parts overridable. For example:
+  - Allowing custom implementations of `panic_out_of_bounds` and `panic_overflowing_add`, etc.
+    (The Rust for Linux project would make use of this.)
+  - Allowing overriding `write_to_stdout` and `write_to_stderr`.
+    (This enables custom testing frameworks to capture output. It is also extremely useful on targets like wasm.)
+- This could possibly be extended to groups of functions in the form of a `trait` that can be globally implemented.
+  (E.g. `extern impl AsyncRuntime`, to say that there must be a global implementation of that trait.)
+  - Given an `extern impl Trait` feature, could we provide a compatibility mechanism so that a crate providing an `extern impl fn` can migrate to an `extern impl Trait` in a compatible way, such that crates doing `impl fn` will still be compatible with the new `extern impl Trait`?
```
Nemo157, line 151: One approach to compatibility would be to start with this syntax already, and just artificially restrict the trait used to only have `Self`-less associated functions, I believe that is isomorphic to this proposal of defining global functions, and could then just be extended to support more trait features in the future (eventually I assume any object-safe trait could be used).

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,151 @@
...
+[RFC 2494 "Existential types with external definition"](https://github.com/rust-lang/rfcs/pull/2492)
+has been proposed before, which basically does this for *types*. Doing this for functions (as a start) saves a lot of complexity.
+
+# Unresolved questions
+
+- Should we provide a mechanism to set an `extern impl fn` using `=` from an existing `fn` value, rather than writing a body? For instance, `impl fn x::y = a::b;`
+- Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+- Should not having an implementation be an error when the function is never called (after dead code elimination)?
```
joshtriplett, line 136: Leaving aside the issues with optimization-dependence, there is a *lot* of value in being able to figure out cases like the following:

- Crate `a` provides `extern impl fn hook()`, with no default.
- Crate `a` provides various APIs, only one of which, `uses_hook()`, ever calls `hook()`.
- Binary crate `b` depends on `a`.
- `b` does not provide an impl of `hook()`.
- `b` never calls `uses_hook()`.

I think this case should link and work. I see many use cases for "if you want to call X you need to provide an impl of Y, but you don't have to provide an impl of Y if you never call X".

To be clear, the case I'm concerned about does not depend on optimization-based dead-code elimination (e.g. `if false { uses_hook(); }`). The case I'm concerned about is where it's *statically* obvious to the compiler that nothing calls `uses_hook()` (either because nothing calls it directly, or because anything calling it directly is itself never called).

text/0000-externally-implementable-functions.md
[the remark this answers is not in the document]
joshtriplett: I posted a note there about the case that motivated this. I agree that we shouldn't have optimization-dependent cases that require DCE. But I think we shouldn't error if we can *statically* determine that the function is never called without requiring the optimizer/DCE at all.

joshtriplett:
@rfcbot resolved github-suggestions

joshtriplett:
@rfcbot concern statically-unused-extern-impl-fn

text/0000-externally-implementable-functions.md
juntyr: I fear that by the point at which the knowledge (that the hook is never used) becomes available, we could already be past some early error stages, making this a very late error to discover. In any case, crate a could always feature-gate the hook and all relevant methods to make such a case explicit.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,137 @@
...
+```rust
+// core::panic:
+
+extern impl fn panic_handler(_: &PanicInfo) -> !;
+
+// user:
+
+impl fn core::panic::panic_handler(_: &PanicInfo) -> ! {
+    loop {}
+}
```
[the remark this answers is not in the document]
m-ou-se: Given that users aren't allowed to have an impl block for `fn` types anyway, and that the syntax is unambiguous regardless, I'm not too worried about this.

But I also don't care that much about the syntax. We can consider other syntaxes before stabilization of course.

text/0000-externally-implementable-functions.md
bjorn3: Supporting this would require delaying codegen of `uses_hook()` to all call sites. Otherwise without `--gc-sections` (because of `-Clink-dead-code=yes` or because of a linker which doesn't support `--gc-sections`) the linker will complain about the missing symbol definition.

It also leaks the internal implementation details of functions unless the function signature is explicitly required to mention all externally implementable functions that are transitively used.

m-ou-se:
I posted an alternative that does this for `static` instead of `fn`: https://github.com/rust-lang/rfcs/pull/3635

They're pretty much equivalent in power. Doing it for `fn` is a bit more ergonomic in quite a few cases, but doing it for `static` is maybe a bit more consistent. I don't have much of a preference, but we should consider both.

text/0000-externally-implementable-functions.md
m-ou-se: @joshtriplett How do you feel about the fact that one needs to provide a `#[panic_handler]` in a no_std program even if the entire crate is just `pub fn start() {}` or something like that, where nothing can possibly call the panic handler?

I agree there could be value in only requiring it when necessary, but I think there is also a lot of value in keeping this feature consistent with how things like `#[panic_handler]` work.

joshtriplett: @m-ou-se I can appreciate that it's hard for the compiler to statically determine "nothing ever panics and thus nothing ever uses the panic handler". The panic handler in particular seems potentially difficult in that way.

It *seems* like it should be easier to determine "nothing ever calls this function".

@bjorn3
> Supporting this would require delaying codegen of `uses_hook()` to all call sites. Otherwise without `--gc-sections` (because of `-Clink-dead-code=yes` or because of a linker which doesn't support `--gc-sections`) the linker will complain about the missing symbol definition.

I think it'd be *perfectly fine* if this errored out if you use `-Clink-dead-code=yes`, on the basis that the dead code you're including has calls to the externally implementable function. Does that make it easier?

-----

That said, I can see another possible way to handle this, as well. If a crate wants to make a set of functionality only available given a specific externally implementable function, it can put *both* the functionality and the externally implementable function behind the same feature flag, and tell users to turn that feature on and then provide the function. That also addresses @bjorn3's point about exposing implementation details, since the feature flag would define the set of things that need the externally implementable function.

Given that, I'll submit a suggestion documenting this, and move this case to future work.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Unresolved questions
+
+- Should we provide a mechanism to set an `extern impl fn` using `=` from an existing `fn` value, rather than writing a body? For instance, `impl fn x::y = a::b;`
+- Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
+- What should the syntax be once we stabilize this?
+- How should this work in dynamic libraries?
+- Should there be a way to specify that implementing the function is unsafe, separately from whether the function itself is unsafe?
+- Should not having an implementation be an error when the function is never called (after dead code elimination)?
+- If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
```
joshtriplett, line 164:
```suggestion
- If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
- An `extern impl fn` that's marked as `pub(crate)` but is nonetheless pub to *implement* could surprise people. Is there some way we can make this less surprising?
```

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,179 @@
...
+There are two kinds of visibilities to be considered for externally implementable functions:
+who can *implement* the function, and who can *call* the function.
+
+Not allowing the function to be implemented by other crates nullifies the functionality, as the entire point of externally implementable functions is that they can be implemented in another crate. This visibility is therefore always (implicitly) "pub".
+
+Allowing a more restricted (that is, not `pub`) visibility for *calling* the function can be useful. For example, today's `#[panic_handler]` can be defined by any crate, but can not be called directly. (Only indirectly through `panic!()` and friends.)
+
+A downside is that it is not possible to allow this "only implementable but not publicly callable" visibility through an alias.
+
+An alternative could be to use the same visibility for both implementing an calling, which would simply mean that the function (or an alias to it) will always have to be `pub`.
```
joshtriplett, line 149:
```suggestion
An alternative could be to use the same visibility for both implementing an calling, which would simply mean that the function (or an alias to it) will always have to be `pub`. Given that the API of the function is already a stability guarantee of the crate in order for third parties to implement it, this might not be an issue.
```

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,179 @@
...
+
```
joshtriplett, line 150:
```suggestion

## Configuration

An `extern impl fn` may have a `#[cfg(...)]` applied to it as usual. For instance, a crate may only provide an `extern impl fn` with a given feature flag enabled, and might then use the same feature flag to conditionally provide make other functions depending on that `extern impl fn`. This is a useful pattern for crates that don't want to provide a default implementation but want to avoid producing a compilation error unless the function is needed.

## Stability

`extern impl fn` may have stability markers (e.g. `#[unstable]`) applied to it, in which case implementing the function requires opting into the unstable feature (and a compiler that permits opting into unstable features).

```

text/0000-externally-implementable-functions.md
m-ou-se: The issue is not stability. The issue is that the function might not be designed to be called from anywhere.

text/0000-externally-implementable-functions.md
m-ou-se: Not sure if this is a useful addition to the RFC.

- cfg() doesn't only apply to `extern impl fn` items, also to `impl fn` items. But since it just works the same for all items, this doesn't really seem to add anything new.

- Stability attributes are unstable, so how that works is more of an implementation detail than part of the language.

m-ou-se:
Process question for @rust-lang/lang: should this be an experimental feature gate instead of an RFC?

joshtriplett:
@m-ou-se Definitely not "instead of", but possibly "in addition to". This feature absolutely needs an RFC in order to be a future stable feature, but it could *also* be an experimental feature gate in order to implement it without waiting for the RFC to be merged.

That said, I really hope we accept and merge this soon, and I'm hoping we get to it in the next lang meeting. I would like to prioritize it.

text/0000-externally-implementable-functions.md
joshtriplett: @m-ou-se Sure, but that seems like a reason to *possibly* have non-`pub` functions in the future, not a reason to require them immediately as part of the initial implementation.

text/0000-externally-implementable-functions.md
joshtriplett: @m-ou-se I realize that configuration and stability attributes apply to just about everything. But it seems particularly relevant to observe that configuration attributes apply, because that serves as a viable substitute in lieu of handling a missing implementation when an externally implementable function is never called. I wrote this patch as an attempt to resolve the "statically-unused-extern-impl-fn" concern.

(That purpose is served entirely by the "Configuration" section, and I agree that the "Stability" section is less important since it's documenting the interaction with an unstable mechanism.)

m-ou-se:
> That screams "use traits for this"

Yeah, I agree that would be a good fit, but then we basically end up with https://github.com/rust-lang/rfcs/pull/2492, which was [not accepted](https://github.com/rust-lang/rfcs/pull/2492#issuecomment-486996239) at the time because it involved too much complexity.

I'd be happy if we could pick that route and make that all work. This RFC is just an attempt to do something much more basic to start with, since a much more complicated change like https://github.com/rust-lang/rfcs/pull/2492 seems unlikely to work out any time soon.

text/0000-externally-implementable-functions.md
m-ou-se: Yeah I think it depends a bit on the situation how useful the template is. For complex language features, the template works well I think. But for smaller/simpler features or subtle changes it often doesn't work. For example, in [RFC 3007](https://rust-lang.github.io/rfcs/3007-panic-plan.html) I used a problems+solutions structure, which worked much better for a set of subtle changes that fix a set of subtle problems.

The risk of not following the RFC template is missing important parts, or producing a hard to follow document with unclear structure, or producing something that's not easily digestible by all target audiences. It's definitely a useful template.

But in the end the point of an RFC is to just clearly lay out all the relevant information. So I mostly see the template as a guide to cover all the information that should be in the document, not as a strict form that needs to be filled in in one exact way. (Especially if that'd just lead to a less clear presentation of the information, as that doesn't benefit anyone.)

Perhaps we should add a note to the template. Or provide multiple templates so one can pick whichever fits best.

burdges:
As written, linkers can do this resolution, no?  Isn't that enough reason this should still exist even if some trait based scheme emerges in 5 years or whatever?

jdonszelmann:
Just like [global registration](https://github.com/rust-lang/rust/issues/125119) I don't think you want to implement this feature through the linker, at the very least initially. Personally, I'd be worried about the errors that are generated and which are hard to bring to the same standard as other rust compiler errors. I also like [this comment](https://github.com/rust-lang/rfcs/pull/3632#issuecomment-2105273291) about it.

bjorn3:
> I'd be worried about the errors that are generated and which are hard to bring to the same standard as other rust compiler errors.

My suggestion is to have the check for a downstream crate implementing it be in rustc, but the doing the actual tying up with the linker, like is currently done for `#[panic_handler]` and `#[global_allocator]` (for the latter the default is handled using a compiler generated shim, but when you actually use `#[global_allocator]` no shim is used at all.). In the end some form of tying up by the linker is necessary anyway unless you want to have a global constructor which at runtime sets a static to point to the external implementation of the function.

text/0000-externally-implementable-functions.md
m-ou-se: Alright. Added the note on configuration.

joshtriplett:
@rfcbot resolve statically-unused-extern-impl-fn

Resolved by the use of `cfg`.

petrochenkov:
@m-ou-se 
> > That screams "use traits for this"
> 
> Yeah, I agree that would be a good fit, but then we basically end up with #2492, which was [not accepted](https://github.com/rust-lang/rfcs/pull/2492#issuecomment-486996239) at the time because it involved too much complexity.
> 
> I'd be happy if we could pick that route and make that all work. This RFC is just an attempt to do something much more basic to start with, since a much more complicated change like #2492 seems unlikely to work out any time soon.

I don't see why impls used for this feature cannot be restricted to prohibit, for example, associated types or generics, if they complicate the minimal design.

joshtriplett:
I'd be happy to see an implementation using traits that is *initially* restricted in what can appear in the trait, if that makes it simple enough to ship.

I do hope eventually we can ship a version that allows, for instance, associated constants, and that we can use those associated constants in generic bounds on functions in the standard library. (For instance, conditionally providing `impl From<u64> for usize`.) However, I know that'd be more complex, and I think we should ship something without that support first if we can do that more easily.

m-ou-se:
> I don't see why traits/impls used for this feature cannot be restricted to prohibit, for example, associated types or generics, if they complicate the minimal design.

So you're proposing basically https://github.com/rust-lang/rfcs/pull/2492 with some restrictions to make the implementation much simpler?

If you have time to write down a more concrete proposal (not necessariliy an RFC, but some clear examples or something), that would be valuable.

petrochenkov:
I don't have a concrete proposal, just a general suggestion to resyntax this

```rust
// log crate:

extern impl fn logger() -> Logger {
    Logger::default()
}

// user:

impl fn log::logger() -> Logger {
    Logger::to_stdout().with_colors()
}
```

into something like

```rust

// log crate:

#[some_extern_impl_attribute]
trait LoggerInterface {
  fn create_logger() -> Logger {
    Logger::default() // either default body for the default
  }
}

// or a separate impl for the default
struct DefaultLoggerCreator;
#[maybe_some_other_extern_impl_attribute_if_really_necessary]
impl LoggerInterface for DefaultLoggerCreator {
  fn create_logger() -> Logger {
      Logger::default()
  }
}

// user:

struct MyLoggerCreator;
#[maybe_some_third_extern_impl_attribute_if_really_necessary]
impl log::LoggerInterface for MyLoggerCreator {
  fn create_logger() -> Logger {
      Logger::to_stdout().with_colors()
  }
}
```

traviscross:
We discussed this in the lang meeting today.  We developed a consensus that this is addressing an important problem and one that we would like to solve.

In the meeting, there were various alterations and alternatives put forward, including by @Amanieu and @tmandry.  We also wanted to cross-check this against the recenty-accepted RFC:

- https://github.com/rust-lang/rfcs/pull/3484

That RFC adopts a conceptual separation between the unsafety of *declaring* an extern block (and verifying that the signatures within are correct) and the unsafety of *calling* (or otherwise using) an extern item that may have other invariants that may need to be upheld.  We just need to check that whatever we do here is consistent with that conceptually (maybe it already is).

While we allow some time for these things, in the interest of not blocking experimentation, we've decided to approve this to go forward as a lang experiment under our process for that.  @joshtriplett has offered to be the liaison.

We've opened a tracking issue for this experiment here:

- https://github.com/rust-lang/rust/issues/125418

traviscross:
@rustbot labels -I-lang-nominated

Since the next step here is to discuss the full set of options, let's nominate the tracking issue in place of this RFC.

Amanieu:
I wrote up my alternative proposal in #3645, which is heavily based on this one. It keeps the basic idea of just having functions that defined downstream and resolved by the linker, but changes the syntax to look more like traits. This works better for things like the global allocator which consists of multiple functions and allows safety to be defined separately on the trait (unsafe to implement) and its functions (unsafe to call).

tmandry:
### Move extern impls to blocks

My proposal is to move both the declarations and implementations into blocks. That would let us differentiate between functions that are unsafe to implement and functions that are unsafe to call. It would look something like this:[^extern-mod]

```rust
// alloc::global:

extern unsafe impl {
    fn allocate(layout: Layout) -> Result<NonNull<[u8]>, AllocError>;
    unsafe fn deallocate(ptr: NonNull<u8>, layout: Layout);
}

// user:

// Note the use of a path here – already allowed, except it's a module not a type!
// This means we won't have to add a new case to the syntax, and keeps things nicely grouped together.
unsafe impl alloc::global {
    fn allocate(layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
        todo!()
    }
    
    unsafe fn deallocate(ptr: NonNull<u8>, layout: Layout) {
        todo!()
    }
}
```

[^extern-mod]: As I prepare to post this I see that it's quite close to a [proposal for `extern mod`](https://github.com/rust-lang/rfcs/pull/3645#issuecomment-2125878668) in the other thread, though it isn't clear where to hang the `unsafe` for impls in that proposal.

From here we can, optionally, do the following:

### Unify declarations with `extern "Rust" {}` blocks

We can make `extern "Rust" {}` work exactly like `extern impl` above. This includes preserving namespacing of function names, unlike the `"C"` ABI.

If we do this we should transition the default ABI inside `extern {}` blocks to be `"Rust"`. This can be done over an edition.

Also as a delta to #3484, `extern "Rust" {}` blocks would not need `unsafe`. In fact, it wouldn't make sense to mark them as such, because the compiler checks the signatures for you. We would not want to use `unsafe extern` to mean "unsafe to override"; instead, we should keep the `extern unsafe impl {}` syntax.

The first proposal is forward-compatible with this one.

Jules-Bertholet:
> Unify declarations with `extern "Rust" {}` blocks

No, just because you happen to be using the Rust ABI for some `extern` declarations, you should not thereby be forced to adopt the new compiler checks? (For example, maybe you are linking to an object file that was compiled with the same version of `rustc`, but is otherwise completely opaque to you.

Also, this "checked `extern`" feature would IIUC be able to support `fn`s with generic type and const parameters, which is not something today's `extern` blocks can support; which suggests to my mind that these are distinct and non-unifiable features, despite their similarity.

tmandry:
For the record I don't think `extern "Rust" {}` is useful for anything today (possibly due to a bug); it seems to rename the function according to the C ABI, but a matching definition does not. All my attempts to use it resulted in linker errors and the only results on Github are uses of the cxx crate.

> For example, maybe you are linking to an object file that was compiled with the same version of rustc, but is otherwise completely opaque to you.

Given what I said above, that's a new feature. I would argue that the default Rust ABI should check externs and require an rmeta file at minimum. But we could add unsafe "unchecked" Rust externs in the future that work across, e.g., staticlib/cdylib boundaries.

> Also, this "checked `extern`" feature would IIUC be able to support `fn`s with generic type and const parameters, which is not something today's `extern` blocks can support; which suggests to my mind that these are distinct and non-unifiable features, despite their similarity.

I can see your argument here. Ultimately it is just a question of how much we decide to semantically group the features, despite various differences in their capabilities. It would be helpful to imagine how we might represent stable ABI boundaries and see how these would fit in (cc @Amanieu).

As I said in my comment above though, it would be fine to defer the question of unifying with `extern "Rust"` until later, since the first part of what I propose is forward compatible with it.

tmandry:
For the record I don't think `extern "Rust" {}` is useful for anything today (possibly due to a bug); it seems to rename the function according to the C ABI, but a matching definition does not. All my attempts to use it resulted in linker errors and the only results on Github are uses of the cxx crate.

> For example, maybe you are linking to an object file that was compiled with the same version of rustc, but is otherwise completely opaque to you.

Given what I said above, that's a new feature. I would argue that the default Rust ABI should check externs and require an rmeta file at minimum. But we could add unsafe "unchecked" Rust externs in the future that work across, e.g., staticlib/cdylib boundaries.

> Also, this "checked `extern`" feature would IIUC be able to support `fn`s with generic type and const parameters, which is not something today's `extern` blocks can support; which suggests to my mind that these are distinct and non-unifiable features, despite their similarity.

I can see your argument here. Ultimately it is just a question of how much we decide to semantically group the features, despite various differences in their capabilities. It would be helpful to imagine how we might represent stable ABI boundaries and see how these would fit in (cc @Amanieu).

As I said in my comment above though, it would be fine to defer the question of unifying with `extern "Rust"` until later, since the first part of what I propose is forward compatible with it.

Jules-Bertholet:
> For the record I don't think `extern "Rust" {}` is useful for anything today (possibly due to a bug); it seems to rename the function according to the C ABI, but a matching definition does not.

Even with `#[no_mangle]`? That sounds like a bug

programmerjake:
> Even with `#[no_mangle]`? That sounds like a bug

I think he's complaining that `#[no_mangle]` is required rather than that `#[no_mangle]` doesn't work..,

lolbinarycat:
> to conditionally provide make other functions 

i assume this is a typo?

also, if i'm not mistaken, this same functionality can be provided with no_mangle and `extern "Rust"`, cam it not?

programmerjake:
> also, if i'm not mistaken, this same functionality can be provided with no_mangle and `extern "Rust"`, cam it not?

if there's a default body: only if your platform happens to support weak symbols, and even then it's `unsafe`.

text/0000-externally-implementable-functions.md
```diff
@@ -0,0 +1,184 @@
...
+- Should there be a way to specify that implementing the function is unsafe, separately from whether the function itself is unsafe?
+- Should not having an implementation be an error when the function is never called?
+- If we do end up designing and providing an `extern impl Trait` feature in addition to `extern impl fn`, should we *only* provide `extern impl Trait`, or is there value in still providing `extern impl fn` as well? This RFC proposes that we should still have `extern impl fn`, for the simpler case, rather than forcing such functions to be wrapped in traits.
+- An `extern impl fn` that's marked as `pub(crate)` but is nonetheless pub to *implement* could surprise people. Is there some way we can make this less surprising? Should we require that all `extern impl fn` have `pub` visibility?
+
+# Future possibilities
+
+- Adding a syntax to specify an existing function as the impl. E.g. `impl core::panic_handler = my_panic_handler;`
+- Doing this for `static` items too. (Perhaps all items that can appear in an `extern "Rust" { … }` block.)
+- Using this for existing overridable global behavior in the standard library, like the panic handler, global allocator, etc.
```
hanna-kruppe, line 175: Not a problem for this RFC since it's just a future possibility, but I wanted to write this up *somewhere* before I forget it again since it's non-obvious and I haven't seen it spelled out before: the "only one impl allowed, except that the declaration can include a default body" semantics work fine for the panic handler but not for the global allocator. The panic handler is declared and used in `core`, `std` supplies a definition that can't be overridden once it's pulled in. In contrast, the global allocator is declared and used in `alloc`, but `std` effectively adds a "default" implementation that can be overridden.

boozook:
Usage-counter of eii-implementation should inherit number of uses (references/calls) of a eii-definition, imho. 

Currently if 
- crate "eii-def" contains definition (with default body)
- crate "eii-impl" contains implementation
- crate "end-client" depend on both, calls `eii-def::eii_definition`

so then DCE believes that `eii-impl` code is not used and removes it, and I getting call of default impl defined in the `eii-def` (default body) or error if there is no default.

- - -

_Maybe I'm wrong, it looks like it minimizes duplicate conflicts and gives more manual control, but requires you to explicitly import the implementation by-hands._

bjorn3:
I wonder if this is just a lack of `--whole-archive` causing the linker to skip all object files in `eii-impl`, which then results in the weak definition in `eii-def` taking effect. If so this has nothing to do with DCE at all.

boozook:
@bjorn3, not sure really, because just found [this](https://github.com/rust-lang/rfcs/pull/3632#discussion_r1596708814):
> - multiple impl's are allowed
> - __the root crate must import the impl they want__
> - the normal-default impl is imported via the prelude

bjorn3:
As currently implemented, `extern crate eii_impl;` will not be considered unused by rustc. Even the unused crate lint doesn't trigger. In any case if you have a concrete example where it goes wrong, please open an issue with the reproducer attached.

kanashimia:
An alternative to this RFC is to have something like what other languages "implicit parameters" or "contextual parameters" as in Scala, Kotlin, and Haskell.
Basically similar idea to what this RFC proposes, but instead of a single global singleton it is something that is lexically scoped.
If it is scoped then overriding definition for some part of the code is obvious, by just adding another nested scope.
That is quite a bit more useful.

bjorn3:
Implicit parameters don't work when you need to go through the C ABI, through function pointers or through virtual calls of trait object methods and it adds runtime overhead. As such it can't replace the implementations of `#[global_allocator]`, `#[panic_handler]`, `#[alloc_error_handler]` (which are already global singletons) and all the other similar features we may want to add in the future. Implicit parameters do have uses, but their use cases only partially overlap with EII. So IMHO if we add implicit parameters, we should add them in addition to EII rather than instead of EII.

PR #3635, opened 2024-05-13
[RFC] externally definable statics

m-ou-se:
This is an alternative to https://github.com/rust-lang/rfcs/pull/3632

[Rendered](https://github.com/m-ou-se/rfcs/blob/extern-static/text/0000-externally-definable-statics.md)

Tracking:
- https://github.com/rust-lang/rust/issues/125418

### DIFF
--- a/text/0000-externally-definable-statics.md
+++ b/text/0000-externally-definable-statics.md
@@ -1,0 +1,186 @@
+- Feature Name: `extern_static`
+- Start Date: 2024-05-13
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+
+A mechanism for defining a static whose definition can be provided (or overridden) in another crate.
+
+Example 1:
+
+```rust
+// core::panic:
+
+extern static PANIC_HANDLER: fn(_: &PanicInfo) -> !;
+
+// user:
+
+impl static core::panic::PANIC_HANDLER = |panic_info| {
+    eprintln!("panic: {panic_info:?}");
+    loop {}
+};
+```
+
+Example 2:
+
+```rust
+// log crate:
+
+extern static LOGGER: Logger = Logger::default();
+
+// user:
+
+impl static log::LOGGER = Logger::to_stdout().with_colors();
+```
+
+# Motivation
+
+We have several items in the standard library that are overridable/definable by the user crate.
+For example, the (no_std) `panic_handler`, the global allocator for `alloc`, and so on.
+
+Each of those is a special lang item with its own special handling.
+Having a general mechanism simplifies the language and makes this functionality available for other crates, and potentially for more use cases in core/alloc/std.
+
+# Explanation
+
+A `static` can be defined as "externally definable" using the `extern` keyword as follows:
+
+```rust
+// In crate `x`:
+
+// Without a default:
+extern static A: fn();
+
+// With a default:
+extern static B: fn() = || {
+    println!("default impl");
+};
+```
+
+Another crate can then provide (or override) the definition of these statics using `impl static` syntax (using their path) as follows:
+
+```rust
+// In another crate:
+
+impl static x::A = || {
+    println!("my implementation of A");
+};
+
+impl static x::B = || {
+    println!("my implementation of B");
+};
+```
+
+# Details
+
+## Signature
+
+The type on the `impl static` item is optional, as it is taken from the `extern static` item referred to by the path.
+If it is given, the types must match.
+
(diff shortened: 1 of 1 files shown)


davidhewitt:
This strikes me a lot as a simpler form of "distributed slice" a la dtolnay's [`linkme`](https://github.com/dtolnay/linkme).

The extensible nature of such a slice would solve more use cases over setting a single value. An easy to see motivation is for custom benchmark or test harnesses to collect cases together via proc macros. Another use case familiar to me is for PyO3 to support collecting methods from multiple impl blocks annotated with `#[pymethods]` to expose to Python.

Perhaps that can be mentioned as a future possibility to this RFC?

bjorn3:
There is a big difference between this RFC and distributed slices: This RFC is trivially implementable across pretty much all linkers in existence. (excluding the default implementation support) As far as linkers are concerned there isn't really a difference between crate_a referencing a symbol and crate_b implementing it (an externally defined static) and vice versa (a regular static referenced from another crate) (except for cycles with static libraries, for which some linkers require `--start-group`/`--end-group` or mentioning the static libraries twice) Distributed slices however require either platform specific mechanisms to either concatenate sections or run static constructors or they need rustc to wrap the linker such that it can generate a definition for the distributed slice based on the crate metadata of all dependencies.

Adding distributed slices as future possibility to this RFC is fine with me.

davidhewitt:
That is indeed how `linkme` currently implements them; I was hoping that a first-class language feature may be able to find an alternative implementation pathway instead of reliance on the platform & linker. Perhaps that's wishful thinking.

kennytm:
OK I forgot one thing about `#[global_allocator]`: the *type* of the `static` is not a fixed type but anything that implements `GlobalAlloc`. Meaning you need either one of the following to reproduce `#[global_allocator]`

```rust
// (1) type the static as `dyn Trait`, 
// perhaps the most faithful representation but perhaps requires unsized_locals (rust-lang/rust#48055)
extern static ALLOCATOR: dyn GlobalAlloc;
impl static ALLOCATOR: dyn GlobalAlloc = SimpleAllocator { ... };

// (2) type the static as `&dyn Trait`,
// incurring double reference though.
extern static ALLOCATOR: &dyn GlobalAlloc;
impl static ALLOCATOR: &dyn GlobalAlloc = &SimpleAllocator { ... };

// (3) type the static as `impl Trait`, 
// maybe not linker-friendly? and not sure if non-object-safe traits can really be used.
extern static ALLOCATOR: impl GlobalAlloc;
impl static ALLOCATOR: impl GlobalAlloc = SimpleAllocator { ... };
```

fbstj:
maybe a silly question: is this not similar to weak linkage?

bjorn3:
In the case where there is a default, yes weak linkage is a possible implementation strategy. For the case where there is no default, it is no different from a regular extern reference in C. No weak linkage need and should be involved as linking must fail if the function is not provided.

jdonszelmann:
@davidhewitt Distributed slice, what you're talking about is actually an open issue in the testing devex team: https://github.com/rust-lang/testing-devex-team/issues/3. I'm currently (slowly) working towards a proposal for that: https://internals.rust-lang.org/t/global-registration-a-kind-of-pre-rfc/20813/23. I'm afraid it will take a little, cause we'll likely do some experimentation in a T-lang experiment. That's distinct from this issue. 

About this rfc: I think I liked the other one slightly better. I think that these externally definable statics are very often going to be functions, and having to use closure syntax for them all the time might be a bit awkward. I guess how Mara phrased it in the alternative RFC: that one might be a bit more ergonomic, and I like that. I like having both functions and statics be available to define externally.

text/0000-externally-definable-statics.md
```diff
@@ -0,0 +1,186 @@
...
+An alternative could be to use the same visibility for both defining an accessing, which would simply mean that the static (or an alias to it) will always have to be `pub`.
+
+# Prior art
+
+[RFC 2494 "Existential types with external definition"](https://github.com/rust-lang/rfcs/pull/2492)
+has been proposed before, which basically does this for *types*. Doing this for statics (as a start) saves a lot of complexity.
+
+# Unresolved questions
+
+- Should we allow some form of subtyping, similarly to how traits allow trait impls to do subtyping?
```
joshtriplett, line 171: I think we should, but I think it'd be fine to move this to "future work" rather than requiring this in the initial version. It's a backwards-compatible addition, and *not* allowing subtyping for now doesn't close the door for doing it in the future.

Evian-Zhang:
It appears to me that the externally defined statics and functions are generalized form of "associated consts" and "required methods", generalized from trait to crate. The root library crate behaves like a trait to define those items, and the leaf binary crate behaves like a struct to implement such items. Is it better to make these two semantics have similar syntax to make it more consistent?

PR #3642, opened 2024-05-22
[RFC] Thread spawn hook

m-ou-se:


m-ou-se:
cc @epage - You said you wanted this for your new test framework. :)

m-ou-se:
Implementation: https://github.com/rust-lang/rust/pull/125405

text/3642-thread-spawn-hook.md
```diff
@@ -0,0 +1,238 @@
...
+pub fn add_spawn_hook<F, G>(hook: F)
+where
+    F: 'static + Sync + Fn(&Thread) -> std::io::Result<G>,
+    G: 'static + Send + FnOnce();
+```
+
+# Implementation
+
+The implementation could simply be a static `RwLock` with a `Vec` of
+(boxed/leaked) `dyn Fn`s, or a simple lock free linked list of hooks.
```
epage, line 114: As an alternative, we could use [global registration](https://github.com/rust-lang/rust/issues/125119)
- Limits what you can do in your `hook`
- Would require a macro and some complexity
- Would only work for static linking with `std`
- Blocks this on a language feature proposal

the8472:
Could this be made more controllable through `thread::Builder`? We could make it `Clone`, modify a global default and `new()` would then clone from the default, giving users the option to modify it further (such as modifying the set of hooks) before spawning additional threads.

text/3642-thread-spawn-hook.md
m-ou-se: Are there any advantages to that approach? I can only think of downsides ^^'.

text/3642-thread-spawn-hook.md
```diff
@@ -0,0 +1,245 @@
...
+That way, a hook added by `add_spawn_hook` will only affect the current thread
+and all (direct and indirect) future child threads of the current thread,
+not other unrelated threads.
+
+Both are relatively easy and efficient to implement (as long as removing hooks
+is not an option).
+
+However, the first (global) behavior is conceptually simpler and allows for more
+flexibility. Using a global hook, one can still implement the thread local
+behavior, but this is not possible the other way around.
```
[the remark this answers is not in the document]
m-ou-se: I don't see how Command::pre_exec is related. That provides a closure to run for one specific Command, not a way to hook into all future Commands.

text/3642-thread-spawn-hook.md
epage: Removes all overhead when not used and keeps the overhead minimal.  Likely won't be a problem.  I mostly mentioned this because I had the thought to use this when it was posted but as I wrote up my post I realized how long the downsides were and thought it worth recording in case others have the thought.

m-ou-se:
> Could this be made more controllable through `thread::Builder`? We could make it `Clone`, modify a global default instance and `new()` would then clone from the default, giving users the option to modify it further (such as modifying the set of hooks) before spawning additional threads.

That might make sense for the stack size, but not for the thread name. And those are (at least today) the only two settings that a `Builder` has.

the8472:
The default thread name would be empty, as it is today.
And people have asked for more features in builders such as affinity or thread priorities (or hooks that can take care of those for specific threads):  https://github.com/rust-lang/libs-team/issues/195

> I don't think it makes sense to allow 'modifying' the set of hooks. If you want to change/remove any hook, you need a way to identify them. 

The current proposal provides `add`. With the builder we could at least have `clear` to opt-out for some subset of threads.
Since the closures have to be 'static anywau we could just do it based on typeID? The types could be made nameable via existential types.

m-ou-se:
> > 
> 
> The current proposal provides `add`. With the builder we could at least have `clear` to opt-out for some subset of threads. Since the closures have to be 'static anyway we could just do it based on typeID? The types could be made nameable via existential types.

Why would you want to fully `clear` the hooks? Without knowing which hooks are registered, you can't know which things you're opting out of, so you basically can't know what `clear` even means. E.g. would you expect clearing these hooks to break a `#[test]` with threads?

I strongly believe this should just work like `atexit()` (global, only registering, no unregistering).

(Even if we were to allow some way to skip the hooks for a specific builder/thread, it doesn't make sense to me to have a global "clear" function to clear all the registered hooks (from the 'global builder' as you propose), because you can't really know which other hooks (of other crates) you might be removing.)

> Since the closures have to be 'static anyway we could just do it based on typeID?

TypeId is not enough. Two hooks might have the same type but a different value (and different behaviour), e.g. with different captures. (Also if you register already boxed `Box<dyn Fn>` as hooks, they will all just have the (same) TypeId of that box type.)

the8472:
> Why would you want to fully clear the hooks? Without knowing which hooks are registered, you can't know which things you're opting out of, so you basically can't know what clear even means.

The RFC already states that threads can be in an unhooked state through pthread spawn. And I assume some weird linkage things (dlopening statically linked plugins?) you could end up with a lot more threads not executing hooks?
So unlike atexit it should already be treated as semi-optional behavior.

Additionally I'm somewhat concerned about a tool intended for automatic state inheritance being a global. In the Enterprise Java world such kind of implicit behavior/magic automatisms occasionally get abused to do terribly bloated things which means it's great when that bloat can be limited to a smaller scope or at least be opted out of.
For similar *increasingly complex inheritable process properties* problems linux systems are trying to occasionally shed this kind of state  (e.g. `sudo` vs. systemd's `run0`, or process spawn servers instead of `fork`)


In the context of a test framework I can imagine tests running in sets and then clearing global state (or at least enumerating it to print information what didn't get cleaned up) to decouple tests from each other.

> E.g. would you expect clearing these hooks to break a #[test] with threads?

"break" in the sense of not doing output capture, as it already is today? Yeah, sure, if the tests need a clean environment for whatever reason that might be a side-effect.

text/3642-thread-spawn-hook.md
m-ou-se: FWIW, the cost of spawning a new thread (not only allocating its stack, its Thread object, etc., but also the actual creation of the thread through a syscall) completely dwarfs any overhead from an indirect function call, so I don't think overhead is an argument here.

(For similar reasons, `std::panic::set_hook` is also just a simple function, rather than a magical attribute like no_std's `#[panic_handler]`.)

I'll add a note in the RFC on this.

m-ou-se:
> as it already is today?

Today, inheriting the output capturing is done unconditionally when `std` spawns a thread. There is no way to disable it or bypass it for specific threads (other than just using pthread or something directly). What I'm proposing here is a drop-in replacement for that exact feature, but more flexible so it can be used for other things than just output capturing. It will work in the exact same situations.

I'm not aware of anyone doing anything specifically to avoid inheriting libtest's output capturing (such as using pthread_create directly for that purpose).

> Additionally I'm somewhat concerned about a tool intended for automatic state inheritance being a global.

The destructors for e.g. `thread_local!{}` are also registered globally (depending on the platform). `atexit()` is global. All these things start with a global building block.

I agree it'd be nice to have some sort of `with_thread_local_context(|| { ... })` to create a scope in which all threads spawned will have a certain context, but the building block you need to make that feature is what is proposed in this RFC.

text/3642-thread-spawn-hook.md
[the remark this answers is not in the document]
joshtriplett: It seems theoretically possible to implement the global behavior via the thread-local behavior, using a thread-local hook that installs itself in the new thread. But I agree that that's not ideal.

joshtriplett:
I do think we're likely to want deregistration as well, but I think we'd want to do that via a scoped function rather than naming. I don't think it *has* to be in the initial version in any case.

ChrisDenton:
Are there use cases other than testing frameworks in mind? Do they have needs that differ from the testing framework case?

joshtriplett:
I think hooks are going to be quite prone to potential deadlocks. That's not a blocker for doing this, just something that'll need to be documented.

text/3642-thread-spawn-hook.md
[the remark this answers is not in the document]
kennytm: @m-ou-se If the local effect isn't `thread::Builder::add_spawn_hook()` I'm not sure what is meant by "local effect" in this section then :thinking:

tmccombs:
> Are there use cases other than testing frameworks in mind?

I can think of a few other use cases:

-  creating a new span in a tracing framework for the new thread that uses the thread local context of the parent thread as the parent span.
-  creating a new logger in TLS with configuration based on the local logger of the parent thread
-  inheriting a value for a thread local variable across spawning a new thread could have several use cases
-  possibly recording metrics on when threads are started, although that would probably be most useful if there was some way to hook into thread destruction as well.
- Initialize a thread local random number generator seeded by the parent thread's random number generator. I'm not sure if that is advantageous over current approaches to thread local RNGs, but it would be possible with this.

programmerjake:
> * possibly recording metrics on when threads are started, although that would probably be most useful if there was some way to hook into thread destruction as well.

TLS drop implementations could likely be used for that.

text/3642-thread-spawn-hook.md
```diff
@@ -0,0 +1,263 @@
...
+(boxed/leaked) `dyn Fn`s, or a simple lock free linked list of hooks.
+
+Functions that spawn a thread, such as `std::thread::spawn` will eventually call
+`spawn_unchecked_`, which will call the hooks in the parent thread, after the
+child `Thread` object has been created, but before the child thread has been
+spawned. The resulting `FnOnce` objects are stored and passed on to the child
+thread afterwards, which will execute them one by one before continuing with its
+main function.
+
+# Downsides
```
kpreid, line 123: When this feature is used as intended, it will now be the case, where it wasn't before, that Rust threads inherit some characteristics from _the thread which spawned them._ This could result in unintended nondeterministic behavior when a thread is spawned lazily. For example, `rayon` lazily spawns its global thread pool the first time a parallel operation is used.

Therefore, I think that

* The above should be listed as a downside (it's one more piece of complexity that library authors may need to keep in mind).

* There should be a way to opt out in `thread::Builder`, resulting in the new thread having the same state as if it was created first thing in `main()`.

  (In another design I have seen for tasks inheriting state, for a language other than Rust, there was no opt-out, *but* there was also no equivalent of `static` items, so it was always *possible* for the action that captured the local state (here, thread spawning) to be be done exactly at the time of construction of the thing that interacts with the spawned thread, rather than lazily later. Because Rust has `static` items but a thread can't be spawned at program loading time, I think we need this option.)

text/3642-thread-spawn-hook.md
```diff
@@ -0,0 +1,263 @@
...
+  already been added (e.g. in a static `AtomicBool`).
+
+- The hooks will not run if threads are spawned through e.g. pthread directly,
+  bypassing the Rust standard library.
+  (However, this is already the case for output capturing in libtest:
+  that does not work across threads when not spawned by libstd.)
+
+# Rationale and alternatives
+
+## Use of `io::Result`.
```
kpreid, line 140: A hazard here is that lazy library authors might write an insufficiently contextful error. In particular, they might use the `?` operator to return a real IO error from some kind of per-thread resource creation; such errors are already cryptic, and become more so when separated from their original context.

Therefore, it might make sense to ensure that, if `thread::Builder::spawn()` fails due to a hook, the error it returns is a *wrapped* error whose `to_string()` is something along the lines of "failed to execute thread spawn hook" and whose `Error::source()` is the error the hook returned. (The message could even include the [`Location`](https://doc.rust-lang.org/std/panic/struct.Location.html) of the `set_hook()` call, so that when debugging, one knows _which_ hook to blame.)

If that is done, then the error type can, and perhaps should, be something other than `io::Error`, because there is no advantage to the hook _producing_ an `io::Error` specifically, and making it an IO error might be misleading. (Use `Box<dyn Error + Send + Sync>` if nothing better comes to mind.)

text/3642-thread-spawn-hook.md
```diff
@@ -0,0 +1,263 @@
...
+
+# Downsides
+
+- The implementation requires allocation for each hook (to store them in the
+  global list of hooks), and an allocation each time a hook is spawned
+  (to store the resulting closure).
+
+- A library that wants to make use of inheriting thread locals will have to
+  register a global hook, and will need to keep track of whether its hook has
+  already been added (e.g. in a static `AtomicBool`).
```
kpreid, line 131: `AtomicBool` isn't sufficient; there are more than two states required to implement this initialization without any race conditions. The right tool is [`std::sync::Once`](https://doc.rust-lang.org/std/sync/struct.Once.html), which is exactly for this shape of problem.

```suggestion
- A library that wants to make use of inheriting thread locals will have to
  register a global hook, and will need to keep track of whether its hook has
  already been added (e.g. in a static `std::sync::Once`).
```

m-ou-se:
I updated the RFC:

1. The effect is now thread-local instead of global. Adding a hook only affects future child threads of the current thread.
2. Removed the use of io::Result.
3. Added an opt-out method to the thread builder.

m-ou-se:
> I think hooks are going to be quite prone to potential deadlocks. That's not a blocker for doing this, just something that'll need to be documented.

It'd be quite unusual to do anything interesting with locks in a hook. They are mainly used with thread locals (e.g. a Cell) rather than globals (e.g. a Mutex).

Joj501 (approved):

text/3642-thread-spawn-hook.md
```diff
@@ -0,0 +1,278 @@
...
+comes at a significant cost.
+
+## `impl` vs `dyn` in the signature
+
+An alternative interface could use `dyn` instead of generics, as follows:
+
+```rust
+pub fn add_spawn_hook<F, G>(
+    hook: Box<dyn Fn(&Thread) -> io::Result<Box<dyn FnOnce() + Send>> + Send + Sync>
+);
```
juntyr, line 232:
```suggestion
pub fn add_spawn_hook<F, G>(
    hook: Box<dyn Fn(&Thread) -> Box<dyn FnOnce() + Send> + Send + Sync>
);
```

ian-h-chamberlain:
> Are there use cases other than testing frameworks in mind? Do they have needs that differ from the testing framework case?

Just to advocate for one other use case, I think a proposal like https://github.com/rust-lang/libs-team/issues/195 _could_ be implemented in terms of a feature like this (which would be great!), but I am not sure it's possible in its current form?

Since the hook takes a `&Thread`, it doesn't seem like it would be possible to influence the creation of the thread using APIs like `pthread_attr_setaffinity_np`, `pthread_attr_setschedparam`, etc. — unless there were some way to influence the attributes themselves by extension APIs on `std::thread::Thread`.

For this reason, I'd argue in favor of something like the aforementioned `ThreadSpawnHookHandle`, even if it was just a thin wrapper around `Thread` for now, to be more extensible in the future for influencing OS-specific thread spawning behavior:
```rs
struct ThreadSpawnHookHandle {
	thread: Thread, // or &Thread ?
	// possible future extension, or a sys:: abstracted version of this:
	#[cfg(unix)]
	attrs: libc::pthread_attr_t,
}
```
Then it could be used with something like this:
```rs
std::thread::add_spawn_hook(|handle| {
	// Maybe something like JoinHandleExt:
	let attrs = handle.pthread_attr_t_mut();
	let my_custom_sched = todo!();
	unsafe { 
		libc::pthread_attr_setschedparam(attrs, &my_custom_sched);
	}
	|| {}
});
```

I think having the flexibility to add APIs that affect thread spawning in the future, whether in a standard or OS-specific way, would be worth the trouble of having the (relatively small) inconvenience of this wrapper type, if there are no other reasons `&Thread` is needed specifically.

m-ou-se:
@ian-h-chamberlain Yeah, that possibility is mentioned in one of the unresolved questions in the RFC:

> - Should the hook be able to access/configure more information about the child thread? E.g. set its stack size.

m-ou-se:
We discussed this in the libs-api meeting; this seems ready for FCP.

@rfcbot merge

the8472:
> Since the hook takes a &Thread, it doesn't seem like it would be possible to influence the creation of the thread using APIs like pthread_attr_setaffinity_np, pthread_attr_setschedparam, etc. 

You could use `pthread_self`, `pthread_setschedparam`, `pthread_setaffinity_np` instead. Once they're set they're inherited by the OS anyway so they don't need inheritable hooks. So a one-shot hook in the thread builder would be sufficient for that. At least on platforms that have those, but the same applies to `pthread_attr_setaffinity_np`, hence the `_np`.

ian-h-chamberlain:
> You could use `pthread_self`, `pthread_setschedparam`, `pthread_setaffinity_np` instead. Once they're set they're inherited by the OS anyway so they don't need inheritable hooks. So a one-shot hook in the thread builder would be sufficient for that. At least on platforms that have those, but the same applies to `pthread_attr_setaffinity_np`, hence the `_np`.

Yeah, but one of the motivations for https://github.com/rust-lang/libs-team/issues/195 was to support platforms that _don't_ allow setting these kinds of parameters for the main thread /current thread (i.e. must be set on the attributes used to spawn the thread). My main concern with the proposed API as written is that there won't be any way to extend it to support those use cases in the future, without some invasive changes to how `Thread` spawning works in `std` today.

m-ou-se:
Tracking issue: https://github.com/rust-lang/rust/issues/132951

PR #3683, opened 2024-08-25
`#[derive(Default)]` on enum variants with fields

estebank:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3683)*

Support the following:

```rust
#[derive(Default)]
enum Foo {
    #[default]
    Bar {
        x: Option<i32>,
        y: Option<i32>,
    },
    Baz,
}
```

[Rendered](https://github.com/estebank/rfcs/blob/derive-default-enum-with-fields/text/0000-derive-default-enum-with-fields.md)

### DIFF
--- a/text/0000-derive-default-enum-with-fields.md
+++ b/text/0000-derive-default-enum-with-fields.md
@@ -148,7 +148,12 @@ This expands on [RFC-3107]. No other changes are needed.
 
 [perfect derives]: https://smallcultfollowing.com/babysteps/blog/2022/04/12/implied-bounds-and-perfect-derive/
 
-The usual drawback of increasing the complexity of the language applies. However, the degree to which complexity is increased is not substantial.
+The usual drawback of increasing the complexity of the implementation applies.
+However, the degree to which complexity is increased is not substantial. If
+anything, the complexity of the concepts needed to be understood is reduced, as
+there are fewer special cases users need to keep in mind when using
+`#[derive(Default)]`, as well as allow us to remove `impl Default`s from the
+standard library.
 
 [The same](https://github.com/rust-lang/rust/issues/26925) issue highlighted on
 [RFC-3107] of current `#[derive(Default)]` on `struct`s producing `impl`s with
--- a/text/0000-derive-default-enum-with-fields.md
+++ b/text/0000-derive-default-enum-with-fields.md
@@ -1,0 +1,267 @@
+- Feature Name: `derive-default-enum-with-fields`
+- Start Date: 2024-08-25
+- RFC PR: [rust-lang/rfcs#3683](https://github.com/rust-lang/rfcs/pull/3683)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Allow `#[derive(Default)]` on `enum` variants with data.
+
+```rust
+#[derive(Default)]
+enum Foo {
+    #[default]
+    Bar {
+        x: Option<i32>,
+        y: Option<i32>,
+    },
+    Baz,
+}
+```
+
+Previously, only unit enum variants were allowed to derive `Default`, by marking
+them with `#[default]`. This feature extens this support to tuple and struct
+enum variants with fields when they all implement `Default`. By extension this
+also means that tuple and struct enum variants with no fields are also suitable
+to be marked with `#[default]`.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, `#[derive(Default)]` is not usable on `enum` variants with data. To
+rectify this situation, we expand the existing `#[default]` attribute
+implementation to support tuple and struct variants.
+
+This allows you to use #[derive(Default)] on enums wherefore you can now write:
+
+```rust
+#[derive(Default)]
+enum Padding {
+    #[default]
+    Space {
+        n: i32,
+    },
+    None,
+}
+```
+
+This feature allows for more cases where `Default` can be derived, instead of
+explicitly implemented. This reduces the verbosity of Rust codebases.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+In the same way that `struct`s can be annotated with `#[derive(Default)]`:
+
+```rust
+#[derive(Default)]
+struct Bar {
+    x: Option<i32>,
+    y: Option<i32>,
+}
+```
+
+which expands to:
+
+```rust
+impl Default for Bar {
+    fn default() -> Bar {
+        Bar {
+            x: Default::default(),
+            y: Default::default(),
+        }
+    }
+}
+```
+
+The same annotation on an `enum` with a variant annotated with `#[default]`:
+
+```rust
+#[derive(Default)]
+enum Foo {
+    #[default]
+    Bar {
+        x: Option<i32>,
+        y: Option<i32>,
+    },
+    Baz,
+}
+```
+
+expands to:
+
+```rust
+impl Default for Foo {
+    fn default() -> Foo {
+        Foo::Bar {
+            x: Default::default(),
+            y: Default::default(),
+        }
+    }
+}
+```
+
+Because the expanded code calls `Default::default()`, if the fields do not
+implement `Default` the compiler will emit an appropriate error pointing at the
+field that doesn't meet its requirement.
+
+```
+error[E0277]: the trait bound `S: Default` is not satisfied
+ --> src/main.rs:4:5
+  |
+2 | #[derive(Default)]
+  |          ------- in this derive macro expansion
+3 | enum Foo {
+3 |     Bar {
+4 |         x: S,
+  |         ^^^^ the trait `Default` is not implemented for `S`
+  |
+  = note: this error originates in the derive macro `Default` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider annotating `S` with `#[derive(Default)]`
+  |
+1 + #[derive(Default)]
+2 | struct S;
+  |
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+[`default_enum_substructure`]: https://github.com/rust-lang/rust/blob/6bb4656ee2ad88425917e3d4ad7ec11a033f181c/compiler/rustc_builtin_macros/src/deriving/default.rs#L96C4-L96C29
+
+[`extract_default_variant`]: https://github.com/rust-lang/rust/blob/6bb4656ee2ad88425917e3d4ad7ec11a033f181c/compiler/rustc_builtin_macros/src/deriving/default.rs#L154C4-L154C27
+
+[`default_struct_substructure`]: https://github.com/rust-lang/rust/blob/6bb4656ee2ad88425917e3d4ad7ec11a033f181c/compiler/rustc_builtin_macros/src/deriving/default.rs#L63C4-L63C31
+
+In `rustc_builtin_macros/src/deriving/default.rs`, we change
+[`extract_default_variant`] to not filter *only* on `VariantData::Unit`, and
+[`default_enum_substructure`] to expand the `impl` in a similar way to
+[`default_struct_substructure`].
+
+[RFC-3107]: https://rust-lang.github.io/rfcs/3107-derive-default-enum.html.
+
+This expands on [RFC-3107]. No other changes are needed.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+[perfect derives]: https://smallcultfollowing.com/babysteps/blog/2022/04/12/implied-bounds-and-perfect-derive/
+
+The usual drawback of increasing the complexity of the language applies. However, the degree to which complexity is increased is not substantial.
+
+[The same](https://github.com/rust-lang/rust/issues/26925) issue highlighted on
+[RFC-3107] of current `#[derive(Default)]` on `struct`s producing `impl`s with
+incorrect bounds (non-[perfect derives]) applies to this proposal as well.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+[`derivative`]: https://crates.io/crates/derivative
+
+[`smart-default`]: https://crates.io/crates/smart-default
+
+As shown by the existence of [`derivative`] and [`smart-default`], there is a
+desire to fill this perceived gap in flexibility that the built-in
+`#[derive(Default)]` support has. We can do nothing and let the ecosystem sort
+this gap out.
+
+
+# Prior art
+[prior-art]: #prior-art
+
+## Procedural macros
+
+There are a number of crates which to varying degrees afford macros for default field values and
+associated facilities.
+
+
+### `#[derive(Derivative)]`
+
+[`derivative`]: https://crates.io/crates/derivative
+
+The crate [`derivative`] provides the `#[derivative(Default)]` attribute. With it, you may write:
+
+```rust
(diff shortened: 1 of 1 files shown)


estebank:
Follow up to https://github.com/rust-lang/rfcs/pull/3107. Noticed as an easy gap to fill during implementation of https://github.com/rust-lang/rfcs/pull/3681.

Lokathor:
Under drawbacks, I'd argue that this actually *reduces* complexity of the standard library.

People know about `derive(Default)` giving some sort of basic Default impl. When there's less special cases when we have to say "except it won't work here, here, and here", that means there's less complexity in the standard library.

estebank:
@Lokathor I'm unsure if I'm reading you right, but that sounds like a *positive* to me 😄

text/0000-derive-default-enum-with-fields.md
```diff
@@ -0,0 +1,267 @@
...
+to be marked with `#[default]`.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, `#[derive(Default)]` is not usable on `enum` variants with data. To
+rectify this situation, we expand the existing `#[default]` attribute
+implementation to support tuple and struct variants.
+
+This allows you to use #[derive(Default)] on enums wherefore you can now write:
```
jplatte, line 36:
```suggestion
This allows you to use `#[derive(Default)]` on enums wherefore you can now write:
```

Lokathor:
Yes, exactly my point. The RFC text says this is a drawback, to have increased complexity. That might be true from a compiler internals perspective, or whatever. However, from an end user perspective, things that "just work" without needing exceptions listed for when they won't work in rare cases, those things are less complex.

So, in other words, I'd delete that part of the Drawbacks section. This doesn't increase complexity from a user's perspective.

jplatte (commented):
I think the reason this was excluded in the previous RFC that added `#[default]` is that its interactions with generics aren't obvious - this RFC should likely be updated to address that.

Lokathor:
@jplatte that's listed in the Drawbacks section i think, could you be more specific about what you expect the RFC to say about generics?

jplatte:
I was wondering what the generated bounds would be on a generic enum that uses this feature.

Currently, `#[derive(Default)]` on a generic enum never emits any bounds: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=efec38bf6b014835f97773e9cc30067c

Let's consider this 

```rust
#[derive(Default)]
enum MyEnum<T, U> {
    #[default]
    Foo {
        field: Option<T>,
    }
    Bar(U),
}
```

I guess there wouldn't be any bounds on `U`, otherwise that would be inconsistent with the existing behavior. But would there be a `T: Default` bound, or an `Option<T>: Default` bound? I think the latter _could_ be justified with enums not supporting private fields, unlike structs.

Lokathor:
There would have to be a bound on `Option<T>` being default, because the expansion is to have the code be `Foo{ field: Default::default() }`, and `field` has the `Option<T>` type. That's the minimum amount of bounding.

It's possible that additional bounds might end up in the mix, because we don't always have perfect derives, as mentioned in the Drawbacks section. For example, the "current behavior" could arguably only apply to unit variants of enums being the default, and field-having variants would (accidentally) drag in all the generics.

jplatte:
> There would have to be a bound on `Option<T>` being default, because the expansion is to have the code be `Foo{ field: Default::default() }`

I don't think there would have to be. The more "obvious" non-perfect-derive expansion would be `T: Default`, just like [for structs](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7e8c80d3dc3beee620fa644a927e80f29).

Lokathor:
Well, `T` being `Default` has nothing really to do with if `Option<T>`, the actual type of the field, is also `Default`.

Or to be more general, if we have a field type using a generic, `F<T>` say, then we only care that `F<T>` is `Default`, not `T` itself.

* It could be the case that `F<T>` is *not* `Default` even when `T` is.
* It could be the case that `F<T>` is *always* `Default` even when `T` is not.
* It could be the case that `F<T>` is `Default` or not depending on some bounds for `T`.

And the Proc Macro doesn't know which case it is when it generating the impl. So, just generating a bound based on just `T` itself is almost fundamentally incorrect, and will generate bad impls in extremely trivial cases.

jplatte:
> The only way of doing perfect derive is with type system access.

How so? You could generate `where Option<T>: Default` as a bound if that's the field type.

> But with default field values, if a field is given an explicit default then `Default::default()` on it doesn't need to be called

Unfortunately this doesn't seem true to me. I would actually argue that #3681 makes things more complicated: Imagine if `Option` had a `const fn default_some() -> Self` associated fn that returns `Some(T::default())`. There would be no way for the `Default` derive to know that `field: Option<T> = Option::default_some()` requires the bound. I guess for the `= None` case, things are indeed simpler though because that's not a function call so it can't have any bounds of its own.

estebank:
> How so? You could generate `where Option<T>: Default` as a bound if that's the field type, no?

The only issue that I could think of when doing that blindly is the case of a recursive type, like the following case:

```rust
#[derive(Clone)]
struct A<T> {
    x: Option<Box<A<T>>,
}
```

Currently the expansion in that case is

```rust
impl<T: Clone> Clone for A<T> {
    fn clone(&self) -> A<T> {
        A {
            self.x.clone(),
        }
    }
}
```

The naïve expansion of this with an attempt at perfect derives could be

```rust
impl<T> Clone for A<T> where A<T>: Clone {
    fn clone(&self) -> A<T> {
        A {
            self.x.clone(),
        }
    }
}
```

which would be accepted by the compiler, but never allow anyone to call `<A<T> as Clone>::clone` for *any* `T`. A simple side-step might be to identify that `A` is used in any of the fields and then revert to the current behavior of imperfect derives. It would make `#[derive(X)]` imperfect for any recursive type (even if unneeded), but allow for more accurate derives in every other case.

> Imagine if `Option` had a `const fn default_some() -> Self where T: Default`

FWIW, that still requires `Default` to be `const`, which it isn't now, but point taken. I need to think more about that case, but I believe that the obligations are actually correctly propagated: you couldn't have an expression that has an implicit transitive bound that references a type parameter without that type parameter having that bound in the item definition.

Lokathor:
I'm not sure that the fact that the derive might do a weird thing in a small fraction of cases is even particularly a problem, because people are never required to use the derive for Default.

Slightly separately i also think less cases would be weird with Default than with the Clone example you give.

Lokathor:
@ehuss I just noticed that you tagged this T-lang, but isn't this more like a T-libs-api concern?

ehuss:
Attributes are handled by the lang team. I added t-libs-api (similar to #3107). In general, teams should feel welcome to add themselves or other teams (and remove themselves if they want). I also expect teams to exercise good judgement in case other teams should be involved.

estebank:
After a quick conversation with niko, he mentioned that the general version of perfect derives is blocked on resolving the cycle case *when it is indirect*:

```rust
struct A<T> {
    x: Option<Box<B<T>>,
}
struct B<T> {
    x: Option<Box<A<T>>,
}
impl<T> Clone for A<T> where B<T>: Clone {
    fn clone(&self) -> A<T> {
        A {
            self.x.clone(),
        }
    }
}
impl<T> Clone for B<T> where A<T>: Clone {
    fn clone(&self) -> B<T> {
        B {
            self.x.clone(),
        }
    }
}
```

This case is not detectable purely from the token tree/AST. @compiler-errors you're looking at the trait resolution side of this, right?

compiler-errors:
@estebank: Not sure why I got a ping directly lol, but this is blocked on coinduction which is blocked on the new trait solver.

matthieu-m:
As mentioned by @Lokathor above, reducing the number of special-cases is generally a win with regard to complexity (for the end-user).

Following on this, I would suggest moving on with the imperfect derive, as a struct would. Or any other derive than `Default` on an enum. Specifically, I would expand:

```rust
#[derive(Default)]
enum MyEnum<T, U> {
    #[default]
    Foo {
        field: Option<T>,
    }
    Bar(U),
}
```

to

```rust
impl<T: Default, U: Default> Default for MyEnum<T, U> {
    fn default() -> Self {
        Self::Foo { field: Default::default() }
    }
}
```

It's very much imperfect:

 - `U` is dragged in even though it doesn't even appear in this variant.
 - `T` is dragged in even though `Option<T>` doesn't require `T: Default` to be `Default`.

But it's **consistent**, and while consistency may be the hobgoblin of little minds, it does make it easier to learn and use the language.

It's simple enough to write the implementation manually in the few cases where it matters, and it's not like we all aren't used to it. A later RFC/feature can take care of handling smarter derives for all.

<details>
<summary>Deriving `Clone` on enums.</summary>

```rust
#[derive(Clone)]
pub enum Enum<T> {
    Foo(PhantomData<T>),
}
```

Expands to:

```rust
pub enum Enum<T> { Foo(PhantomData<T>), }

#[automatically_derived]
impl<T: ::core::clone::Clone> ::core::clone::Clone for Enum<T> {
    #[inline]
    fn clone(&self) -> Enum<T> {
        match self {
            Enum::Foo(__self_0) =>
                Enum::Foo(::core::clone::Clone::clone(__self_0)),
        }
    }
}
```

When `PhantomData` is cloneable regardless of whether its generic type is. That's how it is...
</details>

jplatte:
@matthieu-m It wouldn't be consistent with how `#[derive(Default)]` works on generic enums right now. I think the least surprising / most consistent option would rather be to add `Default` bounds for all generic parameters _that are mentioned within the `#[default]` variant's field types_.

Lokathor:
Just for reference, and to expand on what jplatte said, when you have a unit variant as the default you get a default impl that doesn't depend on all the generics being Default, because the generics aren't field of the unit type. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=be6e8ae45184fddb99d916ec6a237a25

So yes, a default bound on all generics regardless of where they're used is definitely too much bounding, and would not be consistent.

But again, I don't think we need a bound on the generics used by the default variant, what we would need is a bound on the exact field types used in the variant. The way that the generic is wrapped in other types is important. `i32` is Default, but `&i32` is *not*, so if a field is `&T` then a bound on `T: Default` is the wrong bound.

text/0000-derive-default-enum-with-fields.md
```diff
@@ -0,0 +1,272 @@
...
+  refers to the same trait as the name of the macro is from the perspective of the language's static
+  semantics entirely coincidental.
+
+  However, for users who aren't aware of this, it may seem strange that `SmartDefault` should derive
+  for the `Default` trait.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Should we wait until [perfect derives] are addressed first?
```
estebank, line 248: > But again, I don't think we need a bound on the generics used by the default variant, what we would need is a bound on the exact field types used in the variant. The way that the generic is wrapped in other types is important. `i32` is `Default`, but `&i32` is not, so if a field is `&T` then a bound on `T: Default` is the wrong bound.

@Lokathor but it is *the same* wrong bound we have for `struct`s. What you want is perfect derives, which is explicitly called out in the text: do we wait for perfect derives or land support for the same imperfect behavior we have today for `struct`s.

The reason we *can't* do `Option<T>: Default` is because we only have AST information on these macros, and the only way to provide those bounds *correctly* is to detect cycles, both direct (we can detect those) *and* indirect (we can't detect those unless we have resolve/typeck info).

As I'm writing this though, I'm thinking that it *could* be that we could change the desugaring of `derive` to use the more accurate `where` clauses, and have a late lint check for cycles after the fact and tell people to instead write an explicit `impl` instead 🤔 

But that would be its own RFC for "get closer to perfect derives without trait solver changes", and I'm gonna be off-line for several weeks, so I can't drive that now.

matthieu-m:
> @matthieu-m It wouldn't be consistent with how `#[derive(Default)]` works on generic enums right now. I think the least surprising / most consistent option would rather be to add `Default` bounds for all generic parameters _that are mentioned within the `#[default]` variant's field types_.

It's another possibility yes.

It's inconsistent with the derivation of every other trait, but no other trait has a `#[default]` annotation on a specific variant, so maybe it's good enough.

text/0000-derive-default-enum-with-fields.md
Lokathor: > do we wait for perfect derives or land support for the same imperfect behavior we have today for structs.

If we can have it be better than before, I don't see a reason to not make it better than before. If it means that `enum` can (temporarily) derive more precisely than `struct` that seems completely fine to me.

I mean we could start with a bound on all the field types *and also* on the generics (that's what structs do now), but we literally can't have a bound *just* on the generics, without the bound on the field types. That actually won't work, which is my point.

jplatte: I'm pretty confused, what do you mean when you say we can't do it / it doesn't work? It is what we have for structs now, why couldn't we do the same thing here? Yes, it's not going to always generate an implementation that passes typecheck. But the same is true for structs today, and AFAIK this is not considered a major problem.

Lokathor: That is basically what i mean, yes. That the code will have a build error.

RalfJung:
@estebank the enum in the PR description has no `#[derive(Default)]`. ;)

RalfJung:
> But again, I don't think we need a bound on the generics used by the default variant, what we would need is a bound on the exact field types used in the variant. The way that the generic is wrapped in other types is important. i32 is Default, but &i32 is not, so if a field is &T then a bound on T: Default is the wrong bound.

That's true, but it's not how our derives work, and this was discussed above.

I would find it quite surprising if we had two fundamentally different ways to generate code for derives of "things with fields". If we have a way to generate better derives (adding bounds on the fields, not the generic types), we should do that consistently.

The "enum variant without field" case is fairly harmless, but once there are arbitrary fields this is just going to look like more special cases, that work better in some situations but worse in others.

text/0000-derive-default-enum-with-fields.md
```diff
@@ -0,0 +1,272 @@
...
+[`default_struct_substructure`]: https://github.com/rust-lang/rust/blob/6bb4656ee2ad88425917e3d4ad7ec11a033f181c/compiler/rustc_builtin_macros/src/deriving/default.rs#L63C4-L63C31
+
+In `rustc_builtin_macros/src/deriving/default.rs`, we change
+[`extract_default_variant`] to not filter *only* on `VariantData::Unit`, and
+[`default_enum_substructure`] to expand the `impl` in a similar way to
+[`default_struct_substructure`].
+
+[RFC-3107]: https://rust-lang.github.io/rfcs/3107-derive-default-enum.html.
+
+This expands on [RFC-3107]. No other changes are needed.
```
scottmcm, line 144: I don't think this is sufficient; the RFC should describe the behaviour in a way it can be understood without reference to the current source code.

scottmcm (changes requested):
If I'm understanding this right, this is still pending conversations about what should happen with the generic bounds?  Would a rule of "this errors on non-generic types for now" potentially be a way forward, or would that make this unsufficiently useful?

(The "this case really wants perfect derive" definitely feels strong here, but also that if it's perfect derive it should be the same general perfect derive as everything, not a enum-specific thing.)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3683/46781d0b1de3a07b10c2066d371ba7de2e209f15..1264c3ccd5535d66c441fc4fd60c53ded158036e)*

PR #3689, opened 2024-09-06
Update 0507-release-channels.md

Johnson1s:
[Rendered](https://github.com/Johnson1s/rfcs/blob/patch-1/text/0507-release-channels.md)


davidtwco:
This patch just adds the name of a repo or branch or something like that after the "Summary" header, it isn't fixing any typos or errors in this RFC, closing.

PR #3715, opened 2024-10-22
Unsafe derives and attributes

joshtriplett:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3715)*

Allow declaring proc macro attributes and derive macros as unsafe, and
requiring `unsafe` to invoke them.

[Rendered](https://github.com/joshtriplett/rfcs/blob/unsafe-derives/text/3715-unsafe-derives-and-attrs.md)

### DIFF
--- a/text/0000-unsafe-derives-and-attrs.md
+++ b/text/0000-unsafe-derives-and-attrs.md
@@ -1,0 +1,87 @@
+- Feature Name: `unsafe_derives_and_attrs`
+- Start Date: 2024-10-22
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Allow declaring proc macro attributes and derive macros as unsafe, and
+requiring `unsafe` to invoke them.
+
+# Motivation
+[motivation]: #motivation
+
+Some traits place requirements on implementations that the Rust compiler cannot
+verify. Those traits can mark themselves as unsafe, requiring `unsafe impl`
+syntax to implement. However, trait `derive` macros cannot currently require
+`unsafe`. This RFC defines a syntax for declaring and using unsafe `derive`
+macros.
+
+This RFC also defines a syntax for declaring proc macro attributes as unsafe.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Derives
+
+When declaring a proc macro `derive`, you can add the `unsafe` parameter to the
+`proc_macro_derive` attribute to indicate that the derive requires `unsafe`:
+
+```rust
+#[proc_macro_derive(DangerousTrait, unsafe)]
+pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
+    TokenStream::new()
+}
+```
+
+Invoking this derive requires writing either
+`#[unsafe(derive(DangerousTrait))]` or `#[derive(unsafe(DangerousTrait))]`.
+(The latter syntax allows isolating the `unsafe` to a single derive within a
+list of derives.) Invoking an unsafe derive without the unsafe derive syntax
+will produce a compiler error. Using the unsafe derive syntax without an unsafe
+derive will trigger an "unused unsafe" lint.
+
+A `proc_macro_derive` attribute can include both `attributes` for helper
+attributes and `unsafe` to declare the derive unsafe, in any order.
+
+## Attributes
+
+When declaring a proc macro attribute, you can add the `unsafe` parameter to
+the `proc_macro_attribute` attribute to indicate that the attribute requires
+`unsafe`:
+
+```rust
+#[proc_macro_attribute(unsafe)]
+pub fn dangerous(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    item
+}
+```
+
+Invoking an unsafe attribute requires the unsafe attribute syntax:
+`#[unsafe(dangerous)]`.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Should we support the `#[unsafe(derive(DangerousTrait))]` syntax, or only
+`#[derive(unsafe(DangerousTrait))]`? The former elevates the `unsafe` to be
+more visible, and allows deriving several traits using one `unsafe`. The latter
+isolates the `unsafe` to a specific trait. This RFC proposes supporting both,
+but we could choose to only support the latter instead.
+
+# Prior art
+[prior-art]: #prior-art
+
+RFC 3325 defined unsafe attributes. This RFC provides a natural extension of
+that mechanism to derives.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+When we add support for `macro_rules!`-based attributes and derives, we should
+provide a means for such attributes and derives to declare themselves unsafe as
+well.
+
+We could provide a syntax to declare specific helper attributes of a derive as
+unsafe, without declaring the entire derive unsafe.
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -62,6 +62,10 @@ struct SomeStruct { ... }
 struct AnotherStruct { ... }
 ```
 
+(Note that current rustfmt will place every derive on a line of its own if any
+have a comment. That could be changed in a future style edition, but this RFC
+is not making or advocating any style proposals.)
+
 ## Attributes
 
 When declaring a proc macro attribute, you can add the `unsafe` parameter to
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -93,16 +93,17 @@ instead, put the `unsafe` on the outside: `#[unsafe(derive(DangerousTrait))]`.
 
 Some rationale for putting it on the inside:
 - This encourages minimizing the scope of the `unsafe`, isolating it to a
-  single trait.
-- This allows writing all traits to derive within a single `#[derive(...)]`, if
-  desired. Putting the `unsafe` on the outside requires separate `derive`s for
-  safe and unsafe derives, and potentially multiple if the derives care about
-  ordering.
-- This makes it easy to attach `SAFETY` comments to each individual trait.
-- One way to think of `derive(Trait)` is `derive` is the mechanism to invoke
-  derive macros, and `Trait` is the actual derive macro. In this sense, when
-  writing `derive(unsafe(UnsafeTrait))`, the `unsafe` is attached to the
-  specific derive macro rather than the `derive` directive.
+  single derive macro.
+- This allows writing all derive macros to invoke within a single
+  `#[derive(...)]`, if desired. Putting the `unsafe` on the outside requires
+  separate `derive`s for safe and unsafe derives, and potentially multiple if
+  the derives care about ordering.
+- This makes it easy to attach `SAFETY` comments to each individual derive
+  macro.
+- One way to think of `derive(Macro)` is that `derive(..)` enters a context in
+  which one or more derive macros can be invoked, and naming `Macro` is how we
+  actually invoke the derive macro. When invoking unsafe derive macros, we have
+  to wrap those with `unsafe(..)` as in `derive(unsafe(DangerousDeriveMacro))`.
 
 We could use a different syntax for invoking unsafe derives, such as
 `derive(unsafe Trait)`. However, that would be inconsistent with unsafe
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -35,15 +35,31 @@ pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
 }
 ```
 
-Invoking this derive requires writing either
-`#[unsafe(derive(DangerousTrait))]` or `#[derive(unsafe(DangerousTrait))]`.
-(The latter syntax allows isolating the `unsafe` to a single derive within a
-list of derives.) Invoking an unsafe derive without the unsafe derive syntax
-will produce a compiler error. Using the unsafe derive syntax without an unsafe
-derive will trigger an "unused unsafe" lint.
+Invoking this derive requires writing `#[derive(unsafe(DangerousTrait))]`.
+Invoking an unsafe derive without the unsafe derive syntax will produce a
+compiler error. Using the unsafe derive syntax without an unsafe derive will
+trigger an "unused unsafe" lint.
 
 A `proc_macro_derive` attribute can include both `attributes` for helper
 attributes and `unsafe` to declare the derive unsafe, in any order.
+
+If writing code that enforces `SAFETY` comments for every use of `unsafe`, you
+can write the `SAFETY` comment either prior to the derive (for a single unsafe
+derive) or prior to the specific `unsafe(DangerousTrait)` in a list of derives:
+
+```rust
+/// SAFETY: ...
+#[derive(unsafe(DangerousTrait))]
+struct SomeStruct { ... }
+
+#[derive(
+    /// SAFETY: ...
+    unsafe(DangerousTrait),
+    /// SAFETY: ...
+    unsafe(AnotherDangerousTrait),
+)]
+struct AnotherStruct { ... }
+```
 
 ## Attributes
 
@@ -61,14 +77,28 @@ pub fn dangerous(_attr: TokenStream, item: TokenStream) -> TokenStream {
 Invoking an unsafe attribute requires the unsafe attribute syntax:
 `#[unsafe(dangerous)]`.
 
+If writing code that enforces `SAFETY` comments for every use of `unsafe`, you
+can write the `SAFETY` comment immediately prior to the attribute:
+
+```rust
+/// SAFETY: ...
+#[unsafe(dangerous)]
+```
+
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-Should we support the `#[unsafe(derive(DangerousTrait))]` syntax, or only
-`#[derive(unsafe(DangerousTrait))]`? The former elevates the `unsafe` to be
-more visible, and allows deriving several traits using one `unsafe`. The latter
-isolates the `unsafe` to a specific trait. This RFC proposes supporting both,
-but we could choose to only support the latter instead.
+This RFC proposes the synax `#[derive(unsafe(DangerousTrait))]`. We could,
+instead, put the `unsafe` on the outside: `#[unsafe(derive(DangerousTrait))]`.
+
+Some rationale for putting it on the inside:
+- This encourages minimizing the scope of the `unsafe`, isolating it to a
+  single trait.
+- This allows writing all traits to derive within a single `#[derive(...)]`, if
+  desired. Putting the `unsafe` on the outside requires separate `derive`s for
+  safe and unsafe derives, and potentially multiple if the derives care about
+  ordering.
+- This makes it easy to attach `SAFETY` comments to each individual trait.
 
 We could use a different syntax for invoking unsafe derives, such as
 `derive(unsafe Trait)`. However, that would be inconsistent with unsafe
@@ -80,14 +110,6 @@ modifier to `Trait` (e.g. an unsafe version of `Trait`).
 
 RFC 3325 defined unsafe attributes. This RFC provides a natural extension of
 that mechanism to derives.
-
-# Unresolved questions
-[unresolved-questions]: #unresolved-questions
-
-This RFC proposes accepting both `#[unsafe(derive(MyTrait))]` and
-`#[derive(unsafe(MyTrait))]`, among other reasons to make it easy to write
-`#[derive(SafeTrait, unsafe(MyTrait))]`. Should we allow both, or only allow
-the former?
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -107,6 +107,12 @@ Some rationale for putting it on the inside:
   actually invoke the derive macro. When invoking unsafe derive macros, we have
   to wrap those with `unsafe(..)` as in `derive(unsafe(DangerousDeriveMacro))`.
 
+We could, *if* we used the `unsafe(derive(...))` syntax, additionally restrict
+such derives to only contain a single trait, forcing the developer to only
+invoke a single derive macro per `unsafe(derive(...))`. However, this syntax
+naturally leads people to assume this would work, only to encounter an error
+when they do the natural thing that seems like it should work.
+
 We could use a different syntax for invoking unsafe derives, such as
 `derive(unsafe DangerousDeriveMacro)`. However, that would be inconsistent with
 unsafe attributes (which use parentheses), *and* it has the potential to look
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -1,6 +1,6 @@
 - Feature Name: `unsafe_derives_and_attrs`
 - Start Date: 2024-10-22
-- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- RFC PR: [rust-lang/rfcs#3715](https://github.com/rust-lang/rfcs/pull/3715)
 - Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
 
 # Summary
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -70,6 +70,11 @@ more visible, and allows deriving several traits using one `unsafe`. The latter
 isolates the `unsafe` to a specific trait. This RFC proposes supporting both,
 but we could choose to only support the latter instead.
 
+We could use a different syntax for invoking unsafe derives, such as
+`derive(unsafe Trait)`. However, that would be inconsistent with unsafe
+attributes (which use parentheses), *and* it has the potential to look like a
+modifier to `Trait` (e.g. an unsafe version of `Trait`).
+
 # Prior art
 [prior-art]: #prior-art
 
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -26,6 +26,17 @@ macro that implements the `unsafe` trait in a fashion that's always safe (or
 that fails if some obligation is not met), or an `unsafe` `derive` macro that
 puts the safety obligation on the invoker of the `derive`.
 
+Some examples of potential unsafe derives for `unsafe trait`s:
+
+- `TrustedLen` and `DerefPure` in the standard library. (Currently unstable,
+  but such a derive would be useful when they're stable, serving the function
+  of an `unsafe impl`.)
+- `pyo3::marker::Ungil` in `pyo3`, in place of the current handling of a
+  blanket impl for any `Send` type.
+- A hypothetical derive for `bytes::Buf` or similar. (For cases where a type
+  has fields or trivial expressions corresponding to the current slice and
+  offset.)
+
 This RFC also defines a syntax for declaring proc macro attributes as unsafe.
 
 # Guide-level explanation
--- a/text/3715-unsafe-derives-and-attrs.md
+++ b/text/3715-unsafe-derives-and-attrs.md
@@ -29,34 +29,35 @@ When declaring a proc macro `derive`, you can add the `unsafe` parameter to the
 `proc_macro_derive` attribute to indicate that the derive requires `unsafe`:
 
 ```rust
-#[proc_macro_derive(DangerousTrait, unsafe)]
-pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
+#[proc_macro_derive(DangerousDeriveMacro, unsafe)]
+pub fn derive_dangerous_derive_macro(_item: TokenStream) -> TokenStream {
     TokenStream::new()
 }
 ```
 
-Invoking this derive requires writing `#[derive(unsafe(DangerousTrait))]`.
-Invoking an unsafe derive without the unsafe derive syntax will produce a
-compiler error. Using the unsafe derive syntax without an unsafe derive will
-trigger an "unused unsafe" lint.
+Invoking this derive macro requires writing
+`#[derive(unsafe(DangerousDeriveMacro))]`. Invoking an unsafe derive macro
+without the unsafe derive syntax will produce a compiler error. Using the
+unsafe derive syntax without an unsafe derive macro will trigger an "unused
+unsafe" lint.
 
 A `proc_macro_derive` attribute can include both `attributes` for helper
 attributes and `unsafe` to declare the derive unsafe, in any order.
 
 If writing code that enforces `SAFETY` comments for every use of `unsafe`, you
 can write the `SAFETY` comment either prior to the derive (for a single unsafe
-derive) or prior to the specific `unsafe(DangerousTrait)` in a list of derives:
+derive) or prior to the specific `unsafe(DangerousDeriveMacro)` in a list of derives:
 
 ```rust
 // SAFETY: ...
-#[derive(unsafe(DangerousTrait))]
+#[derive(unsafe(DangerousDeriveMacro))]
 struct SomeStruct { ... }
 
 #[derive(
     // SAFETY: ...
-    unsafe(DangerousTrait),
+    unsafe(DangerousDeriveMacro),
     // SAFETY: ...
-    unsafe(AnotherDangerousTrait),
+    unsafe(AnotherDangerousDeriveMacro),
 )]
 struct AnotherStruct { ... }
 ```
@@ -88,8 +89,9 @@ can write the `SAFETY` comment immediately prior to the attribute:
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-This RFC proposes the synax `#[derive(unsafe(DangerousTrait))]`. We could,
-instead, put the `unsafe` on the outside: `#[unsafe(derive(DangerousTrait))]`.
+This RFC proposes the synax `#[derive(unsafe(DangerousDeriveMacro))]`. We
+could, instead, put the `unsafe` on the outside:
+`#[unsafe(derive(DangerousDeriveMacro))]`.
 
 Some rationale for putting it on the inside:
 - This encourages minimizing the scope of the `unsafe`, isolating it to a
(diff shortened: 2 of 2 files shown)


text/0000-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,87 @@
...
+attributes and `unsafe` to declare the derive unsafe, in any order.
+
+## Attributes
+
+When declaring a proc macro attribute, you can add the `unsafe` parameter to
+the `proc_macro_attribute` attribute to indicate that the attribute requires
+`unsafe`:
+
+```rust
+#[proc_macro_attribute]
```
programmerjake, line 55:
```suggestion
#[proc_macro_attribute(unsafe)]
```

joshtriplett: Argh. Good catch.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,87 @@
...
+
+```rust
+#[proc_macro_derive(DangerousTrait, unsafe)]
+pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
+    TokenStream::new()
+}
+```
+
+Invoking this derive requires writing either
+`#[unsafe(derive(DangerousTrait))]` or `#[derive(unsafe(DangerousTrait))]`.
```
ds84182, line 39: Wondering if you've considered unsafe without parenthesis.`#[derive(unsafe DangerousTrait)]` has less visual noise, at least to my eyes.

programmerjake: the `unsafe` uses parenthesis syntax was already decided on in [RFC 3325](https://github.com/rust-lang/rfcs/blob/master/text/3325-unsafe-attributes.md)

kennytm: `unsafe DangerousTrait` may be mistaken as a #3710-unsafe-flavored variant of the `DangerousTrait` trait, so no it shouldn't be considered.

joshtriplett: Added to the alternatives section with that rationale.

clarfonthey:
Just my 2¢, but I think that the shorthand for `derive(..., unsafe(Trait), ...)` is a bit unprecedented.

You have to separate out derive traits any time there's some different requirement, e.g. `cfg_attr(feature = "serde", derive(Serialize, Deserialize))`, and by keeping unsafe at the top level you can easily grep for `#!?\[unsafe\(` whereas `unsafe(` will catch any call to `method_unsafe(`.

Sure, it's likely it won't make a difference, but I think that only having to check attributes at the top level for unsafe to verify safety is best.

carbotaniuman:
The greppability is already broken by things like `#[cfg_attr(all(), unsafe(no_mangle)]`, and can be restored by just grepping `unsafe(` I think putting the unsafe with the trait makes syntactic sense as it discharges the safety obligation of each trait derive.

GnomedDev:
Any reason the syntax for defining an unsafe derive/attribute isn't `unsafe fn`?
```rust
#[proc_macro_attribute]
pub unsafe fn dangerous(_attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}
```

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,92 @@
...
+list of derives.) Invoking an unsafe derive without the unsafe derive syntax
+will produce a compiler error. Using the unsafe derive syntax without an unsafe
+derive will trigger an "unused unsafe" lint.
+
+A `proc_macro_derive` attribute can include both `attributes` for helper
+attributes and `unsafe` to declare the derive unsafe, in any order.
+
+## Attributes
+
+When declaring a proc macro attribute, you can add the `unsafe` parameter to
```
Noratrieb, line 50: @GnomedDev said
> Any reason the syntax for defining an unsafe derive/attribute isn't `unsafe fn`?
```rust
#[proc_macro_attribute]
pub unsafe fn dangerous(_attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}
```

Noratrieb: because that makes no sense, the unsafety is at a meta-level and there are no safety requirements on the macro function itself

GnomedDev: Is there not? I thought the point of an unsafe derive/attribute was that it was unsafe to trigger and had safety requirements?

Noratrieb: you're confusing the memory safety of the rustc process including the proc macro code with the safety of the generated code. this RFC is about the latter, making the macro an unsafe fn is about the former

GnomedDev: I don't personally think it's that confusing, especially for users who don't know how proc macros are loaded and run in the current implementation of the compiler.

clarfonthey:
> The greppability is already broken by things like `#[cfg_attr(all(), unsafe(no_mangle)]`, and can be restored by just grepping `unsafe(` I think putting the unsafe with the trait makes syntactic sense as it discharges the safety obligation of each trait derive.

I feel kind of silly for literally alluding to this point in my post and missing it somehow. You're right and I retract my original claim.

joshtriplett:
We had a @rust-lang/lang design meeting today on the set of macro RFCs. I've updated the RFC to incorporate the feedback from that design meeting.

Per the feedback in that meeting, I'm starting an FCP to start letting people register consensus for this RFC.

@rfcbot merge

joshtriplett:
@rust-lang/lang I've updated this to match the consensus syntax of `derive(unsafe(UnsafeTrait))`, and provided all the points of rationale that I remember from the discussion. Please feel free to post further suggestions if I missed any points of rationale.

carbotaniuman:
One thing I thought of was that it might make sense to make a derive maybe-unsafe and branch between whether `unsafe` was provided, although it might be better to just spell that out as 2 separate derives... The use case for that would be something where the derive can check the preconditions using `static_assert`s or similar, and one where you can promise the preconditions hold if they aren't checkable.

RalfJung:
> One thing I thought of was that it might make sense to make a derive maybe-unsafe and branch between whether unsafe was provided

I think that would be a big mistake. The presence or absence of `unsafe` should never affect the resulting generated code, it should only affect whether the code is accepted or not.

RalfJung:
> @rust-lang/lang I've updated this to match the consensus syntax of `derive(unsafe(UnsafeTrait))`, and provided all the points of rationale that I remember from the discussion. Please feel free to post further suggestions if I missed any points of rationale.

In terms of the mental model I find `derive(unsafe(Trait))` a bit odd -- this looks like the trait is called `unsafe(Trait)`. In the grammar of attributes (well, in its simplified form in my head), `unsafe($attr)` is also an `$attr`, and `derive($trait)` is an `$attr`, but to fit the proposed scheme into that model we have to do an ad-hoc extension of the grammar to allow `unsafe` inside` derive`.

OTOH, the RFC does make some good points in favor of "inner `unsafe`", so I am a bit torn here.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,122 @@
...
+can write the `SAFETY` comment either prior to the derive (for a single unsafe
+derive) or prior to the specific `unsafe(DangerousTrait)` in a list of derives:
+
+```rust
+/// SAFETY: ...
+#[derive(unsafe(DangerousTrait))]
+struct SomeStruct { ... }
+
+#[derive(
+    /// SAFETY: ...
```
Veykril, line 56: This strikes me as a bit odd, is this supposed to be a line comment (that is non-doc comment)? Documentation comments only have very specific places where they are allowed. Also they represent `#[doc = "content"]` so this would be equivalent to writing
```rust
#[derive(
    #[doc = "SAFETY: ..."]
    unsafe(DangerousTrait)
...
```

traviscross:
> In terms of the mental model I find `derive(unsafe(Trait))` a bit odd -- this looks like the trait is called `unsafe(Trait)`.

The argument we landed on, on this point (setting aside the ergonomic arguments in favor of the inner `unsafe`), is that the `derive` is always safe.  It's just triggering a code transformation.  What you have to satisfy preconditions for is the output of each of the derive macros, and that's what those paths represent.

Sometimes those derive macro names correspond directly to a trait, but sometimes they don't (e.g. `CoercePointee`).

joshtriplett:
@traviscross Thanks for the reminder. IIRC that was one of many possible rationales that came up in the meeting, but it's one that I hadn't captured. I've now done so.

RalfJung:
> The argument we landed on, on this point (setting aside the ergonomic arguments in favor of the inner unsafe), is that the derive word is always safe. It's just signaling that we're going to invoke a code transformation. It's the derive macro that is unsafe. That's what has the proof obligation. And that's what the path represents.

I don't understand that distinction. `no_mangle` also just invokes a particular code generation. But that generated code can lead to unsoundness and that's what makes in unsafe. This applies equally to `derive`.

Or put differently, a function invoking an unsafe function without checking its preconditions needs to be unsafe.

But good point that the name in the `derive` is a macro name, not a trait name. That had temporarily escaped my memory.

traviscross:
> I don't understand that distinction...

That distinction is tied exactly to:

> But good point that the name in the `derive` is a macro name, not a trait name.

That is, I agree that if `derive` were itself doing a `no_mangle`-like code transformation, that it should be unsafe.  But it's actually the derive macro that's doing that.  So `derive(..)` is kind of just a context marker that allows us to invoke one or more derive macros.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,126 @@
...
+instead, put the `unsafe` on the outside: `#[unsafe(derive(DangerousTrait))]`.
+
+Some rationale for putting it on the inside:
+- This encourages minimizing the scope of the `unsafe`, isolating it to a
+  single trait.
+- This allows writing all traits to derive within a single `#[derive(...)]`, if
+  desired. Putting the `unsafe` on the outside requires separate `derive`s for
+  safe and unsafe derives, and potentially multiple if the derives care about
+  ordering.
+- This makes it easy to attach `SAFETY` comments to each individual trait.
```
kennytm, line 101: Also I don't see how I also don't see how this is an advantage of `#[derive(unsafe(Foo))]` vs `#[unsafe(derive(Foo))]`. The latter is even 2 lines shorter.

```rust
#[derive(
    // SAFETY: ...
    unsafe(DangerousTrait),
    // SAFETY: ...
    unsafe(AnotherDangerousTrait),
)]
struct AnotherStruct { ... }

vs.

// SAFETY: ...
#[unsafe(derive(DangerousTrait))]
// SAFETY: ...
#[unsafe(derive(AnotherDangerousTrait))]
struct AnotherStruct { ... }
```

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,126 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This RFC proposes the synax `#[derive(unsafe(DangerousTrait))]`. We could,
+instead, put the `unsafe` on the outside: `#[unsafe(derive(DangerousTrait))]`.
+
+Some rationale for putting it on the inside:
+- This encourages minimizing the scope of the `unsafe`, isolating it to a
+  single trait.
```
kennytm, line 96: one could make `#[unsafe(derive(Foo, Bar))]` invalid syntax.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,126 @@
...
+- This allows writing all traits to derive within a single `#[derive(...)]`, if
+  desired. Putting the `unsafe` on the outside requires separate `derive`s for
+  safe and unsafe derives, and potentially multiple if the derives care about
+  ordering.
```
kennytm, line 100: afaik `#[derive(X)]` cannot see another `#[derive(Y)]` and all a derive-macro does is append an item, which is normally order-agnostic.

> [[items.decl-order]](https://doc.rust-lang.org/reference/items.html#r-items.decl-order)
>
> Items may be defined in **any order**, with the exception of `macro_rules` which has its own scoping behavior.

kennytm (commented):
While to the proc_macro author in `#[derive(Clone, Copy)] struct Foo { ... }` it may be understood as invoking the "macros" `Clone` and `Copy` on `struct Foo { ... }`, to an average user it is highly unintuitive to think like that. 

We would think
1. `#[derive(Clone, Copy)]` is just a short hand for `#[derive(Clone)] #[derive(Copy)]` and 
2. the `derive` and the trait (ok, macro) name is a single unit to drive the code generation. 

So `#[unsafe(derive(UnsafeTrait))]` makes more sense to me, which the `unsafe` is applied the entirety of `derive(UnsafeTrait)`, not the word `derive` alone.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,126 @@
...
+
+When declaring a proc macro `derive`, you can add the `unsafe` parameter to the
+`proc_macro_derive` attribute to indicate that the derive requires `unsafe`:
+
+```rust
+#[proc_macro_derive(DangerousTrait, unsafe)]
+pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
+    TokenStream::new()
+}
+```
```
traviscross, line 36: Given that we're putting the `unsafe` around the derive macro name at the use site, what do we think about matching that at the def site?  E.g.:

```rust
#[proc_macro_derive(unsafe(DangerousDeriveMacro)]
pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
    TokenStream::new()
}
```

```suggestion
When declaring a proc macro `derive`, you can wrap the derive macro name with
`unsafe(..)` in the `proc_macro_derive` attribute to indicate that the derive
requires `unsafe`:

```rust
#[proc_macro_derive(unsafe(DangerousDeriveMacro)]
pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
    TokenStream::new()
}
```
```

File: text/3715-unsafe-derives-and-attrs.md [hunk 9]
```diff
@@ -0,0 +1,126 @@
...
+struct SomeStruct { ... }
+
+#[derive(
+    // SAFETY: ...
+    unsafe(DangerousTrait),
+    // SAFETY: ...
+    unsafe(AnotherDangerousTrait),
+)]
+struct AnotherStruct { ... }
+```
```
traviscross, line 62: Given the distinction we're drawing to support the choice of syntax, it's probably better to use something like `DangerousDeriveMacro` rather than `DangerousTrait` in all of the examples.

This also removes the sentence about ordering with the `unsafe` parameter to match the suggestion above that we write `unsafe(DeriveMacro)` at the def-site.

```suggestion
Invoking this derive requires writing `#[derive(unsafe(DangerousDeriveMacro))]`.
Invoking an unsafe derive without the unsafe derive syntax will produce a
compiler error. Using the unsafe derive syntax without an unsafe derive will
trigger an "unused unsafe" lint.

If writing code that enforces `SAFETY` comments for every use of `unsafe`, you
can write the `SAFETY` comment either prior to the derive (for a single unsafe
derive) or prior to the specific `unsafe(DangerousDeriveMacro)` in a list of
derives:

```rust
// SAFETY: ...
#[derive(unsafe(DangerousDeriveMacro))]
struct SomeStruct { ... }

#[derive(
    // SAFETY: ...
    unsafe(DangerousDeriveMacro),
    // SAFETY: ...
    unsafe(AnotherDangerousDeriveMacro),
)]
struct AnotherStruct { ... }
```
```

Comment (traviscross) [reply text/3715-unsafe-derives-and-attrs.md:62]: (There remain other uses below that we'd probably want to update as well to match.)

File: text/3715-unsafe-derives-and-attrs.md (hunk 8 above)
Comment (carbotaniuman) [reply text/3715-unsafe-derives-and-attrs.md:36]: On one hand I like the symmetry, on the other hand (because we overloaded unsafe to have two meanings) I think this could be read as the derive macro itself discharging a safety obligation, whereas the first form makes it slightly clearer that this is just an unsafe derive.

File: text/3715-unsafe-derives-and-attrs.md [hunk 10]
```diff
@@ -0,0 +1,126 @@
...
+pub fn dangerous(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    item
+}
+```
+
+Invoking an unsafe attribute requires the unsafe attribute syntax:
+`#[unsafe(dangerous)]`.
+
+If writing code that enforces `SAFETY` comments for every use of `unsafe`, you
+can write the `SAFETY` comment immediately prior to the attribute:
```
traviscross, line 81:
```suggestion
You can write the `SAFETY` comment immediately prior to the attribute:
```

"What, you mean there's code somewhere that doesn't require that?"  I'd prefer to not hint that this expectation is somehow not the default.  Probably we can just leave this predicate out entirely.

text/3715-unsafe-derives-and-attrs.md
traviscross, line 96:
```suggestion
- This encourages minimizing the scope of the `unsafe`, isolating it to a
  single derive macro.
```

text/3715-unsafe-derives-and-attrs.md
traviscross, line 100:
```suggestion
- This allows writing all derive macros to invoke within a single
  `#[derive(...)]`, if desired. Putting the `unsafe` on the outside requires
  separate `derive`s for safe and unsafe derives, and potentially multiple if
  the derives care about ordering.
```

text/3715-unsafe-derives-and-attrs.md
traviscross, line 101:
```suggestion
- This makes it easy to attach `SAFETY` comments to each individual derive
  macro.
```

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,126 @@
...
+  single trait.
+- This allows writing all traits to derive within a single `#[derive(...)]`, if
+  desired. Putting the `unsafe` on the outside requires separate `derive`s for
+  safe and unsafe derives, and potentially multiple if the derives care about
+  ordering.
+- This makes it easy to attach `SAFETY` comments to each individual trait.
+- One way to think of `derive(Trait)` is `derive` is the mechanism to invoke
+  derive macros, and `Trait` is the actual derive macro. In this sense, when
+  writing `derive(unsafe(UnsafeTrait))`, the `unsafe` is attached to the
+  specific derive macro rather than the `derive` directive.
```
traviscross, line 105:
```suggestion
- One way to think of `derive(Macro)` is that `derive(..)` enters a context in
  which one or more derive macros can be invoked, and naming `Macro` is how we
  actually invoke the derive macro. When invoking unsafe derive macros, we have
  to wrap those with `unsafe(..)` as in `derive(unsafe(DangerousDeriveMacro))`.
```

traviscross:
Let's cancel the stale FCP and restart...

@rfcbot fcp cancel

traviscross:
This all looks correct and desirable to me with one amendment, which I'm including in this proposed FCP, which is that we change the def site for derive macros to match the use site, and write the `unsafe(..)` around the derive macro name in the `proc_macro_derive` attribute, e.g.:

```rust
#[proc_macro_derive(unsafe(DangerousDeriveMacro))]
pub fn derive_helper_attr(_item: TokenStream) -> TokenStream {
    TokenStream::new()
}
```

@rfcbot fcp merge

text/3715-unsafe-derives-and-attrs.md
joshtriplett: @kennytm This goes along with the previous point about not having to split the `derive`.

text/3715-unsafe-derives-and-attrs.md
joshtriplett: @carbotaniuman That's my feeling as well: `proc_macro_derive(unsafe(DangerousDeriveMacro))` reads to me as though the creation of it is unsafe, rather than it being a safe operation that creates an unsafe derive.

text/3715-unsafe-derives-and-attrs.md
joshtriplett: @traviscross It is, in a very practical sense, *not* currently the default: there's no warn-by-default lint for it. I'm all for changing that, but meanwhile, it'd be inaccurate to imply that it is currently required everywhere.

joshtriplett:
@traviscross As mentioned by @carbotaniuman and myself in response to that suggestion: that makes it look like something that's unsafe to *define* rather than unsafe to *use*. For that reason, I don't think that syntax is an improvement.

text/3715-unsafe-derives-and-attrs.md
joshtriplett: I don't think we should use a syntax that naturally leads people to assume this would work, and then complain when they do the natural thing that seems like it should work.

Nonetheless, I've captured this in the alternatives section.

text/3715-unsafe-derives-and-attrs.md
joshtriplett: @kennytm While I think it's good practice to make derives order-independent, it's very possible to make them order-dependent, in multiple ways. First, as you noted, the are exceptions like macros. Second proc macros can have side effects outside the scope of the generated code.

Nonetheless, I've called out that this is unusual.

text/3715-unsafe-derives-and-attrs.md
joshtriplett: I've added an alternative for this, along with rationale for why we shouldn't do it.

traviscross:
@rfcbot concern pick-def-site-name

traviscross:
We'll need to discuss it on a lang call regardless, and I'm rather curious to hear what others on the team think.  While I'm thinking about it, though, here are my notes on the case for symmetry between the def site and the use site.  (@joshtriplett and I talked these points over, and our interesting and fruitful discussion helped to sharpen them for this quick write-up.  These are my views only though.)

First off, I agree that it's a bit of a pun, but I think it's a kind of attractive pun in this case.  What pushes me over the line is that I just can't imagine that, having picked `unsafe(Macro)` at the use site, that we'd ever use `unsafe(Macro)` at the def site to mean something else.  If I thought we were seriously considering some alternative future meaning for that, I'd raise a concern about using `unsafe(Macro)` at the use site.

Broadly, I think it's fine, given how things are today, if `define(unsafe(Thing))` creates an obligation and `use(unsafe(Thing))` discharges it.  If we're unhappy with that, probably the problem is the reuse of the keyword, not the reuse of how that keyword is placed.  One thing that makes me happy with it here is that I just don't see any other plausible meaning for `define(unsafe(Thing))`.  The `Thing` doesn't exist yet -- we're defining it.  There's no way we could be discharging an obligation for it yet.  So we must be creating an obligation, just like we are in other places that we're defining something using `unsafe`, e.g. in `unsafe fn`.

I'm reminded here too about our recent syntax discussions for const trait impls.  We may not end up marking them at all -- that's what the current RFC draft says -- but when we were thinking of doing that, we had to pick between `const impl Trait for ..` and `impl const Trait for ..`.  There are a lot of good reasons to like `const impl Trait`, e.g. in terms of syntactic and conceptual parity with `const fn` and other uses.  But people really liked `impl const Trait for ..` because it matched the `T: const Trait` bound syntax.  We could say here too that this is just a pun as these are different.  One "requires a const impl" and the other "provides a const impl".  And yet, it seemed attractive to lean into this pun.  That's roughly similar to how I feel here, mapped to "discharging a proof obligation" and "creating a proof obligation".

(The parity we try to maintain between our type syntax and constructor expression syntax is maybe also a place where we could say that it's a bit of a pun between two distinct things, but that we like it.)

Another angle is to imagine what we'd do if we didn't have this feature.  Probably we'd still mark these unsafe derive macros by naming them things like `UnsafeThing`.  Then the def site would look like `define(UnsafeThing)` and the use site like `use(UnsafeThing)`.  So in that way, `define(unsafe(Thing))` and `use(unsafe(Thing))` feel equally fine to me.  It's like we're just binding the unsafety to the identity of the derive macro.

(Here and throughout, I'm using `define(..)` and `use(..)` to abstract the question a bit, and to avoid having to repeat `proc_macro_derive(..)` everywhere.)

Regarding the fact that `unsafe(..)` in attributes today only discharges obligations, what I'd point out is that we only very recently shipped unsafe attributes at all, so I don't assign much in the way of precedent value.  It doesn't surprise me one bit that we started by only using the syntax one way, and I don't think that precludes us from using it in another way, in a different context, if it's appropriate, well motivated, and sufficiently unambiguous.

One reason that I proposed this change is that putting the `unsafe` keyword somewhere arbitrary in an argument list -- as with  `define(Thing, attributes(..), unsafe)` -- just doesn't speak to me strongly enough that we're creating an unsafe obligation.  Everywhere else that we do this, it comes in front and is paired closely with the name of what we're defining that carries that obligation.  So probably I'd like to see us address that somehow.  I had thought too about `define(unsafe Thing)`, and that'd be OK if others don't like the proposed pun, but the arguments above cause me to still lean more toward the symmetric answer.

RalfJung:
We're already using `unsafe` both to mark the point of introducing unsafe obligations and of discharging them. This is a point that often trips people up (it is on the short list of things I would change about Rust 1.0 if I could). For the existing cases at least there's  fairly short list of you have to learn to know which is which. The proposal to use `unsafe(Trait)` for both cases means that in the future one will have to know exactly how all attributes use `unsafe` to understand which of them introduces vs discharges obligations -- that does not seem great.

programmerjake:
it just occurred to me that we may want to mark derive macro helper attributes `unsafe` instead of (or in addition to) the derive macro:
```rust
#[proc_macro_derive(MyDebug, attributes(unsafe(debug_deref)))]
fn my_debug ...
```

```rust
// no `unsafe(MyDebug)` here since it's only unsafe
// when you need to deref pointers, which not all structs have.
#[derive(MyDebug)]
pub struct FooPtr<T> {
    // SAFETY: ptr points to a valid memory chunk allocated with `malloc`
    #[unsafe(debug_deref)]
    ptr: NonNull<T>,
    bar: usize,
}
```

traviscross:
Excellent point.  We'll take that up with the nomination as well.

text/3715-unsafe-derives-and-attrs.md
traviscross: Josh and I talked, but for others here, by wanting to remove the predicate, I don't mean to suggest by this language that it is the default or is required.

If we say, "you can write X here," I don't think that implies that you must write "X" at all.  I read it instead as, "if you're going to write an X, which is entirely your option, you can write it here."

What I don't want to suggest is that writing such safety comments is unusual, and the fussy language referencing extrinsic requirements for doing this makes it feel that way to me.

I proposed wording along the lines of "when adding a safety comment, you can write it here" that both Josh and I liked, and that I think he's planning to update the text to reflect.

tmandry:
The motivation section is really short. Even though this seems like an "obvious" extension, I would appreciate it if @joshtriplett and others could add concrete examples of derives that would use this. Until then further bikeshedding on the spelling seems premature.

joshtriplett:
@programmerjake That exact proposal is in the future possibilities section:
> We could provide a syntax to declare specific helper attributes of a derive as unsafe, without declaring the entire derive unsafe.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,142 @@
...
+
+Some rationale for putting it on the inside:
+- This encourages minimizing the scope of the `unsafe`, isolating it to a
+  single derive macro.
+- This allows writing all derive macros to invoke within a single
+  `#[derive(...)]`, if desired. Putting the `unsafe` on the outside requires
+  separate `derive`s for safe and unsafe derives, and potentially multiple in
+  the (unusual) case where the derives care about ordering.
+- This makes it easy to attach `SAFETY` comments to each individual derive
+  macro.
```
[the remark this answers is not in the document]
joshtriplett: @kennytm That's true, though we could fix that in rustfmt in the future.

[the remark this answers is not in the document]
joshtriplett: I've added a note about this.

joshtriplett:
@moulins
> > it just occurred to me that we may want to mark derive macro helper attributes `unsafe` instead of (or in addition to) the derive macro:
> > [...snip]
> 
> I'd go even further: the same attribute could be either safe or `unsafe`, depending on other attributes of the same macro, so there should be a way to accept both `#[unsafe(debug_deref)]` and `#[debug_deref]`.

We could certainly do that, but I think it's reasonable for that to be future work. *This* proposal introduces always-unsafe derives and always-unsafe attributes. In the future, we could have an RFC for sometimes-unsafe derives and sometimes-unsafe attributes, which would require additional definition and documentation.

joshtriplett:
@tmandry https://github.com/facet-rs/facet/issues/80 is a good example of rationale for this.

nikomatsakis:
@rfcbot reviewed 

I am happy with the syntax and I think you should write `unsafe(Trait)` in both the def and use site. 

I like the idea of unsafe attributes.

traviscross:
@rfcbot resolve pick-def-site-name

We discussed this in the lang call today and aligned around the def-site syntax that matches the use-site syntax.

We also talked about expanding on the motivation section of the document, so as to document more of this for posterity, and @joshtriplett is working on making this update.

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,177 @@
...
+have a comment. That could be changed in a future style edition, but this RFC
+is not making or advocating any style proposals.)
+
+### Helper attributes
+
+A `derive` macro can have helper attributes. You can use the following syntax
+to declare a helper attribute as `unsafe`:
+
+```rust
+#[proc_macro_derive(MyDeriveMacro(, attributes(unsafe(dangerous_helper_attr))]
```
traviscross, line 107:
```suggestion
#[proc_macro_derive(MyDeriveMacro, attributes(unsafe(dangerous_helper_attr)))]
```

traviscross:
In the meeting today, we also aligned around including `unsafe` attributes in the normative part of this RFC rather than as future work, and I see now that @joshtriplett has made that change to the document.

tmandry:
I still haven't seen a worked example of where and how this will be used. The conclusion of https://github.com/facet-rs/facet/issues/80 was that it was not a good candidate for an unsafe derive.

`Send` and `Sync` were brought up and they seem like decent candidates. (Modulo the detail that they if they were in core they would be built in derives and not proc macros.. but if this concept was useful in core it would probably be useful elsewhere.)

The problem I see is that the ordinary derive logic doesn't make sense for `Send` and `Sync`. If I have

```rust
struct Ref<T>(*const T);
```

Then I most likely want my `Send` impl to look like this:

```rust
unsafe impl<T: Sync> Send for Ref<T> {}
```

We don't have a precedent for a built-in derive that works like this. If we go by the way those derives work today, `#[derive(unsafe(Send))]` would give you

```rust
unsafe impl<T: Send> Send for Ref<T> {}
```

which would be subtly wrong (albeit in a conservative direction), and worse, hidden from the user.

Overall I'm concerned by combination of the following:

* Broad, whole-type-behavior safety invariants implied by the use of a `#[derive(unsafe(...)]`
* The inherent subtlety of unsafe impls, which might require more detailed knowledge of a type's behavior than a derive macro can have
* Opaqueness of derive expansions masking potential problems with the above
* The limited syntax-level knowledge of macros and their inability to see through other types

I think many of these challenges can be overcome, but we should have use cases in mind to guide this and related features. That way we will end up shipping a good user experience (at least for those use cases) instead of a feature that's almost-but-not-quite useful.

tmandry:
@rfcbot concern needs fleshed out use case

joshtriplett:
@tmandry A few examples of concrete traits that definitely need unsafe derives *and* don't have any complexity around bounds:

- `TrustedLen` and `TrustedStep`
- `DerefPure`
- [`pyo3::marker::Ungil`](https://docs.rs/pyo3/0.24.2/pyo3/marker/trait.Ungil.html)
- A hypothetical `derive` for [`bytes::Buf`](https://docs.rs/bytes/1.10.1/bytes/buf/trait.Buf.html) or similar. (For cases where your type has fields or trivial expressions for current slice, offset, etc.)

text/3715-unsafe-derives-and-attrs.md
```diff
@@ -0,0 +1,188 @@
...
+that fails if some obligation is not met), or an `unsafe` `derive` macro that
+puts the safety obligation on the invoker of the `derive`.
+
+Some examples of potential unsafe derives for `unsafe trait`s:
+
+- `TrustedLen` and `DerefPure` in the standard library. (Currently unstable,
+  but such a derive would be useful when they're stable, serving the function
+  of an `unsafe impl`.)
+- `pyo3::marker::Ungil` in `pyo3`, in place of the current handling of a
+  blanket impl for any `Send` type.
```
kennytm, line 35: If these are the only motivation examples, I don't get why these *marker* traits `TrustedLen`, `TrustedStep`, `DerefPure`, `Ungil` *require* using the `#[derive]` mechanism over manual `unsafe impl`. 



For the first three the only advantages are not having to explicitly write the `where` bounds, but they have the `Iterator` or `Deref` supertraits which cannot be derived so you gotta repeat those bounds anyway.

joshtriplett: Literally any `unsafe trait` that one might wish to `derive` is a potential example. These were a few samples. If you have other examples of `unsafe trait` that you'd prefer to have cited instead of or in addition to these, I'd be happy to add them.

kennytm: > Literally any `unsafe trait` that one might wish to `derive` is a potential example.

Normally if we wish to `derive` something it is because we know that it is possible to fill out the `impl` programmatically with the type's definition alone, thus saving effort to produce the boilerplate impl. 

If all the derive doing is just implementing a marker trait without anything extra there's certainly no point *from the supply-side* anyone wish to write a proc-macro for it. At least it should be derivable together with its base trait like `#[derive(Clone, Copy)]` i.e. `#[derive(Iterator, unsafe(TrustedLen))]`, but you can't & *can't* derive an Iterator.

Another reason we wish to `derive` a marker trait is to insert some compile-time checks, such as `zerocopy`'s macros. But `zerocopy`'s derive-macro checks at the same proved the conditions required by the `unsafe trait` in the first place, making those traits *safe to derive* even if *unsafe to manually implement*. 

For `libstd` the public, documented, non-sealed unsafe traits are:

1. Traits with no obvious derivable impl:
    * GlobalAlloc
    * Allocator
    * Searcher
    * ReverseSearcher
2. Marker traits
    * TrustedStep
    * TrustedLen
    * DerefPure
    * PinCoerceUnsized
3. TransmuteFrom
4. CloneToUninit

Perhaps `CloneToUninit` is the one most compatible with `#[derive]`:

```rust
#[derive(CloneToUninit)]
struct Packet<Tail: ?Sized> {
    header: Header,
    tail: Tail,
}
```

but the actual impl can satisfy `impl CloneToUninit`'s safety condition without user's intervention, meaning `#[derive(CloneToUninit)]` itself is actually safe.

```rust
unsafe impl<Tail: CloneToUninit + ?Sized> CloneToUninit for Packet<Tail> {
    unsafe fn clone_to_uninit(&self, dst: *mut u8) {
        self.header.clone_to_uninit(dst.add(offset_of_val!(self, header)));
        self.tail.clone_to_uninit(dst.add(offset_of_val!(self, tail)));
    }
}
```

So no I don't have any other examples to support this RFC, i.e. I'm still on the side of the concern "needs fleshed out use case".

PR #3747, opened 2024-12-19
unsized params in traits (follow up)

ionicmc-rs:
Original PR: #3745

RFC proposing a future compatibility lint for checking `T: Sized` in required trait method (they didn't before)

[Rendered](https://github.com/ionicmc-rs/rfcs/blob/RFC_unsized_params_in_traits/text/3696-unsized-params-in-traits.md)


ionicmc-rs:
Closing this PR because i updated the original

PR #3758, opened 2025-01-07
Use rustbot config to configure rendered links

Urgau:
This PR adds the new config related to the rendered links.

cc https://github.com/rust-lang/triagebot/pull/1868
cc @ehuss


ehuss (approved):

Urgau:
(just for testing that rustbot still works)

@rustbot label not-rfc

PR #3767, opened 2025-01-31
Book should say what it is about

hkBst:
This is my best guess at what this book is, but I could be wrong, since this does not seem to be stated anywhere...

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 [Rust RFCs]: #rust-rfcs
 
+This book is a collection of all RFCs that have been accepted.
+
 The "RFC" (request for comments) process is intended to provide a consistent
 and controlled path for changes to Rust (such as new features) so that all
 stakeholders can be confident about the direction of the project.


README.md
```diff
@@ -2,6 +2,8 @@
 
 [Rust RFCs]: #rust-rfcs
 
+This book is a collection of all RFCs that have been accepted.
+
 The "RFC" (request for comments) process is intended to provide a consistent
 and controlled path for changes to Rust (such as new features) so that all
 stakeholders can be confident about the direction of the project.
```
traviscross, line 9: Probably the bit being added would make more sense as the second paragraph, after the nature of an RFC has been introduced.

hkBst: No, I don't think that is right. The very first thing this book should say is what it is about at the highest level possible. Then the details of that thing can be explained later.

Also to be able to understand """The "RFC" process ... the direction of the project.""" it helps greatly if you already know what this book is about.

hkBst:
I see there are some confused emojis, so I will try to explain a bit more. Currently I am unsure whether this book is a collection of RFCs that:
 * have merely been written
 * have also had some review for clarity and consistency
 * have also had some review for desirability of implementation
 * have also been implemented (I'm pretty sure it's not this one, but it's not stated anywhere AFAIK)
 * a combination of all previous categories

Does the book explain this somewhere and I missed it? Which of these options is correct?

hkBst:
[What the process is](https://rust-lang.github.io/rfcs/introduction.html#what-the-process-is) says:
"""
In short, to get a major feature added to Rust, one must first get the RFC merged into the RFC repository as a markdown file. At that point the RFC is “active” and may be implemented with the goal of eventual inclusion into Rust.
"""

Urgau:
We discussed this pull request during @rust-lang/internal-sites triage of non-RFC PRs in [#t-internal-sites > Stale non-RFC PRs @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/538071-t-internal-sites/topic/Stale.20non-RFC.20PRs/near/546011978).

We consider that adding the proposed sentence to the README is going to be confusing for visitors coming to the repository (where the new sentence doesn't make sense).

It is unfortunate that the README is shared between the book and the repository.

Thanks you for attempting to improve the book.

*Feel free to open a new PR (or re-open this PR) with a better proposal.*

hkBst:
@Urgau thanks for discussing this issue; perhaps I'll see if I can find a better place for this information.

PR #3768, opened 2025-02-01
RFC: Add safe blocks

AverseABFun:
This RFC is a proposal for so-called "safe blocks" that would provide the opposite effect of unsafe blocks.

### DIFF
--- a/text/0000-safe-blocks.md
+++ b/text/0000-safe-blocks.md
@@ -1,0 +1,156 @@
+- Feature Name: safe_blocks
+- Start Date: 2025-01-31
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This is a RFC to add safe blocks to the language. Safe blocks would have the opposite effect of unsafe blocks and only be valid within unsafe blocks.
+
+This RFC also proposes adding a lint when you do something like `/* safe context */ unsafe { safe { /* code */ } }`, as this would be an example of an anti-pattern.
+
+# Motivation
+[motivation]: #motivation
+
+Adding safe blocks can be useful in many contexts, an example of which was shared by programmerjake on the Rust Internals forum:
+> That said, I think having some form of safe {} blocks is a great idea that I've been wanting for some time, mostly because it'll be useful in macros, e.g.:
+>
+> ```rust
+> macro_rules! inc {
+>    ($a:expr) => {
+>        unsafe {
+>            // not the best example...other macros much more
+>            // naturally have inputs in the middle of an `unsafe` block
+>            let mut a: u64 = safe { $a };
+>            asm!("inc {}", inlateout(reg) a);
+>            a
+>        }
+>    };
+> }
+> ```
+
+Another example, shown by me (AverseABFun) on the Rust Internals forum:
+
+> [snip] for example, I was writing a function like this(ignore the contents, it's not super important):
+>```rust
+>/// Parses a PC Screen Font into a [PCScreenFont].
+>pub fn parse_pc_screen_font(data: RawPCScreenFont) -> Result<PCScreenFont, crate::Error<'static>> {
+>    unsafe {
+>        let unitable: &[&str] = &[];
+>        let unistr = data.glyphs.byte_add(data.bytes_per_glyph as usize*data.num_glyphs as usize);
+>        /* Safe block start */
+>        for i in 0..(data.num_glyphs as usize) {
+>            let char = (*unistr)[i];
+>            /* Didn't finish writing code here lol */
+>        }
+>        /* Safe block end */
+>        
+>        /* Snip */
+>    }
+>}
+>```
+>
+>And I'd like to have the for loop be safe [snip], however having something like:
+>
+>```rust
+>/// Parses a PC Screen Font into a [PCScreenFont].
+>pub fn parse_pc_screen_font(data: RawPCScreenFont) -> Result<PCScreenFont, crate::Error<'static>> {
+>    unsafe {
+>        let unitable: &[&str] = &[];
+>        let unistr = data.glyphs.byte_add(data.bytes_per_glyph as usize*data.num_glyphs as usize);
+>    }
+>    for i in 0..(data.num_glyphs as usize) {
+>        let char = (*unistr)[i];
+>         /* Didn't finish writing code here lol */
+>    }
+>    unsafe {
+>        /* Snip */
+>    }
+>}
+>```
+>
+>seems too obtrusive. Please let me know if I'm missing something, but here's an example code sample with my same example:
+>
+>```rust
+>/// Parses a PC Screen Font into a [PCScreenFont].
+>pub fn parse_pc_screen_font(data: RawPCScreenFont) -> Result<PCScreenFont, crate::Error<'static>> {
+>    unsafe {
+>        let unitable: &[&str] = &[];
+>        let unistr = data.glyphs.byte_add(data.bytes_per_glyph as usize*data.num_glyphs as usize);
+>        safe {
+>            for i in 0..(data.num_glyphs as usize) {
+>                let char = (*unistr)[i];
+>                /* Didn't finish writing code here lol */
+>            }
+>        }
+>        
+>        /* Snip */
+>    }
+>}
+>```
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+So they're called safe blocks, and are only allowed within unsafe blocks. The idea is that you can mark parts of code as being not allowed to have unsafe code. An example might be if you were writing a macro where it had to be unsafe, however you wanted the inputs to be safe, then you could wrap the input in `safe { }` so that it would be interpreted as safe code. This generally makes it more clear that code is safe than if you ended an unsafe block before the code and started it after, however the same thing applies the other way where it's less clear that code is unsafe again after the safe block. The compiler also gives a lint if you do something like `unsafe { safe { /* code */ } }`, as there's not supposed to be any logical reason to do this since the inside of safe blocks and outside of unsafe blocks are supposed to be identical. This can of course be disabled, though.
+
+An example might be:
+```rust
+macro_rules! inc {
+    ($a:expr) => {
+        unsafe {
+            // not the best example...other macros much more
+            // naturally have inputs in the middle of an `unsafe` block
+            let mut a: u64 = safe { $a };
+            asm!("inc {}", inlateout(reg) a);
+            a
+        }
+    };
+}
+```
+As you can see, the code needs to be unsafe and putting the safe code out of the unsafe block decreases readability, and so safe blocks would likely be the best choice in this scenario.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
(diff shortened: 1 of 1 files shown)


text/3768-safe-blocks.md
```diff
@@ -0,0 +1,156 @@
...
+        }
+    };
+}
+```
+As you can see, the code needs to be unsafe and putting the safe code out of the unsafe block decreases readability, and so safe blocks would likely be the best choice in this scenario.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+I'm not super familar with the internals of rustc, but the parts that would need to be modified are probably mostly the parts involving unsafe blocks. This would directly impact those parts because of the nature of the feature. As mentioned in the summary, this also adds a lint(I don't have any ideas for the name right now; please let me know any ideas) that would trigger upon the anti-pattern of `unsafe { safe { /* code */ } }`(or vice versa of course). I'll try to provide details about how it would be implemented, but I don't know how unsafe blocks are implemented to begin with, however based on the messages it appears as though this would be implemented in the AST passes section of rustc(let me know if this is completely incorrect or if I stated this incorrectly).
```
programmerjake, line 117: the general idea of the reference-level explanation is that, rather than describing how it should be implemented in rustc, you instead describe exactly how the syntax works since the idea is this section of the RFC gets copied to the [Rust Reference](https://doc.rust-lang.org/reference/), you could adapt the text from the reference section on [`unsafe` blocks](https://doc.rust-lang.org/reference/expressions/block-expr.html#unsafe-blocks)

AverseABFun: Ohh, thank you! I'll fix that now. Should I keep my paragraph on rustc or no?

AverseABFun: I'm assuming no.

AverseABFun: Fixed!

kennytm:
As a (not very good) prior art there were the `pop_unsafe!` and `push_unsafe!` macros back in 1.3.0 to temporarily inhibit the unsafe block a.k.a. "unsafety hygiene" (rust-lang/rust#27215).

```rust
push_unsafe!({
    expr_1;
    pop_unsafe!({
        expr_2;
    })
})

// similar to this RFC's
unsafe {
    expr_1;
    safe {
        expr_2;
    }
}
```

It is a very ugly hack and the surface syntax is quickly removed in 1.5.0 (rust-lang/rust#28980), but the remnants are only purged 6 years later in 1.55.0 (rust-lang/rust#85421).

---

As for the RFC itself

1. It would be very unfortunate if we need to turn `safe` into a keyword (see also #2388). [Its usage is pretty common](https://github.com/search?q=lang%3ARust%20%2F(let%7Cfn)%20safe%5Cb%2F&type=code). 
2. As shown above, the push/pop-unsafe implementation treats it like a stack, so that an outer unsafe block can propagate into the macro ([godbolt][]):

    ```rust
    #![feature(pushpop_unsafe)]
    #![allow(dead_code)]
    
    unsafe fn do_unsafe_thing(_: u8) {}
    
    macro_rules! do_something {
        ($e:expr) => {
            push_unsafe!({
                do_unsafe_thing(0);
                pop_unsafe!($e)
            })
        }
    }
    
    fn main() {
        unsafe {
            do_something! {
                do_unsafe_thing(1) // no error here
            }
        }
        
        do_something! {
            do_unsafe_thing(2) // E0133 here
        }
    }
    ```

    The RFC does not specify the nesting behavior. If we rewrite the macro to use a `safe` block, will the `do_unsafe_thing(1)` line still compile fine?

    ```rust
    macro_rules! do_something {
        ($e:expr) => {
            unsafe {
                do_unsafe_thing(0);
                k#safe { $e }
            }
        }
    }
    ```

[godbolt]: https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:2,endLineNumber:25,positionColumn:2,positionLineNumber:25,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:'%23!!%5Bfeature(pushpop_unsafe)%5D%0A%23!!%5Ballow(dead_code)%5D%0A%0Aunsafe+fn+do_unsafe_thing(_:+u8)+%7B%7D%0A%0Amacro_rules!!+do_something+%7B%0A++++($e:expr)+%3D%3E+%7B%0A++++++++push_unsafe!!(%7B%0A++++++++++++do_unsafe_thing(0)%3B%0A++++++++++++pop_unsafe!!($e)%0A++++++++%7D)%0A++++%7D%0A%7D%0A%0Afn+main()+%7B%0A++++unsafe+%7B%0A++++++++do_something!!+%7B%0A++++++++++++do_unsafe_thing(1)+//+no+error+here%0A++++++++%7D%0A++++%7D%0A++++%0A++++do_something!!+%7B%0A++++++++do_unsafe_thing(2)+//+E0133+here%0A++++%7D%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'rustc+1.4.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.4.0+(Compiler+%231)',t:'0'),(h:compiler,i:(compiler:r140,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'',overrides:!((name:env,values:!((name:RUSTC_BOOTSTRAP_KEY,value:'01:09:53')))),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.4.0+(Editor+%231)',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

AverseABFun:
Hmm, for the nesting behavior I don't think it would as the safe block is
the block it's contained within and thus it would be assumed to be safe
code. For the keyword thing, safe would be the most natural name but maybe
it could be something like "resafe" as to prevent conflicts?

programmerjake:
if `safe` is a context-dependent keyword so it's only a keyword when immediately followed by `{`, it shouldn't really conflict with existing code unless that code is really weird. this is kinda like `union`.

if there's enough code that conflicts, `safe` can become a context-dependent keyword in the next edition and we'll just use `k#safe` in edition <= 2024.

tczajka:
Every single motivating example can be solved better by reducing the overly large scope of the `unsafe` block. Extra syntax to undo `unsafe` feels unnecessary to me.

kennytm:
@programmerjake 

> if `safe` is a context-dependent keyword so it's only a keyword when immediately followed by `{`, it shouldn't really conflict with existing code unless that code is really weird. this is kinda like `union`.

First of all, `if safe { … } `.

The reason `safe` cannot be a contextual keyword is the same as that of `try`, please checkout #2388. Contextual keyword works for union declaration because `union IDENTIFIER` is otherwise an invalid sequence everywhere else. But `safe { … }` is already a valid expression in itself.

If this RFC has to use the syntax `IDENTIFIER { … }` there is no choice but to make that identifier a ful keyword the next edition. Or you could pick a different syntax such as `safe! { … }`.

burdges:
I'd think some localized flavor of `#[forbid(unsafe_code)]` makes more sense than the word `safe`, no?

Could `#[forbid(unsafe_code)] ` appear before a statement or code block?

zachs18:
Hmm, yes, `#[forbid(unsafe_code)]` or `#[deny(unsafe_code)]` _can_ be put on a (normal) block inside an `unsafe` block, but currently it doesn't appear to actually fire on unsafe operations in that block. Presumably the `unsafe_code` lint only fires on `unsafe` *blocks* (and unsafe attributes like `#[no_mangle]`), not unsafe *operations*: [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1129e4995bb561039ff7b3f056864112).

Perhaps the `unsafe_code` lint could be updated to also apply to unsafe operations if the lint level is raised inside an `unsafe` block?

programmerjake:
> I'd think some localized flavor of `#[forbid(unsafe_code)]` makes more sense than the word `safe`, no?

I disagree for 2 reasons:
1. forbidding unsafe code would prevent you from having an `unsafe` block inside a macro's argument, `safe` blocks allow you to do that.
2. I don't think rust should rely on lints for safety, since iirc they can be disabled with cap-lints and are disabled by default in dependencies.

text/3768-safe-blocks.md
```diff
@@ -0,0 +1,184 @@
...
+    ($a:expr) => {
+        let mut a: u64 = $a;
+        unsafe {
+            asm!("inc {}", inlateout(reg) a);
+            a
+        }
+    };
+}
+```
+However, this decreases readability in my opinion, however this is a matter of opinion. This design also seems the most natural when you are learning rust as it is very similar to unsafe blocks. If this isn't done, there isn't much direct impact, however it does decrease readability in certain circumstances. As far as I know, this couldn't be implemented in a macro, but I could be wrong. As mentioned, this generally increases readability.
```
JarredAllen, line 165: Another alternative worth mentioning that I think would serve the same purpose: allowing finer-grained control of where `unsafe` is applied (e.g. though [single function call unsafe](https://internals.rust-lang.org/t/pre-rfc-single-function-call-unsafe/22343)), such that you won't need a `safe` block anymore.

Which one is better is a matter of preference. I personally like having only one unsafe operation per `unsafe` block, so I can explicitly document on the block why the preconditions are met, so I'd prefer having finer-grained unsafe blocks, but I know other people like to include the entire code incorporating a safety invariant in one `unsafe` block that details the invariant requirements. And nothing prevents Rust from having both, though the second one would have diminishing marginal returns.

I'm not the arbiter of which approach Rust should take, but I think it'd be worth mentioning in the alternatives section, at least.

ssokolow:
> I disagree for 2 reasons:
> 
>     1. forbidding unsafe code would prevent you from having an `unsafe` block inside a macro's argument, `safe` blocks allow you to do that.

So use `#[deny(unsafe_code)]` instead, since it can be overridden with `#[allow(unsafe_code)]`?

ssokolow:
Overall, I don't get the feeling that this proposal justifies the added complexity of having a whole new keyword and mechanism in the language.

First, as [tczajka](https://github.com/tczajka) said, all the motivating examples have un-idiomatically broad `unsafe` blocks and I don't think we should be implicitly encouraging developers to do that by allowing unsafe→safe→unsafe→safe→... nesting.

Second, this feels like the kind of thing where the main place that it'd be relevant is when a macro is subverting how `unsafe` interacts with scoping in the regular language (i.e. calling a function from inside `unsafe` doesn't grant `unsafe` to its contents but substituting some AST nodes into `unsafe` in a macro does), and I don't think we should be so readily be altering the shape of the language at large to account for situations that only crop up because of quirks of what macros are.

(I'm reminded of what a brain-bender it was to realize that, though there was no trait for me to implement to extend [Keats/validator](https://github.com/Keats/validator) in the way I wanted, I could extend it just by defining my *own* extension trait because it being written using macros meant that `#[validate]` implemented duck-typed method resolution.)

zjp-CN:
I've opened a RFC called [safety tags](https://github.com/rust-lang/rfcs/pull/3842) for semantic granularity instead of visual granularity on unsafe code.

Escarcega1989:
> if `safe` is a context-dependent keyword so it's only a keyword when immediately followed by `{`, it shouldn't really conflict with existing code unless that code is really weird. this is kinda like `union`.
> 
> if there's enough code that conflicts, `safe` can become a context-dependent keyword in the next edition and we'll just use `k#safe` in edition <= 2024.

,:;::CUURRENTLY UNDER REVIEW .
1WEEK § WORTH OF TIME WITH TIME. 
NOT "Gas"
DONT PANIC😁😁!!!
TOPS 4 `(in time)
 2 "Issues "!
/` Apologize for the inconvenience§
∆§
THNXS

%LME

PR #3770, opened 2025-02-10
#[doc(consts)]

lolbinarycat:


text/0000-doc-consts.md
```diff
@@ -0,0 +1,134 @@
...
+Hovever, [showing a constant as written may leak implementation details], 
+and in some cases, there is no possible value 
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `#[doc(consts)]` attribute can be placed on any item to control how contained constant expressions are displayed in rustdoc-generated documentation.
+
+* `#[doc(consts = "fold")]` will show them in their fully-evaluated state.
```
lebensterben, line 27: how about `"eval"` for fully evaluated item and "`as-is`" for item as written?

lolbinarycat: "constant folding" is a well established term used for this exact process, and "eval" is much more general (future display modes would likely also involve evaluating the expression).

on the other hand, i don't feel strongly about `"expr"`, but my second choice would probably be `"verbatim"`.

tgross35: Doesn't Rust ever refer to CTFE as constant folding? The docs use evaluation https://doc.rust-lang.org/reference/const_eval.html, and to me "const folding" is an identify+transform optimization pass (as opposed to compile-time evaluation that must unconditionally be run).

text/0000-doc-consts.md
```diff
@@ -0,0 +1,134 @@
...
+
+Introduce a `#[doc(consts = ...)]` attribute controlling how constant expressions are rendered by rustdoc.
+
+# Motivation
+[motivation]: #motivation
+
+Different crates and items have conflicting requirements for their constants.
+For some, [the exact value of a constant is platform dependant](https://internals.rust-lang.org/t/pre-rfc-doc-consts-attribute/21987/9).
+For others, [constant folding obsurces the meaning of values](https://github.com/rust-lang/rust/issues/128347).
+Hovever, [showing a constant as written may leak implementation details],
```
RustyYato, line 18: is there a link missing here?

lolbinarycat: yes, it's one of the linked issues in https://github.com/rust-lang/rust/pull/99688 (i'll go dig through it in a bit)

fmease (commented):
Thinking about it, we actually want to have a similar attribute for *types*:

1. A subgroup of users would like rustdoc to normalize types. Type normalization almost became the default (https://github.com/rust-lang/rust/pull/77467) but due to limitations of the current (but not next-gen) trait solver (namely: fatal overflows), it had to be gated behind the `-Znormalize-docs` where this behavior currently resides. Also, rustdoc is notoriously bad at properly tracking binders (think higher-ranked lifetimes). It's gotten better but it's not sufficient (cc https://github.com/rust-lang/rust/pull/108503).
2. A subgroup of users would like public type aliases to remain unexpanded and private ones to be folded away. This is what rustdoc currently does for items of the local crate. Infamously, this does not extend to inlined cross-crate re-exports (https://github.com/rust-lang/rust/issues/21903) as they use a different IR (middle IR vs HIR). This has created the multi-year effort of representing type aliases in the "type/trait system IR" (aka middle::ty). See https://github.com/rust-lang/rust/issues/112792.

However, \[1\] (type normalization) is **inherently incompatible** with \[2\] (public type alias preservation). I'm currently driving \[2\] and will block any attempt at stabilizing \[1\] :P (not that there was any push to do so). This screams "we'd like to have configuration".

text/0000-doc-consts.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `#[doc(normalize::consts)]` attribute can be placed on any item to control how contained constant expressions are displayed in rustdoc-generated documentation.
+
+* `#[doc(normalize::consts = "fold")]` will show them in their fully-evaluated state.
+* `#[doc(normalize::consts = "expr")]` will show them as-written.
+* `#[doc(normalize::consts = "hide")]` will cause constant expressions to be replaced with `_` or not shown at all.
```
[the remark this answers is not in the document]
kennytm: > I simply believe that the nested function call syntax is less readable, and like you said, rustdoc has complete control over the `doc(` attribute syntax, so why not use the more readable syntax?

I just don't see how `#[doc(normalize::consts = "x")]` is any more readable then `#[doc(normalize_consts = "x")]`. Even `#[doc(normalize(consts = "x"))]` is not really that bad as the nesting must be limited at level 2, unlike `#![doc(test(attr(deny(dead_code))))]`.

The slight advantage of the nested syntax is allowing the two normalization instructions grouped into a single attribute, `#[doc(normalize(consts = "x", types = "y"))]`, but breaking it into 2 lines `#[doc(normalize_consts = "x")] #[doc(normalize_types = "y")]` isn't really a big disadvantage IMO.

> I think `instruction_set` actually acts as pretty good precedence for this, after all, it could've easily been written as `#[instruction_set(arm(a32))]`.

#2867 starts out as `#[instruction_set(a32)]`, the `arm` namespace was [added on later](https://github.com/rust-lang/rfcs/pull/2867#issuecomment-618623702) for disambiguation. 

[The original suggestion](https://github.com/rust-lang/rfcs/pull/2867#issuecomment-593064378) was `#[instruction_set(arm_a32)]` (not the "real name", thus hurts searchability) and then `#[instruction_set(arm, a32)]` (does not suggest a hierarchical relationship). 

None of the arguments are relevant to `normalize::consts` / `normalize::types`.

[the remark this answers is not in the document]
fmease: > added a `normalize::` prefix, what do you think about that?

Sorry for the late reply. Thanks for incorporating my suggestion/concern/point about types! At this stage of the process, I will refrain from thinking too much about syntax choices because I deem it a bit too early.

However I do agree with kennytm that `#[a(b::c = d)]` looks quite *particular* for lack of a better word due to the mixing of "legacy namespacing" `a(b)` and modern namespacing `a::b`. Drawing a comparison to the `diagnostic` *tool namespace* is slightly misguided I would say since it presumably would not actually leverage `rustc_resolve`'s `ToolMod` logic.

> I think I'll wait for a third party to weigh in before changing the proposed syntax again.

Anyhow, please don't feel the need to change the syntax again, it's too early to paint the shed, I'd like to focus on more pressing things first.

text/0000-doc-consts.md
fmease, line 1: I think one good way to make proper progress on this RFC and bring it out of its draft status, would be to schedule some sort of "deep dive session" where interested parties can come together and investigate whether we as rustdoc can *guarantee* to take this attribute into account in all cases from an implementation perspective or whether we need to downgrade it to a "*hint*".

Re. "cases", I'm referring to local vs. inlined cross-crate or more precisely lowering from HIR vs middle::ty IR / MIR (`ValTree`, `mir::Const`, `ty::Const`).

text/0000-doc-consts.md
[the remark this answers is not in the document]
lolbinarycat: I can understand the position that we should just commit to our decisions, even if they're slightly unoptimal.

~~What things do you think are more pressing and need addressing?~~

EDIT: saw your other comment

text/0000-doc-consts.md
fmease: As part of this, I'm also interested in talking about what "`fold`" and "`expr`" entail *precisely*, namely what guarantees we're willing to make if any.

1. For "`expr`", for example whether we want to promise preserving "lexical information" like number bases (e.g., `0xFF`), underscores (e.g., `1_000`) (some users really *love* to keep this behavior (in the local case)).
   * I don't remember if it's preserved in the HIR but even if it's not, then in *some* cases it's obtainable via `SourceMap`. IIRC rustdoc currently does look at the SourceMap to achieve this
   * Obviously for the cross-crate case, middle::ty IR / MIR doesn't contain these pieces of information (and it will never do), so we'd need to use "JS side files" or something like that etc etc
3. For "`fold`",
   * for example what to do if the const expr "is too generic" or if it diverges[^1] (hard error vs. keeping the unnormalized form around (and emit a lint warning?))
   * or whether we want to refrain from normalizing away "paths" (const aliases/projections is the proper term) referring to public const items similar to type aliases (which is based on rustdoc's IR "clean AST")). Whether that's even possible to do (custom normalization logic) if we have a MIR const or a ValTree assuming we want to throw out our "clean AST"-based normalization logic for good. Etc etc.

[^1]: This comes up with assoc consts that normally don't get eval'ed if not referenced.

PR #3772, opened 2025-02-13
RFC: Deprecate the per-build-target `edition` field in `Cargo.toml`

epage:
Deprecate `lib.edition`, etc in favor of only setting `package.edition`, removing the fields in the next Edition.

### DIFF
--- a/text/0000-build-target-edition.md
+++ b/text/0000-build-target-edition.md
@@ -1,0 +1,97 @@
+- Feature Name: (fill me in with a unique ident, `my_awesome_feature`)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Explain the proposal as if it was already included in the language and you were teaching it to another Rust programmer. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how Rust programmers should *think* about the feature, and how it should impact the way they use Rust. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing Rust programmers and new Rust programmers.
+- Discuss how this impacts the ability to read, understand, and maintain Rust code. Code is read and modified far more often than written; will the proposed feature make code easier to maintain?
+
+For implementation-oriented RFCs (e.g. for compiler internals), this section should focus on how compiler contributors should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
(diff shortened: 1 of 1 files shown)


clarfonthey:
I didn't even know this was an option, and I agree it's very confusing, especially considering how cargo itself is affected by editions.

I honestly would go further and propose working with the existing crates that use this feature to remove it, then disabling it for all editions in a future version. Making the presence of an edition field depend on edition feels weird.

Lokathor:
So the error case is if you set an edition >= X in your `package` and then also set any edition value at all in particular build targets?

text/3772-build-target-edition.md
```diff
@@ -0,0 +1,180 @@
...
+- In Edition 20XX+, an error will be reported when this field is present
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This makes testing macros more difficult as they are limited to either
+  - doctests
+  - creating packages just for the sake of defining tests
+
+# Rationale and alternatives
```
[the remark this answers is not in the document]
epage: This seems to run counter to our compatibility guarantees.  We take this approach to help people as we fix bugs that people accidentally relied on but this was intentional behavior supported in the manifest and documented as part of working with Editions.  In practice, this would mean that new versions of cargo would not be able to build older versions of crates like `linkme`, `nameof`, and `num-derive`.

text/3772-build-target-edition.md
```diff
@@ -0,0 +1,180 @@
...
+
+*(added a testing strategy which was previously left unspoken)*
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A non-`None` edition will be considered deprecated
+- A deprecation message will eventually be shown by Cargo
+  - Timing depends on if this will be blocked on having `[lints]` control over this or not
+- In Edition 20XX+, an error will be reported when this field is present
```
epage, line 155: @Lokathor  from https://github.com/rust-lang/rfcs/pull/3772#issuecomment-2657842887

> So the error case is if you set an edition >= X in your `package` and then also set any edition value at all in particular build targets?

epage: Yes. I will make that more clear.

epage: See 13702bb

text/3772-build-target-edition.md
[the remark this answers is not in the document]
epage: I added d8c992a3 which discusses one edition field affecting another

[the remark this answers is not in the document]
clarfonthey: That's very fair, and I guess my justification comes from the fact that the feature is extremely counter-intuitive and not used very often. However, from what you've mentioned, it *is* used a lot in the ecosystem, just not in very many crates, which means that at least half of my argument is invalid.

So, it's fair to prefer this over an edition boundary rather than removing it outright.

joshtriplett:
@rfcbot merge

ehuss:
I'm checking my box, though I do note that cross-edition testing, particularly of macros, can be challenging. However, it seems evident that it is extremely rare that anyone is doing that, and there are workarounds (such as using separate test packages), and the benefits seem to outweigh that drawback to me.

text/3772-build-target-edition.md
```diff
@@ -0,0 +1,199 @@
...
+
+edition = "2015"       # Deprecated, N/A for Edition 20XX+
+```
+
+...
+
+### The `edition` field
+
+The `edition` field defines the [Rust edition] the target will use. If not
+specified, it defaults to the [`edition` field][package-edition] for the
```
Rustin170506, line 111: Is this intended to be a hyperlink, or is it just meant to indicate 'package-edition'?

Rustin170506 (commented):
Thanks!

text/3772-build-target-edition.md
epage: This was copied out of the Cargo book and edited.  The link will be fixed when put back in the Cargo book.  I can fix it now if it matters.

Rustin170506: I don’t think it matters. I just got confused when I saw it in render mode.

PR #3775, opened 2025-02-19
Guarantee slice representation

cramertj:


text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,167 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+The layout in this RFC is already documented in
+[the Unsafe Code Guildelines Reference.](https://rust-lang.github.io/unsafe-code-guidelines/layout/pointers.html)
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* Should `&[T]` include a pointer when `T` is zero-sized?
```
programmerjake, line 156: `bitvec` specifically depends on `&[()]` containing both a pointer and length field: https://docs.rs/crate/bitvec/1.0.1/source/doc/ptr/BitSpan.md

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,167 @@
+- Feature Name: guaranteed_slice_repr
+- Start Date: 2025-02-18
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC guarantees the in-memory representation of slice and str references.
+Specifically, `&[T]` is guaranteed to have the same layout as:
```
programmerjake, line 10: this should probably state that it's not specifying the function call ABI, so `extern "C" fn(&[u8]) -> &[u8]` still doesn't have an ABI guaranteed to match passing that `repr(C)` struct by value

text/0000-guaranteed-slice-repr.md
programmerjake: considering that the latest version of `bitvec` has 55M downloads, I think we can consider this point resolved due to backwards compatibility.

cramertj: AFAICT `bitvec` is essentially assuming that slice is represented exactly as this RFC proposes, which they probably shouldn't be doing-- we haven't guaranteed this representation up until this point. It is a pretty reasonable assumption, though, so I'm sympathetic to the idea that breakages here would be frustrating.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,174 @@
...
+
+```rust
+#[repr(C)]
+struct Slice<T> {
+    data: *const T,
+    len: usize,
+}
+```
+
+The layout of `&str` is the same as that of `&[u8]`, and the layout of
```
compiler-errors, line 20: Perhaps it's worthwhile to generalize this by saying that `str` acts as if it is defined like `#[repr(transparent)] struct str([u8]);`?

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,174 @@
+- Feature Name: guaranteed_slice_repr
+- Start Date: 2025-02-18
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC guarantees the in-memory representation of slice and str references.
+Specifically, `&[T]` is guaranteed to have the same layout as:
```
compiler-errors, line 10: And `&mut [T]`?

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+  [on MSVC](https://godbolt.org/z/Y8ardrshY), though not
+  [on GCC](https://godbolt.org/z/s4v4xehnG) nor
+  [on Clang](https://godbolt.org/z/qsd1K5oGq). Future changes to guarantee a
+  different layout in the C++ standard (unlikely due to MSVC ABI stabilitiy
+  requirements) could preclude matching the layout with `&[T]`.
+
+* Unlike Rust, `std::span` allows the `data` pointer to be `nullptr`. One
+  possibile workaround for this would be to guarantee that `Option<&[T]>` uses
+  `data: std::ptr::null()` to represent the `None` case, making `std::span<T>`
+  equivalent to `Option<&[T]>` for non-zero-sized types.
```
zachs18, line 129: We would probably also need to guarantee that the length part of the pointer is 0 (or at least not uninitialized) when `Option<&[T]>` is `None`. Note that this is not currently the case, `None::<&[u8]>` is currently represented as `[nullptr, uninit]` (though this is not guaranteed).

text/0000-guaranteed-slice-repr.md
zachs18: When you say "we haven't guaranteed this representation" of a slice reference, are you referring to the in-memory _layout_, or just the idea that a slice reference holds a data pointer and a `usize` length?

While some of bitvec's docs do talk about "Rust strongly guarantee[ing] the in-memory layout of a reference to a slice"[^3] (which is incorrect), AFAICT (at least the latest version of) bitvec's actual _code_ never transmutes or assumes the *layout* of slice pointers[^2]; it always uses `std::ptr::slice_from_raw_parts` to construct pointers and pointer casts to access the data pointer and length.

I think it would be a significant breaking change if slice pointers stopped having a data pointer and a length, e.g. if `std::ptr::slice_from_raw_parts(ptr, len).cast()` starts returning a pointer with a different address than was passed in, or if `std::ptr::slice_from_raw_parts(ptr, len).len()` gave a different length than was passed in, for *any* slice element type. While there could maybe be some language-lawyering to say something like "`&[T]` is not guaranteed to keep the same data pointer address when moved/copied if `T` is a ZST, but `*const [T]` is" to allow some optimization, I don't think that would be worth it without also changing the layout[^1].

[^1]: And we have stable docs that  `&T` and `*const T` have the same layout ([src](https://doc.rust-lang.org/reference/type-layout.html#pointers-and-references-layout)), so any change to `&[T]`'s layout would also have to be reflected in `*const [T]` unless we go back on those guarantees.

[^2]: "Values of this type  are binary-incompatible with slice pointers. Transmutation of these values into any other type will result in an incorrect program, and permit the program to begin illegal or undefined behaviors. This type may never be manipulated in any way by user code outside of the APIs it offers to this `bitvec`; it certainly may not be seen or observed by other crates." docs for `BitSpan`, `bitvec`'s "fake" `*const [bit]` type https://docs.rs/crate/bitvec/1.0.1/source/doc/ptr/BitSpan.md

[^3]: https://github.com/ferrilab/ferrilab/blob/main/guide/bitvec/pointer-encoding.md

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Note: prior to this RFC, the type above is not even properly `repr(C)` since the
+size and alignment of slices were not guaranteed. However, the Rust compiler
+accepts `repr(C)` declaration above without warning.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Slices are represented with a pointer and length pair. Their in-memory layout is
```
zachs18, line 47:
```suggestion
Pointers to slices are represented with a pointer and length pair. Their in-memory layout is
```

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+
+```rust
+#[repr(C)]
+struct Slice<T> {
+    data: *const T,
+    len: usize,
+}
+```
+
+The precise ABI of slices is not guaranteed, so `&[T]` may not be passed by-value
```
zachs18, line 58:
```suggestion
The precise ABI of pointers to slices is not guaranteed, so `&[T]` may not be passed by-value
```

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+or returned by-value from an `extern "C" fn`.
+
+The validity requirements for the in-memory slice representation are the same
+as [those documented on `std::slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html).
+Namely:
+
+* `data` must be non-null, valid for reads for `len * mem::size_of::<T>()` many bytes,
+  and it must be properly aligned. This means in particular:
```
zachs18, line 66:
```suggestion
The validity requirements for the in-memory representation of raw slice pointers are simply the validity requirements of the fields (i.e. initialized).

The validity requirements for the in-memory representation of slice references are the same
as [those documented on `std::slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) for shared slice references, and [those documented on `std::slice::from_raw_parts_mut`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html) for unique slice references.
Namely:

* `data` must be non-null, valid for reads (shared reference) or writes (mutable reference) for `len * mem::size_of::<T>()` many bytes,
  and it must be properly aligned. This means in particular:
```

Raw pointers to slices don't need to necessarily point to anything, and shared and unique references have different (but analogous) reqirements.

-----

Also: We might not want to say "validity requirements" here, as the "validity requirements" of references, and the concept of "validity requirements" in general, are not necessarily settled: https://github.com/rust-lang/unsafe-code-guidelines/issues/412 https://github.com/rust-lang/unsafe-code-guidelines/issues/539

zachs18 (commented):
This should probably clearly distinguish between "slices" and "pointers/references to slices", as they are different types (and the layout of slices themselves is [already defined](https://github.com/rust-lang/unsafe-code-guidelines/issues/539)).

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
+- Feature Name: guaranteed_slice_repr
+- Start Date: 2025-02-18
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC guarantees the in-memory representation of slice and str references.
+Specifically, `&[T]` and `&mut [T]` are guaranteed to have the same layout as:
```
kennytm, line 10: do these apply to `*const [T]` and `*mut [T]`?

zachs18: We have stable docs that "pointers and references have the same layout" ([src](https://doc.rust-lang.org/reference/type-layout.html#r-layout.pointer)), so I think it must.

text/0000-guaranteed-slice-repr.md
dlight, line 58: Wny not go ahead and define this ABI as well?

workingjubilee: Our ABI for this type is not the ABI of the equivalent C struct.

workingjubilee: In particular, its return ABI falls outside what can be obtained via C code on some platforms.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+```rust
+#[repr(C)]
+struct HasString {
+    string: &'static str,
+}
+```
+
+Note: prior to this RFC, the type above is not even properly `repr(C)` since the
+size and alignment of slices were not guaranteed. However, the Rust compiler
+accepts `repr(C)` declaration above without warning.
```
SabrinaJewson, line 42:
```suggestion
accepts the `repr(C)` declaration above without warning.
```

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+* `data` must point to `len` consecutive properly initialized values of type `T`.
+
+* The total size `len * mem::size_of::<T>()` of the slice must be no larger than `isize::MAX`,
+  and adding that size to `data` must not "wrap around" the address space.
+  See the safety documentation of [`pointer::offset`].
+
+## `str`
+
+The layout of `&str` is the same as that of `&[u8]`, and the layout of
+&mut str` is the same as that of `&mut [u8]`. More generally, `str` behaves like
```
SabrinaJewson, line 85:
```suggestion
`&mut str` is the same as that of `&mut [u8]`. More generally, `str` behaves like
```

text/0000-guaranteed-slice-repr.md
p-avital: Given that `&()` is indeed represented as a pointer, it stands to reason that `&[()]` should stay coherent with that and not treat `()` as a singleton-type (which would be the implied philosophy of optimizing the slice's pointer out).

I've taken `()` as example here, but I think this point stands for all ZSTs. While I've always liked the idea of treating ZSTs as singletons, that's never been Rust's implementation, and a lot of code relies on that when implementing opaque pointers.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+  equivalent to `Option<&[T]>` for non-zero-sized types.
+
+* Rust uses a dangling pointer in the representation of zero-length slices.
+  It's unclear whether C++ guarantees that a dangling pointer will remain
+  unchanged when passed through `std::span`. However, it does support
+  dangling pointers during regular construction via the use of
+  [`std::to_address`](https://en.cppreference.com/w/cpp/container/span/span)
+  in the iterator constructors.
+
+Note that C++ also does not support zero-sized types, so there is no naiive way
```
SabrinaJewson, line 138:
```suggestion
Note that C++ also does not support zero-sized types, so there is no native way
```

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+* Consider defining a separate Rust type which is repr-equivalent to the platform's
+  native `std::span<T, std::dynamic_extent>` to allow for easier
+  interoperability with C++ APIs. Unfortunately, the C++ standard does not
+  guarantee the layout of `std::span` (though the representation may be known
+  and fixed on a particular implementation, e.g. libc++/libstdc++/MSVC).
+  Zero-sized types would also not be supported with a naiive implementation of
```
SabrinaJewson, line 181:
```suggestion
  Zero-sized types would also not be supported with a naïve implementation of
```

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+[summary]: #summary
+
+This RFC guarantees the in-memory representation of slice and str references.
+Specifically, `&[T]` and `&mut [T]` are guaranteed to have the same layout as:
+
+```rust
+#[repr(C)]
+struct Slice<T> {
+    data: *const T,
+    len: usize,
```
hanna-kruppe, line 16: Why this specific order? Why not length before pointer? Are there any plausible reasons to prefer one over the other, e.g., based on target architecture?

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,182 @@
...
+code which smuggles pointers through the `data` value in `from_raw_parts` /
+`into_raw_parts`.
+
+## Uninhabited types
+
+Similarly, we could be *extra* tricky and make `&[!]` or other `&[Uninhabited]`
+types into a ZST since the slice can only ever be length zero. This may offer
+modest performance benefits for highly generic code which happens to create
+empty slices of uninhabited types, but this is unlikely to be worth the
+cost of maintaining a special case.
```
Jules-Bertholet, line 110: We could also make them length 1 by omitting the length but keeping the pointer. (I don’t necessarily think we should do this, just raising the possibility. Either way, raw pointers should keep the length field, for compatibility.)

text/0000-guaranteed-slice-repr.md
cramertj: >  We might not want to say "validity requirements" here

Yes, "validity requirements" is an overloaded term. I'd be happy to change it here if you have suggestions for a better one!

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+Note that C++ also does not support zero-sized types, so there is no naive way
+to represent types like `std::span<SomeZeroSizedRustType>`.
+
+## Flexibility
+
+Additionally, guaranteeing layout of Rust-native types limits the compiler's and
+standard library's ability to change and take advantage of new optimization
+opportunities.
+
+# Rationale and alternatives
```
tgross35, line 164: crABI provides a way to pass slices to/from C without locking in our representation https://github.com/rust-lang/rfcs/pull/3470, that is probably worth mentioning somewhere.

text/0000-guaranteed-slice-repr.md
workingjubilee: Note that our current layout algorithm is best able to exploit any niche in `len` if `len` is first in this ordering. And we currently cannot use a "double niche" on two different fields very effectively. I can work on improving this, but I cannot promise a specific result yet.

Due to various quirks of our existing APIs, the `len` niche would not apply to
1. zero-sized types
2. the "raw" version of this: `*mut [T]` and `*const [T]`

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+Slice references are represented with a pointer and length pair. Their in-memory
+layout is the same as a `#[repr(C)]` struct like the following:
+
+```rust
+#[repr(C)]
+struct Slice<T> {
+    data: *const T,
+    len: usize,
```
workingjubilee, line 54: If we wished to extend Rust to allow references to unsized types that encompass unsized types... for example, `&[[T]]`, or more interestingly, `&[dyn Trait]`... then one of the likely choices we would make is to have "triple-wide" (or larger!) references, instead of this "double-wide" one. Thus we would only apply the corresponding layout this RFC describes (or any other such specified layout) in the case of `where T: Sized`.

I feel this "multiple metadata" possibility should be considered and either _explicitly reserved_ as a future possibility, as accepting this RFC could be interpreted as foreclosing it, or dismissed.

workingjubilee (commented):
I am not sure that February 2025 is the right time to stabilize this layout, but I do think it is the right time to start discussing how to do so, why we should do so, and removing any blockers to it. Even if we didn't stabilize this layout, we could provide more ways to statically resolve the layout details so that code could interact with it. Nonetheless, I have a question: what code specifically do we want to enable with this choice? Is it Rust code or external over-FFI code? In the latter case, refusing to stabilize any parameter-passing or return-passing details would remain somewhat frustrating and makes this less of an improvement over the status quo.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
```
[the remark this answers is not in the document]
hanna-kruppe: Another niche that isn't currently exploited but may be in the future is alignment, e.g. the pointer in `&[u32]` is not only non-null but addresses 1, 2, and 3 are also invalid (assuming `align_of::<u32>() == 4`). This is a smaller niche than the length being at most `isize::MAX / size_of::<T>()`, but it also applies for ZSTs.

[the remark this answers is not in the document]
workingjubilee: Yes, perhaps more generally we should be clear that "the representation is stable" still allows for transformations to happen "around it", particularly when it comes to ADT tag layout.

text/0000-guaranteed-slice-repr.md
cramertj: @workingjubile Why would `len` have a niche? I would expect `data` to be the field with a niche, as it is non-null for `&[T]`.

hanna-kruppe: > The total size len * mem::size_of::<T>() of the slice must be no larger than isize::MAX, and adding that size to data must not “wrap around” the address space. See the safety documentation of [pointer::offset](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset).

https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+pointer since we know the length can only be zero.
+
+Either option may offer modest performance benefits for highly generic code
+which happens to create empty slices of uninhabited types, but this is unlikely
+to be worth the cost of maintaining a special case.
+
+## Compatibility with C++ `std::span`
+
+The largest drawback of this layout and set of validity requirements is that it
+may preclude `&[T]` from being representationally equivalent to C++'s
```
tmandry, line 128: I keep getting tripped up reading this section by thinking it says we're committing to something that diverges from C++. I would phrase it as something like

> Because C++ does not guarantee a layout for `std::span`, stabilizing one for Rust now means C++ could choose to stabilize a different layout and prevent the two from ever being compatible. However, this is unlikely to ever happen because MSVC uses the same representation proposed in this RFC and has stringent stability requirements.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+  [on MSVC](https://godbolt.org/z/Y8ardrshY), though not
+  [on GCC](https://godbolt.org/z/s4v4xehnG) nor
+  [on Clang](https://godbolt.org/z/qsd1K5oGq). Future changes to guarantee a
+  different layout in the C++ standard (unlikely due to MSVC ABI stabilitiy
+  requirements) could preclude matching the layout with `&[T]`.
+
+* Unlike Rust, `std::span` allows the `data` pointer to be `nullptr`. One
+  possibile workaround for this would be to guarantee that `Option<&[T]>` uses
+  `data: std::ptr::null(), len: 0` to represent the `None` case, making
+  `std::span<T>` equivalent to `Option<&[T]>` for non-zero-sized types.
```
tmandry, line 142: Maybe it would be worth a perf run to see if there's any measurable impact of doing this.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+  existing code which depends on storing data in the pointer of ZST slices.
+
+  This would break popular crates such as [bitvec](https://docs.rs/crate/bitvec/1.0.1/source/doc/ptr/BitSpan.md)
+  (55 million downloads) and would result in strange behavior such as
+  `std::ptr::slice_from_raw_parts(ptr, len).as_ptr()` returning a different
+  pointer from the one that was passed in.
+
+  Types like `*const ()` / `&()` are widely used to pass around pointers today.
+  We cannot make them zero-sized, and it would be surprising to make a
+  different choice for `&[()]`.
```
tmandry, line 187: Maybe we should have warned against this and made everyone use `*const u8` instead. Seems impossible to make a change like this now though.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+
+Similarly, we could be *extra* tricky and make `&[!]` or other `&[Uninhabited]`
+types into a ZST since the slice can only ever be length zero.
+
+If we want to maintain the pointer field, we could also make `&[!]` *just* a
+pointer since we know the length can only be zero.
+
+Either option may offer modest performance benefits for highly generic code
+which happens to create empty slices of uninhabited types, but this is unlikely
+to be worth the cost of maintaining a special case.
```
tmandry, line 123: On the other hand, we could opt not to guarantee this since no one seems to have much of a use case for this right now anyway.

cramertj:
@workingjubilee 

> Nonetheless, I have a question: what code specifically do we want to enable with this choice? Is it Rust code or external over-FFI code? In the latter case, refusing to stabilize any parameter-passing or return-passing details would remain somewhat frustrating and makes this less of an improvement over the status quo.

My use-case is FFI code (specifically, autogenerated C++ interop). Rust code can usually rely on calls to `.as_ptr`/`len`, so it's much less useful there. I don't particularly care whether the by-value ABI is stable as I'm happy to pass slice references by-pointer in parameter and return values.

text/0000-guaranteed-slice-repr.md
cramertj: Oh, clever! Thanks for describing that-- I'll mention it in the RFC.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+```rust
+#[repr(C)]
+struct Slice<T> {
+    data: *const T,
+    len: usize,
+}
+```
+
+The precise ABI of slice references is not guaranteed, so `&[T]` may not be
+passed by-value or returned by-value from an `extern "C" fn`.
```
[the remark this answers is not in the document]
cramertj: I'm happy to amend the RFC to clarify this point! The sentence you highlighted seems like it covers this to me, but perhaps I need to make it bolder/brighter/a headline.

Do you have other suggestions for how I could make this more straightforward?

text/0000-guaranteed-slice-repr.md
cramertj: I considered this, but IMO it seems difficult to imagine how `&[[T]]` would work-- even with triple-wide references, the items in the collection could be heterogeneous.

I suppose we could make it work if all of the elements were the same type, e.g. coercing `&[String]` into `&[dyn Debug]`.

Personally, I'd be happy to restrict this RFC to specify that this only applies to `&[T] where T: Sized`. Would that resolve your concern?

kennytm: we could make the RFC compatible with `&[[U]]` by simply relaxing the bound to `T: ?Sized`:

```rust
&[[U]]
->
{ data: *const [U], len: usize }
->
{ data: { data: *const U, len: usize }, len: usize }
```

so I don't see how this definition of `Slice<T>` forecloses the multi-meta possibility.

text/0000-guaranteed-slice-repr.md
dlight: Understood, thanks for explaining. That's.. really unfortunate.

One *could* still rigorously define the ABI to be what rustc does right now, and let callers in other languages use per-architecture inline assembly if they want to (this should be possible right?). But that's a much harder sell, and it's reasonable to leave this out of this RFC

text/0000-guaranteed-slice-repr.md
Jules-Bertholet: > I considered this, but IMO it seems difficult to imagine how `&[[T]]` would work-- even with triple-wide references, the items in the collection could be heterogeneous.

You could either require homogeneity like you suggest, or you could have the metadata be its own pointer to a separate region of memory, sharing the lifetime of the data pointer. (I wrote a very experimental crate that does something like the latter: https://github.com/Jules-Bertholet/unsized-vec)

But since current Rust doesn’t support this, and nothing in this RFC would interfere with someday adopting either option, it’s not a concern IMO.

programmerjake: > we could make the RFC compatible with `&[[U]]` by simply relaxing the bound to `T: ?Sized`

another valid extension is to just insert more fields for metadata between the pointer and length or change the length to be a struct-- these match how the pointer metadata APIs work more closely.

programmerjake: > changing the metadata (length) from `usize` to a struct would be [rust-lang/libs-team#246](https://github.com/rust-lang/libs-team/issues/246)

I don't really think so, that's just wrapping the metadata type in a struct whenever you're not using it as part of a pointer type. This is explicitly similar to `mem::Discriminant` which wraps an enum's discriminant in a struct, even though the discriminant might be a well-known type, e.g. `u8` for a `repr(u8)` fieldless enum.

text/0000-guaranteed-slice-repr.md
Jules-Bertholet: > as has been FCPed by opsem in [rust-lang/unsafe-code-guidelines#510](https://github.com/rust-lang/unsafe-code-guidelines/issues/510)

Theoretically, we could have been even stricter, additionally requiring that ptr + length not overflow the address space.

text/0000-guaranteed-slice-repr.md
burdges: I think both homogeneous `&[dyn Trait]` and rectangular `&[[T]]` sound useful, especially because the retangular case benefits from more optimizations.

We could consider unsizing options for more general types too like `struct Foo<T>(T,T)`, so then a homogeneous `Foo<dyn Trait>`.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC guarantees the in-memory representation of slice and str references.
+Specifically, `&[T]` and `&mut [T]` are guaranteed to have the same layout as:
+
+```rust
+#[repr(C)]
```
bjorn3, line 13: For the calling convention slices are actually passed as two separate arguments rather than as a single struct argument. Depending on the calling convention the latter can cause it to be passed on the stack, which would likely be a perf regression.

hanna-kruppe: It’s clarified later in the guide level explanation that the RFC does not intend to say anything about argument passing, only memory layout:

> The precise ABI of slice references is not guaranteed, so `&[T]` may not be
passed by-value or returned by-value from an `extern "C" fn`.

GoldsteinE:
Would be nice to also have this for `Box<[T]>` and `Box<str>`.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+struct HasString {
+    string: &'static str,
+}
+```
+
+Note: prior to this RFC, the type above is not even properly `repr(C)` since the
+size and alignment of slices were not guaranteed. However, the Rust compiler
+accepts the `repr(C)` declaration above without warning.
+
+# Guide-level explanation
```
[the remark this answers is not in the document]
tmccombs: Limiting it to platforms where the size and alignment of usize and a pointer are the same seems unnecessarily restrictive. Is there a difficulty with differing size and alignment I'm not aware of?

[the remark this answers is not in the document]
programmerjake: > Limiting it to platforms where the size and alignment of usize and a pointer are the same seems unnecessarily restrictive. Is there a difficulty with differing size and alignment I'm not aware of?

Yes, IIRC we still haven't decided if `usize` is an integer big enough to contain a pointer (C's `uintptr_t`), or big enough for just the address (effectively C's `size_t`) -- those are different on stuff like CHERI where a pointer also has another 64 bits (or more) of permissions information as well as the 64 bits of address.

Last discussion on it (that I recall at the moment):
https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Pre-RFC.20discussion.3A.20usize.20semantics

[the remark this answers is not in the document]
tmccombs: Sure. But even if it is decided that `usize` should be just enough for just the address (`size_t`), then would there be a problem with a slice being defined as something like
```rust
#[repr(C)]
struct Slice<T> {
    data: NonNull<T>,
    len: usize,
}
```
?

Another consideration is how the value would be passed by value as a function argument over FFI. I believe at least on x86_64, the `repr(C)` struct would probably be passed in two registers, but I'm not sure if that would be the case given a definition entirely based on offsets.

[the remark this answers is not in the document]
programmerjake: > Another consideration is how the value would be passed by value as a function argument over FFI. I believe at least on x86_64, the `repr(C)` struct would probably be passed in two registers, but I'm not sure if that would be the case given a definition entirely based on offsets.

passing by value through function arguments/return is specifically excluded by this RFC because that's much more complex and we may want to change it. In practice Rust passes a slice by value as separate pointer and length arguments, rather than as a struct. I've heard that you can't express what it does for return in C code on x86-64 (or x86? icr).

text/0000-guaranteed-slice-repr.md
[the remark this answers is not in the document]
workingjubilee: Simply be more boring and avoid vague terms like ABI, and specify exactly what you specify: "This guarantees the in-memory layout only, it does not involve the subject of how it is passed or returned through functions".

text/0000-guaranteed-slice-repr.md
the8472: I have a [draft PR](https://github.com/rust-lang/rust/pull/116542) where I tried implementing the niche in the length field, it didn't yield much perf improvements. Maybe I should try separating the metadata and the niche. And I haven't attempted reordering the fields yet either.

text/0000-guaranteed-slice-repr.md
cramertj: I think I lost the thread a bit here in terms of how this should affect the layout. AFAICT, there's no reasonable layout we can specify at this point for `&[T]` where `T` is not `Sized`. I'll change the RFC to specify that this layout guarantee is only applicable where `T: Sized`.

text/0000-guaranteed-slice-repr.md
joshlf, line 164: It could be good to mention slice DST references. Their layout is, in practice, the same as the layout of a slice reference, and so this RFC could easily stabilize their layout as well (albeit with `data: *const ()` instead of `data: *const T`).

@RalfJung may have opinions on why this is more subtle than I realize 😛  If it's actually straightforward, though, then it'd be worth considering including in this RFC.

cc @jswrenn

RalfJung: I'm not aware of any opsem subtleties in this area -- all concerns are of the form "do we really want to make this a hard guarantee we cannot change again later".

text/0000-guaranteed-slice-repr.md
bjorn3: Turns out this splitting behavior only applies to the Rust ABI: https://rust.godbolt.org/z/vfEWj5jja (and arm64: https://rust.godbolt.org/z/GoGPex1GM) That makes me a lot less worried about guaranteeing it for the C calling convention. In I'm now slightly in favor of making this guarantee. (Me actually verifying this claim was prompted by https://internals.rust-lang.org/t/what-prevents-rust-from-making-slices-ffi-safe/23049/7)

Jules-Bertholet: Might it not be better to guarantee that the splitting *does* occur? Not only does it improve performance, but it also leads to function signatures that are more typical for C, where `void some_function(whatever_type foo[], size_t len) { ... }` is the common practice.

hanna-kruppe: I think the RFC is right to only propose anything for memory layout and nothing for calling conventions. There's much more room for regret when committing to a particular calling convention for slices than when committing to a memory layout. And there's also much less upside: if you're passing a slice by value over FFI, it's a *local* decision to take it apart and pass the pointer and length however you want. In contrast, if the memory layout of a slice somewhere in memory isn't what you need (i.e., not possible to work with from outside Rust), you may have to change a `&[T]` field in a `repr(C)` struct to a pointer-length pair. That can affect a much larger amount of code far removed from the FFI calls.

workingjubilee: @Jules-Bertholet I think we would rather simply not promise anything right now about the calling convention of these types, certainly not that it gets split across arguments, because such
1. changes the arity between C and Rust, which is quite painful to reason about
2. closes the door on a future C extension introducing proper slice types and us being ABI compatible with them

text/0000-guaranteed-slice-repr.md
scottmcm: This would be a very weird special case, actually.  I'd worry about it far more in terms of all the places that would have to deal with it for correctness (in our codegen and in unsafe code) more than I'd worry about any perf implications.

Notably, `let x = None;` *only* sets the niche today, in niche-encoded cases like this.  Requiring that it also set the other value in the ScalarPair would be *without precedent* for how we do enums.  That means that this would have a giant impact to all kinds of things.  For example, would it mean that <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.StatementKind.html#variant.SetDiscriminant> now needs to update something *other* than the field storing the tag (be that a separate tag or a niche-encoded tag)?  That's worrisome.

RalfJung: I think we can guarantee that *reading* an all-zero chunk of memory produces a `None::<&[T]>`. But indeed we should not guarantee that *writing* a `None::<&[T]>` produces an all-zero representation.

zachs18: One possible downside is that even guaranteeing `zeroed::<Option<&[T]>>() == None` would probably preclude using the length as the niche in `&[T]` (for non-ZST `T`[^1]) as a future possibility, though. 

[^1]: since the length can be at most `isize::MAX as usize / size_of::<T>()` due to the maximum allocation size being `isize::MAX`

hanna-kruppe: I’m not sure if the compiler internals are equipped to do that, but in theory it’s possible to use the not-null niche for `Option<&[T]>` while using the length niche for other enums that need more niche values. This would probably give marginally better codegen on most targets than using `len = -1` as representation for `None` because zero is usually the cheapest constant to materialize and test against. (And it’s definitely better than `isize::MAX + 1`, which requires several instructions to materialize on some targets.)

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+[summary]: #summary
+
+This RFC guarantees the in-memory representation of slice and str references.
+Specifically, `&[T]` and `&mut [T]` are guaranteed to have the same layout as:
+
+```rust
+#[repr(C)]
+struct Slice<T> {
+    data: *const T,
+    len: usize,
```
newpavlov, line 16: I think it's better to use `NonNull` here:
```rust
struct Slice<T> {
    data: NonNull<T>,
    len: usize,
}
```
`NonNull` is FFI-safe and provides the niche, so `Option<&[T]>` and `Option<Slice<T>>` would behave the same way.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+
+Raw pointers to slices such as `*const [T]` or `*mut str` use the same layout
+as slice references, but do not necessarily point to anything.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Zero-sized types
+
+One could imagine representing `&[T]` as only `len` for zero-sized `T`.
```
newpavlov, line 104: IIUC Rust currently does not supports this optimization and there are no plans for it. `&[()]` and `&[Never]` currently have size of two words and changing it may be considered a breaking change (technically it's not, but some crates in the ecosystem rely on it).

Either way, as the first step it may be worth to guarantee layout for slices _iff_ `T` is `Sized` and non-ZST.

text/0000-guaranteed-slice-repr.md
the8472: "same layout as" here means field order, alignment and size, not validity. Validity requirements are mentioned further down. `len` also isn't really `usize` since it has invalid values which we may exploit in the future.

newpavlov: The `NonNull` struct has the "same layout", but it additionally encodes _some_ of the validity requirements which are important for layout of compound types like `Option<&[T]>`. Without it it's unclear whether the nullable pointer optimization is guaranteed to be applied here.

hanna-kruppe: `Option` has [its own (lack of) guarantees about its representation](https://doc.rust-lang.org/std/option/index.html#representation). Currently nothing is guaranteed about `Option<&[T]>` and nothing would be guaranteed about `Option<Slice<T>>` with `data: NonNull<T>` either. Also, since you mention FFI safety: the RFC only guarantees memory layout, not how slices are passed in calling conventions, but even if it did, [making the pointer `NonNull` doesn't make `Option<Slice<T>>` FFI-safe](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=399bf3485a46c0e556782a433cce1a0e) . Per discussion in https://github.com/rust-lang/rfcs/pull/3775#discussion_r2133460995 it's not obvious which, if any, guarantees can be provided without restricting future optimizations.

newpavlov: `Option` guarantees representation for `Option<&[T]>` if `T: Sized`. I think it's logical (and expected by many) to extend this guarantee to `Option<&[T]>` if we are to guarantee representation of `&[T]`. In other words, I would prefer for this to be explicitly covered in this RFC.

> making the pointer NonNull doesn't make Option<Slice<T>> FFI-safe 

Yes, but it's just a matter of extending `Option` guarantees. Rust _does_ apply the nullable pointer optimization to `Option<Slice<u32>>` and uses the expected ABI for passing values of this type.

hanna-kruppe: I assume you mean `Option<&T>` (for `T: Sized`) guarantees representation. I can understand the expectation that this extends to `Option<&[T]>`, but as discussed in the comments I linked, this is not as obvious as our gut feeling says it should because there's more than one niche to exploit in `&[T]`. I don't think trying to make even more guarantees in the same RFC is a good idea, it's tricky enough to be get the ones right that are currently proposed. I do agree, however, that the RFC should be explicit: if it does *not* intend to guarantee something related that a user may infer, it should explicitly disclaim that (with rationale).

> Yes, but it's just a matter of extending `Option` guarantees. Rust _does_ apply the nullable pointer optimization to `Option<Slice<u32>>` and uses the expected ABI for passing values of this type.

As already discussed in other comment threads, there are good reasons to not guarantee how bare `&[T]` is passed, and the RFC as written does not propose any such guarantee. Unless and until we're sure we can guarantee how bare `&[T]` is passed by value, it's premature to think about how `Option<&[T]>` might be passed by value. Please don't keep mixing up memory layout and calling convention.

text/0000-guaranteed-slice-repr.md
```diff
@@ -0,0 +1,205 @@
...
+
+Note: prior to this RFC, the type above is not even properly `repr(C)` since the
+size and alignment of slices were not guaranteed. However, the Rust compiler
+accepts the `repr(C)` declaration above without warning.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Slice references are represented with a pointer and length pair. Their in-memory
+layout is the same as a `#[repr(C)]` struct like the following:
```
the8472, line 48: Can we use more concrete language than "in-memory layout"? Or link to a definition if that's a spec term.

E.g. 

> Their size, align and field order is the same as...

text/0000-guaranteed-slice-repr.md
workingjubilee: > Personally, I'd be happy to restrict this RFC to specify that this only applies to &[T] where T: Sized. Would that resolve your concern?

Oh, yes, entirely.

PR #3789, opened 2025-03-14
Fix typo in 2093-infer-outlives.md

hkBst:
[Rendered](https://github.com/hkBst/rfcs/blob/patch-2/text/2093-infer-outlives.md)


Noratrieb:
what

Noratrieb:
we generally don't update old RFCs much, but this seems fine.

PR #3796, opened 2025-03-30
Allow `&&`, `||`, and `!` in `cfg`

jhpratt:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3796)*

About as simple as it gets.

[Rendered](https://github.com/jhpratt/rfcs/blob/cfg-logical-ops/text/0000-cfg-logical-ops.md)

### DIFF
--- a/text/0000-cfg-logical-ops.md
+++ b/text/0000-cfg-logical-ops.md
@@ -52,10 +52,10 @@ In terms of formal syntax, the [`[cfg.syntax]`][cfg-syntax] is changed to the fo
 > &nbsp;&nbsp; (`not` | `!`) `(` _ConfigurationPredicate_ `)`
 >
 > _ConfigurationAnd_\
-> &nbsp;&nbsp; _ConfigurationPredicateList_ `&&` _ConfigurationPredicateList_
+> &nbsp;&nbsp; _ConfigurationPredicate_ `&&` _ConfigurationPredicate_
 >
 > _ConfigurationOr_\
-> &nbsp;&nbsp; _ConfigurationPredicateList_ `||` _ConfigurationPredicateList_
+> &nbsp;&nbsp; _ConfigurationPredicate_ `||` _ConfigurationPredicate_
 >
 > _ConfigurationNotOption_\
 > &nbsp;&nbsp; `!` _ConfigurationOption_
--- a/text/0000-cfg-logical-ops.md
+++ b/text/0000-cfg-logical-ops.md
@@ -1,0 +1,94 @@
+- Feature Name: `cfg_logical_ops`
+- Start Date: 2025-03-30
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+`#[cfg]`, `#[cfg_attr]`, and `cfg!()` can use `&&`, `||`, and `!` for `all`, `any`, and `not`,
+respectively.
+
+# Motivation
+[motivation]: #motivation
+
+While there are no technical restrictions to using logical operators, this was not always the case.
+In Rust 1.0, attributes could not contain arbitrary tokens. This restriction was lifted in Rust
+1.34, but the `cfg` syntax was not updated to take advantage of this. By letting developers use
+logical operators, we are _lessening_ the burden of having to remember the `cfg` syntax.
+
+# Explanation
+[explanation]: #explanation
+[cfg-syntax]: https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.syntax
+
+`#[cfg(foo && bar)]` enables the annotated code if and only if both `foo` **and** `bar` are enabled.
+Similarly, `#[cfg(foo || bar)]` enables the annotated code if and only if either `foo` **or** `bar`
+is enabled. Finally, `#[cfg(!foo)]` enables the annotated code if and only if `foo` is **not**
+enabled. `#[cfg_attr]` and `cfg!()` behave the same way. Precedence is the same as in expressions.
+
+In terms of formal syntax, the [`[cfg.syntax]`][cfg-syntax] is changed to the following:
+
+> **<sup>Syntax</sup>**\
+> _ConfigurationPredicate_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _ConfigurationOption_\
+> &nbsp;&nbsp; | _ConfigurationAll_\
+> &nbsp;&nbsp; | _ConfigurationAny_\
+> &nbsp;&nbsp; | _ConfigurationNot_\
+> &nbsp;&nbsp; | _ConfigurationAnd_\
+> &nbsp;&nbsp; | _ConfigurationOr_\
+> &nbsp;&nbsp; | _ConfigurationNotOption_\
+> &nbsp;&nbsp; | `(` _ConfigurationPredicate_ `)`
+>
+> _ConfigurationOption_ :\
+> &nbsp;&nbsp; [IDENTIFIER]&nbsp;(`=` ([STRING_LITERAL] | [RAW_STRING_LITERAL]))<sup>?</sup>
+>
+> _ConfigurationAll_\
+> &nbsp;&nbsp; `all` `(` _ConfigurationPredicateList_<sup>?</sup> `)`
+>
+> _ConfigurationAny_\
+> &nbsp;&nbsp; `any` `(` _ConfigurationPredicateList_<sup>?</sup> `)`
+>
+> _ConfigurationNot_\
+> &nbsp;&nbsp; (`not` | `!`) `(` _ConfigurationPredicate_ `)`
+>
+> _ConfigurationAnd_\
+> &nbsp;&nbsp; _ConfigurationPredicateList_ `&&` _ConfigurationPredicateList_
+>
+> _ConfigurationOr_\
+> &nbsp;&nbsp; _ConfigurationPredicateList_ `||` _ConfigurationPredicateList_
+>
+> _ConfigurationNotOption_\
+> &nbsp;&nbsp; `!` _ConfigurationOption_
+>
+> _ConfigurationPredicateList_\
+> &nbsp;&nbsp; _ConfigurationPredicate_ (`,` _ConfigurationPredicate_)<sup>\*</sup> `,`<sup>?</sup>
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Two ways to express the same thing. This can be somewhat mitigated by a lint for the old syntax.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- The current syntax is verbose and a relic of the past when attributes could not contain arbitrary
+  tokens.
+- Using existing, widely-understood operators makes the syntax more familiar.
+
+# Prior art
+[prior-art]: #prior-art
+
+The `efg` crate is nearly identical to this proposal, the sole difference being not requiring `=`
+for key-value pairs.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+None so far.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+- Pattern-like syntax such as `#[cfg(feature = "foo" | "bar")]` could be allowed as a shorthand for
+  `#[cfg(feature = "foo" || feature = "bar")]`. This would be particularly useful for
+  platform-specific code (e.g. `#[cfg(target_os = "linux" | "windows")]`).
--- a/text/0000-cfg-logical-ops.md
+++ b/text/0000-cfg-logical-ops.md
@@ -7,7 +7,8 @@
 [summary]: #summary
 
 `#[cfg]`, `#[cfg_attr]`, and `cfg!()` can use `&&`, `||`, and `!` for `all`, `any`, and `not`,
-respectively.
+respectively. Due to precedence, `feature = "foo"` must be parenthesized when adjoining any of the
+new operators.
 
 # Motivation
 [motivation]: #motivation
@@ -27,36 +28,39 @@ Similarly, `#[cfg(foo || bar)]` enables the annotated code if and only if either
 is enabled. Finally, `#[cfg(!foo)]` enables the annotated code if and only if `foo` is **not**
 enabled. `#[cfg_attr]` and `cfg!()` behave the same way.
 
-Precedence is the [same as in expressions][precedence], with `=` being treated as `==` for this
-purpose.
+Precedence is the [same as in expressions][precedence], which necessitates using parentheses in some
+situations. This is shown in the table below. `=` is **not** treated the same as `==` for precedence
+purposes; it has lower precedence than all logical operators.
 
 ## Examples
 
-| Syntax                             | Equivalent to                                      | Rationale                                                                           |
-| ---------------------------------- | -------------------------------------------------- | ----------------------------------------------------------------------------------- |
-| `a && b`                           | `all(a, b)`                                        | definition of `&&`                                                                  |
-| `a \|\| b`                         | `any(a, b)`                                        | definition of `\|\|`                                                                |
-| `!a`                               | `not(a)`                                           | definition of `!`                                                                   |
-| `(a)`                              | `a`                                                | definition of `()`                                                                  |
-| `a && b && c && d`                 | `all(a, b, c, d)` (or `all(all(all(a, b), c), d)`) | `&&` is associative                                                                 |
-| `a \|\| b \|\| c \|\| d`           | `any(a, b, c, d)` (or `any(any(any(a, b), c), d)`) | `\|\|` is associative                                                               |
-| `!!!!!!a`                          | `not(not(not(not(not(not(a))))))`                  | `!` can be repeated                                                                 |
-| `((((((a))))))`                    | a                                                  | `()` can be nested                                                                  |
-| `a && b \|\| c && d`               | `any(all(a, b), all(c, d))`                        | `\|\|` has lower precedence than `&&`                                               |
-| `a \|\| b && c \|\| d`             | `any(a, all(b, c), d)`                             | `\|\|` has lower precedence than `&&`                                               |
-| `(a \|\| b) && (c \|\| d)`         | `all(any(a, b), any(c, d))`                        | `()` can be used for grouping                                                       |
-| `!a \|\| !b && !c`                 | `any(not(a), all(not(b), not(c)))`                 | `!` has highest precedence                                                          |
-| `feature="foo" \|\| feature="bar"` | `any(feature="foo", feature="bar")`                | `\|\|` has lower precedence than `=`                                                |
-| `feature="foo" && feature="bar"`   | `all(feature="foo", feature="bar")`                | `&&` has lower precedence than `=`                                                  |
-| `!feature="foo"`                   | _syntax error_                                     | `!` has higher precedence than `=`, which may be confusing, so we ban this syntax   |
-| `!(feature="foo")`                 | `not(feature="foo")`                               | use `()` for grouping                                                               |
-| `!all(x, y)`                       | `not(all(x, y))`                                   | `!` has lower precedence than "function call"                                       |
-| `any(!x \|\| !w, !(y && z))`       | `any(any(not(x), not(w)), not(all(y, z)))`         | `!`, `&&` etc. can be used inside `any`, `all` and `not`                            |
-| `true && !false`                   | `all(true, not(false))`                            | `!`, `&&` etc. can be used on boolean literals (they are syntactically identifiers) |
-| `!accessible(std::mem::forget)`    | `not(accessible(std::mem::forget))`                | `!`, `&&` etc. can be used on `cfg_accessible`                                      |
-| `accessible(std::a \|\| std::b)`   | _syntax error_                                     | … but not inside                                                                    |
-| `!version("1.42.0")`               | `not(version("1.42.0"))`                           | `!`, `&&` etc. can be used on `cfg_version`                                         |
-| `version(!"1.42.0")`               | _syntax error_                                     | … but not inside                                                                    |
+| Syntax                                 | Equivalent to                                      | Rationale                                                                            |
+| -------------------------------------- | -------------------------------------------------- | ------------------------------------------------------------------------------------ |
+| `a && b`                               | `all(a, b)`                                        | definition of `&&`                                                                   |
+| `a \|\| b`                             | `any(a, b)`                                        | definition of `\|\|`                                                                 |
+| `!a`                                   | `not(a)`                                           | definition of `!`                                                                    |
+| `(a)`                                  | `a`                                                | definition of `()`                                                                   |
+| `a && b && c && d`                     | `all(a, b, c, d)` (or `all(all(all(a, b), c), d)`) | `&&` is associative                                                                  |
+| `a \|\| b \|\| c \|\| d`               | `any(a, b, c, d)` (or `any(any(any(a, b), c), d)`) | `\|\|` is associative                                                                |
+| `!!!!!!a`                              | `not(not(not(not(not(not(a))))))`                  | `!` can be repeated                                                                  |
+| `((((((a))))))`                        | a                                                  | `()` can be nested                                                                   |
+| `a && b \|\| c && d`                   | `any(all(a, b), all(c, d))`                        | `\|\|` has lower precedence than `&&`                                                |
+| `a \|\| b && c \|\| d`                 | `any(a, all(b, c), d)`                             | `\|\|` has lower precedence than `&&`                                                |
+| `(a \|\| b) && (c \|\| d)`             | `all(any(a, b), any(c, d))`                        | `()` can be used for grouping                                                        |
+| `!a \|\| !b && !c`                     | `any(not(a), all(not(b), not(c)))`                 | `!` has highest precedence                                                           |
+| `feature="foo" \|\| feature="bar"`     | _syntax error_                                     | `\|\|` has higher precedence than `=`, which may be confusing, so we ban this syntax |
+| `(feature="foo") \|\| (feature="bar")` | `any(feature="foo", feature="bar")`                | use `()` for grouping                                                                |
+| `feature="foo" && feature="bar"`       | _syntax error_                                     | `&&` has higher precedence than `=`, which may be confusing, so we ban this syntax   |
+| `(feature="foo") && (feature="bar")`   | `all(feature="foo", feature="bar")`                | use `()` for grouping                                                                |
+| `!feature="foo"`                       | _syntax error_                                     | `!` has higher precedence than `=`, which may be confusing, so we ban this syntax    |
+| `!(feature="foo")`                     | `not(feature="foo")`                               | use `()` for grouping                                                                |
+| `!all(x, y)`                           | `not(all(x, y))`                                   | `!` has lower precedence than "function call"                                        |
+| `any(!x \|\| !w, !(y && z))`           | `any(any(not(x), not(w)), not(all(y, z)))`         | `!`, `&&` etc. can be used inside `any`, `all` and `not`                             |
+| `true && !false`                       | `all(true, not(false))`                            | `!`, `&&` etc. can be used on boolean literals (they are syntactically identifiers)  |
+| `!accessible(std::mem::forget)`        | `not(accessible(std::mem::forget))`                | `!`, `&&` etc. can be used on `cfg_accessible`                                       |
+| `accessible(std::a \|\| std::b)`       | _syntax error_                                     | … but not inside                                                                     |
+| `!version("1.42.0")`                   | `not(version("1.42.0"))`                           | `!`, `&&` etc. can be used on `cfg_version`                                          |
+| `version(!"1.42.0")`                   | _syntax error_                                     | … but not inside                                                                     |
 
 ## Formal syntax
 
@@ -142,5 +146,9 @@ None so far.
 [future-possibilities]: #future-possibilities
 
 - Pattern-like syntax such as `#[cfg(feature = "foo" | "bar")]` could be allowed as a shorthand for
-  `#[cfg(feature = "foo" || feature = "bar")]`. This would be particularly useful for
+  `#[cfg((feature = "foo") || (feature = "bar"))]`. This would be particularly useful for
   platform-specific code (e.g. `#[cfg(target_os = "linux" | "windows")]`).
+- The use of parentheses could be relaxed in some situations, such as allowing `!feature = "foo"` or
+  `feature = "foo" || feature = "bar"`.
+- A different syntax for key-value pairs, such as `feature("foo")`, could be used for clarity (as it
+  is neither assignment nor equality) and to reduce the need for manual precedence management.
--- a/text/0000-cfg-logical-ops.md
+++ b/text/0000-cfg-logical-ops.md
@@ -1,6 +1,6 @@
 - Feature Name: `cfg_logical_ops`
 - Start Date: 2025-03-30
-- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- RFC PR: [rust-lang/rfcs#3796](https://github.com/rust-lang/rfcs/pull/3796)
 - Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
 
 # Summary
@@ -20,13 +20,46 @@ logical operators, we are _lessening_ the burden of having to remember the `cfg`
 # Explanation
 [explanation]: #explanation
 [cfg-syntax]: https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.syntax
+[precedence]: https://doc.rust-lang.org/reference/expressions.html#expression-precedence
 
 `#[cfg(foo && bar)]` enables the annotated code if and only if both `foo` **and** `bar` are enabled.
 Similarly, `#[cfg(foo || bar)]` enables the annotated code if and only if either `foo` **or** `bar`
 is enabled. Finally, `#[cfg(!foo)]` enables the annotated code if and only if `foo` is **not**
-enabled. `#[cfg_attr]` and `cfg!()` behave the same way. Precedence is the same as in expressions.
+enabled. `#[cfg_attr]` and `cfg!()` behave the same way.
 
-In terms of formal syntax, the [`[cfg.syntax]`][cfg-syntax] is changed to the following:
+Precedence is the [same as in expressions][precedence].
+
+## Examples
+
+| Syntax                             | Equivalent to                                      | Rationale                                                                           |
+| ---------------------------------- | -------------------------------------------------- | ----------------------------------------------------------------------------------- |
+| `a && b`                           | `all(a, b)`                                        | definition of `&&`                                                                  |
+| `a \|\| b`                         | `any(a, b)`                                        | definition of `\|\|`                                                                |
+| `!a`                               | `not(a)`                                           | definition of `!`                                                                   |
+| `(a)`                              | `a`                                                | definition of `()`                                                                  |
+| `a && b && c && d`                 | `all(a, b, c, d)` (or `all(all(all(a, b), c), d)`) | `&&` is associative                                                                 |
+| `a \|\| b \|\| c \|\| d`           | `any(a, b, c, d)` (or `any(any(any(a, b), c), d)`) | `\|\|` is associative                                                               |
+| `!!!!!!a`                          | `not(not(not(not(not(not(a))))))`                  | `!` can be repeated                                                                 |
+| `((((((a))))))`                    | a                                                  | `()` can be nested                                                                  |
+| `a && b \|\| c && d`               | `any(all(a, b), all(c, d))`                        | `\|\|` has lower precedence than `&&`                                               |
+| `a \|\| b && c \|\| d`             | `any(a, all(b, c), d)`                             | `\|\|` has lower precedence than `&&`                                               |
+| `(a \|\| b) && (c \|\| d)`         | `all(any(a, b), any(c, d))`                        | `()` can be used for grouping                                                       |
+| `!a \|\| !b && !c`                 | `any(not(a), all(not(b), not(c)))`                 | `!` has highest precedence                                                          |
+| `feature="foo" \|\| feature="bar"` | `any(feature="foo", feature="bar")`                | `\|\|` has lower precedence than `=`                                                |
+| `feature="foo" && feature="bar"`   | `all(feature="foo", feature="bar")`                | `&&` has lower precedence than `=`                                                  |
+| `!feature="foo"`                   | _syntax error_                                     | `!` has higher precedence than `=`, which may be confusing, so we ban this syntax   |
+| `!(feature="foo")`                 | `not(feature="foo")`                               | use `()` for grouping                                                               |
+| `!all(x, y)`                       | `not(all(x, y))`                                   | `!` has lower precedence than "function call"                                       |
+| `any(!x \|\| !w, !(y && z))`       | `any(any(not(x), not(w)), not(all(y, z)))`         | `!`, `&&` etc. can be used inside `any`, `all` and `not`                            |
+| `true && !false`                   | `all(true, not(false))`                            | `!`, `&&` etc. can be used on boolean literals (they are syntactically identifiers) |
+| `!accessible(std::mem::forget)`    | `not(accessible(std::mem::forget))`                | `!`, `&&` etc. can be used on `cfg_accessible`                                      |
+| `accessible(std::a \|\| std::b)`   | _syntax error_                                     | … but not inside                                                                    |
+| `!version("1.42.0")`               | `not(version("1.42.0"))`                           | `!`, `&&` etc. can be used on `cfg_version`                                         |
+| `version(!"1.42.0")`               | _syntax error_                                     | … but not inside                                                                    |
+
+## Formal syntax
+
+[`[cfg.syntax]`][cfg-syntax] is changed to the following:
 
 > **<sup>Syntax</sup>**\
 > _ConfigurationPredicate_ :\
@@ -36,8 +69,19 @@ In terms of formal syntax, the [`[cfg.syntax]`][cfg-syntax] is changed to the fo
 > &nbsp;&nbsp; | _ConfigurationNot_\
 > &nbsp;&nbsp; | _ConfigurationAnd_\
 > &nbsp;&nbsp; | _ConfigurationOr_\
-> &nbsp;&nbsp; | _ConfigurationNotOption_\
+> &nbsp;&nbsp; | _ConfigurationNegation_\
 > &nbsp;&nbsp; | `(` _ConfigurationPredicate_ `)`
+>
+> _ConfigurationNegatable_ :\
+> &nbsp;&nbsp; _ConfigurationOptionIdent_\
+> &nbsp;&nbsp; | _ConfigurationAll_\
+> &nbsp;&nbsp; | _ConfigurationAny_\
+> &nbsp;&nbsp; | _ConfigurationNot_\
+> &nbsp;&nbsp; | _ConfigurationNegation_ \
+> &nbsp;&nbsp; | `(` _ConfigurationPredicate_ `)`
+>
+> _ConfigurationOptionIdent_ :\
+> &nbsp;&nbsp; [IDENTIFIER]
 >
 > _ConfigurationOption_ :\
 > &nbsp;&nbsp; [IDENTIFIER]&nbsp;(`=` ([STRING_LITERAL] | [RAW_STRING_LITERAL]))<sup>?</sup>
@@ -57,11 +101,14 @@ In terms of formal syntax, the [`[cfg.syntax]`][cfg-syntax] is changed to the fo
 > _ConfigurationOr_\
 > &nbsp;&nbsp; _ConfigurationPredicate_ `||` _ConfigurationPredicate_
 >
-> _ConfigurationNotOption_\
-> &nbsp;&nbsp; `!` _ConfigurationOption_
+> _ConfigurationNegation_\
+> &nbsp;&nbsp; `!` _ConfigurationNegatable_
 >
 > _ConfigurationPredicateList_\
 > &nbsp;&nbsp; _ConfigurationPredicate_ (`,` _ConfigurationPredicate_)<sup>\*</sup> `,`<sup>?</sup>
+
+All future function-like predicates (such as `version` and `accessible`) should be added to
+_ConfigurationNegatable_.
 
 # Drawbacks
 [drawbacks]: #drawbacks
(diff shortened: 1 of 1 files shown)


text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,94 @@
...
+> &nbsp;&nbsp; `any` `(` _ConfigurationPredicateList_<sup>?</sup> `)`
+>
+> _ConfigurationNot_\
+> &nbsp;&nbsp; (`not` | `!`) `(` _ConfigurationPredicate_ `)`
+>
+> _ConfigurationAnd_\
+> &nbsp;&nbsp; _ConfigurationPredicateList_ `&&` _ConfigurationPredicateList_
+>
+> _ConfigurationOr_\
+> &nbsp;&nbsp; _ConfigurationPredicateList_ `||` _ConfigurationPredicateList_
```
programmerjake, line 58: I expect `a, b && c, d` to not be valid where `a, b` is not valid, also I expect `a, b && c, d` to not mean `(a, b) && (c, d)`

```suggestion
> &nbsp;&nbsp; _ConfigurationPredicate_ `&&` _ConfigurationPredicate_
>
> _ConfigurationOr_\
> &nbsp;&nbsp; _ConfigurationPredicate_ `||` _ConfigurationPredicate_
```

text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,94 @@
...
+>
+> _ConfigurationNotOption_\
+> &nbsp;&nbsp; `!` _ConfigurationOption_
```
bluebear94, line 61: This would allow things like `#[cfg(!feature = "foo")]`, which looks strange. Could we restrict this to taking options that consist of a single identifier?

CraftSpider: At least personally, I use `key = "value"` predicates fairly frequently, so I'd expect them to work. I haven't looked closely at the grammar, but if the choice is made to not allow `!feature = "foo"`, I'd hope to at least keep `!(feature = "foo")` or similar, so I can actually use the `!` syntax basically ever.

text/0000-cfg-logical-ops.md
jhpratt: funnily enough I had this originally but misread it when copying into the final format

text/0000-cfg-logical-ops.md
jhpratt: Yes, this would permit that, which was a deliberate decision. If there is not the will for it, I can remove it or restrict per your suggestion, leaving only `!(feature = "foo")`. That would be permitted by the `ConfigurationNot` rule.

text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+While there are no technical restrictions to using logical operators, this was not always the case.
+In Rust 1.0, attributes could not contain arbitrary tokens. This restriction was lifted in Rust
+1.34, but the `cfg` syntax was not updated to take advantage of this. By letting developers use
+logical operators, we are _lessening_ the burden of having to remember the `cfg` syntax.
+
+# Explanation
```
kennytm, line 20: i'd like to see some more complex examples demonstrating what is supported and invalid, such as the strange `#[cfg(!feature = "foo")]`

| Syntax | Expected lowering | Rationale |
|---|---|---|
| `a && b` | `all(a, b)` | Definition of `&&` |
| `a \|\| b` | `any(a, b)` | Definition of `\|\|` |
| `!a` | `not(a)` | Definition of `!` |
| `(a)` | `a` | Definition of `()` |
| `a && b && c && d` | `all(a, b, c, d)` (or `all(all(all(a, b), c), d)`) | `&&` is associative |
| `a \|\| b \|\| c \|\| d` | `any(a, b, c, d)` (or `any(any(any(a, b), c), d)`) | `\|\|` is associative |
| `!!!!!!a` | `not(not(not(not(not(not(a))))))` | `!` can be nested |
| `((((((a))))))` | a | `()` can be nested |
| `a && b \|\| c && d` | `any(all(a, b), all(c, d))` | `\|\|` has lower precedence than `&&` |
| `a \|\| b && c \|\| d` | `any(a, all(b, c), d)` | `\|\|` has lower precedence than `&&` |
| `(a \|\| b) && (c \|\| d)` | `all(any(a, b), any(c, d))` | `()` can be used for grouping |
| `!a \|\| !b && !c` | `any(not(a), all(not(b), not(c)))` | `!` has highest precedence |
| `feature="foo" \|\| feature="bar"` | `any(feature="foo", feature="bar")` | `\|\|` has lower precedence than `=` |
| `feature="foo" && feature="bar"` | `all(feature="foo", feature="bar")` | `&&` has lower precedence than `=` |
| `!feature="foo"` | syntax error | `!` has higher precedence than `=`, which is confusing, so we ban this syntax |
| `!(feature="foo")` | `not(feature="foo")` | Use `()` for grouping |
| `!all(x, y)` | `not(all(x, y))` | `!` has lower precedence than "function call" |
| `any(!x \|\| !w, !(y && z))` | `any(any(not(x), not(w)), not(all(y, z)))` | `!`, `&&` etc can be used inside `any`, `all` and `not` |
| `true && !false` | `all(true, not(false))` | `!`, `&&` etc can be used on boolean literals |
| `!accessible(std::mem::forget)` | `not(accessible(std::mem::forget))` | `!`, `&&` etc can be used on cfg_accessible |
| `accessible(std::a \|\| std::b)` | syntax error | ... but not inside |
| `!version("1.42.0")` | `not(version("1.42.0"))` | `!`, `&&` etc can be used on cfg_version |
| `version(!"1.42.0")` | syntax error | ... but not inside |

text/0000-cfg-logical-ops.md
RalfJung: I feel fairly strongly that we should not allow `!feature = "foo"`. This part of the syntax goes against existing precedent in Rust and is visually quite confusing. In code we write `!(a == b)`, or `a != b`, so that's what we should do here as well.

jhpratt: I'll remove this from the text later today, requiring parentheses in this situation.

text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,94 @@
...
+> &nbsp;&nbsp; `any` `(` _ConfigurationPredicateList_<sup>?</sup> `)`
+>
+> _ConfigurationNot_\
+> &nbsp;&nbsp; (`not` | `!`) `(` _ConfigurationPredicate_ `)`
+>
+> _ConfigurationAnd_\
+> &nbsp;&nbsp; _ConfigurationPredicate_ `&&` _ConfigurationPredicate_
+>
+> _ConfigurationOr_\
+> &nbsp;&nbsp; _ConfigurationPredicate_ `||` _ConfigurationPredicate_
```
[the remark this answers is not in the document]
jhpratt: I was almost positive this was required coming in. Is there a "typical" way of expressing this in formal grammar? Easier to ask than spend time figuring it out myself.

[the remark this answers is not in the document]
CosmicHorrorDev: I believe you would define the grammar to follow transitions through lower precedence to higher, so something like. An example with just `||`, `&&`, `!`, and `()` (in the order of increasing precedence)

With the starting state of ConfigurationPredicate

_Transitions to climb precedence with a single value alone:_
- ConfigurationPredicate -> ConfigurationOr
- ConfigurationOr -> ConfigurationAnd
- ConfigurationAnd -> ConfigurationNot
- ConfigurationNot -> ConfigurationParens

_And the actual transitions with operators:_
- ConfigurationOr -> ConfigurationOr `||` ConfigurationAnd
- ConfigurationAnd -> ConfigurationAnd `&&` ConfigurationNot
- ConfigurationNot -> `!` ConfigurationParens
- ConfigurationParens -> `(` ConfigurationOr `)`

_Terminating the grammar with idents:_
- ConfigurationParens -> Ident

---

Consider `a && b || c && (d || !e)`

1. ConfigurationPredicate -> ConfigurationOr
2. ConfigurationOr -> ConfigurationOr `||` ConfigurationAnd with `a && b` `||` `c && (d || !e)`

Considering the chunk at ConfigurationOr with `a && b`

1. ConfigurationOr -> ConfigurationAnd
2. ConfigurationAnd -> ConfigurationAnd `&&` ConfigurationNot with `a` `&&` `b`
3. `a` and `b` go through all of the transitions to Ident

Considering the chunk at ConfigurationAnd with `c && (c || !e)`

1. ConfigurationAnd -> ConfigurationAnd `&&` ConfigurationNot with `c` `&&` `(c || !e)`
2. `c` goes all the way to Ident

Considering the chunk at ConfigurationNot with `(c || !e)`

1. ConfigurationNot -> ConfigurationParens
2. ConfigurationParens -> `(` ConfigurationOr `)` with `(` `c || !e` `)`
3. ConfigurationOr -> ConfigurationOr `||` ConfigurationAnd with `c` `||` `!e`
4. ConfigurationAnd -> ConfigurationNot
5. ConfigurationNot -> `!` ConfigurationParens with `!` `e`
6. ConfigurationParens -> Ident with `e`

This grammar avoids ambiguity (in the formal sense of only having a single tree that can represent a given expression) and handles precedence correctly (mentally you can throw parens around the chunks to see that they get grouped right. It'd be easier to show as a tree, but that's a lot of effort)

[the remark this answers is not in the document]
programmerjake: rust's expression grammar in the rust reference doesn't use that method to avoid ambiguity and handle precedence, instead it just has a table for expression precedence: https://doc.rust-lang.org/reference/expressions.html#expression-precedence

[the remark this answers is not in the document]
CosmicHorrorDev: A table definitely sounds a lot simpler :grin: 

Thanks, I'm not familiar with the project specifics, so that seems like a better approach

text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,94 @@
+- Feature Name: `cfg_logical_ops`
+- Start Date: 2025-03-30
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+`#[cfg]`, `#[cfg_attr]`, and `cfg!()` can use `&&`, `||`, and `!` for `all`, `any`, and `not`,
```
Jules-Bertholet, line 9: One alternative would be to use `&` and `|` instead of `&&` and `||`.

text/0000-cfg-logical-ops.md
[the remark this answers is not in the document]
jhpratt: I can explicitly state that `&&` has higher precedence, leaving the grammar as-is from the sounds of it?

clarfonthey:
I think that a `feature != "string"` syntax as an alternative to `not(feature = "string")` would be a nice addition here. `!(feature = "string")` feels weirder to me than `not(feature = "string")`, honestly.

jhpratt:
The problem I see with `feature != "foo"` is that it could plausibly be interpreted as "any feature other than `foo`".

text/0000-cfg-logical-ops.md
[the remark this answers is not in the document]
programmerjake: maybe just say it has the same precedence rules as rust expressions? if you want an explicit table, maybe:
| Operator/Punctuation | Associativity |
|-|-|
| parenthesis and `all(...)`/`any(...)`/`not(...)` | &nbsp; |
| `!` operator | &nbsp; |
| `&&` operator | left to right |
| `\|\|` operator | left to right |
| `=` syntax | &nbsp; |
| `,` punctuation | &nbsp; |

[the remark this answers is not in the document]
jhpratt: writing things up now, and I'm simply linking to the reference

clarfonthey:
That interpretation feels like a stretch to me, but it's also understandable. Perhaps a `feature("string")` syntax might be a bit more natural with `!feature("string")`, and also extensible to `feature("string", "other string")`.

jhpratt:
Updated the following:

- `!feature = "foo"` is no longer permitted. Negation can be followed by parentheses, a single identifier (so `#[cfg(!unix)]`), and any function-like predicate.
- `&`/`|` added as alternative to `&&`/`||`
- `feature != "foo"` added as alternative to `!(feature = "foo")`
- added table with examples

text/0000-cfg-logical-ops.md
```diff
@@ -20,13 +20,46 @@ logical operators, we are _lessening_ the burden of having to remember the `cfg`
...
 
 `#[cfg(foo && bar)]` enables the annotated code if and only if both `foo` **and** `bar` are enabled.
 Similarly, `#[cfg(foo || bar)]` enables the annotated code if and only if either `foo` **or** `bar`
 is enabled. Finally, `#[cfg(!foo)]` enables the annotated code if and only if `foo` is **not**
-enabled. `#[cfg_attr]` and `cfg!()` behave the same way. Precedence is the same as in expressions.
-
-In terms of formal syntax, the [`[cfg.syntax]`][cfg-syntax] is changed to the following:
+enabled. `#[cfg_attr]` and `cfg!()` behave the same way.
+
+Precedence is the [same as in expressions][precedence].
```
programmerjake, line 30: you'll want to note `=` is treated as `==` for precedence purposes, since assignment has much lower precedence.

jhpratt: done

m-ou-se:
I read `#[cfg(feature = "asdf")]` similar to how I read things like `#[serde(default = "asdf")]` or `#[link_name = "asdf"]: The `=` syntax is used as a key-value pair, like in all attributes.

As mentioned above, `feature = "asdf"` is not at all like a Rust expression. It doesn't behave like a `=` expression (it's not an assignment). It doesn't behave like an `==` expression (it doesn't check equality either). There is no variable/concept named `feature` that is compared to anything.

It isn't even commutative: `#[cfg("asdf" = feature)]` doesn't work.

Using `key = "value"` syntax makes sense as it matches the syntax for all other attributes. But I don't think it makes much sense to take `&&` and `||` from Rust expressions and trying to force them into attribute syntax. The result feels weirdly inconsistent, with things like `!(feature = "asdf")` not being the same as `feature != "asdf"`, and so on.

So, if we want to go in this direction, we should also change the `feature = "asdf"` syntax to something that fits a Rust expression. E.g. `#[cfg(has_feature("asdf") && has_target_feature("asdf"))]`.

m-ou-se:
Let's look at a real world example:
```rust
#[cfg(any(target_os = "android", target_os = "linux", target_os = "emscripten", target_os = "freebsd"))]
```

This proposal would allow rewriting that as:

```rust
#[cfg(target_os = "android" || target_os = "linux" || target_os = "emscripten" || target_os = "freebsd"))]
```

Which looks a bit nicer.

But instead of just replacing `any()` by a series of `||`, I think we can do much better without giving up on the `key = value` syntax with something like:

```rust
#[cfg(target_os = "android" | "linux" | "emscripten" | "freebsd")]
```

I don't want to derail this thread with alternative proposals, but I do think that `&&` and `||` is not getting us that much. If we go in that direction of making cfg() accept expression-like syntax, then we might be closing ourselves off from improvements to the `key = value` syntax that might serve the actual use cases better.

RalfJung:
>  The result feels weirdly inconsistent, with things like !(feature = "asdf") not being the same as feature != "asdf"

How would they not be the same? The RFC, I think, intents for both of these to mean the same thing: the `feature` variable does not contain "asdf".

m-ou-se:
`!(feature = "asdf")` would work, while `feature != "asdf"` would be an error.

RalfJung:
Ah. That seems entirely consistent to me, it's `=` and not `==` after all.

I like `target_os = "android" | "linux" | "emscripten" | "freebsd"` but it doesn't when you need to look at more than one key.

m-ou-se:
My point is that we need tot look at real world examples of code we're trying to simplify.

For example, quite often, `any()` is used with the same key, such as when selecting for a list of different operating systems. We should optimize the syntax for patterns that occur in actual code.

Large cfg() attributes can be very confusing. But replacing `any()` and `all()` by `||` and `&&` isn't making it much better. Large boolean expressions will still be confusing, especially if their syntax is a hybrid of expressions and (key=value) attribute syntax with inconsistent precedence and operator meaning.

m-ou-se:
Rust parses `feature = "asdf" && unix` as `feature = ("asdf" && unix`)`, _not_ as `(feature = "asdf") && unix`.

One might argue that `=` in attributes is different, but how would we expect an attribute proc macro to parse this today?

```
#[my_attribute_macro(expr = a && b, expr2 = c || d)]
```

I would assume it interprets that as two `key = value` pairs with the values being `a && b` and `c || d`. (It's up to the proc macro to decide, this is what you get if you parse it with a Rust expression parser.)

So even in an attribute, I'd argue that `&&` should have higher precedence than the `key=value` operator.

programmerjake:
maybe embrace that they are sets and allow using `"abcd" in feature` instead of `feature = "abcd"`? though then that makes me think we should have named it `features`.

kennytm:
> I would go so far as to argue that `any()` and `all()` are _better for formatting_. Binary operators used repeatedly have an awkward indentation problem. 
>
> ...
>
> Whereas a similar boolean expression is less regular and longer:
>
> ...
>
> (And there are probably even messier cases to present with `&&` and `!` but I don't feel up to producing an illustrative example right now.) I won’t say that I haven’t wished for conventional binary operators in _simple_ cases, but switching to them completely would have significant disadvantages (as would having a mix of both as standard style, simply due to having two things to learn and the choice of which one to use).

Yes the formatting is strange, but I'd argue that the binary operator makes the `cfg` more *readable* for reviewers because you can be sure it is a chain of `any` rather than `all` at the first glance (spot the difference:

<table><tr><th>Existing syntax</th><td>

```rust
#[cfg(any(
    feature = "foo",
    feature = "bar",
    feature = "baz",
    feature = "quux",
    feature = "magic",
    test
))]
```

</td><td>

```rust
#[cfg(all(
    feature = "foo",
    feature = "bar",
    feature = "baz",
    feature = "quux",
    feature = "magic",
    test
))]
```

</td></tr><tr><th>This RFC</th><td>

```rust
#[cfg(
    (feature = "foo")
        || (feature = "bar")
        || (feature = "baz")
        || (feature = "quux")
        || (feature = "magic")
        || test
)]
```

</td><td>

```rust
#[cfg(
    (feature = "foo")
        && (feature = "bar")
        && (feature = "baz")
        && (feature = "quux")
        && (feature = "magic")
        && test
)]
```

</td></tr><tr><th>Alternative proposal</th><td>

```rust
#[cfg(any(
    feature = "foo" 
        | "bar" 
        | "baz" 
        | "quux" 
        | "magic",
    test,
))]
```
</td><td>

```rust
#[cfg(all(
    feature = "foo" 
        & "bar" 
        & "baz" 
        & "quux" 
        & "magic",
    test,
))]
```

</td></tr></table>

SabrinaJewson:
`cfg(any())` and `cfg(all())` are very useful, often in macros, as always-false and always-true `cfg` values respectively. This RFC does not _need_ to supplant all existing use cases of `all` and `any`, but it does seem weird to replace most things except this one edge case.

Jules-Bertholet:
> `cfg(any())` and `cfg(all())` are very useful, often in macros, as always-false and always-true `cfg` values respectively.

https://github.com/rust-lang/rust/issues/131204

joshtriplett:
@m-ou-se wrote:
> If we do want to make cfg() take something that looks like an expression, we should also change the `feature = "asdf"` syntax to something that fits a Rust expression. E.g. `#[cfg(has_feature("asdf") && has_target_feature("asdf"))]`.

Agreed completely. I don't think it would be ambiguous here to reuse "feature" without adding "has_": `cfg(feature("asdf"))`. And this would also be consistent with other parts of cfg syntax.

kennytm:
> 

> I don't think it would be ambiguous here to reuse "feature" without adding "has_": `cfg(feature("asdf"))`. And this would also be consistent with other parts of cfg syntax.

Technically this change of syntax will conflict with the cfg_version feature but since it is unstable it is fine...? (plus can't use `version="x"` on stable at all, it somehow triggers cfg_version too)

```rust
#![feature(cfg_version)]
fn main() {
	dbg!(cfg!(version("1.69.0")));
	dbg!(cfg!(version("2.0.0")));
	dbg!(cfg!(version = "1.69.0"));
	dbg!(cfg!(version = "2.0.0"));
}
```
```console
$ rustc +nightly --cfg 'version="2.0.0"' 1.rs
$ ./1
[1.rs:3:2] cfg!(version("1.69.0")) = true
[1.rs:4:2] cfg!(version("2.0.0")) = false
[1.rs:5:2] cfg!(version = "1.69.0") = false
[1.rs:6:2] cfg!(version = "2.0.0") = true
```

jdonszelmann:
I feel like many problems raised (by various people) here can be partially mitigated when we were to replace `feature = "something"` with `feature("something")` and equivalently for the other key-values such as `target_arch("x86_64")`  
I didn't come up with this, it felt like a slightly nicer way of writing @m-ou-se's `has_feature` which I like less, and now after having written the rest of this comment and reading back the thread I see that @clarfonthey proposed this quite literally.

Having said that, let this comment be a motivation for *why* I think this mitigates so many of the concerns. The advantages I see are:

1. The confusion about the meaning of  `=` (whether it means `==` or "assignment `=`" or a third kind of `=`) becomes irrelevant.
2. Because we use parentheses, we automatically disambiguate any use of `&&` and `||`. In other words, `#[my_attribute_macro(expr = a && b, expr2 = c || d)]` is instead written as  `#[my_attribute_macro(expr(a) && b && expr2(c) || d)]` Though of course, whether `&&` has precedence over `||` is still here
3. Whether to permit `feature != "a"` is irrelevant. it'd always be `!feature("a")`
4. it's consistent with `any` and `all` which are already parenthesized.
5. If we choose to permit alternation within an attribute (i.e. `feature = "a" | "b" | "c"` which with my proposal becomes `feature("a" | "b" | "c")`) this becomes clearly unambiguous from `feature("a") | b | c`

jhpratt:
I'll defer to the team here. I'm happy to write up whatever changes are desired, which at this point sounds like it may be `feature("foo")`, `key(val1 | val2 | val3)`, etc. If this is the case, I have one question: should we permit mixing and matching or only support all-or-nothing?

tmccombs:
I like the proposal in https://github.com/rust-lang/rfcs/pull/3796#issuecomment-2775430532, but the `feature = "abc"` syntax still needs to be supported for backwords compatibility, so we need to define how that interacts with the new syntax. Do we

1. Disallow the = syntax  in combination with && or ||
2. Require any use of = to be within parentheses, ex. `(feature="abc") && (target_os="linux")`
3. As originally described in this RFC
4. Something else

camsteffen:
I propose that we add `key(val | val)` syntax _without_ adding `&&`, `||`, `!`. That's where the bigger win is, and it would be less churn.

In a world of boolean logic, the existence of any/all/not functions is indubitably justified, and they deserve our respect.

jhpratt:
@camsteffen Is churn the primary (or even sole) concern? I think the :+1:s on this PR show that there's unquestionably the desire for logical operators — it feels obvious imo.

jdonszelmann:
I actually like @camsteffen's proposal a lot. It makes many scenarios simpler right now without changing ther original feature = name syntax. And the existing situation isn't so bad. And, if we feel like adding boolean operators later that still works!

camsteffen:
@jhpratt Not just churn, but the fact of having two ways to express something and a split in the ecosystem of which one is used. The motivation may be to make things simpler by matching other syntax, but I think the reality is that it's increasing complexity with another thing to learn. I would be more open to it if I could snap my fingers and change all the code in the world to the new syntax. But I also agree with @kpreid that any/all is better with regard to formatting.

Jules-Bertholet:
> > I would go so far as to argue that `any()` and `all()` are _better for formatting_. Binary operators used repeatedly have an awkward indentation problem…
> 
> Yes the formatting is strange, but I'd argue that the binary operator makes the `cfg` more _readable_ for reviewers…
>
> […]
>
> #### Alternative proposal	
> 
> ```rust
> #[cfg(any(
>     feature = "foo" 
>         | "bar" 
>         | "baz" 
>         | "quux" 
>         | "magic",
>     test,
> ))]
> ```
>
> ```rust
> #[cfg(all(
>     feature = "foo" 
>         & "bar" 
>         & "baz" 
>         & "quux" 
>         & "magic",
>     test,
> ))]
> ```


A little more indentation makes this look even better:

```rust
#[cfg(any(
    feature = "foo" 
            | "bar" 
            | "baz" 
            | "quux"
            | "magic",
    test,
))]
```

```rust
#[cfg(all(
    feature = "foo" 
            & "bar" 
            & "baz" 
            & "quux" 
            & "magic",
    test,
))]
```

kennytm:
@Jules-Bertholet the default rustfmt will not use vertical alignment.

----

@camsteffen Rust isn't afraid of having multiple ways to express the same concept, we have `for x in it {}` vs `while let Some(x) = t.next() {}`; `async fn f() -> R` vs `fn f() -> impl Future<Output=R>`; `fn g(a: impl Read)` vs `fn g<R: Read>(a: R)` vs `fn g<R>(a: R) where R: Read+,` etc. Even [your proposal](https://github.com/rust-lang/rfcs/pull/3796#issuecomment-2781102724) it going to introduce `#[cfg(feature = "name")]` vs `#[cfg(feature("name"))]`. IMO "splitting the ecosystem" is FUD and is no ground to reject a proposal for Rust.

This proposed syntax `#[cfg(key("val1" | "val2" | "val3"))]` works well when everything involved has the same "key", but won't help when multiple "keys" are involved, e.g.:

<table><thead><tr><td></td><th>

[Exhibit A](https://github.com/rust-lang/rust/blob/e643f59f6da3a84f43e75dea99afaa5b041ea6bf/library/std/src/os/mod.rs#L19-L25)

</td><th>

[Exhibit B](https://github.com/rust-lang/rust/blob/e643f59f6da3a84f43e75dea99afaa5b041ea6bf/library/unwind/src/lib.rs#L28-L34)

</td></tr></thead><tbody><tr><th>

Existing code

</th><td>

```rust
#[cfg(all(
    doc,
    any(
        all(target_arch = "wasm32", not(target_os = "wasi")),
        all(target_vendor = "fortanix", target_env = "sgx")
    )
))]
``` 

</td><td>

```rust
#[cfg(any(
    unix,
    windows,
    target_os = "psp",
    target_os = "solid_asp3",
    all(target_vendor = "fortanix", target_env = "sgx"),
))]
```

</td></tr><tr><th>This RFC</th><td>

```rust
#[cfg(doc && (
    (target_arch = "wasm32") && !(target_os = "wasi") 
        || (target_vendor = "fortanix") && (target_env = "sgx")
))]
```

</td><td>


```rust
#[cfg(unix 
    || windows 
    || (target_os = "psp") 
    || (target_os = "solid_asp3")
    || (target_vendor = "fortanix") && (target_env = "sgx")
)]
```

</td></tr><tr><th>key("value") only</th><td>

```rust
#[cfg(all(
    doc,
    any(
        all(target_arch("wasm32"), not(target_os("wasi"))),
        all(target_vendor("fortanix"), target_env("sgx"))
    )
))]
``` 

</td><td>

```rust
#[cfg(any(
    unix,
    windows,
    target_os("psp" | "solid_asp3"),
    all(target_vendor("fortanix"), target_env("sgx")),
))]
```

</td></tr><tr><th>This RFC + key("value")</th><td>

```rust
#[cfg(doc && (
    target_arch("wasm32") && !target_os("wasi") 
        || target_vendor("fortanix") && target_env("sgx")
))]
```

</td><td>

```rust
#[cfg(unix 
    || windows 
    || target_os("psp" | "solid_asp3")
    || target_vendor("fortanix") && target_env("sgx")
)]
```

</td></tr></tbody></table>

RalfJung:
Oddly, my first reaction to those examples that I'd like to use "any" and "&&"... I can't really explain why. But I do feel like the use of infix operators makes it less clear that we are looking at a list of cases.

RalfJung:
For those that are curious about the "nearly" part: the large family of ML-based languages including OCaml use `!` to mean "load from a reference" (that's a bit like `*` in Rust but not quite). "not" is written `~` in those languages (I think this is intended to be close the `¬`, the typical logical symbol for negation).

kornelski:
I really like this proposal. Not because it shortens some expressions, but because it's a step towards reducing the amount of unique microsyntaxes inside the Rust language.

kornelski:
As for the `=` bikeshed, how about `feature.name`/`target_os.wasi`, pretending that it's a struct with fields for all the options.

programmerjake:
> As for the `=` bikeshed, how about `feature.name`/`target_os.wasi`, pretending that it's a struct with `bool` fields for all the options.

what happens when values aren't a simple number/identifier? e.g. `--cfg=foo="abc def"`: https://rust.godbolt.org/z/bKfKGorbW

kennytm:
> As for the `=` bikeshed, how about `feature.name`/`target_os.wasi`, pretending that it's a struct with `bool` fields for all the options.

A [feature name](https://doc.rust-lang.org/cargo/reference/features.html#the-features-section) may consist of `[_0-9\p{XID_Start}][-+.\p{XID_Continue}]*`, and even the crates.io-restricted set is still `[_0-9a-zA-Z][-+_0-9a-zA-Z]*`. It would be very confusing for the parser to allow [`#[cfg(feature.c++20)]`](https://docs.rs/crate/cxx/latest/features).

Furthermore a custom config `--cfg 'thing="!@#$%^&"'` allows any string literal as the value (including empty string `--cfg 'thing=""'`, which is distinct from `--cfg 'thing'`). I don't think the quotation marks should be omitted.

```rust
#[cfg(feature = "c++20")]
#[cfg(feature("c++20")]
#[cfg(feature."c++20")]
```

programmerjake:
> I don't think the quotation marks should be omitted.

we could do something like TOML paths where you can omit the quotation marks if it's a valid identifier or number (equivalent to the corresponding quoted version) or you can have a string literal with any random text at all (except for commas since I think it's split into separate values using commas...).

so, with `--cfg=a="123" --cfg=b="abc" --cfg=c="123-4,abc"` that works with all of:
* `cfg(a.123)`
* `cfg(a."123")`
* `cfg(b.abc)`
* `cfg(b."abc")`
* `cfg(c."123-4")`
* `cfg(c.abc)`
* `cfg(c."abc")`

as well as all the existing `cfg(a = "123")` syntax.

kornelski:
Without inventing *new* syntax that isn't in Rust, this could also be `feature["#$%@!"]`.

RalfJung:
`key("value")` is not inventing new syntax.

text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,146 @@
...
+| `any(!x \|\| !w, !(y && z))`       | `any(any(not(x), not(w)), not(all(y, z)))`         | `!`, `&&` etc. can be used inside `any`, `all` and `not`                            |
+| `true && !false`                   | `all(true, not(false))`                            | `!`, `&&` etc. can be used on boolean literals (they are syntactically identifiers) |
+| `!accessible(std::mem::forget)`    | `not(accessible(std::mem::forget))`                | `!`, `&&` etc. can be used on `cfg_accessible`                                      |
+| `accessible(std::a \|\| std::b)`   | _syntax error_                                     | … but not inside                                                                    |
+| `!version("1.42.0")`               | `not(version("1.42.0"))`                           | `!`, `&&` etc. can be used on `cfg_version`                                         |
+| `version(!"1.42.0")`               | _syntax error_                                     | … but not inside                                                                    |
```
Urgau, line 59: How would it interact with the `target(os = "linux", ...)` shorthand from [RFC 3239](https://github.com/rust-lang/rfcs/pull/3239)?

kennytm: well it's being removed rust-lang/rust#130780

```rust
#[cfg(all(target_os = "linux", target_arch = "arm"))]
#[cfg(target_os("linux") && target_arch("arm"))]
#[cfg(target(os = "linux", arch = "arm"))]
```

Urgau: Last comments from T-lang members tend to the opposite, so probably going to stay.

tmandry:
Things I'm quite sure I like:

* `&&` instead of `all(...)`. It's too difficult to find a visual distinction between `all` and `any` and I often miss it.
* `feature("string")` instead of `feature = "string"`. This was already confusing to begin with.

Things I'm less sure of but still open to:

* `||` instead of `any(...)`. We should probably allow it for consistency, but I do find myself agreeing with @RalfJung that it ends up looking nicer to use `any()`. The main thing for me is having a clear visual distinction between the two, as I said above, and we technically only need one operator to accomplish that.
* `!`. I think this will be much improved when combined with `feature("string")`. I like that it's "loud" visually, but most times it has to be used like `!()` which never feels that great.
* Incorporating pattern syntax. We might need to extend it beyond `|` to include `&`. This does seem like it could be a win, but it implies to me that we would want to unify these as far as we sensibly could. I'd want to discuss the possible implications of that in more depth.

tgross35:
When deciding upon a pattern here, it's probably worth thinking about the standard library's close-to-stable [`cfg_match!`](https://github.com/rust-lang/rust/issues/115585). It would be great if the same syntax was usable in both `cfg(...)` and `cfg_match!{ ... }`.

kennytm:
since `cfg_match!` is just [implemented as a decl-macro](https://github.com/rust-lang/rust/blob/847e3ee6b0e614937eee4e6d8f61094411eadcc0/library/core/src/macros/mod.rs#L238-L259) that expands to `#[cfg]` i don't see why a syntax is only usable in one place but not the other.

kennytm:
The problem adapting to `cfg_match!` is that `a && b` etc. are not valid `:meta` in a macro. This is not a problem of `cfg_match!` alone, any other decl-macro expecting `#[cfg($cfg:meta)]` will be affected

joshtriplett:
@rust-lang/lang, shall we accept the RFC as currently proposed, for adding `&&`, `||`, and `!` to `cfg` syntax?

As currently written, the RFC defers all other potential additions to future proposals.

@rfcbot merge

jhpratt:
That is my understanding of the current state as well. If accepted, I will almost certainly follow up with a function-like `feature("foo")` proposal as a standalone RFC.

Thanks for moving this forward, Josh!

traviscross:
First, thanks to @jhpratt for putting this together.  The current syntax of `cfg` overall does seem less than ideal, and there's appeal in trying to do better.

Some thoughts:

**One**, I agree with @m-ou-se that, on its own, the degree of improvement here is somewhat limited, with @RalfJung that `any(..)` has some virtues, and with @kpreid that infix is often going to result in worse formatting.

**Two**, I'm certain that `S = e` is a flawed way to represent `e ∈ S`, but I'm uncertain about representing that as `S(e)` here for user-specified `cfg` options (like `feature`, which is just a convention as far as the language is concerned).  Maybe.  But it would mean that later adding any new operator would be a breaking change.  And if we're touching it, it does seem that we should also consider having a concise way to represent "the set `X` contains at least one element of the set `Y` (i.e., `X ∩ Y ≠ ∅`)", as @m-ou-se suggested, and correspondingly "the set `X` contains every element of the set `Y` (i.e. `X ⊇ Y`)".  It seems worth exploring some options for how best to encode those.

That's not directly part of this RFC, of course, but I think probably this RFC makes less sense if we're not headed in a direction like that, so I do maybe want to consider more of this together.

**Three**, this RFC specifies an alternate precedence for `=` in this context.  I understand why it does that, but our story around precedence is already something I'd like to see us try to simplify (with editions) as we go about specifying the language, and so adding a little carve-out here raises a concern for me, including for the reasons that @m-ou-se mentioned.  If we were to otherwise accept this RFC, I'd rather keep normal precedence and require the parens.

@rfcbot concern uncomfortable-with-precedence-carve-out

RalfJung:
> I'm certain that S = e is a flawed way to represent e ∈ S, but I'm uncertain about representing that as S(e) here for user-specified cfg options

FWIW I find `S(e)` to be a quite natural way to represent `e ∈ S`, but that is probably because of my background in type theory where sets are often modeled as predicates ("functions with return type `bool`", basically), and so asking "is this value in the set" is literally a function call.

traviscross:
> FWIW I find `S(e)` to be a quite natural way to represent `e ∈ S`, but that is probably because of my background in type theory where sets are often modeled as predicates ("functions with return type `bool`", basically), and so asking "is this value in the set" is literally a function call.

Yes; I have a similar intuition for it, and agree that it's natural in that way.  I'm just not sure about putting this set-membership predicate in the same syntax space as arbitrary predicates, and thereby making it a breaking change to add any new such predicates.

text/0000-cfg-logical-ops.md
```diff
@@ -0,0 +1,146 @@
...
+| `(a \|\| b) && (c \|\| d)`         | `all(any(a, b), any(c, d))`                        | `()` can be used for grouping                                                       |
+| `!a \|\| !b && !c`                 | `any(not(a), all(not(b), not(c)))`                 | `!` has highest precedence                                                          |
+| `feature="foo" \|\| feature="bar"` | `any(feature="foo", feature="bar")`                | `\|\|` has lower precedence than `=`                                                |
+| `feature="foo" && feature="bar"`   | `all(feature="foo", feature="bar")`                | `&&` has lower precedence than `=`                                                  |
+| `!feature="foo"`                   | _syntax error_                                     | `!` has higher precedence than `=`, which may be confusing, so we ban this syntax   |
+| `!(feature="foo")`                 | `not(feature="foo")`                               | use `()` for grouping                                                               |
```
tgross35: (starting an inline discussion for brainstorming `key = "val"` syntax stemming from https://github.com/rust-lang/rfcs/pull/3796#issuecomment-2993831077)

> > FWIW I find `S(e)` to be a quite natural way to represent `e ∈ S`, but that is probably because of my background in type theory where sets are often modeled as predicates ("functions with return type `bool`", basically), and so asking "is this value in the set" is literally a function call.
> 
> Yes; I have a similar intuition for it, and agree that it's natural in that way. I'm just not sure about putting this set-membership predicate in the same syntax space as arbitrary predicates, and thereby making it a breaking change to add any new such predicates.

Could we add a new named predicate for this? `set(feature, "foo")` is terse and could work, though I don't love that the intuition here would probably indicate a 1:1 mapping. `contains` or `includes` are probably how we would spell this on a collection, but I think `contains(feature, "foo")` sounds a bit worse than plural would be `contains(features, "foo")`.

jhpratt: I'll have an RFC together soon, though deciding what the syntax should be will definitely be better before it's fully written.

Random thought, which admittedly isn't very Rust-like, is `feature?("foo")`, along with `feature?("foo" | "bar")` and `feature?("foo" & "bar")`. There is some precedence in Ruby where `?` means a function that returns a boolean.

traviscross: > There is some precedence in Ruby where `?` means a function that returns a boolean.

It comes from Scheme, which uses the same convention for predicates, e.g. `odd?`.

jmjoy:
Simple Boolean operations are one thing, but what I’m really hoping for is the ability to compare values, something that’s quite common in C macros, like:

`#if FOO > 1000`

But in Rust, there’s currently no way to express this, which leads to some rather awkward workarounds.

Right now, the values in Rust’s `cfg` are all strings, so it seems difficult to support something like:

`#[cfg(FOO > 1000)]`

jhpratt:
@jmjoy That would fundamentally change what `cfg` is.

text/0000-cfg-logical-ops.md
kornelski: To me the important effect of adding `&&`, etc. to `cfg` is making it more consistent with the regular Rust syntax, so that it's less of a second micro language within a language. 

Using `?` in cfg in a way that is different from regular Rust code would be just another version of the same problem of using `=` in cfg in a way that is different from regular Rust code.

`feature("val")` is not too bad. It's probably the tersest syntax that regular Rust code could implement for the same purpose.

For multiple values, maybe something like this:

`feature(any("val1", "val2"))` 

This would also be implementable in a regular Rust syntax with a bit of generics.

clarfonthey: One potential issue is that `feature(..)` is currently used for enabling language features, whereas `feature = ".."` is used for checking crate features. Not really sure how to reconcile this, honestly, but worth pointing out that making the syntax closer to `feature(..)` may not be best.

kornelski: Rust typically uses `foo()` for getters and `set_foo()` for setters, and `cfg()` is a context that is querying the config rather than setting it, so to be this doesn't look like a setter. 

`has_feature()` would be even clearer.

joshtriplett:
There have been proposals for key-value features, and I'd expect such a proposal to talk about how to use them in `cfg`, but that's off-topic for *this* RFC.

Jules-Bertholet:
It would be nice if this RFC also specified a new `macro_rules!` fragment specifier to match an arbitrary `cfg` predicate, for use by macros like `cfg_if`. Currently, you can use `meta`, which excludes only `true` and `false`, but with this RFC, that would become completely inadequate.

traviscross:
> It would be nice if this RFC also specified a new `macro_rules!` fragment specifier to match an arbitrary `cfg` predicate, for use by macros like `cfg_if`. Currently, you can use `meta`, which excludes only `true` and `false`, but with this RFC, that would become completely inadequate.

That's a good point.  This document should, I think, discuss the effect of this on macro fragment specifiers and propose solutions, perhaps using our guidance in https://github.com/rust-lang/rfcs/pull/3531, if appropriate.

@rfcbot concern discuss-effect-on-macro-fragment-specifiers

jhpratt:
`meta` already works, though, as it matches the _entire_ attribute inside the brackets, not just what's inside `cfg()`. [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e7f9759194868b40c1e68a567e274f2b) for proof.

Sure, things like `cfg_if` and `cfg_select` need to work around some things because they match on the inner parts, but isn't that already the case?

Jules-Bertholet:
> Sure, things like `cfg_if` and `cfg_select` need to work around some things because they match on the inner parts, but isn't that already the case?

Right now, the only thing they “need to work around” is `true`/`false`:

- `cfg_if` simply [doesn’t support them](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d314684c216d5241bc19e8eb8e5e5660).
- `cfg_select` in the stdlib [doesn’t use `macro_rules!`](https://doc.rust-lang.org/src/core/macros/mod.rs.html#238).
- https://github.com/rust-lang/rust/pull/146281 (a PR I wrote that modifies the std `thread_local` such that it parses `#[cfg_attr(…)]`) uses [extra](https://github.com/rust-lang/rust/blob/1bfbfbfa0e41e5d7fb9a257199eecb7d828de1f1/library/std/src/sys/thread_local/os.rs#L75-L91) [match](https://github.com/rust-lang/rust/blob/1bfbfbfa0e41e5d7fb9a257199eecb7d828de1f1/library/std/src/thread/local.rs#L196-L224) [arms](https://github.com/rust-lang/rust/blob/1bfbfbfa0e41e5d7fb9a257199eecb7d828de1f1/library/std/src/thread/local.rs#L266-L284) just for `true` and `false`, which is annoying[^1] but doable. Supporting this RFC’s syntax, in contrast, would require complex TT-munching, which is *really* unappealing, especially given that the macro has far too much of that going on already.

[^1]: If this RFC were to be rejected, I would probably propose changing the `meta` specifier to accept `true` and `false`, in order to address this issue—it would be a backward-compatible change that would not require an edition. But if this RFC is accepted, a dedicated `cfg` specifier would make more sense, and there wouldn’t be much point in changing `meta`.

jhpratt:
How common are macros that need to handle the _internal_ part of `cfg`? Ultimately that's what we're talking about here. The fact that `meta`, prior to `true`/`false`, matched the internal part perfectly was largely a historical artifact from before arbitrary tokens could be accepted. That restriction was lifted in April 2019.

Essentially, my argument is that this scenario is sufficiently uncommon that it is not worth the added language complexity for a rarely used macro fragment specifier.

Jules-Bertholet:
> How common are macros that need to handle the _internal_ part of cfg?

I just cited three examples.

If you feel that defining an entirely new specifier is too much complexity, one alternative would be to re-use `expr`.

jhpratt:
> I just cited three examples.

I meant in general, everyday code. Defining macros like that is not common, presumably, hence my question. My belief is that _manipulating_ the contents of `cfg` is quite rare, as nearly all cases will want to merely pass the contents on to the output, for which purpose `meta` or `$(x:tt)*` would suffice. All existing macro fragment specifiers are used with sufficient frequency (or are fundamentally required due to "follows" ambiguity) that having one here seems overly niche.

For what it's worth, `meta` does not appear to have been intended to be used in this manner in the first place; it seems to be a historical quirk that the internal syntax followed the same pattern as the attribute as a whole.

Jules-Bertholet:
Actually, I realized that `cfg_if` doesn’t need a dedicated fragment specifier, it can just switch to `tt`: https://github.com/rust-lang/cfg-if/pull/99. That wouldn’t work for `cfg_match` or https://github.com/rust-lang/rust/pull/146281, however.

tgross35:
See https://github.com/rust-lang/rust/issues/145060 for previous discussion on `meta` matching `true` and `false`. Quoting @petrochenkov:

> `meta` is not supposed to match syntax of `cfg` predicates, or other syntaxes that other attributes may define, it matches the stuff inside the attribute brackets `#[...]`.
> 
>     * `#[attr]` - ok, `attr` is `meta`
>     * `#[attr { a b c }]` - ok, `attr { a b c }` is `meta`
>     * `#[true]` - not ok, `true` is not `meta`

Jules-Bertholet:
> one alternative would be to re-use `expr`.

Increasingly I am thinking that this is probably the best solution. `expr` already accepts all the syntax in this RFC, the restriction on its use in `cfg` simply needs to be lifted.

traviscross:
@rfcbot resolve discuss-effect-on-macro-fragment-specifiers

Jules-Bertholet:
> Increasingly I am thinking that this is probably the best solution. `expr` already accepts all the syntax in this RFC, the restriction on its use in `cfg` simply needs to be lifted.

https://github.com/rust-lang/rust/pull/146961

jhpratt:
Regarding precedence, would it be easier to pass the RFC if instead we required `(feature = "foo")`, erroring on the unparenthesized version? That would avoid the precedence carveout while being forward compatible with anything we want.

traviscross:
What would be an example of something that would not be accepted under that rule but would be accepted with `=` having its normal precedence?  I.e., is there any case where giving `=` its normal precedence would not already force the parentheses?

jhpratt:
I believe they would be identical — that's what I meant in my previous comment. Eliminate the carveout and require parentheses.

traviscross:
That would indeed resolve my concern.

jhpratt:
@traviscross I've added a commit removing the precedence carveout while also adding a couple notes.

traviscross:
Thanks.  That resolves this concern:

@rfcbot resolve uncomfortable-with-precedence-carve-out

traviscross:
Since the proposal has now changed, and as there are no outstanding filed concerns to track, for procedural reasons, I'm going to cancel the proposed FCP.

@rfcbot fcp cancel

I'm not going to restart immediately, since I personally want to see other work, such as @tmandry's upcoming `cfg_version` RFC, play out a bit first, but anyone else on the team who wants to propose a new FCP can of course do so.

jhpratt:
A year on from when FCP was initially proposed and 8 months out from the most recent comment, what does @rust-lang/lang think of this? There has been time to consider interaction with `cfg_version`, hopefully.

PR #3818, opened 2025-05-24
RFC: New method for `str`: `str::chunks(usize)`

tkr-sh:
tldr;`slice::chunks` but for `&str` with chunks being substrings as `&str` of `chunk_size` length.


tkr-sh:
yes!
> Given the issues related to UTF-8 boundaries causing potential foot-guns [...]

So this RFC doesn't have the same "problems" as the previous one

ehuss:
I'm going to close in favor of https://github.com/rust-lang/libs-team/issues/592.

PR #3820, opened 2025-05-25
RFC: changes to target_feature attribute

veluca93:
tldr: make it easier/safer to use `target_feature` in traits.


veluca93:
@hanna-kruppe @traviscross @RalfJung

Correct RFC link here :-)

text/3820-target-feature-traits.md
```diff
@@ -0,0 +1,37 @@
...
+- allows to use target features in safe trait methods
+- disallows using features in a trait impl that were not expected in the trait definition
+
+For the second case, we plan to emit a lint and to make this into a hard-error in a future edition.
+
+### New `#[unsafe(target_feature(force = "x"))]` attribute
+
+This `unsafe` attribute can be applied to free functions, trait method implementations or trait method definitions.
+
+It comes with the following soundness requirement: the a function the signature of the function the attribute is applied to must _only be callable if the force-enabled features are guaranteed to be present_ (this can be done, for example, if the function takes arguments that carry appropriate safety invariants).
```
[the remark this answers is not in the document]
hanna-kruppe: I get that the details of the compiler establishing safety of extra target features via arguments' types are very tricky. But it would be extremely unfortunate to first drag the ecosystem through the churn of a FCW, and then one edition later deprecate the style that the FCW promoted in favor of something new that solves the same use case better. (Of course, it would also be unfortunate to block any improvement here on some speculative far-future language evolution.)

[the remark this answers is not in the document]
calebzulawski: My 2 cents-- @veluca93 and I had a (very) long discussion comparing all of the options on Zulip a while back.

This RFC is probably the most minimal change that opens up the most new possibilities for tying target features to types. Unlike #3525, this RFC maintains the premise that target features are tied to the codegen of specific functions. #3525 also introduces questions about whether generic functions (in general) should have variable ABIs specified by their parameters, whether codegen attributes in general would have to be deprecated in favor of types for consistency, and so on.

#3525 can be broken down into a few somewhat separable use cases:
* target feature proof types
* static dispatch of multiversioned functions from said proof type
* a target feature effect system

This RFC mostly enables the static dispatch component (in addition to minimal support for proof types, and simply making target features in traits a lot more ergonomic). The effect system is the most controversial, most difficult to implement, and also not essential to providing the other features. My personal preference would be to accept this RFC and implement the remaining features in a way that complements this RFC.

text/3820-target-feature-traits.md
```diff
@@ -0,0 +1,37 @@
...
+
+Compared to the current status, this:
+- allows to use target features in safe trait methods
+- disallows using features in a trait impl that were not expected in the trait definition
+
+For the second case, we plan to emit a lint and to make this into a hard-error in a future edition.
+
+### New `#[unsafe(target_feature(force = "x"))]` attribute
+
+This `unsafe` attribute can be applied to free functions, trait method implementations or trait method definitions.
```
calebzulawski, line 27: This might benefit from some examples, e.g. proof types:

```rust
struct Avx;

impl Avx {
    fn new() -> Option<Self> {
        if is_x86_feature_detected!("avx") {
            Some(Self)
        } else {
            None
        }
    }
}

#[unsafe(target_feature(force = "avx"))]
fn requires_avx(_proof: Avx) {
    // AVX is enabled, but can be called with an entirely safe abstraction
}
```
An expanded version with static dispatch depending on feature would be nice, too

traviscross: From our earlier thread in https://github.com/rust-lang/rust/issues/139368, here's a worked example of what can be done today, demonstrating the relevant safety analysis:

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=%2F%2F+This+demonstrates+a+pattern+for+using+%60target_feature%60+on+functions%0A%2F%2F+in+a+trait+impl+in+a+sound+way+along+with+a+proposed+safety%0A%2F%2F+analysis.%0A%2F%2F%0A%2F%2F+The+key+factor%2C+in+my+view%2C+is+that+despite+the+trait+function%0A%2F%2F+being+marked+as+%60unsafe%60%2C+it+must+actually+be+always+safe+to+call%0A%2F%2F+%28at+least+as+far+as+%60target_feature%60+is+concerned%3B+it+could+have%0A%2F%2F+unrelated+preconditions%29.++If+this+were+not+the+case%2C+then+it+would%0A%2F%2F+need+to+obligate+the+caller+to+never+call+it+in+a+generic+context%2C%0A%2F%2F+and+that+wouldn%27t+be+a+very+appealing+kind+of+precondition+for+us%0A%2F%2F+at+a+language+level.%0A%2F%2F%0A%2F%2F+Since+it+is+always+safe+to+call%2C+the+implementor+must+ensure+to+not%0A%2F%2F+add+any+preconditions.++This+is+an+inherent+requirement+for+trait%0A%2F%2F+implementors%3B+the+trait+does+not+need+to+state+this+in+its+safety%0A%2F%2F+contract.++Here%2C+that+means+that+the+implementor+must+take%0A%2F%2F+precautions+to+ensure+that%2C+when+using+%60target_feature%60%2C+the%0A%2F%2F+function+can+only+be+called+when+that+feature+is+available+%28or+when%0A%2F%2F+someone+has+unsafely+asserted+that+it+is%29.%0A%2F%2F%0A%2F%2F+See%3A%0A%2F%2F%0A%2F%2F+-+https%3A%2F%2Fgithub.com%2Frust-lang%2Frust%2Fissues%2F139368%23issuecomment-2906958561%0A%2F%2F%0A%2F%2F+Author%3A+TC%0A%2F%2F+Date%3A+2025-05-24%0A%0Apub+trait+Tr+%7B%0A++++%2F%2F%2F+SAFETY%3A+This+is+always+safe+to+call.%0A++++%2F%2F%0A++++%2F%2F+SAFETY%28internal%29%3A+We+mark+this+function+as+unsafe+because+when%0A++++%2F%2F+implementing+this+function+we+need+to+apply+the%0A++++%2F%2F+%60target_feature%60+attribute%2C+and+that+is+only+allowed+on+trait%0A++++%2F%2F+functions+when+the+function+is+marked+unsafe.%0A++++%2F%2F%0A++++%2F%2F+Nonetheless%2C+this+is+always+safe+to+call%2C+because+a+valid%0A++++%2F%2F+implementation+of+this+function+cannot+strengthen+the+safety%0A++++%2F%2F+requirements%2C+and+so+when+applying+%60target_feature%60%2C+it+must%0A++++%2F%2F+ensure+that+it%27s+only+possible+to+call+the+function+when+that%0A++++%2F%2F+target+feature+is+available.%0A++++unsafe+fn+f%28%26self%2C+xs%3A+%26mut+%5Bu64%3B+1024%5D%29%3B%0A%7D%0A%0Apub+struct+Fallback%3B%0A%0Aimpl+Fallback+%7B%0A++++pub+fn+new%28%29+-%3E+Self+%7B%0A++++++++Self%0A++++%7D%0A%7D%0A%0Aimpl+Tr+for+Fallback+%7B%0A++++%2F%2F%2F+SAFETY%3A+This+is+still+always+safe+to+call.%0A++++unsafe+fn+f%28%26self%2C+xs%3A+%26mut+%5Bu64%3B+1024%5D%29+%7B%0A++++++++xs.iter_mut%28%29.for_each%28%7Cx%7C+%2Ax+%3D+x.wrapping_add%281%29%29%3B%0A++++%7D%0A%7D%0A%0A%2F%2F+SAFETY%3A+We+must+ensure+that+a+value+of+this+type+cannot+be%0A%2F%2F+constructed+unless+the+%60avx2%60+target+feature+is+available+%28or%0A%2F%2F+someone+has+unsafetly+asserted+that+it+is%29.%0Apub+struct+Avx2%28%28%29%29%3B%0A%0Aimpl+Avx2+%7B%0A++++pub+fn+new%28%29+-%3E+Self+%7B%0A++++++++%2F%2F+SAFETY%3A+This+check+is+what+is+actually+upholding+our%0A++++++++%2F%2F+obligation+to+the+trait+below.%0A++++++++if+%21is_x86_feature_detected%21%28%22avx2%22%29+%7B%0A++++++++++++panic%21%28%29%3B%0A++++++++%7D%0A++++++++Self%28%28%29%29%0A++++%7D%0A%7D%0A%0Aimpl+Tr+for+Avx2+%7B%0A++++%2F%2F%2F+SAFETY%3A+This+is+still+always+safe+to+call.%0A++++%2F%2F%0A++++%2F%2F+SAFETY%28internal%29%3A+We%27re+applying+%60target_feature%60+to+this+impl%0A++++%2F%2F+item%2C+so+to+uphold+our+obligation+per+the+trait+item%27s+safety%0A++++%2F%2F+contract+that+this+must+always+be+safe+to+call%2C+we+must+ensure%0A++++%2F%2F+that+it+cannot+be+called+unless+the+target+feature+is%0A++++%2F%2F+available.++Since+a+%60Self%60+value+must+be+available+to+call+this%0A++++%2F%2F+function%2C+we+do+this+by+ensuring+that+%60Self%60+can+only+be%0A++++%2F%2F+constructed+when+the+target+feature+is+available.%0A++++%23%5Btarget_feature%28enable+%3D+%22avx2%22%29%5D%0A++++%2F%2F~%5E+This+attribute+should+be+unsafe%2C+as+we+need+to+discharge+an%0A++++%2F%2F+obligation+at+this+point%2C+as+per+the+above+safety+comment.%0A++++unsafe+fn+f%28%26self%2C+xs%3A+%26mut+%5Bu64%3B+1024%5D%29+%7B%0A++++++++%2F%2F+There%27s+no+%60unsafe%60+here%2C+but+the+generated+code+is+likely%0A++++++++%2F%2F+to+use+AVX2+instructions.%0A++++++++xs.iter_mut%28%29.for_each%28%7Cx%7C+%2Ax+%3D+x.wrapping_add%281%29%29%3B%0A++++%7D%0A%7D%0A%0Afn+caller%3CT%3A+Tr%3E%28s%3A+%26T%2C+xs%3A+%26mut+%5Bu64%3B+1024%5D%29+%7B%0A++++%2F%2F+SAFETY%3A+Per+the+trait%27s+contract+for+%60f%60%2C+this+function+is%0A++++%2F%2F+always+safe+to+call.%0A++++unsafe+%7B+s.f%28xs%29+%7D%3B%0A%7D%0A%0Afn+main%28%29+%7B%0A++++let+mut+xs+%3D+%5B0%3B+1024%5D%3B%0A++++caller%28%26Fallback%3A%3Anew%28%29%2C+%26mut+xs%29%3B%0A++++caller%28%26Avx2%3A%3Anew%28%29%2C+%26mut+xs%29%3B%0A++++dbg%21%28xs.iter%28%29.sum%3A%3A%3Cu64%3E%28%29%29%3B%0A%7D%0A)

text/3820-target-feature-traits.md
```diff
@@ -0,0 +1,37 @@
...
+
+It comes with the following soundness requirement: a function with the signature of the function the attribute is applied to must _only be callable if the force-enabled features are guaranteed to be present_ (this can be done, for example, if the function takes arguments that carry appropriate safety invariants).
+
+Because of the soundness requirement, applying this attribute does not impose additional the requirements for calling this function on the callers (which is why the attribute can be applied to arbitrary trait methods).
```
obi1kenobi, line 31: typo fix:
```suggestion
Because of the soundness requirement, applying this attribute does not impose additional requirements for calling this function on the callers (which is why the attribute can be applied to arbitrary trait methods).
```

text/3820-target-feature-traits.md
obi1kenobi: From the playground link:
```
// The key factor, in my view, is that despite the trait function
// being marked as `unsafe`, it must actually be always safe to call
// (at least as far as `target_feature` is concerned; it could have
// unrelated preconditions).  If this were not the case, then it would
// need to obligate the caller to never call it in a generic context,
// and that wouldn't be a very appealing kind of precondition for us
// at a language level.
```

I'm curious if you see the "it must always be safe to call" part as specific to the example, or whether it should apply in general?

I didn't quite understand the connection between having additional safety requirements and using it in a generic context, nor whether in that case we're discussing the trait or function being generic (or possibly both).

text/3820-target-feature-traits.md
```diff
@@ -0,0 +1,37 @@
...
+
+The effect of the attribute is otherwise equivalent to `#[target_feature(enable = "x")]`.
+
+Note that this version of the attribute can be used in the use-cases where `#[target_feature(enable = "x")]` would no longer be allowed.
+
+Also note that the safety requirements of `#[unsafe(target_feature(force = "x"))]` only depend on the function's *signature*. This implies that an implementation (whether it is overriding a provided method or implementing a required method) should also be allowed to enable a feature (without needing to discharge any further safety requirements) if the corresponding definition had the `#[unsafe(target_feature(force = "x"))]` applied to it.
```
obi1kenobi, line 37: > without needing to discharge any further safety requirements

Just checking: I believe this is intended to refer specifically to target-feature requirements, not all safety requirements in general?

An example may be good here as well.

obi1kenobi (commented):
Broadly, I'm in favor of this. I think it addresses the problem in https://github.com/rust-lang/rust/issues/139368 nicely. Thanks for putting it together!

text/3820-target-feature-traits.md
```diff
@@ -0,0 +1,37 @@
...
+- allows to use target features in safe trait methods
+- disallows using features in a trait impl that were not expected in the trait definition
+
+For the second case, we plan to emit a lint and to make this into a hard-error in a future edition.
+
+### New `#[unsafe(target_feature(force = "x"))]` attribute
+
+This `unsafe` attribute can be applied to free functions, trait method implementations or trait method definitions.
+
+It comes with the following soundness requirement: a function with the signature of the function the attribute is applied to must _only be callable if the force-enabled features are guaranteed to be present_ (this can be done, for example, if the function takes arguments that carry appropriate safety invariants).
```
DJMcNab, line 29: I'd like for this line to be weakened slightly; I believe it would also be fine for the safety invariant to be that these functions are only called in contexts where the target feature is know to be present.
So for reachable `pub` functions, that would be equivalent to this stricter condition, but intra-module (and especially intra-function), the weaker condition would better reflect what users want.

For example, if I write a macro which wants to generate:

```rust
#[unsafe(target_feature(force = "avx2"))]
fn do_my_thing(arg1: __mm128) -> __mm128 {
    // User provided code
}
if is_x86_target_feature_enabled("avx2") {
     do_my_thing(user_provided_arg1);
}
```

Where the safety here is guaranteed because `do_my_thing`  cannot be called except in the block which performs the check due to macro hygiene.
This is of course equivalent to:

```rust
struct Token;
#[unsafe(target_feature(force = "avx2"))]
fn do_my_thing(token: Token, arg1: __mm128) -> __mm128 {
    // User provided code
}
if is_x86_target_feature_enabled("avx2") {
     do_my_thing(Token, user_provided_arg1);
}
```

Which meets this stricter condition because it's impossible to name `Token` outside of the macro, and the macro only constructs `Token` once the features have been checked. But that's pointless busiwork only introduced by this condition being so strict.

--- 

For context, this is thinking about this feature in the context of Fearless SIMD, specifically https://github.com/linebender/fearless_simd/pull/108. That currently cannot currently allow the addition of arbitrary attributes to the generated function (e.g. if the user wanted to add `#[inline(never)]`, `#[track_caller]`, `#[allow/expect]`, etc.) safely - the user could always add another `#[target_feature(enable = "...")]`, breaking the safety guarantees.

For the specific example in the macro-generated code of https://github.com/linebender/fearless_simd/pull/108, we already have a suitable value of the equivalent `Token` to hand, so it isn't actually any cost to the requirement for that PR. But I don't see any reason why the much simpler macro generating the first code block shouldn't be useful and wouldn't be valid.

---

This being weaker would need a little bit more care - in particular, it would need to be written to exclude [`#[panic_handler]`](https://doc.rust-lang.org/reference/panic.html#the-panic_handler-attribute) and `main`. This would be similar to [`[attributes.codegen.target_feature.allowed-positions]`](https://doc.rust-lang.org/reference/attributes/codegen.html#r-attributes.codegen.target_feature.allowed-positions), which I don't think would implicitly apply to `unsafe(target_feature)` (because this RFC introduces `#[unsafe(target_feature)]` as a "new" attribute, rather than as an extension to the old attribute).

DJMcNab (commented):
As a maintainer of [Fearless SIMD](https://github.com/linebender/fearless_simd), a SIMD abstraction library, I believe this RFC would be a significant improvement for us - it would make generic SIMD libraries significantly more auditable. Currently, we have huge numbers of unsafe blocks, which aren't actually doing anything interesting, they're just telling the compiler that we know *a* target features is enabled, but we can't tell you which one. And as that `unsafe`  is an "all or nothing" situation, the compiler doesn't give us the tools to validate that:

1) We haven't called an "actually unsafe" intrinsic, or if we have that we've met the safety conditions
2) The intrinsic we've called is actually justified by the level which we know is available (this is tracked in https://github.com/linebender/fearless_simd/issues/124).

In my mind, this is the natural extension from target feature 1.1. I presume the next step from our side to help move this forward would be experimentally porting to the version of this which is currently in nightly? (That is of course presuming that it's not  since been removed and I haven't noticed).

(As an aside, it isn't entirely clear to me why that is under `#![feature(effective_target_features)]` rather than the `#![feature(target_feature_traits)]` of this RFC. I have to presume that must be because this RFC hasn't landed yet, and adding a feature gate either needs a lang experiment or a merged RFC? Edit: Implemented in https://github.com/rust-lang/rust/pull/145573)

PR #3823, opened 2025-05-29
add edit link to each page

tshepang:


Noratrieb:
I don't think this is a good idea. While we generally take small fixes to RFC text, they are generally not intended to be edited, so I wouldn't want to encourage it.

traviscross:
@ehuss and I talked about this, and we found ourselves indifferent to it, so we're happy to go with @Noratrieb and close it out.

PR #3829, opened 2025-06-02
de-RFC: Remove unsized_locals

Noratrieb:
[Original RFC](https://github.com/Noratrieb/rfcs/blob/ununsize_locals/text/1909-unsized-rvalues.md)

[New RFC](https://github.com/Noratrieb/rfcs/blob/ununsize_locals/text/3829-de-rfc-unsized-locals.md)

[Rendered](https://github.com/Noratrieb/rfcs/blob/ununsize_locals/text/3829-de-rfc-unsized-locals.md)

### DIFF
--- a/text/1909-unsized-rvalues.md
+++ b/text/1909-unsized-rvalues.md
@@ -5,7 +5,7 @@
 
 > ⚠ Update 8 years later ⚠
 >
-> [The team decided to un-accept](https://github.com/rust-lang/rfcs/pull/3829) this RFC and the feature has been removed.
+> [The team decided to un-accept](https://github.com/rust-lang/rfcs/pull/3829) the portions of this RFC not related to function parameters and the `unsized_locals` feature has been removed.
 
 
 # Summary
--- a/text/1909-unsized-rvalues.md
+++ b/text/1909-unsized-rvalues.md
@@ -3,6 +3,11 @@
 - RFC PR: [rust-lang/rfcs#1909](https://github.com/rust-lang/rfcs/pull/1909)
 - Rust Issue: [rust-lang/rust#48055](https://github.com/rust-lang/rust/issues/48055)
 
+> ⚠ Update 8 years later ⚠
+>
+> [The team decided to un-accept](https://github.com/rust-lang/rfcs/pull/3829) this RFC and the feature has been removed.
+
+
 # Summary
 [summary]: #summary
 
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -95,6 +95,8 @@ The main difference here is that recursion is a lot harder to do on accident.
 When recursion is used, it is usually used on purpose, and while sometimes the potential for stack overflows is overlooked, the general feature is usually used on purpose.
 That said, recursion can certainly be dangerous in some contexts, but prior existing features are not a good reason to introduce more ways to blow the stack.
 
+The Linux kernel has spent a lot of time on getting VLAs (C's cousin to `unsized_locals`) [removed from the codebase](https://www.phoronix.com/news/Linux-Kills-The-VLA).
+
 # Guide-level obfuscation
 [guide-level-obfuscation]: #guide-level-obfuscation
 
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -1,0 +1,146 @@
+- Feature Name: `unsized_locals`
+- Start Date: 2025-06-02
+- RFC PR: [rust-lang/rfcs#3829](https://github.com/rust-lang/rfcs/pull/3829)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+
+_Following the great success of the [type ascription de-RFC], we now present the next one in the series!_
+
+# Summary
+[summary]: #summary
+
+Unsized locals ([RFC 1909], called "unsized rvalues" originally)
+has been a merged RFC for eight years with no clear path to stabilization.
+
+There is still a very large gap in the implementation in rustc that hasn't been adressed for many years and there are several language problems with this feature. 
+
+This RFC intends to advocate for the feature being removed entirely with a fresh RFC being necessary to add a similiar feature again.
+
+# Demotivation
+[demotivation]: #demotivation
+
+The `unsized_locals` feature is simple to explain on the surface: local variables no longer have to be `Sized`.
+
+```rust
+#![feature(unsized_locals)]
+fn main() {
+    let x = *Box::from("hello, world!");
+}
+```
+
+This will dynamically allocate space on the stack for the string.
+C has a similiar feature, [`alloca`] and variable length arrays (VLA), the latter having been made optional to implement in C11.
+
+## Lack of Proper Implementation
+
+This feature has never been properly implemented in rustc.
+The variable length array form proposed in the RFC still doesn't exist at all, only the unsized local variable does.
+It is implemented in the type checker and codegen, but lacking MIR semantics and therefore unimplemented in the compile time function evaluator.
+This is very significant, as MIR semantics govern how the feature should behave precisely in the first place.
+Without them, they cannot work in `const` and optimizations are likely broken around them.
+Because of this lack of implementation quality, the `unsized_locals` feature was already accepted for removal from rustc two years ago in [MCP 630].
+This removal hasn't yet been implemented.
+
+## Implicit Danger
+
+While lack of implementation quality is a sign of lack of interest for the feature, it is not the primary reason for this RFC,
+which is purely about the langauge design of the feature.
+
+The original RFC was very short, and especially short on motivation and rationale for the design.
+
+Dynamic stack allocation in geneneral has a rather significant downside: it makes it easy to accidentally overflow the stack if you allocate a lot.
+Stacks are usually rather small (on the order of few megabytes, depending on the platform), which means that dynamically allocating user-controlled input on the stack is often rather dangerous.
+While stack overflows are not considered memory unsafe by Rust, they still cause crashes which can lead to denial of service vulnerabilities and unrealiableness in general.
+Dynamic stack allocation as currently implemented interacts especially poorly with loops.
+Allocations are not freed until the function returns, so the following example overflows the stack:
+
+```rust
+#![feature(unsized_locals)]
+fn main() {
+    let s = "A".repeat(1000);
+    for i in 0..1000000 {
+        let x: str = *Box::from(s.as_str());
+        std::hint::black_box(&s);
+    }
+}
+```
+There are ways around this (rewinding the stack pointer at the end of the loop to free up the memory), but they are not currently implemented.
+
+Dynamic stack allocation also has its upsides.
+It is generally faster than heap allocation and can therefore improve performance in cases where the previously mentioned downsides are not a concern.
+Therefore, this RFC is not necessarily a rejection of the idea of dynamic stack allocation, but merely the way the `unsized_locals` feature exposes it.
+
+Which is where we get to the major argument of this RFC: The `unsized_locals` feature integrates far too naturally into Rust.
+This makes the feature very **easy to use**, and especially **easy to use accidentally**.
+
+As previously mentioned, dynamic stack allocation can be dangerous and should not be used lightly.
+It's an advanced optimization feature that is best left untouched by default.
+As such, it behaves similarly to `unsafe` (but is not actually `unsafe`).
+With `unsized_locals`, the use of dynamic stack allocation is completely implicit.
+When you create an unsized local, it is often not obvious that dynamic stack allocation is happening.
+In the example from the start, we need to be aware of all the involved types (which are often inferred in practice) to know that this is a potentially problematic unsized local that we have to audit more carefully instead of a normal sized local.
+Especially around strings, which are often user-controlled, this easily lead to accidental dangerous situations.
+
+Rusts strings, reference types, and `Sized` are a part of the language that can often be hard to understand for beginners coming from garbage collected languages.
+By allowing people to create a dynamic stack allocation without being aware of what is happening, we open the doors for people, especially new Rust programmers who are not intimately familiar with the tradeoffs of dynamic stack allocation, to shoot themselves in the foot and become vulnerable to unexpected crashes.
+
+As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_.
+Therefore, `unsized_locals` must go.
+
+# Guide-level obfuscation
+[guide-level-obfuscation]: #guide-level-obfuscation
+
+The `unsized_locals` feature is removed from the compiler and [RFC 1909] is officially unaccepted.
+
+If someone wants to bring dynamic stack allocation into Rust again, a new design will have to be designed from scratch, considering all the problems laid out in this RFC.
+
+# Drawforwards
+[drawforwards]: #drawforwards
+
+This feature has the previously mentioned performance upsides that users could profit from if it was stabilized.
+But this benefit applies to other ways to expose dynamic stack allocation too, and other ways are likely to be easier to implement correctly and stabilize.
+
+# Irrationale and alternatives
+[irrationale-and-alternatives]: #irrationale-and-alternatives
+
+If nothing is done on the language side, the feature will likely still be removed from the compiler.
+This puts the feature into a really bad position, but it may be readded in the future if someone desires.
+With this RFC, the fate of `unsized_locals` is sealed and it becomes clear to anyone what the state of the feature is.
+
+The intent of this RFC is not to proposed an alternative way to solve dynamic stack allocation,
+but there are some listed alternatives here that may be considered in the future if desired.
+While this RFC doesn't explicitly encourage people to revisit this topic, it may result in new activity around dynamic stack allocation in Rust.
+
+[RFC 1808] proposed an `alloca` function, which was rejected because `alloca` does not really behave like a function.
+
+[RFC 1808] then changed to propose the the VLA syntax instead.
+It was rejected in favor of more general unsized values, which culminated in [RFC 1909].
+
+The [`alloca` crate](https://crates.io/crates/alloca) implements dynamic stack allocation via a closure indirection and FFI with C.
+
+The `unsized_fn_params` feature doesn't suffer from the same problems as `unsized_locals` and will still be kept around.
+It is independent of this RFC.
+
+# Posterior art
+[posterior-art]: #posterior-
+
+The best prior art for this removal is of course the inspiration for the de-RFC format, the [type-ascription de-RFC].
+[MCP 630] can also be seen as prior art
+
+# Unresolved answers
+[unresolved-answers]: #unresolved-answers
+
+None
+
+# Future probabilities
+[future-probabilities]: #future-probabilities
+
+In the future, dynamic stack allocation may be re-added to Rust via some other feature that solves the explicitness problems outlined in the motivation.
+
+Alternatively, it could be decided, explicitly or implicitly through inaction, that dynamic stack allocation is not a fit for Rust and will not be added.
+
+[type-ascription de-RFC]: https://rust-lang.github.io/rfcs/3307-de-rfc-type-ascription.html
+[`alloca`]: https://man7.org/linux/man-pages/man3/alloca.3.html
+[MCP 630]: https://github.com/rust-lang/compiler-team/issues/630
+[RFC 1808]: https://github.com/rust-lang/rfcs/pull/1808
+[RFC 1909]: https://rust-lang.github.io/rfcs/1909-unsized-rvalues.html
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -94,6 +94,10 @@ The `unsized_locals` feature is removed from the compiler and [RFC 1909] is offi
 
 If someone wants to bring dynamic stack allocation into Rust again, a new design will have to be designed from scratch, considering all the problems laid out in this RFC.
 
+This does not have a negative effect on features that feature unsized values in function signatures like `unsized_fn_params`.
+Their behavior is much more clear and they are implemented differently.
+In fact, `unsized_fn_params` is currently needed in the standard library to implement `Box<dyn FnOnce()> as FnOnce()>`.
+
 # Drawforwards
 [drawforwards]: #drawforwards
 
@@ -129,6 +133,7 @@ The best prior art for this removal is of course the inspiration for the de-RFC 
 
 # Unresolved answers
 [unresolved-answers]: #unresolved-answers
+
 
 None
 
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -59,7 +59,7 @@ There are ways around this (rewinding the stack pointer at the end of the loop t
 ## Implicit Danger
 
 While lack of implementation quality is a sign of lack of interest for the feature, it is not the primary reason for this RFC,
-which is purely about the langauge design of the feature.
+which is purely about the language design of the feature.
 
 The original RFC was very short, and especially short on motivation and rationale for the design.
 
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -87,6 +87,13 @@ By allowing people to create a dynamic stack allocation without being aware of w
 As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_.
 Therefore, `unsized_locals` must go.
 
+`unsized_locals` is not the only feature that can cause unbounded stack allocation and eventually lead to stack overflow in Rust, one has to look no further than Dijkstra's favorite: recursion.
+A program with unbounded recursion is easily found and fixed, but especially recursive parsers suffer from similar problems where user input can cause stack overflows.
+What makes recursion different from dynamic stack allocation?
+The main difference here is that recursion is a lot harder to do on accident.
+When recursion is used, it is usually used on purpose, and while sometimes the potential for stack overflows is overlooked, the general feature is usually used on purpose.
+That said, recursion can certainly be dangerous in some contexts, but prior existing features are not a good reason to introduce more ways to blow the stack.
+
 # Guide-level obfuscation
 [guide-level-obfuscation]: #guide-level-obfuscation
 
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -12,7 +12,7 @@ _Following the great success of the [type ascription de-RFC], we now present the
 Unsized locals ([RFC 1909], called "unsized rvalues" originally)
 has been a merged RFC for eight years with no clear path to stabilization.
 
-There is still a very large gap in the implementation in rustc that hasn't been adressed for many years and there are several language problems with this feature. 
+There is still a very large gap in the implementation in rustc that hasn't been addressed for many years and there are several language problems with this feature. 
 
 This RFC intends to advocate for the feature being removed entirely with a fresh RFC being necessary to add a similiar feature again.
 
@@ -29,7 +29,7 @@ fn main() {
 ```
 
 This will dynamically allocate space on the stack for the string.
-C has a similiar feature, [`alloca`] and variable length arrays (VLA), the latter having been made optional to implement in C11.
+C has a similar feature, [`alloca`] and variable length arrays (VLA), the latter having been made optional to implement in C11.
 
 ## Lack of Proper Implementation
 
@@ -48,9 +48,9 @@ which is purely about the langauge design of the feature.
 
 The original RFC was very short, and especially short on motivation and rationale for the design.
 
-Dynamic stack allocation in geneneral has a rather significant downside: it makes it easy to accidentally overflow the stack if you allocate a lot.
+Dynamic stack allocation in general has a rather significant downside: it makes it easy to accidentally overflow the stack if you allocate a lot.
 Stacks are usually rather small (on the order of few megabytes, depending on the platform), which means that dynamically allocating user-controlled input on the stack is often rather dangerous.
-While stack overflows are not considered memory unsafe by Rust, they still cause crashes which can lead to denial of service vulnerabilities and unrealiableness in general.
+While stack overflows are not considered memory unsafe by Rust, they still cause crashes which can lead to denial of service vulnerabilities and unreliability in general.
 Dynamic stack allocation as currently implemented interacts especially poorly with loops.
 Allocations are not freed until the function returns, so the following example overflows the stack:
 
@@ -79,9 +79,9 @@ As such, it behaves similarly to `unsafe` (but is not actually `unsafe`).
 With `unsized_locals`, the use of dynamic stack allocation is completely implicit.
 When you create an unsized local, it is often not obvious that dynamic stack allocation is happening.
 In the example from the start, we need to be aware of all the involved types (which are often inferred in practice) to know that this is a potentially problematic unsized local that we have to audit more carefully instead of a normal sized local.
-Especially around strings, which are often user-controlled, this easily lead to accidental dangerous situations.
+Especially around strings, which are often user-controlled, this easily leads to accidentally-dangerous situations.
 
-Rusts strings, reference types, and `Sized` are a part of the language that can often be hard to understand for beginners coming from garbage collected languages.
+Rust's strings, reference types, and `Sized` are a part of the language that can often be hard to understand for beginners coming from garbage-collected languages.
 By allowing people to create a dynamic stack allocation without being aware of what is happening, we open the doors for people, especially new Rust programmers who are not intimately familiar with the tradeoffs of dynamic stack allocation, to shoot themselves in the foot and become vulnerable to unexpected crashes.
 
 As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_.
@@ -124,8 +124,8 @@ It is independent of this RFC.
 # Posterior art
 [posterior-art]: #posterior-
 
-The best prior art for this removal is of course the inspiration for the de-RFC format, the [type-ascription de-RFC].
-[MCP 630] can also be seen as prior art
+The best prior art for this removal is of course the inspiration for the de-RFC format, the [type ascription de-RFC].
+[MCP 630] can also be seen as prior art.
 
 # Unresolved answers
 [unresolved-answers]: #unresolved-answers
@@ -139,7 +139,7 @@ In the future, dynamic stack allocation may be re-added to Rust via some other f
 
 Alternatively, it could be decided, explicitly or implicitly through inaction, that dynamic stack allocation is not a fit for Rust and will not be added.
 
-[type-ascription de-RFC]: https://rust-lang.github.io/rfcs/3307-de-rfc-type-ascription.html
+[type ascription de-RFC]: https://rust-lang.github.io/rfcs/3307-de-rfc-type-ascription.html
 [`alloca`]: https://man7.org/linux/man-pages/man3/alloca.3.html
 [MCP 630]: https://github.com/rust-lang/compiler-team/issues/630
 [RFC 1808]: https://github.com/rust-lang/rfcs/pull/1808
--- a/text/3829-de-rfc-unsized-locals.md
+++ b/text/3829-de-rfc-unsized-locals.md
@@ -87,8 +87,7 @@ Especially around strings, which are often user-controlled, this easily leads to
 Rust's strings, reference types, and `Sized` are a part of the language that can often be hard to understand for beginners coming from garbage-collected languages.
 By allowing people to create a dynamic stack allocation without being aware of what is happening, we open the doors for people, especially new Rust programmers who are not intimately familiar with the tradeoffs of dynamic stack allocation, to shoot themselves in the foot and become vulnerable to unexpected crashes.
 
-As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_.
-Therefore, `unsized_locals` must go.
+As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_, a major downside of this feature that has not been addressed in prior RFCs.
 
 `unsized_locals` is not the only feature that can cause unbounded stack allocation and eventually lead to stack overflow in Rust, one has to look no further than Dijkstra's favorite: recursion.
 A program with unbounded recursion is easily found and fixed, but especially recursive parsers suffer from similar problems where user input can cause stack overflows.
@@ -105,6 +104,8 @@ The Linux kernel has spent a lot of time on getting VLAs (C's cousin to `unsized
 The `unsized_locals` feature is removed from the compiler and [RFC 1909] is officially unaccepted.
 
 If someone wants to bring dynamic stack allocation into Rust again, a new design will have to be designed from scratch, considering all the problems laid out in this RFC.
+Alternatively, the existing design of `unsized_locals` *could* be RFCed again, if such an RFC was able to work around all the problems.
+This RFC merely unaccepts the previous `unsized_locals` RFC, it does not necessarily banish `unsized_locals` from existence forever.
 
 This does not have a negative effect on features that feature unsized values in function signatures like `unsized_fn_params`.
 Their behavior is much more clear and they are implemented differently.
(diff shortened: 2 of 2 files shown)


text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+_Following the great success of the [type ascription de-RFC], we now present the next one in the series!_
+
+# Summary
+[summary]: #summary
+
+Unsized locals ([RFC 1909], called "unsized rvalues" originally)
+has been a merged RFC for eight years with no clear path to stabilization.
+
+There is still a very large gap in the implementation in rustc that hasn't been adressed for many years and there are several language problems with this feature.
```
workingjubilee:
```suggestion
There is still a very large gap in the implementation in rustc that hasn't been addressed for many years and there are several language problems with this feature. 
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+```rust
+#![feature(unsized_locals)]
+fn main() {
+    let x = *Box::from("hello, world!");
+}
+```
+
+This will dynamically allocate space on the stack for the string.
+C has a similiar feature, [`alloca`] and variable length arrays (VLA), the latter having been made optional to implement in C11.
```
workingjubilee:
```suggestion
C has a similar feature, [`alloca`] and variable length arrays (VLA), the latter having been made optional to implement in C11.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+This removal hasn't yet been implemented.
+
+## Implicit Danger
+
+While lack of implementation quality is a sign of lack of interest for the feature, it is not the primary reason for this RFC,
+which is purely about the langauge design of the feature.
+
+The original RFC was very short, and especially short on motivation and rationale for the design.
+
+Dynamic stack allocation in geneneral has a rather significant downside: it makes it easy to accidentally overflow the stack if you allocate a lot.
```
workingjubilee:
```suggestion
Dynamic stack allocation in general has a rather significant downside: it makes it easy to accidentally overflow the stack if you allocate a lot.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+None
+
+# Future probabilities
+[future-probabilities]: #future-probabilities
+
+In the future, dynamic stack allocation may be re-added to Rust via some other feature that solves the explicitness problems outlined in the motivation.
+
+Alternatively, it could be decided, explicitly or implicitly through inaction, that dynamic stack allocation is not a fit for Rust and will not be added.
+
+[type-ascription de-RFC]: https://rust-lang.github.io/rfcs/3307-de-rfc-type-ascription.html
```
kennytm:
```suggestion
[type ascription de-RFC]: https://rust-lang.github.io/rfcs/3307-de-rfc-type-ascription.html
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+## Implicit Danger
+
+While lack of implementation quality is a sign of lack of interest for the feature, it is not the primary reason for this RFC,
+which is purely about the langauge design of the feature.
+
+The original RFC was very short, and especially short on motivation and rationale for the design.
+
+Dynamic stack allocation in geneneral has a rather significant downside: it makes it easy to accidentally overflow the stack if you allocate a lot.
+Stacks are usually rather small (on the order of few megabytes, depending on the platform), which means that dynamically allocating user-controlled input on the stack is often rather dangerous.
+While stack overflows are not considered memory unsafe by Rust, they still cause crashes which can lead to denial of service vulnerabilities and unrealiableness in general.
```
workingjubilee:
```suggestion
While stack overflows are not considered memory unsafe by Rust, they still cause crashes which can lead to denial of service vulnerabilities and unreliability in general.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+Which is where we get to the major argument of this RFC: The `unsized_locals` feature integrates far too naturally into Rust.
+This makes the feature very **easy to use**, and especially **easy to use accidentally**.
+
+As previously mentioned, dynamic stack allocation can be dangerous and should not be used lightly.
+It's an advanced optimization feature that is best left untouched by default.
+As such, it behaves similarly to `unsafe` (but is not actually `unsafe`).
+With `unsized_locals`, the use of dynamic stack allocation is completely implicit.
+When you create an unsized local, it is often not obvious that dynamic stack allocation is happening.
+In the example from the start, we need to be aware of all the involved types (which are often inferred in practice) to know that this is a potentially problematic unsized local that we have to audit more carefully instead of a normal sized local.
+Especially around strings, which are often user-controlled, this easily lead to accidental dangerous situations.
```
workingjubilee:
```suggestion
Especially around strings, which are often user-controlled, this easily leads to accidentally-dangerous situations.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+Rusts strings, reference types, and `Sized` are a part of the language that can often be hard to understand for beginners coming from garbage collected languages.
```
workingjubilee:
```suggestion
Rust's strings, reference types, and `Sized` are a part of the language that can often be hard to understand for beginners coming from garbage-collected languages.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+The [`alloca` crate](https://crates.io/crates/alloca) implements dynamic stack allocation via a closure indirection and FFI with C.
+
+The `unsized_fn_params` feature doesn't suffer from the same problems as `unsized_locals` and will still be kept around.
+It is independent of this RFC.
+
+# Posterior art
+[posterior-art]: #posterior-
+
+The best prior art for this removal is of course the inspiration for the de-RFC format, the [type-ascription de-RFC].
+[MCP 630] can also be seen as prior art
```
workingjubilee:
```suggestion
[MCP 630] can also be seen as prior art.
```

workingjubilee (commented):
Typos.

kennytm:
I hope that this won't affect feasibility of #2884 (basically unsized_return).

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+The [`alloca` crate](https://crates.io/crates/alloca) implements dynamic stack allocation via a closure indirection and FFI with C.
+
+The `unsized_fn_params` feature doesn't suffer from the same problems as `unsized_locals` and will still be kept around.
+It is independent of this RFC.
+
+# Posterior art
+[posterior-art]: #posterior-
+
+The best prior art for this removal is of course the inspiration for the de-RFC format, the [type-ascription de-RFC].
```
Noratrieb:
```suggestion
The best prior art for this removal is of course the inspiration for the de-RFC format, the [type ascription de-RFC].
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+It is independent of this RFC.
+
+# Posterior art
+[posterior-art]: #posterior-
+
+The best prior art for this removal is of course the inspiration for the de-RFC format, the [type ascription de-RFC].
+[MCP 630] can also be seen as prior art.
+
+# Unresolved answers
+[unresolved-answers]: #unresolved-answers
```
workingjubilee, line 148: Given @kennytm's concern, it may be useful to clarify this should not really affect unsized values in signatures per se, though it may cause some integration problems that would likely have to be solved anyways (as anyone who was hoping to have the `unsized_locals` design carry any unsized handling for parameters or returns was obviously betting on the wrong horse).

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+Therefore, this RFC is not necessarily a rejection of the idea of dynamic stack allocation, but merely the way the `unsized_locals` feature exposes it.
+
+Which is where we get to the major argument of this RFC: The `unsized_locals` feature integrates far too naturally into Rust.
+This makes the feature very **easy to use**, and especially **easy to use accidentally**.
+
+As previously mentioned, dynamic stack allocation can be dangerous and should not be used lightly.
+It's an advanced optimization feature that is best left untouched by default.
+As such, it behaves similarly to `unsafe` (but is not actually `unsafe`).
+With `unsized_locals`, the use of dynamic stack allocation is completely implicit.
+When you create an unsized local, it is often not obvious that dynamic stack allocation is happening.
```
dlight, line 83: Accidentally creating unsized locals is bad, but recursion is *also* dangerous in the same way, and also done implicitly (recursive calls, and in special mutually recursive calls, unfortunately look like any other call). You also don't need to annotate very large types in local variables (such as large arrays) that will be replicated alongside recursive calls. There's tail recursion in nightly but it inverts the syntactical burden (recursion is implicit, tail recursion is explicitly marked as such)

Maybe if you wanted to *prove* the stack is never exhausted (a feature Rust doesn't have, but maybe it should - something like, with the right annotations rustc will fail to compile if the stack can possibly grow beyond 8MB or something), it would be very hard to have this hypothetical feature with unsized locals without upstreaming something like [flux](https://github.com/flux-rs/flux). But that's true regardless of how much syntactical noise unsized locals have.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+Dynamic stack allocation also has its upsides.
+It is generally faster than heap allocation and can therefore improve performance in cases where the previously mentioned downsides are not a concern.
+Therefore, this RFC is not necessarily a rejection of the idea of dynamic stack allocation, but merely the way the `unsized_locals` feature exposes it.
+
+Which is where we get to the major argument of this RFC: The `unsized_locals` feature integrates far too naturally into Rust.
+This makes the feature very **easy to use**, and especially **easy to use accidentally**.
+
+As previously mentioned, dynamic stack allocation can be dangerous and should not be used lightly.
+It's an advanced optimization feature that is best left untouched by default.
+As such, it behaves similarly to `unsafe` (but is not actually `unsafe`).
```
Jules-Bertholet, line 81: I think this exaggerates the danger. For example, maybe you are writing code for an embedded system with a low amount of memory. You want to use trait objects to save on code size, but also don’t want to pull in an allocator. In this case, dynamic alloca could be very helpful for *reducing* memory usage.

And, as others have mentioned, there are many other safe and easy ways of overflowing the stack in Rust. We don’t generally call e.g. recursion an “advanced optimization feature” that’s so scary it’s “similar to `unsafe`”.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+#![feature(unsized_locals)]
+fn main() {
+    let s = "A".repeat(1000);
+    for i in 0..1000000 {
+        let x: str = *Box::from(s.as_str());
+        std::hint::black_box(&s);
+    }
+}
+```
+There are ways around this (rewinding the stack pointer at the end of the loop to free up the memory), but they are not currently implemented.
```
Jules-Bertholet, line 59: This belongs in the “lack of proper implementation” section. It’s a bug in the current implementation, not a problem with the feature as such.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+Dynamic stack allocation also has its upsides.
+It is generally faster than heap allocation and can therefore improve performance in cases where the previously mentioned downsides are not a concern.
+Therefore, this RFC is not necessarily a rejection of the idea of dynamic stack allocation, but merely the way the `unsized_locals` feature exposes it.
```
Jules-Bertholet, line 74: I agree that we still need to figure out the best way to expose the feature, but I think that the best way to do that is to implement it properly on nightly and enable experimentation. Right now, we have vague concerns about footguns and blowing the stack, but without concrete user feedback, it’s impossible to determine how severe the issue is or how the problem could be mitigated.

Jules-Bertholet: E.g. maybe lints could address the danger. Or maybe all unsized locals should be annotated in some way (e.g. `let #[unsized] foo = …`). We can’t try those things out unless we have a proper implementation.

text/3829-de-rfc-unsized-locals.md
workingjubilee: In practice, I rarely see recursive code in Rust, but I do see `let` bindings a lot, so this doesn't seem like a very strong argument.

text/3829-de-rfc-unsized-locals.md
workingjubilee: Then go implement it if you think it's such a good idea.

text/3829-de-rfc-unsized-locals.md
workingjubilee: ~~It is part of the feature's UX as currently designed. Discussing hypothetical features with wildly different UX and then pretending they are still the same because they accomplish a vaguely similar design goal is a vice of this community.~~ Meh, incorrect.

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: My point is, it’s hard to experiment with the frontend/syntax parts of the feature, if the codegen/backend is missing/broken.

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: [RFC 1909](https://rust-lang.github.io/rfcs/1909-unsized-rvalues.html) does not specify “unsized locals in a loop leak a bunch of memory for no reason”. It’s not part of the “design”; it’s a codegen bug. We know how to fix it, and that fix would need to happen before stabilization.

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: You do see function calls a lot, though. And function calls can be accidentally recursive, just as `let`s can be accidentally unsized

kennytm: On one hand the stack usage by recursion is O(n) --- you need to call itself too many times before blowing up the stack,
and alloca is O(1) --- you just need to one wrong `let` to crash the entire program, so `unsized_locals` is indeed much more dangerous than recursion.

On the other hand, the same applies for *sized* locals
```rust
#[expect(clippy::large_stack_arrays)] // <- this is pedantic level lint, not enabled by default
let a = [2; 1_000_000_000_000];
assert_eq!(a.last(), Some(&2));
```
(that 1_000_000_000_000 can be hidden from the reader through const calculations like `size_of::<T>() * 100` or some multi-dimensional arrays)

For sure this is not *dynamic* stack allocation, with alloca the compiler cannot verify that the length is bounded by some number and thus cannot deterministically issue that equivalent of `clippy::large_stack_arrays`, but if `[2; 1_000_000_000_000]` is allowed to be safe there seems no reason to say alloca behaves like unsafe. :thinking:

workingjubilee: this is true, but `[2; 1_000_000_000_000]` seems slightly more obvious in being a gunshot at your foot.

@Jules-Bertholet I'm not going to contest that you can have accidental recursion, but we're still talking about several orders of magnitude less in frequency than "`let` binding in a loop". Indeed, the fact that loops (or iterator combinators) are so often the more ergonomic choice in Rust is part of why I see recursion less.

Jules-Bertholet: > "`let` binding in a loop"

Again, the fact that this blows the stack is a (show-stopper) codegen bug in the current implementation. Unsized locals should not be stabilized without that bug being fixed, but it does not invalidate the feature as such.

The actual stack overflow footgun with the unsized locals feature looks more like:

```rust
fn foo(this_slice_might_be_huge: &[u32]) {
    let local = *this_slice_might_be_huge; // oops
}
```

burdges:
It'd be cool if unsized_return could return into some future unsized_local, which sounds impossible with the alloca crate.  An unsized_local that's less natural feeling sounds reasonable though.  Anyways, the alloca crate has only three dependents, so maybe nobody cares enough.

text/3829-de-rfc-unsized-locals.md
workingjubilee: ...I think, to actually implement the suggestion, we would have to allow every `let`-binding of an alloca that is nested inside a loop to turn that loop body into a closure which is now driven by the new, restructured loop. The closure would have to capture all the values referenced inside the loop and then return a value representing the control flow state to decide whether we must e.g. `return`. That could force a stack frame into the generated code, but then we would have to hope that the dynamic allocation then doesn't e.g. get inlined? There's some possible alternatives that could use something involving exception handling as an implementation if it performed better, but would otherwise be similar in concept.

While I am obviously hypothesizing based on insufficient information, I expect any such implementation to be a significant-enough optimization barrier that such a feat would detract from the intended usage. And if it isn't, it seems it would be at risk of optimizations piercing through the function boundary and removing the stack frame. That would actually reintroduce the stack overflow risk.

Jules-Bertholet: What? No, you just need to adjust the stack pointer after each iteration of the loop. IIUC, rustc should be emitting [`llvm.stacksave`](https://llvm.org/docs/LangRef.html#llvm-stacksave-intrinsic) and [`llvm.stackrestore`](https://llvm.org/docs/LangRef.html#int-stackrestore) in the LLVM IR. That it does not do this, is a bug.

workingjubilee: huh, I missed those instructions while specifically looking for instructions like that. hurf.

text/3829-de-rfc-unsized-locals.md
Noratrieb: No significant experimentation has happened for the past **eight** years. I wouldn't hold my beer.

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: We’ve all been there…

text/3829-de-rfc-unsized-locals.md
workingjubilee: Right, it's actually the middle-end that we're missing, I think? Because MIR is kinda its own beast?

Jules-Bertholet: > No significant experimentation has happened for the past **eight** years.

Yes, because the middle/back-end implementation is broken, scaring nightly users away. If the compiler team would rather strip out that broken implementation than continue to let it rot, I have no objection. But https://github.com/rust-lang/compiler-team/issues/630 is enough for that. IIUC this de-RFC would require anyone motivated to implement the feature properly, to go through the full RFC process first. I think that’s going too far. My preference would be to “downgrade” #1909 to an eRFC, instead of unaccepting it completely.

workingjubilee: Changing it to an "experimental" RFC will not change that right now it is a gravestone of a feature that no one wants to proceed on. Yes, probably someone *should* go through the RFC process again, because right now there's not even a draft of a proposal for how it may be implemented in rustc's semantic layers.

Jules-Bertholet: >  it is a gravestone of a feature that no one wants to proceed on

I wouldn’t say that. For example, someone recently put in the effort to [fix](https://github.com/rust-lang/rust/pull/111374) the longstanding issue with unsized locals being misaligned. Clearly there are *some* people still interested in this.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+This will dynamically allocate space on the stack for the string.
+C has a similar feature, [`alloca`] and variable length arrays (VLA), the latter having been made optional to implement in C11.
+
+## Lack of Proper Implementation
+
+This feature has never been properly implemented in rustc.
+The variable length array form proposed in the RFC still doesn't exist at all, only the unsized local variable does.
+It is implemented in the type checker and codegen, but lacking MIR semantics and therefore unimplemented in the compile time function evaluator.
+This is very significant, as MIR semantics govern how the feature should behave precisely in the first place.
+Without them, they cannot work in `const` and optimizations are likely broken around them.
```
Jules-Bertholet: I would like to understand the “MIR semantics” issues in more detail. What exactly would the compiler team need to see before it would become possible to move forward? https://github.com/rust-lang/compiler-team/issues/630 says:

> The [MIR semantics] that const eval used to implement - before support was removed - were known to be awkward and surprising.

But doesn’t link to any of that discussion. It’s hard for people to step up and address the problems, without a clear summary of what they are.

workingjubilee: Even then, a detailed summary of the problems would also have to be updated for changes in MIR semantics since then, no? Which may be potentially quite a bit.

text/3829-de-rfc-unsized-locals.md
workingjubilee: I believe that what @tmiasko was touching on there involved fixes to parts of the code that were shared between `unsized_locals` and `unsized_fn_params`, because code that accepts `unsized_fn_params` sometimes, due to the way codegen works, creates unsized temporaries, i.e. locals. That then requires handling that for `unsized_fn_params`. Our current rude hack that addresses this involves blocking inlining for `unsized_fn_params`, since that's one of the conditions to reach that.

Also see:
- https://github.com/rust-lang/rust/issues/111355
- https://github.com/rust-lang/rust/issues/111175

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: I found https://github.com/rust-lang/rust/issues/48055#issuecomment-1837424794 which seems to be what I was looking for, or at least a start.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+- RFC PR: [rust-lang/rfcs#3829](https://github.com/rust-lang/rfcs/pull/3829)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+
+_Following the great success of the [type ascription de-RFC], we now present the next one in the series!_
+
+# Summary
+[summary]: #summary
+
+Unsized locals ([RFC 1909], called "unsized rvalues" originally)
```
scottmcm, line 12: suggestion: also include a change to that RFC to add a header mentioning this one, like how https://github.com/rust-lang/rfcs/pull/3219/files added a header about unaccepting part of 2115.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+This feature has never been properly implemented in rustc.
+The variable length array form proposed in the RFC still doesn't exist at all, only the unsized local variable does.
+It is implemented in the type checker and codegen, but lacking MIR semantics and therefore unimplemented in the compile time function evaluator.
+This is very significant, as MIR semantics govern how the feature should behave precisely in the first place.
+Without them, they cannot work in `const` and optimizations are likely broken around them.
+Because of this lack of implementation quality, the `unsized_locals` feature was already accepted for removal from rustc two years ago in [MCP 630].
+This removal hasn't yet been implemented.
+
+## Implicit Danger
```
workingjubilee: It may be worth noting that projects like the Linux kernel entirely ban the C cousin to this proposed Rust feature, the VLA: https://github.com/torvalds/linux/commit/5e88c48cb4bcbd0ea683373a106268933bb4edd2

text/3829-de-rfc-unsized-locals.md
kennytm: @Jules-Bertholet `let local = *this_slice_might_be_huge` won't compile because `[u32]` is not Copy.

This works though:

```rust
#![feature(unsized_locals)]

fn g(a: Box<[u32]>) -> u32 {
    let b = *a;
    // thread '<unknown>' has overflowed its stack
    // fatal runtime error: stack overflow, aborting
    *b.last().unwrap()
}

fn main() {
    let a = vec![2; 10_000_000].into_boxed_slice();
    assert_eq!(g(a), 2);
}
```

Compare with `unsized_fn_params` which runs fine

```rust
#![feature(unsized_fn_params)]

fn g(a: [u32]) -> u32 {
    *a.last().unwrap()
}

fn main() {
    let a = vec![2; 10_000_000].into_boxed_slice();
    assert_eq!(g(*a), 2);
}
```

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: Yes, the reasons for this are explained here: https://www.phoronix.com/news/Linux-Kills-The-VLA

The biggest concern seems to have been the risk of blowing the stack; Linus believed it was essentially always better, at least for the kernel, to just allocate a fixed-sized array buffer of sufficient size.

I think that’s fair. I more or less agree with criticisms of VLAs in C more broadly, and that they are rarely the right choice. I also agree that if unsized locals were to be stabilized in Rust, we would want some mechanism to warn people away from using them with slices.

*However*, Rust, compared to C, has a second kind of unsized type: trait objects. And I suspect that unsized locals would be both far more useful, and far less dangerous, for them.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,146 @@
...
+
+While lack of implementation quality is a sign of lack of interest for the feature, it is not the primary reason for this RFC,
+which is purely about the langauge design of the feature.
```
Fulgen301:
```suggestion
which is purely about the language design of the feature.
```

text/3829-de-rfc-unsized-locals.md
scottmcm: > _However_, Rust, compared to C, has a second kind of unsized type: trait objects. And I suspect that unsized locals would be both far more useful, and far less dangerous, for them.

With generators and futures, I think that's become less true, because we already know that people sometimes have problems with a whole bunch of state in their futures resulting in big enough futures to cause stack size issues just putting them into boxes.

scottmcm:
I think de-RFCing the local variables part of this makes sense.

The original FCP was

> I am not recommending an experimental RFC because I don't think we'll be making vast changes to the technical plan laid out here

But it seems like what we'd ended up hitting is that we really do need a fairly drastic re-work of how this works internally (since we don't have a MIR semantic for it) and plausibly also a re-work of how the syntax should behave (because we don't necessarily want this to *silently* happen).

Notably, like how the type ascription de-RFC wasn't a "we don't want type ascription at all" statement, I don't see this de-RFC as a statement that "we don't want allocas ever".  If a feature with similar motivation comes back that's fine; this is just recognizing that the current state here doesn't seem to be stabilization-track and probably getting there deserves a new RFC.

My understanding is that this de-RFC is intentionally not touching `unsized_parameters`, which sounds appropriate to me.  [EDIT: the RFC has been updated to say this very directly as well.]

@rfcbot fcp merge

text/3829-de-rfc-unsized-locals.md
Noratrieb: I added a mention of recursion in abb5788400405fd912a5468740c8955a52ad6f6a.

text/3829-de-rfc-unsized-locals.md
Noratrieb: This makes sense, I moved it in a104109606c3adbe64b30a5892adb7e713f768b5

text/3829-de-rfc-unsized-locals.md
Noratrieb: I added a link to the comment in 5bc11846302f5fe76326f065c33b554339979e91

text/3829-de-rfc-unsized-locals.md
Noratrieb: 0e0b1a55f6825d4cad326c4a743fcbe1cf741e51

text/3829-de-rfc-unsized-locals.md
Noratrieb: Added in 969b1392001a4b398b77ae4b2c192e54d6ee01f3

text/3829-de-rfc-unsized-locals.md
```diff
@@ -94,6 +94,10 @@ The `unsized_locals` feature is removed from the compiler and [RFC 1909] is offi
 
 If someone wants to bring dynamic stack allocation into Rust again, a new design will have to be designed from scratch, considering all the problems laid out in this RFC.
 
+This does not have a negative effect on features that feature unsized values in function signatures like `unsized_fn_params`.
+Their behavior is much more clear and they are implemented differently.
+In fact, `unsized_fn_params` is currently needed in the standard library to implement `Box<dyn FnOnce()> as FnOnce()>`.
```
scottmcm, line 112: Thanks; good to have this listed out very explicitly.

text/1909-unsized-rvalues.md
```diff
@@ -3,6 +3,11 @@
 - RFC PR: [rust-lang/rfcs#1909](https://github.com/rust-lang/rfcs/pull/1909)
 - Rust Issue: [rust-lang/rust#48055](https://github.com/rust-lang/rust/issues/48055)
 
+> ⚠ Update 8 years later ⚠
+>
+> [The team decided to un-accept](https://github.com/rust-lang/rfcs/pull/3829) this RFC and the feature has been removed.
```
scottmcm: nit:
```suggestion
> [The team decided to un-accept](https://github.com/rust-lang/rfcs/pull/3829) the portions of this RFC not related to function parameters and the `unsized_locals` feature has been removed.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,161 @@
...
+
+# Summary
+[summary]: #summary
+
+Unsized locals ([RFC 1909], called "unsized rvalues" originally)
+has been a merged RFC for eight years with no clear path to stabilization.
+
+There is still a very large gap in the implementation in rustc that hasn't been addressed for many years and there are several language problems with this feature. 
+
+This RFC intends to advocate for the feature being removed entirely with a fresh RFC being necessary to add a similiar feature again.
```
joshtriplett:
```suggestion
This RFC intends to advocate for the feature being removed entirely with a fresh RFC being necessary to add a similar feature again.

Note that the acceptance of this RFC should not be taken as judgement on a future RFC. A fresh RFC with a new design would be required, and this RFC may provide input on design constraints for such an RFC, but this should not be taken as pre-rejecting such an RFC.
```

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: Consider this bog-standard generic function:

```rust
fn foo(arg: impl SomeTrait) {
    // do stuff with `arg`, including storing it in locals
}
```

To my mind, if the above is considered totally mundane and not-scary, then this:

```rust
fn foo(arg: dyn SomeTrait) {
    // do stuff with `arg`, including storing it in locals
}
```

Should not be scary either. The risk of `dyn SomeTrait` being huge and blowing the stack, is no different from the risk of `impl SomeTrait` being huge and blowing the stack. Unfairly penalizing the latter pushes users towards monomorphization, making it *harder* to limit memory usage.

workingjubilee: The discussion of `dyn Trait` is a separate matter. I'm resolving this suggestion.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,163 @@
...
+Therefore, `unsized_locals` must go.
+
+`unsized_locals` is not the only feature that can cause unbounded stack allocation and eventually lead to stack overflow in Rust, one has to look no further than Dijkstra's favorite: recursion.
+A program with unbounded recursion is easily found and fixed, but especially recursive parsers suffer from similar problems where user input can cause stack overflows.
+What makes recursion different from dynamic stack allocation?
+The main difference here is that recursion is a lot harder to do on accident.
+When recursion is used, it is usually used on purpose, and while sometimes the potential for stack overflows is overlooked, the general feature is usually used on purpose.
+That said, recursion can certainly be dangerous in some contexts, but prior existing features are not a good reason to introduce more ways to blow the stack.
+
+The Linux kernel has spent a lot of time on getting VLAs (C's cousin to `unsized_locals`) [removed from the codebase](https://www.phoronix.com/news/Linux-Kills-The-VLA).
```
Jules-Bertholet, line 99: Reiterating the point I made [here](https://github.com/rust-lang/rfcs/pull/3829#discussion_r2126412382) in a now-resolved thread:

Rust has two main kinds of unsized types: slices and trait objects. Putting unsized slices on the stack is analogous to VLAs in C, and is footgunny and useless to about the same degree. However, putting trait objects on the stack—something which has no analogue in C—is much less likely to cause an overflow, and also offers far more utility.

text/3829-de-rfc-unsized-locals.md
Jules-Bertholet: I started a discussion about fixing the MIR on Zulip: https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/MIR.20semantics.20for.20unsized.20locals

traviscross:
While I'm OK with de-RFCing this, as probably I do want someone to write a fresh RFC, if we were to actually do this[^1], that made the case for how this should be done in the context of Rust as it is today, I don't want to commit to the following as a normative position:

> Which is where we get to the major argument of this RFC: The `unsized_locals` feature integrates far too naturally into Rust. This makes the feature very easy to use, and especially easy to use accidentally.

Maybe that's correct, maybe it's not (as @Jules-Bertholet is arguing above).  Probably I'd want to see an RFC arguing the case for what the right approach is (and read through the threads of all the prior RFCs), and I don't want to prejudge that.

In accepting an RFC, we generally aren't necessarily accepting the motivation or other non-normative sections as the position of the team, so perhaps I could check a box anyway.  But there's a lot of normative-sounding language in this RFC, e.g.,

> As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_. Therefore, `unsized_locals` must go.

and it's not clear to me what the intended normative sections are.  If this can be adjusted a bit so that it's clear that this RFC isn't committing normatively to this, but is rather raising this as one item that a future RFC should clearly speak to, and that the position taken here is the author's only, then I'll happily check a box.

I'd be happy both to do that and for @Jules-Bertholet or others to pursue an experiment to see if dynamic stack allocation can be done correctly in Rust, and to explore the best way to expose that in the language.

[^1]: ...after an acceptable experimental implementation were to be done.

traviscross:
@rfcbot concern clarify-what-is-normative

As [mentioned](https://github.com/rust-lang/rfcs/pull/3829#issuecomment-2941822651), I'd like to avoid taking a normative position on how this might be done in the future, but in any event, we should clarify what exactly is normative here and therefore what our proposed FCP represents, so I'll file a concern to that effect.

It'd be OK, I think, if the only normative aspect of this RFC is that, by virtue of this de-RFC, someone needs to write a new RFC in order to eventually stabilize this.

Jules-Bertholet:
> > Which is where we get to the major argument of this RFC: The `unsized_locals` feature integrates far too naturally into Rust. This makes the feature very easy to use, and especially easy to use accidentally.
> 
> Maybe that's correct, maybe it's not (as @Jules-Bertholet is arguing above).

To be clear, though I think the RFC’s language on this is too strong, I do agree that there is a real and serious concern here, and it would have to be addressed before stabilization.

> It'd be OK, I think, if the only normative aspect of this RFC is that, by virtue of this de-RFC, someone needs to write a new RFC in order to eventually stabilize this.

I fully agree with this, this would address all my objections.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,163 @@
...
+This puts the feature into a really bad position, but it may be readded in the future if someone desires.
+With this RFC, the fate of `unsized_locals` is sealed and it becomes clear to anyone what the state of the feature is.
+
+The intent of this RFC is not to proposed an alternative way to solve dynamic stack allocation,
+but there are some listed alternatives here that may be considered in the future if desired.
+While this RFC doesn't explicitly encourage people to revisit this topic, it may result in new activity around dynamic stack allocation in Rust.
+
+[RFC 1808] proposed an `alloca` function, which was rejected because `alloca` does not really behave like a function.
+
+[RFC 1808] then changed to propose the the VLA syntax instead.
```
bjorn3, line 133:
```suggestion
[RFC 1808] then changed to propose the VLA syntax instead.
```

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,163 @@
...
+The [`alloca` crate](https://crates.io/crates/alloca) implements dynamic stack allocation via a closure indirection and FFI with C.
+
+The `unsized_fn_params` feature doesn't suffer from the same problems as `unsized_locals` and will still be kept around.
+It is independent of this RFC.
+
+# Posterior art
+[posterior-art]: #posterior-
+
+The best prior art for this removal is of course the inspiration for the de-RFC format, the [type ascription de-RFC].
+[MCP 630] can also be seen as prior art.
```
bjorn3, line 145: Should we add a reference to this de-RFC to the type ascription de-RFC as posterior art?

text/3829-de-rfc-unsized-locals.md
scottmcm: I think it probably depends greatly on the trait involved.  Trivially `dyn AsRef<[u8]>` is not materially better than `[u8]` on the stack, and dyn coroutines of various types can also be much too large to be good on the stack.

Though I do agree that `dyn SqlBackend` or similar might have much more of a reasonable size expectation that might make putting it on the stack less worrisome.

Jules-Bertholet: > Trivially `dyn AsRef<[u8]>` is not materially better than `[u8]` on the stack

And `impl AsRef<[u8]>` is also going to be equally bad, but we allow that today. https://github.com/rust-lang/rfcs/pull/3829#discussion_r2127287041

scottmcm:
@Noratrieb Do you have what you need to address the [concern](https://github.com/rust-lang/rfcs/pull/3829#issuecomment-2941831706) above?  AFAICT the RFC hasn't been updated since that was filed.

Noratrieb:
Apart from memory of this RFCs existence, which was evidently severely lacking in the last two months, yeah. I'll add it within the next week.

Noratrieb:
@traviscross I have pushed cff645d5f3d510e3f56692579e46b2d5201aa989, does that address your concerns?

traviscross:
No, not quite.  It says, e.g.:

> Alternatively, the existing design of `unsized_locals` *could* be RFCed again, if such an RFC was able to work around all the problems. This RFC merely unaccepts the previous `unsized_locals` RFC, it does not necessarily banish `unsized_locals` from existence forever.

But part of what this RFC is doing is setting out an opinion of what those problems are, so it almost sounds normative to suggest that those same problems would need to be worked around for another RFC to be accepted.  E.g.:

> As a _dangerous_ feature, dynamic stack allocation must be explicit and obvious - and `unsized_locals` makes it implicit _by design_, a major downside of this feature that has not been addressed in prior RFCs.

I want to be clear that we're not accepting as a normative position that "dynamic stack allocation must be explicit and obvious".  Maybe it should, maybe it shouldn't, but that's not a position I think we need to take to de-RFC this.

Probably I'd just do more to mark some sections as non-normative and then explicitly set out that the single normative effect is to unaccept the existing RFC such that someone would need to write a new one and have that accepted.

If you want (and are patient for me to get to it), I'm happy to take a pass at marking things clearly here.  I know it might just be difficult to guess what I'm looking to see.

Noratrieb:
Feel free to edit it to your likings :)

burdges:
Is there still some "commitment" to the original RFC motivation of passing unsized values?  Ala
```
fn takes_closure(f: dyn FnOnce()) { f(); }
```
It feels like more explicit schemes might break this, and and require borrows here, which maybe fine.

NobodyXu:
> Is there still some "commitment" to the original RFC motivation of passing unsized values?  Ala
> 
> ```
> 
> fn takes_closure(f: dyn FnOnce()) { f(); }
> 
> ```
> 
> It feels like more explicit schemes might break this, and and require borrows here, which maybe fine. 

[RFC 1909](https://rust-lang.github.io/rfcs/1909-unsized-rvalues.html) proposes something similar for unsized arguments and VLA

Noratrieb:
See the RFC.
> This does not have a negative effect on features that feature unsized values in function signatures like unsized_fn_params. Their behavior is much more clear and they are implemented differently. In fact, unsized_fn_params is currently needed in the standard library to implement Box<dyn FnOnce()> as FnOnce()>.

text/3829-de-rfc-unsized-locals.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Lack of Proper Implementation
+
+This feature has never been properly implemented in rustc.
+The variable length array form proposed in the RFC still doesn't exist at all, only the unsized local variable does.
+It is implemented in the type checker and codegen, but [lacking MIR semantics](https://github.com/rust-lang/rust/issues/48055#issuecomment-1837424794) and therefore unimplemented in the compile time function evaluator.
+This is very significant, as MIR semantics govern how the feature should behave precisely in the first place.
+Without them, they cannot work in `const` and optimizations are likely broken around them.
+Because of this lack of implementation quality, the `unsized_locals` feature was already accepted for removal from rustc two years ago in [MCP 630].
+This removal hasn't yet been implemented.
```
RalfJung, line 44: FWIW, it has meanwhile been implemented (https://github.com/rust-lang/rust/pull/141811).

PR #3841, opened 2025-07-23
Demote `x86_64-apple-darwin` from Tier 1 to Tier 2 with host tools

shepmaster:
[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/3841-demote-x86_64-apple-darwin.md)

### DIFF
--- a/text/3841-demote-x86_64-apple-darwin.md
+++ b/text/3841-demote-x86_64-apple-darwin.md
@@ -1,0 +1,154 @@
+- Feature Name: none
+- Start Date: 2025-07-23
+- RFC PR: [rust-lang/rfcs#3841](https://github.com/rust-lang/rfcs/pull/3841)
+- Rust Issue: [rust-lang/rust#145252](https://github.com/rust-lang/rust/pull/145252)
+
+# Summary
+[summary]: #summary
+
+Demote target `x86_64-apple-darwin` from Tier 1 to Tier 2 with host tools as this platform's lifetime is limited.
+
+Tier 2 with host tools means that the `x86_64-apple-darwin` target,
+including tools like `rustc` and `cargo`,
+is guaranteed to build but is not guaranteed to pass tests.
+
+This RFC does **not** propose removing the target completely from the codebase.
+
+# Motivation
+[motivation]: #motivation
+
+The `x86_64-apple-darwin` target has no long-term future.
+Upcoming changes will affect Rust's ability to ensure that the target meets the Tier 1 requirements,
+so we should demote it to Tier 2 with host tools in a controlled fashion.
+
+The most immediate critical change is that the free GitHub Actions macOS x86\_64 runners that the Rust project relies on will be [discontinued soon][macos-13-sunset].
+There is no known long-term replacement for these runners.
+
+## A brief timeline
+
+- 2020-06-22: Apple [announced plans][trans] to shift away from the x86\_64 architecture.
+- 2020-12-31: Rust [promoted `aarch64-apple-darwin`][aarch-tier-2] to Tier 2 with host tools.
+- 2023-06-05: Apple [announced the replacement][trans] of the last x86\_64 hardware.
+- 2023-10-02: GitHub [announces public GitHub Actions runners][m1-runners] for Apple silicon.
+- 2024-10-17: Rust [promoted `aarch64-apple-darwin`][aarch-tier-1] to Tier 1.
+- **2025-07-23**: This RFC opened.
+- 2025-09-01: GitHub [will discontinue][macos-13-sunset] providing free macOS x86\_64 runners for public repositories.
+- 2025 (Fall): [macOS 26][tahoe] will be the last macOS to support the x86\_64 architecture.
+- 2027: The [Rosetta 2][trans] compatibility layer will be mostly removed.
+
+[trans]: https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon
+[aarch-tier-2]: https://blog.rust-lang.org/2020/12/31/Rust-1.49.0/#64-bit-arm-macos-and-windows-reach-tier-2
+[aarch-tier-1]: https://blog.rust-lang.org/2024/10/17/Rust-1.82.0/#macos-on-64-bit-arm-is-now-tier-1
+[m1-runners]: https://github.blog/changelog/2023-10-02-github-actions-apple-silicon-m1-macos-runners-are-now-available-in-public-beta/
+[macos-13-sunset]: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down
+[tahoe]: https://en.wikipedia.org/wiki/MacOS_Tahoe
+
+## `x86_64-apple-darwin` popularity
+
+Looking at the [public download statistics][dl-stats] for the previous month (retrieved on 2025-07-21),
+we can see that `x86_64-apple-darwin` has substantially fewer downloads than `aarch64-apple-darwin`:
+
+[dl-stats]: https://p.datadoghq.com/sb/3a172e20-e9e1-11ed-80e3-da7ad0900002-60425c7cb1b7beb2e8959a305a301c0c?fromUser=false&refresh_mode=sliding&from_ts=1750525313022&to_ts=1753117313022&live=true
+
+### `rustc`
+
+| platform                   | downloads | percentage |
+|----------------------------|----------:|-----------:|
+| `x86_64-unknown-linux-gnu` |   194.38M |     81.58% |
+| `aarch64-apple-darwin`     |     7.15M |      3.00% |
+| `x86_64-apple-darwin`      |     2.74M |      1.15% |
+
+### `std`
+
+| platform                   | downloads | percentage |
+|----------------------------|----------:|-----------:|
+| `x86_64-unknown-linux-gnu` |    95.12M |     66.20% |
+| `aarch64-apple-darwin`     |     4.82M |      3.35% |
+| `x86_64-apple-darwin`      |     2.76M |      1.92% |
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The first release after this RFC is merged will be the last one with Tier 1 support for the `x86_64-apple-darwin` target.
+The release after that will demote the target to Tier 2 with host tools,
+which means we no longer guarantee that it will be tested by CI.
+
+Once this RFC is merged,
+a blog post will be published on the main Rust Blog announcing the change to alert users of the demotion.
+
+The demotion will also be mentioned in the release announcement for the last
+release with Tier 1 support, as well as the first release with Tier
+2 support.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The CI setup for [`rust-lang/rust`][r-l/r] will be modified to change the `dist-x86_64-apple` builder to no longer build the tests or run them.
+
+[r-l/r]: https://github.com/rust-lang/rust
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Without automated testing,
+this target will likely deteriorate more quickly.
+
+Users may be relying on Rust's Tier 1 support to provide confidence for their own artifacts.
+These users will be stuck on an old compiler version.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Rust CI could use emulation,
+  such as that provided by [Rosetta 2][rosetta-2].
+  Lightweight experiments show that this may increase CI times by a factor of 3
+  (e.g. a step taking 200 seconds would now take 600 seconds).
+  This would be a temporary solution,
+  as eventually Apple will [sunset Rosetta 2][trans].
+
+  We may choose to run tests in emulation even after the target is demoted to Tier 2 with host tools.
+  That change would be evaluated independently from this RFC and in a similar fashion to other non-Tier-1 targets with extra testing.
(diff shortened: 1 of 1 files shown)


text/3841-demote-x86_64-apple-darwin.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- None
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+`aarch64-apple-darwin` could be demoted to Tier 3 or support completely removed.
```
jdno:
```suggestion
`x86_64-apple-darwin` could be demoted to Tier 3 or support completely removed.
```

text/3841-demote-x86_64-apple-darwin.md
```diff
@@ -0,0 +1,146 @@
...
+is guaranteed to build but is not guaranteed to pass tests.
+
+This RFC does **not** propose removing the target completely from the codebase.
+
+# Motivation
+[motivation]: #motivation
+
+The `x86_64-apple-darwin` target has no long-term future.
+Upcoming changes will affect Rust's ability to ensure that the target meets the Tier 1 requirements,
+so we should demote it to Tier 2 with host tools in a controlled fashion.
```
RalfJung, line 22:
```suggestion
so we should demote it to Tier 2 with host tools in a controlled fashion.
Specifically, the free macOS x86\_64 runners that rustc CI is currently relying on will be discontinued soon, and there is no good long-term replacement for them available.
```
I felt like the RFC lacked some context for which part of the timeline below is the most immediately critical for this concrete step.

text/3841-demote-x86_64-apple-darwin.md
```diff
@@ -0,0 +1,149 @@
...
+
+The most immediate critical change is that the free GitHub Actions macOS x86\_64 runners that the Rust project relies on will be discontinued soon.
```
marcoieni, line 24:
```suggestion
The most immediate critical change is that the free GitHub Actions macOS x86\_64 runners that the Rust project relies on will be [discontinued](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down) soon.
```

Noratrieb:
Starting FCP already, we don't have many options here.
@rfcbot fcp merge

text/3841-demote-x86_64-apple-darwin.md
```diff
@@ -0,0 +1,149 @@
...
+| `x86_64-unknown-linux-gnu` |    95.12M |     66.20% |
+| `aarch64-apple-darwin`     |     4.82M |      3.35% |
+| `x86_64-apple-darwin`      |     2.76M |      1.92% |
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The first release after this RFC is merged will be the last one with Tier 1 support for the `x86_64-apple-darwin` target.
+The release after that will demote the target to Tier 2 with host tools,
+which means it will not be tested by CI.
```
RalfJung: FWIW, we do run some tests for non-tier-1-targets, e.g. some i586 targets. We could probably do the same for x86_64-apple-darwin? IOW, the consequence of this RFC is that we stop *promising* that we run all the tests, and we'll drop what is impractical, but we don't have to drop everything immediately. We could, for instance, keep running the library tests under rosetta, which shouldn't take too long.

jieyouxu: Yes, it may be *feasible* for `rust-lang/rust` to run a subset of tests still for `x86_64-apple-darwin`, this RFCs just drops that *promise*. This wording is somewhat accurate in that it's definitely a *possibility* for us to not run *any* tests (as per Tier 2 support level), even if that seems unlikely.

RalfJung: I think that would be good to communicate clearly, also in the blog post: we don't actually plan to drop *all* tests immediately, we'll continue testing some things under Rosetta where feasible, etc.

Or is our plan really to drop all tests entirely when we stop using the free macos-13 runners?

jieyouxu: > Or is our plan really to drop all tests entirely when we stop using the free macos-13 runners?

I think the intention is to keep that open as a possibility in case it becomes necessary.

RalfJung: Right, that's what I figured. I think this needs to be quite explicit, or people will conclude that we are dropping all tests immediately.

shepmaster: My intention is:

1. Make it clear in this RFC that there is *no guarantee* that tests will be run. 
2. When the RFC is accepted, immediately stop building the tests. This implies no longer running them, as well as no longer blocking merging PRs or releases if those tests fail, etc. This is the fastest solution to avoid spending money on CI that we'd rather not.

At that point, the infra team would be happy to collaborate if someone investigates how to run the tests in emulation. We can then evaluate that change like we have for any of the other non Tier-1 targets that have similar setups, including aspects like CI complexity, flakiness, execution time, ability of contributors to have access to the hardware to fix issues, etc. Any such setup will also be at the whim of various Rust teams to reduce / remove at any point.

I worry that by saying anything in this RFC about the slightest possibility of tests being run, people will disregard the whole thing and then if we don't run tests / stop running tests, will be upset that we misled them.

jieyouxu:
Friendly nudge for the remaining FCP reviewers (would not have done so if it's not as time-sensitive), as on **2025-09-01** GitHub [will discontinue](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down) providing free macOS x86_64 runners for public repositories, which won't leave us a lot of time for doing the demotion accounting for FCP time.

cc @cjgillot @nagisa @spastorino @wesleywiser

text/3841-demote-x86_64-apple-darwin.md
shepmaster: > At that point [...] remove at any point.

Would it meet your goals if I put something like this paragraph into the "future possibilities" section?

RalfJung: > When the RFC is accepted, immediately stop building the tests.

Okay that is not what I expected. I thought we'd continue to run some of the tests, as we do for a whole bunch of other tier 2 targets. Given the use this target still has, removing *all* testing seems excessive. IIUC, we can keep running tests on the newer macos runners with a 3x slowdown, which should make it fairly easy to still keep some test coverage.

RalfJung:
@rfcbot concern dont-commit-to-dropping-all-test-coverage

(Not sure if I can raise concerns here? I'm in t-compiler but not a "maintainer".)

I think we shouldn't prescribe in the RFC that we stop all tests. I think we should just say we stop *guaranteeing* that all tests are run. That's a big difference. The stats show there's still a lot of x86_64-apple users out there, so it's too early to drop all test coverage entirely.

text/3841-demote-x86_64-apple-darwin.md
jieyouxu: > I thought we'd continue to run some of the tests, as we do for a whole bunch of other tier 2 targets.

I don't think we can necessarily guarantee that[^tier-2]. We will need to do the demotion, *even if* it means no tests are run. *Even if* it might be possible to run a subset of tests under emulation.

That is, it's probably clearest if this RFC is interpreted as the "worst case" scenario, where all testing has to be dropped. It might be possible to re-add some test coverage for x86_64 macos, but this is not guaranteed.

[^tier-2]: Even other Tier 2 tests are by no means guaranteed at all.

jieyouxu: (Forcing an inline thread) @RalfJung you said

> I think we shouldn't prescribe in the RFC that we stop all tests. I think we should just say we stop guaranteeing that all tests are run. That's a big difference. The stats show there's still a lot of x86_64-apple users out there, so it's too early to drop all test coverage entirely.

What wording do you have in mind instead? I share the concern about

> I worry that by saying anything in this RFC about the slightest possibility of tests being run, people will disregard the whole thing and then if we don't run tests / stop running tests, will be upset that we misled them.

I don't think we can make the *guarantee* or *imply* that we *shall* retain test coverage for `x86_64-apple-darwin` in the RFC, *even if* that ends up viable in practice and ends up being implemented. Be it through emulation or whatever.

That is,

> I think we should just say we stop guaranteeing that all tests are run.

to me sounds like it implies "a subset of tests will be guaranteed to run", which is not quite true, because AFAIUI we cannot make that guarantee.

jieyouxu: I think a midway wording re.

> I think we shouldn't prescribe in the RFC that we stop all tests.

might be something like

> As per guarantees for Tier 2 targets, users should *expect* that testing of the `x86_64-apple-darwin` target in CI will be stopped, even though some subset of test coverage of the target may be retained or reimplemented where feasible.

But again, this wording sounds like "we guarantee a subset of tests will still be run", which isn't something we can guarantee AFAIK.

RalfJung: > I don't think we can necessarily guarantee that

The RFC currently basically suggests that we *guarantee that we don't run any tests*.
That is not the same as *not guaranteeing that we run all the tests*.

RalfJung: > to me sounds like it implies "a subset of tests will be guaranteed to run"

It very clearly doesn't say that. It makes no sense to say something we don't want to do out of fear that someone isn't able to understand what we are actually saying.

> But again, this wording sounds like "we guarantee a subset of tests will still be run", 

It doesn't. Not at all.

jieyouxu:
@rfcbot concern test-coverage-wording

> I think we shouldn't prescribe in the RFC that we stop all tests. I think we should just say we stop guaranteeing that all tests are run. That's a big difference. The stats show there's still a lot of x86_64-apple users out there, so it's too early to drop all test coverage entirely.
>
> -- https://github.com/rust-lang/rfcs/pull/3841#issuecomment-3127771502

text/3841-demote-x86_64-apple-darwin.md
jieyouxu: I'm not a native speaker, so I'll delegate the wording assessment to other people. In any case, I just want to emphasize that it should be taken as "we can't *guarantee* *any* test coverage for `x86_64-apple-darwin` even if a subset of test coverage ends up being feasible".

To clarify, IIUC, I think your emphasis is for the RFC wording to not imply that we *must* drop *all* test coverage for `x86_64-apple-darwin`, right?

jieyouxu: (In any case, I think this particular comment thread is basically just https://github.com/rust-lang/rfcs/pull/3841#discussion_r2237048012, so collapsing it in favor of the other thread.)

jieyouxu: > The RFC currently basically suggests that we _guarantee that we don't run any tests_. That is not the same as _not guaranteeing that we run all the tests_.

So something like

> This means that it is no longer guaranteed that the `x86_64-apple-darwin` target will be tested in CI

Which includes the possibility that no tests for `x86_64-apple-darwin` is run at all?

RalfJung: Yes. It's definitely a possibility, but it's a matter of continuous evaluation.

shepmaster: Is this better?

```diff
diff --git a/text/3841-demote-x86_64-apple-darwin.md b/text/3841-demote-x86_64-apple-darwin.md
index e1e72361..942446ae 100644
--- a/text/3841-demote-x86_64-apple-darwin.md
+++ b/text/3841-demote-x86_64-apple-darwin.md
@@ -71,7 +71,7 @@ we can see that `x86_64-apple-darwin` has substantially fewer downloads than `aa
 
 The first release after this RFC is merged will be the last one with Tier 1 support for the `x86_64-apple-darwin` target.
 The release after that will demote the target to Tier 2 with host tools,
-which means it will not be tested by CI.
+which means we no longer guarantee that it will be tested by CI.
 
 Once this RFC is merged,
 a blog post will be published on the main Rust Blog announcing the change to alert users of the demotion.
```

RalfJung: Yes, I would prefer that.

jieyouxu:
Resolving test-coverage-wording concern given [positive feedback](https://github.com/rust-lang/rfcs/pull/3841#discussion_r2237872339) and wording adjustment.

@rfcbot resolve test-coverage-wording

stacksjb:
This seems extremely early, given that Apple is continuing to [support MacOS Intel until 2028](https://www.pcmag.com/news/apple-confirms-end-of-support-for-intel-macs-after-macos-tahoe) 

Is the primary issue an inability to provide ongoing support due to GitHub Runner deprecation (which is still largely up in the air, as newer OS images can likely run on AMD64 - see https://github.com/actions/runner-images/issues/12520 )

emilyalbini:
> Is the primary issue an inability to provide ongoing support due to GitHub Runner deprecation (which is still largely up in the air, as newer OS images can likely run on AMD64 - see https://github.com/actions/runner-images/issues/12520 )

Newer OS images are available for x86_64, but only on the paid large runners. We can only use the free macOS runners. The part of the announcement relevant to us is that `macos-13` is the only x86_64 image available on free runners.

OBITOONDEADO1MZ (approved):

xtqqczze:
> Newer OS images are available for x86_64, but only on the paid large runners. We can only use the free macOS runners. The part of the announcement relevant to us is that `macos-13` is the only x86_64 image available on free runners.

FYI: The `macos-26-intel` (x86_64) standard runner is available until 2028.

PR #3854, opened 2025-09-12
RFC for `#[stable(since)]`

kornelski:
RFC for the issue https://github.com/rust-lang/rust/issues/74182

[Rendered](https://github.com/kornelski/rfcs/blob/stable_since/text/0000-stable_since.md)

### DIFF
--- a/text/0000-stable_since.md
+++ b/text/0000-stable_since.md
@@ -13,12 +13,12 @@ Allow crates to specify `#[stable(since = "version")]` on items for Rustdoc to r
 
 This functionality is already implemented for the standard library. Other crates also expand their public API over time, and it would be helpful to inform users about the minimum crate version required for each item.
 
-It's possible to automatically infer in which version each item has been added by comparing public APIs of crates' public releases, but this is too expensive and complicated to perform each time when generating documentation. The `#[stable(since)]` attribute behaves like a first-class cache for this information for `rustdoc` and other tools.
+It's possible to automatically infer in which version each item has been added/changed by comparing public APIs of crates' public releases, but this is too expensive and complicated to perform each time when generating documentation. The `#[stable(since)]` attribute can provide this information in a way that is readily available for `rustdoc` and other tools.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-When added to an item, it specifies in which version of the crate this item has been added:
+When added to an item, it specifies the oldest version of the crate that has this item available, and implemented in a way that is compatible with the crate's current version:
 
 ```rust
 #[stable(since = "2.25.0")]
@@ -29,19 +29,27 @@ Rustdoc will include the version in the documentation of the item, with a descri
 
 To ease development of unreleased features, there is no restriction on the version range, and it may refer to a future not-yet-released version.
 
+The version in `since` must be updated when the interface changes in a semver-incompatible way.
+
+The version in `since` should be updated when the behavior of the item changes significantly. If in doubt, it should be the later version. Dependency management tools may use this version to suggest bumping miniumum required version in `Cargo.toml`.
+
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-The `version` must parse as SemVer. It refers to a version of the crate that defines the item this attribute belongs to.
+The version in `since` refers to a version of the crate that defines the item this attribute belongs to. The version should parse as Cargo-compatible SemVer, but this won't be enforced by `rustc`, only `clippy` (same as `#[deprecated]`).
 
-`rustc` doesn't need to be made aware of crates' versions, because there's no restriction on the version range.
+`rustdoc` should not display the attribute on items defined in other crates.
 
-`rustdoc` should not display the attribute on items re-exported from other crates.
+This attribute on `pub use` refers to availability of the path.
+
+This attribute may be set on public items that are not accessible outside of the crate. This attribute set on private items may trigger a warning.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-This attribute may be incorrect if added manually. It may be confused with rustc version compatibility.
+It's not obvious how this attribute should be used with APIs that change over time. It's simpler to only specify when an item has been added, but this can be misleading if the latest version differs significantly from the earliest version. Specifying the lowest *compatible* version is more useful for choosing the minimum required crate version, but this type of compatibility is hard to define (semver is well understood when upgrading, but downgrading loses new features and reintroduces old bugs, so even a semver-patch downgrade may be breaking).
+
+This attribute may be incorrect if added manually, or become stale if not updated after a significant change.
 
 It specifies only a single version per item, which may not be enough to fully explain availability of items that are available conditionally or under different paths.
 
@@ -50,32 +58,42 @@ Versions on re-exported items are not relevant for the crate re-exporting them, 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-- The entire `#[stable(feature)]`/`#[unstable(feature)]` functionality could be stabilized for 3rd party crates
-- API stability could be stored outside of the source code, e.g. in a file similar to `rustdoc`'s JSON
-- It could be shortened to `#[since("version")]`
-- It could be expanded to `#[stable(added = "version", changed = "version", rust_version = "msrv")]`
+The version used in third party crates is unlikely to be confused with MSRV, because very few crates have only versions in the 1.xx range relevant for Rust versions.
+
+`since` is used for consistency with `#[deprecated(since)]`.
+
+Alternatives:
+
+- The entire `#[stable(feature)]`/`#[unstable(feature)]` functionality could be stabilized for 3rd party crates.
+- API stability could be stored outside of the source code, e.g. in a file similar to `rustdoc`'s JSON.
+- It could be generalized to `#[changed(version = "semver", note = "how")]` that provides a changelog for each item. This could also be implemented by allowing multiple instances of `#[stable(since = "version", note = "changes")]`.
+- docs.rs could generate `rustdoc` JSON for all crates.io crates, making it easier to diff releases and annotate API changes automatically.
 
 # Prior art
 [prior-art]: #prior-art
 
-The `#[stable(since = "version")]` attribute is a subset of standard library's `#[stable(feature, since)]`, but to keep the scope small, this RFC does not include support for feature flags nor unstable APIs outside of the standard library.
+The `#[stable(since = "version")]` attribute is a subset of standard library's `#[stable(feature, since)]`. This RFC does not include support for feature flags nor unstable APIs outside of the standard library. The standard library has unusual backwards-compatibility constraints, so it doesn't serve as a guide for hanlding APIs with breaking changes.
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-Should crates reset the `version` when making semver-breaking changes to the item?
+Should it specify the oldest version regardless of compatibility? Should it be bumped only on semver-breaking API changes, or also when the behavior changes?
 
-Should the `version` allow a placeholder value like `UNRELEASED`?
+Should it be allowed on private items? (there's `--document-private-items`, but those items won't be accessible from outside of the crate).
 
-Is it clear enough that the version is the crate's own version and not the minimum requierd Rust version?
-
-How to support items re-exported from other crates? Could `use` support overriding `#[stable(since)]`?
+Should it support specifying different kinds of stability, like `const_stable`?
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
 The entire `#[stable(feature)]`/`#[unstable(feature)]` functionality could be stabilized for 3rd party crates.
 
-Tools like rust-analyzer or clippy could help users bump versions in `Cargo.toml` when their crate uses items from a newer version of a dependency than the minimum version specified in `Cargo.toml`.
+A placeholder like `NEXT` or `UNRELEASED` could be supported, and either automatically updated or rejected by `cargo publish` (clippy allows "TBD" in `deprecated(since)`).
+
+A new `#[changed]` attribute could be added for tracking history of incompatible changes or extensions. This RFC proposes `#[stable]` to track the latest incompatible change, so `#[changed]` would provide information about partial compatibility with older versions, or a changelog.
+
+Tools like rust-analyzer or clippy could help users bump versions in `Cargo.toml` when their crate uses items from a newer version of a dependency than the minimum version specified in `Cargo.toml` (like `clippy::incompatible_msrv` for crates).
 
 These attributes could be automatically generated by tools like `cargo-public-api` or `cargo-semver-checks`.
+
+These attributes could be used by tools that auto-generate changelogs.


text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+
+Versions on re-exported items are not relevant for the crate re-exporting them, because it matters when the re-export has been added.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- The entire `#[stable(feature)]`/`#[unstable(feature)]` functionality could be stabilized for 3rd party crates
+- API stability could be stored outside of the source code, e.g. in a file similar to `rustdoc`'s JSON
+- It could be shortened to `#[since("version")]`
+- It could be expanded to `#[stable(added = "version", changed = "version", rust_version = "msrv")]`
```
kennytm: :thinking: Does "msrv" really belong here? If the `rustc` version is too old the crate would simply refuse to be compiled, regardless if the you the tagged API or not. Besides you can't bump the MSRV without publishing the package, so it can be inferred from the `changed` crate version already.

Lokathor: I could see the argument that a person would maybe want to know "if I use this then I can't go back in rust version farther then X", but that's extremely rare. Usually people seem to have a particular (possibly locked) version of rust available, and then want to know what they can do while on that same version of rust.

text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+The `#[stable(since = "version")]` attribute is a subset of standard library's `#[stable(feature, since)]`, but to keep the scope small, this RFC does not include support for feature flags nor unstable APIs outside of the standard library.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Should crates reset the `version` when making semver-breaking changes to the item?
+
+Should the `version` allow a placeholder value like `UNRELEASED`?
+
+Is it clear enough that the version is the crate's own version and not the minimum requierd Rust version?
```
Lokathor: Are you thinking that a person would see a 1.0-series crate with "stable since 1.34" or something, and then think that it means that the crate provides an item as long as the rust version is 1.34? Because that's... Not How It Works for anything else in rust crates. A particular user *could* think that, but as soon as they ask almost anyone else I'm sure they'd be corrected.

text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+[prior-art]: #prior-art
+
+The `#[stable(since = "version")]` attribute is a subset of standard library's `#[stable(feature, since)]`, but to keep the scope small, this RFC does not include support for feature flags nor unstable APIs outside of the standard library.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Should crates reset the `version` when making semver-breaking changes to the item?
+
+Should the `version` allow a placeholder value like `UNRELEASED`?
```
Lokathor: I think a rule like "you must put a string that can be a cargo crate version" is a simple enough rule, and however people want to get fancy with it is up to the ecosystem. For example, people could put the next minor version to be released, or put 0.0.0 as a placeholder, or whatever fits the project's tools and workflow.

text/0000-stable_since.md
hanna-kruppe: Note, however, that there *are* real world examples of crates providing their own APIs based on the Rust version (e.g., `camino` and `fs-err` wrap std APIs and use Rust version detection to support new additions without constantly bumping MSRV).

text/0000-stable_since.md
kornelski: It is indeed, but if people are going to use `999.999.999` or `0.0.0-unreleased` etc. as placeholders, we could as well provide some placeholder out of the box.

text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This attribute may be incorrect if added manually. It may be confused with rustc version compatibility.
+
+It specifies only a single version per item, which may not be enough to fully explain availability of items that are available conditionally or under different paths.
+
+Versions on re-exported items are not relevant for the crate re-exporting them, because it matters when the re-export has been added.
+
+# Rationale and alternatives
```
epage, line 58: There can be multiple ways something can be stable.  Today, `std` also uses `const_stable`.  I assume this should at least be acknowledged in case it can affect the design.  Today they are separate attributes.  Should we instead generalize `stable` to make it work for both?  What impact may that have on this attribute?

text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+Is it clear enough that the version is the crate's own version and not the minimum requierd Rust version?
+
+How to support items re-exported from other crates? Could `use` support overriding `#[stable(since)]`?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+The entire `#[stable(feature)]`/`#[unstable(feature)]` functionality could be stabilized for 3rd party crates.
+
+Tools like rust-analyzer or clippy could help users bump versions in `Cargo.toml` when their crate uses items from a newer version of a dependency than the minimum version specified in `Cargo.toml`.
```
epage: I would love to have something like `clippy::incompatible_msrv` for dependencies.  We'd have to tell the "compiler" (clippy in this case) the minimum version of every direct dependency (and maybe their pub dependencies)

text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+
+Should crates reset the `version` when making semver-breaking changes to the item?
+
+Should the `version` allow a placeholder value like `UNRELEASED`?
+
+Is it clear enough that the version is the crate's own version and not the minimum requierd Rust version?
+
+How to support items re-exported from other crates? Could `use` support overriding `#[stable(since)]`?
+
+# Future possibilities
```
epage, line 86: `cargo update` could tell people to run some new command, like `cargo report news clap@before clap@after` and see what is newly available.

Even better if we can get some form of "behavior change" and "remove attributes" (along with the stable `deprecated`) along with build-flow analysis and tailor those parts of the report specifically to your use of the library (plus the new stuff)

text/0000-stable_since.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- The entire `#[stable(feature)]`/`#[unstable(feature)]` functionality could be stabilized for 3rd party crates
+- API stability could be stored outside of the source code, e.g. in a file similar to `rustdoc`'s JSON
+- It could be shortened to `#[since("version")]`
+- It could be expanded to `#[stable(added = "version", changed = "version", rust_version = "msrv")]`
+
+# Prior art
```
epage, line 72: Should we also talk about `#[deprecated]` as it serves a similar role.

text/0000-stable_since.md
epage: Would love placeholder support as well as a teaching aid to use it (e.g. a clippy lint saying that the `since` is too new)

How does `#[deprecated(since)]` deal with any of this?

epage:
Unsure what has been holding this up but would really love to have this!

clarfonthey:
One thing worth considering here that distinguishes crate APIs from the standard library is that so far, the standard library has not introduced any truly breaking changes to things, but crates do all the time. It might be worth investigating how to properly distinguish that the stability attributes specifically indicate the stability of the current version of an API, and not just when the API was introduced.

text/0000-stable_since.md
kennytm: https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.deprecated.syntax

"`rustc` does not currently interpret the string, but external tools like Clippy may check the validity of the value."

`#[deprecated(since = "abcdef")]` currently does not produce any lints or errors, but clippy denies with [`clippy::deprecated_semver`](https://rust-lang.github.io/rust-clippy/master/index.html#deprecated_semver). It does recognize the special string `"TBD"` though (rust-lang/rust-clippy#11850).

text/0000-stable_since.md
epage: I feel like `changed` will need its own attribute, stacked on top of each other.

Should this be in future possibilities?  How I've been visualizing all of this in my head, `changed` is another important element I would like us to have a serious discussion about at some point.

Huh, I guess `changed` could be added as an unstable attribute and applied to `home` as a way to start playing with it.

epage:
@clarfonthey 

> One thing worth considering here that distinguishes crate APIs from the standard library is that so far, the standard library has not introduced any truly breaking changes to things, but crates do all the time. It might be worth investigating how to properly distinguish that the stability attributes specifically indicate the stability of the current version of an API, and not just when the API was introduced.

I would prefer "first add" and we provide something else to track changes, like a `#[changed]` attribute, see also https://github.com/rust-lang/rfcs/pull/3854#discussion_r2349997019

text/0000-stable_since.md
Lokathor: Changed can't really be a future possibility, because at the rate these sorts of language things happen, people will have `stable(since)` for X many months already, and already think of the attribute as meaning one thing, and then eventually another attribute comes in and it changes how people should think about that.

So either it's in the initial design and all stabilized at once, or it can't really happen at all.

Lokathor:
Personally, with the way I usually think about this, I'd like `stable(since)` to mean "stable in this form since", and the fact that something might have existed in some other way before the current iteration is just not as important. I don't need to read about something being stable since 1.6 and changed in 2.0 and then changed in 3.0, the 1.6 and 2.0 version info doesn't really help.

text/0000-stable_since.md
kornelski: I don't think `changed` must be *stabilized* together. It should be sufficient to have some plan for future extensions. The `stable` attribute can define how it's supposed to be used with APIs that changed, and attributes for providing extra information can be added later.

jhpratt:
While I have a vision for reforming stability attributes in general and exposing them to all crates, I don't see any scenario where this is _not_ a subset of what is ultimately decided.

Huge :+1: from me. My primary suggestion is to include a recommendation (but not a hard requirement) for a lint for future versions. Having `since = "NEXT"` or similar (as mentioned) would also be very useful for scenarios where you don't know if it's a patch, minor, or major version that'll be released next.

text/0000-stable_since.md
kpreid: >  If the `rustc` version is too old the crate would simply refuse to be compiled

This is not true if the package detects the `rustc` version in use, which is already sometimes done through build scripts, and which #3857 proposes to make possible with a simple `cfg`.

text/0000-stable_since.md
jhpratt: I think it's best to start with the simplest possible case and worry about the full situation later. I have some notes and design sketches for my thoughts if you'd like to discuss it on Zulip.

jmillikin:
If this attribute is intended for documentation and won't be enforced by the compiler, would it make sense to put it in the `#[doc(...)]` attribute?

```rust
#[doc(since = "2.25.0")]
pub fn add_trombone_emoji() {}
```

text/0000-stable_since.md
```diff
@@ -0,0 +1,99 @@
...
+The `#[stable(since = "version")]` attribute is a subset of standard library's `#[stable(feature, since)]`. This RFC does not include support for feature flags nor unstable APIs outside of the standard library. The standard library has unusual backwards-compatibility constraints, so it doesn't serve as a guide for hanlding APIs with breaking changes.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Should it specify the oldest version regardless of compatibility? Should it be bumped only on semver-breaking API changes, or also when the behavior changes?
+
+Should it be allowed on private items? (there's `--document-private-items`, but those items won't be accessible from outside of the crate).
+
+Should it support specifying different kinds of stability, like `const_stable`?
```
jmillikin, line 84: I think this would be useful, because marking a function as `const` is forward-compatible.

If the `#[doc(...)]` syntax is used then an item might be annotated `#[doc(since="1.1", const_since="1.23")]`.

kornelski:
When considering *why* anyone needs this information, the primary use-case seems to be choosing the minimum required version of a crate (especially if the latest version of the crate requires a very new MSRV).

`#[stable(since = "added")]` with the first version in which an item has appeared seems more natural given the "stable since" terminology. However, it may be unsuitable for choosing the correct minimum required version if the oldest implementation isn't a compatible substitute for the latest one. What exactly is breaking for downgrades is hard to define, even harder than semver for upgrades.

There may be multiple kinds of features/behaviors stabilized separately. libstd has `const_stable`, but there may be library-specific functionality added. For example, [`wgpu` features](https://docs.rs/wgpu/latest/wgpu/struct.FeaturesWGPU.html) may be added to Vulcan backend much earlier than to WebGPU backend, and this distinction matters a lot to users who build for browsers.

So I'm wondering whether some changelog-like attribute would be better. If multiple `#[changelog(version, note)]` attributes were allowed, it would be easy to log multiple kinds of changes over time. `min(version)` would be "stable since", and `max(version)` would be a conservative minimum required crate version.

kornelski:
There's going to be a `cfg()` for Rust version: https://github.com/rust-lang/rust/issues/64796 which together with `doc(cfg)` could be a way to display MSRV for each item in the documentation.

The `since` version in libstd is effectively both MSRV and the minimum libstd crate version, but in 3rd party crates these two could diverge, and theoretically both could be available at the same time.

jmillikin:
> `#[stable(since = "added")]` with the first version in which an item has appeared seems more natural given the "stable since" terminology. However, it may be unsuitable for choosing the correct minimum required version if the oldest implementation isn't a compatible substitute for the latest one. What exactly is breaking for downgrades is hard to define, even harder than semver for upgrades.

Let's say `foolib-v1.1` adds a new function `fn decode_u64(buf: &[u8]) -> u64`. That symbol could be annotated with `#[doc(since = "1.1")]` to aid the reader in knowing which minimum version of `foolib` they need to depend on.

If `foolib-v1.2` changes the behavior of `decode_u64` to support a new encoded format then it doesn't make sense to change its availability annotation. Similarly if there was a bug that returned the wrong value for some inputs and v1.1.1 fixed that bug then it wouldn't make sense to bump the "available since" annotation.

There could be an annotation to mark the most recent version that modified a given function's observable behavior, but that seems like it would be difficult to keep in sync with the actual code -- imagining a deep call graph, does a fix in some inner core logic cause the entire API surface to get its "API mtime" bumped...?

> So I'm wondering whether some changelog-like attribute would be better. If multiple `#[changelog(version, note)]` attributes were allowed, it would be easy to log multiple kinds of changes over time. `min(version)` would be "stable since", and `max(version)` would be a conservative minimum required crate version.

This seems like it would be a better fit for prose, such as a summary of changes in the API comment or a traditional `CHANGELOG.txt`. The `rustdoc` extended Markdown syntax for tables can be used to good effect for inline changelogs, because it's more flexible than a strict `(version, note)` tuple.

text/0000-stable_since.md
epage: A flat `doc(since)` isn't specific enough. `const_since` is at least more specific. Neither allows an associate description which this doesn't need but future possibilities do.

kornelski: `doc` is a rustdoc-specific attribute, and controls [how the docs are generated](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#at-the-item-level). 

I think availability and compatibility of APIs is more general than this. It's a semver/public API annotation that could be used beyond rustdoc.

Apart from that whether it's `doc(since)` or `stable(since)` is a bikeshed color.

Lokathor: It is a bike shed, but I think it's the correct color for this shed. We're giving documentation. It's orderly and specific, but it's still documentation information about the item/method.

kornelski:
@jmillikin you say it doesn't make sense to change the availability annotation, but don't justify why. If `decode_u64` started supporting a new format in v1.1, then for users needing the new format it would be important to know they can only use `decode_u64` since v1.1, and earlier versions won't work. I'm trying to address that problem by suggesting to bump the `since` version, so that users get a version that will be compatible with the behavior specified in the documentation they're reading. 

An attribute only specifying when an item has become available for the first time is simpler to define, but what would you use this information for? I'm trying to look at it from a perspective of use-cases to avoid merely adding facts because it's possible. This version needs to be useful for something. 


Maybe the issue could be redefined in a documentation-centric way (justifying `doc(since)` attr): saying that this documentation as written applies to the item since the given version. So if the docs say "the decode function supports the new format", it needs to refer to some later library version that added the new format support, not the oldest version that added the function without new format support.

PR #3859, opened 2025-09-22
enable range-diff triagebot feature

programmerjake:
as mentioned [on zulip](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Experimental.20range-diff.20for.20force-push/near/540688241)
https://forge.rust-lang.org/triagebot/range-diff.html

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -12,6 +12,8 @@ trigger-files = ["text/"]
 
 [shortcut]
 
+[range-diff]
+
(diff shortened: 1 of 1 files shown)


fmease:
While the repository rules explicitly state "don't rebase / squash / force-push" (see the `README.md`). I guess if an RFC author accidentally disobeys them, it's still handy to post the range-diff.

PR #3865, opened 2025-10-02
Emphasize do not squash/rebase commits

jieyouxu:
Could _maybe_ make e.g. https://github.com/rust-lang/rfcs/pull/3851#issuecomment-3317282798 less likely to happen.


jieyouxu:
r? internal-sites

Urgau:
Sure, why not.

Urgau (approved):

Urgau:
> `r? internal-sites`

Btw, we may want to make us auto-assigned for anything that doesn't touch the text directory.

jieyouxu:
Ah, good point. Even if we can't review it, we should try to help find reviewers who can.

PR #3867, opened 2025-10-05
forgettable auto trait

mcmah309:


bjorn3:
This is missing any discussion about `Rc` cycles that were the cause of the leakpocalypse and the reason why `mem::forget` was made safe. And it also doesn't provide a solution to how to handle `Rc` cycles in the face of `Forgettable`/`Leak`/... without breaking changes.

mcmah309:
This RFC isn't trying to solve the problem of `Rc` cycles. Rust's current safety guarantees do not include a guarantee that destructors will always run. A program can create a reference cycle using `Rc`,  call `process::exit` to exit without running destructors, etc. And this RFC does not attempt to change that. Instead it attempts to add the invariant that a `!Forgettable` will always have it's `Drop` ran before being freed, to ensure the rest of the program continues safely. In practice, this is less useful then saying "a `!Forgettable` will always have it's `Drop` ran before going out of scope", but this can not be accomplished without addressing the `Rc` cycle issue. ~~For the [scoped_static](https://github.com/mcmah309/scoped_static) example mentioned in the motivation section of the RFC, `Rc` cycles are not an issue since `ScopeGuard` contains a lifetime and it is safe for the duration of that lifetime.~~

Though this part of the RFC could be reworded to reflect this:
> Safe code can rely on `Drop` being ran for `T` where `T: !Forgettable` to prevent undefined behavior.

Edit: I stand corrected, I investigated this further and leaking through `Rc` cycles (thus forgetting) remains an issue for [scoped_static](https://github.com/mcmah309/scoped_static) crate:
```rust
fn cycle_forget<T>(data: T) {
    use std::rc::Rc;
    use std::cell::RefCell;

    struct Leak<T> {
        cycle: RefCell<Option<Rc<Rc<Leak<T>>>>>,
        data: T,
    }

    let e = Rc::new(Leak {
        cycle: RefCell::new(None),
        data: data,
    });
    *e.cycle.borrow_mut() = Some(Rc::new(e.clone()));
}
```
Which was a major motivation for this RFC. In light of such, the current proposed implementation does not force the invariant that "`!Forgettable` will always have it's Drop ran before being freed". Thus closing.

workingjubilee:
thank you for playing!

PatchMixolydic:
> I investigated this further and Rc cycle issue remains for [scoped_static](https://github.com/mcmah309/scoped_static) crate.

For posterity's sake, here's a worked example ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=955ff47dc09a9e24c673cd059c93a262)):
```rust
struct GuardLeaker<'a> {
    guard: Option<ScopeGuard<'a, Box<f64>>>,
    leaker_2: Mutex<Option<Arc<GuardLeaker<'a>>>>,
}

impl<'a> GuardLeaker<'a> {
    fn new(guard: Option<ScopeGuard<'a, Box<f64>>>) -> Self {
        Self {
            guard,
            leaker_2: Mutex::new(None),
        }
    }
}

impl<'a> Drop for GuardLeaker<'a> {
    fn drop(&mut self) {
        // This is never printed — this is only here to prove that this is leaked
        println!("GuardLeaker drop");
    }
}

fn create_leaker<'a>(guard: ScopeGuard<'a, Box<f64>>) -> Arc<GuardLeaker<'a>> {
    let leaker = Arc::new(GuardLeaker::new(Some(guard)));
    *leaker.leaker_2.lock().unwrap() = Some(leaker.clone());
    leaker
}

// Based on the readme example
#[tokio::main]
async fn main() {
    let value = Box::new(1.0);
    let ref_value = &value;

    let guard_leaker = create_leaker(ScopeGuard::new(ref_value));
    let lifted: Scoped<Box<f64>> = guard_leaker.guard.as_ref().unwrap().lift();

    // does not join
    tokio::spawn(async move {
        // `lifted` moved here
        let value = **lifted + 1.0;
        assert_eq!(value, 2.0);
        // `lifted` dropped
    });

    // `guard_leaker` drops and leaks `guard` while `lifted` is alive!
}
```

On the playground's copy of Rust 1.90.0, this yields:
```
thread 'tokio-runtime-worker' panicked at src/main.rs:125:9:
assertion `left == right` failed
  left: 1.0
 right: 2.0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Miri 2025-10-04 catches the UB (the primary span here is pointing into `<Scoped<T> as Deref>::deref`):
```
error: Undefined Behavior: constructing invalid value: encountered a dangling reference (use-after-free)
   --> src/main.rs:81:9
    |
 81 |         self.0.as_ref()
    |         ^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE on thread `tokio-runtime-w`:
    = note: inside `<Scoped<std::boxed::Box<f64>> as std::ops::Deref>::deref` at src/main.rs:81:9: 81:24
note: inside closure
   --> src/main.rs:124:22
    |
124 |         let value = **lifted + 1.0;
    |                      ^^^^^^^
```

Banning unforgettable types in `Rc` and `Arc` is a possible workaround, but that would not account for existing user-defined smart pointers, which would still allow leaking unforgettable types in safe Rust[^1].

(`forget` can also be replicated by sending an unforgettable `Send` value to a different thread that ignores all received values and runs until process termination, though I'm not sure how relevant this is to other proposals for unforgettable types. It doesn't seem that different from calling `exit(0)` at the end of `main`, though it certainly seems more annoying to detect without requiring that unforgettable types are also unsendable.)

At least some of these issues may be mitigated by making `Forgettable` a default bound, like `Sized`, which has already been proposed in [RFC 3782](https://github.com/rust-lang/rfcs/pull/3782).

[^1]: Unless such smart pointers are retroactively declared unsound, which seems counter to Rust's stability policy.

mcmah309:
Thanks for giving a full example.

> Banning unforgettable types in Rc and Arc is a possible workaround, but that would not account for existing user-defined smart pointers, which would still allow leaking unforgettable types in safe Rust[1](https://github.com/rust-lang/rfcs/pull/3867#user-content-fn-1-a4b1ce506f59dce7f922e9fe7d67c412).

Yes I think too often users will write code that can violate this invariant and it will be marked as safe.

In light of my discovery, which you explained in good detail, I don't believe a `Forgettable` default bound is very useful if an `!Forgettable` can still be forgotten through safe code and trigger UB. Thus my reason for closing.

mcmah309:
A possible solution to all this would be some sort of `defer` construct.

```rust
impl<'a, T: 'static> ScopeGuard<'a, T> {
    // notice `defer` keyword, like `const`, needs to be called from a `defer` block
    pub defer fn new(value: &'a T) -> Self {
       ...
    }
...
}

impl<'a, T: 'static> Defer for ScopeGuard<'a, T> {
    fn defer(self) {
     // We don't need anything here since drop will now run
   }
}

#[tokio::main]
async fn main() {
    let value = Box::new(1.0);
    let ref_value = &value;
    // `guard` ensures no derived "lifted" values exist when dropped and `defer` runs the implemented `Defer` code at the end of scope
    let guard: &mut ScopeGuard<Box<f64>> = defer { ScopeGuard::new(ref_value)) };
    // `lifted` holds a `'static` reference to `'ref_value`
    let lifted: Scoped<Box<f64>> = guard.lift();
    tokio::spawn(async move {
        // lifted moved here
        let value = **lifted + 1.0;
        assert_eq!(value, 2.0);
        // lifted dropped
    })
    .await
    .unwrap();
   /// Forgetting has no effect since a guard is a reference
   std::mem::forget(guard);
   // `Defer` code runs which drops the `guard`
}
```

Another more extensible possibility would be to not require `Defer` to be implemented on the type directly. So one can write arbitrary defer code.
```rust
impl<'a, T: 'static> ScopeGuard<'a, T> {
    pub defer fn new(value: &'a T) -> impl Defer<Self> {
       ...
       ScopeGuardDefer(guard)
    }
...
}

struct ScopeGuardDefer<'a, T: 'static>(ScopeGuard<'a, T>);

impl<'a, T: 'static>  Defer for ScopeGuardDefer<'a, T> {
    fn defer(self) {
   }
}

#[tokio::main]
async fn main() {
    ...
    let guard: &mut ScopeGuard<Box<f64>> = defer { ScopeGuard::new(ref_value)) };
    ...
}
```

Ddystopia:
@PatchMixolydic hi, your example is fully covered in the RFC you linked in a great detail. For details you can read its guide level explanation, while in a nutshell your example won't compile because you won't be able to send `!Forget` value using `Arc`-like channel, while if sent via borrowing channel borrow checker will reject lifetimes.

PR #3877, opened 2025-11-08
Add mdbook-extra-watch-dirs

ehuss:
This helps when using `mdbook serve` so that mdbook will automatically reload whenever a file changes.

This doesn't handle adding new files because that requires running `generate-book.py`, but that should be fine.

r? internal-sites

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -12,3 +12,6 @@ heading-split-level = 0
 
 [output.html.playground]
 runnable = false
+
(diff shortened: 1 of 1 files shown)


Urgau:
I was wondering how `serve` could know that the files in `text/` are linked to those in `src/` (as generated by `generate-book.py`), turns out they are just symlinks.

PR #3878, opened 2025-11-08
Apply several markup fixes

ehuss:
This includes several commits to fix various markdown errors in the rfc book. This is in preparation for mdbook 0.5 which is a little more sensitive to markdown errors (mostly generating warnings, or not rendering as expected). This should be compatible with both 0.5 and 0.4.

r? internal-sites

[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/0048-traits.md)

### DIFF
--- a/text/3537-msrv-resolver.md
+++ b/text/3537-msrv-resolver.md
@@ -31,7 +31,7 @@ likely only needing to follow the FCP process.
 
 ## Status Quo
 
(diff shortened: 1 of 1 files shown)


text/0048-traits.md
```diff
@@ -26,7 +26,7 @@ put onto a surer footing.
 
 ## Use cases
 
-### Poor interaction with overloadable deref and index <a name=overload>
+### Poor interaction with overloadable deref and index { #overload }
```
Urgau, line 29: I didn't knew this was a syntax, very fancy.

PR #3883, opened 2025-11-22
Update to mdbook 0.5

ehuss:
Changelog: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-051

[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/0048-traits.md)


Urgau:
The new sidebar is really nice. Exciting to use it in other places.

Regarding that sidebar, I noticed that it doesn't take into account Level 1 heading, is that expected? should we update all the RFCs to reduce the heading by one?

*We are missing the Summary, Motivation, ... headings on all RFCs.*

image

ehuss:
Hm, that's a good point. Yea, it is expected as mdbook expects a chapter to have only one level-1 heading (usually the chapter title at the top).

At the risk of causing complete chaos, I pushed a commit that corrects the heading levels. Pages generally shouldn't have multiple h1 headings anyways.

Urgau:
I'm fine updating all the headings.

However we should also update the level of all the other headings (h2 -> h3, h3 -> h4, ...), otherwise the sections no longer have the right parent.

| Your patch | Expected output |
|------|-----|
| image | image |

ehuss:
Well shoot. I thought my replacements handled that. I have re-pushed the commit which should fix that. I ended up using pulldown-cmark to more accurately detect the headings. Unfortunately a large number of RFCs had weird heading levels. So I went through manually visually inspected each RFC to see if the nesting looked odd and fixed any RFC that looked weird (that was maybe 10-15% of them).

Urgau (approved):
Well done. Looks good to me.

PR #3895, opened 2025-12-11
Removed incorrect sentence about the procedure

Noratrieb:
This is not true today at all.

What *may* need an RFC is adding a new capability that doesn't exist today, like uninit `freeze`, but those things are very rare so I think it doesn't make sense to put it here.


Urgau (approved):
Indeed, this is very much not true today.

*(I approve with my T-compiler hat on)*

Urgau:
And merge with my T-internal-sites hat on.

PR #3903, opened 2026-01-03
RFC: Trait Implementation Privacy with `permits`

HackingRepo:
[Rendered](https://github.com/HackingRepo/rfcs/blob/patch-1/text/0000-trait-permits.md)

### DIFF
--- a/text/0000-trait-permits.md
+++ b/text/0000-trait-permits.md
@@ -34,6 +34,8 @@ The `permits` clause allows trait authors to specify which crates may provide im
 pub trait Test permits crate, mycrate_extra {
     fn run(&self);
 }
+```
+
 crate refers to the defining crate.
 
 Other identifiers refer to external crates by name.


text/0000-trait-permits.md
```diff
@@ -0,0 +1,105 @@
...
+
+Introduce a new syntax for restricting trait implementations to specific crates:
+
+```rust
+trait Test permits mycrate_extra, crate {
+    fn run(&self);
+}
+```
+
+Motivation
```
kennytm: please format the section titles properly

text/0000-trait-permits.md
```diff
@@ -0,0 +1,105 @@
...
+- Start Date: 2026-01-03
+- RFC PR: rust-lang/rfcs#0000
+- Rust Issue: rust-lang/rust#0000
+
+## Summary
+
+Introduce a new syntax for restricting trait implementations to specific crates:
+
+```rust
+trait Test permits mycrate_extra, crate {
```
kennytm, line 13: how does the upstream trait declaration knows the name of the downstream crate name provide the impls? any crate can call themselves `mycrate_extra`, so this does not prevent "malicious impls" mentioned on line 21 at all

text/0000-trait-permits.md
```diff
@@ -0,0 +1,105 @@
...
+    fn run(&self);
+}
+```
+
+Motivation
+Rust currently allows any downstream crate to implement traits for types they own, subject to the orphan rules. While flexible, this can lead to:
+
+Accidental or malicious impls: External crates can implement traits in ways that break invariants.
+
+Audit difficulty: It is hard to know which crates are allowed to implement a trait.
```
Lokathor, line 23: It is generally very simple to work out how the Orphan Rule applies. The `fundamental` types work a little different, but there's only a handful of such types.

text/0000-trait-permits.md
HackingRepo: The permits clause isn’t based on string names alone. Rust already tracks crate identity through Cargo’s dependency resolution and compiler metadata. When I write permits mycrate_extra, the compiler enforces that only the actual crate resolved as mycrate_extra in the dependency graph can implement the trait. Another crate cannot simply call itself mycrate_extra and bypass this crate identity is tied to compilation units, not arbitrary names. In that sense, permits builds on the same machinery as the orphan rule: it uses the compiler’s notion of crate ownership to restrict trait implementations, preventing unauthorized or malicious impls.

text/0000-trait-permits.md
HackingRepo: You’re right — the orphan rule is generally simple to apply, with only a handful of exceptions for fundamental types like references, slices, arrays, and function pointers. But the orphan rule only governs who can implement a trait at all (must own the trait or the type). My RFC is tackling a different axis: even if a crate owns the type, it shouldn’t be able to implement the trait unless it’s explicitly permitted. In other words, the orphan rule ensures coherence, while permits would enforce privacy and trust boundaries. The two are complementary orphan rule prevents conflicting impls, permits prevents unauthorized impls.

text/0000-trait-permits.md
```diff
@@ -0,0 +1,105 @@
...
+
+Boilerplate sealed traits: Developers often use the "sealed trait" pattern to prevent external impls, which is verbose and indirect.
+
+By introducing permits, Rust gains a first-class mechanism for trait implementation privacy, improving safety and clarity.
+
+Detailed Design
```
kennytm: these are still not a title (and the current RFC template does not really have a "Detailed Design" section)

why aren't you copying `0000-template.md` and edit from there?

text/0000-trait-permits.md
```diff
@@ -0,0 +1,203 @@
...
+
+The `permits` clause allows trait authors to specify which crates may provide implementations of the trait.
+
+```rust
+pub trait Test permits crate, mycrate_extra {
+    fn run(&self);
+}
+crate refers to the defining crate.
+
+Other identifiers refer to external crates by name.
```
scottmcm, line 41: your markdown is broken

text/0000-trait-permits.md
```diff
@@ -0,0 +1,205 @@
...
+
+```rust
+pub trait Test permits crate, mycrate_extra {
+    fn run(&self);
+}
+```
+
+crate refers to the defining crate.
+
+Other identifiers refer to external crates by name.
```
JarredAllen, line 41: Is there a use for "I want this crate and a specific set of dependent crates named here to be the only crates who can implement this trait" restrictions? I can't think of any Rust code I've read/written where anything between "only this crate can implement" or "the whole world can implement" would be useful.

HackingRepo: Extension crates maintained by the same team: Large projects often split functionality across multiple crates (e.g. serde + serde_json, tokio + tokio-util). The core crate may want to restrict trait impls to itself and its officially maintained extension crates, while preventing arbitrary downstream impls that could break invariants.

Auditability and safety: Allowing only a known set of crates to implement a trait makes it easier to audit and reason about behavior. For example, if a trait encodes safety‑critical invariants, you don’t want random downstream crates introducing unsafe impls.

Avoiding boilerplate sealed traits: Today, the sealed trait pattern is used to enforce this restriction, but it’s verbose and indirect. A permits clause would make the intent explicit and first‑class.

Ecosystem clarity: It communicates to downstream users which crates are “officially sanctioned” to extend the trait, reducing confusion and accidental misuse.

Important! the crate should be listed in Cargo.toml, That not for arbiratry crates string.

text/0000-trait-permits.md
```diff
@@ -0,0 +1,205 @@
...
+```rust
+trait Test permits mycrate_extra, crate {
+    fn run(&self);
+}
+```
+
+## Motivation
+Rust currently allows any downstream crate to implement traits for types they own, subject to the orphan rules. While flexible, this can lead to:
+
+Accidental or malicious impls: External crates can implement traits in ways that break invariants.
```
teor2345, line 21: If a malicious impl can cause memory safety issues, then the trait must be marked `unsafe`. Or rewritten to defend against malicious impls.

If it can't, then API design, documentation, assertions, and tests are usually the way this gets resolved. Are there  specific scenarios where these aren't enough?

workingjubilee: To be charitable to the idea for a moment, a malicious impl *could* be internally unsound, technically, even if the functions are safe.

I am charitable only because that more thoroughly demonstrates the problem: one is *always* gambling that all the code you are calling is not internally unsound, and in that case e.g. a `Debug` or `PartialEq` impl is no different than any other trait's impls.

teor2345: In practice, malicious (or buggy) `PartialEq` is worse than a local trait, because it's called by a lot of library code, and by language operators. So reviewing / auditing all implementations and uses of `PartialEq` for bugs is tricky.

But at least it's a safe trait, so callers have to code defensively, because they have to accept incorrect `eq()` return values without any unsoundness.

Would an unsafe trait achieve some of the goals of this RFC?
Then maintaining invariants would be the trait implementer's responsibility - and callers would have clear safety preconditions.

It feels like that's roughly what's happening here, you're shifting responsibility for implementing the trait correctly on to the permitted implementers.

text/0000-trait-permits.md
```diff
@@ -0,0 +1,205 @@
...
+
+Audit difficulty: It is hard to know which crates are allowed to implement a trait.
+
+Boilerplate sealed traits: Developers often use the "sealed trait" pattern to prevent external impls, which is verbose and indirect.
+
+By introducing permits, Rust gains a first-class mechanism for trait implementation privacy, improving safety and clarity.
+
+## Guide-level explanation
+
+The `permits` clause allows trait authors to specify which crates may provide implementations of the trait.
```
teor2345, line 31: Is there an incremental way of introducing the changes in this RFC?

For example, standardising sealed traits would reduce the boilerplate, and make them more discoverable (and easier to teach).

Could we start with something like:
```rust
pub trait SealedTrait sealed {
    …
}
```

Then add syntax for additional permitted crates once there's a demonstrated need?

If that's the first step, can it be done outside the compiler by a macro?
(And then if the macro becomes popular, that's an argument for standardisation.)

text/0000-trait-permits.md
```diff
@@ -0,0 +1,205 @@
+# RFC: Trait Implementation Privacy with `permits`
+
+- Feature Name: `trait_permits`
+- Start Date: 2026-01-03
+- RFC PR: rust-lang/rfcs#0000
+- Rust Issue: rust-lang/rust#0000
+
+## Summary
+
+Introduce a new syntax for restricting trait implementations to specific crates:
```
ChayimFriedman2, line 10: [There is already an accepted RFC for syntax sugar for sealed traits](https://github.com/rust-lang/rfcs/pull/3323). It is less flexible than this RFC by not allowing putting external crates (which IMO is a bad idea), but more flexible by allowing specific paths within the current crate.

text/0000-trait-permits.md
ChayimFriedman2, line 41: It is worth noting than crates have no names in Rust currently. I'm not sure Cargo is even telling rustc the crate name, and I know rust-analyzer does not require a crate name. Only dependencies have names. So this is a new concept.

HackingRepo: The crates of course should be in Cargo.toml and installed when `use test_crate;`, it should exist in cargo toml after that the samething `permits test_crate`

PR #3906, opened 2026-01-07
Let `Option` derive `#[must_use]`

dhardy:
Let `Option` and `Box` derive `#[must_use]` from their generic parameter `T`.

[Rendered](https://github.com/dhardy/rfcs/blob/derive-must-use/text/0000-derive-must-use.md)

### DIFF
--- a/text/0000-derive-must-use.md
+++ b/text/0000-derive-must-use.md
@@ -1,0 +1,64 @@
+- Feature Name: Let `Option` derive `#[must_use]`
+- Start Date: 2026-01-07
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Let `Option` and `Box` derive `#[must_use]` from their generic parameter `T`.
+
+# Motivation
+[motivation]: #motivation
+
+If we write:
+```rust
+#[must_use]
+struct Redraw;
+
+fn do_thing() -> Option<Redraw> {
+    // Do some thing which requires a redraw...
+    Some(Redraw)
+}
+```
+then `do_thing` should be `#[must_use]`, and while we can apply the `#[must_use]` attribute to the function `do_thing`, we shouldn't have to (remember to do so).
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `Option` and `Box` types will "magically" have the `#[must_use]` attribute if and only if their generic parameter `T` does.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This will be an internal detail of the standard library. It may use another special attribute like `#[derive_must_use_from(T)]`, but for the purposes of this RFC, the `derive_must_use_from` attribute may remain unstable forever.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I see no drawbacks besides the small amount of complexity involved.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The only obvious (non-empty) alternative is to add (and stabilise) a new `#[derive_must_use_from(T)]` attribute and apply this to `Option<T>` (and [other types](#future-possibilities)).
+
+This would not be a strict alternative in that nothing prevents this from being done later.
+
+# Prior art
+[prior-art]: #prior-art
+
+[RFC #3737](https://github.com/rust-lang/rfcs/pull/3737) is vaguely related (only in that it also pertains to `#[must_use]`). Otherwise I do not see any prior art.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Possibly a few other standard library types would benefit from this derivation of `#[must_use]`:
+
+- `Box<T>` can do so (included in this RFC, though motivation is weaker)
+- `RefCell<T>` and `Mutex<T>` *could* do so but it is unlikely of any use
+- `Rc<T>`, `Arc<T>` and various reference types *should not* since they do/may not have exclusive ownership of the value
+- `Vec<T>` and other containers *could* do so
--- a/text/0000-derive-must-use.md
+++ b/text/0000-derive-must-use.md
@@ -1,6 +1,6 @@
 - Feature Name: Let `Option` derive `#[must_use]`
 - Start Date: 2026-01-07
-- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- RFC PR: [rust-lang/rfcs#3906](https://github.com/rust-lang/rfcs/pull/3906)
 - Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
 
 # Summary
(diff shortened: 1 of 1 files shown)


burdges:
Is there some reason #[must_use] should not be a marker trait?

Lokathor:
I would say that actually must_use is more a property of a particular function, not really of types, and sticking it on types is a shorthanded way to put it on a lot of functions "for free" cognitively. I'd say it's best if we, as an ecosystem, stick to having must_use on functions/methods rather than types.

dlight:
> I would say that actually must_use is more a property of a particular function, not really of types, and sticking it on types is a shorthanded way to put it on a lot of functions "for free" cognitively. I'd say it's best if we, as an ecosystem, stick to having must_use on functions/methods rather than types.

So #[must_use] could be a marker trait on the function's type where it is applied (each function has a unique type) *and* on each type it is applied. Then you add a blanked impl on functions whenever their output type impls the trait.

(it could be two traits, one for types you apply #[must_use] and another for functions, so that returning the function doesn't trigger it)

dhardy:
> I would say that actually must_use is more a property of a particular function, not really of types

Yes, but `#[must_use]` on types already has stable support (since 1.0 I think).

> So #[must_use] could be a marker trait

Perhaps, but that would be quite a significant revision. Definitely out-of-scope for this RFC.

dlight:
Agreed! For this RFC a trait is not necessary. If extending this feature to custom types is desired, a trait would be useful

kennytm:
Note that it is already possible to mark a trait as `#[must_use]` (which is how rustc can warn about [unused `Future`](https://doc.rust-lang.org/1.92.0/src/core/future/future.rs.html#30)). It does not help this RFC though.

text/0000-derive-must-use.md
```diff
@@ -0,0 +1,66 @@
...
+    // Do some thing which requires a redraw...
+    Some(Redraw)
+}
+```
+then `do_thing` should be `#[must_use]`, and while we can apply the `#[must_use]` attribute to the function `do_thing`, we shouldn't have to (remember to do so).
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `Option` and `Box` types will "magically" have the `#[must_use]` attribute if and only if their generic parameter `T` does.
```
soqb, line 29: why "magically"?

why not permit `#[must_use]` on the parameters themselves, e.g.

```rust
pub struct Option<#[must_use] T> {
    Some(T),
    None,
}
```

"desugars"[^trait] to

```rust
impl<T: MustUse> MustUse for Option<T> {}
```

[^trait]: There will not necessarily be an actual `MustUse` trait in the implementation.

dhardy: > impl<T: MustUse> MustUse for Option<T> {}

`#[must_use]` is not currently a trait. If you want to make an RFC for that, please go ahead. It may be a good idea, but it is an idea for another RFC, not this one.

Since `#[must_use]` is not currently a trait, we lack the language to describe derive behaviour, hence use of "magically" here.

> `Option<#[must_use] T>`

Use of `#[must_use]` on parameters is novel syntax that I don't particularly like.

dhardy:
Since there have been a number of suggestions about how to generalise this RFC, let me be clear: *this RFC is extremely specific on purpose*:

- This RFC affects only two `std` types in a way which should be compatible with any more general "must_use derive" mechanism
- It addresses what I see as the largest short-comings of `#[must_use]` without the complexity of a general "attribute derive mechanism" or turning `#[must_use]` into a trait

tmccombs:
It seems strange to me for it to magically apply to these two types, without a more general mechanism. 

And why not Result? Or Pin?

text/0000-derive-must-use.md
soqb: sorry, i thought i was clear in the original comment that "the `MustUse` trait" was mentioned purely as an analogy, and not as a normative suggestion, but perhaps not. fwiw i think a `MustUse` trait would be bad since
1. trait solver is slow and heavy compared to just looking at some simple hard-coded rules based on  named types/traits.
2. it gives perhaps too much flexibility to the user. do we really want people to be able to write `impl<T: Copy> MustUse for Thing<T>`, for instance.

i think users _should_ have access to more fine-grained `#[must_use]` annotations, but that doesn't per-se have to be part of this change since `Option` and `Box` alone _would_ be clear usability improvements in my book.

i don't want to overlitigate, but i still reckon this RFC doesn't really justify the "magic" design as written: it's clearly conceptually easier to special case these two generic types, but i don't see it as better from a maintenance perspective if we do want to rip it out for something holistic. for instance `is_ty_must_use` is already [more complicated than ideally it should be](https://github.com/rust-lang/rust/blob/9f6cd6defbd7ef13f6777aa8e43b14d69f0a830a/compiler/rustc_lint/src/unused.rs#L308-L323) (sidebar: and, seemingly, incomplete; see [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e283d114d7b1b6eca94b81425be46f34)). should we keep adding additional branches ad infinitum?

kennytm: > Use of `#[must_use]` on parameters is novel syntax that I don't particularly like.

Attribute on type parameter is already in use since #3621 ([ref](https://doc.rust-lang.org/std/marker/derive.CoercePointee.html#multiple-type-parameters)), it is no longer a novel syntax.

```rust
#[derive(CoercePointee)]
#[repr(transparent)]
struct MySmartPointer<#[pointee] T: ?Sized, U> {
    ptr: Box<T>,
    _phantom: PhantomData<U>,
}
```

dhardy: > `Option<#[must_use] T>`

Something like this would be clearer (though I do find `derive`'s "magic" handling of generic parameters unfortunate relative to making this explicit, as with [autoimpl](https://crates.io/crates/impl-tools)):
```rust
#[derive(must_use)]
enum Option<T> { /* ... */ }
```

That may be besides the point though if this RFC adds more complexity than value. (I had no idea that `Result<T, Infallible>` would not trigger a "must use" warning.)

burdges:
Appears `#[must_use]` could be a trait based upon the above comments.  I think here the reason is just simplicity: 

`Result` is `#[must_use]` now.  The issue here is `Option<Result<..>>` and `Box<Result<..>>` are not `#[must_use]` now, so you must manually put a `#[must_use]` on `fn`s returning them.  

Afaik, no pointer types are `#[must_use]` so `Pin` does not come up, but sure `Pin<Box<Result<..>>>` becomes a question if `Box<Result<..>>` becomes `#[must_use]`.

text/0000-derive-must-use.md
```diff
@@ -0,0 +1,66 @@
+- Feature Name: Let `Option` derive `#[must_use]`
+- Start Date: 2026-01-07
+- RFC PR: [rust-lang/rfcs#3906](https://github.com/rust-lang/rfcs/pull/3906)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Let `Option` and `Box` derive `#[must_use]` from their generic parameter `T`.
```
tgross35, line 9: 2cents: I think "inherit" may be a better term ot use than "derive". For me at least, in reading this RFC title I expected something like `#[derive(MustUse)]` to be involved.

lebensterben: `derive` may cause confusion with derive macro.

but `inherit` may also cause confusion with that in OOP sense.

I think `propagate` is a better word.

PR #3909, opened 2026-01-20
`git mv text/0000-export-visibility.md text/3834-export-visibility.md`

anforowicz:
https://github.com/rust-lang/rfcs/pull/3834 got initially merged while still being contained in `text/0000-export-visibility.md`.  This PR renames the RFC so that it has a correct/actual number.

/cc @tmandry 

[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/3834-export-visibility.md)


anforowicz:
AFAICT 3874 is free - the last biggest number is 3873:

```
$ ls text | sort --numeric-sort | tail
3772-build-target-edition.md
3791-crate-attr.md
3809-derive-from.md
3817-promote-aarch64-pc-windows-msvc-to-tier-1.md
3841-demote-x86_64-apple-darwin.md
3848-asm-const-ptr.md
3849-Project-Goals-2025h2.md
3872-crates-io-security.md
3873-build-std-context.md
3874-export-visibility.md
```

zachs18:
IIUC, the RFC number should be the PR number that added the RFC, so would be `3834-...` here.

In particular, it should not be named `3874-...`, as RFC 3874 https://github.com/rust-lang/rfcs/pull/3874 is something else (not yet merged, but in FCP with disposition merge).

anforowicz:
> IIUC, the RFC number should be the PR number that added the RFC, so would be `3834-...` here.

Ooops.  Fixed.  Thanks for helping me figure out the logistical process here.

PR #3927, opened 2026-03-12
Crate deletion allowances

clarfonthey:
This RFC proposes adding a number of "reasonable changes" to the current crates.io deletion policy that will allow the deletion of several "reasonable to delete" crates which currently do not qualify.

They are definitely, directly, shamelessly motivated by the personal experience of the RFC author. Please let them know if anything was poorly or hastily worded.

> [!IMPORTANT]  
>

[Rendered](https://github.com/clarfonthey/rust-rfcs/blob/deletion-allowances/text/0000-crate-deletion-allowances.md)

### DIFF
--- a/text/0000-crate-deletion-allowances.md
+++ b/text/0000-crate-deletion-allowances.md
@@ -35,7 +35,7 @@ The additions from this RFC are directly inspired by these two examples, althoug
 
 The crate deletion criteria receives the following change, specifically regarding reverse dependencies which are considered by the deletion criteria:
 
-* Yanked versions which are over a year old are not considered as valid reverse dependencies. If the depended crate passes the deletion criteria and gets deleted, these yanked versions may also get deleted.
+* Yanked versions which are over a year old are not considered as valid reverse dependencies. If the depended crate passes the deletion criteria and gets deleted, this still won't delete those yanked versions; they'll just have an invalid dependency.
 * If a version of a crate has been failing to compile on the latest stable for over 4 years (~1 edition cycle), it may be yanked without the approval of the crate author. Since these versions are necessarily at least four years old, they also satisfy the requirements for the above criteria.
 
 The full deletion criteria becomes:


text/0000-crate-deletion-allowances.md
```diff
@@ -0,0 +1,90 @@
...
+* Yanked versions which are over a year old are not considered as valid reverse dependencies. If the depended crate passes the deletion criteria and gets deleted, these yanked versions may also get deleted.
+* If a version of a crate has been failing to compile on the latest stable for over 4 years (~1 edition cycle), it may be yanked without the approval of the crate author. Since these versions are necessarily at least four years old, they also satisfy the requirements for the above criteria.
+
+The full deletion criteria becomes:
+
+* The crate has been published for less than 72 hours,
+* or if all the following conditions are met:
+    * The crate has a single owner,
+    * The crate has been downloaded less than 100 times for each month it has been published,
+    * The crate is not depended upon by any other crate on crates.io (i.e. it has no reverse dependencies), excluding yanked versions of crates that were published over a year ago
```
Diggsey: With the proposed changes it becomes important to specify whether this is a direct or a transitive dependency.

(eg. consider [regular crate] -> [yanked crate] -> [crate to be deleted])

kennytm: this sentence reads like the "yanked versions of [this] crates ..." are *excluded* from being considered for deletion 🤔

clarfonthey: I decided to just elaborate this section more fully to make it clearer; what do you think now?

clarfonthey: Yeah, trying to fit this in one line was a bad idea. Hopefully it seems good now

text/0000-crate-deletion-allowances.md
```diff
@@ -0,0 +1,88 @@
...
+* [value](https://crates.io/crates/value), which has a single reverse-dependency through a yanked version that almost certainly used it due to a typo
+
+The additions from this RFC are directly inspired by these two examples, although they are also designed to apply as broadly as possible.
+
+## Guide-level explanation (Proposal)
+[guide-level-explanation]: #guide-level-explanation
+
+The crate deletion criteria receives the following change, specifically regarding reverse dependencies which are considered by the deletion criteria:
+
+* Yanked versions which are over a year old are not considered as valid reverse dependencies. If the depended crate passes the deletion criteria and gets deleted, these yanked versions may also get deleted.
```
joshtriplett:
```suggestion
* Yanked versions which are over a year old might not considered as valid reverse dependencies. This evaluation may take into account whether the dependency appears to be accidental/unused.
```

We should not be *deleting* old yanked versions of established crates; unless there's some malice involved, they should stick around for archival purposes. Also, consider that a crate might have a feature-flagged dependency using a feature that isn't intended for most people to use. I realize that "may also get deleted" is broad and subject to discretion, but I don't think we should leave that possibility open.

We could, potentially, say something like "If the depended crate passes the deletion criteria and gets deleted, that would be taken into account for any subsequent evaluation of the depending crate or its yanked versions."

text/0000-crate-deletion-allowances.md
```diff
@@ -0,0 +1,88 @@
...
+* If a version of a crate has been failing to compile on the latest stable for over 4 years (~1 edition cycle), it may be yanked without the approval of the crate author. Since these versions are necessarily at least four years old, they also satisfy the requirements for the above criteria.
```
joshtriplett, line 39: As written, this criteria is much too broad. For instance, it's legitimate for a crate to only work on nightly, not on stable.

It seems like this is written in order to enable the previous requirement. But we shouldn't be force-yanking a crate version. We *might*, *potentially* want to *ignore* it for the purposes of evaluating the deletion criteria.

```suggestion
* If a version of a crate has been failing to compile on the latest stable for over 4 years (more than a full edition cycle), it may be ignored for the purposes of evaluating reverse dependencies. This evaluation may take into account whether the dependency appears accidental/unused.
```

text/0000-crate-deletion-allowances.md
```diff
@@ -0,0 +1,88 @@
...
+
+The full deletion criteria becomes:
+
+* The crate has been published for less than 72 hours,
+* or if all the following conditions are met:
+    * The crate has a single owner,
+    * The crate has been downloaded less than 100 times for each month it has been published,
+    * The crate is not depended upon by any other crate on crates.io (i.e. it has no reverse dependencies):
+        * The only crate versions that directly depend on the to-be-deleted crate (including dev and build dependencies) were published at least a year ago and have been yanked for at least a month
+        * Versions of crates which have failed to compile on stable Rust for approximately four years may be yanked without their authors' approval to help satisfy this requirement
```
joshtriplett, line 49: (These would need updating per the suggestions above.)

text/0000-crate-deletion-allowances.md
clarfonthey: The thought process here is to consider the case where *only* yanked versions contain the dependency: even just one non-yanked version having the dependency overrides this requirement and means that both versions remain.

That said, you're right that deleting the versions is probably excessive; we already have crates that might not compile on crates.io and that's fine, so, there's no harm in keeping yanked versions up. (I will edit that probably later today.)

text/0000-crate-deletion-allowances.md
clarfonthey: The idea behind yanking is that ultimately, the version does not compile, and so it's definitely doing something weird and obscure that probably doesn't apply any more. Given Rust's strong stability guarantees, this usually only happens for inference changes that have known-minimal breakages (meaning this was already evaluated to be within those known, minimal breakages) or explicit soundness issues, which again, would have already been evaluated.

And, we also can still keep yanked versions around, just, they've been signalled to not be recommended.

text/0000-crate-deletion-allowances.md
clarfonthey: (Have since updated the text to remove that bit.)

text/0000-crate-deletion-allowances.md
oskgo: I think yanking versions that are broken on the default toolchain (latest stable) goes beyond what yanking has traditionally been for. There are legitimate reasons for a crate version to have a different toolchain in mind, such as latest nightly or fixed version/nightly.

Maybe it's relying on a feature that's on the path to stabilisation. This is the case with the Rust for Linux `kernel ` crate for example, though that one isn't published to crates.io. Old versions with `#![feature(...)]` are going to fail to compile on stable, even when those features have been stabilised.

There are also crates that pin themselves to a certain version/nightly that's moved intermittently, usually because they rely on implementation details. The pinned version won't be the latest. I think this used to be (is?) the case for some crates in the formal verification ecosystem, but I don't recall the details. Force yanking old versions here might make it more difficult to verify old artifacts for projects that don't put Cargo.lock in version control.

text/0000-crate-deletion-allowances.md
```diff
@@ -0,0 +1,88 @@
...
+* [value](https://crates.io/crates/value), which has a single reverse-dependency through a yanked version that almost certainly used it due to a typo
+
+The additions from this RFC are directly inspired by these two examples, although they are also designed to apply as broadly as possible.
+
+## Guide-level explanation (Proposal)
+[guide-level-explanation]: #guide-level-explanation
+
+The crate deletion criteria receives the following change, specifically regarding reverse dependencies which are considered by the deletion criteria:
+
+* Yanked versions which are over a year old are not considered as valid reverse dependencies. If the depended crate passes the deletion criteria and gets deleted, this still won't delete those yanked versions; they'll just have an invalid dependency.
```
fintelia, line 38: What happens if the crate author tries to un-yank the version? Are they blocked from doing so?

clarfonthey: I had no idea that unyanking was even supported by Cargo, but yeah, I guess that `cargo yank --undo` exists, now that I'm looking at it.

I actually have no idea about the specifics of this, and whether it's time-gated or not, but it appears that this is a valid action, so, will have to look more into it.

text/0000-crate-deletion-allowances.md
clarfonthey: You are right and I guess that we would have to pave a path specifically for allowing the current nightly if it turns out that a feature genuinely exists in the compiler for a full edition cycle, since it would still compile. I'll think a bit on how to word that.

text/0000-crate-deletion-allowances.md
```diff
@@ -0,0 +1,88 @@
...
+
+See the [guide-level explanation](#guide-level-explanation) for the full list.
+
+## Motivation
+[motivation]: #motivation
+
+[RFC 3660](https://rust-lang.github.io/rfcs/3660-crates-io-crate-deletions.html) currently details the requirements for deleting a crate which seem very reasonable:
+
+* The crate has been published for less than 72 hours,
+* or if all the following conditions are met:
```
tarcieri, line 19: Isn't it problematic to delete any crate with dependent crates, regardless of its age?

clarfonthey: Did you read the rest of the RFC?

PR #3938, opened 2026-03-28
RFC: Associated traits

sandersaares:
Allow traits to declare **associated traits** — named trait constraints that are defined abstractly in a trait and given concrete values in implementations. Just as associated types let a trait abstract over *which type* is used, associated traits let a trait abstract over *which constraints* are imposed. This is the trait-level analog of associated types.

```rust
#![feature(associated_traits)]

// 1. Declare an associated trait in a trait definition.
trait Container {
    trait Elem;
    fn process<T: Self::Elem>(&self, item: T);
}

// 2. Each impl chooses the concrete constraints.
impl Container for MyContainer {
    trait Elem = Send + Clone;
    fn process<T: Self::Elem>(&self, item: T) { /* ... */ }
}

// 3. Generic code outside the impl uses `C::Elem` as a bound.
//    This is the primary use case: the caller is generic over
//    the constraints without knowing what they are.
fn process_item<C: Container, T: C::Elem>(container: &C, item: T) {
    container.process(item);
}

// 4. Fully-qualified syntax also works.
fn process_qualified<C: Container, T: <C as Container>::Elem>(container: &C, item: T) {
    container.process(item);
}
```

Previous discussion in #2190.

Prototype implementation: https://github.com/sandersaares/rust/tree/associated-traits

[Rendered](https://github.com/rust-lang/rfcs/blob/eda316fd2f04ad50873b2e047235e3e54866064f/text/0000-associated-traits.md)

> [!IMPORTANT]  
>

### DIFF
--- a/text/0000-associated-traits.md
+++ b/text/0000-associated-traits.md
@@ -1,6 +1,6 @@
 - Feature Name: `associated_traits`
-- Start Date: 2026-03-21
-- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Start Date: 2026-03-28
+- RFC PR: [rust-lang/rfcs#3938](https://github.com/rust-lang/rfcs/pull/3938)
 - Rust Issue: [rust-lang/rfcs#2190](https://github.com/rust-lang/rfcs/issues/2190)
 
 ## Summary
--- a/text/0000-associated-traits.md
+++ b/text/0000-associated-traits.md
@@ -1,0 +1,489 @@
+- Feature Name: `associated_traits`
+- Start Date: 2026-03-21
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rfcs#2190](https://github.com/rust-lang/rfcs/issues/2190)
+
+## Summary
+[summary]: #summary
+
+Allow traits to declare **associated traits** — named trait constraints that are defined abstractly in a trait and given concrete values in implementations. Just as associated types let a trait abstract over *which type* is used, associated traits let a trait abstract over *which constraints* are imposed. This is the trait-level analog of associated types and the Rust equivalent of Haskell's `ConstraintKinds`.
+
+```rust
+trait Container {
+    trait Elem;
+    fn process<T: Self::Elem>(&self, item: T);
+}
+
+impl Container for MyContainer {
+    trait Elem = Send + Clone;
+    fn process<T: Self::Elem>(&self, item: T) { /* ... */ }
+}
+```
+
+## Motivation
+[motivation]: #motivation
+
+### The problem
+
+Rust programmers frequently need to write traits that are *generic over constraints*. Today, the only way to parameterize a trait over which bounds should apply to some type is to fix them at the trait definition site, or to duplicate the entire trait hierarchy for each set of constraints.
+
+Consider a plugin framework:
+
+```rust
+// Without associated traits — constraints are baked in.
+trait Plugin {
+    fn run<T: Send + 'static>(&self, task: T);
+}
+```
+
+Every implementor must accept `Send + 'static`, even if some plugin systems (e.g., single-threaded ones) don't need `Send`. The only workaround today is to duplicate the trait:
+
+```rust
+trait SendPlugin {
+    fn run<T: Send + 'static>(&self, task: T);
+}
+
+trait LocalPlugin {
+    fn run<T: 'static>(&self, task: T);
+}
+```
+
+This duplication cascades through every consumer of the trait, every generic function, and every downstream crate.
+
+### Use cases from the community
+
+[rust-lang/rfcs#2190](https://github.com/rust-lang/rfcs/issues/2190) (76 👍, 12 👀) has collected real-world use cases since 2017. The most prominent:
+
+**Async runtime agnosticism.** A runtime trait can abstract over whether futures must be `Send`:
+
+```rust
+trait Runtime {
+    trait FutureConstraint;
+    fn spawn<F: Future<Output = ()> + Self::FutureConstraint>(f: F);
+}
+
+impl Runtime for TokioRuntime {
+    trait FutureConstraint = Send + 'static;
+    // ...
+}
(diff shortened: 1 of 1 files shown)


text/0000-associated-traits.md
```diff
@@ -0,0 +1,618 @@
...
+}
+```
+
+**Declaration bounds** (the part after `:`) are *requirements* on every impl's value. If `trait Elem: Clone;`, then every impl must provide a value that implies `Clone`. These bounds are also available to callers: a function with `T: C::Elem` may assume `T: Clone` when the declaration includes `: Clone`.
+
+**Defaults** (the part after `=`) provide a value that impls may omit, analogous to default associated types. If an impl omits the associated trait, the default is used.
+
+### Associated trait implementations
+
+In a trait impl, the associated trait's value is provided with `trait Name = Bounds;`:
```
Nashenas88, line 357: I'm curious whether this intends to support "no additional bounds". For example https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c5774b6a2c62aa4352b5f87ac941f3d0 where the two functions show you can have a where clause with an empty bound. See the where clause section in the reference https://doc.rust-lang.org/reference/items/generics.html#grammar-WhereClause . The syntax you supplied does not consider the Bound optional. Maybe it's helpful to add a dedicated grammar section here too?

It seems less useful to me in the example I provided, but I can imagine wanting to write a test-only trait impl without any additional bounds on the associated trait.

text/0000-associated-traits.md
```diff
@@ -0,0 +1,618 @@
...
+```rust
+fn print_element<C: Container, T: C::Elem>(x: T)
+where
+    C::Elem: Debug,  // The impl's value for Elem must include Debug
+{
+    println!("{:?}", x);
+}
+```
+
+This is different from `T: C::Elem + Debug` (which constrains `T` independently). The value constraint `C::Elem: Debug` constrains the *Container's impl* — if the impl provides `trait Elem = Send` (no Debug), the call is rejected even if `T` happens to implement Debug.
```
theemathas, line 281: It's not clear to me what `C::Elem: Debug` means. Here are some possible interpretations:

* The list of traits in the `trait Elem = ....` declaration in the impl, must literally contain `Debug` in the list.
* The list of traits (same as above), must contain `Debug`, or contain a subtrait of `Debug`.
* If the impl has a `trait Elem = ....`. Then, the compiler checks whether `X: ....` would imply `X: Debug` for all types `X`. This might hold if, say, we have a blanket `impl<X: ....> Debug for X`.

theemathas: Also, it's not clear to me why we would want only this direction of constraints. For example, why wouldn't we want to write something like `where Trait: C::Elem`? Such a bound would, for example, maybe allow something like:

```rust
trait Consumer {
    trait Requirements;
    fn consume<T: Self::Requirements>(&self, value: T);
}

fn consume_debug<C: Consumer>(c: C)
where
    Debug: C::Requirements
{
    // c can consume anything that implements Debug
    c.consume(1);
    c.consume("abc");
}
```

text/0000-associated-traits.md
```diff
@@ -0,0 +1,618 @@
...
+
+### Where associated traits *cannot* appear
+
+- **Type position**: `let x: T::Elem = ...` is an error — associated traits are constraints, not types.
+- **`dyn` position**: `dyn T::Elem` is an error. Rust type-checks generic function bodies once, before monomorphization, so the set of traits behind `T::Elem` is not yet known. Since `dyn` types require a compile-time-known vtable layout, `dyn` with an associated trait that depends on a type parameter is fundamentally incompatible with Rust's generics model. This is the same reason `dyn T` where `T` is a type parameter is rejected.
```
theemathas, line 290: Should something like `dyn Container<Elem = Debug>` be allowed?

text/0000-associated-traits.md
sandersaares: Oh, neat! I did not realize this possibility existed. Sure, supporting an empty bound sounds reasonable.

text/0000-associated-traits.md
```diff
@@ -0,0 +1,618 @@
...
+
+```rust
+trait SendPlugin {
+    fn run<T: Send + 'static>(&self, task: T);
+}
+
+trait LocalPlugin {
+    fn run<T: 'static>(&self, task: T);
+}
+```
```
Jules-Bertholet, line 65: To what degree can HRTB/[`non_lifetime_binders`](https://github.com/rust-lang/rust/issues/108185) replace this feature? E.g.

```rust
trait Plugin<T> {
    fn run(&self, task: T);
}

trait SendPlugin = for<T: Send + 'static> Plugin<T>;
trait LocalPlugin = for<T: 'static> Plugin<T>;
```

kennytm: I think this is similar to how HRTB can simulate #1598 (GATs). *"... a hacky attempt to work around the limitation"*.

text/0000-associated-traits.md
sandersaares: Incorporated empty bound possibility into RFC text, supporting both `trait Item;` and `trait Item = ;` forms (equivalent).

text/0000-associated-traits.md
sandersaares: Clarified RFC text that the second interpretation is the intended one (i.e. must be a declared transitive supertrait). The third one with support for blanket impls is an interesting option but feels complex, so to keep the RFC simple I just listed it in future possibilities section.

The question about direction of constraints needs more analysis. I make no comment on that at the moment.

nikomatsakis:
Hi all! We reviewed this in our lang-team meeting. We all agreed that **eventually** we would like to do something like this, but that neither we nor the @rust-lang/types team is really ready to consider it right now. (Eventually, I'd like to extend our set of "kinds" to include "bounds" so that one could do all kinds of interesting things...).

If you want to pursue fundamental types systems work of this kind, I would encourage you to talk to some of us on the types team about doing work around the new trait solver or other similar projects, as we need that sort of foundational support before we can take on projects like this. (Feel free to ping me on Zulip to discuss!)

Thanks very much for the RFC. Interesting stuff.

PR #3953, opened 2026-04-21
Support heterogeneous try blocks (`try_blocks_heterogeneous`) RFC

MusicalNinjaDad:
[RFC 3721](https://github.com/rust-lang/rfcs/pull/3721) implemented default support for homogeneous `try {...}` blocks, where all `?`s return the same error type. This RFC aims to provide support for explicit annotation of the returned error type from a `try {...}` block.

#### Note to reviewers
This RFC provides post-hoc documentation of the current status-quo to act as an anchor point for moving forwards and a consolidation point for historical comments across multiple tickets. The changes described are _already implemented in nightly as described_.

I appreciate that particularly the current syntax is need of bikeshedding 🚳 - I propose that the RFC is first merged to reflect the current implementation, effectively documentation to the state that matches reality. Then I'd be happy to lead, or hand over, driving the discussion on refinements to syntax etc. Having something implemented that should have had an RFC first, but didn't, leaves a gap which makes moving forward difficult and puts barriers in place for anyone trying to experiment with the feature & give feedback.

This is created at the request of the implementor (@scottmcm) based on a [brief zulip chat](https://rust-lang.zulipchat.com/#narrow/channel/122652-new-members/topic/Mike.3A.20I.27d.20like.20to.20help.20stabilising.20features.20I.20love.20to.20use/near/584765588) and written in a way to mimic those RFCs that I, as a nightly user, find valuable.

> [!IMPORTANT]
>

r? @scottmcm

#### AI Statement

- The text is all my own work, hand-written, by a human, no AI.
- The research and analysis is all my own work, no AI assistance.
- The only AI assistance came in the form of an early review for typos but that was so unreliable I switched it off too

[Rendered](https://github.com/MusicalNinjaDad/rust-rfcs/blob/first_draft/text/0000-try-blocks-heterogeneous.md)

### DIFF
--- a/text/0000-try-blocks-heterogeneous.md
+++ b/text/0000-try-blocks-heterogeneous.md
@@ -134,16 +134,84 @@ fn heterogeneous_into_anyhow() {
 ## Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-Explain the proposal as if it was already included in the language and you were teaching it to another Rust programmer. That generally means:
+_Assuming the explanation for try blocks is implemented as per RFC 3712_
 
-- Introducing new named concepts.
-- Explaining the feature largely in terms of examples.
-- Explaining how Rust programmers should _think_ about the feature, and how it should impact the way they use Rust. It should explain the impact as concretely as possible.
-- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
-- If applicable, describe the differences between teaching this to existing Rust programmers and new Rust programmers.
-- Discuss how this impacts the ability to read, understand, and maintain Rust code. Code is read and modified far more often than written; will the proposed feature make code easier to maintain?
+> This behaviour is what you want in the vast majority of simple cases.  In particular,
+> it always works for things with just one `?`, so simple things like `try { a? + 1 }`
+> will do the right thing with minimal syntactic overhead.  It's also common to want
+> to group a bunch of things with the same error type.  Perhaps it's a bunch of calls
+> to one library, which all use that library's error type.  Or you want to do
+> [a bunch of `io` operations](<https://github.com/rust-lang/rust/blob/d6f3a4ecb48ead838638e902f2fa4e5f3059779b/>> > compiler/rustc_borrowck/src/nll.rs#L355-L367) which all use `io::Result`.  Additionally, `try` blocks work with
+> `?`-on-`Option` as well, where error-conversion is never needed, since there is only `None`.
+>
+> It will fail to compile, however, if not everything shares the same error type.
+> Suppose we add some formatting operation to the previous example:
+>
+> ```rust,edition2021,compile_fail
+> let pair_result = try {
+>     let a = std::fs::read_to_string("hello")?;
+>     let b = std::fs::read_to_string("world")?;
+>     let c: i32 = b.parse()?;
+>     (a, c)
+> };
+> ```
+>
+> The compiler won't let us do that:
+>
+> ```text
+> error[E0308]: mismatched types
+>   --> src/lib.rs:14:32
+>    |
+>    |     let c: i32 = b.parse()?;
+>    |                           ^ expected struct `std::io::Error`, found struct `ParseIntError`
+>    = note: expected enum `Result<_, std::io::Error>`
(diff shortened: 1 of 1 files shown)


text/0000-try-blocks-heterogeneous.md
```diff
@@ -0,0 +1,536 @@
...
+with associated adjustments to `visit::walk_fn`, `assert!`
+
+#### Parse & pretty-print syntax `try bikeshed T {...}`
+
+1. Add (temporary) `bikeshed` keyword (see [unresolved-questions]) to available Tokens
+
+    ```rust
+    enum TokenType {
+    ...
+        SymBikeshed,
```
kennytm, line 245: ???

we are obviously not going to use `try bikeshed Type {}` as the final syntax. 

also the rustc implementation is not interesting to anyone — it does not enrich the spec which the general users read, while compiler-team members obviously knows how to connect the dots.

plus, the `try bikeshed Type {}` placeholder syntax [already works today](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&code=%23%21%5Bfeature%28try_blocks_heterogeneous%29%5D%0Afn+main%28%29+%7B%0A++++let+f+%3D+try+bikeshed+Option%3Cu32%3E+%7B%0A++++++++5%0A++++%7D%3B%0A%7D) so IMO this section is totally useless.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3953/f63771800b506c3874b760c28acc8fbeb32f7824..80ad9a31101ea47b2ee02dce24195bf19b7b9f21)*

text/0000-try-blocks-heterogeneous.md
```diff
@@ -0,0 +1,536 @@
...
+This is also a good place to "dump ideas", if they are out of scope for the
+RFC you are writing but otherwise related.
+
+If you have tried and cannot think of any future possibilities,
+you may simply state that you cannot think of anything.
+
+Note that having something written down in the future-possibilities section
+is not a reason to accept the current or a future RFC; such notes should be
+in the section on motivation or rationale in this or subsequent RFCs.
+The section merely provides additional information.
```
kennytm: the placeholder text is still here 😅 

```suggestion
```



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3953/f63771800b506c3874b760c28acc8fbeb32f7824..80ad9a31101ea47b2ee02dce24195bf19b7b9f21)*

text/0000-try-blocks-heterogeneous.md
MusicalNinjaDad: Background: I put this RFC together on request from scottmcm (on zulip, when I offered to help out with stabilisation in Try trait etc.). It's my first one - so will have missed reaching the right target audience a few times, I'm sure

kennytm: maybe you should first send a Pre-RFC to internals.rust-lang.org. 

we don't know what is being discussed in Zulip (you can add a link actually), but this RFC did not give satisfying responses to the unresolved questions raised in rust-lang/rust#149488 (syntax, shape of annotated type), so I think it's too early to actually request-for-comments.

lyphyser:
Why not instead just fix type inference so that `let x: Result<_, B>  = try {}` works, as well as variants like returning the try value, etc.?

MusicalNinjaDad:
> Why not instead just fix type inference so that `let x: Result<_, B> = try {}` works, as well as variants like returning the try value, etc.?

Does the explanation added in [5101305](https://github.com/rust-lang/rfcs/pull/3953/commits/51013059bc6fa8126f4c858491156ebd518bd42c) help @lyphyser ?

MusicalNinjaDad:
Hi @scottmcm - is this anywhere near your expectations? Would you like me to re-roll the reviewer, so this gets "taken off your plate"?

MusicalNinjaDad:
r? lang

clarfonthey:
Note that RFCs are not reviewed by single people like ordinary PRs; they're reviewed by the entire team, so, you don't want to do `r?` for these.

PR #3961, opened 2026-05-06
Fix 3931 rfmf number and name

ehuss:
This fixes the 3931 filename and PR number.

[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/3931-rfmf-rust-foundation-maintainer-fund.md)


ehuss:
cc @oli-obk Just FYI, there's usually several steps that need to be taken before merging an RFC ([docs](https://forge.rust-lang.org/lang/rfc-merge-procedure.html)).

PR #3963, opened 2026-05-26
Merge RFC Propose the concept of a crates.io username for identity

carols10cents:
This merges https://github.com/rust-lang/rfcs/pull/3946 that has completed FCP. 🎉 

[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/3946-crates-io-username-identity.md)


programmerjake:
usually RFCs are merged by adding commits in the original PR for moving the file to the correct location and adding the RFC number and tracking issue, then merging the original PR, rather than creating a new PR.

carols10cents:
> usually RFCs are merged by adding commits in the original PR for moving the file to the correct location and adding the RFC number and tracking issue, then merging the original PR, rather than creating a new PR.

yes, i agree what i did was weird. i was following the forge instructions though, which i suspect were outdated/confusingly worded, so i've [sent in a PR to update](https://github.com/rust-lang/rust-forge/pull/1056).

PR #3965, opened 2026-05-27
`--allow-unstable-flags`: Allow unstable flags on stable

jyn514:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3965)*

This is a follow-up to https://github.com/rust-lang/rfcs/pull/3882 which is scoped only to CLI flags, and has improved caching mechanisms by going through Cargo.

> [!IMPORTANT]
>

[Rendered](https://github.com/jyn514/rfcs/blob/allow-flags/text/0000-rustc-bootstrap.md)


jyn514:
cc some of the stakeholders:
- @weihanglo (Cargo)
- @celinval (rustc_public)
- @obi1kenobi (cargo-semver-checks)
- @ojeda (rust-for-linux)
- @Veykril (rust-analyzer)
- @vlad20012 (RustRover)
- @calebcartwright (Rustfmt)

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,283 @@
...
+Enabling RUSTC_BOOTSTRAP for one part of the toolchain enables it for *all* parts of the toolchain; in particular:
+- `proc-macro2` uses `cargo:rerun-on-env-changed=RUSTC_BOOTSTRAP`, causing cache thrashing whenever this env var changes.
+- rust-analyzer wants to enable RUSTC_BOOTSTRAP only for cargo and libtest, but the variable enables features for rustc as well.
+  `RUSTFLAGS="-Z allow-features="` fixes this for lang features, but at the price of thrashing the cache; and there is no equivalent way to disable unstable CLI features.
+
+An important design constraint here is that the "end-user" (whoever is running the build) should always have control over which features are enabled.
+To the extent that tools act as a "buffer" between feature breakage and the end-user,
+they should only take responsibility for exactly the features whose breakage they know how to handle.
+
+We mark language features out of scope, because we expect the new flags to reduce most of the use cases for RUSTC_BOOTSTRAP, and therefore to reduce the amount of needless cache thrashing going on.
```
zackw, line 100: Because what I care about is opening up testing and experimentation with unstable features to stakeholders who can't or won't use nightly toolchains, I would like to see this restriction lifted if at all possible.  However, it would be fine with me to _start_ with non-language features, as long as there is a path to inclusion of language features in the future. (see also below)

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,283 @@
...
+We do not want to decouple our stability policy for language features,
+because there's no possibility there of the library acting as a buffer between other projects and breakage[^2].
+
+[^2]: Theoretically libraries can write a build script that does feature detection, but this slows down the build for everyone, and it's very very hard to write that build script properly.
```
zackw, line 104: Feature detection should IMO be built into the language.  Of course that's a big Twinkie all by itself, and I don't want to scope-creep your proposal to death. I bring it up to make clear that the current need for a difficult-to-write build script is itself a limitation that could be lifted, and therefore should not stand in the way of expanding this proposal to cover language features.

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,283 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This encourages using unstable features on stable, explicitly going against our goals as a project.
+  But people are doing that anyway, and keeping the status quo does not help us prevent them, while causing many other issues.
+  - Note that while this could be seen as encouragement at the *policy* level, it's actually more restrictive at the *technical* level,
+    since it requires people to make an exhaustive list of all features they're using.
+- This may make it less likely that people help stabilize features.
+  But stabilizing features is [very very hard](https://medium.com/@ElizAyer/organizational-boundary-problems-too-many-cooks-or-not-enough-kitchens-2ddedc6de26a),
+  and in the meantime people have very little recourse when they need to use an unstable feature.
```
zackw, line 223: I believe that implementation of this proposal will actually _encourage_ people to help stabilize features, by enabling people who can't or won't use nightly toolchains to test and experiment with unstable features.  I think this possibility should be considered as a plus rather than a minus for the proposal.

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,283 @@
...
+## Stability policy
+
+Despite being usable on stable, this is an unstable feature.
+Like any other unstable feature, we reserve the right to change or remove this feature in the future, as well as any other unstable feature that it enables.
+Using this feature is opting out of the normal stability/backwards compatibility guarantee of stable.
+
+Although we do not take technical measures to prevent it from being used, we strongly discourage using this feature.
+If at all possible, please contribute to stabilizing the features you care about instead of bypassing the Rust project's stability policy.
+
+If you do use this to enable an unstable feature, please contact a member of the project who works on the feature in question, so that we know who is exposed to breakage.
```
zackw, line 143: As I have been saying, this mechanism _enables_ people who currently cannot to contribute to the stabilization of features.  Therefore, IMO this strikes a wrong note and should be revised along the lines of "We encourage use of this mechanism, to facilitate testing and experimentation with unstable features.  However, please be aware that by using it you are exposing yourself to potential breakage that your use of the stable toolchain would otherwise insulate you from.  If you use this feature, please reach out to and stay in touch with a member of the project ..."

zackw (commented):
As a user and occasional contributor, I wholeheartedly support relaxation of the stability policy to provide opt-in, per-feature access to unstable features -- _all_ unstable features, not just those that take the form of CLI flags -- on the stable toolchain.

Right now, there are several important groups of stakeholders (enumerated in this very RFC) who _cannot_ test or experiment with unstable features, because they cannot use the nightly toolchain.  There are also an unknown but probably large number of developers who _will not_ risk using the nightly toolchain (and thus exposing themselves to _all_ the potential breakage) _even though_ they would like to experiment with one or two unstable features.  In my opinion, therefore, the stability policy in its present state _actively discourages_ people from testing and experimenting with unstable features, and is a major reason why it is so hard to get enough feedback to complete stabilization.

With that position in mind, a few suggested improvements to the RFC:

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
jyn514: 👍 I think we do have that path, I discuss it a bit in "Drawbacks" under "lang features".

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+still requiring feature gates, but allowing feature gates to be enabled on stable.
+
+It does so in two ways:
+1. Extend `-Z unstable-options` to take a list of option names, rather than being a simple boolean.
+   Then, rename it to `--allow-flags`. `allow-flags` is always available, even on stable.
+2. Add a new `--allow-flags` flag to Cargo, which propagates it to all invoked commands with proper caching.
+3. Change Rustfmt to give a hard error if a feature gate is disabled.
+
+This RFC is *not* intended as a general purpose mechanism for Rust developers to use nightly features on stable;
+it's specifically targeted at build systems wrapping cargo, such as distro packagers, external tools shipped with the toolchain, and large projects that build a custom Rust toolchain from source.
```
ojeda: Does it have a use case for build systems wrapping `rustc` directly? i.e. this sentence says that it is specifically targeted only to those that use `cargo`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+3. Encourage people to help stabilize the features they care about.
+
+There are some cases in which none of those goals are applicable, but we still prevent people from using nightly features.
+This is particularly bad when projects *must* depend on unstable features to ship another feature they care about.
+Some examples:
+- rust-analyzer and RustRover need `./some-libtest-binary --format=json` to determine the list of possible tests to run.
+- rust-analyzer and RustRover need all values in `rustc --print=cfg` to build the standard library.
+  (see [#139892](https://github.com/rust-lang/rust/issues/139892#issuecomment-2808505610) for an explanation of why this is affected by unstable features)
+- `cargo semver-checks` needs `rustdoc --output-format=json` in order to work at all.
+- Rust for Linux (RfL) needs a way to build a custom version of core.
```
ojeda, line 59: Hmm... I may be confused. Building `core` will be eventually handled by `build-std` as far as I know, but even when that happens (and when we do not need any unstable language feature), we will still need to enable certain compiler flags etc., and IIUC this is meant to work for those, no?

I guess this may just be a subset (since it is an example, after all), but perhaps it helps to say we could use it for other things too (perhaps linking to https://github.com/Rust-for-Linux/linux/issues/2).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
+- Feature Name: `allow-flags`
+- Start Date: 2025-05-18
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes decoupling the two components of our stability policy, for CLI flags only:
```
alice-i-cecile, line 9: It would be helpful to readers to clearly articulate what those two components are in this sentence :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
Lokathor: If people can't or won't use Nightly, why would they allow themselves to activate an unstable feature on Stable?

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+still requiring feature gates, but allowing feature gates to be enabled on stable.
+
+It does so in two ways:
+1. Extend `-Z unstable-options` to take a list of option names, rather than being a simple boolean.
+   Then, rename it to `--allow-flags`. `allow-flags` is always available, even on stable.
+2. Add a new `--allow-flags` flag to Cargo, which propagates it to all invoked commands with proper caching.
+3. Change Rustfmt to give a hard error if a feature gate is disabled.
```
alice-i-cecile: This is a third way :) I also think that this is primarily cleanup, and doesn't need to be mentioned in the summary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+
+This RFC is *not* intended as a general purpose mechanism for Rust developers to use nightly features on stable;
+it's specifically targeted at build systems wrapping cargo, such as distro packagers, external tools shipped with the toolchain, and large projects that build a custom Rust toolchain from source.
+As such, it does not attempt to address unstable lang features, which we consider adequately addressed by `RUSTC_BOOTSTRAP=crate_name`.
```
alice-i-cecile:
```suggestion
As such, it does not attempt to address the use of unstable lang features with a stable Rust compiler version, which we consider adequately addressed by `RUSTC_BOOTSTRAP=crate_name`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+## Why allow using unstable features on stable?
+
+Rust's stability policy has two components:
+1. To the extent possible, each unstable feature comes with a feature gate, and is disabled when that feature gate is inactive. [^1]
```
alice-i-cecile:
```suggestion
1. To the extent possible, each unstable feature comes with its own feature gate, and is disabled when that feature gate is inactive. [^1]
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+2. Enabling feature gates is only allowed on the nightly toolchain.
+
+[^1]: There are some exceptions to this, such as https://github.com/rust-lang/rust/issues/139892#issuecomment-2808505610.
+     But in general we attempt to make sure all unstable features have a feature gate.
```
alice-i-cecile, line 38: This line can just be cut IMO. Exception that proves the rule and all that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+- Nightly is not the same as stable-with-`RUSTC_BOOTSTRAP`.
+  In particular, stable contains backports and nightly does not.
+
+## Why this exact mechanism?
+
+Currently, these tools use [`RUSTC_BOOTSTRAP=1`][rustc-bootstrap] as a workaround. But this workaround has many downsides:
+
+[rustc-bootstrap]: https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/rustc-bootstrap.html
+
+Enabling RUSTC_BOOTSTRAP for one part of the toolchain enables it for *all* parts of the toolchain; in particular:
```
kpreid: Both of these paragraphs end with a colon, and the referent of the first one is unclear. Perhaps they should be combined? Or if the “many downsides” are something other than the following bullet list, then it seems unclear what they are.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
Lokathor: They are explained in the sentence, immediately after the `:`, on line 10

text/0000-rustc-bootstrap.md
ojeda: It would be nice to mention right away in the introduction that it takes a list of flags, or rather show the syntax, e.g. `--allow-flags=<flag1>,...,<flagN>` or similar.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The following documentation will live in the [unstable book] (or Cargo's [unstable features][cargo-unstable] section), not in the [rustc book].
+
+[unstable book]: https://doc.rust-lang.org/nightly/unstable-book/
+[cargo-unstable]: https://doc.rust-lang.org/cargo/reference/unstable.html#allow-features
+[rustc book]: https://doc.rust-lang.org/nightly/rustc/
+
+The `--allow-flags` Rustc option allows you to control precisely which unstable options are used by a given crate.
```
kpreid: Speaking only for my own sense of things, `--allow-flags` doesn’t sound scary enough — or even really specific enough. How about `--allow-unstable-flags`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+## Why allow using unstable features on stable?
+
+Rust's stability policy has two components:
+1. To the extent possible, each unstable feature comes with a feature gate, and is disabled when that feature gate is inactive. [^1]
+2. Enabling feature gates is only allowed on the nightly toolchain.
+
+[^1]: There are some exceptions to this, such as https://github.com/rust-lang/rust/issues/139892#issuecomment-2808505610.
+     But in general we attempt to make sure all unstable features have a feature gate.
+
+Our motivation for 1 (having feature gates) is to make sure that people do not unknowingly rely on unstable features.
```
alice-i-cecile, line 40: This is missing some rationale about granularity, but that's probably not super relevant here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+
+Why are these uses ok? Three reasons:
+- Each of these, except for rustc_public, is an external tool, not a library.
+  They do not need unstable language features, only unstable tool output.
+- Each of these tools accept responsibility for breakage.
+  `semver-checks` and RfL both explicitly adapt to each new release of rustc, and their feedback on breakage is very useful for improving the features they use.
+  rust-analyzer and RustRover don't break at all for `--print=cfg`—they're not using it in code, only in the CLI—and adapt to any changes in libtest json format.
+- These tools act as a "buffer" between other projects and breakage.
+  For example, semver-checks hides the breaking changes behind its own interface such that downstream projects are not affected.
+  Similar, RfL backports breakage fixes to stable branches such that old versions of the kernel keep building with new rust toolchains.
```
ojeda:
```suggestion
  Similarly, RfL backports breakage fixes to stable branches such that old versions of the kernel keep building with new rust toolchains.
```

I am not sure if I understand what "buffer" means here, i.e. we do adapt and backport, but it seems this paragraph is about avoiding issues for third-party repositories. I guess it may mean that the mainline branch of Linux is the "buffer" for the stable branches? I guess it is OK, though it is the same project, so it may be a bit confusing.

Nevertheless, it is true that Rust for Linux can act as a "buffer" for others, in the sense that we do build the kernel in upstream Rust CI, and thus some issues may be found there pre-merge, and thus before they are found in other projects. On the other hand, the breakage has been fairly small so far.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+- `rustc_public`'s entire mission is to wrap unstable APIs with stable ones and therefore needs access to all `rustc_private` features.
+
+Why are these uses ok? Three reasons:
+- Each of these, except for rustc_public, is an external tool, not a library.
+  They do not need unstable language features, only unstable tool output.
+- Each of these tools accept responsibility for breakage.
+  `semver-checks` and RfL both explicitly adapt to each new release of rustc, and their feedback on breakage is very useful for improving the features they use.
+  rust-analyzer and RustRover don't break at all for `--print=cfg`—they're not using it in code, only in the CLI—and adapt to any changes in libtest json format.
+- These tools act as a "buffer" between other projects and breakage.
+  For example, semver-checks hides the breaking changes behind its own interface such that downstream projects are not affected.
```
ojeda:
```suggestion
  For example, `semver-checks` hides the breaking changes behind its own interface such that downstream projects are not affected.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
Lokathor: I don't need it to be scary, but being more specific is of great benefit.

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+Currently, these tools use [`RUSTC_BOOTSTRAP=1`][rustc-bootstrap] as a workaround. But this workaround has many downsides:
+
+[rustc-bootstrap]: https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/rustc-bootstrap.html
+
+Enabling RUSTC_BOOTSTRAP for one part of the toolchain enables it for *all* parts of the toolchain; in particular:
+- `proc-macro2` uses `cargo:rerun-on-env-changed=RUSTC_BOOTSTRAP`, causing cache thrashing whenever this env var changes.
+- rust-analyzer wants to enable RUSTC_BOOTSTRAP only for cargo and libtest, but the variable enables features for rustc as well.
+  `RUSTFLAGS="-Z allow-features="` fixes this for lang features, but at the price of thrashing the cache; and there is no equivalent way to disable unstable CLI features.
+
+An important design constraint here is that the "end-user" (whoever is running the build) should always have control over which features are enabled.
```
ojeda: Do you mean "the project" (i.e. not Rust, but some crate or Rust for Linux etc.) instead of "the end-user" here? i.e. I would expect the "end-user" to not control the set of features, precisely so that they are shielded from that by the project's build system. I may be missing what you are trying to say here, though.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+[alona]: https://github.com/aDotInTheVoid/
+[Zulip]: https://rust-lang.zulipchat.com/
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Shared rules
+
+`allow-flags` is a comma-separated list of CLI flag names, with the leading `-Z` or `--` (if any) removed.
+Flag values are not supported, only names.
```
ojeda: Is the reason to not pass the prefixes like `-Z` that one does not need to change it later if it gets stabilized? It could be nice to explain that here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+Stable/beta Rustfmt now errors instead of warning when an unstable option is set without also setting `--allow-flags`.
+
+libtest runners accept `--allow-flags`.
+
+Each non-Cargo flag takes one of the following strings as a value:
+- An empty string, which indicates that no flags are allowed (default on stable)
+- A comma-separated list of flag names.
+
+If `allow-flags` is passed multiple times, the *intersection* of all values is used, not the union.
+This matters in cases where two parties don't trust each other, such as running `cargo build --allow-flags=rustc=x` in a workspace with `build.rustflags=--allow-flags=x,y`: this should be equivalent to `rustc --allow-flags=x`.
```
ojeda: So the `cargo` level one is intended to act as a "limit"? I guess this related to my other comment, i.e. when the "end-user" needs to control this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+We suggest, but do not require, that the flag is made part of the fingerprint tracking, not unit cache tracking,
+so that changing the enabled features overwrites the cache rather than adding to it.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This encourages using unstable features on stable, explicitly going against our goals as a project.
+  But people are doing that anyway, and keeping the status quo does not help us prevent them, while causing many other issues.
+  - Note that while this could be seen as encouragement at the *policy* level, it's actually more restrictive at the *technical* level,
+    since it requires people to make an exhaustive list of all features they're using.
```
ojeda, line 220: Would `RUSTC_BOOTSTRAP=1` still exist? i.e. it requires to pass the exhaustive list only if it `RUSTC_BOOTSTRAP=1` wasn't passed?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+  We always have the option to extend the syntax in the future; and because this whole feature is perma-unstable, we have the option to rename existing flags as well.
+- We could use environment variables instead of flags.
+  This requires no coordination with Cargo and makes the feature seem less "official", which might discourage people from using it.
+  But it makes the caching situation much worse, and runs into platform-specific limitations like not being able to set an env var more than one time, or hitting implementation limits on the number of env vars that can be set.
+  Flags avoid this by using [response files] and allowing allow-features to be additive.
+- We can """simply""" tell people to stop using nightly features on stable (either politely, or with technical measures).
+  This will have a large negative impact on the ecosystem -
+  rust-analyzer and RustRover will not support running unit tests on stable; `cargo semver-checks` will not work at all on stable; Rust for Linux will break entirely.
+- We can leave the status quo.
+  This is in many ways the worst of all worlds - people still use unstable features on stable, but in hacky ways that break.
```
ojeda, line 248: I think the "worst of all worlds" would be the previous bullet, i.e. breaking existing projects, no? i.e. the current situation works fine in practice, after all.

Regarding "in hacky ways that break", it depends on how one looks at it: if there is no other reasonable way to do it, then it isn't really hacky! :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+- We can separate the toolchain into "stable" and "unstable" channels, and tell distros to package both.
+  This is a big ask from distros, and does not actually help with many of the problems
+  (for example cargo semver-checks cannot rely on it being installed, and rust-analyzer will still have caching issues).
```
ojeda, line 251: Yeah, I think we discussed this in the All-Hands of last year (2025) -- it complicates things for both distros and end-users and doesn't help too much.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+
+This cannot be done in a library or macro.
+
+# Prior art
+[prior-art]: #prior-art
+
```
ojeda, line 262: Another example are C compilers like e.g. Clang which has `*experimental*` flags, but they don't need an `--allow-flags` I think.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+[free-threaded python]: https://docs.python.org/3/whatsnew/3.13.html#whatsnew313-free-threaded-cpython
+[kubernetes-features]: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+[experimental]: https://docs.scala-lang.org/scala3/reference/other-new-features/experimental-defs.html
+[preview]: https://docs.scala-lang.org/scala3/reference/preview/index.html#
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Will this cause flags to be de-facto stable, even moreso than they are now?
+  We could emit a future-compat warning whenever this flag is used; is that sufficient?
```
ojeda:
```suggestion
- Will this cause flags to be de-facto stable, even moreso than they are now?
  We could emit a future-compat warning whenever this flag is used; is that sufficient?
```

If a warning is emitted, it should have a way to be silenced, i.e. we don't want to "spam" end users.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

ojeda:
> Right now, there are several important groups of stakeholders (enumerated in this very RFC) who _cannot_ test or experiment with unstable features, because they cannot use the nightly toolchain.

Do you mean they cannot use `RUSTC_BOOTSTRAP=1`?

text/0000-rustc-bootstrap.md
zackw: If they activate one or two unstable features on stable, that exposes them to much less potential breakage than jumping to nightly would.  It's easier to understand the risk they're taking, and presumably they have a concrete need for those unstable features (even if only curiosity) so the reward for the risk is clearer.

zackw:
> > Right now, there are several important groups of stakeholders (enumerated in this very RFC) who _cannot_ test or experiment with unstable features, because they cannot use the nightly toolchain.
> 
> Do you mean they cannot use `RUSTC_BOOTSTRAP=1`?

No, I mean they cannot use the nightly toolchain, which is currently the only officially sanctioned way of trying out unstable features.  They _could_ use `RUSTC_BOOTSTRAP=1` with the stable toolchain, but that's not officially sanctioned, therefore they (reasonably) may expect that feedback will be rejected.  And it turns on _all_ the unstable features available in that version of the compiler, which is likely to be more instability risk than they are comfortable with.

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,287 @@
...
+- rust-analyzer and RustRover need all values in `rustc --print=cfg` to build the standard library.
+  (see [#139892](https://github.com/rust-lang/rust/issues/139892#issuecomment-2808505610) for an explanation of why this is affected by unstable features)
+- `cargo semver-checks` needs `rustdoc --output-format=json` in order to work at all.
+- Rust for Linux (RfL) needs a way to build a custom version of core.
+  In particular, they mentioned they need to disable float support, because using float registers can cause unsoundness.
+- `rustc_public`'s entire mission is to wrap unstable APIs with stable ones and therefore needs access to all `rustc_private` features.
+
+Why are these uses ok? Three reasons:
+- Each of these, except for rustc_public, is an external tool, not a library.
+  They do not need unstable language features, only unstable tool output.
```
DaAlbrecht: Tool output to me sounds a bit too specific, especially since that is a subcategory of the unstable [cargo features](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#list-of-unstable-features)

```suggestion
  They do not need unstable language features, only unstable tool features.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..6353abf11da599225ee3775a7eda3adc44214d5b)*

ojeda:
> They could use RUSTC_BOOTSTRAP=1 with the stable toolchain, but that's not officially sanctioned, therefore they (reasonably) may expect that feedback will be rejected.

That is what I mean: it is possible `RUSTC_BOOTSTRAP=1` with stable toolchains (which is what we do in Rust for Linux), and we do contribute to the stabilization and testing of features upstream etc.

So that is why I was asking, i.e. those other stakeholders could, in principle, use it too.

In other words, the RFC may be good idea, but it is not strictly needed in order to test those features by those stakeholders.

>  And it turns on all the unstable features available in that version of the compiler, which is likely to be more instability risk than they are comfortable with.

Using `RUSTC_BOOTSTRAP=1` still requires one to e.g. enable the unstable language features one needs and to pass the flags etc., i.e. it is not automatically enabling them. My understanding is that `--allow-flags` would work similarly but for a subset, so it may not be a substantial difference in that regard, if I understand correctly.

Lokathor:
@zackw I do not understand how "Stable with an unstable feature enabled" is somehow meaningfuly different from "Nightly with the same feature enabled". Can you point to some specific non-gated instability as an example of the dangers of Nightly compared to Stable that a person would have to contend with?

jyn514:
@lokathor i discuss this in detail in the "motivation" section, search for "can't people just use nightly".

please use threads for any further comments.

text/0000-rustc-bootstrap.md
jyn514: The "buffer" here is between breakage in nightly Rust and the distros that package Rust For Linux. My understanding is that they coordinate with you to determine which toolchains they can package, right? So if something breaks, you have a chance to fix it, rather than things instantly breaking on Debian Stable or whatever.

text/0000-rustc-bootstrap.md
jyn514: I mean something like "a self-contained build". The thing I'm contrasting to here is `cargo build` on a library crate, which is not self-contained because other people can pull in that library as a dependency from crates.io. When distros package Linux, to my knowledge, they don't change its build flags, although they might change `make menuconfig` or something like that.

I'm not sure how to reword this ...

text/0000-rustc-bootstrap.md
jyn514: *Both* of the flags are a limit. I do not attempt to decide who is "the most" top-level — *all* flags are AND-ed, not OR-ed.

text/0000-rustc-bootstrap.md
jyn514: Yes. But I want to get people away from RUSTC_BOOTSTRAP=1 because it's caused us so many issues in the past. See the "Future possibilities" section.

text/0000-rustc-bootstrap.md
jyn514: No, warning each time is intentional. I do not want people to be completely unaware that they're depending on unstable features.

text/0000-rustc-bootstrap.md
ojeda: We have an MSRV, currently the latest Debian Stable's `rustc`, i.e. 1.85.0, and then distributions pick whatever version they want. So e.g. Ubuntu LTS may pick 1.93.1, and that's it, i.e. they don't really see/use nightly. So I am not sure what you mean by Debian Stable breaking.

With distributions like Fedora that release packages for the latest Rust release within a few days of the release there is of course way less room, but it is still a few weeks due to the beta, and anyway we do our best to fix things while on nightly (ideally even before the PR is merged, thanks to the Rust CI building the kernel).

text/0000-rustc-bootstrap.md
jyn514: Yes, for wrapping rustc directly also. I changed the wording.

text/0000-rustc-bootstrap.md
jyn514: FWIW you (RfL) control Rustc's invocation, so you could (e.g.) post-process the JSON messages and filter out the ones about unstable warnings.

text/0000-rustc-bootstrap.md
ojeda: As long as it is still there, it should be fine, i.e. to avoid immediate breakage everywhere. Thanks!

text/0000-rustc-bootstrap.md
ojeda: Yeah, what I am trying to understand is the use case, e.g. the "Cargo level" sounds like "end-user can limit the flags used", but I am not sure I get why one would want to do that.

text/0000-rustc-bootstrap.md
jyn514: I've linked that issue to clarify that this isn't just limited to build-std.

text/0000-rustc-bootstrap.md
ojeda: > No, warning each time is intentional. I do not want people to be completely unaware that they're depending on unstable features.

In that case, I cannot agree with the RFC, sorry... We cannot spam end users, distributions and other projects with warnings they cannot (and shouldn't) do anything about.

I mean, I understand that you want to prevent people from using features unaware. That is good, but if a user intentionally adds yet another flag to acknowledge the warning, then they are already aware, no? Unless you mean you want end users (rather than the developers of a project) to know about it, but then I don't understand why you would want that to begin with.

> FWIW you (RfL) control Rustc's invocation, so you could (e.g.) post-process the JSON messages and filter out the ones about unstable warnings.

That sounds like a workaround to silence a warning that shouldn't be emitted to begin with since it is expected.

Generally speaking, any warning that cannot be acknowledged by a build system is likely a mistake. (My understanding is that there were `rustc` warnings introduced in the past that do not have a way to allow them due to how they implemented/were designed initially, but introducing more of those on purpose sounds quite bad.)

text/0000-rustc-bootstrap.md
jyn514: Added a footnote.

text/0000-rustc-bootstrap.md
jyn514: Because in most cases the author of the package is a different person than the person who's passing flags to Cargo (and indirectly to Rustc). For RfL these are the same (you write all the makefiles yourself), but the author of Cargo.toml is usually a different person than whoever's running `cargo build`. If the two disagree, neither of them should silently get opted into more unstable features.

text/0000-rustc-bootstrap.md
ojeda: Ah, so you mean the developer of the project. I would suggest picking some other word than "end-user" in that case, since to me that sounds like the actual user, e.g. someone that may not even know Rust, if that makes sense.

(Regarding Linux: distributions tweak the config which in turn definitely changes flags, but at least those are expected. They may also use different toolchains and a wide range of versions for each, e.g. GCC and Clang, which are already a big change of its own too. And, in principle, they may also use the variables that Kbuild provides to tweak the flags too, if they know what they are doing.)

jyn514: No, the developer of the project is not the same as the person running the build in the case of someone writing a Cargo library.

text/0000-rustc-bootstrap.md
ojeda: Hmm... I am not sure I follow. Why would a developer want to limit the flags "manually" in the command-line vs. configuring it elsewhere somehow?

jyn514: Because this may be a one-off invocation where they are not configuring the *project*, only the *tool invocation*. See "why allow using unstable features", specifically the bullets around rust-analyzer and semver-checks.

text/0000-rustc-bootstrap.md
ojeda: By developer I mean the developer of the project using another crate. Either way, if they are a developer, they are not an end user, if that makes sense.

i.e. my original comment was about the confusing with "end user", which I think usually is understood as users of whatever final products, rather than intermediate crates etc.

text/0000-rustc-bootstrap.md
ojeda: Ah, alright. I thought it could have been more about a developer e.g. testing the flags needed.

text/0000-rustc-bootstrap.md
jyn514: > Generally speaking, any warning that cannot be acknowledged by a build system is likely a mistake.

The problem is, again, how to distinguish a developer from a self-contained build. I don't want people just running `cargo build` to have unstable features hidden if they're e.g. enabled and hidden in `.cargo/config.toml`, which is controlled by the project, not the person running the build.

The thing I came up with was "able to post-process the output", which is true of all of RfL, semver-checks, and RA. If we get rid of that criteria and add a lint flag, then people can 1. run `cargo build` 2. with unstable features 3. on someone else's project they're using as a library 4. on stable 5. without ever seeing warnings. I am unwilling to accept that.

jyn514: I guess we could add back some environment variable and have *that* be the way to silence the future-compat warning, but that feels kinda icky.

Lokathor: The .cargo/config.toml can just set that env variable if it wanted. Unless you block the variable, in which case just block the flag instead and save yourself an env variable (because env variables are a PITA on windows and should broadly be avoided for that alone).

jyn514: maybe .cargo/config.toml is the wrong thing to think about here ... i actually don't care so much if individual libraries enable things, as long as they're at the top-level. i'm mostly worried about them being used as dependencies.

in that case i think an env var probably works? you can enable it in .cargo/config.toml but not in Cargo.toml, so it will be ignored if you're depending on the crate as a library.

Lokathor: Then still, why not make it a Cargo.toml setting that's ignored when the crate isn't the top level build?

Darksonn: Some experience from the `--cfg tokio_unstable` option we use to implement unstable features in Tokio ... for us the philosophy is that the project that decides to opt-in or not is the project that performs the final build.

So for libraries in the crates ecosystem that wish to use unstable features, they may be able to enable them locally with `.cargo/config.toml`, but this does not get published on crates.io, so if anyone uses that library, then the person using it from their project will have to add the same logic to *their* downstream `.cargo/config.toml` to opt-in.

I think this works well as a deterrent, and I don't think warnings are needed here.

Lokathor: I don't disagree but we already have settings in Cargo.toml that aren't passed up to users of the crate as a library (eg: build.profile), so why aren't we also thinking of this as being a similar setting that is **both** in Cargo.toml but also doesn't float up to users of the library part of the crate?

Darksonn: I don't think it's important whether it's in `Cargo.toml` or `.cargo/config.toml` as long as it has the behavior that it doesn't apply when downloaded as a library from crates.io.

zackw: This is wandering off topic but I really want to see `.cargo/config.toml` stop being a thing people put in source trees (see https://internals.rust-lang.org/t/cargo-project-config-unification/24082/5 for why).

jyn514: I agree with @Darksonn's reasoning. I've moved the force-warn to "Rationale and alternatives" and out of the main RFC.

ojeda: Thanks for removing the warning from the RFC!

> The problem is, again, how to distinguish a developer from a self-contained build.

I think that is a concern at the Cargo layer, i.e. it could be special cased it in Cargo one way or another if one really wants such a warning, but having an unconditional warning at the `rustc` level is what sounded to me like a bad idea.

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+
+`allow-unstable-flags` is accepted on all channels.
+When it's not present, the default on stable/beta is to ban all unstable flags,
+and the default on nightly is to allow all unstable flags.
+
+Unrecognized flag names in `allow-unstable-flags` are a hard error.
+
+## Non-cargo rules
+
+Stable/beta Rustfmt now errors instead of warning when an unstable option is set without also setting `--allow-unstable-flags`.
```
epage, line 171: This could probably be relaxed to "if the absence of the setting would cause a reformat, it should error". This is basically the approach we take in Cargo when deciding whether use of an unstable feature should be a warning or an error.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+If `allow-unstable-flags` is passed multiple times, the *intersection* of all values is used, not the union.
+This matters in cases where two parties don't trust each other, such as running `cargo build --allow-unstable-flags=rustc=x` in a workspace with `build.rustflags=--allow-unstable-flags=x,y`: this should be equivalent to `rustc --allow-unstable-flags=x`.
+
+## Cargo rules
+
+Build scripts cannot set these flags; `cargo::rustc-flags` continues to only accept `-l` and `-L` flags.
+
+Each Cargo flag takes a value that starts with a tool name, then the string '=', then a valid value for a non-Cargo flag.
+The tool can be `cargo`, in which case the flag applies to the unstable flags of Cargo itself.
+Tool names are the name of the exact binary that will be spawned: `rustdoc`, `clippy-driver`, etc.
```
epage, line 188: Cargo has no knowledge of `clippy-driver`; it is just a `RUSTC_WORKSPACE_WRAPPER`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+If `RUSTC_WRAPPER` or `RUSTC_WORKSPACE_WRAPPER` is set, the intersection of the flags for `rustc` and the wrapper are passed; this requires additional work from the user but avoids silently passing unstable flags to more tools than intended.
+
+
+If `allow-unstable-flags` is passed multiple times, tools are unioned, but values are intersected.
+In other words, `cargo doc --allow-unstable-flags=rustc=x --allow-unstable-flags=rustdoc=y` will pass `--allow-unstable-flags=x` to Rustc and `--allow-unstable-flags=y` to Rustdoc.
+
+`cargo <cmd> --allow-unstable-flags` applies to both Cargo and all tools it spawns.
+That is, it passes the flag to Rustc when Rustc is spawned.
+This applies to all packages, not just the current workspace.
+In practice, the only tool passing unstable flags to Cargo is `cargo-semver-checks`, which is building all dependencies in any case.
```
epage, line 199: I feel I'm missing something here. A lot of tools pass unstable flags to Cargo. RustRover is an easy case that comes to mind.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+
+Unrecognized tool names are an error.
```
epage, line 201: This might be a bit messy to determine

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+libtest runners accept `--allow-unstable-flags`.
+
+Each non-Cargo flag takes one of the following strings as a value:
+- An empty string, which indicates that no flags are allowed (default on stable)
+- A comma-separated list of flag names.
+
+If `allow-unstable-flags` is passed multiple times, the *intersection* of all values is used, not the union.
+This matters in cases where two parties don't trust each other, such as running `cargo build --allow-unstable-flags=rustc=x` in a workspace with `build.rustflags=--allow-unstable-flags=x,y`: this should be equivalent to `rustc --allow-unstable-flags=x`.
+
+## Cargo rules
```
epage, line 182: I feel like cargo, or all tools, should have a hard warning when this feature is used, maybe saying something similar to what is said under "Stability policy".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
Veykril: so does rust-analyzer too

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+
+[alona]: https://github.com/aDotInTheVoid/
+[Zulip]: https://rust-lang.zulipchat.com/
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Shared rules
+
+`allow-unstable-flags` is a comma-separated list of CLI flag names, with the leading `-Z` or `--` (if any) removed.
```
epage, line 157: The Cargo team recently discussed changing `cargo install` from ignoring the lockfile to acting like any other cargo command by default (instead of requiring the misnamed `--locked`). When discussing the possibility of a flag to restore the old behavior, something we were considering was having the name better communicate "here be dragons".

I wonder if such a naming scheme should be used here as well. It is helped by being longer than `-Z` and in having `unstable`. `allow` feels a bit more passive.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Cargo rules
+
+Build scripts cannot set these flags; `cargo::rustc-flags` continues to only accept `-l` and `-L` flags.
+
+Each Cargo flag takes a value that starts with a tool name, then the string '=', then a valid value for a non-Cargo flag.
+The tool can be `cargo`, in which case the flag applies to the unstable flags of Cargo itself.
+Tool names are the name of the exact binary that will be spawned: `rustdoc`, `clippy-driver`, etc.
+If `RUSTC` or `RUSTDOC` is set, the tool name is still `rustc`/`rustdoc`, not the overridden value.
+If `RUSTC_WRAPPER` or `RUSTC_WORKSPACE_WRAPPER` is set, the intersection of the flags for `rustc` and the wrapper are passed; this requires additional work from the user but avoids silently passing unstable flags to more tools than intended.
```
epage, line 190: Alternatively, we could define this to operate in parallel to `RUSTFLAGS` and `RUSTDOCFLAGS`, ie `rust=` is used anywhere `RUSTFLAGS` is used and `rustdoc=` is used anywhere `RUSTDOCFLAGS` is used.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

epage, line 190: This isn't too clear. Maybe it would be good to provide an example of `RUSTC_WORKSPACE_WRAPPER` being used.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+- Nightly is not the same as stable-with-`RUSTC_BOOTSTRAP`.
+  In particular, stable contains backports and nightly does not.
+
+## Why this exact mechanism?
+
+Currently, these tools use [`RUSTC_BOOTSTRAP=1`][rustc-bootstrap] as a workaround.
+But enabling RUSTC_BOOTSTRAP for one part of the toolchain enables it for *all* parts of the toolchain; in particular:
+- `proc-macro2` uses `cargo:rerun-on-env-changed=RUSTC_BOOTSTRAP`, causing cache thrashing whenever this env var changes.
+- rust-analyzer wants to enable `RUSTC_BOOTSTRAP` only for cargo and libtest, but the variable enables features for rustc as well.
+  `RUSTFLAGS="-Z allow-features="` fixes this for lang features, but at the price of thrashing the cache; and there is no equivalent way to disable unstable CLI features.
```
epage, line 89: Whats the cache thrashing scenario here? We used to thrash the caches on RUSTFLAGS changes but we fixed that a year or two ago and now create dedicated cache entries.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

text/0000-rustc-bootstrap.md
```diff
@@ -0,0 +1,310 @@
...
+## Implementation notes
+
+### Caching
+
+Currently, changing `RUSTC_BOOTSTRAP` does not invalidate Cargo's build cache by itself, but in practice can cause build scripts deep in the dependency tree to re-run.
+With `--allow-unstable-flags`, we separate the mechanism for enabling flags from the mechanism for enabling lang features,
+greatly decreasing how often build scripts that detect `RUSTC_BOOTSTRAP` rerun.
+
+We suggest, but do not require, that the flag is made part of the fingerprint tracking, not unit cache tracking,
+so that changing the enabled features overwrites the cache rather than adding to it.
```
epage, line 212: What is the motivation for this? I thought we were trying to avoid cache thrashing but this would cause it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3965/c488c3bb7a058c81460ea0161732d4ca5bfc8885..d6779ea48b0a166496688e1b15df900be7ef3f26)*

PR #3966, opened 2026-05-31
BTF relocations

vadorovsky:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3966)*

The RFC proposes the `btf_relocatable_types` feature gate, that provides a `#[repr(Btf)]` representation and low-level field-info intrinsics that emit [BTF (BPF TypeFormat)][btf] [CO-RE (Compile Once, Run Everywhere)][co-re] relocations.

It also documents the relationship to `offset_of!`, ordinary field projection and LLVM BPF lowering.

This feature was originally proposed as [pre-RFC][pre-rfc].

[btf]: https://docs.kernel.org/bpf/btf.html
[co-re]: https://nakryiko.com/posts/bpf-portability-and-co-re/
[pre-rfc]: https://internals.rust-lang.org/t/pre-rfc-btf-relocations/24161/25

[Rendered](https://github.com/vadorovsky/rfcs/blob/btf-relocations/text/0000-btf-relocations.md)

### DIFF
--- a/text/0000-btf-relocations.md
+++ b/text/0000-btf-relocations.md
@@ -1,0 +1,450 @@
+- Feature Name: `btf_relocations`
+- Start Date: 2026-05-30
+- RFC PR: [rust-lang/rfcs#3966](https://github.com/rust-lang/rfcs/pull/3966)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+## Summary
+[summary]: #summary
+
+Add experimental Rust support for [Compile Once, Run Everywhere (CO-RE)][co-re]
+relocations based on the [BPF Type Format (BTF)][btf]. The feature introduces a
+`#[repr(Btf)]` representation for structs and unions whose field layout must be
+queried through BTF-aware operations, and adds BTF-aware macros for accessing
+the fields:
+
+* `core::btf::field_byte_offset!`
+* `core::btf::field_byte_size!`
+* `core::btf::field_exists!`
+
+The user-facing feature is gated by `#![feature(btf_relocations)]`.
+
+## Motivation
+[motivation]: #motivation
+
+[BTF][btf] is the type metadata format used by the Linux kernel and eBPF
+tooling. eBPF loaders such as [Aya][aya] and [libbpf][libbpf] use BTF for
+relocations: the compiled program records which field or array element it
+intended to access, and the loader rewrites the bytecode to match the layout of
+the kernel it is about to run on.
+
+Clang and GCC are capable of emitting such relocations.
+
+Rust can already target eBPF, but it does not currently have a way to emit
+these BTF access relocations. In practice, that means Rust eBPF programs often
+have to pick one of three inconvenient options:
+
+- Vendor the exact kernel type definitions and rebuild for each supported kernel
+  layout.
+- Avoid typed field access and manually encode offsets, sacrificing readability
+  and maintainability.
+- Write a module in C solely for accessing kernel types and use `build.rs` to
+  link it to the Rust project.
+
+`offset_of!` is the wrong primitive for this purpose: it intentionally folds to a
+compile-time layout constant, so backend codegen no longer knows which source
+field was being queried. Ordinary Rust field projection is also insufficient:
+once it becomes a normal memory access, the BTF field identity needed for CO-RE
+relocation has been lost.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `btf_relocations` feature is for Rust code that models external BTF
+types, primarily Linux kernel types used by eBPF programs.
+
+A type that should participate in BTF relocation is written with
+`#[repr(Btf)]`:
+
+```rust
+#![feature(btf_relocations)]
+
+#[repr(Btf)]
+pub struct task_struct {
+    pub pid: i32,
+    pub tgid: i32,
+}
+```
+
+`#[repr(Btf)]` is intentionally not just a layout hint. It marks a type as one
+whose fields should not be accessed through ordinary Rust field projection for
+accesses that are meant to be relocatable. For such types, direct field access
+is rejected:
+
+```rust
+#![feature(btf_relocations)]
+
+#[repr(Btf)]
+pub struct task_struct {
+    pub pid: i32,
+}
+
+fn pid(task: &task_struct) -> i32 {
+    task.pid
+    // error: cannot access fields of a `#[repr(Btf)]` type directly
+}
+```
+
+The same restriction applies to `offset_of!`:
+
+```rust
+#![feature(btf_relocations)]
+
+#[repr(Btf)]
+pub struct task_struct {
+    pub pid: i32,
+}
+
+const PID_OFFSET: usize = core::mem::offset_of!(task_struct, pid);
+// error: cannot use `offset_of!` with a `#[repr(Btf)]` type
+```
+
+Instead, code that needs field metadata uses BTF-aware queries. These macros
+take a root carrier type and a field path:
+
+```rust
+#![feature(btf_relocations)]
+
+#[repr(Btf)]
+pub struct task_struct {
+    pub pid: i32,
+    pub tgid: i32,
+}
+
+impl task_struct {
+    #[inline]
+    pub fn has_pid(&self) -> bool {
+        core::btf::field_exists!(task_struct, pid)
+    }
+
+    #[inline]
+    pub fn pid_offset(&self) -> Option<usize> {
+        self.has_pid().then(|| {
+            Some(core::btf::field_byte_offset!(task_struct, pid))
+        })
+    }
+
+    #[inline]
+    pub fn pid_size(&self) -> Option<usize> {
+        self.has_pid().then(|| {
+            Some(core::btf::field_byte_size!(task_struct, pid))
+        })
+    }
+
+    #[inline]
+    pub fn pid(&self) -> Option<&i32> {
+        self.has_pid().then(|| {
+            let offset = core::btf::field_byte_offset!(task_struct, pid);
+            let ptr = self as *const task_struct as *const u8;
+
+            // SAFETY: the BTF relocation says that `se.vruntime` exists in the
+            // target layout, and the returned offset is relative to `task_struct`.
+            Some(unsafe { &*(ptr.add(offset) as *const i32) })
+        })
+    }
+
+    #[inline]
+    pub fn has_tgid(&self) -> bool {
+        core::btf::field_exists!(task_struct, tgid)
+    }
+
+    #[inline]
+    pub fn tgid_offset(&self) -> Option<usize> {
+        self.has_tgid().then(|| {
+            Some(core::btf::field_byte_offset!(task_struct, tgid))
+        })
+    }
+
+    #[inline]
+    pub fn tgid_size(&self) -> Option<usize> {
+        self.has_tgid().then(|| {
+            Some(core::btf::field_byte_size!(task_struct, tgid))
+        })
+    }
+
+    #[inline]
+    pub fn tgid(&self) -> Option<&i32> {
+        self.has_tgid().then(|| {
+            let offset = core::btf::field_byte_offset!(task_struct, tgid);
+            let ptr = self as *const task_struct as *const u8;
+
+            // SAFETY: the BTF relocation says that `se.vruntime` exists in the
+            // target layout, and the returned offset is relative to `task_struct`.
+            Some(unsafe { &*(ptr.add(offset) as *const i32) })
+        })
+    }
+}
+```
+
+Nested field paths are supported. For example, access to the fields of
+`sched_entity` that is nested in `task_struct` can be done with one macro call:
+
+```rust
+#![feature(btf_relocations)]
+
+#[repr(Btf)]
+pub struct load_weight {
+    pub weight: usize,
+}
+
+#[repr(Btf)]
+pub struct sched_entity {
+    pub load: load_weight,
+    pub vruntime: u64,
+}
+
+#[repr(Btf)]
+pub struct task_struct {
+    pub se: sched_entity,
+}
+
+impl task_struct {
+    #[inline]
+    pub fn sched_vruntime(&self) -> Option<&u64> {
+        core::btf::field_exists!(task_struct, se.vruntime).then(|| {
+            let offset = core::btf::field_byte_offset!(task_struct, se.vruntime);
+            let ptr = self as *const task_struct as *const u8;
+
+            // SAFETY: the BTF relocation says that `se.vruntime` exists in the
+            // target layout, and the returned offset is relative to `task_struct`.
+            Some(unsafe { &*(ptr.add(offset) as *const u64) })
+        })
+    }
+
+    #[inline]
+    pub fn sched_load_weight(&self) -> Option<&usize> {
+        core::btf::field_exists!(task_struct, se.load.weight).then(|| {
+            let offset = core::btf::field_byte_offset!(task_struct, se.load.weight);
+            let ptr = self as *const task_struct as *const u8;
+
+            // SAFETY: the BTF relocation says that `se.load.weight` exists in the
+            // target layout, and the returned offset is relative to `task_struct`.
+            Some(unsafe { &*(ptr.add(offset) as *const usize) })
+        })
+    }
+}
+```
+
+The offset returned for a nested path is relative to the root carrier type,
+`task_struct`.
+
+On BPF targets with BTF-capable backend support and debug info enabled, these
+queries lower to CO-RE relocations. On targets or backends without BTF
+relocation support, they fall back to the current compilation unit's ordinary
+layout information.
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Feature gate
+
+The language feature is named `btf_relocations`.
+
+The feature gate controls the user-facing BTF relocation surface, including
+`#[repr(Btf)]` and the `core::btf` field-info macros.
+
+### `#[repr(Btf)]`
+
+The `repr` attribute accepts a new representation hint:
+
+```rust
+#[repr(Btf)]
+struct S {
+    field: u32,
+}
+```
+
+`#[repr(Btf)]` is accepted on structs and unions. It is rejected on other item
+kinds.
+
+A `#[repr(Btf)]` type uses C-compatible field layout. In compiler terms,
+`repr(Btf)` implies the layout constraints of `repr(C)` and also marks the type
+as BTF-relocatable. This gives the backend stable field ordering and offsets for
+the compile-time fallback while preserving a distinct marker for type checking
+and codegen.
+
+Direct field projection from a `#[repr(Btf)]` ADT is rejected. This includes
+projections reached through autoderef:
+
+```rust
+task.pid
+```
+
+The `offset_of!` macro is also rejected when any container in the queried path is
+a `#[repr(Btf)]` ADT.
+
+These restrictions avoid silently producing non-relocatable code for operations
+that appear to query a relocatable type. Code that genuinely wants a normal
+non-relocatable Rust type should not use `#[repr(Btf)]`.
+
+### Field-info macros
+
+The following macros are added under `core::btf`:
+
+```rust
+core::btf::field_byte_offset!(Carrier, field.path) -> usize
+core::btf::field_byte_size!(Carrier, field.path) -> usize
+core::btf::field_exists!(Carrier, field.path) -> bool
+```
+
+`Carrier` is the root local Rust type whose BTF graph describes the access.
+The second argument is a dot-separated Rust field path starting from `Carrier`.
+The compiler type checks this path using the same field lookup rules as
+`offset_of!`, except that `#[repr(Btf)]` containers are accepted for these
+BTF-aware queries.
+
+The macros have the following meanings:
+
+* `field_byte_offset!(Carrier, field.path)` returns the byte offset of the
+  complete field path from the root carrier.
+* `field_byte_size!(Carrier, field.path)` returns the byte size of the terminal
+  field.
+* `field_exists!(Carrier, field.path)` returns whether the complete field path
+  exists in the target BTF type.
+
+These macros do not perform memory access. They are metadata queries and do not
+require the caller to uphold memory-safety invariants.
+
+Nested paths are supported:
+
+```rust
+core::btf::field_byte_offset!(task_struct, se.load.weight)
+```
+
+These macros are the only user-facing API for BTF field-info queries.
+Implementations may lower them to a dedicated internal compiler operation that
+carries the root type, resolved field path, and query kind; that operation is
+not exposed as a callable API.
+
+### Backend lowering
+
+For LLVM BPF codegen with debug info enabled, each resolved field-info query
+lowers to `llvm.bpf.preserve.field.info` with the corresponding BPF field-info
+kind:
+
+* byte offset: `0`
+* byte size: `1`
+* field exists: `2`
+
+The frontend does not expose LLVM's
+`llvm.preserve.{struct,array,union}.access.index` operations. The codegen
+backend constructs the required access-index chain internally from the Rust
+carrier type and compiler-generated field path.
+
+The result of the LLVM intrinsic is an integer value. Offset and size queries
+are zero-extended to `usize`. Existence queries are compared against zero and
+return `bool`.
+
+If the target, backend, or codegen mode cannot emit BTF field relocations, the
+field-info queries fall back to ordinary layout-computed values:
+
+* `field_byte_offset!` returns the complete field-path offset from the
+  current compilation layout.
+* `field_byte_size!` returns the field size from the current compilation
+  layout.
+* `field_exists!` returns `true` for a field path present in the current
+  compilation layout.
+
+BTF CO-RE relocation emission is only meaningful for BPF targets, and it
+requires the debug metadata used to describe the relevant types.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+This is a niche feature aimed at one target family and one ecosystem workflow.
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+### Emit BTF relocations in bpf-linker
+
+The main alternative is to emit BTF relocations in [bpf-linker][bpf-linker],
+which is a bitcode linker used exclusively for BPF targets. However, it
+prevents us from supporting ld type of linkers (e.g. binutils, lld) for BPF
+targets in future.
+
+### Use `offset_of!`
+
+`offset_of!` is intentionally a constant layout query. It does not preserve the
+field identity needed to emit a BTF relocation. Reusing it would either silently
+produce non-relocatable code or require changing the meaning of an existing
+stable API in target-specific ways.
+
+### Make ordinary field access relocatable
+[relocatable-field-access]: #relocatable-field-access
+
+The compiler could try to make `task.pid` on selected types emit relocatable
+accesses automatically. This is what Clang currently does for CO-RE field
+accesses: ordinary C field projection can be preserved and lowered to BTF
+relocations. This is attractive ergonomically, but it requires a more intrusive
+change to MIR and the design of a proper abstract machine with operational
+semantics. It also has some similarities to the [`Sized` hierarchy
+RFC][sized-hierarchy], which has not yet been accepted.
+
+This overlaps with the accepted [Field Projections project goal][field-projections],
+which is exploring virtual places as a general mechanism for custom field
+projection. This RFC deliberately does not depend on that work: it provides the
+low-level BTF field metadata queries needed for CO-RE today, while leaving a
+future field-projection-based ergonomic surface open.
+
+Providing the field-info queries proposed in this RFC does not rule out
+exploring this alternative in the future. On the contrary, Clang provides both
+explicit field-info builtins and field projection. It makes sense to treat these
+as separate RFCs.
+
+### Use a separate `#[relocatable]` attribute
+
+A standalone `#[relocatable]` attribute was considered. `#[repr(Btf)]` is more
+specific about the external format and makes clear that the marker affects
+representation-related compiler behavior. It also groups the feature with other
+layout and representation attributes.
+
+### Expose LLVM intrinsics directly
+
+LLVM already provides BPF intrinsics for preserving access indices and querying
+field information. Exposing those directly would leak backend-specific details
+into Rust code and make the API unusable for non-LLVM codegen backends.
+
+The proposed Rust field-info queries are backend-neutral. LLVM-specific lowering
+remains an implementation detail. Lowering for GCC might be implemented in the
+future.
+
+### Implement this entirely in libraries
+
+Libraries can provide ergonomic wrappers, but they cannot make `offset_of!`
+preserve field identity after type checking and MIR lowering, nor can they
+reliably construct backend metadata for CO-RE relocation emission. The compiler
+must participate.
+
+## Prior art
+[prior-art]: #prior-art
+
+Clang and LLVM support BPF CO-RE through builtins and LLVM intrinsics such as
+`__builtin_preserve_access_index`, `__builtin_preserve_field_info`, and
+`llvm.bpf.preserve.field.info`. C BPF programs commonly use libbpf macros such
+as `BPF_CORE_READ` and `bpf_core_field_exists` to generate these relocations.
+
+Rust BPF projects such as [Aya][aya] need access to the same relocation
+model. Today, they generally rely on generated bindings, helper macros, or
+backend behavior outside Rust's stable language surface.
+
+This RFC follows the same underlying CO-RE model as C/Clang while avoiding a
+direct dependency on C syntax or LLVM-specific frontend intrinsics.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* How should non-LLVM codegen backends expose equivalent relocation support?
+
+## Future possibilities
+[future-possibilities]: #future-possibilities
+
+Once the [`Sized` hierarchy RFC][sized-hierarchy] is accepted,
+[ordinary field access could be made relocatable][relocatable-field-access].
+
+[btf]: https://docs.kernel.org/bpf/btf.html
+[co-re]: https://nakryiko.com/posts/bpf-portability-and-co-re/
+[aya]: https://github.com/aya-rs/aya
+[libbpf]: https://github.com/libbpf/libbpf
+[bpf-linker]: https://github.com/aya-rs/bpf-linker
+[field-projections]: https://github.com/rust-lang/rust-project-goals/issues/390
+[sized-hierarchy]: https://github.com/rust-lang/rfcs/pull/3729
--- a/text/0000-btf-relocations.md
+++ b/text/0000-btf-relocations.md
@@ -80,7 +80,7 @@ pub struct task_struct {
 
 fn pid(task: &task_struct) -> i32 {
     task.pid
-    // error: cannot access fields of a `#[repr(Btf)]` type directly
+    // error: cannot access fields of a `#[btf_relocatable]` type directly
 }
 ```
 
@@ -96,7 +96,7 @@ pub struct task_struct {
 }
 
 const PID_OFFSET: usize = core::mem::offset_of!(task_struct, pid);
-// error: cannot use `offset_of!` with a `#[repr(Btf)]` type
+// error: cannot use `offset_of!` with a `#[btf_relocatable]` type
 ```
 
 Instead, code that needs field metadata uses BTF-aware queries. These macros
--- a/text/0000-btf-relocations.md
+++ b/text/0000-btf-relocations.md
@@ -412,9 +412,10 @@ Clang and LLVM support BPF CO-RE through builtins and LLVM intrinsics such as
 `llvm.bpf.preserve.field.info`. C BPF programs commonly use libbpf macros such
 as `BPF_CORE_READ` and `bpf_core_field_exists` to generate these relocations.
 
-Rust BPF projects such as [Aya][aya] need access to the same relocation
-model. Today, they generally rely on generated bindings, helper macros, or
-backend behavior outside Rust's stable language surface.
+Rust BPF projects need to run on different kernels. Today, they usually achieve
+that either by re-generating the types and compiling separately for each
+kernel, or by linking a C module that is used only for interacting with kernel
+types.
 
 This RFC follows the same underlying CO-RE model as C/Clang while avoiding a
 direct dependency on C syntax or LLVM-specific frontend intrinsics.
--- a/text/0000-btf-relocations.md
+++ b/text/0000-btf-relocations.md
@@ -14,7 +14,6 @@ the fields:
 
 * `core::btf::field_byte_offset!`
 * `core::btf::field_byte_size!`
-* `core::btf::field_exists!`
 
 The user-facing feature is gated by `#![feature(btf_relocations)]`.
 
@@ -112,28 +111,18 @@ pub struct task_struct {
 
 impl task_struct {
     #[inline]
-    pub fn has_pid(&self) -> bool {
-        core::btf::field_exists!(task_struct, pid)
-    }
-
-    #[inline]
     pub fn pid_offset(&self) -> Option<usize> {
-        self.has_pid().then(|| {
-            Some(core::btf::field_byte_offset!(task_struct, pid))
-        })
+        core::btf::field_byte_offset!(task_struct, pid)
     }
 
     #[inline]
     pub fn pid_size(&self) -> Option<usize> {
-        self.has_pid().then(|| {
-            Some(core::btf::field_byte_size!(task_struct, pid))
-        })
+        core::btf::field_byte_size!(task_struct, pid)
     }
 
     #[inline]
     pub fn pid(&self) -> Option<&i32> {
-        self.has_pid().then(|| {
-            let offset = core::btf::field_byte_offset!(task_struct, pid);
+        core::btf::field_byte_offset!(task_struct, pid).map(|offset| {
             let ptr = self as *const task_struct as *const u8;
 
             // SAFETY: the BTF relocation says that `se.vruntime` exists in the
@@ -143,28 +132,18 @@ impl task_struct {
     }
 
     #[inline]
-    pub fn has_tgid(&self) -> bool {
-        core::btf::field_exists!(task_struct, tgid)
-    }
-
-    #[inline]
     pub fn tgid_offset(&self) -> Option<usize> {
-        self.has_tgid().then(|| {
-            Some(core::btf::field_byte_offset!(task_struct, tgid))
-        })
+        core::btf::field_byte_offset!(task_struct, tgid)
     }
 
     #[inline]
     pub fn tgid_size(&self) -> Option<usize> {
-        self.has_tgid().then(|| {
-            Some(core::btf::field_byte_size!(task_struct, tgid))
-        })
+        core::btf::field_byte_size!(task_struct, tgid)
     }
 
     #[inline]
     pub fn tgid(&self) -> Option<&i32> {
-        self.has_tgid().then(|| {
-            let offset = core::btf::field_byte_offset!(task_struct, tgid);
+        core::btf::field_byte_offset!(task_struct, tgid).map(|offset| {
             let ptr = self as *const task_struct as *const u8;
 
             // SAFETY: the BTF relocation says that `se.vruntime` exists in the
@@ -200,8 +179,7 @@ pub struct task_struct {
 impl task_struct {
     #[inline]
     pub fn sched_vruntime(&self) -> Option<&u64> {
-        core::btf::field_exists!(task_struct, se.vruntime).then(|| {
-            let offset = core::btf::field_byte_offset!(task_struct, se.vruntime);
+        core::btf::field_byte_offset!(task_struct, se.vruntime).map(|offset| {
             let ptr = self as *const task_struct as *const u8;
 
             // SAFETY: the BTF relocation says that `se.vruntime` exists in the
@@ -212,8 +190,7 @@ impl task_struct {
 
     #[inline]
     pub fn sched_load_weight(&self) -> Option<&usize> {
-        core::btf::field_exists!(task_struct, se.load.weight).then(|| {
-            let offset = core::btf::field_byte_offset!(task_struct, se.load.weight);
+        core::btf::field_byte_offset!(task_struct, se.load.weight).map(|offset| {
             let ptr = self as *const task_struct as *const u8;
 
             // SAFETY: the BTF relocation says that `se.load.weight` exists in the
@@ -278,9 +255,8 @@ non-relocatable Rust type should not use `#[repr(Btf)]`.
 The following macros are added under `core::btf`:
 
 ```rust
-core::btf::field_byte_offset!(Carrier, field.path) -> usize
-core::btf::field_byte_size!(Carrier, field.path) -> usize
-core::btf::field_exists!(Carrier, field.path) -> bool
+core::btf::field_byte_offset!(Carrier, field.path) -> Option<usize>
+core::btf::field_byte_size!(Carrier, field.path) -> Option<usize>
 ```
 
 `Carrier` is the root local Rust type whose BTF graph describes the access.
@@ -295,8 +271,6 @@ The macros have the following meanings:
   complete field path from the root carrier.
 * `field_byte_size!(Carrier, field.path)` returns the byte size of the terminal
   field.
-* `field_exists!(Carrier, field.path)` returns whether the complete field path
-  exists in the target BTF type.
 
 These macros do not perform memory access. They are metadata queries and do not
 require the caller to uphold memory-safety invariants.
--- a/text/0000-btf-relocations.md
+++ b/text/0000-btf-relocations.md
@@ -331,10 +331,10 @@ targets in future.
 The compiler could try to make `task.pid` on selected types emit relocatable
 accesses automatically. This is what Clang currently does for CO-RE field
 accesses: ordinary C field projection can be preserved and lowered to BTF
-relocations. This is attractive ergonomically, but it requires a more intrusive
-change to MIR and the design of a proper abstract machine with operational
-semantics. It also has some similarities to the [`Sized` hierarchy
-RFC][sized-hierarchy], which has not yet been accepted.
+relocations. This is attractive ergonomically, but it requires a more
+intrusive change to MIR and the design of a proper operational semantics for
+these relocatable accesses. It also has some similarities to the [`Sized`
+hierarchy RFC][sized-hierarchy], which has not yet been accepted.
 
 This overlaps with the accepted [Field Projections project goal][field-projections],
 which is exploring virtual places as a general mechanism for custom field
--- a/text/0000-btf-relocations.md
+++ b/text/0000-btf-relocations.md
@@ -12,8 +12,8 @@ relocations based on the [BPF Type Format (BTF)][btf]. The feature introduces a
 be queried through BTF-aware operations, and adds BTF-aware macros for
 accessing the fields:
 
-* `core::btf::field_byte_offset!`
-* `core::btf::field_byte_size!`
+* `core::arch::bpf::field_byte_offset!`
+* `core::arch::bpf::field_byte_size!`
 
 The user-facing feature is gated by `#![feature(btf_relocations)]`.
 
@@ -115,17 +115,17 @@ pub struct task_struct {
 impl task_struct {
     #[inline]
     pub fn pid_offset(&self) -> Option<usize> {
-        core::btf::field_byte_offset!(task_struct, pid)
+        core::arch::bpf::field_byte_offset!(task_struct, pid)
     }
 
     #[inline]
     pub fn pid_size(&self) -> Option<usize> {
-        core::btf::field_byte_size!(task_struct, pid)
+        core::arch::bpf::field_byte_size!(task_struct, pid)
     }
 
     #[inline]
     pub fn pid(&self) -> Option<&i32> {
-        core::btf::field_byte_offset!(task_struct, pid).map(|offset| {
+        core::arch::bpf::field_byte_offset!(task_struct, pid).map(|offset| {
             let ptr = self as *const task_struct as *const u8;
 
             // SAFETY: the BTF relocation says that `se.vruntime` exists in the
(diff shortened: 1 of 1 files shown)


text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+#![feature(btf_relocations)]
+
+#[repr(Btf)]
+pub struct task_struct {
+    pub pid: i32,
+}
+
+fn pid(task: &task_struct) -> i32 {
+    task.pid
+    // error: cannot access fields of a `#[repr(Btf)]` type directly
```
workingjubilee: I would rather we do not use another `repr` that has the same "feature" as `repr(packed)`, for which we seem to *still* be encountering new issues: https://github.com/rust-lang/rust/issues/157011

I would prefer that we find some new way to express it syntactically that makes it harder for us to make more such mistakes in the future. Essentially, because these are sort of an Alien Being relative to normal structs, maybe they should not use the syntax of normal structs at all?

**But** I also wish to acknowledge that I do not have any objection to the fundamental *behavior* proposed, so this should not block any initial experimentation. It is more of a yet-to-be-resolved-concern type of deal. Unanswered question? "Can anyone think of something better?"

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+```rust
+task.pid
+```
+
+The `offset_of!` macro is also rejected when any container in the queried path is
+a `#[repr(Btf)]` ADT.
+
+These restrictions avoid silently producing non-relocatable code for operations
+that appear to query a relocatable type. Code that genuinely wants a normal
+non-relocatable Rust type should not use `#[repr(Btf)]`.
```
workingjubilee: It isn't clear to me that this achieves its intended goal? Or I don't understand the goal?

The current proposal includes a "compile-time fallback" to normal layout behavior, several lines back.

```md
On targets or backends without BTF
relocation support, they fall back to the current compilation unit's ordinary
layout information.
```

This is a fallback path that is sometimes permitted based on some variable that can differ based on the circumstances of the compilation. So if we do not trust the compiler to correctly handle the matter, it can always choose a "wrong" path and silently produce directly non-relocatable accesses *despite* using all these special macros. According to my understanding, we have gone to some length to supposedly prevent silent failure via introducing syntactic noise, and then allowed for silent failure?

Thus we must trust the compiler anyways, right? So what are we trying to guard against?

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+```
+
+`#[repr(Btf)]` is accepted on structs and unions. It is rejected on other item
+kinds.
+
+A `#[repr(Btf)]` type uses C-compatible field layout. In compiler terms,
+`repr(Btf)` implies the layout constraints of `repr(C)` and also marks the type
+as BTF-relocatable. This gives the backend stable field ordering and offsets for
+the compile-time fallback while preserving a distinct marker for type checking
+and codegen.
```
workingjubilee: Is this intended to use **actually compatible with C** layout or is it intended to simply use a linear layout? See RFC #3845 for a description of that problem and a proposed solution, which is orthogonal to this RFC except insofar as it implies the "what do you really mean?" question.

AFAIK the platforms that this relocatable-fields proposal is relevant for are ones where the notional `repr(C_actual)` and `repr(ordered_fields)` are identical... ish... but that sort of statement is prone to breaking down in the face of new discoveries, so we might as well get ahead of that.

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+    #[inline]
+    pub fn pid(&self) -> Option<&i32> {
+        self.has_pid().then(|| {
+            let offset = core::btf::field_byte_offset!(task_struct, pid);
+            let ptr = self as *const task_struct as *const u8;
+
+            // SAFETY: the BTF relocation says that `se.vruntime` exists in the
+            // target layout, and the returned offset is relative to `task_struct`.
+            Some(unsafe { &*(ptr.add(offset) as *const i32) })
+        })
```
workingjubilee, line 134: So a typo in a field name can result in simply turning this into a `None` and never taking the "active" code path, right? Since the relocation is functionally a set of runtime checks, and we would have to assume we take the field name for granted, and the entire reason to use these relocations is the struct definition *per se* is actually external to the program...?

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+relocation has been lost.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `btf_relocations` feature is for Rust code that models external BTF
+types, primarily Linux kernel types used by eBPF programs.
+
+A type that should participate in BTF relocation is written with
+`#[repr(Btf)]`:
```
workingjubilee: Something to consider with this as `repr`: Is this meant to be exclusive with other `repr` hints?
  - align?
  - packed?

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+}
+```
+
+`#[repr(Btf)]` is accepted on structs and unions. It is rejected on other item
+kinds.
+
+A `#[repr(Btf)]` type uses C-compatible field layout. In compiler terms,
+`repr(Btf)` implies the layout constraints of `repr(C)` and also marks the type
+as BTF-relocatable. This gives the backend stable field ordering and offsets for
+the compile-time fallback while preserving a distinct marker for type checking
```
workingjubilee: What do we mean here, or what do we achieve, by having a "compile-time fallback"?

There's an implicit alternative here: No fallback. What are we giving up if we choose that instead?

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+### Make ordinary field access relocatable
+[relocatable-field-access]: #relocatable-field-access
+
+The compiler could try to make `task.pid` on selected types emit relocatable
+accesses automatically. This is what Clang currently does for CO-RE field
+accesses: ordinary C field projection can be preserved and lowered to BTF
+relocations. This is attractive ergonomically, but it requires a more intrusive
+change to MIR and the design of a proper abstract machine with operational
+semantics. It also has some similarities to the [`Sized` hierarchy
+RFC][sized-hierarchy], which has not yet been accepted.
```
workingjubilee: This statement seems to be wearing its shirt backwards, so to speak. We do have an abstract machine, but rather we would need to fit these magic accesses into that semantics, which seems like a lot.
```suggestion
relocations. This is attractive ergonomically, but it requires a
more intrusive change to MIR and the design of a proper
operational semantics for these relocatable accesses.
It also has some similarities to the [`Sized` hierarchy
RFC][sized-hierarchy], which has not yet been accepted.
```

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+
+If the target, backend, or codegen mode cannot emit BTF field relocations, the
+field-info queries fall back to ordinary layout-computed values:
+
+* `field_byte_offset!` returns the complete field-path offset from the
+  current compilation layout.
+* `field_byte_size!` returns the field size from the current compilation
+  layout.
+* `field_exists!` returns `true` for a field path present in the current
+  compilation layout.
```
workingjubilee: This would just become a name resolution question, right? It doesn't have anything to actually do with layout.

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+
+The result of the LLVM intrinsic is an integer value. Offset and size queries
+are zero-extended to `usize`. Existence queries are compared against zero and
+return `bool`.
+
+If the target, backend, or codegen mode cannot emit BTF field relocations, the
+field-info queries fall back to ordinary layout-computed values:
+
+* `field_byte_offset!` returns the complete field-path offset from the
+  current compilation layout.
```
workingjubilee: It basically becomes just equal to `offset_of`, right? i.e. it can have a purely "library" implementation when BPF isn't involved (though mostly because the library already exposes such a language primitive...)

workingjubilee (commented):
Mostly minutiae that would benefit from being clarified.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/6787515f4abad9ffec4539b8c4df0cf5a1cee672..54a1461930e7b9a7d77d82fc26e6f03723265ace)*

text/0000-btf-relocations.md
vadorovsky: The reason I proposed the fallback mechanism is that most of Linux BPF projects consist of two parts - the actual BPF program running in the kernel, and the user-space application. The user-space part grabs the data produced by the BPF program from a BPF map (usually a ring buffer, or a hash map). BPF maps, at least in Linux, can store any type with C representation. It's not uncommon to store a BTF-relocatable kernel struct there.

But now you made me realize that this can be solved with `#[cfg_attr(target_arch = "bpf", ...)]`, so I agree with you, we could carry on without a fallback and emit a compilation error if BTF relocations are requested on non-BPF target. That absolutely makes sense from the "trust the compiler" perspective.

text/0000-btf-relocations.md
vadorovsky: I'm fine with replacing `repr` with something else, although the only (not sophisticated at all) idea I have now is an attribute (e.g. `#[relocatable]`, `#[btf_relocatable]`. Your other comment https://github.com/rust-lang/rfcs/pull/3966#discussion_r3486931443 is an another good argument against `repr`, and in general it should be possible to make a `repr(C)`, a `repr(Rust)` or `repr(align)` struct BTF-relocatable.

text/0000-btf-relocations.md
vadorovsky: I must admit, while writing the RFC, I was assuming actual compatibility with C layout, given that Linux kernel's BPF VM is written in C and writing Linux BPF programs means dealing with C types.

But after giving it more though, using a linear layout should be enough.

text/0000-btf-relocations.md
vadorovsky: Yes, that's right. Although instead of saying "is actually external", I would say that the struct definition is a _reference point_ for the runtime checks.

To be fair, I proposed a field projection (like in clang) in my pre-RFC, but given the complexity, I'm not pursuing it for now.

text/0000-btf-relocations.md
vadorovsky: Not really, actually I can't think of anything preventing BTF-relocatable types be `align` or `packed`, in theory. Seems like that's an another argument against `repr`.

workingjubilee: Part of the reason I'm not super-happy with `repr` in general (just... in Rust, and to be clear I'm not asking you to fix all of Rust) is indeed the sort of arbitrary composition rules, where sometimes a `repr` is exclusive (`repr(C)` and `repr(transparent)`, for instance) and sometimes it isn't (`repr(align(N))`, unless it's `repr(transparent)`...) so I don't think "another `repr` has kinda weird/arbitrary composition rules" is disqualifying, but it does mean we always have to think about the interactions.

text/0000-btf-relocations.md
workingjubilee: This only applies to the "fallback" case anyways, right?

text/0000-btf-relocations.md
workingjubilee: To me, it almost feels like it's some kind of exotic `extern struct`, so I would start thinking in that direction and ping-pong it a little with whoever I asked on T-lang to sponsor things.

workingjubilee:
Syntax questions and my nits aside, this will want a sponsor from T-lang for proceeding as a nightly experiment (which will later come back to re-informing this RFC). I have nominated it so that if you do not find a sponsor in your own time they eventually see it appear on their agenda and one of them can either take up the gauntlet or they can decline it. And this comment as explanation for why I nominated it, of course. This may be a very *eventual* eventually, so I do suggest finding a sponsor in your own time.

text/0000-btf-relocations.md
Mark-Simulacrum: I don't think repr(C) is exclusive? You can definitely have a `repr(C, align(...))` struct for example (there's even an example in the reference: https://doc.rust-lang.org/nightly/reference/type-layout.html#r-layout.repr.align-packed).

I *do* think that the RFC should spell out what should be allowed / disallowed.

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+}
+```
+
+The offset returned for a nested path is relative to the root carrier type,
+`task_struct`.
+
+On BPF targets with BTF-capable backend support and debug info enabled, these
+queries lower to CO-RE relocations. On targets or backends without BTF
+relocation support, they fall back to the current compilation unit's ordinary
+layout information.
```
Mark-Simulacrum: Generally speaking, most Rust features that are not very obviously codegen backend specific (e.g., `-Cllvm-args`), we try to avoid "backend support" as a caveat here. Is there a reason to caveat this in this way? Do we have expected use cases for this being a no-op when not supported?

Maybe this wants to say that if you're compiling repr(Btf) types for a non-bpf target it's a no-op in that case, but on bpf targets we mandate it does what's expected?

(What is the failure mode as a user of this feature if I accidentally don't have BTF relocation support when I expected it? Is it reliably a BPF verifier error?)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/6787515f4abad9ffec4539b8c4df0cf5a1cee672..54a1461930e7b9a7d77d82fc26e6f03723265ace)*

traviscross:
We talked about this in the lang call today.  We were interested to hear what the people working on @rust-lang/rust-for-linux think about this, as well as what those working on reflection (@oli-obk, @scottmcm, rust-lang/rust-project-goals#406) and projection (@BennoLossin, @dingxiangfei2009, @tmandry, rust-lang/rust-project-goals#390) think.

If this generally makes sense to people, I'll be happy to champion a lang experiment here.

cc @lqd @Nadrieril

Darksonn:
Also needs input from sized hierarchy side. cc @davidtwco 

[#t-lang > BTF relocations proposal @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/BTF.20relocations.20proposal/near/584609492)

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+This overlaps with the accepted [Field Projections project goal][field-projections],
+which is exploring virtual places as a general mechanism for custom field
+projection. This RFC deliberately does not depend on that work: it provides the
+low-level BTF field metadata queries needed for CO-RE today, while leaving a
+future field-projection-based ergonomic surface open.
+
+Providing the field-info queries proposed in this RFC does not rule out
+exploring this alternative in the future. On the contrary, Clang provides both
+explicit field-info builtins and field projection. It makes sense to treat these
+as separate RFCs.
```
Darksonn, line 335: I think this RFC could be written to make it more clear that we expect field access and `offset_of!` to work in the future. For example, the `offset_of!` section is phrased as if we could never support `offset_of!` with btf relocations.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/6787515f4abad9ffec4539b8c4df0cf5a1cee672..54a1461930e7b9a7d77d82fc26e6f03723265ace)*

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+
+`Carrier` is the root local Rust type whose BTF graph describes the access.
+The second argument is a dot-separated Rust field path starting from `Carrier`.
+The compiler type checks this path using the same field lookup rules as
+`offset_of!`, except that `#[repr(Btf)]` containers are accepted for these
+BTF-aware queries.
+
+The macros have the following meanings:
+
+* `field_byte_offset!(Carrier, field.path)` returns the byte offset of the
```
oli-obk, line 257: Considering this is UX wise equivalent to normal offset_of, could we just reuse the same intrinsic but make the carrier type just not impl a new marker trait constly, while all normal types implement it constly. This way you get an error trying to use it in const contexts which obviously can't load the relocation info, but codegen can just switch to a runtime op instead of a compile time op. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/6787515f4abad9ffec4539b8c4df0cf5a1cee672..54a1461930e7b9a7d77d82fc26e6f03723265ace)*

oli-obk: Ah I see it below

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,450 @@
...
+The main alternative is to emit BTF relocations in [bpf-linker][bpf-linker],
+which is a bitcode linker used exclusively for BPF targets. However, it
+prevents us from supporting ld type of linkers (e.g. binutils, lld) for BPF
+targets in future.
+
+### Use `offset_of!`
+
+`offset_of!` is intentionally a constant layout query. It does not preserve the
+field identity needed to emit a BTF relocation. Reusing it would either silently
+produce non-relocatable code or require changing the meaning of an existing
```
oli-obk: How is the new macro with the same API change anything? Why isn't that also silently producing wrong things when e.g. a field doesn't exist at all in the relocation table? 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/6787515f4abad9ffec4539b8c4df0cf5a1cee672..54a1461930e7b9a7d77d82fc26e6f03723265ace)*

text/0000-btf-relocations.md
workingjubilee: Yes, the problem with `repr`s in general is basically there's no obvious way to tell what is an "exclusive" repr attribute like `C`, `transparent`, and `Rust`, vs. the ones that are "non-exclusive" like `align`. Except even that general principle is not enough because `transparent` is exclusive with `align`, for instance. You just have to memorize the pile of rules, which gets more complex every time we add a new one.

text/0000-btf-relocations.md
vadorovsky: Good point, I'm going to rephrase this (while still making it clear it's out of scope of this RFC).

text/0000-btf-relocations.md
vadorovsky: I agree that `repr` is not ideal for the reasons mentioned.

text/0000-btf-relocations.md
vadorovsky: As agreed above, I'm going to remove the "compile-time fallback" from the RFC.

text/0000-btf-relocations.md
vadorovsky: Yes, I will reword this.

text/0000-btf-relocations.md
vadorovsky: > Generally speaking, most Rust features that are not very obviously codegen backend specific (e.g., `-Cllvm-args`), we try to avoid "backend support" as a caveat here. Is there a reason to caveat this in this way? Do we have expected use cases for this being a no-op when not supported?

As already agreed with @workingjubilee in https://github.com/rust-lang/rfcs/pull/3966#discussion_r3496687650, I'm going to ditch the idea of a no-op fallback in favor of a compiler error, when BTF relocations are requested, but can't be emitted.

That said, BTF relocations have to be explicitly supported by compiler backends. Both LLVM and GCC support them. But then Cranelift doesn't support BPF at all. For now there is no compiler backend that supports BPF target without supporting BTF relocations, but in theory such combination _could_ exist, and GCC used to be in such state in between version 10 (when BPF targets were introduced) and 12 (when relocations were introduced). If BPF support ever appears in Cranelift, I would also expect its features to be implemented gradually, not all at once. Therefore I don't think we can entirely avoid considering backend support.

> Maybe this wants to say that if you're compiling repr(Btf) types for a non-bpf target it's a no-op in that case, but on bpf targets we mandate it does what's expected?

I think what would make sense is emitting an error if `repr(Btf)` appears during compilation:

* To non-BPF target.
* To BPF target, but with a backend that does not support relocations. Again, no such backend exists _now_, but GCC _used to_ be like that, and there is non-zero chance Cranelift could fall into this territory in future.

> (What is the failure mode as a user of this feature if I accidentally don't have BTF relocation support when I expected it? Is it reliably a BPF verifier error?)

BTF relocations are done by user-space libraries that load the BPF programs (e.g. [Aya](https://github.com/aya-rs/aya), [cilium/ebpf](https://github.com/cilium/ebpf), [libbpf](https://github.com/libbpf/libbpf)), and they are supposed to perform them before the program is loaded into the kernel. I don't think there is any library that doesn't support BTF relocations, but in theoretical scenario where such library exists, and it loads a BPF program with a BTF-relocatable struct, where the layout carried by the program differs from the kernel, then the original program's bytecode would remain unchanged, the field access would be done according to the old offset, and the verifier would likely detect an invalid field access and reject the program.

text/0000-btf-relocations.md
vadorovsky: > How is the new macro with the same API change anything?

When I submitted the [pre-RFC](https://internals.rust-lang.org/t/pre-rfc-btf-relocations/24161), where I initially proposed emission of BTF relocations in regular field projections and `offset_of!` , the main feedback was that it goes against the principle that `Sized` types should have layout known at compile time, which is not the case with BTF-relocatable types.

Then the discussion went towards the hierarchy of Sized types RFC https://github.com/rust-lang/rfcs/pull/3729 which would eliminate this problem. But then we also agreed treating BTF-relocatable types as `!Sized` and allowing access to them only with dedicated macros/intrinsics could be a good way to keep the feature rolling without hard dependency on #3729.

> Why isn't that also silently producing wrong things when e.g. a field doesn't exist at all in the relocation table?

Not sure if I got this question right - are you asking about what `field_byte_offset!` and `field_byte_size!` macros produce when a field does not exist? That's a great question and actually made me realize that they should probably return an `Option`:

```
core::btf::field_byte_offset!(Carrier, field.path) -> Option<usize>
core::btf::field_byte_size!(Carrier, field.path) -> Option<usize>
```

vadorovsky:
> We talked about this in the lang call today. We were interested to hear what the people working on @rust-lang/rust-for-linux think about this, as well as what those working on reflection (@oli-obk, @scottmcm, [rust-lang/rust-project-goals#406](https://github.com/rust-lang/rust-project-goals/issues/406)) and projection (@BennoLossin, @dingxiangfei2009, @tmandry, [rust-lang/rust-project-goals#390](https://github.com/rust-lang/rust-project-goals/issues/390)) think.
> 
> If this generally makes sense to people, I'll be happy to champion a lang experiment here.
> 
> cc @lqd @Nadrieril

Thank you, @traviscross, for the discussion and for offering to champion this. I appreciate the lang team's interest. I'm going to modify the RFC according to the comments later today.

text/0000-btf-relocations.md
oli-obk: Jup, that makes sense now, thx

text/0000-btf-relocations.md
workingjubilee: Yeah, saying "the compiler must produce an error if this code generation feature is not supported" seems fine. Especially if failing to handle BTF relocations means the program is rejected as ill-formed by the BPF interpreter anyways? Then there doesn't seem to be much benefit to permitting it.

"The codegen backend is allowed to give you a useless program" is usually something we consider an occasional and avoidable side effect of things which have significant benefits for *all* programs, but require such permission. Here, there's nothing we're trading that off for.

And assuming we do use an attribute of some kind to indicate these semantics, the `cfg_attr` case does handle any fallback someone might deliberately want.

text/0000-btf-relocations.md
vadorovsky: > It's not uncommon to store a BTF-relocatable kernel struct there.

Actually, after giving it more thought, this should be avoided, and it's good that we will make it difficult with a compile error. BTF-relocatable struct has an unknown layout. Anything stored in BPF maps should have deterministic layout across kernel versions, so the user-space application behaves the same regardless of the kernel version. A good pattern here is always to copy specific fields from a relocatable type to a non-relocatable type before storing it in a map.

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,440 @@
...
+non-relocatable Rust type should not use `#[repr(Btf)]`.
+
+### Field-info macros
+
+The following macros are added under `core::btf`:
+
+```rust
+core::btf::field_byte_offset!(Carrier, field.path) -> usize
+core::btf::field_byte_size!(Carrier, field.path) -> usize
+core::btf::field_exists!(Carrier, field.path) -> bool
```
workingjubilee: If the first two are made `Option<usize>` as you mentioned in https://github.com/rust-lang/rfcs/pull/3966/changes#r3518749148 then, if the intent of the proposal is to be minimal, the `bool` macro can be dropped. This seems silly, of course.

Conversely, if the intent of the proposal is to have things that would be useful, then `btf::field_ref!` and `btf::field_mut!` for obtaining `Option<&T>` and `Option<&mut T>` might be helpful, assuming references to CO-RE types are sensible. Or just a `field_ptr` maybe?

I would assume such things wouldn't land in the very first implementation attempt, but before the experiment is concluded it may be useful to try implementing them. Then we learn of the drawbacks, can document them in this RFC. and have that for later reference when we consider more-direct support for `&relocatable_type.field`.

Again, a sort of non-blocking comment as "nah, that seems like more effort than necessary" also seems fine.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/037f71163b768372b28fa715b815b3fce0c9abb0..6292c33a0cda52563b7b28a4b4661d0f5bed1a6b)*

vadorovsky: > If the first two are made `Option<usize>` as you mentioned in https://github.com/rust-lang/rfcs/pull/3966/changes#r3518749148 then, if the intent of the proposal is to be minimal, the `bool` macro can be dropped. This seems silly, of course.

Good point, if we make two other macros always return `Option`, then having the `bool` macro is redundant and would incentivize writing suboptimal code with redundant checks. I'll drop it.

> Conversely, if the intent of the proposal is to have things that would be useful, then `btf::field_ref!` and `ptr::field_mut!` for obtaining `Option<&T>` and `Option<&mut T>` might be helpful, assuming references to CO-RE types are sensible. Or just a `field_ptr` maybe?

Given that these `field_ref!`/`field_mut!` macros, e.g.:

```rust
core::btf::field_ref!(task_struct, se.vruntime);
```

Would likely end up being an equivalent of:

```rust
if core::btf::field_exists!(task_struct, se.vruntime) {
    let offset = core::btf::field_byte_offset!(task_struct, se.vruntime);
    let ptr = self as *const task_struct as *const u8;
    Some(unsafe { &*(ptr.add(offset) as *const u64) })
} else {
    None
}
```

I don't think we could design the `field_*` macros in a way they produce any more optimal IR and bytecode. I think both snippets would lower to something like:

```
r0 = 0
r2 = 1  ; FIELD_EXISTS; emits a relocation, program loader can patch it
if r2 == 0 goto +4

r2 = 123  ; FIELD_BYTE_OFFSET; emits a relocation, program loader can patch it
r1 += r2
r0 = *(u64 *)(r1 + 0)

[...]
```

I would skip them for the sake of making the compiler work minimal, and instead focus on implementing such abstractions in Aya. In general, I don't expect any BPF program developers to use these intrinsics directly, unless someone wants to invent own library/framework for BPF development.

> I would assume such things wouldn't land in the very first implementation attempt, but before the experiment is concluded it may be useful to try implementing them. Then we learn of the drawbacks, can document them in this RFC. and have that for later reference when we consider more-direct support for `&relocatable_type.field`.

Not sure if implementing `field_ref!`/`field_mut!` will be helpful as a reference for the field projection support.

vadorovsky: The first point addressed in 6116056b8fc3f02ad044b762afb92697f42cb0f8

text/0000-btf-relocations.md
vadorovsky: Done 6116056b8fc3f02ad044b762afb92697f42cb0f8

text/0000-btf-relocations.md
vadorovsky: Done 97478ed575da9bd44c4bd67c0d7af49fd24d0884

text/0000-btf-relocations.md
workingjubilee: Oh yes. If they do not actually seem helpful for your given use-case then that is certainly a reason not to bother!

text/0000-btf-relocations.md
vadorovsky: Actually the entire bullet point is gone since 6116056b8fc3f02ad044b762afb92697f42cb0f8

text/0000-btf-relocations.md
vadorovsky: Done 0a3bcf74ce8c844626d60e8278130362d6ea6aca

text/0000-btf-relocations.md
vadorovsky: For now I went ahead with `#[btf_relocatable]` attribute: 09663bd1e4285a4931f5c782af7ee7ca56d51f74

text/0000-btf-relocations.md
vadorovsky: This should be addressed - I removed both the fallback mechanism (03549d3ad5888c1dee7b24a4607d3a5f45967be4) and `#[repr(Btf)]` (09663bd1e4285a4931f5c782af7ee7ca56d51f74) from the RFC.

text/0000-btf-relocations.md
vadorovsky: > But after giving it more though, using a linear layout should be enough.

Actually not even linear layout will be needed with LLVM 23 - https://github.com/llvm/llvm-project/pull/205396 this change will make BTF and types with layout reordered by Rust compiler work just fine.

Besides, today we have an ugly hack in bpf-linker that makes it work:

https://github.com/aya-rs/bpf-linker/blob/46cbc2896f9ef36c79169f6beedd796c0a932002/src/llvm/di.rs#L130-L133

Of course this hack will go away with LLVM 23.

Let's conclude this topic that there is no restriction against using `#[btf_relocatable]` with different representations, including `#[repr(Rust)]`.

text/0000-btf-relocations.md
vadorovsky: I believe this is addressed by 09663bd1e4285a4931f5c782af7ee7ca56d51f74 and moving away from the `repr` idea entirely.

text/0000-btf-relocations.md
vadorovsky: Done 03549d3ad5888c1dee7b24a4607d3a5f45967be4

text/0000-btf-relocations.md
vadorovsky: The no-op fallback is removed from the RFC 03549d3ad5888c1dee7b24a4607d3a5f45967be4. Is there any unresolved question here, or something you would like me to clarify?

workingjubilee: This looks good!

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,420 @@
...
+#![feature(btf_relocations)]
+
+#[btf_relocatable]
+#[repr(C)]
+pub struct task_struct {
+    pub pid: i32,
+}
+
+const PID_OFFSET: usize = core::mem::offset_of!(task_struct, pid);
+// error: cannot use `offset_of!` with a `#[repr(Btf)]` type
```
workingjubilee:
```suggestion
// error: cannot use `offset_of!` with a `#[btf_relocatable]` type
```

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,420 @@
...
+
+#[btf_relocatable]
+#[repr(C)]
+pub struct task_struct {
+    pub pid: i32,
+}
+
+fn pid(task: &task_struct) -> i32 {
+    task.pid
+    // error: cannot access fields of a `#[repr(Btf)]` type directly
```
workingjubilee:
```suggestion
    // error: cannot access fields of a `#[btf_relocatable]` type directly
```

text/0000-btf-relocations.md
```diff
@@ -0,0 +1,420 @@
...
+
+`#[btf_relocatable]` is accepted on structs and unions. It is rejected on other
+item kinds.
+
+Direct field projection from a `#[btf_relocatable]` ADT is rejected. This
+includes projections reached through autoderef:
+
+```rust
+task.pid
+```
```
workingjubilee: I actually feel slightly unclear why this code example is here because it doesn't seem to illustrate the paragraph before it.

workingjubilee (commented):
Cool. I resolved my concerns where they seemed no longer useful to think about or seemed to be adequately encompassed into the RFC. Noticed some typos, or at least incomplete substitutions.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3966/037f71163b768372b28fa715b815b3fce0c9abb0..09663bd1e4285a4931f5c782af7ee7ca56d51f74)*

text/0000-btf-relocations.md
vadorovsky: This was supposed to be a code example of a disallowed field projection, but yes, probably unnecessary - it's already being explained multiple times before in the guide-level explanation. I removed it in ad18f899918127f2882c014700b39e992ae472b6.

text/0000-btf-relocations.md
vadorovsky: Sorry, fixed 639bdc605cfa278eb1b932da683f925fc5acac9a

text/0000-btf-relocations.md
vadorovsky: Done 639bdc605cfa278eb1b932da683f925fc5acac9a

text/0000-btf-relocations.md
lqd: > Then the discussion went towards the hierarchy of Sized types RFC #3729 which would eliminate this problem

Just checking, as I'm not familiar with how BTF-relocatable types are used in practice: it seems like the property of having a size that is not known at compile time but known to be constant at runtime (depending on the hopefully future differentiation between `const Sized` vs `Sized` types from a follow-up to #3729) would be enough for you to model field offsets on BTF-relocatable types that are only known at runtime, and you won't need to model additional, separate, properties (e.g. with your own traits), right?

vadorovsky: @lqd I'm so sorry for missing this question!

> Just checking, as I'm not familiar with how BTF-relocatable types are used in practice: it seems like the property of having a size that is not known at compile time but known to be constant at runtime (depending on the hopefully future differentiation between `const Sized` vs `Sized` types from a follow-up to #3729) would be enough for you to model field offsets on BTF-relocatable types that are only known at runtime and you won't need to model additional, separate, properties (e.g. with your own traits), right?

That's right, under the Sized hierarchy RFC, BTF-relocatable types could implement `SizeOfVal`, I don't see any need for any additional trait.

One open question, which would remain even after the Sized hierarchy is accepted, is how do we handle direct field accesses to fields that don't exist during runtime. Do we panic? If so, are we going to generalize the macros (e.g. `field_byte_offset! -> Option<usize>`), that I'm proposing in this RFC, to become a fallible way of accessing potentially non-existing fields, that would apply to all of `SizeOfVal` but not `Sized` types?

vadorovsky:
I believe I’ve addressed all the review feedback.

I have an implementation ready on my branch (https://github.com/vadorovsky/rust/tree/vad/btf-relocations). @traviscross, would you be comfortable championing this as a lang experiment, or do we need more feedback and explicit approvals? In case we can proceed, should I create a tracking issue, or should rust-lang/rust#158412 be converted into one?

traviscross:
I'll champion it.  Go ahead and create a tracking issue if you would.

nagisa:
Given that the long-term plan is to implement support for transparent operation with ordinary rust field projections, how terrible would it be to keep the `core::bpf` macros implemented but unstable until the field projections get implemented? Do these macros otherwise achieve anything that a plain field projection would not be able to provide functionality wise?

vadorovsky:
@traviscross

> I'll champion it. Go ahead and create a tracking issue if you would.

Thank you! I created the tracking issue and the libs-team ACP:

* https://github.com/rust-lang/rust/issues/160616
* https://github.com/rust-lang/libs-team/issues/846

I'm going to send a PR later today.

@nagisa

> Given that the long-term plan is to implement support for transparent operation with ordinary rust field projections, how terrible would it be to keep the `core::bpf` macros implemented but unstable until the field projections get implemented?

It probably depends on how long `Sized` hierarchy and the field projections take to implement. BPF targets are currently in Tier 3, but not because of lack of BTF relocations, but for another reasons:

* They don't support >5 arguments (until LLVM 22, LLVM 23 fixes that).
* They don't support aggregate returns (until LLVM 22, LLVM 23 fixes that).
* They use a custom bitcode linker (bpf-linker). The two previous points are a blocker for using an ld-flavor of a linker (it's impossible to compile a full object file for `libcore` with the limitations above).

LLVM 23 support just got merged and it unblocks all of these points. The first two points should be fixed automatically, the last one will require some work on supporting linking with binutils (so far the only ld-type of linker supporting BPF). It would be also great to add BPF support to lld, mold, wild etc.

If the work on ld linking gets done faster than dynamic `Sized` and BTF field projection, I would be inclined to try moving the BPF targets at least to Tier 2. Tier 3 means nightly-only (with rustup), and it's a massive blocker in Aya's adoption. There are already some popular projects using it (e.g. mitmproxy) and some Linux distributions (e.g. Arch, Gentoo, NixOS) are enabling BPF targets in their Rust stable packages, even though Rust upstream considers BPF targets unstable.

I could imagine, that in situation when:

* Ld flavor of linkers are supported.
* `core::btf` macros are widely used.
* `Sized` hierarchy and field projections are not done.

I would probably be in favor of trying to stabilize `core::btf` macros, together with moving the BPF target to Tier 2, to unblock Rust BPF projects that want to move to Rust stable, and have to use BTF relocations in some way (pretty much all of BPF that's not limited to networking). Just my opinion.

Of course it would be great to avoid such scenario by making all three done in similar time. I'll try to see how I can be helpful in `Sized` hierarchy, and I'm already experimenting on linkers support.

> Do these macros otherwise achieve anything that a plain field projection would not be able to provide functionality wise?

Yes, there is one difference between macros and a plain field projection. Macros return an `Option`, and therefore return `None` if the requested field does not exist during runtime. Field projection would most likely panic in such case (up for decision in an another RFC, once `Sized` hierarchy is done).

It's not uncommon for Linux kernel code to remove, rename or add fields in important structs, and for BPF programs to want to handle layouts of such structs for different kernel versions. Even if the field projection is implemented, the macros will be useful for handling such cases. Also, some kernel structs have fields that depend on kernel configuration.

For example, there is a field called `cg_dead_lnode` in `task_struct` that depends on `CONFIG_PREEMPT_RT` being enabled:

https://elixir.bootlin.com/linux/v7.1.5/source/include/linux/sched.h#L1329-L1331

Writing a BPF program that would panic on absence of this field wouldn't be great, and it would be preferable to write code like:

```rust
// This would be generated by a proc-macro in Aya.
impl task_struct {
    pub fn maybe_cg_dead_lnode(&self) -> Option<&llist_node> {
        core::btf::field_byte_offset!(task_struct, cg_dead_lnode).map(|offset| {
            let ptr = self as *const task_struct as *const u8;
            Some(unsafe { &*(ptr.add(offset) as *const llist_node) })
        })
    }
}

// User's code that doesn't panic.
if let Some(cg_dead_lnode) = task_struct.maybe_cg_dead_lnode() {
    // handle `PREEMPT_RT`
} else {
    // no `PREEMPT_RT`
}
```

nagisa:
A'ight. In that case the only other suggestion I have is that this module should be under `core::arch::bpf`, like all other arch-specific stuff, rather than at the top level in core.

vadorovsky:
> A'ight. In that case the only other suggestion I have is that this module should be under `core::arch::bpf`, like all other arch-specific stuff, rather than at the top level in core.

Good idea, applied in 971eeca07256769ef54dfca78778611ff44870e6.

tamird:
Apologies for showing up late here but I want to address one thing

> Yes, there is one difference between macros and a plain field projection. Macros return an Option, and therefore return None if the requested field does not exist during runtime. Field projection would most likely panic in such case (up for decision in an another RFC, once Sized hierarchy is done).

I think this is not quite correct -- the way that BPF relocations work today in libbpf is more nuanced than this; there are ways to explicitly check for the presence of a particular struct after relocations, and this mechanism is used for generalized feature detection. In other words, a common pattern is to check for the presence of a field once and then access it and other fields whose presence is implied by its presence _without checking_. In case of invalid access the loader rewrites those field access into poison instructions and the program will fail to verify.

So I don't think we should bake Option into these facilities. That would be purpose-defeating.

vadorovsky:
> Apologies for showing up late here but I want to address one thing
> 
> > Yes, there is one difference between macros and a plain field projection. Macros return an Option, and therefore return None if the requested field does not exist during runtime. Field projection would most likely panic in such case (up for decision in an another RFC, once Sized hierarchy is done).
> 
> I think this is not quite correct -- the way that BPF relocations work today in libbpf is more nuanced than this; there are ways to explicitly check for the presence of a particular struct after relocations, and this mechanism is used for generalized feature detection. In other words, a common pattern is to check for the presence of a field once and then access it and other fields whose presence is implied by its presence _without checking_. In case of invalid access the loader rewrites those field access into poison instructions and the program will fail to verify.
> 

To be fair, I diverged from what clang/libbpf does intentionally, assuming that returning an `Option` would be more "rusty" (making it explicit, that the field "maybe exists").

But after giving it more thought, I think you are right. One downside of forcing an `Option` unconditionally is that it forces every program to have a branch (`if`, `match`, `?` etc.) on any field access. That adds to the complexity of the bytecode. Users who are either fine with their program being rejected while encountering non-existing fields, or are fine with doing one "feature check" and then proceeding with unchecked accesses, might want to do so to minimize their program size.

We already trust the BPF program loaders (Aya, libbpf etc.) to do the relocation work correctly, so we might as well trust them to poison the instructions and refuse to load programs that request a non-existent field in unchecked manner. After all, the entire CO-RE feature can't be done without trusting the loaders.

Therefore, we could redefine the macros as:

```rust
core::arch::bpf::field_exists!(Carrier, field.path) -> bool
core::arch::bpf::field_byte_offset!(Carrier, field.path) -> usize
core::arch::bpf::field_byte_size!(Carrier, field.path) -> usize
```

Given these, and given a struct like:

```rust
#[btf_relocatable]
#[repr(C)]
pub struct sched_entity {
    pub deadline: u64,
    pub min_vruntime: u64,
}

#[btf_relocatable]
#[repr(C)]
pub struct task_struct {
    pub sched: sched_entity,
}
```

One could build checked access methods, similar to the ones already mentioned in the RFC:

```rust
impl task_struct {
    #[inline]
    pub fn maybe_sched_deadline(&self) -> Option<&u64> {
        if core::arch::bpf::field_exists!(task_struct, se.deadline) {
            let ptr = self as *const task_struct as *const u8;
            let offset = core::arch::bpf::field_byte_offset!(task_struct, se.deadline);
            Some(unsafe { &*(ptr.add(offset) as *const u64) })
        } else {
            None
        }
    }

    #[inline]
    pub fn maybe_sched_min_vruntime(&self) -> Option<&u64> {
        if core::arch::bpf::field_exists!(task_struct, se.min_vruntime) {
            let ptr = self as *const task_struct as *const u8;
            let offset = core::arch::bpf::field_byte_offset!(task_struct, se.min_vruntime);
            Some(unsafe { &*(ptr.add(offset) as *const u64) })
        } else {
            None
        }
    }
}
```

As well as separate check methods, and unchecked access methods:

```rust
impl task_struct {
    #[inline]
    pub fn has_sched(&self) -> bool {
        core::arch::bpf::field_exists!(task_struct, se)
    }

    #[inline]
    pub fn has_sched_deadline(&self) -> bool {
        core::arch::bpf::field_exists!(task_struct, se.deadline)
    }

    #[inline]
    pub fn has_sched_min_vruntime(&self) -> bool {
        core::arch::bpf::field_exists!(task_struct, se.min_vruntime)
    }

    #[inline]
    pub fn deadline(&self) -> &u64 {
        let ptr = self as *const task_struct as *const u8;
        let offset = core::arch::bpf::field_byte_offset!(task_struct, se.deadline);
        Some(unsafe { &*(ptr.add(offset) as *const u64) })
    }

    #[inline]
    pub fn min_vruntime(&self) -> &u64 {
        let ptr = self as *const task_struct as *const u8;
        let offset = core::arch::bpf::field_byte_offset!(task_struct, se.min_vruntime);
        Some(unsafe { &*(ptr.add(offset) as *const u64) })
    }
}
```

Where they could be used either without any checks at all, where we rely on loader rejecting the program, in case either `sched_entity` or any of the requested fields don't exist:

```rust
let deadline = task.deadline();
let min_vruntime = task.min_vruntime();
```

Or with the "one check" approach that @tamird mentioned, that checks for a certain kernel feature once, but then avoids adding more branches in the bytecode for every field:

```rust
if task.has_sched() {
    let deadline = task.deadline();
    let min_vruntime = task.min_vruntime();
    [...]
}
```

In case any field does not exist, the program still will be rejected.

@oli-obk @lqd Does it sound good to you? I tried to check whether the sized hierarchy RFC has an answer about what are we planning to do in the event of non-existent fields during runtime, but I couldn't find any. I believe, that makes this RFC the first one bringing such an answer.

And we need to decide whether an answer is - *we trust the target to reject the code with unchecked accesses* and *we trust the target to handle conditional field access correctly*.

I'll update the RFC, issues and my implementation once confirmed everyone is happy with that approach.

nagisa:
Does `unwrap_unchecked` of the option not produce the "ideal" codegen here?

vadorovsky:
> Does `unwrap_unchecked` of the option not produce the "ideal" codegen here?

It most likely does. I will give it a try, if it really does, then I would keep the RFC as it is (returning `Option`).

We could hide `unwrap_unchecked` in unchecked methods generated by Aya macros, and consider these methods safe with an explanation that `unwrap_unchecked` on BTF-relocatable type fields is safe and doesn't cause UB.

vadorovsky:
OK, I have an answer now.

> Does `unwrap_unchecked` of the option not produce the "ideal" codegen here?

No, not at all.

This Rust code:

```rust
#![feature(allow_internal_unstable, btf_relocations, decl_macro, no_core)]
#![no_core]
#![no_std]
#![no_main]

extern crate minicore;
use minicore::{*, Option::*};

#[allow_internal_unstable(builtin_syntax)]
pub macro maybe_field_byte_offset($Container:ty, $($fields:expr)+ $(,)?) {{
    if builtin # btf_field_exists($Container, $($fields)+) {
        ::minicore::Option::Some(builtin # btf_field_byte_offset($Container, $($fields)+))
    } else {
        ::minicore::Option::None
    }
}}


#[allow_internal_unstable(builtin_syntax)]
pub macro field_byte_offset($Container:ty, $($fields:expr)+ $(,)?) {{
     builtin # btf_field_byte_offset($Container, $($fields)+)
}}

#[btf_relocatable]
#[repr(C)]
pub struct Inner {
    pub x: u32,
    pub y: u64,
}

#[btf_relocatable]
#[repr(C)]
pub union Payload {
    pub word: u64,
    pub half: u32,
}

#[btf_relocatable]
#[repr(C)]
pub struct Outer {
    pub pad: u32,
    pub inner: Inner,
    pub payload: Payload,
}

#[unsafe(no_mangle)]
pub fn field_offset() -> usize {
    field_byte_offset!(Outer, inner)
}

#[unsafe(no_mangle)]
pub fn field_offset_option() -> Option<usize> {
    maybe_field_byte_offset!(Outer, inner)
}

#[inline(always)]
fn unwrap_unchecked<T>(o: Option<T>) -> T {
    match o {
        Some(o) => o,
        None => unsafe { unreachable() },
    }
}

#[unsafe(no_mangle)]
pub fn field_offset_unwrap_unchecked() -> usize {
    unwrap_unchecked(maybe_field_byte_offset!(Outer, inner))
}
```

Produces the following function definitions in IR:

```llvm
; Function Attrs: nounwind
define i64 @field_offset() unnamed_addr #1 !dbg !24 !guid !29 {
start:
  %0 = getelementptr inbounds %Outer, ptr null, i32 0, i32 2
  %1 = load i32, ptr @"llvm.Outer:0:8$0:1", align 4
  %2 = call i32 @llvm.bpf.passthrough.i32.i32(i32 0, i32 %1)
  %_0 = zext i32 %2 to i64, !dbg !30
  ret i64 %_0, !dbg !31
}
```

```llvm
; Function Attrs: nounwind
define [2 x i64] @field_offset_option() unnamed_addr #1 !dbg !32 !guid !49 {
start:
  %_0 = alloca [16 x i8], align 8
  %0 = getelementptr inbounds %Outer, ptr null, i32 0, i32 2
  %1 = load i32, ptr @"llvm.Outer:2:1$0:1", align 4
  %2 = call i32 @llvm.bpf.passthrough.i32.i32(i32 1, i32 %1)
  %_1 = icmp ne i32 %2, 0, !dbg !50
  br i1 %_1, label %bb3, label %bb6, !dbg !50

bb6:                                              ; preds = %start
  store i64 0, ptr %_0, align 8, !dbg !51
  br label %bb7, !dbg !52

bb3:                                              ; preds = %start
  %3 = getelementptr inbounds %Outer, ptr null, i32 0, i32 2
  %4 = load i32, ptr @"llvm.Outer:0:8$0:1", align 4
  %5 = call i32 @llvm.bpf.passthrough.i32.i32(i32 2, i32 %4)
  %_4 = zext i32 %5 to i64, !dbg !53
  %6 = getelementptr inbounds i8, ptr %_0, i64 8, !dbg !54
  store i64 %_4, ptr %6, align 8, !dbg !54
  store i64 1, ptr %_0, align 8, !dbg !54
  br label %bb7, !dbg !52

bb7:                                              ; preds = %bb3, %bb6
  %7 = load [2 x i64], ptr %_0, align 8, !dbg !55
  ret [2 x i64] %7, !dbg !55
}
```

```llvm
; Function Attrs: nounwind
define i64 @field_offset_unwrap_unchecked() unnamed_addr #1 !dbg !56 !guid !57 {
start:
  %o.dbg.spill.i = alloca [8 x i8], align 8
  %o.i = alloca [16 x i8], align 8
  %_1 = alloca [16 x i8], align 8
  %0 = getelementptr inbounds %Outer, ptr null, i32 0, i32 2
  %1 = load i32, ptr @"llvm.Outer:2:1$0:1", align 4
  %2 = call i32 @llvm.bpf.passthrough.i32.i32(i32 4, i32 %1)
  %_2 = icmp ne i32 %2, 0, !dbg !58
  br i1 %_2, label %bb3, label %bb6, !dbg !58

bb6:                                              ; preds = %start
  store i64 0, ptr %_1, align 8, !dbg !59
  br label %bb7, !dbg !60

bb3:                                              ; preds = %start
  %3 = getelementptr inbounds %Outer, ptr null, i32 0, i32 2
  %4 = load i32, ptr @"llvm.Outer:0:8$0:1", align 4
  %5 = call i32 @llvm.bpf.passthrough.i32.i32(i32 3, i32 %4)
  %_5 = zext i32 %5 to i64, !dbg !61
  %6 = getelementptr inbounds i8, ptr %_1, i64 8, !dbg !62
  store i64 %_5, ptr %6, align 8, !dbg !62
  store i64 1, ptr %_1, align 8, !dbg !62
  br label %bb7, !dbg !60

bb7:                                              ; preds = %bb3, %bb6
  %7 = load i64, ptr %_1, align 8, !dbg !63
  %8 = getelementptr inbounds i8, ptr %_1, i64 8, !dbg !63
  %9 = load i64, ptr %8, align 8, !dbg !63
  store i64 %7, ptr %o.i, align 8
  %10 = getelementptr inbounds i8, ptr %o.i, i64 8
  store i64 %9, ptr %10, align 8
    #dbg_declare(ptr %o.i, !64, !DIExpression(), !71)
  %_2.i = load i64, ptr %o.i, align 8, !dbg !73
  %11 = getelementptr inbounds i8, ptr %o.i, i64 8, !dbg !73
  %12 = load i64, ptr %11, align 8, !dbg !73
  %13 = icmp eq i64 %_2.i, 1, !dbg !74
  br i1 %13, label %_RINvCsgHzIm8B26qX_15btf_relocations16unwrap_uncheckedjEB2_.exit, label %bb1.i, !dbg !74

bb1.i:                                            ; preds = %bb7
  unreachable, !dbg !75

_RINvCsgHzIm8B26qX_15btf_relocations16unwrap_uncheckedjEB2_.exit: ; preds = %bb7
  %14 = getelementptr inbounds i8, ptr %o.i, i64 8, !dbg !76
  %o1.i = load i64, ptr %14, align 8, !dbg !76
  store i64 %o1.i, ptr %o.dbg.spill.i, align 8, !dbg !76
    #dbg_declare(ptr %o.dbg.spill.i, !69, !DIExpression(), !77)
  ret i64 %o1.i, !dbg !78
}
```

`field_offset_unwrap_unchecked` is not even remotely close to being equivalent to `field_offset`, quite the opposite, it ends up being the most bloated.

I will go ahead with what I proposed in https://github.com/rust-lang/rfcs/pull/3966#issuecomment-5226567210 - I'll remove `Option` and make the macros return `usize` / `bool`.

programmerjake:
> `field_offset_unwrap_unchecked` is not even remotely close to being equivalent to `field_offset`, quite the opposite, it ends up being the most bloated.

it looks like you didn't turn optimizations on, `unwrap_unchecked` (and most other Rust code) generally requires you to turn optimizations on for them to be useful in simplifying generated code.

vadorovsky:
My bad! After turning on optimizations, the IR of `field_offset` and `field_offset_unwrap_unchecked` is still not identical:

```llvm
; Function Attrs: nofree nosync nounwind memory(read, argmem: none, inaccessiblemem: none, target_mem: none)
define noundef range(i64 0, 4294967296) i64 @field_offset() unnamed_addr #1 !dbg !24 !guid !29 {
start:
  %0 = load i32, ptr @"llvm.Outer:0:8$0:1", align 4
  %1 = tail call i32 @llvm.bpf.passthrough.i32.i32(i32 0, i32 %0)
  %_0 = zext i32 %1 to i64, !dbg !30
  ret i64 %_0, !dbg !31
}
```

```llvm
; Function Attrs: nofree nosync nounwind memory(read, argmem: none, inaccessiblemem: write, target_mem: none)
define noundef range(i64 0, 4294967296) i64 @field_offset_unwrap_unchecked() unnamed_addr #2 !dbg !55 !guid !56 {
start:
  %0 = load i32, ptr @"llvm.Outer:2:1$0:1", align 4
  %1 = tail call i32 @llvm.bpf.passthrough.i32.i32(i32 3, i32 %0)
  %2 = load i32, ptr @"llvm.Outer:0:8$0:1", align 4
  %3 = tail call i32 @llvm.bpf.passthrough.i32.i32(i32 4, i32 %2)
  %_5 = zext i32 %3 to i64, !dbg !57
  ret i64 %_5, !dbg !58
}
```

But the BPF ASM produced by LLVM seems identical:

```asm
	.section	.text.field_offset,"ax",@progbits
	.globl	field_offset
	.p2align	3
	.type	field_offset,@function
field_offset:
.Lfunc_begin0:
	.cfi_sections .debug_frame
	.cfi_startproc
	.file	1 "/home/vad/src/rust" "tests/codegen-llvm/btf-relocations.rs"
	.loc	1 54 2 prologue_end
.Ltmp0:
.Ltmp1:
	r0 = 8
	exit
.Ltmp2:
.Ltmp3:
.Lfunc_end0:
	.size	field_offset, .Lfunc_end0-field_offset
	.cfi_endproc
```

```asm
	.section	.text.field_offset_unwrap_unchecked,"ax",@progbits
	.globl	field_offset_unwrap_unchecked
	.p2align	3
	.type	field_offset_unwrap_unchecked,@function
field_offset_unwrap_unchecked:
.Lfunc_begin2:
	.cfi_startproc
	.loc	1 72 2 prologue_end
.Ltmp12:
.Ltmp13:
	r0 = 8
	exit
.Ltmp14:
.Ltmp15:
.Lfunc_end2:
	.size	field_offset_unwrap_unchecked, .Lfunc_end2-field_offset_unwrap_unchecked
	.cfi_endproc
```

So I guess it's fine. If you prefer to keep `Option` in the API, I can keep the RFC as it is, and we can make use of `Option::unwrap_unchecked` for implementing unchecked getters in Aya.

tamird:
fwiw I would prefer to avoid unwrap_unchecked.

PR #3968, opened 2026-06-04
RFC for convenient, explicit closure capture using `move($expr)` expressions

nikomatsakis:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3968)*

Add `move($expr)` syntax inside closures, async blocks, and generators. A `move($expr)` evaluates the expression at closure-creation time and captures the result by value. This gives precise control over what a closure captures and when, without needing temporary variables outside the closure.

A prototype implementation is available thanks to @TaKO8Ki 

**Desired feedback:**

* [Choice of syntax](https://github.com/rust-lang/rfcs/pull/3968#pullrequestreview-4435393221)

> [!IMPORTANT]
>

[Rendered](https://github.com/nikomatsakis/rfcs/blob/move-expressions/text/0000-move-expressions.md)

### DIFF
--- a/text/0000-move-expressions.md
+++ b/text/0000-move-expressions.md
@@ -1,0 +1,444 @@
+- Feature Name: `move_expressions`
+- Start Date: 2026-05-14
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#155050](https://github.com/rust-lang/rust/issues/155050)
+
+## Summary
+[summary]: #summary
+
+Add `move($expr)` syntax inside closures, async blocks, and generators. A `move($expr)` evaluates the expression at closure-creation time and captures the result by value. This gives precise control over what a closure captures and when, without needing temporary variables outside the closure.
+
+## Motivation
+[motivation]: #motivation
+
+Closures and futures in Rust use an automatic capture mechanism that detects the places that the closure/future body uses from the surrounding environment and decides automatically how to capture them (e.g., by reference, by value). However, it is frequently useful to capture values in other ways; one common example is wishing to capture the *clone* of a value. Unfortunately Rust's current closure syntax provides no ergonomic way to manage this.
+
+A common pattern for capturing clones, drawn from [a blog post by the Dioxus team][dioxus-blog], is to introduce "dummy variables" into the surrounding scope:
+
+```rust
+let _some_a = self.some_a.clone();
+let _some_b = self.some_b.clone();
+let _some_c = self.some_c.clone();
+tokio::task::spawn(async move {
+    do_something_else_with(_some_a, _some_b, _some_c)
+});
+```
+
+[dioxus-blog]: https://dioxus.notion.site/Dioxus-Labs-High-level-Rust-5fe1f1c9c8334815ad488410d948f05e
+
+The problems with this pattern:
+
+- The temporary variables (`_some_a`, etc.) exist only to shuttle data into the closure. They add noise and separate the clone from its point of use.
+- The `async move` captures *everything* by move, which means later code can't use `self.some_a` even though the intent was only to move the clones.
+- When closures are nested or have many captures, the pre-closure variable block grows unwieldy and obscures the actual logic.
+
+Another option is to introduce let variables into a block:
+
+```rust
+tokio::task::spawn({
+    let some_a = self.some_a.clone();
+    let some_b = self.some_b.clone();
+    let some_c = self.some_c.clone();
+    async move {
+        do_something_else_with(some_a, some_b, some_c)
+    }
+});
+```
+
+This pattern at least avoids the need to introduce "dummy names" like `_some_a`, but it still has many of the same downsides.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` will execute when the closure is created and then moved into a temporary captured by the closure; `move($expr)` is then replaced with a reference to this temporary. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
+
+### `move($expr)` gives explicit control over captures
+
+Within a closure, async block, or generator, you can write `move($expr)` to evaluate an expression at closure-creation time and capture the result by value:
+
+```rust
+|| {
+    move(some_vec).push(22)
+}
+```
+
+This moves `some_vec` into the closure, regardless of how it's used in the body. The expression inside `move(...)` is evaluated when the closure is *created*, not when it's *called*.
+
+Any expression is valid inside `move(...)`:
+
+```rust
+|| {
+    // Move a clone
+    move(data.clone()).process()
+}
+
+|| {
+    // Move a reference
+    let len = move(&big_vec).len();
+}
+
+|| {
+    // Move the result of a function call
+    move(compute_config()).apply()
+}
+```
+
+### A common pattern: listing captures at the top
+
+When you want to be fully explicit about what a closure captures, you can list all captures at the top of the body:
+
+```rust
+|| {
+    let vec = move(input.vec);
+    let data = move(&cx.data);
+    let output_tx = move(output_tx);
+
+    process(&vec, &mut output_tx, data)
+}
+```
+
+### `move ||` as shorthand
+
+The `move` keyword on a closure changes the default so that every captured variable is moved in. You can combine this with `move($expr)` for fine-grained control:
+
+```rust
+move || {
+    process(&input.vec, &mut output_tx, move(&cx.data))
+    //       ---------       ---------       --------
+    //           |               |               |
+    //       moved in by     moved in by    only a reference
+    //       `move ||`       `move ||`      is moved in
+}
+```
+
+Here `move(&cx.data)` creates a reference to `cx.data` at closure-creation time and captures that reference. The closure borrows `cx.data` even though it's a `move` closure.
+
+### Async blocks and generators
+
+`move($expr)` works the same way in async blocks and generators:
+
+```rust
+tokio::task::spawn(async {
+    send_data(move(tx.clone())).await;
+});
+```
+
+### Closures called multiple times (`FnMut`, `Fn`)
+
+When a closure is called multiple times, the captured value persists across calls. A `move($expr)` captures once at creation time:
+
+```rust
+data_source_iter
+    .inspect(|item| {
+        inspect_item(item, move(tx.clone()).clone())
+        //                      ----------  -------
+        //                           |         |
+        //                   clone tx once     |
+        //                   at creation       |
+        //                                     |
+        //                             clone the captured
+        //                             value on each call
+    })
+    .collect();
+
+// `tx` is still usable here
+```
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Syntax
+
+A `move` expression has the form:
+
+```
+move ( <expr> )
+```
+
+It is valid only inside the body of a closure, async block, or generator. Using `move(...)` outside these contexts is a compile error.
+
+### Desugaring
+
+A `move($expr)` desugars into a fresh temporary variable that is:
+1. Bound to the value of `$expr`, evaluated at closure-creation time
+2. Always captured by value (moved into the closure), regardless of how it is used in the closure body
+
+Concretely:
+
+```rust
+|| foo(move(expr))
+```
+
+desugars to the equivalent of:
+
+```rust
+{
+    let tmp = expr;
+    move_capture(tmp) || foo(tmp)
+}
+```
+
+where `move_capture(tmp)` means `tmp` is captured by move even if the closure body only borrows it.
+
+The temporary is not literally introduced in surface-level Rust. It exists in the compiler's intermediate representation. Its type is inferred from `$expr`.
+
+### Evaluation order
+
+When a closure contains multiple `move(...)` expressions, the temporaries are evaluated in source order (left-to-right, top-to-bottom) at closure-creation time:
+
+```rust
+|| f(move(a()), move(b()))
+```
+
+desugars as:
+
+```rust
+{
+    let tmp_1 = a();
+    let tmp_2 = b();
+    move_capture(tmp_1, tmp_2) || f(tmp_1, tmp_2)
+}
+```
+
+`a()` is evaluated before `b()`.
+
+### Interaction with `move ||` closures
+
+In a `move ||` closure, all captured variables are moved by default. A `move($expr)` within such a closure still introduces a temporary that is captured by move, but because the expression can be anything (including a reference) it allows capturing *less* than full ownership:
+
+```rust
+move || foo(move(&x))
+```
+
+desugars as:
+
+```rust
+{
+    let tmp = &x;
+    move_capture(tmp) || foo(tmp)
+}
+```
+
+The closure captures `tmp` (a reference) by move rather than capturing `x` by move. The closure borrows `x`.
+
+### Nested closures
+
+Each `move($expr)` is associated with its immediately enclosing closure. In nested closures, each `move(...)` creates a temporary for its own level:
+
+```rust
+|| {
+    || {
+        move(move(v.clone())).len()
+    }
+}
+```
+
+desugars as:
+
+```rust
+{
+    let tmp_outer = v.clone();
+    move_capture(tmp_outer) || {
+        let tmp_inner = tmp_outer;
+        move_capture(tmp_inner) || {
+            tmp_inner.len()
+        }
+    }
+}
+```
+
+The outer `move(...)` captures `v.clone()` into the outer closure. The inner `move(...)` then moves that value from the outer closure into the inner closure.
+
+### Drop semantics
+
+The temporary introduced by `move($expr)` is dropped when the closure is dropped. If evaluation of the expression panics, the panic propagates at closure-creation time and the closure is never created. Already-evaluated temporaries from prior `move(...)` expressions in the same closure are dropped in reverse order, following standard drop semantics.
+
+### Type inference
+
+The type of a `move($expr)` expression within the closure body is the type of `$expr`. Type inference works as normal. The temporary's type is inferred from the expression, and the use sites within the closure see that type.
+
+### Closures that are never called
+
+If a closure is created but never called, the captured temporaries are still dropped when the closure is dropped. This is the same behavior as any other captured value.
+
+### Parsing
+
+`move` is already a keyword in Rust. The parser recognizes `move(` inside a closure body as the start of a move expression rather than a call expression (since `move` is not a valid identifier).
+
+One ambiguity worth noting: at the statement level, `move(x) || y` could parse as either a move expression followed by a logical-OR, or as part of a closure. This is resolved by context: `move($expr)` is only valid inside a closure/async/generator body, and the parser does not produce a `move` expression in other positions. Within a closure body, `move(x) || y` is parsed as a move expression (`move(x)`) followed by `|| y` (which begins a nested closure literal).
+
+### Feature gate
+
+This feature is gated behind `#![feature(move_expressions)]`. No edition change is required since `move` is already a keyword in all editions.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+### Another way to spell the same thing
+
+Rust already has `move ||` closures and the let-before-closure pattern. Adding `move($expr)` is a third mechanism for the same underlying operation. Users must learn when to use which.
+
+### Readability of `move(...)` as an expression
+
+`move($expr)` looks like a function call but has different evaluation semantics: the expression runs at closure-creation time, not at call time. This is a new concept for readers to internalize. However, the `move` keyword already exists and already means "take ownership," so the meaning is consistent if not immediately obvious.
+
+### Complexity in nested cases
+
+Nested `move(move(...))` requires understanding which closure level each `move` targets. In practice this is uncommon, but the mental model requires understanding the "each `move` targets its immediately enclosing closure" rule.
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+### Why not fully automatic reference counting?
+
+This RFC emerged as part of a larger discussion in how to make reference counting more ergonomic. The original goal of that effort was to address the frequest request from higher-level Rust frameworks to have "fully automatic" reference counting, where users do not need to do anything explicit to `clone` a ref-counted value. This RFC is instead targeting a more modest goal, but one which we believe is also important: making fully explicit clones more ergonomic.
+
+The rationale is this: there are many Rust applications where explicit ref-counting is desirable (e.g., in kernel development or when developing a widely used, perf sensitive library or concurrent data structure). Moreover, even in higher-level applications, there are times when knowing exactly how many references are outstanding is important, such as when working with channels (where dropping the last `tx` handle causes a `rx` handle to return `None`, and hence can be important for correctness) or when working with large ref-counted data structures that consume a lot of memory. Fully automatic reference counting makes some Rust applications much cleaner but does nothing for these scenarios. This RFC, in contrast, helps all Rust users, even if it doesn't go as far as higher-level applications would like. Nothing in the RFC precludes adding fully automatic reference counting as future work.
+
+### Why this design over explicit capture clauses?
+
+[Explicit capture clauses][ecc] (e.g., `move(a.b.c.clone(), ..) || ...`) have been discussed for years. They front-load all capture decisions at the closure head, which helps in some cases but creates problems:
+
+- The capture list grows linearly with captures, duplicating information that's often obvious from the body.
+- They introduce a new syntactic position with its own grammar, including place remapping (`a.b.c = expr`) and open-ended captures (`..`).
+- For short closures, the overhead of the capture list dominates the closure itself.
+- They don't integrate naturally with the "clone then capture" pattern when you want the clone *at the point of use* rather than at the closure head.
+
+Move expressions solve the same problems with less syntax: you write `move($expr)` at the point of use, which is both the declaration and the consumption.
+
+[ecc]: https://smallcultfollowing.com/babysteps/blog/2025/10/22/explicit-capture-clauses/
+
+### Why not RFC #3680's `.use` / `use ||`?
+
+RFC #3680 proposed a `Use` trait and `use ||` closures that automatically clone captures. This was implemented experimentally but received feedback that implicit cloning hides costs. Move expressions take the opposite approach: every clone is written explicitly, but the boilerplate of shuttling it into the closure is eliminated.
+
+### Why not postfix syntax (`expr.move`)?
+
+A postfix operator would look like `expr.move` or `$expr.move`. The problem is that `move` doesn't operate on the final value of the expression; it changes *when* the entire expression is evaluated. Consider:
+
+```rust
+|| process(foo(bar()).move)
+```
+
+When does `bar()` execute? It must be at closure-creation time, but the postfix position suggests it's part of the normal call-time evaluation. A prefix `move(...)` makes the scope of early evaluation explicit.
+
+### What happens if we don't do this?
+
+The clone-before-closure pattern remains the only option. This is verbose but functional. The cost is friction for a common pattern, particularly in async code.
+
+### What is the fully desugaring semantics, canonical form for closures?
+
+As a general rule, it is desirable to have a "fully desugared, most general" form of each piece of Rust syntax. For closures, the "fully general form" is to convert a closure into a move closure that only references new temporaries introduced explicitly to be desugared (determining exactly what is captured into those temporaries is non-trivial and require analyzing the body of the closure to see what places it uses and how). The canonical closure form does not change with this RFC, it is simply extended so that `move` expressions are another way to get one of those temporary values:
+
+```rust
+// Example 1:
+|| foo(&vec)
+// equivalent to:
+{
+    let tmp = &vec;
+    move || foo(tmp)
+}
+
+// Example 2:
+move || foo(&vec)
+// equivalent to:
+{
+    move || foo(&vec)
+}
+
+// Example 3:
+|| foo(&vec, move(bar.clone()))
+// equivalent to:
+{
+    let tmp0 = &vec;
+    let tmp1 = bar.clone();
+    move || foo(tmp0, tmp1)
+}
+```
+
+## Prior art
+[prior-art]: #prior-art
+
+### C++ lambda init-captures (C++14)
+
+C++14 added generalized lambda captures with initializers:
+
+```cpp
+auto closure = [x = std::move(some_vec)] {
+    x.size();
+};
+```
+
+This allows arbitrary expressions in the capture list, evaluated at lambda-creation time. The captured variable `x` is then available in the lambda body.
+
+The design serves the same purpose as `move($expr)`: evaluating an expression at closure-creation time and binding it into the closure. The difference is that C++ places this in a capture list (front-loaded), while Rust's `move($expr)` is inline at the point of use. C++'s approach requires naming the captured variable explicitly in the capture list, which adds verbosity when the variable is used only once.
+
+### Earlier community proposals
+
+This idea has been discussed multiple times in the Rust community:
+
+- [Zachary Harrold proposed it on Zulip][z1] using the `super` keyword, and created a proc-macro prototype called [soupa](https://crates.io/crates/soupa).
+- [@simulacrum proposed using `move`][z2] instead of `super`, which better aligns with Rust's existing terminology.
+
+[z1]: https://rust-lang.zulipchat.com/#narrow/channel/410673-t-lang.2Fmeetings/topic/Design.20meeting.202025-08-27.3A.20Ergonomic.20RC/near/555236763
+[z2]: https://rust-lang.zulipchat.com/#narrow/channel/410673-t-lang.2Fmeetings/topic/Design.20meeting.202025-08-27.3A.20Ergonomic.20RC/near/555643180
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+### Should there be a syntax for moving specific variables?
+
+The compiler's HIR desugaring represents "this variable is captured by move" directly. An explicit syntax like `move[x, y] || ...` would let users write:
+
+```rust
+move[x, y] || {
+    x.process();
+    y.send();
+    z.borrow(); // z captured by ref
+}
+```
+
+On one hand, this is a natural complement to `move($expr)`, letting you move specific variables without a full `move ||`. On the other hand, any closure can already be written using `move($expr)` at each use site, making this redundant:
+
+```rust
+|| {
+    move(x).process();
+    move(y).send();
+    z.borrow();
+}
+```
+
+The tradeoff is conciseness vs. having a single mechanism. This question can be resolved during stabilization.
+
+## Future possibilities
+[future-possibilities]: #future-possibilities
+
+### Diagnostic suggestions
+
+The compiler could suggest `move($expr)` when it detects the let-clone-before-move-closure pattern:
+
+```
+help: consider using a move expression
+  |
+- let tx_clone = tx.clone();
+- tokio::spawn(async move { send(tx_clone).await });
++ tokio::spawn(async { send(move(tx.clone())).await });
+```
+
+This is left to implementation discretion rather than specified by this RFC.
+
+### Integration with a `Share` trait
+
+A future `Share` trait identifying types where cloning creates an alias (rather than an independent copy) would compose naturally with move expressions:
+
+```rust
+tokio::spawn(async {
+    send(move(tx.share())).await;
+});
+```
+
+This is out of scope for this RFC.
+
+### `move($expr)` outside closures
+
+A future extension could allow `move($expr)` in non-closure contexts as a no-op identity expression, making it possible to write generic code that works both inside and outside closures. Whether this is desirable is left to future discussion.
--- a/text/0000-move-expressions.md
+++ b/text/0000-move-expressions.md
@@ -50,7 +50,7 @@ This pattern at least avoids the need to introduce "dummy names" like `_some_a`,
 ## Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` will execute when the closure is created and then moved into a temporary captured by the closure; `move($expr)` is then replaced with a place expression consisting of this temporary. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
+We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` will execute when the closure is created and then moved into a temporary captured by the closure; `move($expr)` is then replaced with a place expression referring to this temporary. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
 
 ### `move($expr)` gives explicit control over captures
 
@@ -311,6 +311,17 @@ The temporary introduced by `move($expr)` is dropped when the closure is dropped
 
 The type of a `move($expr)` expression within the closure body is the type of `$expr`. Type inference works as normal. The temporary's type is inferred from the expression, and the use sites within the closure see that type.
 
+### Place expression semantics
+
+Syntactically, `move($expr)` parses as a value expression. However, like other value expressions (e.g., function calls), it can be promoted to a place expression by the compiler when used in a place context (e.g., `&move(...)` or `&mut move(...)`). The temporary that is introduced is declared as mutable, so `&mut move(x)` is always valid without the parent binding of `x` being declared `mut`.
+
+```rust
+|| {
+    // Valid: the captured temporary is declared mutable
+    process(&mut move(output_tx));
+}
+```
+
 ### Closures that are never called
 
 If a closure is created but never called, the captured temporaries are still dropped when the closure is dropped. This is the same behavior as any other captured value.
@@ -537,6 +548,20 @@ This idea has been discussed multiple times in the Rust community:
 
 ## Unresolved questions
 [unresolved-questions]: #unresolved-questions
+
+### Choice of syntax
+
+I'm not convinced that `move($expr)` is the right syntax. Here are some alternatives keywords I've considered:
+
+* `capture($expr)` (new keyword, though)
+* `super($expr)` (heavily overloaded keyword, I generally prefer `move` as it is already associated with closure capture rules, but open to others' takes)
+
+And then there is the question of `()` vs `{}` (thanks @juntyr for [raising it](https://github.com/rust-lang/rfcs/pull/3968/changes#r3356222419) earlier!)
+
+* `move($expr)`
+* `move { $expr }`
+
+[See discussion on github](https://github.com/rust-lang/rfcs/pull/3968#pullrequestreview-4435393221).
 
 ### Should there be a syntax for moving specific variables?
 
--- a/text/0000-move-expressions.md
+++ b/text/0000-move-expressions.md
@@ -50,7 +50,7 @@ This pattern at least avoids the need to introduce "dummy names" like `_some_a`,
 ## Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` will execute when the closure is created and then moved into a temporary captured by the closure; `move($expr)` is then replaced with a reference to this temporary. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
+We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` will execute when the closure is created and then moved into a temporary captured by the closure; `move($expr)` is then replaced with a place expression consisting of this temporary. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
 
 ### `move($expr)` gives explicit control over captures
 
@@ -91,7 +91,7 @@ When you want to be fully explicit about what a closure captures, you can list a
 || {
     let vec = move(input.vec);
     let data = move(&cx.data);
-    let output_tx = move(output_tx);
+    let mut output_tx = move(output_tx);
 
     process(&vec, &mut output_tx, data)
 }
@@ -143,6 +143,60 @@ data_source_iter
 
 // `tx` is still usable here
 ```
+
+### Common patterns with cloning
+
+Cloning a value into a closure is one of the most common uses of `move($expr)`. The right pattern depends on how the closure uses the cloned value and how many times the closure is called.
+
+**`move(x.clone())` — clone once, consume once (FnOnce)**
+
+When the closure is called exactly once and consumes the value, clone directly into the capture:
+
+```rust
+let handle = tokio::spawn(async {
+    channel.send(move(tx.clone())).await;
+});
+// tx still usable here
+```
+
+**`&move(x.clone())` — clone once, borrow on each call (Fn / FnMut)**
+
+When the closure is called multiple times but only needs a reference to the cloned value, take a reference to the capture. This avoids consuming the captured clone on the first call:
+
+```rust
+let f: Box<dyn Fn()> = Box::new(|| {
+    let data = &move(config.clone());
+    data.validate();
+    data.report();
+});
+f(); // works
+f(); // works again — data is borrowed, not consumed
+```
+
+**`move(x.clone()).clone()` — clone once at creation, clone again per call (Fn / FnMut)**
+
+When the closure is called multiple times and each call needs an owned value, clone the captured value on each invocation:
+
+```rust
+fn processed_items(
+    shared_config: &Rc<SharedConfig>,
+    items: Vec<Item>,
+) -> impl Iterator<Item = ProcessedItem> + 'static {
+    data_source_iter
+        .map(|item| {
+            process(item, move(shared_config.clone()).clone())
+            //            --------------------------- -------
+            //                      |                     |
+            //            clone shared_config once        |
+            //            when the closure is created     |
+            //                                            |
+            //                                  clone from the captured
+            //                                  value on each call
+        })
+}
+```
+
+This is useful when you need a fresh owned copy per call but wish to have the closure own the value (in this example, the closure needs to own the `Rc<SharedConfig>` so it can satisfy the `'static` bound).
 
 ## Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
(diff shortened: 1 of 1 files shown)


nikomatsakis:
This has been the subject of much prior discussion. I'm nominating for brief lang-team discussion in our upcoming Wednesday meeting.

@rustbot label +I-lang-nominated

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+
+The type of a `move($expr)` expression within the closure body is the type of `$expr`. Type inference works as normal. The temporary's type is inferred from the expression, and the use sites within the closure see that type.
+
+### Closures that are never called
+
+If a closure is created but never called, the captured temporaries are still dropped when the closure is dropped. This is the same behavior as any other captured value.
+
+### Parsing
+
+`move` is already a keyword in Rust. The parser recognizes `move(` inside a closure body as the start of a move expression rather than a call expression (since `move` is not a valid identifier).
```
fmease: Nightly rustc's parser also recognizes `move(` as the start of a move expression in all other places where an expression is expected, not just *inside a closure body* meaning it's a normal expression which I agree with. `const _: () = move(0);` is syntactically valid in the reference implementation, justifiably so.

Or are you saying that `move($expr)` should not be an *Expression* but rather an *ExpressionIncludingMove* and grammar rule *ClosureExpression* is to be updated to *… `(` ExpressionIncludingMove | `->` TypeNoBounds BlockExpressionIncludingMove `)`\**? I doubt that as it would be quite unconventional and inconsistent.

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
+- Feature Name: `move_expressions`
+- Start Date: 2026-05-14
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#155050](https://github.com/rust-lang/rust/issues/155050)
+
+## Summary
+[summary]: #summary
+
+Add `move($expr)` syntax inside closures, async blocks, and generators. A `move($expr)` evaluates the expression at closure-creation time and captures the result by value. This gives precise control over what a closure captures and when, without needing temporary variables outside the closure.
```
juntyr, line 9: Why is `move(x)` chosen instead of `move { x }` which might communicate the different scope better?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Closures that are never called
+
+If a closure is created but never called, the captured temporaries are still dropped when the closure is dropped. This is the same behavior as any other captured value.
+
+### Parsing
+
+`move` is already a keyword in Rust. The parser recognizes `move(` inside a closure body as the start of a move expression rather than a call expression (since `move` is not a valid identifier).
+
+One ambiguity worth noting: at the statement level, `move(x) || y` could parse as either a move expression followed by a logical-OR, or as part of a closure. This is resolved by context: `move($expr)` is only valid inside a closure/async/generator body, and the parser does not produce a `move` expression in other positions. Within a closure body, `move(x) || y` is parsed as a move expression (`move(x)`) followed by `|| y` (which begins a nested closure literal).
```
fmease: > and the parser does not produce a `move` expression in other positions

As I just wrote, that would be quite surprising and unorthodox.

> at the statement level, `move(x) || y` could parse as either a move expression followed by a logical-OR, **or as part of a closure** [emphasis mine]

I find the highlighted part a bit too vague; after all `move(` cannot possibly be a closure modifier unless you'd like to reserve the syntax for explicit capture list which I don't think is the case as you only seem to propose `move[…] |…| …` in *Future possibilities*.

I guess you meant to refer to the ambiguity where `move(x) || y` in statement expression context could be interpreted as *two* statements, move expression `move(x)` followed by closure `|| y`? That's fixed by considering move expressions to be [*ExpressionWithoutBlock*](https://doc.rust-lang.org/nightly/reference/expressions.html#grammar-ExpressionWithoutBlock).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

fmease: > Within a closure body, `move(x) || y` is parsed as a move expression (`move(x)`) followed by `|| y` (which begins a nested closure literal).

Could you clarify that? The rustc implementation actually parses both `|| { move(x) || y }` and `|| move(x) || y` ("`move(x) || y` within a closure body") as a closure containing a logical OR which I agree with; there's no "nested closure literal".

If you actually want to parse `|| { move(x) || y }` as a closure containing a move expression followed by a closure, then move expressions need to be an [*ExpressionWithBlock*](https://doc.rust-lang.org/nightly/reference/expressions.html#grammar-ExpressionWithBlock). However, that would imply that `|| { move(a) move(b) move(c) }` would be syntactically valid (i.e., semicolon not needed to terminate expression statement `move($expr)`) which I don't think is intended?

kennytm: Yeah I find the last sentence very confusing. In the current spelling `move(x) || y` only makes sense as a logical-OR expression. Unless this sentence was written when it was spelled as `move { x } || y`?

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+
+### A common pattern: listing captures at the top
+
+When you want to be fully explicit about what a closure captures, you can list all captures at the top of the body:
+
+```rust
+|| {
+    let vec = move(input.vec);
+    let data = move(&cx.data);
+    let output_tx = move(output_tx);
```
kennytm:
```suggestion
    let mut output_tx = move(output_tx);
```

This is used in line 96 as `&mut output_tx` so `output_tx` itself must be `mut`.

This raises a question, what happens in the case we don't list the captures separately?
```rust
|| process(&move(input.vec), &mut move(output_tx), move(&cx.data));
```
I suppose this would error unless `output_tx` in the parent scope is also declared as `mut`?

And similarly, if the content of `move` is an rvalue (i.e. not a place expression), taking `&mut` should be always allowed?
```rust
|| f(&mut move({ output_tx }))
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
nikomatsakis: I agree, this section is too hand-wavy. I'll go make it more precise and compare against the implementation.

text/0000-move-expressions.md
nikomatsakis: Good question. I'll add a FAQ. I'm open to more discussion or bike-shedding on the choice of keyword. The reason for `()` though is that it is often used inside of a context like `move(foo.clone()).bar()`, and I find `move { foo().clone }.bar()` to be ... more distracting? I don't know, it's not a hard opinion. I'm going to open a separate "bikeshed thread" to explore options.

nikomatsakis, line 9: **Desired feedback: choice of syntax**

I'm not convinced that `move($expr)` is the right syntax. Here are some alternatives keywords I've considered:

* `capture($expr)` (new keyword, though)
* `super($expr)` (heavily overloaded keyword, I generally prefer `move` as it is already associated with closure capture rules, but open to others' takes)

And then there is the question of `()` vs `{}` (thanks @juntyr for [raising it](https://github.com/rust-lang/rfcs/pull/3968/changes#r3356222419) earlier!)

* `move($expr)`
* `move { $expr }`

nikomatsakis (commented):
**Desired feedback: choice of syntax**

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
nikomatsakis: The primary goals I think are clarity of semantics -- I am not sure if `move` conveys that well or not. `capture` strikes me as the most clear, but the potential conflict with existing code may be a major issue.

I suppose that `capture { $expr }` could be done as a "contextual keyword" and is unlikely to conflict with existing structs.  I generally dislike contextual keywords.

steffahn: IMO, `move(x)` is bad because it’s too similar to C++’s `std::move(x)` but it does something entirely different.

Also regarding `move { $expr }`, I think that
```rs
move {
    block_contents()
    …
}
```
might look too similar to
```rs
move || {
    block_contents()
    …
}
```

---

I personally would like a new keyword like `capture`. I’m not a big fan of recycling keywords for entirely different purposes, anyway. `move` for closures was to indicate that implicit closure captures should all be done by moving. This new thing is for explicit closure captures (and fully flexible w.r.t. whether things are captured by moving, borrowing, or other preprocessing steps).

I’m unsure about style of parentheses. Both `capture($expr)` and `capture { $expr }` seem reasonable. The latter is perhaps syntactically more noticeable without syntax highlighting. The only downside I could see is that “capture” is a bit long.

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+    }
+});
+```
+
+This pattern at least avoids the need to introduce "dummy names" like `_some_a`, but it still has many of the same downsides.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` will execute when the closure is created and then moved into a temporary captured by the closure; `move($expr)` is then replaced with a reference to this temporary. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
```
steffahn: > `move($expr)` is then replaced with a reference to this temporary

why a reference? Can’t you use the captured thing by-value? Or… wait a minute… do you not actually mean a [reference](https://doc.rust-lang.org/std/primitive.reference.html)? *[I’m only noticing this alternative intepretation on second read.]*

To improve this confusion:

> when the closure is called, `move($expr)` functions as a place expression referring to this temporary

Also, perhaps spelling “temporary” longer as “temporary variable” is easier to read for more people?

> The expression `$expr` will execute when the closure is created and then moved into a temporary variable captured by the closure; when the closure is called, `move($expr)` functions as a place expression referring to this temporary variable.

The use of “moved to” in this sentence is *perhaps a fun trick/mnemonic* if the keyword is still `move`, but we could just as well write “stored in” or “assigned to” or something like that.

Here’s a possible rewording, though less comprehensive edits are of course possible.

```suggestion
We propose to introduce a `move($expr)` expressions that can appear in a closure or a future. The expression `$expr` is evaluated when the closure is created and its value is stored in a temporary variable captured by the closure. When the closure is called, `move($expr)` functions as a place expression referring to this temporary variable. This can be used to capture a clone of a value (`move(vec.clone())`) but also other derived values (e.g., `move(vec.len())`).
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+// equivalent to:
+{
+    let tmp0 = &vec;
+    let tmp1 = bar.clone();
+    move || foo(tmp0, tmp1)
+}
+```
+
+## Prior art
+[prior-art]: #prior-art
```
steffahn, line 523: I think, if *closure syntax* is viewed as a form of *quasiquotation*, then these `move(…)` expressions basically just *antiquotation*, right? I’m not sure what the best concrete prior art to cite on this would be but I believe the comparison is quite useful.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+[Explicit capture clauses][ecc] (e.g., `move(a.b.c.clone(), ..) || ...`) have been discussed for years. They front-load all capture decisions at the closure head, which helps in some cases but creates problems:
+
+- The capture list grows linearly with captures, duplicating information that's often obvious from the body.
+- They introduce a new syntactic position with its own grammar, including place remapping (`a.b.c = expr`) and open-ended captures (`..`).
+- For short closures, the overhead of the capture list dominates the closure itself.
+- They don't integrate naturally with the "clone then capture" pattern when you want the clone *at the point of use* rather than at the closure head.
+
+Move expressions solve the same problems with less syntax: you write `move($expr)` at the point of use, which is both the declaration and the consumption.
+
+[ecc]: https://smallcultfollowing.com/babysteps/blog/2025/10/22/explicit-capture-clauses/
```
steffahn, line 401: IMO, move expressions don’t solve the same problem fully. If a variable us used *multiple times* in the closure, then `move` expressions don’t really allow any significant shortening of boilerplate.

Current Rust (variant 1)
```rs
let foobar_ = foobar.clone();
use_closure(move || {
    foobar_.method1();
    foobar_.method2();
});

foobar.still_usable_here();
```
(variant 2)
```rs
use_closure({
    let foobar = foobar.clone();
    move || {
        foobar.method1();
        foobar.method2();
    }
});

foobar.still_usable_here();
```

With explicit capture (with shorthand supporting a method like `.clone()`):
```rs
use_closure(move(foobar.clone()) || {
    foobar.method1();
    foobar.method2();
});

foobar.still_usable_here();
```

With `move($expr)`:

```rs
use_closure(|| {
    let foobar = move(foobar.clone());
    foobar.method1();
    foobar.method2();
});

foobar.still_usable_here();
```

---

I guess what’s really missing is the potential for any *shorthand* to avoid the need for listing `foobar` *twice*? The point where you need to type every cloned variable name *two extra times* is where things get annoyingly verbose.

That’s especially if you introduce a second usage, e.g. going from
```rs
|| {
    move(foobar.clone()).method1();
}
```
to
```rs
|| {
    let foobar = move(foobar.clone());
    foobar.method1();
    foobar.method2();
}
```
involved introducing *three* mentions of `foobar`.

---

One thing that `move($expr)` does however allow is to shorten things with a simple macro. (Maybe that’s all we really need?)

```rs
use_closure(|| {
    capture_clone!(foobar);
    foobar.method1();
    foobar.method2();
});

foobar.still_usable_here();
```


```rs
use_closure(|| {
    // expands to a list of `let $ident = move($ident);` statements
    capture_clone!(foobar, baz, frobnication);

    foobar.method1(baz);
    baz.call(frobincation.wow());
    foobar.method2(baz, baz);
});

foobar.still_usable_here();
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
kennytm: > I suppose that `capture { $expr }` could be done as a "contextual keyword" and is unlikely to conflict with existing structs.

Haven't we learnt from `try { ... }` :sweat_smile:? As an expression `capture` cannot just be a contextual keyword.

And yes `struct capture` with a small `c` [does exist](https://github.com/Yale-PROCTOR/crat/blob/f598249c3665972a2e173517366bdad00084c95f/benchmarks/rs/libfreenect/examples/wavrecord.rs#L121).

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+```
+move ( <expr> )
+```
+
+It is valid only inside the body of a closure, async block, or generator. Using `move(...)` outside these contexts is a compile error.
+
+### Desugaring
+
+A `move($expr)` desugars into a fresh temporary variable that is:
+1. Bound to the value of `$expr`, evaluated at closure-creation time
```
ds84182, line 217: Perhaps I'm an outlier but many expect that, in the absence of control flow, programs are executed in the order declared within the source file. This is a departure from that, as these expressions are not evaluated in the same place as its neighbors.

Expression evaluation escapes execution.

One can argue that constants do the same thing, but constants have no side effects so the site of evaluation cannot be observed (execution _duration_ is not a side effect; this should not be debated here, panics neither as it does not produce a program to _have_ side effects in).

If this should exist, the expressions must never diverge (as in panic, infinite loop, abort). At that point the side effects of this early evaluation are miniscule, and therefore this would not cause much problems. It would require a new "color" of function for non divergence, which is probably years away at this point. (In other words, it's not const and not pure but a secret third thing 🙂)

But in its current state I think it will be fairly easy to encounter new footguns unique to Rust. This is solving a common problem, meaning it'll be commonly used, and at times accidentally or purposefully misused. A panic on closure creation is unprecedented, and suddenly macros containing closures have to deal with closure creation possibly executing arbitrary code.

Now, if I understand correctly, the suggested `use` idea allows the clone to be syntactically elided because the `Use` trait should only be implemented on types with "cheap clones" (which should not diverge except in extreme circumstances). Therefore this matches the expected behavior of a program as the inserted clone is negligible.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,444 @@
...
+- For short closures, the overhead of the capture list dominates the closure itself.
+- They don't integrate naturally with the "clone then capture" pattern when you want the clone *at the point of use* rather than at the closure head.
+
+Move expressions solve the same problems with less syntax: you write `move($expr)` at the point of use, which is both the declaration and the consumption.
+
+[ecc]: https://smallcultfollowing.com/babysteps/blog/2025/10/22/explicit-capture-clauses/
+
+### Why not RFC #3680's `.use` / `use ||`?
+
+RFC #3680 proposed a `Use` trait and `use ||` closures that automatically clone captures. This was implemented experimentally but received feedback that implicit cloning hides costs. Move expressions take the opposite approach: every clone is written explicitly, but the boilerplate of shuttling it into the closure is eliminated.
```
dlight: This feedback is puzzling, because `.use` isn't implicit and doesn't hide anything. It removes *emphasis* on cloning, just like `?` removes emphasis on the ceremony of forwarding an error to the caller. Sometimes de-emphasizing the cost of a clone is wanted and helps with legibility. But like in the `?` case, there is still a syntactical marker that points to what is happening.

In any way, is `.use` and `move()` compatible with each other? Can we imagine a future where we get both?

I'm asking because `.use` isn't *just* about closure captures, and it made clone-heavy code feel easier to read.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..c22c67ae737c5fd1796044cdcad691c0f2858696)*

text/0000-move-expressions.md
Nadrieril: > in the absence of control flow, programs are executed in the order declared within the source file

I'd argue that's still the case/not less the case! The code inside a closure was never going to be executed right away, it's just that now building the closure may also be executing a few extra operations, in the order in which the `move` expressions are. Not saying it's not surprising but I'd say: before, a closure was one big "hole" in the order of the execution; now it's a hole-with-holes or like a kind of sieve.

text/0000-move-expressions.md
nikomatsakis: The question is not so much whether it exists as whether it's widely used; it would require a new edition obviously but the transition is possible.

text/0000-move-expressions.md
nikomatsakis: I'll rewrite that to be more specific.

text/0000-move-expressions.md
scottmcm: I think https://github.com/rust-lang/rfcs/pull/3943 (cc @Amanieu) particularly pushes me away from using `move($expr)` here, because once that RFC means that we have a semantic difference between copy and move, using `move($place)` to say "yes I really want to move this, not just copy it" would be to me the "obvious" syntax for that opt-in which would be important for things like large arrays of copy types (`[u8; 1024]`, etc).

(TBH it's already unfortunate that `move` doesn't necessarily move, but rather copies for Copy types, but that part is not *this* RFC's problem.)

I don't know that I have a great alternative suggestion, though.  `precompute`, maybe?  Something to emphasize the *ordering* difference would be nice, since I think it's more about evaluation order than about capturing modes.

---

EDIT: The more I've been going through here, with the RFC being called "explicit closure capture" I'm liking the spelling of `capture { ... }`.

And we can do the whole `k#capture` on old editions approach to make it non-breaking.

nikomatsakis:
I've pushed a number of edits in response to comments made by folks on the RFC. I'm going to "resolve" the relevant conversations that are now captured in FAQs etc.

text/0000-move-expressions.md
nikomatsakis: I adjusted the text.

text/0000-move-expressions.md
nikomatsakis: I added a FAQ and explained @Nadrieril's perspective (which I share).

text/0000-move-expressions.md
nikomatsakis: I added a FAQ covering the "cliff" in going from one use to two.

nikomatsakis: I opted not to add discussion of that macro as an alternative because it didn't seem like a "real proposal".

text/0000-move-expressions.md
nikomatsakis: I didn't add anything for this, I wouldn't exactly call it *prior art*, but I guess we could add the comparison.

text/0000-move-expressions.md
nikomatsakis: I made an edit that I think makes it clearer.

text/0000-move-expressions.md
nikomatsakis: Thanks, fixed the original point.

nikomatsakis: > I suppose this would error unless output_tx in the parent scope is also declared as mut?

Actually, I missed this observation, I would have expected it to compile, but it's worth stating explicitly either way (rvalues are generally mutable, and `move()` is an rvalue).

text/0000-move-expressions.md
nikomatsakis: It was just wrong. I've rewritten it.

text/0000-move-expressions.md
nikomatsakis: Rewritten.

text/0000-move-expressions.md
```diff
@@ -0,0 +1,597 @@
...
+
+|| {
+    // Move the result of a function call
+    move(compute_config()).apply()
+}
+```
+
+### A common pattern: listing captures at the top
+
+When you want to be fully explicit about what a closure captures, you can list all captures at the top of the body:
```
scottmcm, line 88: Might not need to be this PR, but if one is intending to be fully-explicit it would be nice to have a "capture nothing I didn't write a `capture(foo)` expression for" opt-in on the closure syntax so that if you're not actually fully-explicit you can get an error about it and find out, rather than accidentally capture something.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..28321a084bc316f209d09b9921b809a8ad5481ae)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,597 @@
...
+    move_capture(tmp_outer) || {
+        let tmp_inner = tmp_outer;
+        move_capture(tmp_inner) || {
+            tmp_inner.len()
+        }
+    }
+}
+```
+
+The outer `move(...)` captures `v.clone()` into the outer closure. The inner `move(...)` then moves that value from the outer closure into the inner closure.
```
Skgland, line 304: I find this a bit confusing.

```
move(move(v.clone()))
          ^^^^^^^^^ argument expression of inner move
     ^^^^^^^^^^^^^^^^ inner move and argument expression of outer move
^^^^^^^^^^^^^^^^^^^^^ outer move
```

I would expect the outer move to capture `move(v.clone())` from inside the outer closure.
In the outer closure I would then expect the inner move to capture `v.clone()` from the surrounding context.



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..28321a084bc316f209d09b9921b809a8ad5481ae)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,597 @@
...
+        //                                     |
+        //                             clone the captured
+        //                             value on each call
+    })
+    .collect();
+
+// `tx` is still usable here
+```
+
+### Common patterns with cloning
```
scottmcm, line 147: (Not a question) Just wanted to say that I really like this section 👍 

(Makes me ponder having the template include something like "Is there something you'd put in a Rust-by-Example page about your feature?" to prompt more bits like this in other RFCs.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..28321a084bc316f209d09b9921b809a8ad5481ae)*

text/0000-move-expressions.md
teohhanhui: As a user this feels more natural to me:

```rust
move || foo(super { expr.clone() })
```

`super` seems like the natural fit to refer to the outer scope.

`move` is too overloaded in the context of closures.

And `{ ... }` feels more like "scope", whereas `(...)` *really* feels like function call syntax.

text/0000-move-expressions.md
```diff
@@ -0,0 +1,597 @@
...
+This allows arbitrary expressions in the capture list, evaluated at lambda-creation time. The captured variable `x` is then available in the lambda body.
+
+The design serves the same purpose as `move($expr)`: evaluating an expression at closure-creation time and binding it into the closure. The difference is that C++ places this in a capture list (front-loaded), while Rust's `move($expr)` is inline at the point of use. C++'s approach requires naming the captured variable explicitly in the capture list, which adds verbosity when the variable is used only once.
+
+### Earlier community proposals
+
+This idea has been discussed multiple times in the Rust community:
+
+- [Zachary Harrold proposed it on Zulip][z1] using the `super` keyword, and created a proc-macro prototype called [soupa](https://crates.io/crates/soupa).
+- [@simulacrum proposed using `move`][z2] instead of `super`, which better aligns with Rust's existing terminology.
```
Mark-Simulacrum, line 544: I'll selfishly flag this comment as well - https://github.com/rust-lang/rfcs/pull/3680#issuecomment-2308559508 :)

There's some possibly useful responses after that I think, though I believe they broadly align with where this RFC's concrete proposal has landed at.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..28321a084bc316f209d09b9921b809a8ad5481ae)*

text/0000-move-expressions.md
Skgland: Maybe it would help to show the desugaring in two steps

```rust
|| {
    || {
        move(move(v.clone())).len()
    }
}
```

desugars as:

```rust
|| {
    let tmp_inner = move(v.clone());
    move_capture(tmp_inner) || {
        tmp_inner.len()
    }
}
```

desugars as:

```rust
{
    let tmp_outer = v.clone();
    move_capture(tmp_outer) || {
        let tmp_inner = tmp_outer;
        move_capture(tmp_inner) || {
            tmp_inner.len()
        }
    }
}
```

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+}
+```
+
+where `move_capture(tmp)` means `tmp` is captured by move even if the closure body only borrows it.
+
+The temporary is not literally introduced in surface-level Rust. It exists in the compiler's intermediate representation. Its type is inferred from `$expr`.
+
+### Evaluation order
+
+When a closure contains multiple `move(...)` expressions, the temporaries are evaluated in source order (left-to-right, top-to-bottom) at closure-creation time:
```
scottmcm, line 241: nit: While it's in some ways obvious, it might be nice to include a note here about nested closures, since of course nested ones aren't exactly evaluated in source order because it depends when the inner closures run.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+
+### Why not postfix syntax (`expr.move`)?
+
+A postfix operator would look like `expr.move` or `$expr.move`. The problem is that `move` doesn't operate on the final value of the expression; it changes *when* the entire expression is evaluated. Consider:
+
+```rust
+|| process(foo(bar()).move)
+```
+
+When does `bar()` execute? It must be at closure-creation time, but the postfix position suggests it's part of the normal call-time evaluation. A prefix `move(...)` makes the scope of early evaluation explicit.
```
scottmcm, line 415: Another excellent call-out!  Agreed that a "warning up-front to a linear reader" is important here.

I think this specifically pushes me towards *braces* in the syntax, because typically (perhaps unintentionally) things that affect the meaning of the whole scope are in braces, like how it's `unsafe { ... }` because it affects the *scope* not the resulting *value*.

And thus being `|| capture { foo.clone() }.clone()` with braces here makes sense to me. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+            //                                            |
+            //                                  clone from the captured
+            //                                  value on each call
+        })
+}
+```
+
+This is useful when you need a fresh owned copy per call but wish to have the closure own the value (in this example, the closure needs to own the `Rc<SharedConfig>` so it can satisfy the `'static` bound).
+
+## Reference-level explanation
```
scottmcm, line 201: Question: what's the closure capture behaviour of a capture expression in a nested thunk?

For example, if I have
```rust
let my_vec = ...;
foo(async {
    bar(|| move(my_vec.len()))
});
```

Is `my_vec` captured by reference in the outer thunk?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+
+### Drop semantics
+
+The temporary introduced by `move($expr)` is dropped when the closure is dropped. If evaluation of the expression panics, the panic propagates at closure-creation time and the closure is never created. Already-evaluated temporaries from prior `move(...)` expressions in the same closure are dropped in reverse order, following standard drop semantics.
+
+### Type inference
+
+The type of a `move($expr)` expression within the closure body is the type of `$expr`. Type inference works as normal. The temporary's type is inferred from the expression, and the use sites within the closure see that type.
+
+### Place expression semantics
```
scottmcm, line 314: Is it a place expression for things like `ptr::addr_of!`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

Nadrieril: It should be yea, the explanatory desugaring suggests so. `addr_of` is `&raw const` anyway, and `&raw const move(..)` surely should work like `&move(..)`.

EDIT: I hadn't read the section in question, expecting the answer to be "it's always a place expression", see my comment below

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+
+Syntactically, `move($expr)` parses as a value expression. However, like other value expressions (e.g., function calls), it can be promoted to a place expression by the compiler when used in a place context (e.g., `&move(...)` or `&mut move(...)`). The temporary that is introduced is declared as mutable, so `&mut move(x)` is always valid without the parent binding of `x` being declared `mut`.
```
Nadrieril, line 316: I think `move(..)` must be a place expression. With what you suggest, `&mut move($expr)` would create a temporary to store the contents of `move($expr)`:
```rust
let mut tmp = move($expr);
&mut tmp
```
which means that the mutation is not preserved across calls to the closure. In fact here if the type of `tmp` is not `Copy`, you can't call the closure twice.

Moreover, value-to-place coercion doesn't happen on the lhs of an assignment, yet we want this to work:
```rust
let mut x = 42;
let f = || move(x) += 1;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

Nadrieril: This is further justified by the proposed desugaring: it says the `move(..)` expression is replaced by a local variable, which is a place expression. (if I may shill my own work) [my proposed desugaring](https://nadrieril.github.io/rust-via-desugarings/pipeline/closure-adt.html) replaces it with a struct field, also a place expression.

text/0000-move-expressions.md
Diddykonga: Kinda orthogonal, but I remember this being mentioned with syntax similar to this before:
```rust
move(a, mut b) || {}
```
Was there a decision to not re-use/extend the implicit part? I thought it made it more clear as to the effect happening at creation-time, rather then during body-execution.

kennytm: > Was there a decision to not re-use/extend the implicit part? I thought it made it more clear as to the effect happening at creation-time, rather then during body-execution.

This is answered in line 390 *"Why this design over explicit capture clauses?"*

text/0000-move-expressions.md
kennytm: IMO nested `move()` warrants at minimum a `complexity`-group clippy lint, it is too complicated for human consumption, and should better be split into multiple non-nested `move()` as 

```rust
|| {
    let tmp_inner = move(v.clone());
    || {
        move(tmp_inner).len()
    }
}
```

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+}
+
+fn main() {
+    println!("{}", parse_as_expr!(move(x))); // prints 0
+}
+```
+
+### Feature gate
+
+This feature is gated behind `#![feature(move_expressions)]`. No edition change is required since `move` is already a keyword in all editions.
```
TaKO8Ki, line 364: Now the feature gate is `move_expr`. We can fix it later if necessary.

Code: https://github.com/rust-lang/rust/blob/485ec3fbcc12fa14ef6596dabb125ad710499c9e/compiler/rustc_feature/src/unstable.rs#L642

The Unstable Book: https://doc.rust-lang.org/beta/unstable-book/language-features/move-expr.html

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
epage: My concern with `name(expr)` is that there is no graphical hint that evaluation order is being affected. This feels like a major gotcha in reading Rust code. I assume there were similar arguments around `expr.await`. I suspect that `await` has a bit more innocuous of an effect on understanding control flow, particularly when it exists inside of an `async` block / fn. Here, something from inside the middle of a closure is actually running as part of the closure's construction. There are no hints to expect this when looking at the top, requires full inspection to identify the presence, and is likely less common / familiar for people approaching Rust such that they would likely guess wrong what is happening.

The closest syntactic analog feels like a closure itself which is why I lean towards `name { expr }` or even `name || { expr }` (e.g. `capture || { data.clone() }`)

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+let f: Box<dyn Fn()> = Box::new(|| {
+    let data = &move(config.clone());
+    data.validate();
+    data.report();
+});
+f(); // works
+f(); // works again — data is borrowed, not consumed
+```
+
+**`move(x.clone()).clone()` — clone once at creation, clone again per call (Fn / FnMut)**
```
epage, line 176: I suspect the existence of `&move(x.clone())` would be surprising to people who might instead write `move(x.clone()).clone()`. If possible, might be worth a clippy lint.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
epage: The idea of closures being quasiquotation (https://github.com/rust-lang/rfcs/pull/3968#discussion_r3362435446) is also a good analogy for what could inspire the syntax (which I guess would be closest to `name { expr }`?)

text/0000-move-expressions.md
steffahn: @nikomatsakis I didn't mean to bring up the macro to be an alternative, but instead a possible use case. You can't currently define a macro like this, with `move(expr)` you can.

text/0000-move-expressions.md
kennytm: `capture || { $expr }` would be totally wrong because this expression still have same type as `$expr`, while the closure-like syntax makes it looks like it is of type `impl Fn() -> T`.

Nadrieril: On Zulip, using macro-looking syntax (e.g. `capture!(x.clone())`) was suggested as a way to hint that the contained tokens may not be evaluated in the way you expect.

camsteffen: I do think capture is more intuitive and direct for conveying semantics: "execute this at capture time". Moving a value is the desired effect, but changing execution time is the actual core semantic of the feature. I think I also like curly braces since it conveys a distinct scope, and feels more amenable to using any expression vs. requiring just a local. In the case of multiple captures, I might even prefer to use a capture block returning a tuple (in some cases).

On one hand I do like move for its consistency with existing syntax, and its brevity. On the other hand, I think this feature feels semantically distinct enough to warrant a new keyword. And I think the bit of overhead from a longer word and curly braces won't be too painful since implicit captures will continue to serve the majority of cases. Actually I'd consider it a negative if the feature encourages people to abandon implicit captures altogether, and making it into a block syntax could be a good deterrent from that.

text/0000-move-expressions.md
camsteffen: I'm pretty sure it has to be the other way around? The inner move is for the outer closure. This would be more consistent with evaluation order of nested function calls for example. Consider cases where the type changes across move expressions.

```rust
x = move(move(y).z());
```

camsteffen: I suggest that we simplify this by changing to:

> `move` always evaluates when the *innermost* closure is created. `move(move(..))` is invalid.

Nested move is just way too complicated to be worth the convenience IMO. There's always a way to break it into separate moves, one per closure.

text/0000-move-expressions.md
ds84182: > The code inside a closure was never going to be executed right away

This is a form of control flow. There was never an expectation that the code in a closure would be executed immediately, if at all. The call which executes the code is a change in the control flow of the program.

Closures today still do not violate the expected execution order of the program, as written.

> Not saying it's not surprising but I'd say: before, a closure was one big "hole" in the order of the execution; now it's a hole-with-holes or like a kind of sieve.

But in this case branches that are not taken can have side effects. So this "hole" in the order of execution is also a hole that is completely _extralogical_.

```rust
|| {
  if false {
    move(panic!())
  }
}
```

This panics on closure construction, despite never being reachable. This is unexpected behavior to anyone who isn't a Rust expert.

Sure, in a closure you can capture by placing expressions in unreachable code. But here you can seemingly _execute_ code in what is otherwise unreachable code.

---

The issue I want to hammer home is that this behavior is unprecedented in Rust today. Macros relying on closure/future construction being pure have no way to defend against this. Currently a closure that has nothing to capture has nothing to drop, and therefore the presence of a closure with no captures does not alter program behavior. That is no longer true when this is introduced.

kennytm: @ds84182 since this thread is already resolved I suggest you start a new thread on Lines 451–464 otherwise people can't see this reply.

text/0000-move-expressions.md
csmoe: if `move` is for closures, what about adopting C++'s closure syntax?

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+A common pattern for capturing clones, drawn from [a blog post by the Dioxus team][dioxus-blog], is to introduce "dummy variables" into the surrounding scope:
+
+```rust
+let _some_a = self.some_a.clone();
+let _some_b = self.some_b.clone();
+let _some_c = self.some_c.clone();
+tokio::task::spawn(async move {
+    do_something_else_with(_some_a, _some_b, _some_c)
+});
+```
```
yoshuawuyts, line 25: The motivation section spends quite some time on this example: once using dummy variables, and once more using expressions which return futures. I was expecting this RFC would show how this example would be improved by using `move($expr)` later on, but it doesn't do that. I feel like it should?

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+{
+    let tmp0 = &vec;
+    let tmp1 = bar.clone();
+    move || foo(tmp0, tmp1)
+}
+```
+
+### Does `move($expr)` change evaluation order away from source order?
+
+Arguably not more so than closures already do. The body of a closure is already deferred — none of it executes at the point where the closure expression appears in the source. A `move($expr)` extracts a sub-expression from that deferred body and evaluates it at the point where the closure is created, which *is* source order for the enclosing scope. This is exactly what the let-before-closure pattern does today:
```
yoshuawuyts, line 453: > ### Does `move($expr)` change evaluation order away from source order?
> Arguably not more so than closures already do. The body of a closure is already deferred — none of it executes at the point where the closure expression appears in the source.

From the perspective of the _closure body_ the evaluation order does change in ways that no other language feature does today: a line of code defined _later_ on in the body will now evaluate _before_ any of the other code in the body. Consider this example:

```rust
let file = File::open("./my-file.md")?;
(|| {
    write!(&file, "first")?;
    write!(&file, "second")?;
    write!(&file, "third")?;
    // 97 more lines here
})();
```

What are the first three lines we've written to `my-file.md`? With the addition of `move($expr)` we can no longer answer that question without first having read all 100 lines of the closure, because any line might contain a `move($expr)` that will be run before the rest of the body. This is a change in evaluation order (run earlier) in the opposite direction of closures (run later).

yoshuawuyts (commented):
When this RFC was discussed in the lang meeting three weeks ago I unfortunately had to drop early. I did still want to elaborate on some of the concerns I have though, so I wrote it out in a post instead: [Hoisting Expressions - Yosh Wuyts](https://blog.yoshuawuyts.com/hoisting-expressions). The main concern I have with this proposed feature are the changes to the evaluation ordering. I've added an in-line comment on this RFC with an example that illustrates what I mean.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
QuartzLibrary: As someone reading through the RFC for the first time, my first thought was dislike of the fact that it was inline (vs. previous `move` prefix proposals). 
I then realised that's silly because capturing variables is the whole point of closures and inline is better.

Given that, the syntax still felt off, and -> `move {}` -> `capture {}` seemed like the intuitive improvement.
I was happy to see it mentioned at the end, though split across alternatives.

text/0000-move-expressions.md
Mark-Simulacrum: You have to read all the way to `()` to know that any of the code in the closure is going to get executed though, right? And in most cases the closure is passed into some other function / stored in a local variable / spawned into a runtime to run at an arbitrary point in time, which further obscures that relationship. Outside of an IIFE, it would be straightforward to copy your example in a way that sneaks an extra write in (at least in general - you might need an Arc but I wouldn't expect it to be hard to disguise that the file is a shared handle).

```rust
let file = File::open("./my-file.md")?;
spawn(|| { // runs at arbitrary time
    write!(&file, "first")?;
    write!(&file, "second")?;
    write!(&file, "third")?;
    // 97 more lines here
});
write!(&file, "sneaking in before first")?;
```

I think it *is* true that this feature provides risk of making code hard to read, but I think the vast majority of code won't do that in practice, and will benefit from avoiding the need to pre-declare a bunch of variables just for moving clones or other types into the closure; I think on-net the readability of code will improve as a result of this.

One potentially worthwhile callout: if it turns out this feature is terrible for readability, I think it would be fairly straightforward to remove it in a future edition (or requite sign-posting of some kind, etc.). There's clear syntax to attach migrations to and injecting migrated code shouldn't be too hard.

kennytm: While I don't think hoisting-in-the-middle is an incurable problem in reality[^1], the counter in the above comment is pretty weak.

1. You don't have to read all the way to `()`, as soon as you see `(||` you know there's something fishy going on. IIFE is still a common practice in Rust because `try` is still unstable and it's the easiest way to simulate the feature, and I expect IIFE could be classified as an anti-pattern after `try` is stabilized, and as a side effect invalidate Yosh's original "***change** the evaluation order*" claim. 

2. The `"sneaking in before first"` example is missing the point entirely. When reviewing code, those 100 lines of closures are in a different indentation level than the final `write!()`. If I decide to skip reviewing the closure content I can collapse the block in an IDE or just quickly scroll through eyeing on the spacing. This makes the final `write!()` very prominent to the reviewer. While a `move()` at the 37th line within the closure could be easily missed.

I agreed that hoisting-in-the-middle won't change the evaluation order away from source order in the technical sense, but I do think this would indeed make reviewing more straining, since the body of a closure may participate in evaluation in the definer's scope, so a strict review must read the code in 2 passes, and can't simply skip the closure entirely.

| first pass | second pass |
|---|---|
| 1st pass to read the code outside (but need to realize the several `move {}` in the middle | 2nd pass to read the content closure | 


[^1]: for instance we could have a `style`/`pendantic`-category clippy lint that requires all `move()` usage must appear within the first few lines

Nadrieril: I think it's simply bad practice to use a `move` expression that does something non-trivial. Clippy could lint against that. Writing to a file is very much never what I'd expect to be doing there, exactly for the reasons mentioned.

So sure in an adverserial review we have to read the whole closure to know what is getting executed when, but in common practice I'd blame the author for having written something weird, just like I might blame an author who writes a macro that accidentally executes its input twice.

yoshuawuyts: We are introducing an entirely new kind of expression kind in `move($expr)`/`capture { $expr }`/`hoist { $expr }`. If a new kind of expression is too general and we feel it should immediately be reigned in via lints, perhaps that's a good sign that we would actually prefer a more constrained solution?

scottmcm: (Said in the abstract, without taking a position on the concrete thing either way) One big advantage of lints is that we can tune them over time without it being breaking changes and without needing to update the reference with the exact details.  If for various reasons it's awkward to specify a grammar or lexical rule for what should be blocked, it can be better to intentionally allow a bit more than people "should" use to keep the "here's what the reference says you *can* do" simple, then add nuance to lints instead.

text/0000-move-expressions.md
CheaterCodes: I'm definitely in favor of using block syntax rather than function call syntax: I see a strong analogy to `const {}` blocks, where `const` means "execute at compile time" and `move/capture` means "execute at capture creation time.

Having this behavior hidden somewhere in the closure body feels a bit iffy, but I agree that this is somewhat the point of closures, which makes me feel better about it.

I do want to point out that `move || {...}` and `|| move {...}` look too similar for comfort.

oxalica: > * `move($expr)`
> * `move { $expr }`

This syntax is quite generic and does not ties directly to closure. It feels like it can also be used for literally explicit moving, in order to avoid auto-deref, field-wise-capture (currently done by `{ variable }`), or auto-`Copy` (currently impossible, will be relevant once we have https://github.com/rust-lang/rfcs/pull/3943).

Using `move($expr)` or `move { $expr }` for closure capturing will prevent future use of the most natural syntax for explicit moving.

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+
+```rust
+// Today: `expr` evaluates here, in source order
+let tmp = expr;
+|| use(tmp)
+
+// With this RFC: same semantics, just written inline
+|| use(move(expr))
```
nicopap, line 461: I'm confused by the example.

Is `use` here supposed to be a function? I guess so, but `use` is a rust keyword. Maybe `use` is something I don't know, I don't know all of the rust experimental syntax. If it's meant to be a function call, I'd recommend swapping `use` for something else, like `consume`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
```diff
@@ -0,0 +1,622 @@
...
+    z.borrow();
+}
+```
+
+The tradeoff is conciseness vs. having a single mechanism. This question can be resolved during stabilization.
+
+## Future possibilities
+[future-possibilities]: #future-possibilities
+
+### Diagnostic suggestions
```
nicopap, line 593: This is tangential, but to add on diagnostic suggestions:

The move-expression syntax is an opportunity to make more actionable error messages on `async` blocks and closure captures.

If I end up with a let-clone-before-move, it's often because my closure needed to `impl Fn` and the `move` forced it into a `FnOnce`. The compiler errors in those cases are often non actionable, and it's only through experience that I know to let-clone-before-move.

The compiler could recommend using move-expression in those situations. IIrc, I encountered similar situations with non-static futures/closures.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3968/6787515f4abad9ffec4539b8c4df0cf5a1cee672..5e68928e1dd91ef1d7c0a8f24932e80c1f1907ee)*

text/0000-move-expressions.md
nicopap: I've no strong opinions, but some points:

* I like `move($expr)` because it's 6 chars, including parenthesis
* Parenthesis is confusing. The only examples in the language of `keyword($tt)` are scope qualifiers `pub(super | in $path)`. It looks like a function call.
* It makes more sense to have `move { $expr }`, `keyword { $body }` is already used in rust eg `async { $body }`
* `move { $expr }` is 9 chars, including spaces
* The comment about the similarity with unquoting is very relevant and helped me better understand the feature
* Having random `move` in the body bubbling up to closure creation time is footgunny (same thing can be said of the `defer` keyword in go though)
* I like the symmetry between `move || { $body }` and `move` inside closure
* Although, the fact that `move || { $body }` and `|| move { $body }` are possible is confusing.
* `async || { $body }` and `|| async { $body }` is as confusing.
* `capture { $body }` is getting long, 12 chars
* Others mentioned the possibility of a `capture!` macro that could take advantage of move-expression. I really like it, it can completely replace the C++-like closure capture syntax.
* Having the `move` inside macros is as footgunny as in the middle of a closure with many LoC, if not worse.
    * The same can be said of `return` and `break` keywords though.
* I like to imagine move-expression as a lower level building block that allows more complex features, like the already-mentioned `capture!` macro.

### Suggestions

* Maybe have two keywords? I don't like those suggestions though, but the idea:
    * `move in { $expr }`
    * `as move { $expr }`
    * Maybe this can inspire someone to suggest something better 😅 
* Maybe add ability to explicitly name the closure, to not be as confusing as the nested `move` proposition. Eg:
    * `in $closure_name { $expr }`
    * `[do] capture [in $closure_name] { $expr }`
    * Not sure how feasible it is, since closures/async blocks are expressions. Maybe could work with `'labels`?

nicopap: More bikeshedding and random thoughts:

* `capture` is programming language jargon. We all know here that "the closure captures the environment", but for someone who doesn't have a CS degree (I remember being one of those) "capture" is not as evocative.
* I don't know of any language with a `capture` keyword. So there isn't any expectations about its meaning.
* Since move-expressions are a finer-grained version of the `move` keyword on closure, I think `move` is a good keyword to use.
* I thought of `paste`, and `stick`: It conveys that we evaluate once and re-use.
* There is value in not introducing new keywords.
* Using `move` here doesn't preclude from using a different syntax for the "future possibilities" in #3943. "most natural syntax" is subjective.

text/0000-move-expressions.md
nicopap: The `soupa!` macro has the advantage to work in arbitrary situations.

There is a `soupa!/super` duality similar to `async/await`, with a keyword specifying a scope, and a keyword specifying the block with special control flow inside the scope.

I think it makes a lot of sense to keep the concept of scope. Imagine a `soupa!` macro that you could combine with async blocks and closures like so:

```rust
    data_source_iter
        //   vvvvvv -- instead of soupa!{|item| {...}}
        .map(soupa! |item| {
            process(item, super { shared_config.clone() }.clone())
        })
```

It could be a keyword, or ☃️ , just using `soupa!` for illustration

The advantage of having a keyword at the scope level is that it makes explicit the fact that we have those "reverse deferred" expressions, which makes reviewing code easier.

Also it might be possible to combine this with labels to make it explicit which nested `soupa!` scope a move-expression should be initialized in.

PR #3977, opened 2026-06-27
Method chain as item

contactomorph:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3977)*

Introduce method chains: a native item whose parameter list is split across named, dot-separated sections, as an additive alternative to named/optional arguments.

> [!IMPORTANT]
>

[Rendered](https://github.com/rust-lang/rfcs/blob/ae5ca9739f54045a472e4d81ebe3877bf1417a2e/text/0000-method-chains-as-items.md)

### DIFF
--- a/text/0000-method-chains-as-items.md
+++ b/text/0000-method-chains-as-items.md
@@ -1,0 +1,640 @@
+- Feature Name: method-chains-as-items
+- Start Date: 2026-06-25
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+## Summary
+[summary]: #summary
+
+Introduce *method chains* as a native kind of Rust item: a function whose parameter list is
+split across a sequence of named, dot-separated sections instead of a single parenthesized list.
+Each section has a name and its own parameter list (which may be empty, or hold several
+parameters at once) and only the last section carries a return type and a body. For example:
+
+```rust
+fn define_movie<'a>(name: &'a str)
+    .released_in(release_year: usize)
+    .directed_by(director_name: &'a str) -> Movie
+{
+    Movie {
+        name: name.to_string(),
+        release_year,
+        director_name: director_name.to_string(),
+    }
+}
+```
+
+is called as a sequence of named method calls:
+
+```rust
+let movie = define_movie("The Lobster")
+    .released_in(2015)
+    .directed_by("Yorgos Lanthimos");
+```
+
+A method chain call is a single, atomic expression: only the complete chain, written out in
+full, is ever a valid call (see Guide-level explanation). This RFC defines only that observable
+behavior, not any particular implementation (see Reference-level explanation). Method chains can
+be declared as free items, as associated functions inside `impl` blocks, or as required/provided
+methods inside `trait` definitions.
+
+This pattern is already available today, via code generation, as the third-party proc-macro
+crate [`assemblist`](https://crates.io/crates/assemblist), which this RFC's author also
+maintains. This RFC proposes turning the underlying call shape into a compiler-native item,
+presenting it as a self-contained, additive alternative to the broader, long-debated proposal of
+adding named and optional parameters to Rust.
+
+## Motivation
+[motivation]: #motivation
+
+Rust deliberately has no named or optional parameters. Every call site is positional, every
+parameter of a function must be supplied, and there is no syntax to label an argument with the
+name of the parameter it fills. This is a long-standing, frequently revisited point of friction:
+proposals to add named and/or optional parameters surface regularly in the Rust community, and
+just as regularly stall. Part of the reason is that many separate sub-questions all need an
+answer at once, and reviewers tend to disagree on each one individually: which parameters can be
+named, whether they can be skipped, and in what order they can then be supplied. Whatever the
+answer, it also has to respect two things that already hold today: an existing call like
+`f(a, b, c)` has to keep meaning exactly what it already means, and a function name has to keep
+resolving to exactly one item regardless of which arguments are passed, since Rust functions,
+unlike some other languages' functions, are not overloaded by argument count or type.
+
+In the absence of named and optional parameters, the idiomatic workaround is to hand-write a
+chain of methods (commonly called a builder) that incrementally accumulates the values a
+caller supplies before producing the final result. There is no single canonical shape for this:
+authors are free to structure it however they like. Two broad strategies recur, though, each with
+its own trade-off: mutating a single instance through methods callable in any order, which is
+simple to write but only checked for completeness at runtime; or encoding which values have been
+supplied so far into the type itself (often called the *typestate* pattern) which pushes that
+check to compile time, at the cost of meaningfully more code to design and maintain. Whichever
+strategy an author picks, or wherever they land between the two, real costs remain:
+
+- **Boilerplate.** Every constructible type that needs this needs its own hand-written chain of
+  methods (or a derive macro tying one to a specific struct's fields) designed and written from
+  scratch.
+- **A trade-off between simplicity and compile-time completeness.** As above: an author gets one
+  property or the other, rarely both, out of a single hand-written type.
+- **Documentation spread across pages that aren't individually interesting.** Whatever shape the
+  builder takes, a consumer mostly cares about one thing: the complete sequence of calls and what
+  they produce together. But that behavior is documented piecemeal, split across the builder
+  type's own page and each of its methods', rather than presented as the single, cohesive thing it
+  actually is.
+
+Method chains avoid this trade-off entirely, while staying purely additive. They introduce a new
+kind of item, they do not change the meaning of any existing function or call expression:
+
+- They generate no boilerplate beyond the chain's own declaration: there is nothing to design or
+  hand-write beyond the chain itself (no separate builder type, no type-level tracking of what's
+  been supplied so far).
+- It is checked for completeness entirely at compile time, with no runtime check and no extra
+  type-level machinery needed to get there (see Guide-level explanation).
+- The whole chain is documented as a single item, not fragmented across a builder type's page and
+  its methods' pages (see Rationale and alternatives for why a macro-based implementation cannot
+  achieve this on its own).
+
+Because this RFC proposes a new kind of item rather than new call syntax for existing functions,
+it sidesteps the two constraints described above that proposals about named/optional parameters
+must consider. Nothing about how `f(a, b, c)` resolves today changes, because the syntax for method
+chains just mimics an ordinary chain of method calls and does not require altering the syntax for
+individual functions/methods. A method chain's own name still resolves to exactly one item, the same way an
+ordinary function's does (see Reference-level explanation): nothing about it depends on which
+arguments are passed, so no question of arity-based overloading ever comes up. This still addresses
+the same underlying need: letting callers supply grouped, named, and conditionally-shaped sets of
+parameters without writing or maintaining a hand-rolled builder.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Think of a method chain as a function whose argument list has been split into a sequence of
+named sections. Instead of:
+
+```rust
+fn define_movie(name: &str, release_year: usize, director_name: &str) -> Movie { /* … */ }
+
+define_movie("The Lobster", 2015, "Yorgos Lanthimos");
+```
+
+you write:
+
+```rust
+fn define_movie<'a>(name: &'a str)
+    .released_in(release_year: usize)
+    .directed_by(director_name: &'a str) -> Movie
+{
+    Movie {
+        name: name.to_string(),
+        release_year,
+        director_name: director_name.to_string(),
+    }
+}
+```
+
+and call it the same way you'd call any chain of methods:
+
+```rust
+let movie = define_movie("The Lobster")
+    .released_in(2015)
+    .directed_by("Yorgos Lanthimos");
+```
+
+Each `.section_name(…)` looks like an ordinary method call, but the whole expression (from
+`define_movie("The Lobster")` to the final `.directed_by("Yorgos Lanthimos")`) is really just
+one call, spread across several dot-separated sections instead of one parenthesized list. Writing
+only part of it, like `define_movie("The Lobster")` alone, or
+`define_movie("The Lobster").released_in(2015)` alone, isn't a smaller, valid version of that
+call: it's an incomplete one, and the compiler rejects it, naming the section it expected to come
+next. There is no point during evaluation where "a movie chain with the name set, waiting for a
+release year" exists as something you could store in a `let`, return from a function, or pass to
+something else. The chain simply isn't a value at all until every section has been supplied.
+
+This has direct, practical consequences:
+
+- **You cannot skip a section, call sections out of order, call a section twice, or stop
+  partway.** Only `define_movie(…).released_in(…).directed_by(…)`, written out in full, in
+  that order, in one expression, is something the compiler recognizes as a call at all.
+- **An incomplete chain is a compile error, not a runtime concern.** There is no "half-built
+  movie" state to accidentally use as if it were finished, because there is no such state to
+  begin with (see above).
+- **The chain reads top to bottom as documentation of its own parameters.** A caller (or an IDE's
+  autocomplete) sees `.released_in(`, gets a hint for `release_year: usize`, then sees
+  `.directed_by(`, and gets a hint for `director_name: &str`. Every parameter arrives carrying
+  the name its author gave it, without Rust needing named arguments inside a single `(…)` list.
+
+A section's parameter list can hold any number of parameters, including none at all, the way
+`.times()` does above. For instance:
+
+```rust
+fn begin(i: i32).next().finish(a: i32, b: i32) -> i32 {
+    i + a + b
+}
+
+begin(1).next().finish(2, 3);
+```
+
+### Method chains as methods
+
+A method chain can also be declared inside an `impl` block, optionally starting with a receiver,
+exactly like an ordinary method:
+
+```rust
+struct PlayList { tracks: Vec<Track> }
+
+impl PlayList {
+    pub fn insert(&mut self, index: usize)
+        .of_kind(kind: TrackKind)
+        .titled(title: String) -> &mut Track
+    {
+        let track = Track::new(kind, title);
+        self.tracks.insert(index, track);
+        &mut self.tracks[index]
+    }
+}
+
+playlist.insert(0).of_kind(TrackKind::Audio).titled("Intro".to_string());
+```
+
+Any receiver form an ordinary method accepts today (`self`, `&self`, `&mut self`, or an
+explicit-type receiver such as `self: Box<Self>` or `self: Rc<Self>`) can appear as the chain's
+receiver, with the one restriction that it must be the very first parameter of the chain's first
+section. There's nothing about splitting the rest of the parameter list across sections that
+requires restricting how `self` itself is taken.
+
+### What doesn't change
+
+A method chain is declared with the same `fn` keyword, the same visibility modifiers (`pub`,
+`pub(crate)`, …), and can be marked `async`, `const`, or `unsafe` exactly like an ordinary function.
+`async`, `const`, and `unsafe` apply to the chain as a whole. There's nothing to `.await` until
+every section has been supplied and the chain's body actually runs. Reading and reviewing existing
+Rust code is entirely unaffected:
+nothing about ordinary functions, methods, or call sites changes. A method chain is something a
+function's *author* opts into when declaring it; everything else in the language behaves exactly
+as it does today.
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Grammar
+
+A method chain item has the shape:
+
+```text
+chain := [attrs] [vis] [«unsafe»] [«const»] [«async»] «fn» ident [generics] «(» [receiver «,»]
+         params «)» ( «.» section )+ «->» ReturnType [where_clause] «{» body «}»
+
+section := ident [generics] «(» params «)»
+```
+
+with the following constraints:
+
+- `params`, in any section, may be empty or hold any number of parameters. There is no
+  requirement that a section introduce exactly one.
+- The optional `receiver` may appear only in the first section, but is otherwise unrestricted:
+  any receiver form valid for an ordinary method today (`self`, `&'a self`, `&'a mut self`, or
+  an explicit-type receiver such as `self: Box<Self>`) is valid here too, and any future
+  extension to the set of receiver forms ordinary methods accept (e.g. further
+  arbitrary-self-types work) applies equally to method chains with no separate decision required.
+  The only chain-specific restriction is positional: a receiver can only be the chain's very
+  first parameter.
+- Each section's `[generics]` introduces type and lifetime parameters that are in scope for that
+  section's own parameters and return type, and for every later section (exactly like a
+  function's generics, but introduced incrementally). A later section's bounds may reference an
+  earlier section's generic parameters (e.g. a section taking `n: T` can be followed by a section
+  bounded on `T: Display`).
+- Only the *last* section carries `-> ReturnType { body }`; every earlier section is purely a
+  parameter-list extension of the same underlying call.
+- A `where` clause, if present, appears once for the whole chain, after the final section's
+  return type and before its body, exactly like an ordinary generic function's. It can bound any
+  generic parameter introduced by any section, including ones introduced well before the final
+  section. There is no per-section `where` clause, and one cannot be written between two
+  sections: collecting every bound at the end, rather than next to the section that introduced
+  the corresponding generic parameter, keeps a chain's `where` clause exactly where a reader
+  already expects to find one.
+- `vis`, `unsafe`, `const`, and `async` apply once, to the chain as a whole, exactly like an
+  ordinary function's; there is no per-section visibility, safety, constness, or asyncness in
+  this proposal (see Future possibilities for the branching extension, where each alternative
+  reintroduces its own `fn` and, with it, its own modifiers). An `unsafe` chain requires an
+  `unsafe` block at the call site and runs its final section's body in an unsafe context, the
+  same as an ordinary `unsafe fn`. A `const` chain is held to ordinary `const fn` rules over the
+  single fused call described in Semantics; the one caveat is that a later section's `where`
+  clause bounding an earlier section's generic parameter (e.g. `T: Display`) is, for a `const`
+  chain, subject to whatever rules `const fn` already imposes on generic trait bounds today
+  (`~const` bounds), independently of this proposal.
+
+### Name resolution
+
+A method chain's first section's identifier lives in the same namespace as an ordinary function,
+associated function, or method name. Declaring two method chains (or a method chain and an
+ordinary function/method, since an ordinary function is just a method chain with a single
+section) with initial sections under the same name in the same scope is exactly as much a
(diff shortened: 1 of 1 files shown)


text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+
+is called as a sequence of named method calls:
+
+```rust
+let movie = define_movie("The Lobster")
+    .released_in(2015)
+    .directed_by("Yorgos Lanthimos");
+```
+
+A method chain call is a single, atomic expression: only the complete chain, written out in
```
ds84182, line 35: I don't understand how this is beneficial if the entire chain is required. Optionality compliments named parameters, and this lacks it entirely.

This is also quite similar to method chains in Dart. e.g. `var foo = new Foo()..a()..b()..c();` which calls a, b, and c on the new instance, discards their return values, and then initializes the variable with the new instance. However Dart's version is much more versatile. A Rust equivalent would work for methods that take self by-ref and by-mut. This allows for optionality.

But Dart-like method chaining is still missing required named parameters. While this proposed feature can emulate those, it adds arbitrary restrictions to the ordering of the "method" invocations.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+next. There is no point during evaluation where "a movie chain with the name set, waiting for a
+release year" exists as something you could store in a `let`, return from a function, or pass to
+something else. The chain simply isn't a value at all until every section has been supplied.
+
+This has direct, practical consequences:
+
+- **You cannot skip a section, call sections out of order, call a section twice, or stop
+  partway.** Only `define_movie(…).released_in(…).directed_by(…)`, written out in full, in
+  that order, in one expression, is something the compiler recognizes as a call at all.
+- **An incomplete chain is a compile error, not a runtime concern.** There is no "half-built
```
kennytm, line 155: I think forcing a compilation error is not gonna work cleanly, since `x.y().z()` itself is a valid expression the error can't happen at parser level, so it must be enforced by typeck which the compiler need to invent some type which can only be used if directly followed by the rest of the chain. This can be partially fulfilled by making the intermediate type unsized. With the current situation of having `unsized_fn_params` minus `unsized_locals`, you'll get these behavior:

```rust
// allowed
let movie_0 = define_movie(t0).release_in(y0).directed_by(n0);

// also allowed (!)
let movie_1 = ( define_movie(t1).release_in(y1) ).directed_by(n1);

// compile error (currently an ICE)
let movie_2 = { define_movie(t2).release_in(y2) }.directed_by(n2);

// allowed, but you can't use `partial_movie_3` to do anything safe.
let partial_movie_3 = &(define_movie(t3).release_in(y3));
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
kpreid: This proposal is similar to Smalltalk/Objective-C style method calls (e.g. `receiver moveWidgetFrom: arg1 to: arg2`) in that it defines a fixed sequence of identifiers (`"moveWidgetFrom:to:"`) which constitute mandatory naming for a fixed parameter list (`arg1, arg2`). However:

> Declaring two method chains … with initial sections under the same name in the same scope is exactly as much a duplicate-definition error as declaring two ordinary functions under that name is today.

That is, the declaration of the chain `.foo().bar()` requires that the name `foo` does not collide with other chains or methods. That makes this proposal far more limited than Smalltalk/Objective-C, where `foo:bar:` and `foo:baz:` are two distinct methods that can coexist.

Per #future-possibilities, @contactomorph sees this as something that can be added later. Thus, this RFC is serving two purposes, as I see it:

1. Adding a mechanism to rigorously name arguments, without the need to define a builder type and spread the documentation of the operation around that builder's items.
2. Being the foundation to later add branching/optional/repeated chain elements, thus becoming a means of overloading and a replacement for some but not all builders.

Given that, the questions that need answering are:

* Is (1) worthwhile on its own?
    * If so, is this design a good way to achieve (1)?
* Should Rust accept (1) now for the sake of getting (2) later?

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+
+Introduce *method chains* as a native kind of Rust item: a function whose parameter list is
+split across a sequence of named, dot-separated sections instead of a single parenthesized list.
+Each section has a name and its own parameter list (which may be empty, or hold several
+parameters at once) and only the last section carries a return type and a body. For example:
+
+```rust
+fn define_movie<'a>(name: &'a str)
+    .released_in(release_year: usize)
+    .directed_by(director_name: &'a str) -> Movie
```
clarfonthey, line 17: This example already strikes me as contentious due to the fact that the two strings share the same lifetime. How do you expect this to work? Should the compiler automatically shorten the lifetimes to be the intersection of the two (failing if this cannot occur), or should it refuse to accept a second string if its lifetime is shorter than the first?

To be clear, I think the former is the *right* response here, but I would say this example is not nearly as simple as it implies it should be, and already asks a lot of questions that would need to be answered.

And importantly, these are questions that creators of builder-pattern structs need to resolve today, so, this also makes it clear that this way of doing things doesn't necessarily get rid of those problems.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+
+Because this RFC proposes a new kind of item rather than new call syntax for existing functions,
+it sidesteps the two constraints described above that proposals about named/optional parameters
+must consider. Nothing about how `f(a, b, c)` resolves today changes, because the syntax for method
+chains just mimics an ordinary chain of method calls and does not require altering the syntax for
+individual functions/methods. A method chain's own name still resolves to exactly one item, the same way an
+ordinary function's does (see Reference-level explanation): nothing about it depends on which
+arguments are passed, so no question of arity-based overloading ever comes up. This still addresses
+the same underlying need: letting callers supply grouped, named, and conditionally-shaped sets of
+parameters without writing or maintaining a hand-rolled builder.
```
clarfonthey, line 103: So this is actually something I recently have thought a bit about, and I think that information like this belongs in the rationale of the RFC, not the motivation. Effectively, the motivation is where you pitch the problem, the explanations are the solution, and the rationale is where you justify the solution. This feels like a justification of method chains specifically rather than describing any particular problem.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+- **A trade-off between simplicity and compile-time completeness.** As above: an author gets one
+  property or the other, rarely both, out of a single hand-written type.
+- **Documentation spread across pages that aren't individually interesting.** Whatever shape the
+  builder takes, a consumer mostly cares about one thing: the complete sequence of calls and what
+  they produce together. But that behavior is documented piecemeal, split across the builder
+  type's own page and each of its methods', rather than presented as the single, cohesive thing it
+  actually is.
+
+Method chains avoid this trade-off entirely, while staying purely additive. They introduce a new
+kind of item, they do not change the meaning of any existing function or call expression:
```
clarfonthey, line 84: I would disagree mostly for reasons that feel a bit pedantic, but are important. Particularly, while it does not strictly affect any *existing* code which does not use this feature, it would affect the meaning of *future* code, because now I need to ask myself whenever I see a method chain, is this a Method Chain, or just a chain of method calls? And this invokes a nontrivial amount of cognitive overhead that should be considered.

For example, let's imagine the method call `a().b().c().d().e()`. You can add parameters if you think it matters. Currently, there is one canonical way to read this: `a()` is a function which returns some value (call it `A`), and then you call `A::b`, then `B::c`, etc.

With this new framework, I need to verify that first, there does not exist a method chain called `a()`, and then depending on the method chain definition, I interpret this differently. Using parentheses to disambiguate method chains, this could be either `(a().b()).c().d().e()`, `(a().b().c()).d().e()`, `(a().b().c().d()).e()`, or `(a().b().c().d().e())`. And, of course, the original, which could be thought of as a "trivial method chain," but I'm not going to do that here.

And then things get even more complicated if you have optional methods.

Sure, this merely adds something to the language from the perspective of it not breaking past code, but that is a requirement for any new features, not a benefit of any particular feature. It changes how people think about the code, which is not merely additive.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+expression spanning every section from the first to the last. The language only defines a
+meaning for the invocation in this complete form:
+
+- Writing only a prefix of a chain's sections (`ident(args0)` alone, `ident(args0).section1(args1)`
+  alone, and so on up to, but not including, the full chain) is not a valid expression in any
+  position: not as a statement, not as the right-hand side of a `let`, not as a function
+  argument, not as a return value. There is no point during resolution or evaluation at which
+  "the result so far" exists as a value of any type, nameable or not, that a program could refer
+  to, store, or pass around. A chain that is not completed in one expression is a compile-time
+  error naming the section the compiler expected to come next.
```
clarfonthey, line 285: This… is not the kind of semantics I would expect for a feature like this. I would expect that the exact semantics are, an incomplete prefix is an *incomplete* expression, and will fail to compile.

Specifically, I would expect the following to still work:

```rust
macro_rules! wrap {
    ($x:expr) => { $x }
}

let _ = wrap!(wrap!(a()).b()).c();
```

if `a().b().c()` is a valid method chain. It's very natural for macros to build up *normal* method chains like this, and I would expect Method Chains proper to only fail in type-checking.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+- A complete chain invocation behaves exactly as if it were a single call to a function taking
+  the concatenation of every section's parameters (plus the receiver, if any) in declared order,
+  and running the final section's body. There are accordingly no auto-trait, `Copy`/`Clone`, or
+  drop-timing questions to answer for "the result of a non-final section": no such value exists
+  for any such property to apply to. The only types that matter to a method chain are each
+  section's own parameter types and the chain's final return type, exactly as for an ordinary
+  function.
```
clarfonthey, line 292: How do you refer to the type of such a function? Are `extern "C"` method chains possible?

(Note: later points make it clear that this is *not* possible, but I think it's worth pointing out since even if you don't expose this to the user, the compiler *will* need to come up with some sort of type for it anyway.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+  `define_movie` or some hypothetical `define_movie.released_in.directed_by`.
+
+This RFC defines only this observable behavior, and deliberately imposes no implementation
+strategy beyond it. A compiler may lower a method chain to a single physical function (the most
+direct strategy, since nothing about its semantics requires any intermediate value to exist), to a
+sequence of single-method generic types the way today's `assemblist` macro does, or to anything
+else that reproduces the same observable behavior. Whichever strategy is used, the language
+defines no name for any such intermediate construct: it is never shown to `rustdoc`, never
+participates in name resolution, and is not something a Rust programmer is ever expected to
+reason about.
```
clarfonthey, line 310: I don't really like this framing because, well, first, *we do* show this to rustdoc. rustdoc has full access to the compiler internals and therefore will have visibility of whatever compiler representation we choose for this, and it needs to be able to convert that into usable documentation. Therefore, access to the original *is required* for rustdoc to function in some manner.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+  sections. See Future possibilities.
+
+Every example in the Guide-level explanation above is a complete chain invocation in this sense:
+`define_movie("The Lobster").released_in(2015).directed_by("Yorgos Lanthimos")` is one call,
+equivalent to calling a single function over `name`, `release_year`, and `director_name`
+together; `playlist.insert(0).of_kind(TrackKind::Audio).titled("Intro".to_string())` is one call
+equivalent to calling a single method over `index`, `kind`, and `title` together with the
+`&mut self` receiver.
+
+## Drawbacks
```
clarfonthey, line 406: I mention a pretty clear drawback when responding to an earlier point in the RFC: https://github.com/rust-lang/rfcs/pull/3977#discussion_r3523669504

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+also closer to how `assemblist` already behaves today.
+
+Its drawback is exactly what motivated moving away from it: it forces a compiler to actually
+generate, and a chain's intermediate calls to actually produce, a real value of a real type for
+every non-final section, even when, as in every example in this RFC, the whole chain is written
+out in one continuous expression and nothing was ever going to inspect, store, or outlive that
+intermediate value. Committing to that as part of the language's observable behavior also commits
+to answering, for every method chain, questions this RFC currently avoids needing to answer at
+all: what auto traits that intermediate value has, when it is dropped, whether it is `Send`, and
+so on, all for a value whose only legitimate use is to immediately call the next section.
```
clarfonthey, line 455: To be clear, the compiler is probably going to have to do this *anyway* just by the nature of how it works today. When parsing the expression, it's going to be identical to a regular method chain before name resolution happens. In order to properly resolve the names, it's going to have to assign a type for the initial call, and keep a status of the various methods chained onto it until it finishes, at which point, it will have a full type for the expression.

It would be ridiculous IMHO to fundamentally change how the compiler works just to accept this one feature. Instead, I would expect this feature to bend to how method chains normally work internally.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

text/0000-method-chains-as-items.md
```diff
@@ -0,0 +1,640 @@
...
+proposals did: #257's review had to invent a dedicated disambiguation rule purely to keep
+reordered, partially-named calls unambiguous ("if you alter the declaration parameter ordering in
+the middle of a function invocation, you are forced to specify the parameter names for the rest
+of the invocation"). This RFC's fixed order avoids needing any such rule: the order at the call
+site simply *is* the order at the declaration site. It also keeps the door open for the branching
+and optional/repeated-section extensions sketched in Future possibilities, which would be
+substantially harder to design and to reason about if sections could additionally be taken in any
+order.
+
+### Doing nothing
```
clarfonthey, line 471: I think one obvious thing missing from this section is how all of these individual points could be improved on their own, rather than simply offering this solution as the solution to all of them.

For example, you're right that the builder pattern makes things difficult because everything is spread on separate pages. Could we have some attributes for rustdoc to improve this? Maybe rustdoc could even generate some state diagrams for us?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3977/96d0d79254512b8b7d1c6b31bfb512a8b321c144..ae5ca9739f54045a472e4d81ebe3877bf1417a2e)*

joshtriplett:
I don't think we should make this change.

I'm not suggesting that we should never try to make the builder pattern better, or provide a better replacement for it. But this takes something we can already do (e.g. with builders, and the various macros people write to help construct them), and turns it into a native language mechanism *without most of the flexibility of builders*. No optional arguments, no type-state to figure out what has been supplied, and in particular the intermediate "methods" in the "method chain" don't seem to have return values.

So, rather than being an orthogonal language feature that interacts well with other things, it feels like an atomic feature that introduces various new concepts, including the reuse of syntax already used for function calls but in a way that's not actually acting like function calls. So, if we try to *add back in* some of the things we can already do with builders, like optional arguments or typestate, we end up either modifying those mechanisms to fit with this, or fixing the divergence between this and what we already have.

scottmcm:
I agree with Josh.  I want to lean on https://github.com/rust-lang/rfcs/pull/3681 at least for now.  Let's close.

nikomatsakis:
I agree on closing this *but* I want to say that the language I designed for my PhD thesis, Harmonic, had this feature and I loved it, mostly because I really wanted smalltalk. Ha. Basically the idea was to look like Java, so I wanted the normal thing to be `File.open(f) withOptions(bar)` but have `File.open(f)` for "back compatibility". But we are not trying to be smalltalk. =)

ChayimFriedman2:
Procedural note: can we close a RFC without a FCP (and without agreement from the author)?

contactomorph:
> Procedural note: can we close a RFC without a FCP (and without agreement from the author)?

I think it is extremely hasty and unfair. I never had the opportunity to answer.

> So, rather than being an orthogonal language feature that interacts well with other things, it feels like an atomic feature that introduces various new concepts

What _new concepts_ ? The idea of this RFC was precisely to be minimal.

> in a way that's not actually acting like function calls.

An alternative was included that makes them acting like normal function calls.

> without most of the flexibility of builder

Again because the idea was to propose the minimal core of the feature and leave the rest for future. For a long time now, Rust community has encouraged people to create small RFCs with mimimal changes and iterate on them. And I'm reproached exactly to do that ?

@joshtriplett have you even read the proposal completely ?

scottmcm:
> Procedural note: can we close a RFC without a FCP (and without agreement from the author)?

Yes.  FCPs are for one-way doors where the FCP period is the "speak now or hold your peace" time.  That's critical for stabilizing things that people can start to rely on and thus can't change further.

Closing an RFC is not irreversible. A feature can always be added later.  (A classic example is `dyn`, where https://github.com/rust-lang/rfcs/pull/1603 was closed but then later https://github.com/rust-lang/rfcs/pull/2113 was accepted.)

traviscross:
> Procedural note: can we close a RFC without a FCP (and without agreement from the author)?

We can, and we have in the past.  We talked about the process for triaging RFCs at some length in the meeting today where we went through this and many other open RFCs.  What follows is my own view and framing:

We use FCPs to establish team consensus for one-way doors (e.g., stabilizations), for actions of irreversible consequence (e.g., breakage), and where we want to establish precedent.

*Some* RFCs should be closed by FCP because we want to close that door.  We want to signal that "to do *X*, you must overcome an existing team consensus against doing *X*."

That's a high bar, though.  I'd want to review carefully before checking a box on that.  That creates a lot of work (and hence delay) for closing RFCs even when they're extremely unlikely to be accepted.

*In general*, when we close an RFC, we're not meaning to assert something that strong.  We mean to assert only that the probability of the RFC being accepted is so low that it's misleading to leave it open.  I think it's OK, actually, for any lang member (and probably even our lang advisors) to simply close an RFC, at that person's own discretion and on that person's own authority, when that's believed to be the case.  This doesn't set any precedent and is a two-way door.  Any other lang member could simply reopen it and say, "actually, I think this is interesting and we should consider it."

That's what's happening here.  There's simply no constituency for this RFC on the lang team at the moment.  It's extremely unlikely to move forward.  So it felt, to us on the call, fairer to the author and others for us to indicate that by leaving the comments above and closing this.

I appreciate (and I think we all appreciate) the author for having put this forward.  As @nikomatsakis highlighted, it's an interesting idea and I'm glad to have had a look at it.  It's just not something we're likely to adopt at this time.

kennytm:
> Procedural note: can we close a RFC without a FCP (and without agreement from the author)?

Today 5 RFCs are insta-closed[^1], all citing decision from triage meeting. Procedurally it is *documented* in https://github.com/rust-lang/rfcs/blob/master/lang_changes.md:

<table><tr><td>

... The result of triage will either be that the PR is assigned to a member of the sub-team for shepherding, the PR is closed as postponed because the subteam believe it might be a good idea, but is not currently aligned with Rust's priorities, ***or the PR is closed because the sub-team feel it should clearly not be done and further discussion is not necessary.*** In the latter two cases, the sub-team will give a detailed explanation. ...

</td></tr></table>

Similar clause exists for T-compiler too. **This is not mentioned in the "process" described in the main README though** https://github.com/rust-lang/rfcs/blob/master/README.md#what-the-process-is

I do find it is not looking good to massively close this many non-spam RFCs without the usual 10-day FCP waiting from community input. 

Additionally, not going through the rfc-bot means the normal `disposition-*` labels are not applied automatically so we can't tell whether the RFC is closed as "postponed" or "rejected" just by viewing from the PR list.

[^1]: #3204 - deferred to T-compiler, closed as "not yet relevant";<br> #3918 - deferred to T-libs, closed as "should be an ACP";<br> #3938 - closed as "postponed";<br> #3951 - closed as "rejected";<br> #3977 - closed as "rejected"

clarfonthey:
For what it's worth, I've opened a (team-only for now) [thread on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/546943-all.2Fprivate/topic/Closing.20RFCs.20without.20FCP.3F/with/609052401) to discuss the status quo on closing RFCs without FCP, and what documentation should be updated. We could potentially move this into a public discussion thread later, or maybe just have an RFC to discuss how the documentation should be updated and what way would best communicate to people opening RFCs to set expectations.

I don't think that necessarily discussing this *here* is the right option though, since it's kind of off-topic for the RFC itself and I think we should be trying to stay on-topic even for closed PRs.

> @joshtriplett have you even read the proposal completely ?

Just calling this out here: Josh is an active team member who does not make statements like the one he made without seriously considering proposals. I don't think it's acceptable to respond in the way you did, but I do understand the frustration on closing the RFC without adequate warning, which is why I think we should have a serious discussion with the team to help set expectations for RFC authors rather than surprising people.

And to clarify further why I think this is important to point out:

> What *new concepts* ? The idea of this RFC was precisely to be minimal.

I provided [explicit feedback](https://github.com/rust-lang/rfcs/pull/3977#discussion_r3523669504) on this point which you did not address, so, I think that it's worth understanding why team members make decisions like this. Adding new syntax is an *incredibly* high bar for the language, and we want sufficient justification. It is *incredibly easy* to miss details and you are not at all blamed for this, since well, writing RFCs is very hard and we intentionally want it to be a collaborative process.

This is why I think setting expectations is important; there are genuinely no hard feelings for closing an RFC, and as always, it's important for people to reconsider things as the language evolves. It's just that in this case, the bar was set so high and the justification didn't seem to get close to it that the team made the decision to preemptively close without an FCP. In fact, the existence of your crate implementing this feature is ironically a valid reason to decline implementing the feature: modifying the compiler and language is a lot of work and if the feature can be supported by a crate instead, maybe it's not worth it for now. It might be worth it in the future, but at least, not right now.

joshtriplett:
One key detail here: this evaluation was *not* based solely on the current state of the RFC; it also incorporated the anticipated/potential future development of it, as well. The specific framing of this RFC (e.g. taking a minimalist approach) did not harm it in this evaluation.

The issue is not "method chains can't possibly do these things" (e.g. optional arguments, repeated arguments). The issue is that the manner in which we would *want* to do that is fundamentally that it should use and build on and interact with existing language features, rather than replacing them with something that looks similar but operates differently and largely separately. For instance, with function calls, we allow `let b = builder(); let b = b.thing(arg); let x = b.build();`. And `b` has a type. (It may even have a different type after `.thing(arg)`.) The primary presentation of method chains, here, doesn't support that. And if we were to add all of that, the question then becomes whether we want to add a distinct mechanism for this, or extend function calls and structures. (Or provide a builder-helper library.)

I did see the bits about branching and continuations in your future work section. But as far as I can tell, those are still tied into the idea of not invoking any code until all the parameters are present. That is *a* useful mechanism, and would cover some use cases people currently use builders for, but not all.

We have proposals in progress for, for instance, structs with default values. We may wish to extend those proposals to more generally make it easier to create builders, or create something more ergonomic than builders. But a mechanism added to make that more ergonomic still needs to work in terms of the same orthogonal underlying language features (e.g. function calls, types, structures), rather than being a construct that feels "separate", if that makes sense.

On top of that, as others have noted: this is a problem that crates in the ecosystem (including yours) can solve, so if we're going to add something to the language, it needs to provide substantially *more* benefit than a crate is capable of.

contactomorph:
> I don't think it's acceptable to respond in the way you did

Your sentence is a moral judgment. Why is that not acceptable? @joshtriplett's previous message contains statements and suggestions that **factually** contradicts the RFC. So to me it really looked like he had not read the details.

> I provided explicit feedback on this point which you did not address, so, I think that it's worth understanding why team members make decisions like this.

I had already created a draft for responses for your messages. I wanted to make them clean and correct (I am not an English native speaker) before submitting them all together. Sorry, I had been busy these days but I was not expecting my RFC was so close to be closed, in particular because most of you feedback felt like suggestions on how to _improve_ the RFC.

> the existence of your crate implementing this feature is ironically a valid reason to decline implementing the feature: modifying the compiler and language is a lot of work and if the feature can be supported by a crate instead

I'm explicitely mentioning this and I would have liked it to be discussed, precisely because I see the limitation of a crate like mine (this is stated in the RFC).


> For instance, with function calls, we allow let b = builder(); let b = b.thing(arg); let x = b.build();. And b has a type.

This was proposed as an alternative possibility. The very reason I decided *not* to propose this as the main design, but only as an alternative, was precisely because this behaviour could have also been added *later* on if needed. Current design was *intentionally* forward-compatible with the idea of making every section behave as a real function/method. So again I decided to propose the current design first and to consider evolving the RFCs later if the general feedback had been in favor of the alternative.

> But as far as I can tell, those are still tied into the idea of not invoking any code until all the parameters are present. 

No they were not.

> On top of that, as others have noted: this is a problem that crates in the ecosystem (including yours) can solve

Not the documentation part, which is to me a bigger issue than you think it is.

jackh726:
Hey all, I'm going to lock this thread, since I think what has been said so far covers what is needed to be said.

@contactomorph I'm sorry if you feel like the closing this RFC was an unreasonable step. I would read carefully what others have said because they do a good job of explaining the reasoning behind this.

If you'd still like to discuss this potential feature, then come chat about it on the [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang) that is a *much* better forum to have these conversations than a Github RFC thread.

PR #3981, opened 2026-07-02
RFC: Store registry tokens in the OS credential store by default

quinnjr:
This RFC proposes making Cargo store registry tokens in the operating system's credential store (Keychain, Credential Manager, Secret Service) by default instead of plaintext `credentials.toml`, using a single `cargo:keyring` built-in provider backed by the `keyring` crate.

Key points:

- Default provider list becomes `["cargo:keyring", "cargo:token"]` — keyring first, plaintext file as fallback so headless/CI environments are unaffected.
- Existing plaintext tokens are migrated on any credential read (`cargo login`, `cargo publish`, `cargo yank`, `cargo owner`), not just on re-login, so the installed base actually moves.
- `CARGO_REGISTRY_TOKEN` and `--token` keep precedence and are never migrated.
- Builds on the credential-provider infrastructure from RFC 2730, stabilized in 1.74; this RFC only changes the default.

[Rendered](https://github.com/quinnjr/rfcs/blob/keyring-credential-default/text/0000-keyring-credential-default.md)


text/0000-keyring-credential-default.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The obvious alternative is the status quo: the secure providers exist, just turn them on yourself. But a security measure that requires discovering a config key does not protect the default install, and the default install is the threat model here. Defaults matter precisely because most users never change them, and that cuts both ways. Right now it cuts toward plaintext.
+
+Migrating only on `cargo login` would be simpler, but it leaves every existing user in plaintext indefinitely, since a working token means there's no reason to ever run `login` again. Publish-time migration is what actually moves the installed base.
+
+The Rust Maintainers could also keep the three separate per-platform providers and just change the default list per platform. That works, but it's three implementations to keep behaviorally in sync where one crate already does the job, and it leaves the BSDs out.
```
zachs18: Another alternative could be to print a warning/notice when `cargo publish` (or another subcommand that requires registry authentication) is used interactively with a `credentials.toml`, prompting users to migrate manually (with the default for *new* `cargo login`s using the OS credential store). This wouldn't help much for users who publish infrequently, though.

To aid manual migration, maybe something like `cargo login --bikeshed-migrate-credentials-to-keyring` could be added that doesn't (re)login but just does the migration from `credentials.toml` to the OS key store.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3981/037f71163b768372b28fa715b815b3fce0c9abb0..0d81861b34d0039571459e5916a23fc8c5c299c4)*

text/0000-keyring-credential-default.md
```diff
@@ -0,0 +1,67 @@
...
+
+Entries are keyed by service `cargo-registry:<index-url>` so alternative registries each get their own credential. On Linux the Secret Service connection is made over D-Bus at runtime, not linked at build time, so a missing D-Bus daemon just means the provider reports that no store is available and the chain moves on to the file provider.
+
+The existing three per-platform providers stay for a deprecation period so nobody's explicit config breaks, but they'd eventually become aliases for `cargo:keyring`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Keychain access can prompt the user, which is surprising the first time it happens mid-script, though the env-var and file fallbacks mean scripted environments never hit this in practice. Migrating during `publish` makes that first prompt more likely to appear in the middle of a workflow the user thinks of as non-interactive, which is why the notice needs to be loud about what just happened. It also adds the `keyring` crate (and its platform dependencies) to Cargo's dependency tree, which is real supply-chain surface for a security-sensitive code path and would need the usual vetting.
```
teor2345, line 45: Should there be a timeout for the initial transfer to the keychain, in case the user is running non-interactively, but with a GUI/terminal keychain prompt enabled?

Without a timeout, the build process will appear to freeze, and if the user can't see the reason, it will look like a bug. For example, the prompt could appear on a different (virtual/invisible) screen or terminal.

Similar unexpected prompts have led to denial of service issues, although this one isn't remotely triggerable.

text/0000-keyring-credential-default.md
teor2345: I think a manual migration command would be useful whatever we do here, and also for testing.

text/0000-keyring-credential-default.md
```diff
@@ -0,0 +1,67 @@
...
+
+Migrating only on `cargo login` would be simpler, but it leaves every existing user in plaintext indefinitely, since a working token means there's no reason to ever run `login` again. Publish-time migration is what actually moves the installed base.
+
+The Rust Maintainers could also keep the three separate per-platform providers and just change the default list per platform. That works, but it's three implementations to keep behaviorally in sync where one crate already does the job, and it leaves the BSDs out.
+
+# Prior art
+[prior-art]: #prior-art
+
+Cargo's own credential-process RFC ([RFC 2730](https://rust-lang.github.io/rfcs/2730-cargo-token-from-process.html)) and the 1.74 stabilization laid all the groundwork here. This RFC is really just about the default. Outside Rust: `docker-credential-helpers`, `gh auth`, and Python's `keyring` package all made the same move, and their experience (including the headless-fallback problem) is directly applicable.
```
Eh2406, line 59: You're absolutely right, The other languages and Tools Migrations would be extremely helpful in understanding where the pitfalls are. What did they do? What worked? What didn't? Where can I read more?

Are there other tools that did different migrations or made different decisions? If so what decisions did they do and why?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3981/037f71163b768372b28fa715b815b3fce0c9abb0..0d81861b34d0039571459e5916a23fc8c5c299c4)*

text/0000-keyring-credential-default.md
```diff
@@ -0,0 +1,67 @@
...
+
+Migration happens in the credential-resolution path, not just in `login`. When resolution falls through to the file provider and finds a token, and the keyring provider reported itself available, Cargo writes the token to the keyring, verifies the write by reading it back, and only then removes it from `credentials.toml`. If the keyring write or read-back fails, the file is left untouched and the token is used as before. This ordering means an interrupted migration can duplicate a credential but never lose one. Tokens supplied via `CARGO_REGISTRY_TOKEN` or `--token` bypass migration entirely, since they were never Cargo's to store.
+
+Entries are keyed by service `cargo-registry:<index-url>` so alternative registries each get their own credential. On Linux the Secret Service connection is made over D-Bus at runtime, not linked at build time, so a missing D-Bus daemon just means the provider reports that no store is available and the chain moves on to the file provider.
+
+The existing three per-platform providers stay for a deprecation period so nobody's explicit config breaks, but they'd eventually become aliases for `cargo:keyring`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
teor2345, line 43: Another possible drawback is that the file no longer exists in the filesystem (or at least, it's not  in the same place). This has multiple drawbacks:
1. for backup/migration, if a user has carefully configured their system to backup or transfer everything they need, and they switch to a different device, the credential may be lost on the old machine
2. for shared filesystems, the credential will only be on one machine
3. for shared credential stores, the credential could be exposed to unexpected users

~~Since credentials can be re-created via GitHub, the first two issues seem manageable.~~
Are there any precautions we could take for the third issue?

teor2345 (commented):
There are some surprising security and availability risks here, which it would be good to mention in the drawbacks section, and mitigate if possible.

Overall I support this change though!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3981/037f71163b768372b28fa715b815b3fce0c9abb0..0d81861b34d0039571459e5916a23fc8c5c299c4)*

text/0000-keyring-credential-default.md
teor2345: Actually, only crates.io credentials can be re-created via GitHub.

if this change applies to other registry credentials, then moving those credentials into personal keychains could potentially violate corporate asset or security policies.

kennytm: if the corporate registry mandates use of plaintext tokens they can specify it in the config:

```toml
# .cargo/config.toml

[registries.example]
index = 'https://example.com/index'
credential-provider = ['cargo:token']
```

also i don't think that a threat model that - (1) saving valuable corporate asset (the registry login token) into `~/.cargo/credentials.toml` on employee's personal computer is fine but (2) being placed on their personal keychain is bad - make any sense at all. Either the corporate issue company computers so there's no "personal" keychain in the first place, or that the employee is totally trusted with the token so placing it anywhere secure is ok

text/0000-keyring-credential-default.md
quinnjr: I would agree for testing but I don't agree for the full practice.

Too many people, in my opinion, are going to just never do the migration if it's allowed to be manual. The purpose of the RFC is a forced stance on moving to a safety first approach to benefit users who are less savey about proper security procedures.

text/0000-keyring-credential-default.md
quinnjr: I don't think we need to worry about corporate access policies being an issue.

Prior art (Github CLI, pip/uv opt-in, etc) have already made the change and there isn't much covered engineering-news wise about lashing back from a corporate policy standpoint.

In my 5 last years, moves to secured credentials via company-owned password vaults and mandating things like 2FA, Passkeys on personal devices, etc, has been the norm, not the outlier.

text/0000-keyring-credential-default.md
```diff
@@ -0,0 +1,67 @@
...
+
+If you're on a headless server or in CI, nothing changes at all. `CARGO_REGISTRY_TOKEN` still takes precedence and is never migrated anywhere, and if there's no credential store to talk to, Cargo quietly falls back to the file just like today. Nobody's build breaks because D-Bus isn't running.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Rather than maintaining three separate built-in providers, Cargo's default provider would be a single `cargo:keyring` provider built on the `keyring` crate, which already abstracts over the macOS Security Framework, Windows Credential Manager, and Secret Service (with a kernel keyutils fallback on Linux, and Secret Service coverage on FreeBSD and OpenBSD). The default provider list becomes `["cargo:keyring", "cargo:token"]`: try the keyring, fall back to the file.
```
Eh2406: The cargo team rarely specifies exactly what dependencies we're going to use. I suspect we will be more comfortable merging this RFC if it directly specified the semantics; justifying that those semantics are possible to implement by referencing your recommended implementation.

Something like:
> This RFC Described adding a combined Cross-platform built in provider under a common alias like `cargo:keyring`. This would have the semantics of using The system correct built in provider on any system. This can be implemented by wrapping the existing providers, or building a new provider based on the `keyring` crate.

If we do take your advice and use the `keyring` crate. We will also need to have a deep discussion about what infrastructure supports the ongoing maintenance of that crate. Who is on call for any security problems found? Will they still be around and available in three years or five years?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3981/037f71163b768372b28fa715b815b3fce0c9abb0..0d81861b34d0039571459e5916a23fc8c5c299c4)*

text/0000-keyring-credential-default.md
teor2345: To be clear, I think people should be able to manually migrate *before* they publish.
(In addition to automatic migration on publish.)

quinnjr:
The rationale section of this RFC notes that keeping the credential providers in-tree (rather than adopting the `keyring` crate) is viable only alongside a separate PR closing the BSD gap in Cargo's built-in providers. That PR now exists: rust-lang/cargo#17197 widens the `cargo:libsecret` cfg gate from `target_os = "linux"` to all Unix platforms except macOS and mobile targets, verified on a FreeBSD 15.1 VM with libsecret and gnome-keyring (login/get/logout round-trip through the Secret Service).

Whichever direction this RFC lands on for the default provider implementation, that change removes the BSD platform gap as an adoption blocker.

text/0000-keyring-credential-default.md
```diff
@@ -0,0 +1,69 @@
...
+
+Entries are keyed by service `cargo-registry:<index-url>` so alternative registries each get their own credential. On Linux the Secret Service connection is made over D-Bus at runtime, not linked at build time, so a missing D-Bus daemon just means the provider reports that no store is available and the chain moves on to the file provider.
```
UndeadRiceCrispy: Curious Question on the entry keying: what normalization does `<index-url>` get before it becomes the keyring key? The same registry can be expressed several ways in config >> `sparse+https://index.crates.io/` vs. the legacy git index URL, with or without a trailing slash, and if those don't canonicalize to a single key, migration could write the token under one form while later lookups resolve to another. During the write-verify-delete sequence that's the bad case: the file copy is already gone, the keyring entry exists but under a key lookups don't hit, and the user gets an authentication failure on their next publish with no obvious cause.

Is the intent to reuse Cargo's existing source/registry canonicalization for the key, and if so, is it worth the RFC pinning that down explicitly? Cheap to specify now you'd think, a bit more expensive to discover in a migration bug report.

text/0000-keyring-credential-default.md
UndeadRiceCrispy: Building on the backup/loss point above the sharpest version might be: keyring write succeeded, file deleted, and the keyring entry later becomes inaccessible (locked by policy, encrypted home not mounted, desktop environment switched). Is the recovery path anything other than re-authenticating with the registry? Either answer is fine, but it seems worth the RFC stating it if re-login is the designed recovery, saying so turns a scary-looking failure into a documented one.

UndeadRiceCrispy: @kennytm 

I think both of these points surely land for the plaintext-mandate case but there's a narrower version of the concern that survives them: users who *explicitly* configured `cargo:wincred` (not `cargo:token`) may depend on Windows Credential Manager specifics -> storage paths, ACLs, audit events their compliance tooling watches. When `cargo:wincred` becomes an alias for `cargo:keyring`, is the keyring crate's wincred backend behaviorally identical at the API level, or can the alias silently change audit/ACL behavior under an unchanged config?

If identical, worth the RFC asserting it; if not, the deprecation period becomes a silent behavior change for exactly the users who opted into a specific provider deliberately.

PR #3983, opened 2026-07-15
`f16b` type

Jamesbarford:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3983)*

Add a `f16b` type to Rust in `core::num`, with **no** support for scalar arithmetic (though this is discussed as a future possibility). This has been spun off from the RFC [Add `bf16`, `f64f64` and `f80` types](https://github.com/rust-lang/rfcs/pull/3456)

> [!IMPORTANT]
>

[Rendered](https://github.com/Jamesbarford/rust-rfcs/blob/rfc/bf16/text/3983-bf16.md)


text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
+# `bf16` primitive floating-point type
```
kennytm: the "Feature Name / Start Date / RFC PR / Rust Issue" header is gone?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+For a literal suffix `0bf16` may clash with the lexer's integer suffix `0b` for binary numbers thus;
+
+```rust
+let a = 0bf16;
+```
+
+Would need to be written;
+```rust
+let a = 0.0bf16;
+```
```
kennytm: It could also be disambiguated as `0_bf16`.

Is the "unresolved" thing about whether to support the literal `0bf16`? There is no other ways this conflict may arise. Since this is just a single special case IMO it is not harmful to also support this. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

clarfonthey: To be honest, I had no idea that `0f64` and similar worked by itself. I would have expected the decimal part to be required always, as it is when the type isn't specified.

This feels particularly dubious considering how `0x1f64` is acceptable, but refers to the literal 8036 and not `1.0f64`; `0b1f64` rightfully errors. Personally, I would just say maybe a good option to consider is just always requiring the decimal part for float literals just to avoid the need to disambiguate.

Or, at least, in this particular case, allow both `0_bf16` and `0.0bf16` but always lint float literals that don't have a decimal part, due to the dubiousness.

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+Add a primitive type, `bf16`, to Rust for the 16-bit **Brain Floating Point** format.
+
+`bf16` is a 16-bit floating-point format derived from `f32`. It keeps the same 1-bit sign and 8-bit exponent layout as `f32`, but stores only the top 7 fraction bits. This gives it approximately the same dynamic range as `f32`, but much lower precision. The primary role of `bf16` is representing bfloat16 values in memory, in APIs, in FFI signatures where the target ABI supports bfloat16, and as an element type for SIMD and matrix operations. It is expected to be uncommon for a `bf16` to be used as a general-purpose arithmetic type in the same way as `f32` or `f64`.
+
+`bf16` is not an IEEE-754 binary floating-point format however this RFC specifies the proposed `bf16` type as following portable IEEE-754 semantics for conversions and scalar arithmetic. These are specified as as widening to `f32`, performing the corresponding `f32` operation, and narrowing back to `bf16` using round-to-nearest, ties-to-even (RNE) without flushing subnormals to zero. As such, implementations may use native `bf16` instructions for scalar arithmetic and scalar conversion only where they produce identical results.
```
tgross35: There's a double "as as" here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+Increasingly `bf16` is found in hardware across architectures to support these workloads. The below is not an exhaustive list but illustrative of `bf16`'s prevalence in hardware.
+
+On AArch64, `bf16` is supported by Armv8.6-A and later, implementations that include the `bf16` architectural extension, include platforms based on [Neoverse V1][Neoverse V1], [Neoverse N2][Neoverse N2], and [Neoverse V2][Neoverse V2]. On x86_64, `bf16` is supported through `AVX512_BF16` on platforms such as [Intel Cooper Lake][Intel Cooper Lake], [Intel Sapphire Rapids][Intel Sapphire Rapids] and later platforms with `AMX_BF16`, and [AMD Zen 4][AMD Zen 4] and later platforms with `AVX-512 BF16` support. On [RISC-V][RISC_V], `bf16` is supported through the ratified `Zfbfmin`, `Zvfbfmin`, and `Zvfbfwma` extensions, providing scalar conversion, vector conversion, and vector widening multiply-accumulate respectively.
+
+Consequently, `bf16` is not limited to accelerator hardware; it is a natively supported scalar/vector format on modern CPUs, with availability discoverable through target-feature mechanisms.
```
tgross35: I think the chip list can go in an appendix or footnotes (or link to wikipedia instead), IMO that's not especially interesting as long as the extension has been around for a while. Maybe it would be helpful to have something tabular in the reference-level explanation? E.g.:

> * aarch64 `foofeature`: ...
> * aarch64 `barfeature`: ...
> * x86 `bazfeature`: ...
> * RISC-V `Zfbfmin`: scalar `f32` <-> `bf16`, ...
> * RISC-V `Zvfbfmin`: vector `f32` <-> `bf16`, ...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+  | S | E E E E E E E E | M M M M M M M | discarded lower f32 frac |
+  +---+-----------------+---------------+--------------------------+
+    \___________________ _______________/
+            kept as bf16 high 16 bits
+
+  bf16:
+  +---+-----------------+---------------+
+  | S | E E E E E E E E | M M M M M M M |
+  +---+-----------------+---------------+
+```
```
tgross35: This seems more fitting for an appendix or external link than guide-level explanation

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+For more details, see reference-level explanation.
+
+### Scalar arithmetic
+
+The main motivation for `bf16` is SIMD and matrix support, not scalar arithmetic. Scalar `bf16` operations are specified portably as widening operands to `f32`, performing the corresponding Rust `f32` operation, then narrowing the result to `bf16` using round-to-nearest, ties-to-even (RNE). Subnormal results are not flushed to zero by default.
+
+Implementations may use native `bf16` instructions only when they produce the same observable result. Instructions with different rounding, flushing, semantics remain available through target-specific intrinsics. NaN payload behaviour is not strengthened beyond Rust's existing floating-point guarantees from [RFC-3514][RFC-3514].
+
+For more details, see reference-level explanation.
```
tgross35: For the guide-level explanation I think it's more relevant to say:

1. What math traits will be implemented?
2. What `as` casts will be supported?
3. There will be vector versions in `core::arch`
4. Do we guarantee IEEE semantics?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+## Reference-level explanation
+
+### ABI
+
+This RFC does not propose a new platform ABI for `bf16`.
+
+On AArch64, [the AAPCS64 specification defines `__bf16` as the Brain floating-point format with byte size 2 and natural alignment 2][aapcs64-bf16]. On x86_64, `bf16` ABI compatibility should likewise follow the platform psABI and toolchain support for `__bf16`, which is stated as having ["the same size, alignment, parameter passing and return rules as _Float16"][x86_64-bf16].
+
+Within Rust, `bf16` shall have a size 2 and alignment 2, all bit patterns are valid values, and its memory representation is the 16-bit Brain floating-point encoding described above. These are properties of the Rust type itself, not a new foreign calling convention.
```
tgross35: Is this something we want/need to guarantee? I think it's fine to say it will _usually_ size/alignment 2, but not tie ourselves into the alignment in case some platform does something different.

Also the last sentence seems a bit redundant, I'm not sure "not a new foreign calling convention" is needed

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+With a primitive `bf16` type, Rust APIs can model these operations using the natural element type of the source operands:
+
+```rust
+// Example Neon intrinsic that would be added to `std::arch::aarch64`
+fn vbfdot_lane_f32<const LANE: i32>(r: float32x2_t, a: bfloat16x4_t, b: bfloat16x4_t) -> float32x2_t;
+```
+
+Without a scalar `bf16` type, these APIs must instead represent `bf16` data using `u16`, wrapper types, or backend-specific encodings. Using `u16` preserves the raw bits, but it loses the meaning of the value. It also cannot use the usual floating-point surface of the language, such as casts, literals, constants, formatting, arithmetic traits, or generic bounds over floating-point types. That makes the APIs less clear, less portable, and less directly connected to the hardware operations they expose. If adopting a wrapper approach like `core::arch::x86_64::bf16` Rust would need a `bf16` type per architecture, this could cause confusion especially if names for the type diverged.
```
tgross35: Talking about tradeoffs against `u16` and doesn't seem like it should be in the reference-level explanation about vectors, could this be narrowed down to discuss what we're actually adding?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

tgross35: Actually these tradeoffs come up 3-4ish assorted places, they should be consolidated to motivation and/or rationale+alternatives.

Also were we seriously thinking of using `u16` in `std::arch` to represent `bf16`? I'm not sure that's really worth much attention.

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+Therefore, target-specific SIMD, dot-product, and matrix APIs should be understood as exposing the hardware operation, not as promising target-independent bit-exact results. Portable Rust APIs may use `bf16` as the element type for clarity and type safety, but portability of the source-level API should not be read as portability of every numerical detail of the underlying target-specific instruction. This is true today with intrinsics for "packed/vector float min" for example;
+
+```
+x86_64: _mm_min_ps(-0.0, +0.0) -> +0.0 bits 0x00000000  
+AArch64: vminq_f32(-0.0, +0.0) -> -0.0 bits 0x80000000
+```
+
+While ostensibly similar they produce different results.
+
+### Scalar Arithmetic
```
tgross35: Maybe move this above vector arithmetic since it's more fundamental?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+By contrast, x86_64's [`VDPBF16PS`][intel_a64] specifies `bf16` dot products accumulated into `f32` using round-to-nearest ties to even for the FMA accumulation, while also treating input denormals as zero, flushing output denormals to zero, and ignoring `MXCSR`.
+
+Therefore, target-specific SIMD, dot-product, and matrix APIs should be understood as exposing the hardware operation, not as promising target-independent bit-exact results. Portable Rust APIs may use `bf16` as the element type for clarity and type safety, but portability of the source-level API should not be read as portability of every numerical detail of the underlying target-specific instruction. This is true today with intrinsics for "packed/vector float min" for example;
+
+```
+x86_64: _mm_min_ps(-0.0, +0.0) -> +0.0 bits 0x00000000  
+AArch64: vminq_f32(-0.0, +0.0) -> -0.0 bits 0x80000000
+```
+
+While ostensibly similar they produce different results.
```
tgross35: Are we going to hit cases like https://github.com/rust-lang/rust/issues/129880? Cc @RalfJung

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+### Scalar Arithmetic
+
+While this section is detailed, the primary use case of `bf16` is **not** expected to be scalar arithmetic but SIMD/matrix support. Moreover a compelling argument can be made for **not** making scalar arithmetic portable which is laid out in the [rationale and alternatives](#scalar-arithmetic-alternatives) section. So too has a case for not including scalar arithmetic at all. Nevertheless this RFC proposes portable semantics.
```
tgross35: It's valid enough to store `bf16` in a vector instead of `f32` to save some space. It might be more accurate to phase this as doing arithmetic on a bf16 directly vs. via f32 or something in `arch`, rather than scalar vs. vector.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+For narrowing an `f32` to `bf16`, the conversion shall use the IEEE-754-2019 [default rounding-direction attribute for binary floating-point results, namely `roundTiesToEven`][ieee-754]. For subnormals the default delivers a rounded result that may or may not be zero. In other words subnormals are not flushed to zero by default.
+
+This is a semantic specification, not an implementation requirement. An implementation may use any instruction sequence, including native `bf16` instructions, only when the observable result is identical to the specified model for all non-NaN inputs. The result of scalar `bf16` arithmetic must not depend on whether the target has native `bf16` arithmetic, on optimisation level, or on whether the expression is evaluated at compile time or run time.
```
tgross35: This reads as though it only applies to the narrowing operation, but I think it's meant for all scalar ops?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+AArch64: vminq_f32(-0.0, +0.0) -> -0.0 bits 0x80000000
+```
+
+While ostensibly similar they produce different results.
+
+### Scalar Arithmetic
+
+While this section is detailed, the primary use case of `bf16` is **not** expected to be scalar arithmetic but SIMD/matrix support. Moreover a compelling argument can be made for **not** making scalar arithmetic portable which is laid out in the [rationale and alternatives](#scalar-arithmetic-alternatives) section. So too has a case for not including scalar arithmetic at all. Nevertheless this RFC proposes portable semantics. 
+
+For narrowing an `f32` to `bf16`, the conversion shall use the IEEE-754-2019 [default rounding-direction attribute for binary floating-point results, namely `roundTiesToEven`][ieee-754]. For subnormals the default delivers a rounded result that may or may not be zero. In other words subnormals are not flushed to zero by default.
```
tgross35: Maybe instead of this add a note at the end of the arithmetic section saying `f32`<->`bf16` conversions for the purpose of arithmetic are the same as `f32`<->`bf16` semantics, rather than repeating something similar to that section here.

"by default" also seems to imply that FTZ can be enabled.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+This is a semantic specification, not an implementation requirement. An implementation may use any instruction sequence, including native `bf16` instructions, only when the observable result is identical to the specified model for all non-NaN inputs. The result of scalar `bf16` arithmetic must not depend on whether the target has native `bf16` arithmetic, on optimisation level, or on whether the expression is evaluated at compile time or run time.
+
+Scalar arithmetic for `bf16` where no native scalar instruction are present are specified as if both operands are first converted exactly to `f32`, the corresponding Rust `f32` operation is performed, and the resulting `f32` value is then narrowed to `bf16` using round-to-nearest, ties-to-even (RNE). This keeps things portable between targets and the rounding mode the same as the IEEE-754 specification.
```
tgross35: "Scalar arithmetic for `bf16` where no native scalar instruction are present"

Are the semantics different when they are present? (I hope not)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+let c = a + b;
+```
+
+Which is semantically equivalent to:
+
+```rust
+let c = (f32::from(a) + f32::from(b)) as bf16;
+```
+
+Where the `as bf16` cast performs the narrowing under round-to-nearest, ties-to-even, as specified above.
```
tgross35:
````suggestion
In other words, this code:

```rust
let a = 1.1bf16;
let b = 2.2bf16;

let c = a + b;
```

Must be semantically equivalent to:

```rust
let c = (f32::from(a) + f32::from(b)) as bf16;
```
````

Those are just our regular semantics right? I don't think we need to specify or break them down further.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

File: text/3983-bf16.md [hunk 16]
```diff
@@ -0,0 +1,521 @@
...
+
+As a direct consequence:
+
+- Native `bf16` arithmetic operations that round directly to `bf16`, rather than producing the same result as the specified through the above sequence, are **not** legal lowerings for scalar `bf16` arithmetic. They produce different result bits for some inputs. Such operations remain available through explicit target-specific intrinsics, where the user has opted in to native semantics.
+- The narrowing step always rounds to nearest, ties-to-even. Truncation is not a conforming narrowing.
+- The arithmetic performs two IEEE rounding steps: the inner operation rounds to `f32` precision, and the narrow then rounds that result to `bf16` precision.
+
+The two-step rounding may produce results that differ by one [ULP][ULP_WIKI] from a hypothetical single-rounded native `bf16` operation for a small fraction of inputs. This is a deliberate trade-off in favour of portable, bit-exact semantics; targets without native `bf16` arithmetic could not otherwise be supported uniformly. Code that requires single-rounded native semantics should use target-specific intrinsics.
```
tgross35: > targets without native `bf16` arithmetic could not otherwise be supported uniformly

They could - if an asm lowering matching needed semantics isn't available, use a libcall. That's how other float types work.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+bf16::MIN
+bf16::MIN_10_EXP
+bf16::MIN_EXP
+bf16::MIN_POSITIVE
+bf16::NAN
+bf16::NEG_INFINITY
+bf16::RADIX           // 2
+```
+
+The `half` crate documents `bf16::MANTISSA_DIGITS` as 8, including the implicit bit, and `DIGITS` as 2 decimal digits. However, a crate-defined type cannot fully substitute for a primitive type. Though its use is demonstrative of users' need for a `bf16` type.
```
tgross35: This is another motivation within reference-level explanation that isn't mentioned in motivation / rationale

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+- `Add`, `Sub`, `Mul`, `Div`, `Rem`
+- `Neg`
+- `PartialEq`, `PartialOrd`
+- `Copy`, `Clone`, `Default`
+- `Debug`, `Display`, and the other formatting traits
+- `Sum`, `Product`
+- `FromStr`
+
+These scalar operations must be implemented in conformance to the widen -> `f32` operations -> narrow using RNE as specified above.
```
tgross35: This could all probably go in the guide-level explanation, it's API surface. Also consider `TryFrom`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+`bf16` follows the same general pattern as the other primitive floats, supporting:
+
+- `Add`, `Sub`, `Mul`, `Div`, `Rem`
+- `Neg`
+- `PartialEq`, `PartialOrd`
+- `Copy`, `Clone`, `Default`
+- `Debug`, `Display`, and the other formatting traits
+- `Sum`, `Product`
+- `FromStr`
```
tgross35: "Traits" is nested under "Conversions"

(I don't think the `####` section headers within "Conversions" are adding much anyway, they're pretty tiny sections)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+#### The Tradeoffs
+
+A path only `core::num::bf16` does not consume a globally reserved name which signals that this is a specialist interchange type rather than a general-purpose arithmetic type which arguably a `bf16` could be. This could also set a precedent for future numeric formats without committing prelude real estate each time. If we want `bf16` to be something that is only used as an exchange format then `use core::num::bf16;` could be a more natural fit.
+
+However a global namespace is consistent with `f16/f128` and the rest of the primitives, requires no import, and is what users would need if they are wanting to suffix floating point numbers with `bf16` to define `bf16`'s. Though whether this is desirable needs to be ironed out.
+
+This decision is reversible across an edition. Prelude membership is edition-scoped, so `bf16` can ship path-only and later be added to a future edition's prelude (or the reverse) without breaking existing code. This RFC proposes a global namespace first.
```
tgross35: Let's not plan around introducing something only to back it out

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+An alternative is to make `bf16` a compiler-known primitive that is reached by path, as `core::num::bf16`, and is not added to the prelude. This mirrors the placement proposed for `Complex` in [RFC 3892][RFC-3892], which positions a numeric interchange type in `core::num` rather than the global namespace.
+
+#### The Tradeoffs
+
+A path only `core::num::bf16` does not consume a globally reserved name which signals that this is a specialist interchange type rather than a general-purpose arithmetic type which arguably a `bf16` could be. This could also set a precedent for future numeric formats without committing prelude real estate each time. If we want `bf16` to be something that is only used as an exchange format then `use core::num::bf16;` could be a more natural fit.
+
+However a global namespace is consistent with `f16/f128` and the rest of the primitives, requires no import, and is what users would need if they are wanting to suffix floating point numbers with `bf16` to define `bf16`'s. Though whether this is desirable needs to be ironed out.
```
tgross35: Counter: `bf16` is kind of a weird type and probably shouldn't be preferred over one of the `f*` types unless a specific usecase is met. Keeping it from being available by default guides users in the right direction.

I believe `core::arch:bf16` was proposed before as well, along with not allowing the `bf16` suffix.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+Portable scalar arithmetic could prove useful if predictable behaviour is expected. Baking this into the language could lessen the possible footguns from the myriad of disparate, and somewhat confusing, hardware support available. Thus inefficiencies in using this type for scalar arithmetic is deemed an acceptable tradeoff favouring predictable behaviour for a feature that is not anticipated to be the main usecase of the type.
+
+### Naming `bf16`
+
+The name `bf16` maps directly to hardware-provider naming conventions, is a simple abbreviation of the official name [`bfloat16`][bfloat16_name], is similar to the [`__bf16` extension name in C++][__bf16_extension_name] and is immediately distinguishable from the similarly named `f16` by the `b` prefix, meaning `brain` and `f` denoting `float`. This abbreviation also follows Rust's naming conventions for the currently supported numerical types where `float` is abbreviated with `f` or `integer` with an `i`.
```
tgross35: When we talked about this last I believe `bf16`, `f16b`, `fb16`, `bfloat16`, and `__bf16` all had some thought. I agree that `bf16` is fine, just noting that there might be some alternatives floating around the other RFC.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+This decision is reversible across an edition. Prelude membership is edition-scoped, so `bf16` can ship path-only and later be added to a future edition's prelude (or the reverse) without breaking existing code. This RFC proposes a global namespace first.
+
+### Scalar Arithmetic Alternatives
+
+Having a `bf16` type in the global namespace could reasonably set the expectation of having scalar operations baked in. All other globally available primitive numerical types have scalar operations. However omitting scalar arithmetic for a `bf16` means that the rather complicated and, depending on support, possibly inefficient scalar operations could be omitted for `bf16` which makes the type simpler and prevents a user being confused with the resultant code of widening to `f32`, performing the operations and narrowing to `bf16` along with preventing misuse of the type. It seems reasonable a user could utilise the `From` and `as` casts themselves, perform scalar arithmetic in an `f32` and convert back to `bf16` without relying on built in support. Moreover vendor specific intrinsics could be used, where available, to perform scalar operations.
```
tgross35: > However omitting scalar arithmetic for a `bf16` means that the rather complicated and, depending on support, possibly inefficient scalar operations could be omitted for `bf16`

Omitting scalar arithmetic means scalar operations could be omitted - huh?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

tgross35: Honestly I'm really struggling to understand this paragraph and think it could be rewritten

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+The `usize` precedent is weaker than first appears as `usize` is deterministic per target: given a target triple its behaviour is fully specified and stable. Its variation is structural 32-bits vs 64-bits and the type system prevents it being mixed with fixed sized integers. However `bf16` would vary on the same target triple depending on what target features have been made available. This problem is described in the [Prior Art](#c-compiler-support).
+
+Portable scalar arithmetic could prove useful if predictable behaviour is expected. Baking this into the language could lessen the possible footguns from the myriad of disparate, and somewhat confusing, hardware support available. Thus inefficiencies in using this type for scalar arithmetic is deemed an acceptable tradeoff favouring predictable behaviour for a feature that is not anticipated to be the main usecase of the type.
```
tgross35: I don't think `usize` is any kind of precedent here - a different type size isn't remotely close to the complexities introduced by IEEE vs. FTZ or other similar variation.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

tgross35: Why was it determined that we should have bf16->f32->bf16 semantics rather than bf16 semantics? I would expect that most users either want:

1. Do all operations at `f32` without spending time at intermediate rounding, which I believe C does
2. Correct bf16 semantics because you're doing something mathy

`f32` with intermediate rounding gives predictable results but I'm not sure it brings an advantage over either of those two. _(edit: does the difference even matter for the specified ops? https://github.com/rust-lang/rfcs/pull/3983#discussion_r3594160660)_

My preference would be to require bf16 semantics by default, then add methods like [`algebraic_add`](https://doc.rust-lang.org/std/primitive.f32.html#method.algebraic_add) that can do entire operations as `f32`. That is, `a + b + c` only ever gives you a correctly rounded `bf16` result, but `a.algebraic_add(b).algebraic_add(c)` could get lowered to any of the following:

```rust
// simplest, fastest on soft float targets and if there is hardware supporting
// bf16 arithmetic directly
addbf16(addbf16(a, b), c); 

// semantics defined in this RFC: allowed but unlikely
addf32(addf32(a as f32, b as f32) as bf16 as f32, c as f32) as bf16;

// fastest with hardware f32
addf32(addf32(a as f32, b as f32), c as f32) as bf16;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

tgross35:
> This has been spun off from the RFC [Add `bf16`, `f64f64` and `f80` types](https://github.com/rust-lang/rfcs/pull/3456)

Pinging @ecnelises who authored that RFC

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+### ABI
+
+This RFC does not propose a new platform ABI for `bf16`.
+
+On AArch64, [the AAPCS64 specification defines `__bf16` as the Brain floating-point format with byte size 2 and natural alignment 2][aapcs64-bf16]. On x86_64, `bf16` ABI compatibility should likewise follow the platform psABI and toolchain support for `__bf16`, which is stated as having ["the same size, alignment, parameter passing and return rules as _Float16"][x86_64-bf16].
+
+Within Rust, `bf16` shall have a size 2 and alignment 2, all bit patterns are valid values, and its memory representation is the 16-bit Brain floating-point encoding described above. These are properties of the Rust type itself, not a new foreign calling convention.
+
+For `extern "C"` and other foreign ABIs, Rust should not independently choose how `bf16` is passed or returned. Instead, on targets where the platform defines and toolchains implement an ABI for the corresponding C/C++ `__bf16` type, Rust's `bf16` should be ABI-compatible with that type. On platforms where the ABI is undefined, see [Unresolved Questions](#unresolved-questions).
```
tgross35: Do GCC and Clang agree on targets they support? This kind of thing has been most of the `f16` bugs preventing stabilization thus far.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
kennytm: I think `0_bf16` is clear enough to say it is a "0" of type `bf16`. Use lint [`clippy::unseparated_literal_suffix`](https://rust-lang.github.io/rust-clippy/master/index.html?search=literal#unseparated_literal_suffix) against `0bf16`. There is no lint requiring like `5f32` must be written as `5.0f32` though.

text/3983-bf16.md
RalfJung: That depends on how these operations are represented in LLVM IR.

It is also unclear to me what this part of the RFC is even talking about. It would help to concretely say which operations this is about -- I think it refers to things in `core::arch`, but that should be clarified. I would find it concerning if anything outside `core::arch` had any target-specific behavior.

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+## Guide-level explanation
+
+### Naming the 16-bit Brain floating-point `bf16`
+
+The proposed name for the Brain Floating Point type is `bf16`. See rationale as to why.
+
+### Usage
+
+This RFC proposes the `bf16` type to behave like the other primitive floating-point types in Rust, with standard traits implemented. It can be used in bindings, passed to functions, stored in arrays and structs, converted to and from other numeric types, and used with the usual floating-point operations. It follows the IEEE-754 standard for floating point types for scalar operations and only diverge when using vendor intrinsics gated by a target feature. Details for scalar and vector operations are discussed in greater depth below.
```
RalfJung:
```suggestion
This RFC proposes the `bf16` type to behave like the other primitive floating-point types in Rust, with standard traits implemented. It can be used in bindings, passed to functions, stored in arrays and structs, converted to and from other numeric types, and used with the usual floating-point operations. It follows the IEEE-754 standard for floating point types for scalar operations (albeit with a non-standard binary format) and only diverge when using vendor intrinsics gated by a target feature. Details for scalar and vector operations are discussed in greater depth below.
```
That binary format should be described in the Reference-level section.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
```diff
@@ -0,0 +1,521 @@
...
+
+The main motivation for `bf16` is SIMD and matrix support, not scalar arithmetic. Scalar `bf16` operations are specified portably as widening operands to `f32`, performing the corresponding Rust `f32` operation, then narrowing the result to `bf16` using round-to-nearest, ties-to-even (RNE). Subnormal results are not flushed to zero by default.
+
+Implementations may use native `bf16` instructions only when they produce the same observable result. Instructions with different rounding, flushing, semantics remain available through target-specific intrinsics. NaN payload behaviour is not strengthened beyond Rust's existing floating-point guarantees from [RFC-3514][RFC-3514].
+
+For more details, see reference-level explanation.
+
+## Reference-level explanation
+
+### ABI
```
RalfJung, line 103: Given that ABI is a very niche aspect that we usually make basically no guarantees about, this is an odd way to start the reference-level section. I would have expected it to start by saying:
- There's a new type, this is its representation
- This are the main (scalar) operations on the type, and these are their specs

Start with the basics, only then descend into the weeds by discussing vector operations and ABI concerns.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
RalfJung: Also this goes without saying, that's how our spec works everywhere.

text/3983-bf16.md
RalfJung: Why is this paragraph hidden somewhere in the middle of the section? It's the main thing that this section needs to convey.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

text/3983-bf16.md
RalfJung: This paragraph makes basically the same point as the item above

> Native `bf16` arithmetic operations that round directly to `bf16`, rather than producing the same result as the specified through the above sequence, are **not** legal lowerings for scalar `bf16` arithmetic.

The two should be merged. Currently the RFC uses way more words than necessary to convey its information, making it unnecessarily hard to find the key points.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..3662bd79407aebc7d07558df3ecfc4fc2b3e69ec)*

RalfJung: Oh wow. This is extremely surprising. Does the [double rounding theorem](https://dl.acm.org/doi/pdf/10.1145/221332.221334) not apply here? bf16 has less than half the mantissa of f32 so I would have expected that it applies.

Do you have a concrete example of inputs where the two specs differ?

tgross35: I thought this is only referring to FMA and sanity checked that indeed that is the only one that fails https://gist.github.com/tgross35/a91c60bd655e24f912c7c7659a53cefd. But hm - I'm reading again and `mul_add` doesn't seem to be mentioned.

Jamesbarford: > Oh wow. This is extremely surprising...

I need to check this, for some reason I thought it wouldn't however I think I'm wrong.

programmerjake: imo `mul_add` should be defined to work directly on `bf16` so you get the extra precision people expect, ~~the fallback implementation is just `(a as f64).mul_add(b as f64, c as f64) as bf16`~~ (edit: turns out to be more complicated than that for `mul_add`). defining all of add/sub/mul/div/sqrt to work directly on `bf16` is equivalent to defining them to convert to `f32` before the op and round back to `bf16` after using nearest-ties-to-even mode.
for the other ops (except maybe reciprocal sqrt), defining them as using `f32` arithmetic and convering back to `bf16` after every op seems fine to me, since those aren't generally expected to provide bit-exact answers anyway (and where they are, `bf16` can generally also represent those exact answers)

RalfJung: Yeah fully agreed, if it's just mul_add then we should be consistent with the other float types and require maximal precision. sqrt should also get maximal precision since it is a "core" IEEE operation. For transcendental functions we already make no precision guarantees so there's no reason to do anything else here.

Overall then this becomes an entirely normal IEEE-style float type, except that the binary format is not listed in the standard. But most of the standard is parametric over the format so that's not a big deal.

text/3983-bf16.md
ds84182: I feel the same way with the other proposed floating point types (f64f64 and f80). f16 is much more likely to be supported (x64 F16C (2009!), universal GPU support, RISC-V Zfh, ARM(?)). no comment on f128 though, haha.

If bf16 and its oddball friends become new "primitive" types in the prelude I'll make a (joke) RFC to add [f24](https://www.3dbrew.org/wiki/GPU/Shader_Instruction_Set#Floating-Point_Behavior).

Jamesbarford: @ds84182 As noted in the RFC, `bf16` is supported by multiple hardware vendors and is widely used in ML and AI workloads. Intel is also adding further support in upcoming instruction-set extensions; see [Chapter 7 of the Intel® Advanced Vector Extensions 10.2 Architecture Specification](https://www.intel.com/content/www/us/en/content-details/919693/intel-advanced-vector-extensions-10-2-intel-avx10-2-architecture-specification.html).

For the other types you mentioned, to the best of my knowledge:

* `f64f64` has native hardware support only on PowerPC
* `f80` has native hardware support only on x86
* `f24` has no native support from major hardware vendors

In my view, the combination of broad hardware support and substantial real-world use is enough to distinguish `bf16` from those types. C++ has reached a similar conclusion by standardising `std::bfloat16_t`.

The RFC covers these arguments in more detail, so I will avoid repeating all of them here.

programmerjake: > For the other types you mentioned, to the best of my knowledge:
> 
> * `f64f64` has native hardware support only on PowerPC

it's not really native support, it's just calling standard library functions that use the normal `f64` instructions to implement `f64f64`

> * `f80` has native hardware support only on x86

iirc it's also supported on m68k

text/3983-bf16.md
tgross35: First relevant LLVM bug I noticed, probably of many: incorrect bf16 fma https://github.com/llvm/llvm-project/issues/131531.

I assume there's an easy fix by changing that to a libcall, which could likely use the same algorithm as https://github.com/rust-lang/compiler-builtins/blob/dfe80e1d870b56f34a52bd8b3ce47479cdb95a2d/libm/src/math/generic/fma_wide.rs (for fmaf32 via f64 unfused ops)

programmerjake: thanks for spotting the fma bug, apparently I didn't think hard enough when I said `f64` is enough.

VitWW:
I think should be an unification of **float** types names: thy should all starts with "**f**" letter.
And the name should have a marker "**x**" of alternative **exponent** to IEEE standard.
So, the better _default_ name for our type  is `fxb16` - it is clear that it is float type with 16 bit representation with alternative to IEEE standard format.

```rust
let a: fxb16 = 6.7;
let b: fxb16 = 42.69;
let c = a + b;
println!("{}", c);
```

text/3983-bf16.md
Jamesbarford: I've moved what was written here with a view of following it up shortly with a table as you have suggested

text/3983-bf16.md
Jamesbarford: Moved to the appendix :+1:

text/3983-bf16.md
Jamesbarford: I've softened the language on this

text/3983-bf16.md
Jamesbarford: I've moved all of the tradeoff's vs `u16` comments that were peppered throughout the RFC into the Rationale & Alternatives section

text/3983-bf16.md
Jamesbarford: I've reworded this to improve the clarity

text/3983-bf16.md
Jamesbarford: I've shortened this section considerably as it was unnecessarily verbose.

However I think it's worth splitting Scalar vs Vector as the portability with Scalar operations seem quite different from the concerns of Vector operations which are in essence saying they do what the hardware does. Or perhaps I'm missing your point?

text/3983-bf16.md
Jamesbarford: I've re-done these paragraphs to clarify.

text/3983-bf16.md
Jamesbarford: I've reworked the entire section and removed "default".

The only way to change the rounding mode or FTZ behaviour would be through manipulating something like the `FPCR` which feels separate from this RFC

text/3983-bf16.md
Jamesbarford: Re-done the section. No they should produce the same results.

text/3983-bf16.md
Jamesbarford: I've removed the nesting, I think it could be quite useful for navigating the document in the UI to have these headings? I'm quite happy to remove them if you think they don't add anything

text/3983-bf16.md
Jamesbarford: > Counter: bf16 is kind of a weird type and probably shouldn't be preferred over one of the f* types unless a specific usecase is met. Keeping it from being available by default guides users in the right direction.
>
> I believe core::arch:bf16 was proposed before as well, along with not allowing the bf16 suffix.

I've added having no `bf16` suffix to Rationale/Alternatives

text/3983-bf16.md
Jamesbarford: Added those names

text/3983-bf16.md
Jamesbarford: I've reworded it. Somehow it made sense to me when I initially wrote it :sweat_smile:

text/3983-bf16.md
Jamesbarford: True, it is however a primitive type that has different behaviours depending on the target. However if you think this is too much of a reach, I'm happy to remove this point

Jamesbarford: > Why was it determined that we should have bf16->f32->bf16 semantics rather than bf16 semantics

From trawling around I could not find any `FADD BF16_REG, BF16_REG` of `BF16ADD REG, REG` type instruction on any hardware. I subsequently looked at what GCC and Clang do. They convert the operands to `f32` and cast back to `bf16`, how they do this differs and is described in the C++ section.

So I went down the avenue of `bf16->f32->bf16`.

I've tried to clarify the point I'm now making that if a `FADD BF16_REG, BF16_REG` exists and it has IEEE-754 semantics, we would use that. Otherwise we would do `bf16->f32->bf16`.

With regard to making a `a+b+c+d` expression all `f32` and then converting the result back to `bf16` I'd have thought the results could possibly differ from `bf16->f32->bf16`?

text/3983-bf16.md
Jamesbarford: I've moved the section headers around, I hope that addresses your concern

text/3983-bf16.md
RalfJung: `usize` just has a target-dependent *size*, the behavior is otherwise fully portable and identical to u32 or u64.

text/3983-bf16.md
Jamesbarford: For this RFC, are the complexities surrounding `mul_add` something that would belong in the appendix section as opposed to the main body? Where the resolution would be to emit a library call for this op?

text/3983-bf16.md
```diff
@@ -0,0 +1,505 @@
...
+By contrast, x86_64's [`VDPBF16PS`][intel_a64] specifies `bf16` dot products accumulated into `f32` using round-to-nearest ties to even for the FMA accumulation, while also treating input denormals as zero, flushing output denormals to zero, and ignoring `MXCSR`.
+
+Therefore, target-specific SIMD, dot-product, and matrix APIs should be understood as exposing the hardware operation, not as promising target-independent bit-exact results. Portable Rust APIs may use `bf16` as the element type for clarity and type safety, but portability of the source-level API should not be read as portability of every numerical detail of the underlying target-specific instruction. This is true today with intrinsics for "packed/vector float min" for example;
+
+```
+x86_64: _mm_min_ps(-0.0, +0.0) -> +0.0 bits 0x00000000  
+AArch64: vminq_f32(-0.0, +0.0) -> -0.0 bits 0x80000000
+```
+
+While ostensibly similar they produce different results.
```
RalfJung: This entire paragraph is basically just saying "the only types and operations for vectors of bf16 are provided by core::arch (for now), and as usual for those operations, their semantics are architecture-specific", right?

I feel like that point can be made in fewer than 6 paragraphs. ;)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..0e1f5c4d79429158fc6cfe3a66f59de0b936d9a8)*

text/3983-bf16.md
RalfJung: The question was about the semantics, not about how it is compiled. So I am quite confused by the answer. We are not specifying our operations in terms of assembly instructions after all, so sentences like "if instruction X exists and has behavior Y then it may be used" are entirely out of place here.

What I hope the RFC will say is just: All scalar operations behave according to the IEEE 754-2019 specification when instantiated with the parameters for the bf16 format. In particular, they will return the value representable in the bf16 format that is closest to the infinite-precision result, using roundTiesToEven and fully making use of subnormal numbers. End of spec, that already says everything there is to say in the reference section about the semantics of the scalar operations.

In some other section (appendix?) it may then be worth talking about how those semantics could actually be implemented on the most relevant architectures. For all operations except FMA, it can be implemented by casting to f32 and doing the operations there. (That was the conclusion, right?) FMA will need a more complicated fallback if there is no hardware support. But people like me that care primarily about the opsem and not the asm can then just skip that other section. ;)

text/3983-bf16.md
RalfJung: (This thread overlaps a lot with https://github.com/rust-lang/rfcs/pull/3983#discussion_r3593178086 )

> are the complexities surrounding mul_add something that would belong in the appendix section as opposed to the main body?

It depends. What is the spec for mul_add? If it is "we require the result to be as-if the mul and add are done in infinite precision with a single roundTiesToEven step to bf16 at the end" then there is no complexity for mul_add on the spec/Reference level and that discussion should be moved to a clearly separate section that discusses implementation concerns.

text/3983-bf16.md
Jamesbarford: Right :+1:, I've shortened this point and moved the `bf16->f32->bf16` to an appendix while also showing that there are alternative valid lowering schemes. And where that is not valid, as with the case of FMA, emit a library call; 5321ed6256c1596766456a4ca4f4bec6563b6609

text/3983-bf16.md
```diff
@@ -124,26 +124,9 @@ These scalar operations must be implemented in conformance to the widen -> `f32`
 
 ### Scalar Arithmetic
 
-Scalar `bf16` arithmetic without native scalar instructions are defined by exactly converting both operands to `f32`, performing the corresponding Rust `f32` operation, and narrowing the result to `bf16` using [IEEE-754-2019][ieee-754] `roundTiesToEven`, without flushing subnormals to zero. Native `bf16` operations that round directly to `bf16` are legal only if they produce the same result as this sequence, ensuring consistent and portable behaviour across targets.
+`bf16` scalar operations behave according to the [IEEE-754-2019][ieee-754] specification. The scalar operation will return the value representable in the `bf16` format that is closest to the infinite-precision result with `roundTiesToEven` and preserving subnormals without flushing to zero. This will be true on targets woth with native scalar arithmetic instructions and those without.
```
RalfJung: typo: "woth with"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..5321ed6256c1596766456a4ca4f4bec6563b6609)*

text/3983-bf16.md
```diff
@@ -0,0 +1,512 @@
...
+
+Add a primitive type, `bf16`, to Rust for the 16-bit **Brain Floating Point** format.
+
+`bf16` is a 16-bit floating-point format derived from `f32`. It keeps the same 1-bit sign and 8-bit exponent layout as `f32`, but stores only the top 7 fraction bits. This gives it approximately the same dynamic range as `f32`, but much lower precision. The primary role of `bf16` is representing bfloat16 values in memory, in APIs, in FFI signatures where the target ABI supports bfloat16, and as an element type for SIMD and matrix operations. It is expected to be uncommon for a `bf16` to be used as a general-purpose arithmetic type in the same way as `f32` or `f64`.
+
+`bf16` is not an IEEE-754 binary floating-point format however this RFC specifies the proposed `bf16` type as following portable IEEE-754 semantics for conversions and scalar arithmetic. These are specified as widening to `f32`, performing the corresponding `f32` operation, and narrowing back to `bf16` using round-to-nearest, ties-to-even (RNE) without flushing subnormals to zero. As such, implementations may use native `bf16` instructions for scalar arithmetic and scalar conversion only where they produce identical results.
```
RalfJung: "These are specified as widening to ..." is outdated now, that's not how they are specified any more.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..5321ed6256c1596766456a4ca4f4bec6563b6609)*

text/3983-bf16.md
Jamesbarford: 376dbcf8 :+1: thanks

text/3983-bf16.md
```diff
@@ -348,26 +367,55 @@ Specifying portable vector semantics therefore requires settling several questio
...
+let c: bf16 = 3.3bf16;
+
+let d = a + b + c;
+```
+ 
+Which could semantically equate to any of the following;
+
+```rust
+// Widen to `f32` first
+let d: bf16 = (f32::from(a) + f32::from(b) + f32::from(c)) as bf16;
```
programmerjake: when doing `bf16` addition, it isn't correct to convert 3 values to `f32`, do 3 `f32` adds, and then convert the result to `bf16` -- you have to convert back to `bf16` after *every* addition.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..e491009c23014da8a1770c13c6a0d6d02044d5ef)*

Jamesbarford: I think I perhaps got the wrong end of the stick with; https://github.com/rust-lang/rfcs/pull/3983#discussion_r3593178086

I've removed this line.

tgross35: If this comes from https://github.com/rust-lang/rfcs/pull/3983#discussion_r3593178086, I was talking specifically about `algebraic_add` there

text/3983-bf16.md
Jamesbarford: 04a5572b :+1:

text/3983-bf16.md
```diff
@@ -147,32 +147,17 @@ These scalar operations must be implemented in conformance to the widen -> `f32`
 Please see [footnotes on `bf16` scalar arithmetic](#bf16-scalar) for a more indepth exploration into hardware considerations and support.
 
 ### Vector operations
-[reference-vector-operations]: #vector-operations
+[reference-vector-operations]: #reference-vector-operations
 
-A first-class `bf16` type is primarily useful for target-specific SIMD and Matrix APIs. These APIs commonly operate on vectors or tiles of `bf16` values and accumulate into `f32`, reflecting the way `bf16` is typically used in machine learning and numerical code.
+The types and operations for vectors will be provided by `core::arch`, for the moment. As is usual for these operations their semantics are architecture specific. Where similar instructions between plaftorms should not be expected to produce the same results.
```
RalfJung:
```suggestion
The types and operations for vectors will be provided by `core::arch`, for the moment. As is usual for these operations, their semantics are architecture-specific: similar instructions between platforms should not be expected to produce the same results.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..04a5572bb9bab8beeedc07357f19ed4e8d07a52b)*

Jamesbarford: f756210d I need to install a spell checker in Vim :facepalm:

text/3983-bf16.md
Jamesbarford: I've significantly re-worked this section. :+1:

text/3983-bf16.md
Jamesbarford: I've move where this lives and added `TryFrom` to the list

text/3983-bf16.md
Jamesbarford: I've removed it and shortened the section :+1:

text/3983-bf16.md
```diff
@@ -247,9 +244,7 @@ This decision is reversible across an edition. Prelude membership is edition-sco
...
-
-The `usize` precedent is weaker than first appears as `usize` is deterministic per target: given a target triple its behaviour is fully specified and stable. Its variation is structural 32-bits vs 64-bits and the type system prevents it being mixed with fixed sized integers. However `bf16` would vary on the same target triple depending on what target features have been made available. This problem is described in the [Prior Art](#c-compiler-support).
+Another option would be target-dependent behaviour for scalar arithmetic thus producing plausibly faster code at the cost of portability. For instance the aforementioned `VADDBF16` instruction flushes subnormals to zero, which does not follow IEEE-754 and is not portable. Ostensibly being more performant that converting to `f32` doing the operation and narrowing back to `bf16`. Some portability issues are presently observable in clang when converting from an `f32` to `bf16` which are discussed in [Prior Art c++ compiler support](#c-compiler-support).
```
RalfJung: I assume we will offer that instruction in core::arch, so it remains available for people if they need that kind of performance in a hot loop?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..07390f16d2c54260bbb00f4300ca17d4cd119b4e)*

Jamesbarford: I have this;
> Implementations may use native `bf16` instructions only when they produce the same observable result. Instructions with different rounding, flushing, semantics remain available through target-specific intrinsics. NaN payload behaviour is not strengthened beyond Rust's existing floating-point guarantees from [RFC-3514][RFC-3514].

In the Guide-level -> Scalar Arithmetic section, which I feel covers that?

RalfJung: Seems worth mentioning here when discussing target-dependent scalar behavior.

text/3983-bf16.md
Jamesbarford: This section has been shortened and added as a future possibility that is no longer part of this RFC

text/3983-bf16.md
```diff
@@ -0,0 +1,462 @@
...
+Initially the proposed `bf16` type will _not_ be a primitive type living in the global namespace. There have been difficult challenges with the `f16` type which almost certainly would crop up when implementing a primitive `bf16` type. Thus for the time being the `bf16` would live in `core::num::bf16` and could, at a later date, be included in the prelude and its functionality extended.
+
+Having only the path `core::num::bf16` signals that this is a specialist interchange type rather than a general-purpose arithmetic type, although a `bf16` arguably is one to some extent, and helps prevent misuse of the type.
+
+Neither of these opinions is to say that the `bf16` could not be implemented as a primitive and be added to the prelude at a later point in time, nor do these proposed decisions box a `bf16` into a particular path.
+
+### Omitting Scalar Arithmetic 
+[omitting-scalar-arithmetic]: #omitting-scalar-arithmetic
+
+Omitting scalar arithmetic keeps the initial implementation of `bf16` simpler and avoids potentially complex or inefficient implementations as well as the minefield of conversion issues discussed in the [conversions][conversions] section.
```
kennytm: Would the type still implement `PartialEq`? `Default`? `PartialOrd`?

The entire trait list is deleted from 88d5308e34cc114b403ffbe5f0d85c8fc0be55a0, but unlike `Add`/`Sub` etc it is not spelled out whether these non-arithmetic traits aren't supported.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..88d5308e34cc114b403ffbe5f0d85c8fc0be55a0)*

text/3983-bf16.md
```diff
@@ -0,0 +1,462 @@
...
+    let retained_lsb = (bits >> 16) & 1;
+    let rounding_bias = 0x7FFF + retained_lsb;
+    Self((bits.wrapping_add(rounding_bias) >> 16) as u16)
+}
+```
+
+### Display
+[display]: #display
+
+`Display` for `bf16` should follow the same principles as Rust's existing primitive floats: producing a decimal representation that parses back to the same `bf16` value via `FromStr`, preferably using the shortest such representation.
```
kennytm: similar question for `FromStr`, `UpperExp` and `LowerExp`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..88d5308e34cc114b403ffbe5f0d85c8fc0be55a0)*

text/3983-bf16.md
Jamesbarford: c843595d8ee0a4a576586d60384e5f643894ebbe, I think `Default` yes: it would be the bit pattern `0x0`. However for `PartialEq` and `PartialOrd`, in the RFC I've used language to err on the side of caution as they may encounter the same stumbling blocks as some of the scalar arithmetic operations.

**EDIT**
As per the feedback in another comment I've added `PartialEq` and `PartialOrd` to the list of traits

text/3983-bf16.md
Jamesbarford: c843595d8ee0a4a576586d60384e5f643894ebbe added them :+1:. (Marking this as resolved as the other comment overlaps)

text/3983-bf16.md
```diff
@@ -0,0 +1,478 @@
...
+- `Display` [discussed below][display]
+- `Default`, defaulting to `0x0`.
+- `FromStr`
+- `Copy`
+- `Clone`
+- `Debug`
+- `UpperExp`
+- `LowerExp`
+
+Depending on the complexity, the traits `PartialEq` and `PartialOrd` could be considered. These traits, however, may encounter the same [conversion pitfalls][conversions] as discussed in the rationale for _not_ adding scalar arithmetic straight off the bat. The final decision is deferred to the implementation stage so that any platform-specific or native-instruction nuances can be considered.
```
programmerjake: the comparison impls are very easy since losslessly converting from `bf16` to `f32` is just `f32::from_bits((bf16::to_bits(v) as u32) << 16)`, the hard part is when you need to round to `bf16`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/f635361c8e3d927208afd6097d4fc5f30e703b43..c843595d8ee0a4a576586d60384e5f643894ebbe)*

Jamesbarford: 0601c45e0caafda6e8e900296a20229efe71cbaf

tmandry:
We [discussed](https://hackmd.io/-hIVbXK_RI6WO-ujJj2ZZg?view) this RFC in today's lang team design meeting. The overall feeling was positive on the RFC, and we were interested in moving forward quickly with something that has minimal design questions to answer. There were a few main points of feedback on the RFC.

### Naming

We settled on calling this `f16b`, with the idea that floating point types in Rust should start with `f`. This is close enough to the common name that it should be easy to remember. @traviscross proposed that we refer to it as the _balanced_ variant of `f16`, which I like, because it has a similar number of exponent and mantissa bits (8 and 7, respectively, compared to `f16`'s 5 and 10).

We are supportive of including `bf16` aliases in `core::arch` for architectures that document that name explicitly. Our convention in `core::arch` is to use architecture-specific names, so the name makes sense there.

### Trait implementations

There were reservations expressed in the meeting about including any trait implementations where we have to make non-obvious design choices today. For example, `Display` impls might convert to `f32` before printing, but this is arguably not correct behavior. The eventual behavior details would be a libs question, and I think adding the impls could probably be handled as part of their ACP process. We are happy to move forward without them for now.

For `From`, we are in favor of including `impl From<f16b> for f32`, but not the lossy other direction. We're okay with the only constructor being `from_bits` for now, which allows us to defer the decision about truncating vs rounding for conversions from other floating point types.

### `core::arch::x86_64::bf16`

We should unify these types right away and commit to not shipping a separate type under `core::arch::x86_64`. It makes sense to unify them ahead of stabilization, in case that surfaces any issues.

## Feedback on future work

### Primitive type

We didn't have consensus on whether this should be a true primitive type (with `as` casts and literals) in the future, but this can be decided later. We were okay to defer this question for now and would stabilize it as described in the RFC.

### ABI instability

@folkertdev brought up the issue of the ABI of this type in `extern "C"` functions being unspecified on many platforms, and thus subject to change in the future. There was general consensus that we should lint against usages of these types in places where the ABI is not determined.

One reason to do this as warn-by-default and not deny-by-default is that the ABI stability varies by platform. It would be preferable to warn universally in cases of non-portable code than to not compile on specific platforms where the type is supported but the ABI may change. That would also let us use the existing `improper_ctypes` lint. We didn't get explicit consensus on this point, and I can see a case being made either way, but for my part I'm pretty sure we want it to be part of `improper_ctypes`.

tmandry:
I'm kicking off a joint FCP between lang and libs-api to start collecting checkboxes. Filing concerns based on the feedback above.

@rfcbot fcp merge lang,libs-api

@rfcbot concern naming
@rfcbot concern defer-formatting-trait-behavior
@rfcbot concern commit-to-no-other-bf16

traviscross:
Thanks @Jamesbarford for the RFC, and thanks @tmandry for writing that out.

@rfcbot reviewed

VitWW:
Section "f16b Scalar Operations" sitll consists `bf16`.

tmandry:
@rfcbot resolve naming

Note that the other two concerns are still in effect.

text/3983-bf16.md
```diff
@@ -0,0 +1,467 @@
...
+
+
+### Naming `f16b`
+[naming-f16b]: #naming-f16b
+
+Other such proposed names have been; `bf16`, `f16b`, `fb16`, `bfloat16`, and `__bf16`.
+
+This RFC prefers `f16b`.
+
+The name `f16b` is in keeping with the Rust naming convention for floating point types which all start with an `f` prefix.
```
kennytm, line 150: Shedbike.

I really dislike this new name `f16b` (and `fb16` is doubly worse), because all other places using the type always put the "b" before the "f" as shown in all the Prior Art below (C++23 `std::bfloat16_t`, clang `__bf16`, C# [`System.Numerics.BFloat16`](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.bfloat16?view=net-11.0), all the SIMD/GPU instruction names, etc). The names for existing floating-point types like `f64` is intelligible for other languages (e.g. C and Go) that may choose a longer name like `float64_t`, but `f16b` would break that.

The only place `f16b` appears is our previous #2690 which was retracted.

`f16b` is also bad for searchability. If you search the web for `bf16` you get information about the type. If you search for `f16b` you get introduced a fighter jet.

The naming convention makes sense for `f16`/`f32`/`f64`/`f128` which strictly follows IEEE-754. But bfloat16 does not come from IEEE-754 exactly, I see no obligation it must start with the letter "f". 

IMO even `std::num::BFloat16` would be better than `f16b`. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/c6f630fbbd58f9bb104dc750f6eda746817475c4..397111bfc545e56c2077aeab7e53b92161aec832)*

text/3983-bf16.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Prior Art
+[prior-art]: #prior-art
+
+### C++ compiler support
+[cpp-compiler-support]: #cpp-compiler-support
+
+The `C++23` standard defines a `std::bfloat16_t` type in `<stdfloat>` as part of the extended floating-point types introduced by [P1467][P1467]. This was motivated by the prevalence of availability of `bf16` on hardware, library support through class wrappers being cumbersome to use and a desire for more portable code. [(for the section on adding extended floating point types see here)][cpp_bf16_motivation]
+
+GCC 13 implements the C++23 [P1467][P1467] extended floating-point types, including `std::bfloat16_t` and the `f16b` literal suffix. Clang's support for extended floating-point types, at the time of writing, is more limited and varies by feature; see the [Clang C++ status page][clang_status_page] (then search the page for `P1467`). Both GCC and Clang nevertheless support `bfloat16` through the `__bf16` type.
```
kennytm: The C++ suffix is `bf16`! Seems like bulk-replace mistake.

```suggestion
GCC 13 implements the C++23 [P1467][P1467] extended floating-point types, including `std::bfloat16_t` and the `bf16` literal suffix. Clang's support for extended floating-point types, at the time of writing, is more limited and varies by feature; see the [Clang C++ status page][clang_status_page] (then search the page for `P1467`). Both GCC and Clang nevertheless support `bfloat16` through the `__bf16` type.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3983/c6f630fbbd58f9bb104dc750f6eda746817475c4..397111bfc545e56c2077aeab7e53b92161aec832)*

text/3983-bf16.md
liigo: > If you search for f16b you get introduced a fighter jet.

then just search `f16b rust`, most people already do this. other langs already name it differently, `f16b` fits rust-style well.

kennytm: > other langs already name it differently, `f16b` fits rust-style well.

So name those other langs that divert from "bf16" or "bfloat16"? 

The situation here is unlike `f64f64` which there isn't a consistent naming (ppc_fp128? __ibm128?). You're breaking the external consistency of a well-established term used everywhere else, for this so-called internal consistency which itself may not even be a real thing:

* The "rust-style" only formed [ad-hocly](https://github.com/rust-lang/rfcs/pull/3451#discussion_r1246047671) from the names `f32`/`f64` and then joined by `f16`/`f128`. There is never a document declaring that "all floating-point types must start with the letter `f`", and [not even a universal agreement](https://github.com/rust-lang/rfcs/pull/3451#discussion_r1246065657).
* Also, to reiterate, bfloat16 is not ~~IEEE-754-abiding~~ **a basic IEEE-754 format**, so there's no obligation to be consistent with `f16`/`f32`/`f64`/`f128`. It's like if Rust somehow supported one's-complement signed integers there's no obligation to name those type starting with `i`.
* In the current form of the RFC almost none of the usual `fNN` operation is available, they are all punted into "future possibilities", so the user-experience of this new type will be inconsistent in basically all cases except the name.

Additionally, as the type proposed right now is merely a struct with lang-item *not* a primitive, designed as "*specialist interchange type rather than a general-purpose arithmetic type*", so it shouldn't require the special lower-case type naming in the first place.

programmerjake: > * Also, to reiterate, bfloat16 is not IEEE-754-abiding

the exponent and significand bit widths are not one of the options listed in IEEE 754, but it otherwise follows all the rules of IEEE 754 to the same extent that `f32`/`f64` do. technically, it is a specific instance (with b=2, p=8, emax=127) of an *__extendable precision format__* as defined in IEEE 754-2019.

kennytm: @programmerjake Thanks for clarification. In IEEE 754-2019 §3.7 "Extended and extendable precisions", 

> * An **extendable precision format** is an arithmetic format with a precision and range that are defined under user control.

the lack of arithmetic operations is still a blocker, though that's a not a fault of the format itself.

> These formats [...] shall:
> * provide all the representations of floating-point data defined in terms of those parameters in 3.2 and 3.3 :heavy_check_mark: 
> * provide all the operations of this standard, as defined in Clause 5, for that format :thinking: 

And I think what §3.7 wants is to support a `float</*p*/8, /*w*/7>` type (b=2, emax=2<sup>w-1</sup>-1) just like #3686. 

> Language standards should define mechanisms supporting extendable precision for each supported radix. Language standards supporting extendable precision shall permit users to specify *p* and *emax*. Language standards shall also allow the specification of an extendable precision by specifying *p* alone; in this case
*emax* shall be defined by the language standard to be at least 1000 × *p* when *p* is ≥ 237 bits in a binary format or *p* is ≥ 51 digits in a decimal format.

This was the approach explored by Zig (ziglang/zig#3148) naming bf16 → `f16_7`, [tf32](https://en.wikipedia.org/wiki/TensorFloat-32) → `f19_10`, ATI fp24 → `f24_16`, though I think the language has abandoned the idea as I can't find the equivalent on their new repository. And while this approach *is* consistent it is also incredibly non-intuitive (i'm not going to memorize that bf16 has 7-bit precision :upside_down_face:) that is only useful if a truly generic `float<P, W>` exists. I get that this naming variant (`f16m7`/`f16e8`) has already been [discussed](https://hackmd.io/@rust-lang-team/Hk8aSkWLMx#Bikeshed).

programmerjake: yeah, I agree that it would fit into the standard as `float</*p*/8, /*w*/8>` (7 isn't correct for `w`, since $127=2^{8-1}-1=2^{w-1}-1$), however I also think bfloat and bf16 are the standard names for that type, and that basically no one would recognize `float<8, 8>` as being what they want without getting deep into the weeds of float formats.

tmandry:
@rfcbot resolve defer-formatting-trait-behavior

folkertdev:
How do we make progress on the one remaining open concern?

* commit-to-no-other-bf16 (https://github.com/rust-lang/rfcs/pull/3983#issuecomment-5196867183)

Should the RFC just be updated to mention that the `core::arch::x86_64` should be removed and `core::num::f16b` used instead? Or is there more to it than that?

The RFC text already contains:

> `f16b` would be defined in `core::num` and re-exported in `std::num`, like use `core::num::f16b` or use `std::num::f16b`. This can then be re-exported as a `bf16` alias in `core::arch` which presently homes `core::arch::x86::bf16`.
> `core::arch::x86::bf16`'s should use the `f16b`.

Jamesbarford:
> How do we make progress on the one remaining open concern?
> 
>     * commit-to-no-other-bf16 ([`f16b` type #3983 (comment)](https://github.com/rust-lang/rfcs/pull/3983#issuecomment-5196867183))
> 
> 
> Should the RFC just be updated to mention that the `core::arch::x86_64` should be removed and `core::num::f16b` used instead? Or is there more to it than that?
> 
> The RFC text already contains:
> 
> > `f16b` would be defined in `core::num` and re-exported in `std::num`, like use `core::num::f16b` or use `std::num::f16b`. This can then be re-exported as a `bf16` alias in `core::arch` which presently homes `core::arch::x86::bf16`.
> > `core::arch::x86::bf16`'s should use the `f16b`.

I have reworded the points to be a tad stronger;

> - `f16b` would be defined in `core::num` and re-exported in `std::num`, for example `use core::num::f16b` or `use std::num::f16b`.
> - `f16b` will be Rust's only language-provided brain floating-point type. The existing `core::arch::x86::bf16` type will instead use `core::num::f16b` and be re-exported from `core::arch` as the `bf16` alias.

Is this wording satisfactory to address "commit-to-no-other-bf16" ([`f16b` type #3983 (comment)](https://github.com/rust-lang/rfcs/pull/3983#issuecomment-5196867183)) ?

tmandry:
@rfcbot resolve commit-to-no-other-bf16

nikomatsakis:
@rfcbot reviewed

PR #3984, opened 2026-07-15
RFC: Refactor the libs team

nia-e:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3984)*

Restructure and simplify the libs team, per conversations with @Amanieu as the lead & broader team membership.

cc @rust-lang/libs @rust-lang/libs-api @rust-lang/libs-contributors 

[Rendered](https://github.com/rust-lang/rfcs/blob/master/text/3984-libs-team-refactor.md)


text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+[coc]: https://www.rust-lang.org/policies/code-of-conduct
+
+### Libs FCP team
+[libs-fcp-team]: #libs-fcp-team
+
+A subset of the libs team is to be granted an *FCP checkbox* in order to decide on matters relevant to the standard library more broadly. This team is to consist of at least 5 and at most 8 members, with the exact number free to oscillate between these extremes.
+
+#### Selection process
+
+Every 12 months, the membership of the libs-fcp team is reshuffled. Any member of the libs team may self-nominate for the position, including those already on the libs-fcp team. The old libs-fcp team (by means delegating this responsibility to a [facilitator], if they so choose) is then to pick between 5 and 8 (inclusive) of the nominees, which will constitute the new team. The new membership list must be approved by all former members of the team; that is, there must be no active objection to the new list.
```
programmerjake:
```suggestion
Every 12 months, the membership of the libs-fcp team is reshuffled. Any member of the libs team may self-nominate for the position, including those already on the libs-fcp team. The old libs-fcp team (by means including delegating this responsibility to a [facilitator], if they so choose) is then to pick between 5 and 8 (inclusive) of the nominees, which will constitute the new team. The new membership list must be approved by all former members of the team; that is, there must be no active objection to the new list.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..719a5dcddf7aa2a5953d7db8dab6af279baf1a76)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### Expected activities
+
+Libs team members should make an effort to remain engaged with the Rust project in some capacity, but such criteria are to be interpeted loosely. Engagement in pull request or issue discussions, being on *review rotation*, activity in team meetings, and/or activity on Zulip or other official platforms of the project in matters relevant to the standard library and related crates all constitute relevant engagement with the project.
+
+More broadly, members of the libs team constitute a core part of the Rust project and thus are bound by the expectations placed upon such members, including upholding both [the spirit and the letter of the Code of Conduct][coc].
+
+After some member of libs has been inactive for at least twelve months, they may be asked if they wish to remain on the team; otheriwse, they may be moved into alumni, with privileges revoked. An alum may at any point self-nominate to be reinstated, requiring a second and 10 day period without any objections as if they had been nominated by a team member in order to be reinstated.
+
+#### ACP process
```
clarfonthey, line 88: This is I guess more of a plan for the std dev guide than this RFC (I.e., not blocking the RFC), but I do think we should encourage all major libs changes to go through the ACP process even if they don't necessarily change the API surface. Stuff like changing behaviour that we're allowed to change (ex. platform-specific stuff) or large refactors should be encouraged to go through the ACP process instead of going straight to a PR as they sometimes do today.

We should also feel encouraged to open ACPs just to track other discussions happening on places like Zulip just for visibility etc.

(I think that the definition of the ACP process here is broad enough to allow changing all that, just, I think it's important enough to mention here in case it does affect the RFC wording slightly.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..719a5dcddf7aa2a5953d7db8dab6af279baf1a76)*

nia-e: a philosophical point @joshtriplett raised in private conversation that somewhat informed the way i wrote this rfc was to try to only change processes that are directly impacted by the team restructure. the acp process has admittedly partly been an exception since the changes are moderately broader. i think something like what you're suggesting *would* be interesting direction to go in but shouldn't explicitly be enshrined in rfc (as then changing it would take an rfc). does any of the wording in that section prevent expanding the scope as you suggest? if so i'm happy to loosen said wording ofc

clarfonthey: So, *I* would interpret the RFC as the ACP being a broad process the libs team has the right to change, though it presumably would require an FCP like anything else on the std dev guide. I just am not 100% confident in that interpretation which is why I kind of wanted to express that interpretation and just kind of confirm it fits in line with what is written.

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+This fuzzy overlap between top level teams results in occasional friction during their respective team meetings, with time often spent attempting to determine whether a given agenda point is indeed relevant for said meeting or should be discussed in the other top level team's meeting. Attendance of these meetings also does not necessarily follow either the distinction between the teams or the separation between libs/libs-api and libs-contributors, with the FCP rights of the top-level teams not matching either those involved in conversations or those doing work on various aspects of libraries.
+
+Furthermore, the libs-contributors team now numbers 20 members, alongside auxiliary teams that strongly overlap in responsibilities such as crate-maintainers and libc. The permissions afforded to the members of these teams are approximately equal between each other, and the distinction moreso represents the historical circumstances whereby a given member entered the broader libs team family than their concrete current areas of involvement.
```
tgross35: libc is just under crate-maintainers, for what it's worth

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+Further, several terms which appear throughout this RFC should be understood to have a particular meaning:
+- **second**
+    - In the situations listed below where team members have the option to *second* a proposal or nomination, followed by some period of time passing, the intended meaning is that there must exist some particular seconding which is in effect for the entire duration in question. The longest-standing seconding being withdrawn from some given proposal results in the relevant countdown being turned back to the duration of the next-longest-standing seconding.
+- **deadlock**
+    - A conversation is understood to be in *deadlock* when, following extensive conversation, neither party to the conversation is willing or able to reach a mutual compromise with the other(s). The mechanisms about deadlock resolution specifically apply when these parties are members of the libs team.
+- **self-nominate**
+    - Self-nominations to various positions is to be effected privately via a message to a team lead. This does not necessarily need to be in writing. It is the duty of the team lead(s) to present the list of nominees to the appropriate body afterwards.
```
tgross35: What's the difference between "privately via a message" and "in writing"?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+[libs-team]: #libs-team
+
+The top-level team consists of all members of any team previously under the libs team family. That is, all former members of libs, libs-api, and libs-contributors are to be consolidated into a new libs team. Members of the crate-maintainers team are automatically extended an invitation to join the new libs team should they so desire. Members are granted *umbrella privileges* as above, with the possibility of adding themselves to *review rotation*.
+
+It is expected that the libs-fcp team schedules at least one weekly libs team meeting, wherein at least items nominated for the libs team are discussed. The libs-fcp team is free to decide internally when its meetings are held, their duration, and which topics are or are not relevant to any given meeting, so long as they are held on a consistent schedule which reasonably accomodates for the availabilities of all libs-fcp members and enables participation from other libs members. All meetings are to be open to the public unless the libs-fcp team decides by consensus to restrict attendance for some particular meeting. Minutes and agendas from public meetings must also be posted publicly on Zulip.
```
tgross35: Should this be s/libs-fcp/libs? This is the first time libs-fcp is mentioned in the doc.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### Selection process
+
+Anyone whose addition may be valuable to the development of the Rust standard library or related crates under the purview of the libs team may be nominated for membership in the team by an extant member. This nomination does not necessarily have to be expansive, but should be sufficient to assure the team that the nominee can be trusted with *umbrella privileges*. This may be as brief as a single sentence if there is no serious ground for concerns. Note that contributions need not necessarily consist of code, but also include documentation, productive engagement in complex conversations, and anything that could be seen as facilitating the work of other members.
```
tgross35, line 70: May want to mention where this takes place. I assume a "Nominating Ferris Jr." thread on Zulip similar to t-compiler?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+Any preexisting member may issue an objection to a nominee. Objections are considered blocking in the nomination process, and are to be accompanied by motivating reasoning. The criteria which may justify an objection are left intentionally unspecified, so as to allow for flexibility in the decision-making process. By courtesy, objections are to be sent privately to the team leads instead of being announced publicly; however the motivation should, if possible, be made public.
```
tgross35: What exactly is an objection vs. its motivation that should be shared separately?

For t-compiler you can just mention any concerns on the (team-private) nomination thread. It seems helpful if they are at least visible to team members so others can agree or disagree.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+Members may also second a nominee if they believe them to be a valuable addition to the team, without necessarily requiring any motivation.
+
+If a period of 10 days has passed within which a nominee has been seconded and has no outstanding objections, the nominee is automatically added to the team and granted the relevant privileges. If the nominee was not previously part of the Rust project, moderation will first be asked to ensure there are no potential issues with their joining, with the addition to the team only taking effect after moderation signoff.
+
+If on the contrary, 20 days after a nomination, no seconds are outstanding and/or there are still outstanding objections, a countdown of 10 days begins. If objections persist, or, lacking objections, no seconds are put forward, the nomination is automatically withdrawn. Note that this similarly requires that there exists some singular objection which is outstanding for the entirety of these 10 days.
```
tgross35: 10 days seems a bit fast, some people would probably wind up missing out on discussion. I'm not sure how to qualify this but it would be nice if it could wait until e.g. about half of active team members had left a 👍 or some other indication that they'd at least seen it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+Any libs team member may approve an [API change proposal][acp] that has been submitted (excepting the author of the ACP). It is highly recommended, though not explicitly required, that a waiting period of ten days from submission be observed between an ACP being submitted and it being approved.
+
+Approval of an ACP does not necessarily guarantee that it will be included stably in the standard library, but signals that the relevant proposal may be of interest to the team, and seems *reasonably likely* in the judgment of a libs member to pass a subsequent libs-fcp team FCP without objecions. It is left to the judgement of individual team members whether some given ACP has received the necessary amount of scrutiny before being approved, and approval should signal confidence from the approver in the ACP's chances of future stabilisation. Note that ACPs proposing a [breaking change][breaking] require a subsequent FCP.
+
+An ACP should *not* be approved if its author does not appear to have thoroughly explored alternative solutions and there appear to be, or will be in the near future, outstanding conflicting proposals that have not also been given due consideration.
```
tgross35: There are some existing concerns about accepting ACPs without team discussion - this expands upon that by opening the number of people that may second. Would it be reasonable to maybe require at least 2-3 seconds or a discussion?

> ... but signals that the relevant proposal may be of interest to the team, and seems *reasonably likely* in the judgment of a libs member to pass a subsequent libs-fcp team FCP without objecions

For what it's worth, I don't think I've ever felt this heh. With the exception of completely trivial things like constifications (which tend not to get ACPs) there's _always_ some design surface area that the synchronous discussion seems to bring up.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..fe7aaea71f141682c29ec32bc74b2459793de399)*

tgross35: If the (informal) meeting requirement is being removed, I wonder if there would be value in creating Zulip threads to discuss ACPs similar to compiler MCPs, rather than discussing on the GH thread. That tends to be a bit more conversational than discussion on an issue.

text/3984-libs-team-refactor.md
nia-e: Intent was to signal "this can be an in person conversation or happen on a call" but I'll edit the wording, ty

text/3984-libs-team-refactor.md
nia-e: no, the responsibility is on libs-fcp to schedule said meeting, but I'll clear up wording

text/3984-libs-team-refactor.md
nia-e: the intent was effectively, "X particular person objects to Y's nomination" stays private but the anonymised rationale in the form of "there is an outstanding objection to Y on the grounds of $foo" should be made public. again happy to edit wording to make that clearer but this is moreso a guideline

text/3984-libs-team-refactor.md
nia-e: a heavier mechanism was originally suggested here but the result of the discussion in t-libs/meta on zulip was that even this suggested ACP process constitutes a compromise, with some voices being in favour of relaxing it more. personally i think it's an acceptable failure mode if "bad ACPs" make it through so long as they aren't stabilised, and slowing down discussion or requiring multiple seconds can be demotivating to new contributors

text/3984-libs-team-refactor.md
tgross35: That might be worth a mention in the libs-fcp section, just to point out if they're kind of the ones to steer the team and handle things like scheduling

text/3984-libs-team-refactor.md
tgross35: For the compiler threads, objections tend to lead to follow up like "that's been my experience too", "how long ago was this?", "can you link that PR?", "I have some concerns that I don't want to list here, I'll bring it up with the leads", etc. If it winds up similar for libs then I expect that's going to be difficult to anonymize, especially given the smallish team size, and IMO probably not worth it.

But if a mechanism has already been figured out then it doesn't hurt to try.

text/3984-libs-team-refactor.md
nia-e: hopefully clarified language here to signal better the intent of "ACP accepted" = "we want an API that is vaguely this shape, modulo bikeshed and nitpicks"

text/3984-libs-team-refactor.md
nia-e: added a bit about checking there's been enough discussion (cc @Amanieu since this counts as making your job harder)

text/3984-libs-team-refactor.md
nia-e: id leave the details there to be decided ad hoc, but we should clarify that yes. just unsure that should go into this rfc ig

text/3984-libs-team-refactor.md
nia-e: reworded to match intent better

text/3984-libs-team-refactor.md
tgross35: Is it really more contributor-friendly to accept an ACP and ask for followup changes later, or to potentially delete their work _after_ they've gone through the effort? :)

Currently the bar for a ACP acceptance is loose consensus from libs-api, which typically seems to mean 2/5 or 3/5 people agree. The libs team is about to grow to ~35 people: needing two of them to second would still be a significantly lower bar than we currently have.

A bias for action is good but I think it needs to be balanced with the fact that std doesn't _need_ every ACP, even though you could probably find at least one person who likes each idea.

clarfonthey: The bar actually isn't loose consensus from libs-api; we still require only one person to accept it, although inject more scrutiny if something isn't obviously useful. The general idea is that ACP acceptance means we *won't* just straight-up delete someone's work; it may dramatically change shape and ship-of-theseus itself into something else, but it's unlikely to be deleted.

Also, more scrutiny generally gets applied for larger changes. Here's an example of a small change where one person's acceptance was totally possible and we actually questioned it a bit because we weren't sure: rust-lang/libs-team#252

The proposal? Add `char::MIN`, which is just `\0`, but similarly `usize::MIN` is just `0` and we have that. Seems reasonable, one person needed to accept it, no controversy.

nia-e: this is approximately my thought process. the listed relaxed acp process certainly puts more trust in team members, but mostly only brings it in line with the acp process that had been used in the past (where the acp section of meetings was moreso sitting in a call and separately sorting thru acps, giving approvals or rejections based on individual opinion and nominating for group discussion as an exception). and if something is controversial, that can always be brought up for sync discussion. 

i will also cc @burntsushi here since the language around objections was minimally tightened since the last public draft, per more feedback i received

text/3984-libs-team-refactor.md
joshtriplett: To the extent we changed the ACP process here, much of it was an attempt to make sure that process made sense with the new team structure. We can always change it more as needed.

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+- **review rotation**
+  - Members on the *review rotation* will be randomly assigned to new pull requests submitted to the standard library. Being on the review rotation is one of the best ways for members to help the libs team and learn new parts of the library. Review capacity is one of the most important resources that the team has, as it enables our progress in the standard library's continued development and maintenance. There is no concrete expectation that libs members be on the review rotation, but presence on rotation may be used as one of multiple indicators of a member's activity in the project and all members of the libs team are encouraged to add themselves to the rotation.
+- **FCP checkbox**
+  - Members with an *FCP checkbox* are those who have checkboxes with rfcbot to sign off on decisions that require the final comment period process. That is, in order for an FCP to be resolved, there must be no outstanding concerns, at most two unticked checkboxes may be outstanding, and this must stay the case for 10 days. Each member with an *FCP checkbox* contributes one to the total count of checkboxes that exist.
```
yaahc: Might be nice to more explicitly call out or alternatively frame this as "members whose consent is required"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..e44329837832af6481ea4b19a21306140fe8ad8e)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+An FCP, or Final Comment Period, is required for any change which would either be visible in a stable release or which is seen as a major addition to the standard library. The process is handled by [rfcbot], wherein all those with an *FCP checkbox* may signal agreement with a proposal or lodge an objection. If 10 days pass wherein at most two checkboxes are outstanding and no objections are lodged, the FCP will close automatically in the direction it was proposed (i.e. in favour or against).
+
+If an FCP appears unresolvably deadlocked due to blocking concerns from non-libs-fcp members, libs-fcp consensus may overrule those concerns.
+
+#### Ad-hoc subteams
+
+It may be the case that the libs-fcp team wishes to separate its activities along various lines, such as distinct areas of the standard library, API design versus implementation, etc.; in such cases, libs-fcp is empowered to delegate its FCP rights in certain cases to any subteam(s) of itself as it sees fit should there be internal consensus to do so. These may take the form of standing subteams with dedicated meetings and agendas, ad-hoc ping groups, topic-specific FCP lists, or anything in between.
```
yaahc, line 142: how do these subteams interact with the annual reshuffling established in `#### Selection Process`? Does that selection process include repopulating these teams? Do the teams all shut down and need to be re-established upon each iteration of libs-fcp?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..e44329837832af6481ea4b19a21306140fe8ad8e)*

yaahc:
This looks fantastic btw.

text/3984-libs-team-refactor.md
nia-e: i left this underspecified somewhat intentionally, but i (and i think most ppl) would read it as "said subteams are the purview of libs-fcp and libs-fcp can set its own internal guidelines for this". i don't want to set out hard rules here since it'd then be an rfc-level process to change said rules

text/3984-libs-team-refactor.md
nia-e: i like the suggestion, reworded accordingly ^^

text/3984-libs-team-refactor.md
clarfonthey: I'm just going to resolve this then, since I think it's something we can figure out post-RFC.

text/3984-libs-team-refactor.md
clarfonthey: I think the confusion was that currently, crate-maintainers is listed on the library page of the team list but the exact set of crates between libs purview and crate-maintainers purview is a bit fuzzy. (For example, hashbrown is under libs maintainership since it's used in the standard library, but as mentioned, libc is under crate-maintainers.)

The one crate that *is* separated out is regex: https://rust-lang.org/governance/teams/library/#team-regex

And so maybe that was what was intended? Although the regex team being invited is kind of moot since the only person there atm is @BurntSushi who is already on libs-api and thus would be automatically added to the libs team anyway.

text/3984-libs-team-refactor.md
yaahc: makes sense to me

text/3984-libs-team-refactor.md
clarfonthey: FWIW, I wouldn't really call libs a small team; there are currently 29 people on libs-contributors and that's not even combining with the other libs-related teams. Granted, there are 74 in compiler, so definitely more, but I think that the size has passed the point where everyone would be able to identify everyone else based upon small snippets for something like this. YMMV of course.

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+If insufficient nominees exist to reach the minimum size threshold, or if a member of libs-fcp resigns or is removed during the course of their 12-month term dropping the membership count below 5, the team must attempt to explicitly invite libs team members to join libs-fcp until the team meets the minimum membership count again. If no potential candidates have consensus approval, libs-fcp may continue at a reduced size until such candidates are found. Libs-fcp should further attempt to nominate promising candidates from outside of the libs team for membership in libs to facilitate its reassuming of the desired minimum size. The invitation mechanism into libs-fcp may also be invoked any time the libs-fcp team consists of less than 8 members, should the team agree to do so by consensus.
+
+The initial composition of this team is to be determined jointly by members of the previous libs team by the same process, as if the libs team had constituted a previous libs-fcp team composition.
```
BurntSushi: Do you mean members of the previous libs-api team here? e.g., I'm not on libs technically at present.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..6f7405f363793f45fc1c543c08747cf0e719410b)*

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+- **self-nominate**
+    - Self-nomination to various positions is to be effected privately to a team lead. This does not necessarily need to be in writing. It is the duty of the team lead(s) to present the list of nominees to the appropriate body afterwards.
+
+[triagebot]: https://forge.rust-lang.org/triagebot/index.html
+[desktop]: https://forge.rust-lang.org/infra/docs/dev-desktop.html
+
+## Team structure
+[team-structure]: #team-structure
+
+The library team family will be consolidated to consist of only two distinct teams, as below.
```
BurntSushi: Can you explicitly state which teams will be consolidated here? e.g., the regex team is a sub-team of libs. There's also `project-portable-simd`, `wg-allocators` and `crate-maintainers`.

I _assume_ what is meant here is `libs`, `libs-api` and `libs-contributors`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..6f7405f363793f45fc1c543c08747cf0e719410b)*

text/3984-libs-team-refactor.md
nia-e: this has been a pickle for a while and i'm not sure what the right answer is. libs alone is what i spelled out here on the grounds that they are the current official top-level team, so "constitutionally" it makes the most sense (the team lead is lead of libs, libs-api is a subteam, etc.). i also do realise that much (though not all) of the proposed workload of libs-fcp would overlap with libs-api, so a case could be made for "libs + libs-api decides new libs-fcp team". i don't think it would be correct to have it be libs-api alone mainly for the reason that libs-api is officially just a random subteam of libs rn (albeit a very important one).

i also could see the case for "team lead picks" but that's an unfortunate precedent to set, so idk. cc @joshtriplett and @Amanieu since i believe we talked about this a fair bit

BurntSushi (approved):
This looks great! Thank you for putting this together. :-)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..6f7405f363793f45fc1c543c08747cf0e719410b)*

text/3984-libs-team-refactor.md
BurntSushi: We could also just say that the current membership of libs-api becomes libs-fcp for bootstrapping purposes if this is a big pickle.

I think that we should try to follow the spirit of this RFC though, which is that libs-fcp gets to pick their own successors. But not all of libs-api (as currently worded) gets to pick their own successors. So yeah, I get the constitutional argument here, but there's also the spirit of the RFC that I think is important to follow too.

libs+libs-api seems OK to me too. I think that's not a bad way to split the pickle, since I think it satisfies the spirit of the RFC while also respecting the current libs structure.

cc @dtolnay as the only other member of libs-api that isn't part of libs. And cc @cuviper and @thomcc as members of libs that aren't part of libs-api.

text/3984-libs-team-refactor.md
BurntSushi: I think this LGTM.

I continue to like the idea that ACPs should generally be a lightweight pulse check on whether libs-fcp would _likely_ be willing to accept a particular change or not. The original problem they were trying to solve was to avoid contributors doing a bunch of work only to get a "no we don't want that" response.

There will certainly always be some risk that an ACP gets accepted but that the change ultimately doesn't pass FCP. I think that's okay and it should hopefully be pretty rare.

I don't know if it's intentional or not, but this RFC also doesn't explicitly require an ACP for every API change. I kind of think that's a good thing.

nia-e: semi-intentional. i've been of the opinion that there is such a thing as "too small to need an ACP" but that's not a universal opinion. either way, leaving that out means we can determine this internally in libs{-fcp} ^^

text/3984-libs-team-refactor.md
nia-e: done ^^

text/3984-libs-team-refactor.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### Last-resort mechanism for libs-fcp reconstitution
+
+Since libs-fcp selects its own successors, ossification of team membership is a possible concern. This section provides a mechanism for the broader libs team to express no-confidence in the membership of libs-fcp and leadership of libs, and forcibly reconstitute it. This section is written in the hopes that it will never be needed, and that the teams will make every possible effort to resolve conflicts without reaching this point.
+
+Any member of the libs team may request to start a no-confidence vote to this effect. Once started and seconded, a 14-day timer to cast votes begins. Other libs team members (excluding those currently serving as part of libs-fcp) may vote in favour, against, or abstain. Votes may be, but are not required to be, accompanied by motivating reasoning.
```
the8472, line 148: > (excluding those currently serving as part of libs-fcp)

degenerate case: libs team is small, all but 2 members are on libs-fcp, those two members could instant-disband the larger group.
Alternatively, the "at least 50%" rule below, would make it impossible in this case?

I don't think that's a likely scenario, but it sets off an "this doesn't seem robust" itch.



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3984/f635361c8e3d927208afd6097d4fc5f30e703b43..9fee789d5289af3bc7ff78c24f270e2404c47ac7)*

joshtriplett: I think that'd be extremely unlikely to happen, since it'd require either libs-fcp to grow massive or libs to shrink substantially. I also think either version isn't robust; if you included libs-fcp then the mechanism doesn't *function* as long as libs-fcp is sufficiently large compared to libs. So, ether way, I think if we hit such a degenerate case we'd end up needing outside intervention (e.g. moderation/LC).

The question is then which case we think is *more likely*, and I think "libs minus libs-fcp cannot form a supermajority without libs-fcp members" is the more likely case compared to "libs minus libs-fcp is just 2-3 members".

Amanieu:
As per the dicussion in today's meeting.

@rfcbot merge libs,libs-api

nia-e:
squashed commits ^^

clarfonthey:
(don't we generally not do that for RFCs?)

nia-e:
was under the impression we do as usual. i can force-push back to the previous commit /shrug

ChrisDenton:
For RFCs we do like to keep a record of the changes. From the [readme](https://github.com/rust-lang/rfcs#what-the-process-is):

> You can make edits, big and small, to the RFC to clarify or change the design, but make changes as new commits to the pull request, and leave a comment on the pull request explaining your changes. **Specifically, do not squash or rebase commits after they are visible on the pull request.**

(their bold)

nia-e:
fixed, oopsies

clarfonthey:
Huzzah! The libs team is being refactored, and implementing this *is* part of the lib agenda.

Not super sure why we haven't merged this one, but I might as well go ahead and do that since it's accepted and several PRs have been submitted to various repos to try and centralise the docs to use T-libs and not mention T-libs-api so people don't get confused. The final change to implement this on the actual team permissions is in rust-lang/team#2651, which has a few outstanding issues regarding crate ownership before it can be merged.

But, we don't normally wait for implementations before RFC, merge, so…

PR #3985, opened 2026-07-15
Create 0000-enum-variant-cast-patterns.md

ghost:
This RFC proposes allowing unit variants of unit-only enums with an explicit primitive integer #[repr(...)] to be cast directly to that representation type in patterns:

match byte {
    Operation::Banana as u8 => {}
    _ => {}
}

This is narrowly defined syntactic sugar for an equivalent named constant pattern. It introduces no implicit conversions or arbitrary expressions in patterns and does not change enum layout, validity, or exhaustiveness semantics. It is intended primarily for matching integer values from protocols, file formats, hardware interfaces, and FFI while retaining the enum variant’s semantic name.

[Rendered](https://github.com/rust-lang/rfcs/blob/13798ba8859c0fc0564d7dbc35e6deb4aa767f77/text/0000-enum-variant-cast-patterns.md)

### DIFF
--- a/text/0000-enum-variant-cast-patterns.md
+++ b/text/0000-enum-variant-cast-patterns.md
@@ -1,0 +1,470 @@
+- Feature Name: `enum_variant_cast_patterns`
+- Start Date: 2026-07-15
+- RFC PR: rust-lang/rfcs#0000
+- Rust Issue: rust-lang/rust#0000
+
+# Summary
+[summary]: #summary
+
+Permit a unit variant of a unit-only enum with an explicit primitive integer
+representation to be cast to that representation type directly in a pattern.
+
+```rust
+#[repr(u8)]
+enum Operation {
+    Banana = 1,
+    Apple = 2,
+}
+
+fn handle(value: u8) {
+    match value {
+        Operation::Banana as u8 => {}
+        Operation::Apple as u8 => {}
+        _ => {}
+    }
+}
+```
+
+This is syntactic sugar for introducing a named constant of the representation
+type and using that constant as a pattern. It does not introduce an implicit
+conversion, permit general expressions in patterns, or change enum layout,
+validity, or exhaustiveness.
+
+# Motivation
+[motivation]: #motivation
+
+Rust permits a fieldless enum value to be cast to an integer in expression
+position. Explicitly represented enums are consequently useful for describing
+integer protocols and formats while retaining meaningful names in Rust code:
+
+```rust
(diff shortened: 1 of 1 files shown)


text/0000-enum-variant-cast-patterns.md
```diff
@@ -0,0 +1,470 @@
...
+
+Rust may separately gain APIs or syntax for obtaining discriminants from enums
+with data, or derive-based integer-to-enum conversion. Those features address
+different directions of conversion and are not blocked or specified by this
+RFC.
+
+# Acknowledgements
+[acknowledgements]: #acknowledgements
+
+This RFC was developed from the public Zulip proposal
```
clarfonthey, line 469: Developed from?

You posted this proposal four hours before this RFC and it had not yet received any comments.

On its own, this is already a weird thing to say since it's *your* proposal. You wrote an RFC for your proposal; you didn't develop a proposal that was posted somewhere.

This wording immediately begs the question of whether *you* developed this RFC, or if you used a separate tool like an LLM to increase the word count on a proposal that the team already could have told you is unlikely to get accepted because of multiple reasons, including that it is possible with const patterns.

While there isn't an outstanding policy regarding LLM usage on RFCs, this is at least very presumptuous and wasteful, because we've been trying to get people to make smaller proposals before dumping large documents on the team.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rfcs/3985/f635361c8e3d927208afd6097d4fc5f30e703b43..13798ba8859c0fc0564d7dbc35e6deb4aa767f77)*

clarfonthey: Like to be clear, as someone who has in the past effectively just dumped documents on the team, I think there is a lot of value to be drawn from writing these documents even if they aren't accepted. I just think that value equation is totally different when you've just had a tool format an existing proposal into a document with a load of extra words that don't really address the main reason why this would be considered.

The goal is to make the language better and help people write good code. The code examples you gave seem like this extra syntax is not worth the complexity compared to either defining constants like you did (they can even be associated constants, for `Enum::CONST`) or adding `const {}` around the patterns for const patterns.

Instead of anyone drawing value from this discussion it just feels like a big waste of resources spent on a tool to make a document that doesn't need to be read to say there are potential issues with the proposal. And it's extra annoying that you seemingly did the right thing and proposed it more informally on Zulip first but only waited four hours before moving onto the next step.

